Issue #694, opened 2014-07-28
Support custom model mappings


c0rrzin:
I know it's in your pipeline, but I'd like to start scratching some code, because I really need it to my Project.

FYI: I already submitted a registration to your CLA.

jpsim:
We discussed this at our company all-hands week earlier this month and determined that given the wealth of available model mapping libraries, many of which are well built and actively maintained, that we won't be building model-mapping functionality into Realm itself for the foreseeable future.

Here are a few libraries you can use to do this:

**Objective-C**

* [Mantle](/Mantle/Mantle), using the `MTLModel` protocol (Mantle/Mantle#219)
* [Realm-JSON](/matthewcheok/Realm-JSON)
* [KZPropertyMapper](/krzysztofzablocki/KZPropertyMapper)

**Swift**

* [Argo](thoughtbot/argo)
* [Decodable](/Anviking/Decodable)
* [ObjectMapper](/Hearst-DD/ObjectMapper)

There are many other libraries to choose from. We hope that by investing in properly supporting these 3rd party model mapping solutions, we can focus on building features only Realm can do, while supporting these great community projects.

kylebshr:
Thanks for pointing those out — they seem like great options. I was hoping to be able to just throw dictionaries from my json into Objects init, but this feels like the better way to do it.

kylebshr:
I'm having trouble implementing Decodable with Realm — @jpsim, do you have any experience with it, or an example?

jpsim:
> Edit: the Argo link is broken if you'd like to fix that :)

Thanks, fixed.

> I'm having trouble implementing Decodable with Realm — @jpsim, do you have any experience with it, or an example?

There are a number of examples on that project's README. If you're still having trouble implementing that, please file a new issue outlining exactly what you're doing and how it's not working. Either on this repo, or Decodable's, as you deem appropriate.

Payload-Lee:
+1

marksbren:
Has anyone found a model framework that allows for Realm Notifications and works with RealmSwift. I tried ObjectMapper, but mapping objects to JSON must be done in a Realm write block ([the "inout" issue](https://github.com/realm/realm-cocoa/issues/1726)), which triggers the Realm Notification and falls into a infinite loop (mapping to JSON triggers a notification which triggers a mapping to JSON, etc.)

jpsim:
[...]

PR #698, opened 2014-07-28
Add Serialization to JSON String and Cocoa Touch Native Objects

c0rrzin:
references #694

TODO:
- [x] Change Cocoa-touch serializers' method signatures
- [x] Add test cases for circular references
- [x] Use `RLMDynamicGet()` instead of performing primitive selectors


segiddins:
@c0rrzin we're unfortunately unable to merge this pull requests.
For a feature as key as JSON/object mapping, we need to first map out the desired API specification before implementation, from the top down. We'll let you know when we have a pull request with our designs, and we welcome your feedback on it.
Thanks so much for your contribution to Realm, we really appreciate it when people take the time to contribute.

c0rrzin:
@segiddins I really enjoyed spending some time with the project and sure would like to effectively contribute and provide feedback on the design that Realm's team will come up with so it suits its needs.

PR #1026, opened 2014-10-15
Auto migration

alazier:
Changes to migration api to ensure we are always persisting the correct version. You now set a schema version and migration block and migrations are automatically applied as needed.

Also added apis for adding and deleting objects during a migration.

@tgoyne @jpsim

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,7 @@
 * Added method `createObject:withObject:` on `RLMMigration` which allows object creation during migrations.
 * Added method `deleteObject:` on `RLMMigration` which allows object deletion during migrations.
 * Updating to core library version 0.85.0.
+* Implement `objectsWhere:` and `objectsWithPredicate:` for array properties.
 
 ### Bugfixes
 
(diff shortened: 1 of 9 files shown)


rbaulin:
@jpsim @alazier I used to create isolated realms from separate libraries. How do I work with different realms since there is only one global migration block?

yoshyosh:
Hey @rbaulin there is just one global schema, so Realm will know where to apply the changes to when you execute the migration

````obj-c
[migration enumerateObjects:Person.className
                          block:^(RLMObject *oldObject, RLMObject *newObject) { 
...
````
You can see here that we are passing Person.className to the migration block regardless of which realm file Person is located in

rbaulin:
Thanks @yoshyosh, but what if vendor library also uses Realm to store data and we both call `-setVersion:migrationBlock`? Previous API allowed to configure migration individually for each realm path.

alazier:
Good point that we no longer support per Realm migration paths. I think we need to re-add support for this. We will still need to require that a user set a schema version before use though, or else a Realm could get initialized with the incorrect schema version (which was happening in the past). To support the case of vendor libraries setting conflicting versions/migrations, we will probably need to add a per-realm method which allows you to override the global settings.

alazier:
@rbaulin added an issue #1336 to address this.

PR #1042, opened 2014-10-17
Partial updates when calling createOrUpdate

### LINKED ISSUE #1029: Create or update not updating but deleting
I have an api call that returns a list of objects with unique ids but the objects themselves do not contain all the information i.e initial call returns:

```
artist
  id
  name
```

When someone clicks on the aritst to see more detail another server request is made for the specific artist id and would return:

```
artist
   id
   name
   tracks
      track_1
      track_2
      ...
```

The second request works okay but if I go to the main page again and refresh the data with a list of: 

```
artist
  id
  name
```

The tracks get deleted from the artists, any suggestions?

alazier:
Fixes #1029 

@tgoyne

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,9 @@
 * Added method `deleteObject:` on `RLMMigration` which allows object deletion during migrations.
 * Updating to core library version 0.85.0.
 * Implement `objectsWhere:` and `objectsWithPredicate:` for array properties.
+* Add `cancelWriteTransaction` to revert all changes made in a write transaction and end the transaction.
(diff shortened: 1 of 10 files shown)


tamiswa:
@alazier, I don't think this partial update mechanism works in the recursive case.  In my testing I've found that an existing object nested two or more levels deep in a JSON object still experiences the issue in #1029.  An example using that issue's use case.

    artist
      id
      name
      tracks
        id
        name
        files
          id
          location
          type

If the files property isn't there, any file objects in the realm get deleted.

jpsim:
@tamiswa we're aware of this limitation and are tracking the issue in #1362.

PR #1204, opened 2014-12-09
Code sign the browser

### LINKED ISSUE #1202: Sign Realm Browser.app
Semi-related to #680, it would be nice if `Realm Browser.app` in the binary distribution were signed - it would be a bummer if users dropped their whole system's security to run the app instead of treating it like an exception via initial launch from contextual menu.
```sh
$ codesign -dv ~/Downloads/realm-cocoa-0.88.0/browser/Realm\ Browser.app 
/Users/pizthewiz/Downloads/realm-cocoa-0.88.0/browser/Realm Browser.app: code object is not signed at all
```

alazier:
Closes #1202

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -396,7 +396,7 @@ case "$COMMAND" in
     # Browser
     ######################################
     "browser")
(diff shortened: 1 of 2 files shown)


segiddins:
Remind me, why didn't this work?

jpsim:
We couldn't get the signed browser to ever be accepted by GateKeeper.

PR #1246, opened 2014-12-19
[RLMRealm docs] Added doc overview to RLMRealm (fixes #1198)

### LINKED ISSUE #1198: No overview in the API docs for RLMRealm
The RLMRealm class does not have an overview in the appledocs. For a class with such central importance it should really have an explanation of what it does and how it is used.

jpsim:
@alazier @tgoyne anything else that should be added in there?


tgoyne:
:+1:

PR #1272, opened 2014-12-28
Feature/key path mapping transformation

nathanhjones:
Gents - this PR supersedes https://github.com/realm/realm-cocoa/pull/1074. It contains similar functionality, but I found myself needing some of the translation features that were mentioned in the comments for a project. I debated a few different patterns and settled on this. I'm obviously open to discussing things.

This PR does not solve the per-object translation. However, it sets the stage for it in a future PR. An additional initializer on `RLMObject` could be added (along with a read only property) to make this possible - e.g. `initWithTranslation:andObject:`. Currently, this would require that the instance or the translation property itself be passed to the utility function `RLMValidatedDictionaryForObjectSchema`. Noticing that you are making structural changes to that class, I think this requires more discussion before adding.

### DIFF
--- a/Realm.xcodeproj/project.pbxproj
+++ b/Realm.xcodeproj/project.pbxproj
@@ -26,6 +26,15 @@
 		02B8EF5A19E601D80045A93D /* RLMResults.h in Headers */ = {isa = PBXBuildFile; fileRef = 02B8EF5819E601D80045A93D /* RLMResults.h */; settings = {ATTRIBUTES = (Public, ); }; };
 		02B8EF5C19E7048D0045A93D /* RLMCollection.h in Headers */ = {isa = PBXBuildFile; fileRef = 02B8EF5B19E7048D0045A93D /* RLMCollection.h */; settings = {ATTRIBUTES = (Public, ); }; };
 		02B8EF5D19E7048D0045A93D /* RLMCollection.h in Headers */ = {isa = PBXBuildFile; fileRef = 02B8EF5B19E7048D0045A93D /* RLMCollection.h */; settings = {ATTRIBUTES = (Public, ); }; };
+		2F08457E1A49E79A003868BE /* RLMTransformer.h in Headers */ = {isa = PBXBuildFile; fileRef = 2F08457C1A49E79A003868BE /* RLMTransformer.h */; settings = {ATTRIBUTES = (Public, ); }; };
+		2F08457F1A49E79A003868BE /* RLMTransformer.m in Sources */ = {isa = PBXBuildFile; fileRef = 2F08457D1A49E79A003868BE /* RLMTransformer.m */; };
+		2F0845801A49EC72003868BE /* RLMTransformer.h in Headers */ = {isa = PBXBuildFile; fileRef = 2F08457C1A49E79A003868BE /* RLMTransformer.h */; settings = {ATTRIBUTES = (Public, ); }; };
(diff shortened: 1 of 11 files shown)


nathanhjones:
After digging into `RLMObject` initialization behind the scenes a bit, I've got a better understanding and agree that this over complicated and won't scale. My goal was to minimize changes to existing API and the solution suffered for that.

Yes, mapping / transformation should be decoupled from the validation code. I ran some more complex examples through this PR as-is and there are cases where it breaks down given that validation appears to happen when objects are added to a realm as well.

That said, I agree with @jpsim that mapping and transformation should be separate. I like the idea of an `RLMMapping` class that can be passed on initialization. We could then add a couple additional functions to be called in this order:

1. Convert inbound keys to expected keys - `RLMMappedValuesForObjectSchema(id value, NSDictionary *mapping, RLMObjectSchema *objectSchema)`
2. Transform values to correct type / value - `RLMTransformedValuesForObjectSchema(id value, NSDictionary *transforms, RLMObjectSchema *objectSchema)`
3. Run existing validation check `RLMValidatedDictionaryForObjectSchema(id value, RLMObjectSchema *objectSchema, RLMSchema *schema, bool allowMissing)`

@jpsim - I agree, checking for `_realmMapping` is obscure. Feels odd to manipulate an inbound API response before being able to use it. 

Not thrilled with a bunch of new methods, but I think 3 is reasonable given the functionality added. If a dev wants to specify a mapping, other than default (see below), they use the more verbose constructor. The existing methods would just call the new methods passing `nil`, so shouldn't be much overhead.

**Here are a couple more questions:**

1. Do we still provide API for specifying a default `RLMMapping` object? Personally, I like it and I think it will simplify most of the use cases. I know my use cases have very few scenarios where I need more than one mapper, allowing me to use `initWithObject:` vs a more verbose method.
2. What are your thoughts on the block transformation parameters? I could see cases where receiving all values is beneficial:
  * id(^)(id attribute) - e.g. the single mapped value like '123'
  * id(^)(id attributes) - e.g. the `NSDictionary` of all attributes

segiddins:
Some background on where my comments are coming from: I'm a maintainer of [RestKit](https://github.com/RestKit/RestKit) and have spent time developing API interaction-heavy applications.

I don't think a slim API should be the number-one goal of _any_ framework that wants to involve itself in object mapping and value transformation. Instead, I think the focus should be on making any exposed API be as expressive and flexible as possible -- people will want to do weird stuff with object mapping, and we need to allow them to do that without making deviating from 1-1 mappings be a source of frustration.

At first glance, I really like the API that JP is proposing, for several reasons:

1. It clearly separates out the steps of value transformation and object mapping
2. It allows for a flexible _implementation_
3. It allows for an interface that can _optionally_ expose more powerful constructs

How will we supporting mapping to objects by keys that are either compound or not the primary key?

Also, I think it might be cleaner if the `Mapping` object took an explicit `RLMObject` subclass, and then object creation was done through the mapping itself, which would require adding _no_ new public methods to `RLMObject`.

segiddins:
@nathanhjones thanks so much for taking the time to submit this PR, and sorry for leaving it dangling for so long. I'm going to close it now, as we're not quite ready to flesh out this design and maintain the feature at the moment. When we eventually get around to it, we'll be sure to get your feedback. Thanks so much for the PR!

nathanhjones:
No worries @segiddins - I've since focused on some other things and honestly should have closed this a while ago. I wasn't pleased with the second design and should have stuck with my original version that was part of another PR. I appreciate the :eyes: and perhaps We'll revisit in the future and I can be part of that solution...we'll see. Thanks again.

PR #1281, opened 2014-12-30
Roll back the write transaction on schema init if nothing changed

tgoyne:
Eliminates spurious change notifications when opening a Realm flle in multiple processes.

### DIFF
--- a/Realm/RLMObjectStore.mm
+++ b/Realm/RLMObjectStore.mm
@@ -113,11 +113,10 @@ static void RLMCreateColumn(RLMRealm *realm, tightdb::Table &table, RLMProperty 
     }
 }
 
-
-// Schema used to created generated accessors
-static NSMutableArray *s_accessorSchema = [NSMutableArray new];
-
 void RLMRealmCreateAccessors(RLMSchema *schema) {
+    // Schema used to created generated accessors
+    static NSMutableArray * const s_accessorSchema = [NSMutableArray new];
+
     // create accessors for non-dynamic realms
     RLMSchema *matchingSchema = nil;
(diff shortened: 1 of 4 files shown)


tgoyne:
Now does way less stuff since all the other bits have been done better in different PRs.

Realm/RLMObjectStore.mm
```diff
@@ -256,9 +209,71 @@ void RLMRealmCreateTables(RLMRealm *realm, RLMSchema *targetSchema, bool updateE
...
+        // apply migration block if provided
+        if (migrating && migrationBlock) {
+            NSError *error = migrationBlock();
+            if (error) {
+                [realm cancelWriteTransaction];
+                return error;
+            }
+        }
+
+        if (oldVersion != newVersion) {
```
alazier: This is the same as ```migrating```. This should be combined with the previous block, with ```if (migrationBlock)``` in a nested code block.

Realm/RLMObjectStore.mm
```diff
@@ -256,9 +209,71 @@ void RLMRealmCreateTables(RLMRealm *realm, RLMSchema *targetSchema, bool updateE
...
+    @catch (NSException *) {
+        [realm cancelWriteTransaction];
+        @throw;
+    }
+
+    return nil;
 }
 
 
+void RLMRealmSetSchema(RLMRealm *realm, RLMSchema *targetSchema, bool verify) {
```
alazier: Can you move this up so it isn't changed in the diff since there are no code changes?

alazier:
Other than the code factoring issues :+1:

PR #1284, opened 2014-12-31
Additions/Refactor of Realm.framework to support new swift apis

alazier:
This moves over to RLMObjectBase. We also now properly hide private apis using bridging headers.

@jpsim @tgoyne


jpsim:
:+1: nice work

Realm.xcodeproj/project.pbxproj
```diff
@@ -1666,7 +1684,7 @@
 					"$(inherited)",
 				);
 				GCC_PREPROCESSOR_DEFINITIONS = (
-					"DEBUG=1",
+					"REALM_SWIFT=1",
```
tgoyne: This define does not appear to be used anywhere.

tgoyne:
Have you actually tested the dynamic framework with these changes? I am unable to get it to work from either obj-c or swift.

Realm/RLMObjectSchema_Private.hpp
```diff
@@ -22,7 +22,7 @@
 
 // RLMObjectSchema private
 @interface RLMObjectSchema () {
-    @public
+    @private
     // table accessor optimization
     tightdb::TableRef _table;
```
tgoyne: Should move this to RLMObjectSchema.mm now that it doesn't need to be public.

Realm/RLMUtil.mm
```diff
@@ -196,12 +196,16 @@ id RLMValidatedObjectForProperty(id obj, RLMProperty *prop, RLMSchema *schema) {
...
     NSMutableDictionary *outDict = [NSMutableDictionary dictionaryWithCapacity:properties.count];
+    NSDictionary *defaultValues = nil;
     for (RLMProperty *prop in properties) {
         id obj = [value valueForKey:prop.name];
 
         // get default for nil object
         if (!obj && !allowMissing) {
-            obj = objectSchema.defaultValues[prop.name];
+            if (!defaultValues) {
+                defaultValues = RLMDefaultValuesForObjectSchema(objectSchema);
```
tgoyne: `RLMDefaultValuesForObjectSchema` needs to not be so slow if it's going to be called for every object that needs a default value. At the minimum it needs to not make a mutable copy of the dictionary for obj-c objects, and the whole thing should probably be refactored to just grab the values from the swift object directly rather than converting the object to a dictionary.

alazier:
This has been tested against the objc and swift examples

tgoyne:
:+1:

PR #1287, opened 2015-01-01
Improve the exception message when comparing a linked property to another property (unsupported)

### LINKED ISSUE #1286: Support comparing two linked keypaths in a query
See https://groups.google.com/d/msg/realm-cocoa/WRI5oq1Y6zU/sk9-qVIYdjcJ.

jpsim:
Fixes #1286.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,8 @@ x.x.x Release notes (yyyy-MM-dd)
 
 ### Enhancements
 
-* None.
+* Improve the exception message when comparing a linked property to another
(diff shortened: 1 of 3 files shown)


jpsim:
Previous message would read "Column name objectCol.stringCol not found in table". Please review @alazier @tgoyne

tgoyne:
Looks like core does actually support this and it shouldn't be too hard to fix the query building logic.

jpsim:
Closing since https://github.com/realm/realm-core/pull/621 will add support for this.

PR #1292, opened 2015-01-02
Update to core 0.87.04

tgoyne:


alazier:
Did ci actually run before we merged this?

tgoyne:
Yes.

PR #1293, opened 2015-01-02
Roll back changes made when an exception is thrown during a migration.

tgoyne:
Previously it just left things sorta half-migrated.

@alazier


alazier:
:+1:

PR #1296, opened 2015-01-05
Add a check for the count changing during RLMResults fast enumeration

tgoyne:
Without this check we now get an assertion failure when the count goes down, as we'd read past the end of the TableView.

@alazier

### DIFF
--- a/Realm/RLMResults.mm
+++ b/Realm/RLMResults.mm
@@ -144,6 +144,8 @@ static inline void RLMResultsValidateInWriteTransaction(__unsafe_unretained RLMR
         state->extra[1] = self.count;
     }
     else {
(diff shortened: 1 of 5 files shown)


Realm/RLMResults.mm
```diff
@@ -144,6 +144,11 @@ - (NSUInteger)countByEnumeratingWithState:(NSFastEnumerationState *)state
         state->extra[1] = self.count;
     }
     else {
+        if (state->extra[1] != self.count) {
```
alazier: Please add a FIXME to add more comprehensive checking in the future.

alazier:
Should we check on RLMArray as well?

tgoyne:
Added the same check for RLMArray.

alazier:
:+1:

PR #1299, opened 2015-01-06
Add writeEncryptedCopyToPath:key:error:

tgoyne:
This makes it possible to change the encryption key of an existing Realm file as follows:

    // RLMRealm instances are autoreleased and we must ensure the old file is
    // closed before we try to overwrite it
    @autoreleasepool {
        // Write a copy with the new encryption key
        [RLMRealm.defaultRealm writeEncryptedCopyToPath:@"/tmp/out.realm"
                                                    key:newKey
                                                  error:&error];
    }
    // Overwrite the original file with the new copy
    [NSFileManager.defaultManager moveItemAtPath:@"/tmp/out.realm"
                                          toPath:RLMRealm.defaultRealmPath
                                           error:&error];
    // Update the encryption key to use for the default Realm (if using the key caching)
    [RLMRealm setEncryptionKey:newKey forRealmsAtPath:RLMRealm.defaultRealmPath];

    // Open the new file

@alazier

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@ x.x.x Release notes (yyyy-MM-dd)
 
 ### Enhancements
 
-* None.
+* Add ``-[RLMRealm writeEncryptedCopyToPath:key:error:]``.
 
 ### Bugfixes
 
--- a/Realm/RLMRealm.h
+++ b/Realm/RLMRealm.h
@@ -353,6 +353,21 @@ typedef void(^RLMNotificationBlock)(NSString *notification, RLMRealm *realm);
 - (BOOL)writeCopyToPath:(NSString *)path error:(NSError **)error;
 
 /**
+ Write an encrypted and compacted copy of the RLMRealm to the given path.
+
+ The destination file cannot already exist.
+
+ Note that if this is called from within a write transaction it writes the
+ *current* data, and not data when the last write transaction was committed.
+
+ @param path Path to save the Realm to.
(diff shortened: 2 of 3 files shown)


alazier:
I didn't realize you could put emoji in commit messages.

alazier:
We should probably leave the old method which doesn't take a key for the majority of encryption users. Right now it looks like you break everyone's code who uses the existing method.

Might want to have a test for if you try to make a copy to an existing path.

tgoyne:
The old method does still exist.

astigsen:
We should also add support for opening encrypted realms in the browser.

alazier:
@astigsen I'm not sure that is related to this pr. That being said we should still do this, so opened a new issue #1300

alazier:
I was confused by the diff. I see now the old method is still there. 

I know we are following convention with our naming (using the term ```encryptedRealm```, but looking back at our other apis I think I might prefer having the api names be the same with the addition of an parameter ```encryptionKey``` rather than key. Did we ever consider this naming? If we are to ever change it now would probably be the time before adding more methods following the existing convention. Otherwise :+1:

tgoyne:
That was what I originally had when I first added the encryption stuff, and someone objected to it and wanted the encryptedRealm naming.

alazier:
I think this naming is pretty bad - this sounds worse than the other methods to me. Lets discuss renaming before pushing a release. Do you know who had the previous objections? Hopefully it wasn't me.

PR #1301, opened 2015-01-06
Actually encrypt the write logs

tgoyne:
I added support for this to core and then forgot to make the binding actually pass the key to the function...

@alazier


alazier:
:+1:  once ci passes

PR #1308, opened 2015-01-08
Rename methods for encrypted realms

### LINKED ISSUE #1307: Rename api methods which take encryption keys
We kind of depart from convention with our naming for methods which take an encryption key. For instance. Instead of using the phrase ```encryptedRealm```, we should just rename the ```key``` parameter to ```encryptionKey```. We should change the following methods from:
```objc
+ (instancetype)encryptedRealmWithPath:(NSString *)path
                                   key:(NSData *)key
                              readOnly:(BOOL)readonly
                                 error:(NSError **)error;
- (BOOL)writeEncryptedCopyToPath:(NSString *)path key:(NSData *)key error:(NSError **)error;
```
to
```objc
+ (instancetype)realmWithPath:(NSString *)path encryptionKey:(NSData *)key readOnly:(BOOL)readonly error:(NSError **)error;
- (BOOL)writeCopyToPath:(NSString *)path encryptionKey:(NSData *)key error:(NSError **)error;
```
which aligns better with the other method names and with the ```+ (void)setEncryptionKey:(NSData *)key forRealmsAtPath:(NSString *)path;``` method.

@tgoyne mentioned that someone objected to this naming in the past. Does this objection still exist? Thoughts @jpsim @segiddins ?

tgoyne:
Closes #1307.

### DIFF
--- a/Realm/RLMRealm.h
+++ b/Realm/RLMRealm.h
@@ -113,10 +113,10 @@
 
  @return An encrypted `RLMRealm` instance.
  */
-+ (instancetype)encryptedRealmWithPath:(NSString *)path
-                                   key:(NSData *)key
-                              readOnly:(BOOL)readonly
-                                 error:(NSError **)error;
++ (instancetype)realmWithPath:(NSString *)path
+                encryptionKey:(NSData *)key
+                     readOnly:(BOOL)readonly
+                        error:(NSError **)error;
 
 /**
  Set the encryption key to use when opening Realms at a certain path.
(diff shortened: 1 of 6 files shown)


Realm/RLMRealm.h
```diff
@@ -541,7 +541,7 @@ typedef void (^RLMMigrationBlock)(RLMMigration *migration, NSUInteger oldSchemaV
  @param key         64-byte encryption key.
  @return            The error that occurred while applying the migration, if any.
  */
-+ (NSError *)migrateEncryptedRealmAtPath:(NSString *)realmPath key:(NSData *)key;
++ (NSError *)migratedRealmAtPath:(NSString *)realmPath encryptionKey:(NSData *)key;
```
alazier: should be migrateRealmAtPath

alazier:
Other than the typo :+1:

PR #1309, opened 2015-01-09
Eliminate duplicated copies of the header files

### LINKED ISSUE #1303: Duplicate interface definition for class 'RLMResults', compile error when installing Realm via CocoaPods
When installing Realm via CocoaPods for iOS and MacOSX targets within the same Xcode project, CocoaPods is creating duplicate references for each Realm header file when creating the Pods project.

This issue generates compiler errors "Duplicate interface definition for class ...".

A simple Podfile such as this one, will demonstrate the issue:

```
source 'https://github.com/CocoaPods/Specs.git'

target 'target-osx' do
    platform :osx, '10.9'
    pod 'Realm', '~> 0.87'
end

target 'target-ios' do
    platform :ios, '7.0'
    pod 'Realm', '~> 0.87'
end
```

The best explanation for this issue that I could find appears to be caused by an issue in the PodSpec file defining the Pod, and not an issue with CocoaPods itself: [Duplicate header references in build step #2565](https://github.com/CocoaPods/CocoaPods/issues/2565). We are using CocoaPods version 0.35.0, if that makes any difference.

tgoyne:
The podspec was creating two full copies of the includes for the sake of RLMPlatform.h, but the platform checking is unnecessary with CocoaPods, and having duplicated headers breaks #import's duplicate checking when header maps are enabled (which CocoaPods requires) and both the OS X and iOS versions are used.

Closes #1303.

### DIFF
--- a/Realm.podspec
+++ b/Realm.podspec
@@ -23,24 +23,20 @@ Pod::Spec.new do |s|
 
   s.compiler_flags          = "-DTIGHTDB_HAVE_CONFIG -DREALM_SWIFT=0 -DREALM_VERSION='@\"#{s.version}\"'"
   s.prepare_command         = 'sh build.sh cocoapods-setup'
-  s.private_header_files    = 'include-ios/**/*.hpp', 'include-osx/**/*.hpp', '**/*_Private.h'
-  s.source_files            = 'Realm/*.{m,mm}', 'core/**/*.{h,hpp}'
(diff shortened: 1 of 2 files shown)


segiddins:
Does `pod lib lint` pass?

tgoyne:
Does `pod lib lint` test anything useful yet?

segiddins:
http://guides.cocoapods.org/terminal/commands.html#pod_lib_lint

segiddins:
:+1:

PR #1326, opened 2015-01-12
Fix crash when changing or removing primary key from table

kwojtaszek:


alazier:
Do you have a repro case for the crash you are experiencing?

jpsim:
Hi @kwojtaszek from your change, it looks as if you found a scenario in which `objectSchema.className` returns a `Class` instead of an `NSString`. Could you share when that occurs? Ideally by writing a failing test.

From what our continuous integration logs, your change makes our existing unit tests fail, so if this change fixes a specific scenario for you, it also makes the scenarios we're testing fail.

alazier:
We have another pr here #1335 which fixes a similar if not the same issue. Does this fix your issue?

alazier:
This should hopefully be fixed in master after merging #1355

PR #1330, opened 2015-01-12
Clean up header inclusions

tgoyne:
Pull almost all #imports out of header files, eliminate some unused header inclusions, sort imports, and clean up some extern "C" stuff.

@alazier


alazier:
:+1:

PR #1332, opened 2015-01-12
Cache dynamic realms

tgoyne:
Mostly just to make dynamic realms more like normal realms, but also lets us check for a few more invalid things.

Test changes are all just to avoid having the same file open multiple times with different schemas on the same thread.

@alazier


Realm/RLMRealm.mm
```diff
@@ -457,11 +458,12 @@ + (instancetype)realmWithPath:(NSString *)path
                                            reason:@"Realm at path already opened with different inMemory settings"
                                          userInfo:@{@"path":realm.path}];
         }
-        if (realm->_dynamic != dynamic) {
```
alazier: Add a new if statement and proper exception message.

PR #1333, opened 2015-01-13
Update iOS examples for the Swift API

segiddins:
Also allows encryption in Swift

n.b. Migrations aren't fully functional in swift, as demonstrated when running the Migrations example app.

\c @alazier @jpsim

### DIFF
--- a/Realm/RealmSwift/Object.swift
+++ b/Realm/RealmSwift/Object.swift
@@ -62,6 +62,10 @@ public class Object : RLMObjectBase, Equatable {
         super.init()
     }
 
+    public class func createInDefaultRealmWithObject(object: AnyObject) -> Self {
+        return unsafeBitCast(RLMCreateObjectInRealmWithValue(defaultRealm().rlmRealm, className(), object, .allZeros), self)
+    }
+
     public class func createInRealm(realm: Realm, withObject object: AnyObject) -> Self {
         return unsafeBitCast(RLMCreateObjectInRealmWithValue(realm.rlmRealm, className(), object, .allZeros), self)
     }
--- a/Realm/RealmSwift/Object.swift
+++ b/Realm/RealmSwift/Object.swift
@@ -62,10 +62,6 @@ public class Object : RLMObjectBase, Equatable {
         super.init()
     }
 
-    public class func createInDefaultRealmWithObject(object: AnyObject) -> Self {
-        return unsafeBitCast(RLMCreateObjectInRealmWithValue(defaultRealm().rlmRealm, className(), object, .allZeros), self)
-    }
-
     public class func createInRealm(realm: Realm, withObject object: AnyObject) -> Self {
         return unsafeBitCast(RLMCreateObjectInRealmWithValue(realm.rlmRealm, className(), object, .allZeros), self)
     }
--- a/Realm/RealmSwift/Realm.swift
+++ b/Realm/RealmSwift/Realm.swift
@@ -35,6 +35,14 @@ public func defaultRealm() -> Realm {
     return Realm(rlmRealm: RLMRealm.defaultRealm())
 }
 
+public func encryptedRealm(path: String, encryptionKey: NSData, readOnly: Bool, error: NSErrorPointer) -> Realm? {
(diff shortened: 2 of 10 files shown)


Realm/RealmSwift/Object.swift
```diff
@@ -62,6 +62,10 @@ public class Object : RLMObjectBase, Equatable {
         super.init()
     }
 
+    public class func createInDefaultRealmWithObject(object: AnyObject) -> Self {
```
jpsim: I think I had purposely left this out because it doesn't significantly simplify the general case in Swift:

```swift
MyObject.createInDefaultRealmWithObject(obj)
MyObject.createInRealm(defaultRealm(), withObject: obj)
```

segiddins: OK, I can remove it.

Realm/RealmSwift/Realm.swift
```diff
@@ -58,15 +58,31 @@ public class Realm {
...
     }
 
-    public convenience init(path: String, readOnly readonly: Bool, error: NSErrorPointer) {
-        self.init(rlmRealm: RLMRealm(path: path, readOnly: readonly, error: error))
+    public convenience init?(path: String, readOnly readonly: Bool, error: NSErrorPointer) {
+        if let rlmRealm = RLMRealm(path: path, readOnly: readonly, error: error) as RLMRealm? {
+            self.init(rlmRealm: rlmRealm)
+        }
+        else {
+            self.init(rlmRealm: RLMRealm.inMemoryRealmWithIdentifier(NSUUID().UUIDString))
```
jpsim: We should create a no-op function that "returns" a nil RLMRealm in Objective-C and call that instead of creating an in-memory realm. The interface for this can be in `Realm.Private`.

jpsim: Minor point: in the rest of the Swift code, we have the else on the same line as the closing brace.

segiddins: Returning `nil` won't be of any help, since that'd cause a crash.

jpsim: Obviously we can't return nil in the no-op function I mentioned earlier, since that gets us right back to where we were before, but we should still find a minimal way to please the compiler, without having to actually create a legitimate RLMRealm.

examples/ios/swift/Encryption/ViewController.swift
```diff
@@ -68,16 +69,18 @@ class ViewController: UIViewController {
...
         // Reopening with the correct key works and can read the data
         autoreleasepool {
-            let realm = RLMRealm.encryptedRealmWithPath(RLMRealm.defaultRealmPath(),
-                key: self.getKey(), readOnly: false, error: nil)
-
-            self.log("Saved object: \((EncryptionObject.allObjectsInRealm(realm).firstObject()! as EncryptionObject).stringProp)")
+            var error: NSError? = nil
+            if let realm = Realm(path: defaultRealmPath(),
+                encryptionKey: self.getKey(),
+                readOnly: false, error: &error) {
```
jpsim: If we're to have params on new lines, we should put `error` on a new line.

examples/ios/swift/RealmExamples.xcodeproj/project.pbxproj
```diff
@@ -746,7 +811,11 @@
 			buildSettings = {
 				ASSETCATALOG_COMPILER_APPICON_NAME = AppIcon;
 				ASSETCATALOG_COMPILER_LAUNCHIMAGE_NAME = LaunchImage;
-				FRAMEWORK_SEARCH_PATHS = "$(inherited)";
+				FRAMEWORK_SEARCH_PATHS = (
+					"$(inherited)",
+					"/Users/segiddins/Development/Realm/realm-cocoa/build/ios",
```
jpsim: There are a number of absolute paths in this file which should use relative paths.

jpsim: The derived data paths should also be unnecessary.

alazier:
:+1: assuming the examples run on the device

PR #1335, opened 2015-01-13
Refactor schema init and fix a crash

tgoyne:
Pull more of the schema version checking and updating stuff into one place to hopefully make it easier to follow and make explicit and implicit migrations do the exact same thing. Fixes a crash due to RLMRealmCreateMetadataTables not being called when implicitly migrating an existing realm file from being we unconditionally created the primary key table.

This also makes creating a brand new Realm file with a custom schema not set the schema version to the static schema version, since doing so didn't really make any sense.

@alazier


alazier:
Superseded/included in #1355

PR #1337, opened 2015-01-14
Bring Swift public interface to parity with Objective-C

segiddins:
\c @alazier @jpsim

### DIFF
--- a/Realm/RealmSwift/Realm.swift
+++ b/Realm/RealmSwift/Realm.swift
@@ -89,6 +89,10 @@ public class Realm {
         rlmRealm.writeCopyToPath(path, error: error)
     }
 
+    public func writeCopyToPath(path: String, encryptionKey: NSData, error: NSErrorPointer) {
+        rlmRealm.writeEncryptedCopyToPath(path, key: encryptionKey, error: error)
+    }
+
     // MARK: Transactions
 
     public func write(block: (() -> Void)) {
--- a/Realm/RealmSwift/Realm.swift
+++ b/Realm/RealmSwift/Realm.swift
@@ -83,6 +83,12 @@ public class Realm {
         }
     }
 
+    // MARK: Writing a Copy
+
+    public func writeCopyToPath(path: String, error: NSErrorPointer) {
+        rlmRealm.writeCopyToPath(path, error: error)
+    }
+
     // MARK: Transactions
 
     public func write(block: (() -> Void)) {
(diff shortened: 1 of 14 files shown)


Realm/RealmSwift/Realm.swift
```diff
@@ -83,6 +83,12 @@ public class Realm {
         }
     }
 
+    // MARK: Writing a Copy
+
+    public func writeCopyToPath(path: String, error: NSErrorPointer) {
```
alazier: add encryption overload

alazier:
:+1: after you add missing method

jpsim:
In this PR, could you update all the cases where we take an `NSErrorPointer` to have a default value of `nil`? e.g. `error: NSErrorPointer = nil`. This will allow users to omit the error if they don't care.

Realm/RealmSwift/Realm.swift
```diff
@@ -83,6 +83,16 @@ public class Realm {
...
+        rlmRealm.writeCopyToPath(path, error: error)
+    }
+
+    public func writeCopyToPath(path: String, encryptionKey: NSData, error: NSErrorPointer) {
```
jpsim: you should leverage Swift's power and combine `writeCopyToPath(path: String, encryptionKey: NSData, error: NSErrorPointer)` and `writeCopyToPath(path: String, error: NSErrorPointer)` into `writeCopyToPath(path: String, error: NSErrorPointer = nil, encryptionKey: NSData? = nil)`.

This will mean we only have a single copy function declared and users can call as `writeCopyToPath("path.realm")` or `writeCopyToPath("path.realm", error: &error)` or `writeCopyToPath("path.realm", error: &error, encryptionKey: key)`.

alazier: Can we still have the out error as the last argument?

jpsim: That would mean you'd have to explicitly pass in a `nil` encryption key when copying unencrypted realms.

We could have two methods:

```swift
writeCopyToPath(path: String, error: NSErrorPointer = nil)
writeCopyToPath(path: String, encryptionKey: NSData, error: NSErrorPointer = nil)
```

segiddins: which is what we have now

jpsim:
I see now that you've already merged this @segiddins. Could you make my suggested changes in another PR?

segiddins:
@jpsim already opened https://github.com/realm/realm-cocoa/pull/1338

Realm/RealmSwift/Realm.swift
jpsim: with the important difference being the default error values.

segiddins: my other PR adds the default error values

PR #1338, opened 2015-01-14
[Realm] Default NSErrorPointer parameters to nil

segiddins:
\c @jpsim


alazier:
jp suggested in the other pr of only having one copy of these methods.

jpsim:
:+1:

PR #1340, opened 2015-01-14
Support per-Realm migration blocks and schema versions per Realm #1336

### LINKED ISSUE #1336: Support per-Realm migration blocks and schema versions per Realm
Global configuration of migration blocks and version does not work when using third party libraries also using Realm. We need to support version/migration configuration on a per-Realm basis to address this.

rbaulin:
Add api to perform file-based realm migrations:

    + (void)setSchemaVersion:(NSUInteger)version withMigrationBlock:(RLMMigrationBlock)block realmPath:(NSString *)path;

This is required when working with realms in separate environment. All tests passes, example updated. Pull introduces minimal changes to public api.

Closes #1336.


alazier:
Closing this as this issue is also fixed in #1355

PR #1343, opened 2015-01-15
[Examples/Obj-C] Fixed date format and set time zone offset such that the JSON strings are now parsed correctly.

danielpovlsen:
The format YYYY is used for Year (in "Week of Year" based calendars), whereas yyyy is the "normal" calendar year ([doc](https://developer.apple.com/library/mac/documentation/Cocoa/Conceptual/DataFormatting/Articles/dfDateFormatting10_4.html#//apple_ref/doc/uid/TP40002369-SW13)). Along with setting the time zone offset to 0, the example now correctly parses "September 23, 1926" to an NSDate of 1926-09-23 00:00:00 +0000, instead of 1925-12-19 23:00:00 +0000.

### DIFF
--- a/examples/osx/objc/JSONImport/main.m
+++ b/examples/osx/objc/JSONImport/main.m
@@ -37,7 +37,8 @@ int main(int argc, const char * argv[])
 
         NSDateFormatter *dateFormatter = [[NSDateFormatter alloc] init];
         dateFormatter.locale = [NSLocale localeWithLocaleIdentifier:@"en_US_POSIX"];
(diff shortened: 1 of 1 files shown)


tgoyne:
Changes look good to me, but we do need a [CLA](https://docs.google.com/a/realm.io/forms/d/1bVp-Wp5nmNFz9Nx-ngTmYBVWVdwTyKj4T0WtfVm0Ozs/viewform?fbzx=4154977190905366979) even for things like this.

danielpovlsen:
Yes, of course. Done.

tgoyne:
Thanks!

PR #1345, opened 2015-01-15
Remove the NSData size checks

tgoyne:
Core checks the sizes now, so just translate the exceptions to obj-c exceptions.


jpsim:
yuuuuup

alazier:
Do we have tests for each of these cases? If so :+1:

tgoyne:
Only for the NSData case, since we weren't previously checking for overly-long strings.

Issue #935, opened 2014-09-26
Replace object attributes with single method to specify indexed columns


yoshyosh:
```Objective-C
+ (RLMPropertyAttributes)attributesForProperty:(NSString *)propertyName {
    RLMPropertyAttributes attributes = [super attributesForProperty:propertyName];
    if ([propertyName isEqualToString:@"word"]) {
        attributes |= RLMPropertyAttributeIndexed;
    }
    return attributes;
}
```

Feels very verbose and confusing, I also feel like it could be improved in the docs/examples. 

Here's the way you add primary keys:
```Objective-C
+ (NSString *)primaryKey {
    return @"id";
}
```

Could we just do this:
```Objective-C
+ (NSArray *)indexedProperties {
    return @["name", "id"];
}
```

PR #1346, opened 2015-01-15
Allow indexing ints and index int primary keys

tgoyne:
The remaining core issues appear to have been solved.

@alazier @jpsim

### DIFF
--- a/Realm/RLMMigration.mm
+++ b/Realm/RLMMigration.mm
@@ -100,21 +100,22 @@
         RLMProperty *primaryProperty = objectSchema.primaryKeyProperty;
         RLMProperty *oldPrimaryProperty = [[_oldRealm.schema schemaForClassName:objectSchema.className] primaryKeyProperty];
         if (primaryProperty && primaryProperty != oldPrimaryProperty) {
-            // FIXME: replace with count of distinct once we support indexing
-
             // FIXME: support other types
             tightdb::Table *table = objectSchema.table;
+
+            if (!table->has_search_index(primaryProperty.column)) {
+                table->add_search_index(primaryProperty.column);
+            }
+
             NSUInteger count = table->size();
             if (primaryProperty.type == RLMPropertyTypeString) {
-                if (!table->has_search_index(primaryProperty.column)) {
-                    table->add_search_index(primaryProperty.column);
-                }
                 if (table->get_distinct_view(primaryProperty.column).size() != count) {
                     NSString *reason = [NSString stringWithFormat:@"Primary key property '%@' has duplicate values after migration.", primaryProperty.name];
                     @throw [NSException exceptionWithName:@"RLMException" reason:reason userInfo:nil];
                 }
             }
             else {
+                // FIXME: replace with count of distinct once that's supported for more than string
                 for (NSUInteger i = 0; i < count; i++) {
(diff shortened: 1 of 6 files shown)


Realm/Tests/MigrationTests.mm
```diff
@@ -307,6 +307,25 @@ - (void)testStringPrimaryKeyNoIndexMigration {
     [RLMRealm migrateRealmAtPath:RLMTestRealmPath()];
 }
 
+- (void)testIntPrimaryKeyNoIndexMigration {
+    // make string an int
+    RLMObjectSchema *objectSchema = [RLMObjectSchema schemaForObjectClass:MigrationPrimaryKeyObject.class];
+
+    // create without search index
+    objectSchema.primaryKeyProperty.attributes = 0;
```
jpsim: use enum here instead of `0`?

tgoyne: There is no enum value for nothing.

jpsim:
Hurray! :+1:

Realm/Tests/MigrationTests.mm
jpsim: Ah, that's right. well that's the best we can do until we move away from `RLMPropertyAttributes` altogether (#935).

alazier: ```RLMPropertyAttributes``` is gone on al-swift. Once we finish up with that we are good to go. Could also just merge those changes to master if we want them out sooner but doesn't seem worth the effort yet.

Realm/RLMMigration.mm
```diff
@@ -100,21 +100,22 @@ - (void)verifyPrimaryKeyUniqueness {
...
-                if (!table->has_search_index(primaryProperty.column)) {
-                    table->add_search_index(primaryProperty.column);
-                }
                 if (table->get_distinct_view(primaryProperty.column).size() != count) {
                     NSString *reason = [NSString stringWithFormat:@"Primary key property '%@' has duplicate values after migration.", primaryProperty.name];
                     @throw [NSException exceptionWithName:@"RLMException" reason:reason userInfo:nil];
                 }
             }
             else {
+                // FIXME: replace with count of distinct once that's supported for more than string
```
alazier: You tried this for int and it didn't work?

Realm/Tests/MigrationTests.mm
```diff
@@ -307,6 +307,38 @@ - (void)testStringPrimaryKeyNoIndexMigration {
...
+
+    // create realm with old schema and populate
+    @autoreleasepool {
+        RLMRealm *realm = [self realmWithSingleObject:objectSchema];
+        [realm beginWriteTransaction];
+        [realm createObject:MigrationPrimaryKeyObject.className withObject:@[@1]];
+        [realm createObject:MigrationPrimaryKeyObject.className withObject:@[@2]];
+        [realm commitWriteTransaction];
+
+        XCTAssertFalse(realm.schema[MigrationPrimaryKeyObject.className].table->has_search_index(0));
```
alazier: I would prefer the pattern where we have a block and XCTAssertThrows. Every time I see this pattern it takes me a minute to figure out why we XCTAssertFalse which is annoying.

alazier:
:+1:

Realm/RLMMigration.mm
tgoyne: It was implemented after the most recent core release.

PR #1347, opened 2015-01-15
fix build.sh clean

jpsim:
Clean didn't actually delete `build` directories. @tgoyne @alazier

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -191,7 +191,7 @@ case "$COMMAND" in
     # Clean
     ######################################
     "clean")
-        find . -type d -name build -exec rm -r "{}" \;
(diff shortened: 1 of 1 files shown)


build.sh
```diff
@@ -191,7 +191,7 @@ case "$COMMAND" in
     # Clean
     ######################################
     "clean")
-        find . -type d -name build -ls -delete
+        find . -type d -name build -exec rm -r "{}" \;
```
tgoyne: `rm -r "{}" +` is a little faster (passes as many filenames as possible to a single invocation of rm).

tgoyne:
:+1:

PR #1352, opened 2015-01-16
Improve error reporting for incompatible lock files

tgoyne:
Adds a better error message that tries to explain what's wrong.

@alazier @jpsim

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,6 +30,8 @@ x.x.x Release notes (yyyy-MM-dd)
   write transaction with changes were committed without the main thread
   RLMRealm getting a chance to refresh.
 * Fixed incomplete results when querying for non-null relationships.
+* Improve the error message when a Realm file is opened in multiple processes
+  at once.
 
 0.89.2 Release notes (2015-01-02)
 =============================================================
--- a/Realm/RLMRealm.mm
+++ b/Realm/RLMRealm.mm
@@ -76,11 +76,11 @@ static __attribute__((noreturn)) void throw_objc_exception(exception &ex) {
 }
 
 // create NSError from c++ exception
-static NSError *make_realm_error(RLMError code, exception &ex) {
-    NSMutableDictionary *details = [NSMutableDictionary dictionary];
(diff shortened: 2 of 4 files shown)


Realm/RLMRealm.mm
```diff
@@ -302,20 +302,32 @@ - (instancetype)initWithPath:(NSString *)path key:(NSData *)key readOnly:(BOOL)r
...
             error = make_realm_error(RLMErrorFileExists, ex);
         }
-        catch (File::AccessError &ex) {
+        catch (File::AccessError const& ex) {
             error = make_realm_error(RLMErrorFileAccessError, ex);
         }
-        catch (exception &ex) {
+        catch (IncompatibleLockFile const&) {
+            NSString *err = @"Realm file is currently open in another process "
+                             "which cannot shared access with this process. All "
```
jpsim: s/shared/share/

jpsim:
Should we be displaying an error alert in the browser as well?

tgoyne:
Browser really should be displaying all errors when opening realms...

Should be easy to add.

jpsim:
Tracking that in #1354.

jpsim:
:+1:

alazier:
Can we add a test for this?

tgoyne:
No, it requires multiple processes of different architectures.

alazier:
:+1:

alazier:
We could generate lock files from another architecture, and then spawn a process which opens these files. Then it would appear that we are accessing lock files created from another architecture.

tgoyne:
Added a test which mangles the lock file directly.

CHANGELOG.md
```diff
@@ -27,6 +27,8 @@ x.x.x Release notes (yyyy-MM-dd)
 * Fixed a crash when adding primary keys to older realm files with no primary
   keys on any objects.
 * Fixed a crash when removing a primary key in a migration.
+* Improve the erorr message when a Realm file is opened in multiple processes
```
jpsim: error

tgoyne:
Result of opening a file in the browser that's already open in a 32-bit sim now that the browser actually displays errors:

image

Issue #1160, opened 2014-11-28
Skip sorting when calling count on an RLMResults that hasn't been constructed yet


kunalsood:
Continuing the discussion from [this thread] (https://groups.google.com/forum/#!topic/realm-cocoa/3zWhXrCcIic) on the mailing list, I was made aware of how the count property on RLMResults works.

From doing a little more benchmarking, it seems to me that sorting my large query by a single NSDate property is the most expensive operation being done by Realm. Further, it also looks like, Realm is doing the sorting even for count-only queries. I came to this conclusion by testing the following 2 queries.

    [[[Article allObjectsInRealm:realm] sortedResultsUsingProperty:@"date" ascending:YES] count];
    [[Article allObjectsInRealm:realm] count];

The difference was significant. So, if this is indeed the case, do you guys think, not doing sorting on count-only queries is something Realm should handle internally?

segiddins:
It appears that this is already done? (see a03f0c0b8f97)

alazier:
@segiddins that optimization was made before this issue was opened. Calling ```sortedResultsUsingProperty``` requires a tableView to exist which is expensive. Only calling ```count``` doesn't require tableView construction and is usually much faster. It seems that @kunalsood understands this based on his original post (that calling count alone on RLMResults is much faster than calling count on a sorted RLMResults).

I think what is being asking for here is to do tableView creation lazily when calling ```sortedResultsUsingProperty```. This is possible but would require a bit refactoring and storing the sort state both in the binding and in core which is something that would be good to avoid.

jpsim:
The last time we spoke about this, we decided to add two methods to improve transparency and performance: `-[RLMResults evaluate]` and `-[RLMResults isEvaluated]`. With these methods, calling `-[RLMResults count]` would never create the tableView and users would have more control over when table views are actually constructed.

Should we create a GitHub issue to implement this, @alazier @tgoyne?

tgoyne:
`count` already never creates the tableview. Adding `evaluate` is unrelated to this.

alazier:
But sorting does create the tableView. Don't think it makes too much sense to change the current behavior.

kunalsood:
Slightly related question: When a sorted RLMResults which is being used to populate the user interface posts a change notification, does that cause the already constructed tableView in the core to be re-constructed if the UI tries to access some new/changed properties in it?

[...]

PR #1353, opened 2015-01-16
Use a Table-backed RLMResults for allObjects

### LINKED ISSUE #1178: allObjects should return RLMArray backed by a Table
This will improve performance considerably, especially when processing notifications/updates for query-less accessors.

tgoyne:
Closes #1178.

@alazier @jpsim


Realm/Tests/ArrayPropertyTests.m
```diff
@@ -334,4 +333,105 @@ - (void)testSortByMultipleColumns {
...
+    XCTAssertThrowsSpecificNamed([peopleInCompany removeObjectAtIndex:3], NSException, @"RLMException", @"Out of bounds");
+    XCTAssertNoThrow([peopleInCompany removeObjectAtIndex:1], @"Should delete link to employee");
+    [realm commitWriteTransaction];
+
+    XCTAssertEqual(peopleInCompany.count, 2U, @"link deleted when accessing via links");
+    EmployeeObject *test = peopleInCompany[0];
+    XCTAssertEqual(test.age, po1.age, @"Should be equal");
+    XCTAssertEqualObjects(test.name, po1.name, @"Should be equal");
+    XCTAssertEqual(test.hired, po1.hired, @"Should be equal");
+    //XCTAssertEqualObjects(test, po1, @"Should be equal"); //FIXME, should work. Asana : https://app.asana.com/0/861870036984/13123030433568
```
jpsim: That Asana task was marked as completed 6 months ago. Does the test still fail? If so, we should open a new GitHub issue.

Realm/RLMObjectStore.mm
```diff
@@ -431,14 +431,18 @@ void RLMDeleteAllObjectsFromRealm(RLMRealm *realm) {
...
-                                                                            realm:realm];
-    return results;
+
+    if (predicate) {
+        tightdb::Query query = objectSchema.table->where();
+        RLMUpdateQueryWithPredicate(&query, predicate, realm.schema, objectSchema);
+
+        // create and populate array
+        return [RLMResults resultsWithObjectClassName:objectClassName
+                                                query:std::make_unique<Query>(query)
```
alazier: You aligned the wrong :

Realm/RLMResults.mm
```diff
@@ -295,25 +296,24 @@ - (id)objectAtIndexedSubscript:(NSUInteger)index {
     return [self objectAtIndex:index];
 }
 
--(id)minOfProperty:(NSString *)property {
-    RLMResultsValidate(self);
-
-    if (self.count == 0) {
+template<typename Table>
```
alazier: Table for the template parameter name is confusing. TableType might be an improvement.

Realm/Tests/ArrayPropertyTests.m
```diff
@@ -334,4 +333,105 @@ - (void)testSortByMultipleColumns {
     XCTAssertTrue(checkOrder(@[@"age", @"dogName"], @[@NO, @NO], @[b2, a2, b1, a1]));
 }
 
+- (void)testDeleteLinksAndObjectsInArray
+{
+    RLMRealm *realm = [RLMRealm defaultRealm];
+    [realm beginWriteTransaction];
+
+    EmployeeObject *po1 = [[EmployeeObject alloc] init];
```
alazier: These 15 lines could be compacted to
```
[EmployeeObject createWithObject:@[@40, @"Joe", @YES]];
[EmployeeObject createWithObject:@[@ 30, @"John", @NO]];
[EmployeeObject createWithObject:@[@ 25, @"Jill", @YES]];
```

alazier:
Do we have separate tests for any method which has different implementations for the two RLMResults variants?

Might also be nice to verify the subclass is what we expect in the tests.

Also, it would be great if there were a way to factor the test so that the same code could get applied to both variants. Not sure if this is possible though.

jpsim:
While this looks good and mostly safe, I'm worried that our existing tests and these modifications to them don't cover quite enough.

tgoyne:
Any suggestions for other things to test? RLMResults.mm now has 99% test coverage, and the only uncovered things are some error checking unrelated to the changes here. All of the checks which involve code-paths that vary between the two implementations are done on both.

segiddins:
Also closes https://github.com/realm/realm-cocoa/issues/1160 ?

tgoyne:
No, this has no effect once you sort or filter the results in any way.

bmunkholm:
Nice to finally get that done :-)
Did you measure the performance improvements?

jpsim:
@bmunkholm see https://github.com/realm/realm-cocoa/pull/1353/files#diff-1. Hard to tell if there were necessarily *improvements*, but there certainly weren't any regressions based on the performance tests we have.

tgoyne:
This reduces the asymptotic complexity of many operations (O(N<sup>2</sup>) -> O(N) for enumerate-and-mutate, O(N) -> O(1) for just checking the count), so the speedup depends on the size of the data set. For the specific user report that lead to actually implementing this it's about 100x faster.

alazier:
:+1:

PR #1355, opened 2015-01-16
Allow per Realm schema versions and migration

### LINKED ISSUE #1336: Support per-Realm migration blocks and schema versions per Realm

alazier:
Also cleans up the logic for migration and checking/updating schema version.

Closes #1336, #1326 
Replaces #1340, #1335  

@tgoyne


tgoyne:
Examples need to be updated and `migrateRealmAtPath:`'s documentation mentions `setSchemaVersion:withMigrationBlock:`.

tgoyne:
This makes the browser unable to open files with a schema version other than 0, since it makes it check the schema version even for dynamic realms.

tgoyne:
:+1:

Issue #1238, opened 2014-12-17
New Swift APIs


shmidt:
```swift
extension RLMObject {
    // Swift query convenience functions
    public class func objectsWhere(predicateFormat: String, _ args: CVarArgType...) -> RLMResults {
        return objectsWhere(predicateFormat, args: getVaList(args))
    }

    public class func objectsInRealm(realm: RLMRealm, _ predicateFormat: String, _ args: CVarArgType...) -> RLMResults {
        return objectsInRealm(realm, `where`: predicateFormat, args: getVaList(args))
    }
}
```

As for me, I would like to see something like:

```swift
extension RLMObject {
    // Swift query convenience functions
    public class func objects(where predicateFormat: String, _ args: CVarArgType...) -> RLMResults {
        return objectsWhere(predicateFormat, args: getVaList(args))
    }

    public class func objects(inRealm realm: RLMRealm, _ predicateFormat: String, _ args: CVarArgType...) -> RLMResults {
        return objectsInRealm(realm, `where`: predicateFormat, args: getVaList(args))
    }
}
```

alazier:
What's left to be done:
- [ ] documentation
- [ ] swift versions for RLMObjectBase methods `primaryKey`, `ignoaredProperties` etc. - these currently take/return objc types
- [ ] migrations/dynamic interface
- [ ] hide `getGenericListPropertyNames` and other private apis
- [ ] fix/workaround the auto-linking issue
- [ ] comprehensive testing

@jpsim please add other missing items you are aware of.

jpsim:
I've just checked the "walk through Objective-C API and add missing aspects in Swift interface", but that excludes `Migration.swift`.

alazier:
The new apis have been merged to master.

PR #1356, opened 2015-01-17
[Documentation] Document the swift interface

segiddins:
Connected to #1238.

\c @alazier @jpsim


Realm/RealmSwift/List.swift
```diff
@@ -39,30 +47,70 @@ public final class List<T: Object>: ListBase, SequenceType {
...
+    :returns: The index of the given object, or `nil` if the object is not in the list.
+    */
     public func indexOf(predicateFormat: String, _ args: CVarArgType...) -> Int? {
         return notFoundToNil(_rlmArray.indexOfObjectWithPredicate(NSPredicate(format: predicateFormat, arguments: getVaList(args))))
     }
 
     // MARK: Object Retrieval
 
+    /**
+    Returns the obejct at the given `index`.
```
jpsim: s/obejct/object/

Realm/RealmSwift/List.swift
```diff
@@ -72,28 +120,57 @@ public final class List<T: Object>: ListBase, SequenceType {
...
 
     // MARK: Sorting
 
+    /**
+    Returns a sorted version of the list.
+
+    :param: property  The property name to sort by.
+    :param: ascending The direction to sort by.
+
+    :returns: Results containting the objects in the list sorted by the given property.
```
jpsim: s/containting/containing/

Realm/RealmSwift/List.swift
```diff
@@ -72,28 +120,57 @@ public final class List<T: Object>: ListBase, SequenceType {
...
         return Results<T>(_rlmArray.sortedResultsUsingProperty(property, ascending: ascending))
     }
 
-    // MARK: Sequence Support
+    // MARK: Sequence support
 
+    /**
+    Returns a `GeneratorOf<T>` that yields successive elements in the list.
+
+    :returns: Returns a `GeneratorOf<T>` that yields successive elements in the list.
```
jpsim: omit the "Returns":

`:returns: A `GeneratorOf<T>` that yields successive elements in the list.`

Realm/RealmSwift/List.swift
```diff
@@ -105,46 +182,105 @@ public final class List<T: Object>: ListBase, SequenceType {
...
+
+    /**
+    Replaces an object at the given index with a new object.
+
+    Throws an exception when called with an index greater than the number of objects in this List.
+
+    :warning: This method can only be called during a write transaction.
+
+    :param: index  The list index of the object to be replaced.
+    :param: object An object
```
jpsim: missing period

Realm/RealmSwift/Migration.swift
```diff
@@ -18,31 +18,111 @@
...
+/**
+Performs the registered migration block on the Realm at the given path.
+
+This method is called automatically when opening a Realm for the first time and does
+not need to be called explicitly. You can choose to call this method to control
+exactly when and how migrations are performed.
+
+:see: setSchemaVersion(_:_:)
+
+:param: realmPath The path of the Realm to migrate.
```
jpsim: s/realmPath/path/

Realm/RealmSwift/Migration.swift
```diff
@@ -18,31 +18,111 @@
...
 }
 
+/**
+Migration encapsulates the information passed into a MigrationBlock when updating the schema version of a Realm instance.
+*/
 public class Migration {
+    // MARK: Properties
+
+    /// The old schema.
+    /// This is the schema that describes the Realm before the migration is applied.
```
jpsim: Remove the first line here. Total comment could just read `The schema that describes the Realm before the migration is applied.`. Same for the new schema one.

Realm/RealmSwift/Migration.swift
```diff
@@ -18,31 +18,111 @@
...
     }
 
+    /**
+    Delete the given object from a Realm during a migration.
+    This can be called within `enumerate(_:_:)`.
+
+    :param: object The object to be deleted from the Realm being migrated.
+    */
     public func delete(object: RLMObject) {
+        // FIXME: This should take Object
```
jpsim: why not just fix this?

Realm/RealmSwift/Object.swift
```diff
@@ -19,8 +19,48 @@
...
+- `Bool`, `BOOL`, and `bool`
+- `NSDate`
+- `NSData`
+- `Object` subclasses, so you can have many-to-one relationships.
+- `List<T: Object>`, so you can have many-to-many relationships.
+
+### Querying
+
+You can gets `Results` of an Object subclass via tha `objects(_:)` free function or
+`Realm.
```
jpsim: This should read "or `objects()` instance method on `Realm`".

Realm/RealmSwift/Object.swift
```diff
@@ -19,8 +19,48 @@
...
     // Get the names of all properties in the object which are of type List<>
+
+    /**
+    Returns the names of all properties in `obj` that are of type `List<>`.
+
+    :param: obj The object to be reflected.
+
+    :returns: Returns the names of all properties in `obj` that are of type `List<>`.
+    */
     public class func getGenericListPropertyNames(obj: AnyObject) -> NSArray {
```
jpsim: We should call this function `PRIVATE_getGenericListPropertyNames()` and have an all-caps `PRIVATE` disclaimer in the doc comment. It should be private but that isn't possible.

Realm/RealmSwift/ObjectSchema.swift
```diff
@@ -43,8 +58,9 @@ public class ObjectSchema: Equatable {
 
     // MARK: Property Retrieval
 
+    /// Returns the property with the given name, if there it exists.
```
jpsim: > if it exists.

Realm/RealmSwift/Realm.swift
```diff
@@ -21,27 +21,98 @@ import Realm.Private
...
 
-public class Realm {
+/**
+A Realm instance (also referred to as "a realm") represents a Realm
+database.
+
+Realms can either be stored on disk (see `init(path:)`) or in
+memory (see `init(inMemoryIdentifier:)`).
+
+RLMRealm instances are cached internally, and constructing equivalent RLMRealm
```
jpsim: s/RLMRealm/Realm/

jpsim: there are many such offenses in this paragraph

Realm/RealmSwift/Realm.swift
```diff
@@ -21,27 +21,98 @@ import Realm.Private
...
+    Disabling this on an `Realm` without any strong references to it will not
+    have any effect, and it will switch back to YES the next time the `Realm`
+    object is created. This is normally irrelevant as it means that there is
+    nothing to refresh (as persisted `Object`s, `List`s, and `Results` have strong
+    references to the containing `Realm`), but it means that setting
+    `defaultRealm().autorefresh = false` in
+    `application(_:didFinishLaunchingWithOptions:)` and only later storing Realm
+    objects will not work.
+    
+    Defaults to YES.
```
jpsim: s/YES/true/

Realm/RealmSwift/Realm.swift
```diff
@@ -74,6 +184,24 @@ public class Realm {
...
+    process since lldb cannot forward mach exceptions to the process being
+    debugged. Attempting to open an encrypted Realm with lldb attached will result
+    in an EXC_BAD_ACCESS.
+
+    :param: path        Path to the file you want the data saved in.
+    :param: key         64-byte key to use to encrypt the data.
+    :param: readonly    Bool indicating if this Realm is read-only (must use for read-only files).
+    :param: error       If an error occurs, upon return contains an `NSError` object
+                        that describes the problem. If you are not interested in
+                        possible errors, pass in `NULL`.
```
jpsim: s/NULL/nil/

Realm/RealmSwift/Realm.swift
```diff
@@ -74,6 +184,24 @@ public class Realm {
...
+    */
     public convenience init?(path: String, encryptionKey: NSData, readOnly: Bool, error: NSErrorPointer = nil) {
```
jpsim: s/readOnly/readonly/

Realm/RealmSwift/Realm.swift
```diff
@@ -85,52 +213,194 @@ public class Realm {
...
+    This rolls back all objects in the Realm to the state they were in at the
+    beginning of the write transaction, and then ends the transaction.
+
+    This restores the data for deleted objects, but does not re-validated deleted
+    accessor objects. Any `Object`s which were added to the Realm will be
+    invalidated rather than switching back to standalone objects.
+    Given the following code:
+
+    ```swift
+    let oldObject = objects(ObjectType.self).first!
```
jpsim: `objects(ObjectType).first!` is nicer

Realm/RealmSwift/Realm.swift
```diff
@@ -85,52 +213,194 @@ public class Realm {
...
     }
 
+    /**
+    Adds objects in the given sequence to be persisted it in this Realm.
+
+    :see: add(object:)
+
+    :param: objects A sequence which contains objects to be added to this Realm.
+    */
     public func add<S where S: SequenceType>(objects: S) {
```
jpsim: `add<S: SequenceType where S.Generator.Element == Object>(objects: S) {`

Realm/RealmSwift/Realm.swift
```diff
@@ -139,44 +409,102 @@ public class Realm {
...
     }
 
+    /**
+    Adds or updates objects in the given array to be persisted it in this Realm.
+
+    :see: addOrUpdate(object:)
+
+    :param: objects  A sequence of `Object`s to be added to this Realm.
+    */
     public func addOrUpdate<S where S: SequenceType>(objects: S) {
```
jpsim: `addOrUpdate<S: SequenceType where S.Generator.Element == Object>(objects: S)`

Realm/RealmSwift/Realm.swift
```diff
@@ -139,44 +409,102 @@ public class Realm {
...
     public func delete(object: Object) {
         RLMDeleteObjectFromRealm(object)
     }
 
+    /**
+    Deletes the given objects from this Realm.
+
+    :param: object The objects to be deleted.
+    */
     public func delete(objects: [Object]) {
```
jpsim: we should replace this (and the other one) with `delete<S: SequenceType where S.Generator.Element == Object>(objects: S)`

Realm/RealmSwift/Realm.swift
```diff
@@ -139,44 +409,102 @@ public class Realm {
...
     }
 
     // MARK: Notifications
 
+    /**
+    Add a notification handler for changes in this Realm.
+
+
+
+    :param: block   A block which is called to process RLMRealm notifications.
```
jpsim: SO










MUCH






WHITESPACE

Realm/RealmSwift/Realm.swift
```diff
@@ -139,44 +409,102 @@ public class Realm {
...
     public func removeNotification(notificationToken: NotificationToken) {
         rlmRealm.removeNotification(notificationToken)
     }
 
     // MARK: Object Retrieval
 
+    /**
+    Returns all objects of the given type in this Realm.
+
+    :param: type The type of the obejcts to be returned.
```
jpsim: s/obejcts/objects/

Realm/RealmSwift/Realm.swift
```diff
@@ -184,6 +512,12 @@ public class Realm {
 
 // MARK: Notifications
 
+/**
+A notification due to changes to a realm.
+
+- DidChange:       The realm did change.
```
jpsim: Should move these to be above the enum elements.

Realm/RealmSwift/Migration.swift
segiddins: it's out of scope for this PR

Realm/RealmSwift/Object.swift
```diff
@@ -19,40 +19,268 @@
...
+                   thrown if any required properties are not present and no default is set.
+
+                   When passing in an `Array`, all properties must be present,
+                   valid and in the same order as the properties defined in the model.
+    :param: realm  The Realm in which this object is persisted.
+                   The default Realm will be used if this argument is omitted.
+
+    :returns: The created object.
+    */
+    public class func createWithObject(object: AnyObject, inRealm realm: Realm = defaultRealm()) -> Self {
```
jpsim: Both `createWithObject(object:, inRealm:)` and `createOrUpdateWithObject(object:, inRealm:)` are the reverse order compared to the Objective-C API's. I'm proposing this change because it allows omitting the realm argument if the default Realm should be used. I think it simplifies the Swift API's without sacrificing readability, discoverability or functionality, at the expense of being slightly different than the Objective-C equivalents.

Ultimately, this is just a proposal and should be discussed /cc @alazier @tgoyne @segiddins @mrackwitz

jpsim:
I'm done making changes to this and this is ready to be reviewed for merging in `al-swift`. Here's a summary of what this PR does:

1. Adds docs to all declarations. These are mostly copy-pasted from the equivalent Objective-C declarations with minor modifications in function/argument naming and uses ReST instead of the headerdoc format. This hopefully shouldn't be too controversial.
2. Adds missing declarations. A few readonly properties and overridden class functions were added to `Object`, `setEncryptionKey(_:forRealmsAtPath:)` was added.
3. Uses explicit ACL modifiers (private/internal/public).
4. Marks classes as `final` if they're not designed to be subclassed.
5. Unifies wrapper initializers to all take unnamed first arguments.
6. `defaultRealmPath` is now a global variable rather than a pair of free setter/getter functions.
7. Instead of a free function `objects(_:)` and an instance function `Realm().objects(_:)`, the free function is now `objects(_:inRealm:)` with the default realm being used if the realm argument is omitted.
8. The two `notFoundToNil(_:)` implementations were combined.
9. A `SortDescriptor` struct was added, and a `sorted(sortDescriptors:)` instance function was added to both `List` and `Results`.
10. The tests were updated to reflect these changes.

Realm/RealmSwift/Object.swift
tgoyne: Having the `realm` argument second is really awkward when you're passing a large object literal. I'd prefer to just have `createWithObject(obj)` and `createdWithObject(realm, obj)`.

jpsim: I agree it's a bit more awkward:

```swift
LargeModel.createWithObject(
  [
    "someLongPropertyA": [
      "aNestedProperty": 123567,
      "anotherNestedProperty": "asdfkjhasdfk jhasdfjkh asdkfjhas fkhasdf"
    ],
    "prop2": "another value"
  ],
  inRealm: realm)
```

```swift
LargeModel.createInRealm(realm, withObject:
  [
    "someLongPropertyA": [
      "aNestedProperty": 123567,
      "anotherNestedProperty": "asdfkjhasdfk jhasdfjkh asdkfjhas fkhasdf"
    ],
    "prop2": "another value"
  ])
```

The best alternative IMO would be to split in two class methods: `RLMObject.createInRealm(_:withObject:)` and `RLMObject.createWithObject(_:)`. However, I still prefer my suggestion because it's a single method, which helps with autocomplete and keeps our API more concise, it's self-documenting and there's a bit less magic IMO.

jpsim: I'll move this to `RLMObject.createInRealm(_:withObject:)` and `RLMObject.createWithObject(_:)`, then rebase against `mr-swift`, at which point we should review this before doing any new Swift work.

jpsim:
I'm about to rebase this against `al-swift` and add tests for new/modified functions.

Realm/RealmSwift/Migration.swift
```diff
@@ -19,61 +19,263 @@
...
+        set {
+            if self.objectSchema[key].type != .Array {
+                super[key] = newValue
+            }
+            // FIXME: Implement List property setter
         }
-        return super[key]
     }
 
+     /**
```
segiddins: alignment

Realm/RealmSwift/Object.swift
```diff
@@ -19,40 +19,309 @@
...
+
+    // MARK: Customizing
+
+    /**
+    Return an array of property names for properties which should be indexed. Only supported
+    for string and int properties.
+
+    :returns: `Array` of property names to index.
+    */
+    public class func indexedProperties() -> [String]? { return nil }
```
segiddins: any reason to not just return an empty array?

Realm/RealmSwift/Realm.swift
```diff
@@ -21,27 +21,135 @@ import Realm.Private
...
+Returns all objects of the given type in the specified Realm default realm,
+or in the default realm if the `realm` argument is omitted.
+
+:param: type  The type of the objects to be returned.
+:param: realm The Realm instance to query.
+              The default Realm will be used if this argument is omitted.
+
+:returns: Results with all objects of the given type in the specified Realm.
+*/
+public func objects<T: Object>(type: T.Type, inRealm realm: Realm = defaultRealm()) -> Results<T> {
```
segiddins: is the type parameter really necessary? The function will already be parameterized based on the Results type.

Realm/RealmSwift/Realm.swift
```diff
@@ -21,27 +21,135 @@ import Realm.Private
...
+objects (with the same path or identifier) multiple times on a single thread
+within a single iteration of the run loop will normally return the same
+Realm object. If you specifically want to ensure a Realm object is
+destroyed (for example, if you wish to open a realm, check some property, and
+then possibly delete the realm file and re-open it), place the code which uses
+the realm within an `autoreleasepool {}` and ensure you have no other
+strong references to it.
+
+:warning: Realm instances are not thread safe and can not be shared across
+          threads or dispatch queues. You must call this method on each thread you want
```
segiddins: > You must call this method

?

Realm/RealmSwift/Realm.swift
jpsim: Yes, but it looks unwieldy:

```swift
let current = objects(Person)
let proposed: Results<Person> = objects()
// or
let proposed2 = objects as Results<Person>
```

Realm/RealmSwift/Realm.swift
jpsim: I'll change here and in `RLMRealm` to "you must construct a new instance on each thread..."

Realm/RealmSwift/Realm.swift
segiddins: But if you already have a property or a function parameter that's typed...

Yeah, you're right.

tgoyne: Could have an overload which infers the type, but I'm not sure I even like `self.people = objects()`.

Realm/RealmSwift/Object.swift
jpsim: I think that was incompatible with the definition in `RLMObjectBase`, but this isn't the right solution either. I'll comment it out and add a FIXME like the other similar issues in this file.

Realm/Tests/RealmSwift/ListTests.swift
```diff
@@ -292,7 +292,7 @@ class ListTests: TestCase {
         if let realm = array.realm {
             array.append([str1, str2])
 
-            let otherArray = realm.objects(SwiftArrayPropertyObject).first!.array
+        let otherArray = objects(SwiftArrayPropertyObject.self, inRealm: realm).first!.array
```
segiddins: indentation

Realm/RealmSwift/List.swift
```diff
@@ -105,50 +190,115 @@ public final class List<T: Object>: ListBase, SequenceType {
...
-        for obj in objects {
-            _rlmArray.addObject(unsafeBitCast(obj as T, RLMObject.self))
+    /**
+    Appends the objects in the given sequence to the end of the list.
+
+    :warning: This method can only be called during a write transaction.
+
+    :param: objects A sequence of objects.
+    */
+    public func append<S: SequenceType where S.Generator.Element == T>(objects: S) {
```
jpsim: This is one of the few non-doc, non-trivial changes but it should hardly be contentious.

Realm/RealmSwift/Migration.swift
```diff
@@ -19,61 +19,226 @@
...
 
-    public func delete(object: RLMObject) {
-        rlmMigration.deleteObject(object)
+    /**
+    Delete an object from a Realm during a migration. This can be called within
+    `enumerate(_:block:)`.
+
+    :param: object Object to be deleted from the Realm being migrated.
+    */
+    public func delete(object: MigrationObject) {
```
jpsim: Minor interface change.

Realm/RealmSwift/Realm.swift
jpsim: This is one of the few non-doc, non-trivial changes that I kept in this PR because it goes in concert with  the docs for it.

Realm/RealmSwift/Realm.swift
```diff
@@ -53,84 +137,267 @@ public class Realm {
...
 
-    public func add<S where S: SequenceType>(objects: S) {
+    /**
+    Adds objects in the given sequence to be persisted it in this Realm.
+
+    :see: add(object:)
+
+    :param: objects A sequence which contains objects to be added to this Realm.
+    */
+    public func add<S: SequenceType where S.Generator.Element == Object>(objects: S) {
```
jpsim: small API change here

Realm/RealmSwift/Realm.swift
```diff
@@ -139,60 +406,107 @@ public class Realm {
...
 
-    public func addOrUpdate<S where S: SequenceType>(objects: S) {
+    /**
+    Adds or updates objects in the given array to be persisted it in this Realm.
+
+    :see: addOrUpdate(object:)
+
+    :param: objects A sequence of `Object`s to be added to this Realm.
+    */
+    public func addOrUpdate<S: SequenceType where S.Generator.Element == Object>(objects: S) {
```
jpsim: small API change here

Realm/RealmSwift/Realm.swift
```diff
@@ -139,60 +406,107 @@ public class Realm {
...
+    :param: notificationToken The token returned from `addNotificationBlock(_:)`
+                              corresponding to the notification block to remove.
+    */
     public func removeNotification(notificationToken: NotificationToken) {
         rlmRealm.removeNotification(notificationToken)
     }
-
-    // MARK: Object Retrieval
-
-    public func objects<T: Object>(type: T.Type) -> Results<T> {
```
jpsim: This was removed in favor of the `objects(_:inRealm:)` free function.

jpsim:
This PR is now almost exclusively about adding documentation. If there's anything at all unrelated to documentation in here that you're unsure about, let me know and I'll factor it out into a different PR.

Once this is merged, I'll be filing a PR to add the missing aspects of the interface that I just removed.

Please review @alazier @tgoyne @segiddins @mrackwitz

jpsim:
This is now *just* docs. @alazier @tgoyne @segiddins @mrackwitz

Realm/RealmSwift/Aliases.swift
```diff
@@ -20,5 +20,11 @@ import Realm
 
 // These types don't change when wrapping in Swift
 // so we just typealias them to remove the 'RLM' prefix
+
+// MARK: Aliases
+
+/// An enum of the different property types supported by Realm.
```
tgoyne: Should list the values in the docs unless Jazzy handles that automatically somehow.

Realm/RealmSwift/Aliases.swift
```diff
@@ -20,5 +20,11 @@ import Realm
...
 public typealias PropertyType = RLMPropertyType
+
+/// A token that holds onto the Realm and the notification block.
```
tgoyne: "An opaque token used to unregister notification blocks from Realms."

The fact that it has a reference to the RLMRealm and the block isn't exposed in the interface.

Realm/RealmSwift/List.swift
```diff
@@ -19,17 +19,34 @@
...
 
+/**
+`List<T>` is the container type in Realm used to define to-many relationships.
+
+Lists hold a single `Object` subclass (`T`) which defines the "type" of the list.
+
+Lists can be filtered and sorted with the same predicates as `Results<T>`.
+
+List properties on objects are lazily created when accessed, or can be
+obtained by querying a Realm.
```
tgoyne: Querying a Realm never returns a List.

tgoyne: Also, List properties are created eagerly, since doing it lazily didn't seem to be possible.

Realm/RealmSwift/List.swift
```diff
@@ -72,28 +128,57 @@ public final class List<T: Object>: ListBase, SequenceType {
...
+    /// Returns the last object in the list, or `nil` if empty.
     public var last: T? { return _rlmArray.lastObject() as T? }
 
-    // MARK: Subarray Retrieval
+    // MARK: Filtering
+
+    /**
+    Returns `Results` containing list elements that match the given predicate.
+
+    :param: predicateFormat The predicate format string which can accept variable arguments.
```
tgoyne: This sentence is very awkward and doesn't say what it's trying to say.

Realm/RealmSwift/List.swift
```diff
@@ -72,28 +128,57 @@ public final class List<T: Object>: ListBase, SequenceType {
...
 
     // MARK: Sorting
 
+    /**
+    Returns `Results` containing list elements sorted by the given property name.
+
+    :param: property  The property name to sort by.
+    :param: ascending The direction to sort by.
+
+    :returns: `Results` containing list elements sorted by the given property name.
```
tgoyne: "by the given property." or "by the property with the given name."

Realm/RealmSwift/List.swift
```diff
@@ -105,50 +190,115 @@ public final class List<T: Object>: ListBase, SequenceType {
...
+    */
     public func insert(object: T, atIndex index: Int) {
         _rlmArray.insertObject(unsafeBitCast(object, RLMObject.self), atIndex: UInt(index))
     }
 
+    /**
+    Removes the object at the given index from the list. Does not remove the object from the Realm.
+
+    :warning: This method can only be called during a write transaction.
+    :warning: Throws an exception when called with an index greater than the number of objects in the list.
```
tgoyne: or less than 0 (this is relevant to a bunch of things taking an int index).

Realm/RealmSwift/List.swift
```diff
@@ -39,30 +56,69 @@ public final class List<T: Object>: ListBase, SequenceType {
...
+    /// Creates a `List` that holds objects of type `T`.
     public override init() {
         super.init(array: RLMArray(objectClassName: T.className()))
     }
 
+    /**
+    Create a `List` by passing in the `RLMArray` to be wrapped.
+
+    :param: rlmArray `RLMArray`.
+    */
```
tgoyne: Private things should not have formal documentation.

Realm/RealmSwift/Migration.swift
```diff
@@ -19,47 +19,185 @@
...
+    private var rlmMigration: RLMMigration
+
+    // MARK: Altering Objects During a Migration
+
+    /**
+    Enumerates objects of a given type in this Realm, providing both the old and new versions of
+    each object. Object properties can be accessed using subscripting.
+
+    :warning: All objects returned are of a type specific to the current migration and should not be
+              casted to className. Instead you should access them as `Object`s and use subscripting
```
tgoyne: "should not be casted to the normal object type" or something.

Realm/RealmSwift/List.swift
jpsim: How about "The predicate format string, optionally followed by a variable number of arguments."?

Or "The predicate format string and optional format arguments which will be passed to `NSPredicate(format:arguments:)`."?

Realm/RealmSwift/Realm.swift
```diff
@@ -85,52 +228,166 @@ public class Realm {
...
         }
     }
 
+    /**
+    Adds or updates an object to be persisted it in this Realm. The object provided must have a designated
+    primary key. If no objects exist in the Realm instance with the same primary key value, the object is
+    inserted. Otherwise, the existing object is updated with any changed values.
+
+    As with `add`, the object cannot already be persisted in a different
+    Realm. Use `Object(realm:object:)` to copy values to a different Realm.
```
tgoyne: Use `creeateOrUpdate` to copy values to a different Realm.

Realm/RealmSwift/List.swift
jpsim: Jazzy supports a minimum ACL setting and we'll be using `public`, so this won't show up in user-facing documentation.

Realm/RealmSwift/Realm.swift
```diff
@@ -184,13 +497,24 @@ public class Realm {
 
 // MARK: Notifications
 
+/// A notification due to changes to a realm.
 public enum Notification: String {
```
tgoyne: c/p the docs from the obj-c header

Realm/RealmSwift/Results.swift
```diff
@@ -37,83 +41,199 @@ extension Int32: AddableType {}
 extension Int64: AddableType {}
 extension Int: AddableType {}
 
+/**
+`Results` is an auto-updating container type in Realm returned from object
+queries.
+
+Results can be queried with the same predicates as `Object` and `List`
```
tgoyne: `Object` isn't queryable.

Realm/RealmSwift/Realm.swift
jpsim: Which objc header? These enum values are just global constants in objc.

Realm/RealmSwift/List.swift
tgoyne: They make the code far less readable and they're mostly worthless noise when not used for the documentation generator. Good documentation for a user of an API and useful comments for people working on the code have little in common.

jpsim: Fair enough.

Realm/RealmSwift/Results.swift
```diff
@@ -37,83 +41,199 @@ extension Int32: AddableType {}
...
+    :param: property The name of a property conforming to `MinMaxType` to look for a minimum on.
+
+    :returns: The minimum value for the property amongst objects in the Results, or `nil` if the Results is empty.
+    */
     public func min<U: MinMaxType>(property: String) -> U {
         return rlmResults.minOfProperty(property) as U
     }
 
+    /**
+    Returns the maximum (lowest) value of the given property.
```
tgoyne: (highest), although I'm not sure what that's even trying to clarify

Realm/RealmSwift/Results.swift
```diff
@@ -37,83 +41,199 @@ extension Int32: AddableType {}
...
         return rlmResults.averageOfProperty(property) as AnyObject as U
     }
 
     // MARK: Sequence Support
 
+    /**
+    Returns a `GeneratorOf<T>` that yields successive elements in the results.
+
+    :returns: A `GeneratorOf<T>` that yields successive elements in the results.
+    */
```
tgoyne: Just do a single-line description if there's nothing more to be said.

Realm/RealmSwift/Realm.swift
tgoyne: RLMConstants.h documents them.

jpsim: indeed. awesome.

Realm/RealmSwift/Aliases.swift
```diff
@@ -20,5 +20,33 @@ import Realm
...
+// MARK: Aliases
+
+/**
+An enum of the different property types supported by Realm. Object type.
+
+See [Realm Models](http://realm.io/docs/cocoa/latest/#models).
+
+## Primitive types
+
+* Int    = 0
```
segiddins: Is there any reason to document what the enum values map to?

Realm/RealmSwift/List.swift
```diff
@@ -19,17 +19,33 @@
...
 }
 
+/**
+`List<T>` is the container type in Realm used to define to-many relationships.
+
+Lists hold a single `Object` subclass (`T`) which defines the "type" of the list.
+
+Lists can be filtered and sorted with the same predicates as `Results<T>`.
+
+When added as a property on `Object` models, the property must be `let` and cannot be `dynamic`.
```
segiddins: must be declared as `let` ?

Realm/RealmSwift/Aliases.swift
jpsim: No that's just what we do in ObjC. I'll remove the numbers.

Realm/RealmSwift/Migration.swift
```diff
@@ -19,47 +19,183 @@
...
+                         existing objects which require migration.
+:param: oldSchemaVersion The schema version of the `Realm` being migrated.
+*/
 public typealias MigrationBlock = (migration: Migration, oldSchemaVersion: UInt) -> Void
+
+/**
+Provides both the old and new versions of an object in this Realm. Objects properties can be
+accessed using subscripting.
+
+:param: oldObject Object in original `Realm` (read-only)
```
segiddins: period after the param descriptions

Realm/RealmSwift/Migration.swift
```diff
@@ -19,47 +19,183 @@
...
+
+:param: migration        `Migration` object used to perform the migration. The
+                         migration object allows you to enumerate and alter any
+                         existing objects which require migration.
+:param: oldSchemaVersion The schema version of the `Realm` being migrated.
+*/
 public typealias MigrationBlock = (migration: Migration, oldSchemaVersion: UInt) -> Void
+
+/**
+Provides both the old and new versions of an object in this Realm. Objects properties can be
```
segiddins: can _only_ be ?

jpsim:
From our earlier in-person conversation, we've agreed that this should be merged as-is to prevent it from lingering unnecessarily.

PR #1358, opened 2015-01-20
[Podspec] Fix private header definition for using frameworks with CocoaPods

### LINKED ISSUE #1280: error: could not build Objective-C module 'Realm'
I try to build a Mac command line tool to fill a realm database I'd like to use in a iOS app.
I followed the instructions to add the Realm framework to the Mac app.

When I build directly after I have deleted the build folder everything works. When I then change something in a file which imports the Realm module I get the following error:

```
error: could not build Objective-C module 'Realm'
import Realm
       ^
```
I'm using Swift.

Deleting the build folder works but it's far from perfect. Am I missing something?

This does not happen in my iOS project (as far as I can tell).

mrackwitz:
Otherwise private headers will be imported from the umbrella header, which will cause that the framework can't be build.
Fixes #1280.

### DIFF
--- a/Realm.podspec
+++ b/Realm.podspec
@@ -38,5 +38,6 @@ Pod::Spec.new do |s|
 
   s.subspec 'Headers' do |s|
     s.source_files          = 'include/**/*.h'
+    s.private_header_files  = '**/*.hpp', '**/*_Private.h', '**/*_Dynamic.h', 'include/tightdb/**/*.h'
   end
 end
--- a/Realm.podspec
+++ b/Realm.podspec
@@ -21,9 +21,11 @@ Pod::Spec.new do |s|
   s.documentation_url       = "http://realm.io/docs/cocoa/#{s.version}"
   s.license                 = { :type => 'Apache 2.0', :file => 'LICENSE' }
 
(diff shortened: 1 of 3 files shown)


Realm.podspec
```diff
@@ -38,5 +38,6 @@ Pod::Spec.new do |s|
 
   s.subspec 'Headers' do |s|
     s.source_files          = 'include/**/*.h'
+    s.private_header_files  = '**/*.hpp', '**/*_Private.h', '**/*_Dynamic.h', 'include/tightdb/**/*.h'
```
segiddins: any way you could extract this duplicate definition into a variable, so it's in a single place?

mrackwitz: Sure, it's just Ruby. But that will decrease readability slightly, though I doubt that a lot external developers will take a look into the actual podspec.

jpsim: Just change `|s|` to `|subspec|`, so you can do `subspec.private_header_files = s.private_header_files`

mrackwitz: It would need to be `s.attributes_hash[:private_header_files]` as there is no getter.

jpsim: What you're doing now is fine.

tgoyne:
:+1:

PR #1368, opened 2015-01-21
Move getGenericListPropertyNames out of RLMObjectBase

alazier:
I think this is preferable to having the PRIVATE prefix.


alazier:
You ok with this @jpsim ?

Realm/RealmSwift/Object.swift
```diff
@@ -70,3 +52,24 @@ public class Object : RLMObjectBase, Equatable {
 public func == <T: Object>(lhs: T, rhs: T) -> Bool {
     return lhs.isEqualToObject(rhs)
 }
+
+public class ObjectUtil : NSObject {
```
segiddins: does `ObjectUtil` really need to be public?

alazier: If I make it private it seems to be inaccessible from objective-c. Perhaps there is another way?

PR #1373, opened 2015-01-21
Update to core version 0.88.0

tgoyne:


semireg:
(y)

alazier:
:+1:

jpsim:
Shouldn't we also add "Add support for comparing string columns to other string columns in queries." to our changelog? That was included in core 0.88.0.

tgoyne:
No, I still need to expose it in the binding.

jpsim:
Right, thanks.

jpsim:
Looks like https://github.com/realm/realm-core/pull/639 was never reflected in core's release notes, but it was included in 0.87.5. We should add something like "Fixed issue where querying for non-null relationships would return incomplete results.".

Issue #1369, opened 2015-01-21
Include property name in the exception "Inserting invalid object for RLMPropertyTypeAny property"


ishaq:
Including property name in the exception would make it easier to fix the offending code.

```swift
class Post : RLMObject {
  dynamic var content: String = ""
  dynamic var authors: Array<String> = "" // This property is an Array, hence not supported by Realm
}

let post = Post()
post.content = "Hello World"
post.authors = ["ishfaq", "shoaib"]

let realm = RLMRealm.defaultRealm()
realm.beginWriteTransaction()
realm.addObject(post) // This line would throw the exception
realm.commitWriteTransaction()
```

* Realm: 0.89.2
* Xcode: 6.1.1
* OSX: 10.2.2

yoshyosh:
If we can detect that the user is trying to use an array, it would be nice to suggest using RLMArray in the error message

PR #1375, opened 2015-01-21
[RLMAccessor] Improve the exception message when trying to set a property with an object of the incorrect type.

segiddins:
Closes https://github.com/realm/realm-cocoa/issues/1369.

\c @alazier @tgoyne

### DIFF
--- a/Realm/RLMAccessor.mm
+++ b/Realm/RLMAccessor.mm
@@ -310,7 +310,7 @@ static inline void RLMSetValue(__unsafe_unretained RLMObjectBase *obj, NSUIntege
                 return;
         }
     }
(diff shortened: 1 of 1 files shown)


tgoyne:
:+1: once tests pass.

Realm/RLMAccessor.mm
```diff
@@ -310,7 +310,7 @@ static inline void RLMSetValue(__unsafe_unretained RLMObjectBase *obj, NSUIntege
                 return;
         }
     }
-    @throw [NSException exceptionWithName:@"RLMException" reason:@"Inserting invalid object for RLMPropertyTypeAny property" userInfo:nil];
```
alazier: Wont this crash if val is nil? or if the name lookup is nil?

segiddins: No, since everything is a message send

PR #1381, opened 2015-01-21
Add support for interprocess change notifications

tgoyne:


Realm/RLMRealm.mm
```diff
@@ -196,16 +187,26 @@ static void clearRealmCache() {
         for (NSMapTable *map in s_realmsPerPath.allValues) {
             [map removeAllObjects];
         }
-        s_realmsPerPath = [NSMutableDictionary dictionary];
+        [s_realmsPerPath removeAllObjects];
     }
 }
 
+static NSMutableDictionary *s_listenersPerPath = [NSMutableDictionary new];
```
alazier: we should clear this in `resetRealmState`

Realm/RLMRealm.mm
```diff
@@ -196,16 +187,26 @@ static void clearRealmCache() {
...
+// must be called with s_listenersPerPath locked
+static RLMChangeListener *listenerForPath(NSString *path, BOOL inMemory, BOOL create=YES) {
+    RLMChangeListener *listener = s_listenersPerPath[path];
```
alazier: synchronization should be pushed from the caller into this and other listener api methods

Realm/RLMRealm.mm
```diff
@@ -196,16 +187,26 @@ static void clearRealmCache() {
...
             [map removeAllObjects];
         }
-        s_realmsPerPath = [NSMutableDictionary dictionary];
+        [s_realmsPerPath removeAllObjects];
     }
 }
 
+static NSMutableDictionary *s_listenersPerPath = [NSMutableDictionary new];
+// must be called with s_listenersPerPath locked
+static RLMChangeListener *listenerForPath(NSString *path, BOOL inMemory, BOOL create=YES) {
```
alazier: Should split this into two methods `startListening(Realm *realm)` and `stopListening(Realm *realm)`

Realm/RLMChangeListener.h
```diff
@@ -0,0 +1,34 @@
...
+
+// A thread which waits for change notifications on the given path and notifies
+// all registered RLMRealms when a change occurs. Does *not* retain registered
+// RLMRealm instances
+@interface RLMChangeListener : NSThread
+- (instancetype)initWithPath:(NSString *)path
+                    inMemory:(BOOL)inMemory
+                       cache:(NSMutableDictionary *)cache;
+
+// This both must be called with `cache` locked
```
segiddins: s/This/These methods/

Realm/RLMChangeListener.mm
```diff
@@ -0,0 +1,156 @@
...
+}
+
+- (void)run {
+    while (_sharedGroup && _sharedGroup->wait_for_change()) {
+        // we don't have any accessors, so just start a new read transaction
+        // rather than using advance_read() as that does far more work
+        _sharedGroup->end_read();
+        _sharedGroup->begin_read();
+
+        @synchronized (_realms) {
```
alazier: would be nice to have an explicit check that we cancelled/stopped notifications for this path

Realm/RLMChangeListener.mm
```diff
@@ -0,0 +1,156 @@
...
+    std::unique_ptr<SharedGroup> _sharedGroup;
+    NSString *_path;
+    NSMutableArray *_realms;
+    NSCondition *_shutdownCondition;
+}
+
+- (instancetype)initWithPath:(NSString *)path inMemory:(BOOL)inMemory cache:(NSMutableDictionary *)cache {
+    self = [super initWithTarget:self selector:@selector(run) object:nil];
+    if (self) {
+        _cache = cache;
```
alazier: passing in a global cache to be used by all instances seems wrong

Realm/RLMChangeListener.mm
```diff
@@ -0,0 +1,156 @@
...
+        _replication.reset(tightdb::makeWriteLogCollector(path.UTF8String));
+        SharedGroup::DurabilityLevel durability = inMemory ? SharedGroup::durability_MemOnly :
```
alazier: Why not create/open the shared group in the background thread?

Realm/RLMChangeListener.h
```diff
@@ -0,0 +1,34 @@
...
+////////////////////////////////////////////////////////////////////////////
+
+#import <Foundation/Foundation.h>
+
+@class RLMRealm;
+
+// A thread which waits for change notifications on the given path and notifies
+// all registered RLMRealms when a change occurs. Does *not* retain registered
+// RLMRealm instances
+@interface RLMChangeListener : NSThread
```
alazier: Would be preferable to just use a block rather than subclassing NSThread

alazier: Or at least use an NSThread instance instead of subclassing

Realm/RLMChangeListener.mm
```diff
@@ -0,0 +1,156 @@
...
+        _path = path;
+        _realms = [NSMutableArray array];
+        _shutdownCondition = [[NSCondition alloc] init];
+        [self start];
+    }
+    return self;
+}
+
+- (void)addRealm:(RLMRealm *)realm {
+    @synchronized (_realms) {
```
alazier: Would be good to only synchronize on _cache (or the global once it is moved into this file).

Realm/RLMChangeListener.mm
```diff
@@ -0,0 +1,156 @@
...
+        [_realms addObject:[[RLMWeakNotifier alloc] initWithRealm:realm]];
+    }
+}
+
+- (void)removeRealm:(RLMRealm *)realm {
+    @synchronized (_realms) {
+        @autoreleasepool {
+            // The NSPredicate needs to be deallocated before we return or it crashes,
+            // since we're called from -dealloc and so retaining `realm` doesn't work.
+            [_realms filterUsingPredicate:[NSPredicate predicateWithFormat:@"realm != nil AND realm != %@", realm]];
```
alazier: Use a NSMutableSet instead so you don't have to do this query for removal

tgoyne:
From #1531:

> The remaining test failure is a perf regression for realm creation. Not a lot of room left for optimization since 80% of the time is spent in the three `open()` calls. One thing we could do is push the `open()` on the fifo to the background thread, but that'd limit how many error cases we could report gracefully.

tgoyne:
retest me please

tgoyne:
retest this please

Realm/RLMProperty.mm
```diff
@@ -39,7 +39,6 @@ - (instancetype)initWithName:(NSString *)name
         _objectClassName = objectClassName;
         _indexed = indexed;
         [self setObjcCodeFromType];
-        [self updateAccessors];
```
alazier: Why do we no longer need this?

tgoyne: We never needed (there) it in the first place. That initializer is only used for dynamic objects, and dynamic objects don't have accessor properties.

tgoyne: That change can be split out of this PR if desired; it was just a perf optimization to try to make up for the added overhead of opening the named pipe.

alazier: Lets split it out for now.

Realm/RLMRealm.mm
```diff
@@ -395,7 +335,7 @@ + (instancetype)realmWithPath:(NSString *)path
     }
 
     // try to reuse existing realm first
-    RLMRealm *realm = cachedRealm(path);
+    RLMRealm *realm = RLMGetCurrentThreadCachedRealmForPath(path);
```
alazier: This sounds a bit awkward. `RLMGetThreadLocalCachedRealmForPath` sounds a bit better. This isn't a big deal though.

Realm/RLMRealm.mm
```diff
@@ -450,10 +391,17 @@ + (instancetype)realmWithPath:(NSString *)path
...
         }
+
+        if (!dynamic) {
+            RLMCacheRealm(realm);
+        }
     }
 
-    if (!dynamic) {
-        cacheRealm(realm, path);
+    if (!readonly) {
```
alazier: should this be in the synchronized block?

Realm/RLMRealm.mm
tgoyne: ThreadLocal is definitely better here.

Realm/RLMRealm.mm
tgoyne: I don't see any reason why it'd need to be.

Realm/RLMRealmUtil.mm
```diff
@@ -0,0 +1,271 @@
...
+        if (fd != -1) {
+            ::close(fd);
+        }
+        fd = -1;
+    }
+
+public:
+    ~FdHolder() { close(); }
+    operator int() const { return fd; }
+    FdHolder& operator=(int newFd) {
```
alazier: we should make sure this doens't work with FdHolder

Realm/RLMRealmUtil.mm
```diff
@@ -0,0 +1,271 @@
...
+        int pipeFd[2];
+        ret = pipe(pipeFd);
+        if (ret == -1) {
+            return handleError(errno, error);
+        }
+
+        _shutdownReadFd = pipeFd[0];
+        _shutdownWriteFd = pipeFd[1];
+
+        dispatch_async(dispatch_get_global_queue(DISPATCH_QUEUE_PRIORITY_DEFAULT, 0), ^{
```
alazier: spacing looks wrong

Realm/RLMRealmUtil.mm
```diff
@@ -0,0 +1,271 @@
...
+    return self;
+}
+
+- (void)listen {
+    // Create the runloop source
+    CFRunLoopSourceContext ctx{};
+    ctx.info = (__bridge void *)self;
+    ctx.perform = [](void *info) {
+        RLMNotifier *notifier = (__bridge RLMNotifier *)info;
+        if (RLMRealm *realm = notifier->_realm) {
```
alazier: if statement in unecessary

Realm/RLMRealmUtil.mm
```diff
@@ -0,0 +1,271 @@
...
+    // it should be shut down.
+    FdHolder _shutdownReadFd;
+    FdHolder _shutdownWriteFd;
+}
+
+- (instancetype)initWithRealm:(RLMRealm *)realm error:(NSError **)error {
+    self = [super init];
+    if (self) {
+        _realm = realm;
+        _runLoop = CFRunLoopGetCurrent();
```
alazier: we should retain this.

Realm/RLMRealmUtil.mm
```diff
@@ -0,0 +1,271 @@
...
+        CFRunLoopSourceSignal(signal);
+        // Signalling the source makes it run the next time the runloop gets
+        // to it, but doesn't make the runloop start if it's currently idle
+        // waiting for events
+        CFRunLoopWakeUp(_runLoop);
+    }
+}
+
+- (void)stop {
+    notifyFd(_shutdownWriteFd);
```
alazier: should release runloop here

Realm/RLMRealmUtil.mm
```diff
@@ -0,0 +1,271 @@
...
+            // Spurious wakeup; just wait again
+            continue;
+        }
+
+        // Check which file descriptor had activity: if it's the shutdown
+        // pipe, then someone called -stop; otherwise it's the named pipe
+        // and someone committed a write transaction
+        if (event.ident == (uint32_t)_shutdownReadFd) {
+            CFRunLoopSourceInvalidate(signal);
+            return;
```
alazier: release runloop here.

Realm/RLMRealmUtil.mm
```diff
@@ -0,0 +1,271 @@
...
+
+    // Set up the kqueue
+    // EVFILT_READ indicates that we care about data being available to read
+    // on the given file descriptor.
+    // EV_CLEAR makes it wait for the amount of data available to be read to
+    // change rather than just returning when there is any data to read.
+    struct kevent ke[2];
+    EV_SET(&ke[0], _notifyFd, EVFILT_READ, EV_ADD | EV_CLEAR, 0, 0, 0);
+    EV_SET(&ke[1], _shutdownReadFd, EVFILT_READ, EV_ADD | EV_CLEAR, 0, 0, 0);
+    kevent(_kq, ke, 2, nullptr, 0, nullptr);
```
alazier: we should at least check the error here and assert

Realm/RLMRealmUtil.mm
```diff
@@ -0,0 +1,271 @@
...
+            continue;
+        }
+
+        // Check which file descriptor had activity: if it's the shutdown
+        // pipe, then someone called -stop; otherwise it's the named pipe
+        // and someone committed a write transaction
+        if (event.ident == (uint32_t)_shutdownReadFd) {
+            CFRunLoopSourceInvalidate(signal);
+            return;
+        }
```
alazier: should check that we get _notifyFd

Realm/Tests/InterprocessTests.m
```diff
@@ -0,0 +1,329 @@
...
+        // single process do everything
+        nextRun = obj.intCol + arc4random() % 10;
+    }
+}
+
+- (void)testRecoverAfterCrash {
+    if (self.isParent) {
+        [self runChildAndWait];
+        RLMRealm *realm = RLMRealm.defaultRealm;
+        [realm beginWriteTransaction];
```
alazier: could actually write/validate some data to be sure.

Realm/Tests/InterprocessTests.m
```diff
@@ -0,0 +1,329 @@
...
+    }
+    else {
+        // Tell the parent we've launched
+        [realm transactionWithBlock:^{}];
+        // Wait for a commit notification from the parent
+        [self waitForNotification:RLMRealmDidChangeNotification realm:realm block:^{}];
+    }
+}
+
+@end
```
alazier: newline

Issue #1388, opened 2015-01-22
RLMAccessor.h:20:9: 'RLMObjectStore.hpp' file not found - podspec issue


ldiqual:
Related issue: https://github.com/realm/realm-cocoa/issues/1280
Related PR: https://github.com/realm/realm-cocoa/pull/1358

I am now getting this error when installing Realm 0.90.0 via cocoapods 0.36.0.beta.1

```
RLMAccessor.h:20:9: 'RLMObjectStore.hpp' file not found
```

![image](https://cloud.githubusercontent.com/assets/941453/5850872/f050517a-a1b1-11e4-9bf9-606d720950f8.png)

### Podfile:

```
source 'https://github.com/CocoaPods/Specs.git'

platform :ios, '8.0'
inhibit_all_warnings!

pod 'Realm', '0.90.0'
```

segiddins:
RLMAccessor.h should be marked as a private header, right @mrackwitz ?

mrackwitz:
Yes, `RLMAccessor.h` should be private. I will update the podspec, accordingly.

hartbit:
Same problem here.

PR #1389, opened 2015-01-22
Made RLMAccessor a private header

hartbit:
To fix Realm compilation as a CocoaPods framework. #1388


jpsim:
Thanks for the PR, @hartbit but we opted to go a different route, instead explicitly specifying public headers in the podspec in #1393.

Please keep in mind that you'll also have to [accept our CLA](https://docs.google.com/a/jpsim.com/forms/d/1bVp-Wp5nmNFz9Nx-ngTmYBVWVdwTyKj4T0WtfVm0Ozs/viewform?fbzx=4154977190905366979) for us to be able to merge your contributions in the future.

Thanks again!

PR #1392, opened 2015-01-22
Refactor the Realm cache and inter-thread change notifications

tgoyne:
This has a somewhat refactored version of the changes to Realm caching and notifying things that I initially did for the inter-process notifications, minus the actual inter-process notifications part since that's actually fairly separate from this.

Pulls the realm cache and notification logic out of RLMRealm.mm, keeps track of whether we have any pending notifications for a Realm to improve performance under heavy write load, and tracks Realms to notify separately from cached Realms since we're planning to make uncached Realms more of a thing.

@alazier


Realm/RLMRealmCache.h
```diff
@@ -0,0 +1,36 @@
...
+#import <Foundation/Foundation.h>
+
+@class RLMRealm;
+
+// Add a Realm to the weak cache
+void RLMCacheRealm(RLMRealm *realm);
+// Get a Realm for the given path which can be used on the current thread
+RLMRealm *RLMGetCurrentThreadCachedRealmForPath(NSString *path);
+// Get a Realm for the given path
+RLMRealm *RLMGetCachedRealmForPath(NSString *path);
```
alazier: should name RLMGetAnyCachedRealmForPath

Realm/RLMRealmCache.mm
```diff
@@ -0,0 +1,134 @@
...
+            [map removeAllObjects];
+        }
+        [s_realmsPerPath removeAllObjects];
+    }
+    @synchronized (s_notifiersPerPath) {
+        [s_notifiersPerPath removeAllObjects];
+    }
+}
+
+// A weak holder for an RLMRealm to allow calling performSelector:onThread:
```
alazier: Maybe move everything below into a new file RLMListener, or RLMNotifier

Realm/RLMRealmCache.mm
```diff
@@ -0,0 +1,134 @@
...
+
+- (instancetype)initWithRealm:(RLMRealm *)realm;
+- (void)notifyOnTargetThread;
+@end
+
+void RLMStartListeningForChanges(RLMRealm *realm) {
+    @synchronized (s_notifiersPerPath) {
+        NSMutableArray *notifiers = s_notifiersPerPath[realm.path];
+        if (!notifiers) {
+            notifiers = [NSMutableArray new];
```
alazier: use maptable instead of array

PR #1397, opened 2015-01-22
[Podspec] Move headers to subspec only

mrackwitz:
Otherwise they will cause warnings with framework support of CocoaPods 0.36.0-beta1, because they would be added twice. This won't mean any changes for the user, as the `Header` subspec is included by default.

### DIFF
--- a/Realm.podspec
+++ b/Realm.podspec
@@ -21,24 +21,9 @@ Pod::Spec.new do |s|
   s.documentation_url       = "http://realm.io/docs/cocoa/#{s.version}"
   s.license                 = { :type => 'Apache 2.0', :file => 'LICENSE' }
 
-  public_header_files = 'include/Realm/RLMArray.h',
-                        'include/Realm/RLMCollection.h',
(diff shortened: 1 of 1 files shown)


segiddins:
@mrackwitz can you confirm that listing with this podspec is successful, as is using it?

segiddins:
@tgoyne what are your thoughts on this change?

tgoyne:
Seems like the right thing to do if it works.

segiddins:
@mrackwitz :+1: if you confirm this works (both via linting and actually using the pod in an example app that has a test target)

segiddins:
@mrackwitz ping?

jpsim:
I'm closing this since installing Realm with CocoaPods 0.36.0.rc.1 currently works quite well (no warnings, unit testable), using the following Podfile:

```ruby
target 'App' do
  use_frameworks!
  pod 'Realm'
end

target 'AppTests' do
  use_frameworks!
  pod 'Realm'
end
```

Issue #1177, opened 2014-12-04
Tables should be removable during migration when removing an object type


webmagnets:
I have removed the class in my code. Now, how can I remove the table?

segiddins:
Our consensus is to not automatically remove tables, but to provide a method during migrations to manually remove them.

PR #1399, opened 2015-01-23
[RLMObjectStore] Delete tables that are no longer relevant during a migration.

segiddins:
Closes https://github.com/realm/realm-cocoa/issues/1177.

I'll need a little help in adding test coverage for this. Feel free to just say it's wrong :)

\c @tgoyne @alazier

### DIFF
--- a/Realm/RLMMigration.h
+++ b/Realm/RLMMigration.h
@@ -88,13 +88,13 @@ typedef void (^RLMObjectMigrationBlock)(RLMObject *oldObject, RLMObject *newObje
 - (void)deleteObject:(RLMObject *)object;
 
 /**
- Deletes the table for the class with the given name. Throws if that class still is in the schema.
+ Deletes the data for the class with the given name.
  
  @param   name The name of the RLMObject class to delete.
  
- @return whether the table existed and was deleted.
+ @return whether data was deleted.
  */
-- (BOOL)deleteTableForClassName:(NSString *)name;
+- (BOOL)deleteDataForClassName:(NSString *)name;
 
 @end
 
--- a/Realm/RLMMigration.h
+++ b/Realm/RLMMigration.h
@@ -89,10 +89,11 @@ typedef void (^RLMObjectMigrationBlock)(RLMObject *oldObject, RLMObject *newObje
 
 /**
  Deletes the data for the class with the given name.
+ This is used to clean up as removing RLMObject subclasses.
  
- @param   name The name of the RLMObject class to delete.
+ @param  name The name of the RLMObject class to delete.
  
- @return whether data was deleted.
+ @return whether there was any data to delete.
  */
 - (BOOL)deleteDataForClassName:(NSString *)name;
 
--- a/Realm/RLMMigration.h
+++ b/Realm/RLMMigration.h
@@ -87,6 +87,15 @@ typedef void (^RLMObjectMigrationBlock)(RLMObject *oldObject, RLMObject *newObje
  */
 - (void)deleteObject:(RLMObject *)object;
 
+/**
+ Deletes the table for the class with the given name. Throws if that class still is in the schema.
+ 
+ @param   name The name of the RLMObject class to delete.
+ 
+ @return whether the table existed and was deleted.
+ */
+- (BOOL)deleteTableForClassName:(NSString *)name;
+
 @end
 
 
--- a/Realm/RLMMigration.h
+++ b/Realm/RLMMigration.h
@@ -89,7 +89,8 @@ typedef void (^RLMObjectMigrationBlock)(RLMObject *oldObject, RLMObject *newObje
 
 /**
  Deletes the data for the class with the given name.
- This is used to clean up as removing RLMObject subclasses.
+ This deletes all objects of the given class, and if the RLMObject subclass no longer exists in your program,
+ cleans up any remaining metadata for the class in the Realm file.
  
  @param  name The name of the RLMObject class to delete.
  
--- a/Realm/RLMMigration.mm
+++ b/Realm/RLMMigration.mm
@@ -163,4 +163,21 @@
     [_realm deleteObject:object];
 }
 
+- (BOOL)deleteTableForClassName:(NSString *)name {
+    if (!name) {
+        @throw RLMException(@"Must provide a class name.");
+    }
+
+    if ([_realm.schema schemaForClassName:name]) {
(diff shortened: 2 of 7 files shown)


Realm/RLMObjectStore.mm
```diff
@@ -216,7 +216,18 @@ static bool RLMRealmCreateTables(RLMRealm *realm, RLMSchema *targetSchema, bool
         }
     }
 
-    // FIXME - remove deleted tables
+    if (updateExisting) {
+        NSMutableArray *tablesToDelete = [NSMutableArray array];
```
jpsim: You should use an `NSMutableSet` and store the `RLMStringData` instead of the `NSString` to avoid doing another conversion further down.

alazier: Instead of keeping track of tables to delete, you should enumerate in reverse and delete them inline.

alazier:
You need tests for this. Consider changing the migration tests or adding new tests altogether. Would be good to have a test in migrations where data from deprecated objects is inserted into new or existing objects.

Realm/RLMObjectStore.mm
segiddins: > Instead of keeping track of tables to delete, you should enumerate in reverse and delete them inline.

The documentation says that the order of tables is not guaranteed to be stable when you delete a table.

segiddins: > You should use an NSMutableSet and store the RLMStringData instead of the NSString to avoid doing another conversion further down.

Won't an `NSSet` not hold `tightdb::StringData` since it's not an Objective-C object?

jpsim: True, so keep storing the `NSString`s but switch to `NSSet` nonetheless :wink:.

segiddins:
@alazier added a test.

segiddins:
At this point, should I just close this PR?

segiddins:
This has been updated and should be ready for review, @jpsim @bdash.

Realm/Tests/MigrationTests.mm
```diff
@@ -143,6 +144,40 @@ - (void)testPerRealmMigration {
...
+    }
+
+    @autoreleasepool {
+        // verify migration
+        RLMRealm *realm = [self realmWithTestPath];
+        XCTAssertFalse(realm.group->has_table(RLMStringDataWithNSString(RLMTableNameForClass(@"DeletedClass"))), @"The deleted class should not have a table.");
+    }
+
+    [RLMRealm setSchemaVersion:0 forRealmAtPath:RLMTestRealmPath() withMigrationBlock:nil];
+    XCTAssertThrows([RLMRealm migrateRealmAtPath:RLMTestRealmPath()]);
```
bdash: This last assertion doesn't appear to be testing anything related to this change.

bdash:
Changelog entry?

:+1:

segiddins:
\c @tgoyne

tgoyne:
We don't use the word "table" in our public API. I don't have any great suggestions, but maybe something like `deleteDataForRemovedObjectClass:`?

jpsim:
I agree we should continue to avoid talking about tables in our public API, unless absolutely necessary. I'd consider that an implementation detail.

```objc
/**
 Deletes the all data for the class with the given name from the Realm.
 Throws if that class is still in the active schema.
 
 @param className The name of the RLMObject subclass to delete.
 
 @return whether the data was deleted. Returns `NO` if the file does not contain that class.
 */
- (BOOL)deleteDataForClassName:(NSString *)className;
```

This API should also be available from Realm Swift.

tgoyne:
That name suggests that it should clear the table rather than being an error if the object class still exists in the current schema. I think I like that, but I haven't thought too much about if there's a case where it'd be overly dangerous.

segiddins:
`deleteMetadataForClassName` ?

tgoyne:
It's not just metadata that it deletes.

Realm/RLMMigration.mm
```diff
@@ -163,4 +163,21 @@ - (void)deleteObject:(RLMObject *)object {
...
+
+    if ([_realm.schema schemaForClassName:name]) {
+        @throw RLMException([NSString stringWithFormat:@"Cannot delete table for class name '%@', since it is in use.", name]);
+    }
+
+    size_t table = _realm.group->find_table(RLMStringDataWithNSString(RLMTableNameForClass(name)));
+    if (table == realm::not_found) {
+        return false;
+    }
+    _realm.group->remove_table(table);
```
segiddins: also need to remove primary key

segiddins:
This has been updated

Realm/RLMMigration.h
```diff
@@ -87,6 +87,15 @@ typedef void (^RLMObjectMigrationBlock)(RLMObject *oldObject, RLMObject *newObje
  */
 - (void)deleteObject:(RLMObject *)object;
 
+/**
+ Deletes the data for the class with the given name.
+ 
+ @param   name The name of the RLMObject class to delete.
+ 
+ @return whether data was deleted.
```
tgoyne: Whether there was any data to delete.

Realm/RLMMigration.h
```diff
@@ -87,6 +87,15 @@ typedef void (^RLMObjectMigrationBlock)(RLMObject *oldObject, RLMObject *newObje
  */
 - (void)deleteObject:(RLMObject *)object;
 
+/**
+ Deletes the data for the class with the given name.
```
tgoyne: Should mention that you use this to clean up after removing classes.

segiddins:
@tgoyne @jpsim @bdash this look good?

Realm/RLMMigration.h
```diff
@@ -87,6 +87,16 @@ typedef void (^RLMObjectMigrationBlock)(RLMObject *oldObject, RLMObject *newObje
...
+ This is used to clean up as removing RLMObject subclasses.
```
jpsim: This is used to clean up removed RLMObject subclasses from the realm file.

tgoyne: This deletes all objects of the given class, and if the RLMObject subclass no longer exists in your program, cleans up any remaining metadata for the class in the Realm file.

jpsim:
:+1:

segiddins:
I don't believe the test failure is related?

jpsim:
No, test failure is unrelated.

tgoyne:
:+1:

PR #1412, opened 2015-01-26
Fix accessor creation when the first Realm opened is read-only

tgoyne:
When a read-only Realm is missing tables, we can't read the column order from the table, so we were creating accessors where every property was using column 0, and then reusing those accessors for Realms opened later which do have the tables. This fixes the problem by explicitly checking `_column` in `isEqualToProperty:` rather than relying on the order of the properties in the array, and as an optimization skips creating the accessors for missing tables.

Fixes https://secure.helpscout.net/conversation/67013100/364/ and hopefully the other reports of column type mismatches.

@alazier

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,19 @@
+x.x.x Release notes (yyyy-MM-dd)
+=============================================================
+
+### API breaking changes
+
+* None.
+
+### Enhancements
+
+* None.
+
+### Bugfixes
+
+* Fix incorrect column type assertions when the first Realm file opened is a
+  read-only file that is missing tables.
+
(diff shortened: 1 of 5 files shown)


alazier:
Is this something users have been hitting? If unsere, can you ping the users who have hit the column assertion and ask if they are using real-only realms? Would be nice to know if this issue is resolved or not.

tgoyne:
I linked the helpscout thread where a user hit this and gave sample code...

alazier:
I thought we had multiple users hitting the col_type_StringEnum assertion. Is this not the case?

tgoyne:
We've had several people report similar problems, yes.

alazier:
:+1:

PR #1414, opened 2015-01-26
Throw an exception when adding an invalidated or deleted object as a link

jpsim:
Fixes https://secure.helpscout.net/conversation/65032812/293/. @alazier @tgoyne


alazier:
:+1:

PR #1415, opened 2015-01-27
Improvements to RLMObject create* methods when argument is an RLMObject

### LINKED ISSUE #1409: Passing a realm-backed RLMObject to createOrUpdate should throw an exception
Otherwise, people try to do this:

```swift
class StringInt: RLMObject {
    dynamic var string = ""
    dynamic var int = 0

    override class func primaryKey() -> String! {
        return "string"
    }
}

let realm = RLMRealm.defaultRealm()
realm.beginTransaction()
let object = StringInt.createInRealm(realm, withObject: ["0", 0])
object.int = 1 // This already updates the object
StringInt.createOrUpdateInRealm(realm, withObject: object) // This should throw an exception
realm.commitTransaction()
```

Thoughts @tgoyne @alazier ?

jpsim:
- `RLMCreateObjectInRealmWithValue()` is now a no-op if value is an RLMObject
  of the same type already backed by the target realm.
- Moved check for primary key from `createOrUpdateInDefaultRealmWithObject:` to
  `createOrUpdateInRealm:withObject:`

Fixes #1409. @alazier @tgoyne

### DIFF
--- a/Realm/RLMObject.h
+++ b/Realm/RLMObject.h
@@ -156,7 +156,7 @@
  is returned. Otherwise this creates and populates a new instance of this object in the default Realm.
  
  If nested objects are included in the argument, `createOrUpdateInDefaultRealmWithObject:` will be
- called on them.
+ called on them if have a primary key (`createInDefaultRealmWithObject:` otherwise).
  
  This is a no-op if the argument is an RLMObject of the same type already backed by the target realm.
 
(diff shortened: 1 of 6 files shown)


Realm/RLMObject.h
```diff
@@ -148,6 +154,11 @@
  This method can only be called on object types with a primary key defined. If there is already
  an object with the same primary key value in the default RLMRealm its values are updated and the object
  is returned. Otherwise this creates and populates a new instance of this object in the default Realm.
+ 
+ If nested objects are included in the argument, `createOrUpdateInDefaultRealmWithObject:` will be
```
tgoyne: Should mention that linked objects aren't required to have primary keys.

jpsim: Good call, fixed.

tgoyne:
:+1:

PR #1417, opened 2015-01-27
Document Swift interface

jpsim:
Temporary PR to compare `seg-swift-docs` against `mr-swift`. Will force push to `seg-swift-docs` once `mr-swift` is merged into `al-swift`.


jpsim:
`mr-swift` is not ready to be merged. Closing this in favor of merging `seg-swift-docs` as the next step in this Swiftventure.

Realm/RealmSwift/List.swift
```diff
@@ -72,28 +124,68 @@ public final class List<T: Object>: ListBase, SequenceType {
...
+    }
+
+    /**
+    Returns `Results` containing list elements sorted by the given sort descriptors.
+
+    :param: sortDescriptors `SortDescriptor`s to sort by.
+
+    :returns: `Results` containing list elements sorted by the given sort descriptors.
+    */
+    public func sorted(sortDescriptors: [SortDescriptor]) -> Results<T> {
```
alazier: Need a test for multi sort.

Realm/RealmSwift/List.swift
```diff
@@ -105,50 +197,115 @@ public final class List<T: Object>: ListBase, SequenceType {
...
-        for obj in objects {
-            _rlmArray.addObject(unsafeBitCast(obj as T, RLMObject.self))
+    /**
+    Appends the objects in the given sequence to the end of the list.
+
+    :warning: This method can only be called during a write transaction.
+
+    :param: objects A sequence of objects.
+    */
+    public func append<S: SequenceType where S.Generator.Element == T>(objects: S) {
```
alazier: Can you add more details of what we are working around/what it should look like otherwise? Do our existing tests work for this?

Realm/RealmSwift/Migration.swift
```diff
@@ -54,23 +54,26 @@ public class Migration {
...
+    internal init(_ rlmMigration: RLMMigration) {
         self.rlmMigration = rlmMigration
     }
 }
 
 public class MigrationObject : Object {
-    subscript(key: String) -> AnyObject? {
-        if (self.objectSchema[key].type == RLMPropertyType.Array) {
-            return listProperties[key]
+    public override subscript(key: String) -> AnyObject? {
```
alazier: Did you clear build/DerivedData and run the MigrationExample after making these changes? Would be nice to know we didn't break anything. Would also be prudent to run all of the other examples.

Realm/RealmSwift/List.swift
jpsim: All the relevant details are in the type constraint, but I can break it down here if you like:

`S: SequenceType where S.Generator.Element == T` means that `objects` can only be a type that conforms to `SequenceType` where its elements are all the same type as the objects already stored in this `List`. So if someone creates a custom `LinkedList<Object>` type that conforms to `SequenceType`, this will just work.

Realm/RealmSwift/Migration.swift
```diff
@@ -54,23 +54,26 @@ public class Migration {
...
+        get {
+            if (objectSchema[key].type == .Array) {
```
alazier: Please revert this change to subscript. The setter does not work for list properties. This should be handled in a different pr.

Realm/RealmSwift/Migration.swift
jpsim: I ran the tests, but I assume those aren't quite ready yet for the new Migration stuff?

PR #1419, opened 2015-01-27
Update framework copyright notice date range

pizthewiz:
Given how small the change is, I thought a PR might be preferred to an issue.


bmunkholm:
Thanks for the consideration!
But actually it's not needed to update copyright dates. The date merely says when the copyright was first claimed. And it's still valid even after the date. See also this [SO Q/A](http://stackoverflow.com/questions/2390230/do-copyright-dates-need-to-be-updated) for reference.

PR #1420, opened 2015-01-27
Add an example app that uses extensions

### LINKED ISSUE #1377: Interprocess: shared container sample app
Test out interprocess pr in a sample app using shared containers.

segiddins:
Closes #1377.

\c @tgoyne @jpsim


examples/ios/objc/Extension/AppDelegate.h
```diff
@@ -0,0 +1,17 @@
+//
```
tgoyne: all new files need a proper license header

examples/ios/objc/Extension/AppDelegate.m
```diff
@@ -0,0 +1,81 @@
...
+    return YES;
+}
+
+@end
+
+@implementation TickViewController
+
+- (void)viewDidLoad {
+    [super viewDidLoad];
+    [RLMRealm setDefaultRealmPath:[[[NSFileManager defaultManager] containerURLForSecurityApplicationGroupIdentifier:@"group.io.realm.examples.extension"] URLByAppendingPathComponent:@"extension.realm"].path];
```
tgoyne: Shouldn't need to do this twice.

examples/ios/objc/RealmExamples.xcodeproj/project.pbxproj
```diff
@@ -737,6 +1125,192 @@
...
+			buildSettings = {
+				ASSETCATALOG_COMPILER_APPICON_NAME = AppIcon;
+				CLANG_WARN_UNREACHABLE_CODE = YES;
+				CODE_SIGN_ENTITLEMENTS = extension/extension.entitlements;
+				CODE_SIGN_IDENTITY = "iPhone Developer";
+				"CODE_SIGN_IDENTITY[sdk=iphoneos*]" = "iPhone Developer";
+				ENABLE_STRICT_OBJC_MSGSEND = YES;
+				FRAMEWORK_SEARCH_PATHS = (
+					"$(inherited)",
+					"/Users/segiddins/Development/Realm/realm-cocoa/build/ios",
```
tgoyne: path

examples/ios/objc/Extension/AppDelegate.m
```diff
@@ -0,0 +1,81 @@
...
+}
+
+- (void)viewDidAppear:(BOOL)animated {
+    [super viewDidAppear:animated];
+    [self tock];
+    [self updateLabel];
+}
+
+- (void)updateLabel {
+    [self.button setTitle:[NSString stringWithFormat:@"%ld", (long)self.tick.count] forState:UIControlStateNormal];
```
tgoyne: [self.button setTitle:@(self.tick.count).stringValue forState:UIControlStateNormal]

examples/ios/objc/Extension/AppDelegate.h
segiddins: believe I hit all the new source files

Realm/Tests/PerformanceTests.m
```diff
@@ -29,7 +29,7 @@ + (RLMPropertyAttributes)attributesForProperty:(__unused NSString *)propertyName
 }
 @end
 
-#if !DEBUG
+//#if !DEBUG
```
jpsim: This is temporary for CI, I hope?

examples/ios/objc/TodayExtension/TodayViewController.m
```diff
@@ -0,0 +1,81 @@
...
+    self.preferredContentSize = CGSizeMake(0, 200.0);
+    [RLMRealm setDefaultRealmPath:[[[NSFileManager defaultManager] containerURLForSecurityApplicationGroupIdentifier:@"group.io.realm.examples.extension"] URLByAppendingPathComponent:@"extension.realm"].path];
+    self.tick = [Tick allObjects].firstObject;
+    if (!self.tick) {
+        [[RLMRealm defaultRealm] transactionWithBlock:^{
+            self.tick = [Tick createInDefaultRealmWithObject:@[@"", @0]];
+        }];
+    }
+    self.notificationToken = [self.tick.realm addNotificationBlock:^(NSString *notification, RLMRealm *realm) {
+        if (self.tick.count % 19 == 0) {
```
jpsim: could you add an explanatory comment here, and all other notification blocks doing something similar?

Realm/Tests/PerformanceTests.m
segiddins: Ah dammit, I think I messed up the rebase. I'll fix.

jpsim:
You should add these targets to `build.sh examples`, which would make sure they continue to build.

Adding comments explaining the rationale behind the modulo's would be nice, so we don't forget when we look at this again in a few months. It would also be nicer if the 13/17/19 values were kept in a single `Constants.h` file shared by all three targets.

Otherwise, :+1:

segiddins:
The 13/17/19 are just three low prime numbers, so I don't see what use putting them in a Constants file will be. But I will add some comments.

jpsim:
> The 13/17/19 are just three low prime numbers, so I don't see what use putting them in a Constants file will be.

Sure, doesn't matter.

segiddins:
> You should add these targets to build.sh examples, which would make sure they continue to build.

Done.

Realm/Tests/PerformanceTests.m
tgoyne: The `#if !DEBUG` makes them not show up in the Xcode UI, and this is rebased on a WIP branch that I need to be able to test.

segiddins: @tgoyne it should still be there, but I fixed the rebase

PR #1423, opened 2015-01-28
Centralize exception creation

### LINKED ISSUE #1372: Using single code path for creating exceptions/Add library version to user info
Add the version text of the cocoa binding to all exceptions thrown from core. In RLMRealm:throw_objc_exception()
In likely doesn't make sense to include it in make_realm_error.

### LINKED ISSUE #1374: Centralize exception creation in a single function
Will make addressing https://github.com/realm/realm-cocoa/issues/1372 much easier.

segiddins:
Closes #1374.
Closes #1372.

\c @alazier

### DIFF
--- a/Realm/RLMRealm.mm
+++ b/Realm/RLMRealm.mm
@@ -294,7 +294,7 @@ NSString * const c_defaultRealmFileName = @"default.realm";
                              "architecture. For sharing files between the Realm "
                              "Browser and an iOS simulator, this means that you "
                              "must use a 64-bit simulator.";
-            error = [NSError errorWithDomain:@"io.realm"
+            error = [NSError errorWithDomain:RLMErrorDomain
                                         code:RLMErrorIncompatibleLockFile
                                     userInfo:@{NSLocalizedDescriptionKey: err,
                                                @"Error Code": @(RLMErrorIncompatibleLockFile)}];
--- a/Realm/RLMRealm.mm
+++ b/Realm/RLMRealm.mm
@@ -68,12 +68,6 @@ using namespace std;
 using namespace tightdb;
 using namespace tightdb::util;
 
-
-// create NSException from c++ exception
-static __attribute__((noreturn)) void throw_objc_exception(exception &ex) {
-    @throw RLMException(ex);
-}
-
 // create NSError from c++ exception
 static NSError *make_realm_error(RLMError code, exception const& ex) {
     return [NSError errorWithDomain:@"io.realm"
@@ -605,7 +599,7 @@ static void CheckReadWrite(RLMRealm *realm, NSString *msg=@"Cannot write to a re
             // File access errors are treated as exceptions here since they should not occur after the shared
             // group has already been successfully opened on the file and memory mapped. The shared group constructor handles
             // the excepted error related to file access.
-            throw_objc_exception(ex);
+            @throw RLMException(ex);
         }
     } else {
         @throw RLMException(@"The Realm is already in a write transaction");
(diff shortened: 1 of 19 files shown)


Realm/RLMUtil.hpp
```diff
@@ -27,6 +27,9 @@
 
 @class RLMProperty;
 
+NSException *RLMException(NSString *message, NSDictionary *userInfo = nil);
+NSException *RLMException(std::exception exception);
```
tgoyne: `std::exception const& exception`

Passing polymorphic types by value doesn't work (you get a `std::exception` copy inside the function rather than a copy with the original type).

Realm/RLMUtil.mm
```diff
@@ -238,3 +237,18 @@ id RLMValidatedObjectForProperty(id obj, RLMProperty *prop, RLMSchema *schema) {
     }
     return outArray;
 };
+
+NSException *RLMException(NSString *reason, NSDictionary *userInfo) {
+    NSMutableDictionary *info = [NSMutableDictionary dictionaryWithDictionary:userInfo];
```
tgoyne: Does this actually work when `userInfo` is `nil`? I would expect it to throw.

segiddins: That's what `dictionaryWithDictionary` is for, as opposed to `mutableCopy` -- it works when the parameter is `nil`.

Realm/RLMUtil.hpp
segiddins: done

Realm/RLMUtil.mm
alazier: Can you add a test for this?

Realm/RLMRealm.mm
```diff
@@ -71,8 +71,7 @@ - (void)notify;
 
 // create NSException from c++ exception
 static __attribute__((noreturn)) void throw_objc_exception(exception &ex) {
```
alazier: We should just remove this method and throw RLMException wherever this is called.

Realm/RLMRealm.mm
```diff
@@ -71,8 +71,7 @@ - (void)notify;
...
-    NSString *errorMessage = [NSString stringWithUTF8String:ex.what()];
-    @throw [NSException exceptionWithName:@"RLMException" reason:errorMessage userInfo:nil];
+    @throw RLMException(ex);
 }
 
 // create NSError from c++ exception
```
alazier: Should convert make_realm_error to RLMError following the pattern used with RLMException.

Realm/RLMRealm.mm
```diff
@@ -88,9 +87,7 @@ static void setOrThrowError(NSError *error, NSError **outError) {
         *outError = error;
```
alazier: Could make this method a utililty methods as well.

alazier:
Overall looks good. Lets try to finish this up today to minimize conflicts.

Realm/RLMUtil.mm
segiddins: There are currently no tests for the functionality in `RLMUtil`, should I add a new test file?

segiddins:
@alazier added test coverage.

alazier:
Couldn't we just make a function which returns bool and XCTAssertTrue? Would be a bit less confusing I think. Otherwise everything looks good to me.

segiddins:
> Couldn't we just make a function which returns bool and XCTAssertTrue? Would be a bit less confusing I think.

Done.

alazier:
:+1:

PR #1427, opened 2015-01-29
[WIP] Add missing aspects of Migration.swift

jpsim:


Realm/RealmSwift/Migration.swift
```diff
@@ -102,18 +102,58 @@ public func setSchemaVersion(schemaVersion: UInt, realmPath: String, migrationBl
...
+Get the schema version for a Realm at a given path.
+
+:param: realmPath Path to a Realm file.
+:param: error     If an error occurs, upon return contains an `NSError` object
+                  that describes the problem. If you are not interested in
+                  possible errors, omit the argument, or pass in `nil`.
+
+:returns: The version of the Realm at `realmPath` or `nil` if the version cannot be read.
+*/
+public func schemaVersionAtPath(realmPath: String, error: NSErrorPointer = nil) -> UInt? {
```
segiddins: should return an `Int`?

Realm/RealmSwift/Migration.swift
```diff
@@ -102,18 +102,58 @@ public func setSchemaVersion(schemaVersion: UInt, realmPath: String, migrationBl
...
+:param: encryptionKey 64-byte encryption key if Realms at the given path are encrypted.
+                      If the Realms at the given path are not encrypted, omit the argument or pass
+                      in `nil`.
 
 :returns: `nil` if the migration was successful, or an `NSError` object that describes the problem
           that occured otherwise.
 */
-public func migrateRealm(path: String) -> NSError? {
+public func migrateRealm(path: String, encryptionKey: NSData? = nil) -> NSError? {
     return RLMRealm.migrateRealmAtPath(path)
```
segiddins: the `encryptionKey` needs to be used?

Realm/RealmSwift/Migration.swift
```diff
@@ -123,7 +163,7 @@ of a `Realm` instance.
 
 This object provides access to the previous and current `Schema`s for this migration.
 */
-public class Migration {
+public final class Migration {
```
segiddins: @jpsim weren't you thinking of making `Migration` a `struct`?

Realm/RealmSwift/Migration.swift
jpsim: I'd prefer sticking to `UInt` to represent natural numbers, unless there's a strong parallel that can be made with a similar interface in the Swift standard library that does otherwise (e.g. `Array` subscript takes an `Int`).

PR #1431, opened 2015-01-29
Fix for invalid property comparison and reuse during column reorder and migrations

alazier:
Still needs more testing.

@tgoyne

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,10 @@
+0.90.4 Release notes (2015-01-29)
+=============================================================
+
+### Bugfixes
+
+* Fix bug that resulted in columns being dropped and recreated during migrations.
+
 0.90.3 Release notes (2015-01-27)
 =============================================================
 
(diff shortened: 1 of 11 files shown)


tgoyne:
Can the copy for the accessor cache also be a shallow copy? It seems like it should be able to.

tgoyne:
Added some tests which fail on master that this fixes.

alazier:
@tgoyne  The accessor cache schema can be used directly and not mutated.

In general this set of optimizations worries me. They seem to easily risk unintended consequences with future changes. How large is the gain exactly? If schema copying is really a huge deal we can probably remove some of the copies done in RLMCreateTables as well.

tgoyne:
The copy for the accessor class is probably totally irrelevant performance-wise. I was mostly just wondering if I was missing something that made it necessary.

alazier:
I was talking about this set of optimizations, on the accessor on in particular.

I made the change to use the accessor schema directly. Since it is't mutated this should be fine. Also remove the copy from in RLMCreateTables as well. This speed up the testRealmCreation test by 5-10% on the mac - not sure what it will do on the phone.

I'm pretty sure this is safe - ie we make a deep copy if anyone ever needs to mutate properties, and shallow copy for separate table caches, but please review carefully.

tgoyne:
:+1:

PR #1434, opened 2015-01-29
[RLMProperty] Re-add isEqualToProperty:

segiddins:
Restores the user-facing `isEqualToProperty:` method

\c @alazier @tgoyne


alazier:
how about a test?

segiddins:
I'm unsure of the best way to write tests for this, since the public initializer wouldn't expose manipulating the things we'd like to test (such as the fact that we ignore the `column` property).

segiddins:
@alazier added a pair of very basic tests

PR #1437, opened 2015-01-30
renamed reference to wrong method in ref-doc

bmunkholm:
@tgoyne

### DIFF
--- a/Realm/RLMRealm.h
+++ b/Realm/RLMRealm.h
@@ -122,7 +122,7 @@
  Set the encryption key to use when opening Realms at a certain path.
 
  This can be used as an alternative to explicitly passing the key to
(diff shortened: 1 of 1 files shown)


jpsim:
Thanks @bmunkholm!

Realm/RLMRealm.h
```diff
@@ -122,7 +122,7 @@
  Set the encryption key to use when opening Realms at a certain path.
 
  This can be used as an alternative to explicitly passing the key to
- `encryptedRealmWithPath:key:readOnly:error:` each time a Realm instance is
+ `realmWithPath:key:readOnly:error:` each time a Realm instance is
```
segiddins: I believe the method is actually `realmWithPath:encryptionKey:readOnly:error:`

PR #1440, opened 2015-01-30
Replace attributes by property indexed

### LINKED ISSUE #935: Replace object attributes with single method to specify indexed columns

mrackwitz:
@tgoyne 

Closed #935


tgoyne:
:+1: once CI passes.

PR #1446, opened 2015-02-03
[RLMUtil] Only try to import RLMVersion.h if REALM_VERSION is undefined.

segiddins:
Fixes building with CocoaPods.

Verified this fixes things with `pod lib lint`.

\c @tgoyne @mrackwitz

### DIFF
--- a/Realm/RLMUtil.mm
+++ b/Realm/RLMUtil.mm
@@ -23,8 +23,11 @@
 #import "RLMObject_Private.h"
 #import "RLMProperty_Private.h"
 #import "RLMSwiftSupport.h"
-#import "RLMVersion.h"
 #import "RLMSchema_Private.h"
+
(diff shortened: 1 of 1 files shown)


tgoyne:
:+1:

alazier:
It would be better to actually put this check into RLMVersion.h.

Alternatively `#pragma once` in RLMVersion.h would also fix the issue.

segiddins:
RLMVersion.h doesn't exist when installed via CocoaPods.

tgoyne:
It might be nicer to generate RLMVersion.h in `prepare_command` rather than defining REALM_VERSION directly now that the version is used in multiple places.

segiddins:
Sure, we could have the `cocoapods-setup` command create `RLMVersion.h` in `include/Realm`.

PR #1450, opened 2015-02-03
Swift migration improvements and testing

alazier:


RealmSwift/Migration.swift
```diff
@@ -102,19 +102,44 @@ public func setSchemaVersion(schemaVersion: UInt, realmPath: String, migrationBl
...
+                      If the Realms at the given path are not encrypted, omit the argument or pass
+                      in `nil`.
 
 :returns: `nil` if the migration was successful, or an `NSError` object that describes the problem
           that occured otherwise.
 */
-public func migrateRealm(path: String) -> NSError? {
-    return RLMRealm.migrateRealmAtPath(path)
+public func migrateRealm(path: String, encryptionKey: NSData? = nil) -> NSError? {
+    if (encryptionKey == nil) {
```
segiddins: why the parens around the condition?

RealmSwift/ObjectSchema.swift
```diff
@@ -33,7 +33,9 @@ public class ObjectSchema: Equatable {
     var rlmObjectSchema: RLMObjectSchema
 
     /// Array of persisted `Property` objects for an object.
-    public var properties: [Property] { return rlmObjectSchema.properties as [Property] }
+    public var properties: [Property] {
+        return (rlmObjectSchema.properties as [RLMProperty]).map( { Property(rlmProperty: $0) } )
```
segiddins: use the same spacing here as on Schema

alazier:
@tgoyne @segiddins 

Only way I was able to make this work was to add configurations to build the dynamic frameworks. I was unable to make xcode behave properly with any other strategies. If you have other suggestions on how to make this work please share. Otherwise this is ready for review.

RealmSwift/Tests/MigrationTests.swift
```diff
@@ -0,0 +1,215 @@
...
+        }
+        else {
+            migrateRealm(realmPath, encryptionKey: nil)
+        }
+
+        XCTAssertEqual(didRun, shouldRun)
+    }
+
+    // migrate default realm and ensure migration
+    private func migrateAndTestDefaultRealm(shouldRun: Bool = true, schemaVersion: UInt = 1, autoMigration: Bool = false, block: MigrationBlock? = nil) {
```
segiddins: why not make this call into `migrateAndTestRealmAtPath` with the `defaultRealmPath`?

alazier: this tests the `setDefaultRealmSchemaVersion` method and makes sure `defaultRealm()` performs an automigration

alazier: We could still combine the methods though, just not sure it's worth it

segiddins: :+1: just as long as it serves a purpose!

alazier: I will just remove the default version and test the methods in there separately.

alazier:
@tgoyne @segiddins ready for review

Realm/RLMObjectStore.mm
```diff
@@ -289,18 +289,23 @@ static inline void RLMVerifyInWriteTransaction(RLMRealm *realm) {
...
-                                                                        view:object->_row.get_linklist(prop.column)
-                                                                       realm:object->_realm];
-                if (object.class == NSClassFromString(@"RealmSwift.MigrationObject")) {
-                    [(id<RLMSwiftMigrationObject>)object initalizeListPropertyWithName:prop.name rlmArray:array];
-                }
-                else {
-                    auto list = static_cast<RLMListBase *>(object_getIvar(object, prop.swiftListIvar));
-                    list._rlmArray = array;
+            if (prop.type == RLMPropertyTypeArray) {
+                static Class s_swiftMigrationObjectClass =  NSClassFromString(@"RealmSwift.MigrationObject");
```
tgoyne: extra space after =

Realm/RLMObjectStore.mm
```diff
@@ -289,18 +289,23 @@ static inline void RLMVerifyInWriteTransaction(RLMRealm *realm) {
 
 static inline void RLMInitializeSwiftListAccessor(RLMObjectBase *object) {
     // switch List<> properties to linkviews from standalone arrays
-    if ([object isKindOfClass:NSClassFromString(@"RealmSwift.Object")]) {
+    static Class s_swiftObjectClass = NSClassFromString(@"RealmSwift.Object");
```
tgoyne: This function is very deeply nested and would benefit from some early returns/continues.

RealmSwift/Migration.swift
```diff
@@ -102,19 +102,44 @@ public func setSchemaVersion(schemaVersion: UInt, realmPath: String, migrationBl
 }
 
 /**
+Get the schema version for an encrypted Realm at a given path.
```
tgoyne: Also non-encrypted since the key is optional?

RealmSwift/Migration.swift
```diff
@@ -176,26 +201,36 @@ public class Migration {
...
-            return listProperties[key]
+        get {
+            if (self.objectSchema[key].type == RLMPropertyType.Array) {
+                return listProperties[key]
+            }
+            return super[key]
+        }
+        set(value) {
+            if (self.objectSchema[key].type == RLMPropertyType.Array) {
+                fatalError("Setting List properties during migrations is unsupported.")
```
tgoyne: Maybe suggest adding/removing objects from the existing list property?

RealmSwift/Tests/TestCase.swift
```diff
@@ -20,7 +20,7 @@ import XCTest
 import RealmSwift
 import Realm
 
-private func testRealmPath() -> String {
+public func testRealmPath() -> String {
```
tgoyne: internal?

RealmSwift/Migration.swift
```diff
@@ -176,26 +201,36 @@ public class Migration {
...
+            }
+            else {
```
segiddins: this `else` is redundant

RealmSwift/Tests/MigrationTests.swift
```diff
@@ -0,0 +1,281 @@
...
+        // refresh to update realm
+        defaultRealm().refresh()
+
+        // check edited values
+        let object = objects(SwiftObject.self).first!
+        XCTAssertEqual(object.boolCol, false)
+        XCTAssertEqual(object.intCol, 1)
+        XCTAssertEqual(object.floatCol, 1.0 as Float)
+        XCTAssertEqual(object.doubleCol, 10.0)
+        XCTAssertEqual(object.binaryCol, NSData(bytes: "b", length: 0))
```
segiddins: that has a `length` of `1`?

Realm/RLMObjectStore.mm
```diff
@@ -289,19 +289,25 @@ static inline void RLMVerifyInWriteTransaction(RLMRealm *realm) {
...
-                }
-                else {
-                    auto list = static_cast<RLMListBase *>(object_getIvar(object, prop.swiftListIvar));
-                    list._rlmArray = array;
-                }
+                [(id<RLMSwiftMigrationObject>)object initalizeListPropertyWithName:prop.name rlmArray:array];
+            }
+            else {
+                auto list = static_cast<RLMListBase *>(object_getIvar(object, prop.swiftListIvar));
+                list._rlmArray = [RLMArrayLinkView arrayWithObjectClassName:prop.objectClassName
```
segiddins: this array creation logic appears to be duplicated?

alazier: I think this is fine. Better than doing the conditional twice as we were before. Not enough code to warrant another helper function.

PR #1451, opened 2015-02-04
Add assertThrows to RealmSwift tests

segiddins:
\c @alazier @tgoyne


alazier:
We should use this somewhere to make sure it works.

segiddins:
@alazier I manually verified that it works

Issue #1455, opened 2015-02-04
setSchemaVersion and migrateRealm should throw if Realm is already open


alazier:
Right now opened Realms continue to work but are not refreshed automatically.

PR #1456, opened 2015-02-04
[RLMRealm] Throw an exception when trying to set schema version or migra...

segiddins:
...te a realm when that realm is already open

Closes https://github.com/realm/realm-cocoa/issues/1455

\c @alazier @tgoyne

### DIFF
--- a/Realm/RLMRealm.mm
+++ b/Realm/RLMRealm.mm
@@ -509,6 +509,10 @@ static id RLMAutorelease(id value) {
 }
 
 + (void)setEncryptionKey:(NSData *)key forRealmsAtPath:(NSString *)path {
+    if (realmsOpenAtPath(path)) {
+        @throw RLMException(@"Cannot set encryption key for Realms that are already open.");
+    }
+
     setKeyForPath(validatedKey(key), path);
 }
 
--- a/Realm/RLMRealm.mm
+++ b/Realm/RLMRealm.mm
@@ -158,6 +158,10 @@ static NSArray *realmsAtPath(NSString *path) {
     }
 }
 
+static bool realmsOpenAtPath(NSString *path) {
+    return realmsAtPath(path).count > 0;
+}
+
 static void clearRealmCache() {
     @synchronized(s_realmsPerPath) {
         for (NSMapTable *map in s_realmsPerPath.allValues) {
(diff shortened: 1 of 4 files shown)


Realm/RLMRealm.mm
```diff
@@ -807,6 +807,10 @@ + (void)setDefaultRealmSchemaVersion:(NSUInteger)version withMigrationBlock:(RLM
 }
 
 + (void)setSchemaVersion:(NSUInteger)version forRealmAtPath:(NSString *)realmPath withMigrationBlock:(RLMMigrationBlock)block {
+    if (realmsAtPath(realmPath)) {
```
alazier: You want to check count of this object

tgoyne: This check doesn't work. `realmsAtPath` can return an empty array.

Realm/RLMRealm.mm
```diff
@@ -850,6 +854,10 @@ + (NSError *)migrateRealmAtPath:(NSString *)realmPath encryptionKey:(NSData *)ke
 }
 
 + (NSError *)migrateRealmAtPath:(NSString *)realmPath key:(NSData *)key {
+    if (realmsAtPath(realmPath)) {
```
alazier: check count of this object

Realm/Tests/MigrationTests.mm
```diff
@@ -102,6 +103,8 @@ - (void)testGetSchemaVersion {
         [RLMRealm defaultRealm];
     }
 
+    [RLMRealm resetRealmState];
```
alazier: we shouldn't have to call this - better way to clear the cache is to do previous operations in an autoreleasepool. This ensures the RLMRealm objects have actually been disposed of.

tgoyne:
Having to call `resetRealmState` to get things out of the cache always indicates you have a bug somewhere.

segiddins:
Updated

alazier:
check when setting encryption key as well

alazier:
otherwise looks fine

alazier:
need a test though

segiddins:
@alazier test coverage added

alazier:
need to update the CHANGELOG

segiddins:
CHANGELING entry added

alazier:
:+1:

tgoyne:
:+1:

mclark4386:
Um... anyone else having issue with this in Swift since in swift you can't try/catch to handle this gracefully if the realm is already open?

PR #1457, opened 2015-02-05
fixes for ci

alazier:
@segiddins


Realm/Tests/DynamicTests.m
```diff
@@ -18,7 +18,9 @@
 
 #import "RLMTestCase.h"
 
-@import Realm.Private;
+//@import Realm.Private;
```
segiddins: Should just be deleted?

alazier:
@tgoyne @segiddins

Only way I was able to make this work was to add configurations to build the dynamic frameworks. I was unable to make xcode behave properly with any other strategies. If you have other suggestions on how to make this work please share. Otherwise this is ready for review.

PR #1479, opened 2015-02-09
Add OS X tests for doing things with multiple processes

tgoyne:
Will fail on CI due to not using a core build with working wait_for_change().

Any suggestions for other things to test? It covers everything we test for multiple threads along with a few other things that came to mind and an attempt at a brute-force test, but it seems like there must be more that's testable.

@alazier

### DIFF
--- a/Realm/Tests/InterprocessTests.m
+++ b/Realm/Tests/InterprocessTests.m
@@ -1,0 +1,276 @@
+////////////////////////////////////////////////////////////////////////////
+//
+// Copyright 2015 Realm Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+////////////////////////////////////////////////////////////////////////////
+
+#import "RLMMultiProcessTestCase.h"
+
+@interface InterprocessTest : RLMMultiProcessTestCase
+@end
+
+@implementation InterprocessTest
+- (void)testCreateInitialRealmInChild {
+    if (self.isParent) {
+        RLMRunChildAndWait();
+        RLMRealm *realm = [RLMRealm defaultRealm];
+        XCTAssertEqual(1U, [IntObject allObjectsInRealm:realm].count);
+    }
+    else {
+        RLMRealm *realm = [RLMRealm defaultRealm];
+        [realm beginWriteTransaction];
+        [IntObject createInRealm:realm withObject:@[@0]];
+        [realm commitWriteTransaction];
+    }
+}
+
+- (void)testCreateInitialRealmInParent {
+    RLMRealm *realm = [RLMRealm defaultRealm];
+    if (self.isParent) {
+        [realm beginWriteTransaction];
+        [IntObject createInRealm:realm withObject:@[@0]];
+        [realm commitWriteTransaction];
+
+        RLMRunChildAndWait();
+    }
+    else {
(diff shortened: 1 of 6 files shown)


Realm/Tests/RLMMultiProcessTestCase.h
```diff
@@ -0,0 +1,31 @@
...
+@interface RLMMultiProcessTestCase : RLMTestCase
+// if true, this is running the main test process
+@property (nonatomic, readonly) bool isParent;
+
+// spawn a child process running the current test and wait for it complete
+// returns the return code of the process
+- (int)runChildAndWait;
+@end
+
+#define RLMRunChildAndWait() \
```
alazier: This name is a bit confusing - why not make an objc method called ```- (void)runChildAndVerifySuccess```?

tgoyne: The macro is so that errors are reported on the line where RLMRunChildAndWait() appears rather than inside the helper method.

Realm/Tests/RLMMultiProcessTestCase.m
```diff
@@ -0,0 +1,119 @@
...
+    [task launch];
+
+    NSFileHandle *err = [NSFileHandle fileHandleWithStandardError];
+    NSData *delimiter = [@"\n" dataUsingEncoding:NSUTF8StringEncoding];
+    NSMutableData *buffer = [NSMutableData data];
+
+    // Filter the output from the child process to reduce xctest noise
+    while (true) {
+        NSUInteger newline = [buffer rangeOfData:delimiter options:0 range:NSMakeRange(0, buffer.length)].location;
+        if (newline != NSNotFound) {
```
alazier: Should we XCTAssertFail if newLine == NSNotFound?

alazier: Hmm, I'm not quite sure what we are filtering out here. My only concern would be xctest will change its formatting which will break us. Guess there isn't much we can do about that though.

Realm/Tests/RLMMultiProcessTestCase.m
```diff
@@ -0,0 +1,119 @@
...
+            // Skip lines starting with "Test Case", "Test Suite" and "     Executed"
+            const void *b = buffer.bytes;
+            if (newline < 13 || (memcmp(b, "Test Suite", 10) && memcmp(b, "Test Case", 9) && memcmp(b, "     Executed", 13))) {
```
alazier: Since it shouldn't be a performance issue, we could just call `rangeOfString:` for each key phrase we want to filter out.

Realm/Tests/RLMMultiProcessTestCase.m
tgoyne: The child process writing something to stderr that doesn't have a trailing newline would be weird, but not an error.

Now that I look at it again the code is wrong though (it needs to loop until there's no newlines left since `availableData` may return something with multiple lines...).

Realm/Tests/InterprocessTests.m
```diff
@@ -0,0 +1,276 @@
...
+    };
+
+    IntObject *obj = [IntObject allObjects].firstObject;
+    int nextRun = -1;
+
+    // Wait for all of the workers to start up
+    while (obj.intCol < 0) {
+        if (nextRun == -1) {
+            [realm beginWriteTransaction];
+            ++obj.intCol;
```
alazier: This looks so wrong :smile:. Why not obj.intCol++?

alazier:
Overall this looks really good. Some other things to consider testing:
- Contesting write-locks - have parent blocked on child so that notification is fired at beginning of parent's write transaction
- Crashing child - have child crash mid-write transaction and make sure parent can still write after and child has had no effect
- Dueling migrations/schema initializations - ensure our logic is sound if parent and child both try to initialize schema simultaneously. I fear we may assume that there is another thread in the same process which successfully completes this operation?

tgoyne:
> Crashing child - have child crash mid-write transaction and make sure parent can still write after and child has had no effect

I think that doesn't work at the moment due to Darwin not having robust mutexes. If process crashes while holding the write lock then the file has to be fully closed and reopened so that the lock file gets recreated.

tgoyne:
Should at least verify that fully closing and reopening the file after a crash works though.

PR #1482, opened 2015-02-10
set _inWriteTranscation before triggering notifications in beginWriteTransaction

### LINKED ISSUE #1477: _inWriteTransaction should be set to YES before triggering notificaions in beginWriteTransaction
Currently this is not the case, and calling beginWriteTranaction in a notification causes an assert rather than an exception.

alazier:
Fixes #1477

@tgoyne


tgoyne:
:+1:

PR #1489, opened 2015-02-11
writeCopyToPath returns NSError instead of BOOL

alazier:
@tgoyne


tgoyne:
We specifically switched away from this to match the normal Cocoa convention: https://github.com/realm/realm-cocoa/pull/1123

PR #1494, opened 2015-02-12
[Results] Construct via initializer, rather than the free 'objects' function

segiddins:
\c @alazier


segiddins:
We decided we don't like this.

PR #1495, opened 2015-02-12
Add tests that things throw when they should

segiddins:
\c @alazier @tgoyne

### DIFF
--- a/Realm/RLMArrayLinkView.mm
+++ b/Realm/RLMArrayLinkView.mm
@@ -143,6 +143,9 @@ static inline void RLMValidateObjectClass(__unsafe_unretained RLMObject *obj, __
     RLMLinkViewArrayValidateInWriteTransaction(self);
 
     RLMValidateObjectClass(object, self.objectClassName);
+    if (index > _backingLinkView->size()) {
+        @throw RLMException(@"Trying to insert object at invalid index");
+    }
     if (object.realm != self.realm) {
         [self.realm addObject:object];
     }
--- a/RealmSwift/Util.swift
+++ b/RealmSwift/Util.swift
@@ -34,6 +34,6 @@ internal func throwRealmException(message: String, userInfo: [String:AnyObject] 
 
 internal func assertPositive(int: Int, parameterName: String) {
     if int < 0 {
-        throwRealmException("Cannot pass a negative value for '\(parameterName).'")
+        throwRealmException("Cannot pass a negative value for '\(parameterName)'.")
     }
 }
--- a/RealmSwift/Util.swift
+++ b/RealmSwift/Util.swift
@@ -1,0 +1,39 @@
+////////////////////////////////////////////////////////////////////////////
+//
+// Copyright 2015 Realm Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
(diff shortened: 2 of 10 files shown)


RealmSwift/Util.swift
```diff
@@ -0,0 +1,39 @@
...
+        return nil
+    }
+    return Int(index)
+}
+
+internal func throwException(message: String, userInfo: [String:AnyObject] = [:]) {
+    NSException(name: RLMExceptionName, reason: message, userInfo: userInfo).raise()
+}
+
+internal func assertPositive(int: Int, parameterName: String) {
```
alazier: Might be better named `assertIntIsPositive`?

RealmSwift/Util.swift
```diff
@@ -0,0 +1,39 @@
...
+// MARK: Internal Helpers
+
+internal func notFoundToNil(index: UInt) -> Int? {
+    if index == UInt(NSNotFound) {
+        return nil
+    }
+    return Int(index)
+}
+
+internal func throwException(message: String, userInfo: [String:AnyObject] = [:]) {
```
alazier: `throwRealmException`?

RealmSwift/Util.swift
segiddins: No need, since we can always overload on that first parameter.

RealmSwift/Util.swift
segiddins: Inside Realm, I don't know what other kinds of exceptions we'd be throwing. I'll make the change if you really want it,

RealmSwift/Util.swift
alazier: Is there another case where we would actually overload this? When seeing this called in the code it is a bit confusing. I think being a bit more verbose here will improve readability.

RealmSwift/Util.swift
alazier: As is the case with `assertPositive` I just feel the name is a bit vague. It is not obvious what this does to someone unfamiliar with the codebase the first time they see this method. `throwRealmException` is unambiguously clear.

alazier:
Other than naming ambiguity :+1:

RealmSwift/Util.swift
tgoyne: `assertPositive`'s also doubly-wrong: it's a precondition check and not an assertion, and it's checking for not negative numbers, not positive numbers.

Realm/RLMArrayLinkView.mm
```diff
@@ -143,6 +143,9 @@ - (void)insertObject:(RLMObject *)object atIndex:(NSUInteger)index {
     RLMLinkViewArrayValidateInWriteTransaction(self);
 
     RLMValidateObjectClass(object, self.objectClassName);
+    if (index > _backingLinkView->size()) {
```
tgoyne: Should have an obj-c test for this.

RealmSwift/Util.swift
segiddins: @tgoyne in that case, what would you like it to be named?

Realm/RLMArrayLinkView.mm
segiddins: Added

RealmSwift/Util.swift
segiddins: Done

segiddins:
@tgoyne could you take a look at this?

tgoyne:
Still uses the word `assert` for something that is not an assertion. `assertPositive(x)` should be a sanity check in a place where it is not *possible* for `x` to be non-positive, not a precondition check in a place where `x` is not allowed to be non-positive.

segiddins:
@tgoyne changed the naming on `assertPositive`

alazier:
I would prefer the names `ensurePositiveIndex(index: Int)` or `throwForNegativeIndex(index: Int)` and not take a second argument, since it is the same for all cases.

jpsim:
@segiddins please just rename to `throwForNegativeIndex(index: Int)` and let's be done with this. You get my :+1:.

segiddins:
Changed.

PR #1505, opened 2015-02-15
Fix warning in testDataObjectDescription

yuuki1224:
It should be used XCTAssertTrue() instead of XCTAssertNotEqual() when
we confirm the result of rangeOfString, I think because of warning. It
compares int value(NSNotFound) with unsigned int(location property),
ofcourse it's no problem just silence compiler warnings.


segiddins:
I believe NSNotFound is an NSUInteger?

yuuki1224:
@segiddins 
NSNotFound is a NSInteger

alazier:
:+1:

PR #1510, opened 2015-02-17
Add RealmSwift.podspec

segiddins:
Depends upon https://github.com/CocoaPods/CocoaPods/pull/3145.

\c @alazier @tgoyne


Realm.podspec
```diff
@@ -52,7 +52,8 @@ Pod::Spec.new do |s|
   s.osx.vendored_library    = 'core/libtightdb.a'
 
   s.subspec 'Headers' do |s|
-    s.source_files          = 'include/**/*.h'
+    s.source_files          = 'include/**/*.{h,hpp}'
```
tgoyne: The .hpp and private files should not be in the Headers subspec.

Realm/RLMAccessor.mm
```diff
@@ -16,16 +16,16 @@
 //
 ////////////////////////////////////////////////////////////////////////////
 
-#import "RLMAccessor.h"
+#import <Realm/RLMAccessor.h>
```
tgoyne: RLMAccessor.h is not a framework header.

Realm/RLMAccessor.mm
```diff
@@ -16,16 +16,16 @@
...
 
 #import "RLMArray_Private.hpp"
```
tgoyne: Sort the imports after changing them.

Realm.podspec
segiddins: Wait, what's wrong with them being here?

tgoyne: The Headers subspec is purely so that people can use the Realm headers in their unit test target without linking Realm into their unit test target, so there's no reason for it to include anything but the public headers.

Realm/RLMAccessor.mm
segiddins: For some reason, changing it back breaks the build

Realm.podspec
segiddins: But having it include other things (as long as they're not public) causes no harm?

jpsim: It does simplify the podspec a bit, so I say why not.

Realm/RLMAccessor.mm
jpsim: I've seen this before when building Realm as a clang module too. I'd much prefer to find the root cause of the issue, but unless someone can think of a reason this change could cause problems, let's do it.

RealmSwift.podspec
```diff
@@ -0,0 +1,28 @@
...
+                              DESC
+  s.homepage                = "http://realm.io"
+  s.source                  = { :git => 'https://github.com/Realm/realm-cocoa.git', :tag => "v#{s.version}" }
+  s.author                  = { 'Realm' => 'help@realm.io' }
+  s.requires_arc            = true
+  s.social_media_url        = 'https://twitter.com/realm'
+  s.documentation_url       = "http://realm.io/docs/cocoa/#{s.version}"
+  s.license                 = { :type => 'Apache 2.0', :file => 'LICENSE' }
+
+  s.dependency 'Realm', "= #{s.version}"
```
jpsim: will this work, even when using RealmSwift as a local podspec?

jpsim: in other words, will it also grab the local `Realm` dependency, or will it get it from GitHub?

RealmSwift.podspec
```diff
@@ -0,0 +1,28 @@
...
+  s.source_files = 'RealmSwift/*.swift'
+
+  s.ios.deployment_target   = '8.0'
+  s.osx.deployment_target   = '10.8'
```
jpsim: should be 10.9 (source: http://stackoverflow.com/a/24010399/373262)

RealmSwift.podspec
```diff
@@ -0,0 +1,28 @@
+Pod::Spec.new do |s|
```
jpsim: Can you really not define two specs in the same file? We could avoid duplicating a lot of this content.

Realm.podspec
tgoyne: It makes the podspec more complicated, not less (at least assuming that the explicit `private_header_files` isn't pointless) and doesn't have any benefits.

Realm/RLMAccessor.mm
tgoyne: Having to include a seemingly random subset of the private headers as module headers (do you know what's different between RLMRealm_Private.hpp and RLMProperty_Private.hpp? I don't) suggests that the module is *severely* misconfigured and will almost certainly result in a steady stream of bug reports about compilation errors with unhelpful error messages.

Realm.podspec
```diff
@@ -21,25 +22,24 @@ Pod::Spec.new do |s|
...
+                              'include/Realm/RLMObjectBase.h',
+                              'include/Realm/RLMObjectSchema.h',
+                              'include/Realm/RLMPlatform.h',
+                              'include/Realm/RLMProperty.h',
+                              'include/Realm/RLMRealm.h',
+                              'include/Realm/RLMResults.h',
+                              'include/Realm/RLMSchema.h',
+                              'include/Realm/Realm.h'
 
   s.compiler_flags          = "-DTIGHTDB_HAVE_CONFIG -DREALM_SWIFT=0 -DREALM_VERSION='@\"#{s.version}\"'"
```
jpsim: `RealmSwift.podspec` should be set its own compiler flags. Just `-DREALM_SWIFT=1` I think.

Realm.podspec
jpsim: > It makes the podspec more complicated, not less

Ok true, this approach requires to explicitly listing the private header files. I agree with @tg the way it was before is better.

jpsim:
@segiddins ping us again once you've addressed the concerns raised here.

RealmSwift.podspec
mrackwitz: No, podspecs are constrainted in their names to match: `/#{spec.name}.podspec(.json)?/`

RealmSwift.podspec
mrackwitz: As this question may come up again: This depends on your Podfile.

This will install RealmSwift from the local sources, but will fetch Realm from the master spec repo:
```ruby
pod "RealmSwift", path: "vendor/Realm"
```

This will install both from the local sources.
```ruby
pod "RealmSwift", path: "vendor/Realm"
pod "Realm", path: "vendor/Realm"
```

PR #1513, opened 2015-02-17
Fix for deleting objects from the wrong realm

alazier:
@segiddins @tgoyne 

Cherry-picked/merge from swift-realm branch.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@ x.x.x Release notes (yyyy-MM-dd)
 
 * An exception will now be thrown when calling `-beginWriteTransaction` from within a notification
   triggered by calling `-beginWriteTransaction` elsewhere.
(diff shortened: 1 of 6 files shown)


segiddins:
:+1:, but mostly because I wrote that :)

tgoyne:
`RLMArray` and `RLMResults` appear to also have the same behavior of when passing them to a different realm's `deleteObjects:`.

segiddins:
> RLMArray and RLMResults appear to also have the same behavior of when passing them to a different realm's deleteObjects:.

Addressed

alazier:
doesn't look like we are testing either the RLMArray or RLMResults cases.

segiddins:
Tests added

alazier:
:+1:

tgoyne:
:+1:

PR #1515, opened 2015-02-18
[RLMProperty] Disallow creating a property of type RLMObject

### LINKED ISSUE #1465: Creating an object property of type RLMObject should throw
A users was trying to do this in swift which should throw. This should also throw if done from objc.

segiddins:
Closes #1465.

\c @alazier

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ x.x.x Release notes (yyyy-MM-dd)
 * An exception will be thrown when calling `setEncryptionKey:forRealmsAtPath:`, 
   `setSchemaVersion:forRealmAtPath:withMigrationBlock:`, and `migrateRealmAtPath:` 
   when a Realm at the given path is already open.
+* Object and array properties of type `RLMObject` will no longer be allowed.
 
 ### Enhancements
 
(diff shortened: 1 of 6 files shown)


tgoyne:
:+1:

tgoyne:
changelog

alazier:
I think you should instead be changing RLMISSubClass rather than rewriting each instance. Should maybe be renamed to RLMIsObjectClass or something, and we should fix other cases where we have made this change. We should not be duplicating this logic/test throughout the code base.

segiddins:
@alazier done

alazier:
What about the other places `RLMIsSubClass` is called? Should those cases also instead be `RLMIsObjectClass`? If so we can remove `RLMIsSubClass` altogether in favor of the new method.

segiddins:
I replaced all the relevant places, but the new function uses `RLMIsSubclass` under the hood.

segiddins:
(The test failure is `testBeginWriteTransactionsRefreshesRealm`, which has been failing intermittently for a while)

segiddins:
@alazier think I'm good to merge this?

PR #1519, opened 2015-02-18
Avoid an unnecessary write transaction when opening Realms

tgoyne:
Check if any writing actually needs to be done before starting the write transaction, and skip it when possible. Mostly only relevant for the multi-process case, where this makes it so that a write transaction in one process doesn't block other processes from opening the file.

@alazier

### DIFF
--- a/Realm/RLMObjectStore.mm
+++ b/Realm/RLMObjectStore.mm
@@ -162,6 +162,27 @@ void RLMRealmSetSchema(RLMRealm *realm, RLMSchema *targetSchema, bool verify) {
     }
 }
 
+// try to get all of the tables for the realm
+// if they all exist, sets the realm's schema to targetSchema and returns true
+// returns false if any are missing and does not change the schema
+static bool RLMRealmGetTables(RLMRealm *realm, RLMSchema *targetSchema) {
+    if (!RLMRealmHasMetadataTables(realm)) {
+        return false;
+    }
+
+    for (RLMObjectSchema *objectSchema in targetSchema.objectSchema) {
+        NSString *tableName = RLMTableNameForClass(objectSchema.className);
+        TableRef table = realm.group->get_table(tableName.UTF8String);
+        if (!table) {
+            return false;
+        }
+        objectSchema.table = table.get();
+    }
+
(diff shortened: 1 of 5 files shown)


Realm/RLMObjectStore.mm
```diff
@@ -234,22 +255,38 @@ static bool RLMRealmCreateTables(RLMRealm *realm, RLMSchema *targetSchema, bool
...
+
+    return oldVersion != newVersion;
+}
+
 NSError *RLMUpdateRealmToSchemaVersion(RLMRealm *realm, NSUInteger newVersion, RLMSchema *targetSchema, NSError *(^migrationBlock)()) {
-    [realm beginWriteTransaction];
+    // if the schema version matches, try to get all the tables without entering
+    // a write transaction
+    if (!RLMMigrationRequired(realm, newVersion) && RLMRealmGetTables(realm, targetSchema)) {
+            return nil;
```
alazier: indentation is off

Realm/RLMObjectStore.mm
```diff
@@ -162,6 +162,27 @@ void RLMRealmSetSchema(RLMRealm *realm, RLMSchema *targetSchema, bool verify) {
...
+    for (RLMObjectSchema *objectSchema in targetSchema.objectSchema) {
+        NSString *tableName = RLMTableNameForClass(objectSchema.className);
+        TableRef table = realm.group->get_table(tableName.UTF8String);
+        if (!table) {
+            return false;
+        }
+        objectSchema.table = table.get();
+    }
+
+    RLMRealmSetSchema(realm, targetSchema, true);
```
alazier: Think it might be better if RLMSetSchema is called from the calling function instead of here. Same probably goes with RLMCreateTables. The concern is that it isn't clear that RLMRealmSetSchema is being called based on these function names. Could also rename both methods to RLMRealmTryGetTablesAndSetSchema and RLMCreateTablesAndSetSchema but those sound horrible

alazier:
Functionally this looks fine. Would be good if we could add an interprocess test to make sure actually fixes the issue at hand.

tgoyne:
I verified it worked manually, and I'll add a test on the interprocess branch.

PR #1523, opened 2015-02-19
[RLMObjectBase] Prefix realm and objectSchema properties

segiddins:
\c @alazier @tgoyne

### DIFF
--- a/RealmSwift/Object.swift
+++ b/RealmSwift/Object.swift
@@ -82,23 +82,20 @@ public class Object : RLMObjectBase, Equatable {
 
     // MARK: Properties
 
-    // FIXME: Rename to `realm`
     /// The `Realm` this object belongs to, or `nil` if the object
     /// does not belong to a realm (the object is standalone).
-    public var swiftRealm: Realm? {
-        if let rlmRealm = realm {
+    public var realm: Realm? {
+        if let rlmRealm = rlmRealm {
             return Realm(rlmRealm)
         }
         return nil
     }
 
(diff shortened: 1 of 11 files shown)


RealmSwift/Object.swift
```diff
@@ -82,23 +82,19 @@ public class Object : RLMObjectBase, Equatable {
...
-    // FIXME: Rename to `realm`
     /// The `Realm` this object belongs to, or `nil` if the object
     /// does not belong to a realm (the object is standalone).
-    public var swiftRealm: Realm? {
-        if let rlmRealm = realm {
-            return Realm(rlmRealm)
-        }
-        return nil
+    public var realm: Realm? {
+        return Realm(RLMObjectBaseRealm(self))
```
segiddins: still needs the `if let` check

RealmSwift/Object.swift
```diff
@@ -82,23 +82,19 @@ public class Object : RLMObjectBase, Equatable {
...
     }
 
-    // FIXME: Rename to `objectSchema`
     /// The `ObjectSchema` which lists the persisted properties for this object.
-    public var swiftObjectSchema: ObjectSchema { return ObjectSchema(rlmObjectSchema: objectSchema) }
+    public var objectSchema: ObjectSchema {
+        return ObjectSchema(rlmObjectSchema: RLMObjectBaseObjectSchema(self))
```
segiddins: needs the `if let` check

PR #1524, opened 2015-02-19
[RLMObjectBase] Prefix realm and objectSchema properties

segiddins:
\c @alazier @tgoyne


tgoyne:
This is kinda ugly. Would it work to have unprefixed properties on RLMObjectBase that are declared in a header that isn't exposed to Swift?

segiddins:
But then accessing those properties on a swift object wouldn't return the RLMRealm?

-Samuel E. Giddins

On Feb 19, 2015, at 1:44 PM, Thomas Goyne <notifications@github.com> wrote:

This is kinda ugly. Would it work to have unprefixed properties on RLMObjectBase that are declared in a header that isn't exposed to Swift?

alazier:
This will also probably have to be done differently to support a generalized object store. Lets discuss alternatives tomorrow.

PR #1531, opened 2015-02-19
Use a named pipe for interprocess notifications

tgoyne:
Basic idea here is to just create a named pipe, wait for data to become available on it in a background thread (and trigger notifications when it does), and then write a byte to it after each commit. kqueue is used to wait on both the named pipe and an anonymous pipe used locally to tell the waiting thread to exit.

Doing the notifications entirely in the binding turned out to not really be any more complex than using `wait_for_change()` was since eliminating the `SharedGroup` on the background thread makes shutdown and cleanup a lot simpler.

@alazier

### DIFF
--- a/Realm/RLMRealmUtil.h
+++ b/Realm/RLMRealmUtil.h
@@ -29,8 +29,12 @@ RLMRealm *RLMGetAnyCachedRealmForPath(NSString *path);
 // Clear the weak cache of Realms
 void RLMClearRealmCache();
 
-void RLMStartListeningForChanges(RLMRealm *realm);
-void RLMStopListeningForChanges(RLMRealm *realm);
-// notify all of the Realms other than `notifyingRealm` at the same path as
-// `notifyingRealm` that a commit has occurred
-void RLMNotifyOtherRealms(RLMRealm *notifyingRealm);
+@interface RLMWeakNotifier : NSObject
+// listens to changes to the realm's file and notifies it when they occur
+// does not retain the Realm
+- (instancetype)initWithRealm:(RLMRealm *)realm;
+// stop listening for changes
+- (void)stop;
+// notify other Realm instances for the same path that a change has occurred
+- (void)notifyOtherRealms;
+@end
--- a/Realm/RLMRealmUtil.h
+++ b/Realm/RLMRealmUtil.h
@@ -32,7 +32,7 @@ void RLMClearRealmCache();
 @interface RLMNotifier : NSObject
 // listens to changes to the realm's file and notifies it when they occur
 // does not retain the Realm
-- (instancetype)initWithRealm:(RLMRealm *)realm;
+- (instancetype)initWithRealm:(RLMRealm *)realm error:(NSError **)error;
 // stop listening for changes
 - (void)stop;
 // notify other Realm instances for the same path that a change has occurred
--- a/Realm/RLMRealmUtil.h
+++ b/Realm/RLMRealmUtil.h
@@ -29,7 +29,7 @@ RLMRealm *RLMGetAnyCachedRealmForPath(NSString *path);
 // Clear the weak cache of Realms
 void RLMClearRealmCache();
 
-@interface RLMWeakNotifier : NSObject
+@interface RLMNotifier : NSObject
 // listens to changes to the realm's file and notifies it when they occur
 // does not retain the Realm
 - (instancetype)initWithRealm:(RLMRealm *)realm;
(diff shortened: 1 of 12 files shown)


alazier:
Sounds encouraging. Have we run the latency test on this?

tgoyne:
It's ~20% faster than master due to that it doesn't use `performSelector:onThread:`.

Realm/RLMRealmUtil.h
```diff
@@ -29,8 +29,12 @@ RLMRealm *RLMGetAnyCachedRealmForPath(NSString *path);
 // Clear the weak cache of Realms
 void RLMClearRealmCache();
 
-void RLMStartListeningForChanges(RLMRealm *realm);
-void RLMStopListeningForChanges(RLMRealm *realm);
-// notify all of the Realms other than `notifyingRealm` at the same path as
-// `notifyingRealm` that a commit has occurred
-void RLMNotifyOtherRealms(RLMRealm *notifyingRealm);
+@interface RLMWeakNotifier : NSObject
```
alazier: I think we should just rename this to `RLMNotifier`.

Realm/RLMRealmUtil.h
```diff
@@ -29,8 +29,12 @@ RLMRealm *RLMGetAnyCachedRealmForPath(NSString *path);
...
+// listens to changes to the realm's file and notifies it when they occur
+// does not retain the Realm
+- (instancetype)initWithRealm:(RLMRealm *)realm;
+// stop listening for changes
+- (void)stop;
+// notify other Realm instances for the same path that a change has occurred
+- (void)notifyOtherRealms;
+@end
```
alazier: newline

Realm/RLMRealmUtil.mm
```diff
@@ -49,85 +53,116 @@ void RLMCacheRealm(RLMRealm *realm) {
...
-void RLMStartListeningForChanges(RLMRealm *realm) {
-    @synchronized (s_notifiersPerPath) {
-        NSMapTable *notifiers = s_notifiersPerPath[realm.path];
-        if (!notifiers) {
-            notifiers = [NSMapTable mapTableWithKeyOptions:NSPointerFunctionsWeakMemory valueOptions:NSPointerFunctionsStrongMemory];
-            s_notifiersPerPath[realm.path] = notifiers;
-        }
-        [notifiers setObject:[[RLMWeakNotifier alloc] initWithRealm:realm] forKey:realm];
+static void checkError(int ret) {
+    if (ret < 0 && errno != EEXIST) {
```
alazier: comments here could be helpful

Realm/RLMRealmUtil.mm
```diff
@@ -49,85 +53,116 @@ void RLMCacheRealm(RLMRealm *realm) {
...
 
-void RLMStartListeningForChanges(RLMRealm *realm) {
-    @synchronized (s_notifiersPerPath) {
-        NSMapTable *notifiers = s_notifiersPerPath[realm.path];
-        if (!notifiers) {
-            notifiers = [NSMapTable mapTableWithKeyOptions:NSPointerFunctionsWeakMemory valueOptions:NSPointerFunctionsStrongMemory];
-            s_notifiersPerPath[realm.path] = notifiers;
-        }
-        [notifiers setObject:[[RLMWeakNotifier alloc] initWithRealm:realm] forKey:realm];
+static void checkError(int ret) {
```
alazier: Could move this and `notifyFd` outside `RLMNotifier` implementation

alazier:
The use of blocks/closures makes it a bit difficult to see at a glance all of the state variables being used. Might be nice to encapsulate all state used in the objc class, or wrap it all in a struct which is created in one place and used in a block. Right now we have a mix of the two, some members of `RLMNotifier` and some stack variables, which is a bit confusing.

alazier:
Would also be good to have a big juicy comment which describes exactly what is going on here, how the notification system works, and all of the states we can be in.

tgoyne:
It definitely needs some shuffling around and a nice overview of what it's doing, but I think I've been staring at it too long to have any hope of writing it now.  Maybe tomorrow.

alazier:
There are a few things I don't quite understand so lets discuss tomorrow.

tgoyne:
The remaining test failure is a perf regression for realm creation. Not a lot of room left for optimization since 80% of the time is spent in the three `open()` calls. One thing we could do is push the `open()` on the fifo to the background thread, but that'd limit how many error cases we could report gracefully.

PR #1536, opened 2015-02-21
[RLMCollection] Implement custom KVC support

### LINKED ISSUE #1172: Support setValue:forKey: and valueForKey: on RLMCollection
Hi guys,

will it be possible to do bulk updates in the future? I would really like a possibility to update all objects matching a given query without looping them or even all updates in the Realm at once.

segiddins:
Closes #1172.

\c @alazier @tgoyne @jpsim


Realm/RLMUtil.mm
```diff
@@ -240,6 +241,44 @@ id RLMValidatedObjectForProperty(id obj, RLMProperty *prop, RLMSchema *schema) {
     return outArray;
 };
 
+NSArray *RLMCollectionValueForKey(NSString *key, RLMRealm *realm, RLMObjectSchema *objectSchema, size_t count, size_t (^indexGenerator)()) {
```
alazier: the indexGenerator block should just take the index. index can be passed in from the loops in the following two methods rather than being incremented by each caller.

Realm/RLMUtil.mm
```diff
@@ -240,6 +241,44 @@ id RLMValidatedObjectForProperty(id obj, RLMProperty *prop, RLMSchema *schema) {
...
+    if (count == 0) {
+        return @[];
+    }
+    BOOL keyIsSelf = [key isEqualToString:@"self"];
+    NSMutableArray *results = [NSMutableArray arrayWithCapacity:count];
+    RLMObjectBase *accessor = RLMCreateObjectAccessor(realm, objectSchema, 0);
```
alazier: We need to call RLMCreateObjectAccessor for every element for swift list properties to work. Once we do this we no longer have to get the first row twice.

Realm/RLMUtil.mm
```diff
@@ -240,6 +241,44 @@ id RLMValidatedObjectForProperty(id obj, RLMProperty *prop, RLMSchema *schema) {
...
+    tightdb::Table *table = objectSchema.table;
+    for (size_t i = 0; i < count; i++) {
+        size_t rowIndex = indexGenerator();
+        id value;
```
alazier: instead of pulling out value, just add to results in each block of the if/else.

Realm/RLMResults.mm
```diff
@@ -481,6 +499,24 @@ - (NSUInteger)count {
     return _table->size();
 }
 
+- (id)valueForKey:(NSString *)key {
+    RLMResultsValidate(self);
+    const size_t size = _table->size();
+    __block size_t index = 0;
+    return RLMCollectionValueForKey(key, _realm, _objectSchema, size, ^size_t{
+        return _table->get(index++).get_index();
```
tgoyne: This is a very roundabout way of writing `return index++`.

Realm/RLMResults.mm
```diff
@@ -244,6 +244,24 @@ - (NSUInteger)indexOfObject:(RLMObject *)object {
...
+        return _backingView.get(index++).get_index();
+    });
+}
+
+- (void)setValue:(id)value forKey:(NSString *)key {
+    RLMResultsValidateInWriteTransaction(self);
+    const size_t size = _backingView.size();
+    __block size_t index = 0;
+    RLMCollectionSetValueForKey(value, key, _realm, _objectSchema, size, ^size_t{
+        return _backingView.get(index++).get_index();
```
tgoyne: `_backingView.get_source_ndx(index++)`

Also maybe just add `indexInSource` to RLMArrayLinkView, expose it in the _Private headers, and use that rather than the block.

tgoyne:
Does `valueForKeyPath:` magically work with only `valueForKey:` overridden? If so it'd be good to have tests for that (if not, implementing that doesn't need to be part of this PR).

segiddins:
@tgoyne it should, as `-[NSObject valueForKeyPath:]` just recursively decomposes the key path and calls `valueForKey` internally, but I'll add tests

Realm/RLMResults.mm
segiddins: What is `indexInSource`?

Realm/RLMUtil.mm
segiddins: I personally prefer not to duplication the `addObject` call

bmunkholm:
remember changelog update

jpsim:
Please add docs to `RLMCollection.h`:

```objc
/**
 Returns an NSArray containing the results of invoking `valueForKey:` using key on each of the collection's objects.

 @param key The name of the property.

 @return NSArray containing the results of invoking `valueForKey:` using key on each of the collection's objects.
 */
- (id)valueForKey:(NSString *)key;

/**
 Invokes `setValue:forKey:` on each of the collection's objects using the specified value and key.

 @warning This method can only be called during a write transaction.

 @param value The object value.
 @param key   The name of the property.
 */
- (void)setValue:(id)value forKey:(NSString *)key;
```

Realm/RLMResults.mm
```diff
@@ -560,6 +592,15 @@ - (id)objectAtIndex:(NSUInteger)index {
     @throw RLMException(@"Index is out of bounds.", @{@"index": @(index)});
 }
 
+- (id)valueForKey:(NSString *)key {
+    RLMResultsValidate(self);
```
jpsim: No need to validate.

Realm/RLMResults.mm
```diff
@@ -560,6 +592,15 @@ - (id)objectAtIndex:(NSUInteger)index {
...
+    return @[];
+}
+
+- (void)setValue:(__unused id)value forKey:(__unused NSString *)key {
+    RLMResultsValidateInWriteTransaction(self);
```
jpsim: No need to validate.

segiddins: Why not? It would potentially allow the user to think they were doing things on the right thread when they really weren't.

tgoyne: It's not really necessary, but I do think it'd be better to have a consistent interface and do the checks here.

jpsim: Good points, can you add validations elsewhere in this class? That's small enough to do in this PR.

Realm/Tests/ArrayPropertyTests.m
```diff
@@ -262,6 +262,60 @@ - (void)testFastEnumeration
                     @"Adding an object during fast enumeration did not throw");
 }
 
+- (void)testValueForKey {
```
jpsim: Please add standalone tests with employees in both `testValueForKey` and `testSetValueForKey`.

jpsim:
Other than my comments, :+1:

segiddins:
This PR is on hold for a bit, until I have time to finish everything up and make sure we have adequate test coverage.

segiddins:
This is probably ready for review again.

jpsim:
That's a :+1: for me.

segiddins:
Changelog entry added.

Realm/RLMArrayLinkView.mm
```diff
@@ -226,6 +226,22 @@ - (NSUInteger)indexOfObject:(RLMObject *)object {
     return result;
 }
 
+- (id)valueForKey:(NSString *)key {
+    RLMLinkViewArrayValidateAttached(self);
+    const size_t size = _backingLinkView->size();
+    return RLMCollectionValueForKey(key, _realm, _objectSchema, size, ^size_t(size_t index){
```
tgoyne: missing a space before { in all of these

tgoyne:
Needs tests for `@"self"`.

Realm/RLMArrayLinkView.mm
segiddins: we add the preceding space even for blocks?

segiddins: done

segiddins:
> Needs tests for @"self".
done

RealmSwift/Tests/ListTests.swift
```diff
@@ -161,6 +161,19 @@ class ListTests: TestCase {
...
 
+    func testValueForKey() {
+        let expected = map(array) { $0.stringCol }
+        let actual = array.valueForKey("stringCol") as [String]!
+        XCTAssertEqual(expected, actual)
+    }
+
+    func testSetValueForKey() {
+        array.setValue("hi there!", forKey: "stringCol")
+        let expected = map(array as List<SwiftStringObject>) { _ in "hi there!" }
```
tgoyne: `let expected = map(array!) { _ in "hi there!" }`

segiddins: done

Realm/RLMUtil.mm
```diff
@@ -248,6 +249,50 @@ id RLMValidatedObjectForProperty(id obj, RLMProperty *prop, RLMSchema *schema) {
...
 
+NSArray *RLMCollectionValueForKey(NSString *key, RLMRealm *realm, RLMObjectSchema *objectSchema, size_t count, size_t (^indexGenerator)(size_t)) {
+    if (count == 0) {
+        return @[];
+    }
+    BOOL keyIsSelf = [key isEqualToString:@"self"];
+    NSMutableArray *results = [NSMutableArray arrayWithCapacity:count];
+    RLMObjectBase *accessor;
+    tightdb::Table *table = nullptr;
+    if (!keyIsSelf) {
```
alazier: I don't see why we don't just separate out the case `keyIsSelf == YES` completely. Would make the code a bit easier to read without much code duplication.

segiddins: as it is, I don't find it to be very hard to read, but if you _really_ want it separated out, I suppose I could, but I would mean duplicating a lot of the core logic

alazier: I don't think it requires any duplicate logic other than a loop, and seem cleaner to me:
```objc
NSMutableArray *results = [NSMutableArray arrayWithCapacity:count];
 if ([key isEqualToString:@"self"]) {      
    for (size_t i = 0; i < count; i++) {
        size_t rowIndex = indexGenerator(i);
        [results addObject:RLMCreateObjectAccessor(realm, objectSchema, rowIndex) ?: NSNull.null];
    }
    return results;
}

RLMObjectBase *accessor = [[objectSchema.accessorClass alloc] initWithRealm:realm schema:objectSchema];    
tightdb::Table *table = objectSchema.table;
for (size_t i = 0; i < count; i++) {
    size_t rowIndex = indexGenerator(i);
    accessor->_row = (*table)[rowIndex];
    RLMInitializeSwiftListAccessor(accessor);
    [results addObject:[accessor valueForKey:key] ?: NSNull.null];
}
```

segiddins: done

alazier:
Might be good to test `self` for swift as well. Otherwise :+1:

alazier:
Meant to comment, not close...

segiddins:
I believe all comments have been addressed? Time to :shipit: ?

tgoyne:
:+1:

PR #1537, opened 2015-02-21
Updated example projects to work with Swift 1.2

isaacroldan:
The examples stopped working for the new Swift version, a couple of small changes:
- cast with `as!` instead of `as` to force downcast
- changed the signature of the `application:didFinishLaunchingWithOptions` method
- also, changed some optionals to follow the same style in all examples.


segiddins:
Thank you @isaacroldan, but as Swift 1.2 is still unreleased, I think we're best off sticking with the production version.

PR #1539, opened 2015-02-23
Swiftify object customization methods

alazier:
@tgoyne @segiddins

### DIFF
--- a/Realm/RLMObjectBase.h
+++ b/Realm/RLMObjectBase.h
@@ -38,8 +38,18 @@
 
 + (NSString *)className;
 
-+ (NSDictionary *)defaultPropertyValues;
-+ (NSString *)primaryKey;
-+ (NSArray *)ignoredProperties;
+@end
+
+
+
+Class RLMObjectUtilClass(bool isSwift);
+
+@interface RLMObjectUtil : NSObject
+
++ (NSString *)primaryKeyForClass:(Class)cls;
++ (NSArray *)ignoredPropertiesForClass:(Class)cls;
++ (NSArray *)indexedPropertiesForClass:(Class)cls;
+
++ (NSArray *)getGenericListPropertyNames:(id)obj;
 
 @end
--- a/Realm/RLMObjectBase.h
+++ b/Realm/RLMObjectBase.h
@@ -42,7 +42,7 @@
 
 
 
-Class RLMObjectUtilClass(bool isSwift);
+Class RLMObjectUtilClass(BOOL isSwift);
 
 @interface RLMObjectUtil : NSObject
 
--- a/Realm/RLMObjectBase.mm
+++ b/Realm/RLMObjectBase.mm
@@ -88,35 +88,15 @@
     if (self) {
         _realm = realm;
         _objectSchema = schema;
-        if (useDefaults) {
+        if (useDefaults && !schema.isSwiftClass) {
             // set default values
-            NSDictionary *dict = [self.class defaultPropertyValues];
+            NSDictionary *dict = RLMDefaultValuesForObjectSchema(schema);
             for (NSString *key in dict) {
                 [self setValue:dict[key] forKey:key];
             }
         }
     }
     return self;
-}
-
-// default indexed properties implementation
-+ (NSArray *)indexedProperties {
-    return @[];
-}
-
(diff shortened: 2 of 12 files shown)


Realm/RLMObjectBase.h
```diff
@@ -38,8 +38,18 @@
...
 
-+ (NSDictionary *)defaultPropertyValues;
-+ (NSString *)primaryKey;
-+ (NSArray *)ignoredProperties;
-
 @end
+
+
+
+Class RLMObjectUtilClass(bool isSwift);
```
segiddins: should be `BOOL` since this is ObjC

Realm/RLMObjectBase.mm
```diff
@@ -289,3 +269,31 @@ - (NSUInteger)hash {
...
+
++ (NSArray *)indexedPropertiesForClass:(Class)cls {
+    return [cls indexedProperties];
+}
+
++ (NSArray *)getGenericListPropertyNames:(__unused id)obj {
+    return nil;
+}
+
+@end
```
segiddins: newline

Realm/Tests/Swift/SwiftObjectInterfaceTests.swift
```diff
@@ -109,7 +109,7 @@ class SwiftObjectInterfaceTests: SwiftTestCase {
         realm.commitWriteTransaction()
 
         let object = SwiftDefaultObject.allObjectsInRealm(realm).firstObject() as SwiftDefaultObject
-        XCTAssertEqual(object.intCol, 2, "defaultPropertyValues should override native property default value")
+        XCTAssertEqual(object.intCol, 1, "defaultPropertyValues should override native property default value")
```
segiddins: update failure message

RealmSwift/Object.swift
```diff
@@ -119,7 +119,7 @@ public class Object : RLMObjectBase, Equatable {
 
     :returns: `Array` of property names to ignore.
     */
-    public class func swiftIgnoredProperties() -> [String] { return [] }
+    public class func ignoredProperties() -> [String] { return [] }
```
segiddins: remove the `FIXME`

RealmSwift/Object.swift
```diff
@@ -178,6 +178,25 @@ public func == <T: Object>(lhs: T, rhs: T) -> Bool {
...
+        if let type = type as? Object.Type {
+            return type.ignoredProperties() as NSArray?
+        }
+        return nil
+    }
+    @objc private class func indexedPropertiesForClass(type: AnyClass) -> NSArray? {
+        if let type = type as? Object.Type {
+            return type.indexedProperties() as NSArray?
+        }
+        return nil
```
segiddins: should probably assert that we never get a non-`Object` class?

Realm/RLMObjectBase.mm
```diff
@@ -289,3 +269,31 @@ - (NSUInteger)hash {
 }
 
 @end
+
+
+
+Class RLMObjectUtilClass(bool isSwift) {
+    static Class objectUtilObjc = [RLMObjectUtil class];
+    static Class objectUtilSwift = NSClassFromString(@"RealmSwift.ObjectUtil");
+    return isSwift && objectUtilSwift ? objectUtilSwift : objectUtilObjc;
```
segiddins: Why the `&& objectUtilSwift`?

segiddins:
Maybe some of these methods belong more naturally on `RLMObjectSchema`?

Realm/RLMObjectBase.mm
alazier: objectUtilSwift doesn't exist for the old swift apis - this is how we fall back to the objc version.

segiddins: In that case, won't things explode if you have a Swift subclass of `RLMObject` and you also have `RealmSwift`?

alazier: We don't support mixing the apis. ie the old apis are not supported with the RealmSwift framework.

segiddins: Then we should find a clean way to error if the APIs are mixed

segiddins:
This change will make it impossible to provide default values for array properties

PR #1546, opened 2015-02-25
clean up kvc support for lists

alazier:
@tgoyne @segiddins 

I think this addresses the concerns. valueForKey now returns a List. setValue supports both lists and RLMArrays. We only had to forward the fast enumeration method to make this work.

### DIFF
--- a/Realm/RLMListBase.m
+++ b/Realm/RLMListBase.m
@@ -20,6 +20,7 @@
 #import <Realm/RLMArray.h>
 
 @implementation RLMListBase
+
 - (instancetype)initWithArray:(RLMArray *)array {
     self = [super init];
     if (self) {
@@ -31,4 +32,5 @@
 - (NSUInteger)countByEnumeratingWithState:(NSFastEnumerationState *)state objects:(id __unsafe_unretained [])buffer count:(NSUInteger)len {
     return [__rlmArray countByEnumeratingWithState:state objects:buffer count:len];
 }
+
 @end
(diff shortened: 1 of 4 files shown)


Realm/RLMListBase.m
```diff
@@ -26,4 +27,8 @@ - (instancetype)initWithArray:(RLMArray *)array {
     }
     return self;
 }
+
+- (NSUInteger)countByEnumeratingWithState:(NSFastEnumerationState *)state objects:(id __unsafe_unretained [])buffer count:(NSUInteger)len {
+    return [__rlmArray countByEnumeratingWithState:state objects:buffer count:len];
+}
 @end
```
segiddins: newline

segiddins:
Tests for this? You may just want to use what I'd written that uncovered the issue, namely

```swift
func testInitStandaloneObjectWithObjectWithDefaultArrayProperty() {
        let object = SwiftObject(object: [
            "intCol": 200
            ])

        XCTAssertNil(object.realm)
        XCTAssertEqual(object.stringCol, "a", "Should be initialized with default value")
        XCTAssertEqual(object.intCol, 200, "Should be initialized with the passed-in value")
        XCTAssertEqual(object.arrayCol.count, 0)
        XCTAssertNil(object.arrayCol.realm)
    }

    func testInitStandaloneObjectWithObjectWithCustomArrayProperty() {
        let object = SwiftObject(object: [
            "arrayCol" : [[true]]
            ])

        XCTAssertNil(object.realm)
        XCTAssertEqual(object.arrayCol.count, 1)
        XCTAssertEqual(object.arrayCol.first!.boolCol, true)
        XCTAssertNil(object.arrayCol.realm)
    }
```

alazier:
I assumed you were working on these in another pr. These tests do pass now though :)

segiddins:
Please add them, then?

PR #1548, opened 2015-02-25
[RLMObjectBase] Don't crash when calling isEqual: against a non-RLMObject...

### LINKED ISSUE #1547: isEqual: between Realm object and non-Realm object causes a crash
Looks like `[realmObject isEqual:nonRealmObject]` causes a crash — unless `nonRealmObject` implements, for example,
`- (id)realm { return nil; }`

segiddins:
Closes #1547.

\c @alazier @tgoyne

### DIFF
--- a/Realm/RLMObjectBase.mm
+++ b/Realm/RLMObjectBase.mm
@@ -256,12 +256,12 @@
 }
 
 - (BOOL)isEqual:(id)object {
-    if (_objectSchema.primaryKeyProperty) {
-        return [self isEqualToObject:object];
+    if (RLMObjectBase *other = RLMDynamicCast<RLMObjectBase>(object)) {
+	if (_objectSchema.primaryKeyProperty) {
+	    return [self isEqualToObject:other];
+	}
     }
-    else {
-        return [super isEqual:object];
-    }
+    return [super isEqual:object];
 }
 
 - (NSUInteger)hash {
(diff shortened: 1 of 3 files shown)


Realm/RLMObjectBase.mm
```diff
@@ -256,12 +256,12 @@ - (BOOL)isEqualToObject:(RLMObject *)object {
...
-        return [self isEqualToObject:object];
-    }
-    else {
-        return [super isEqual:object];
+    if (RLMObjectBase *other = RLMDynamicCast<RLMObjectBase>(object)) {
+        if (_objectSchema.primaryKeyProperty) {
+            return [self isEqualToObject:other];
+        }
     }
+    return [super isEqual:object];
```
alazier: Wont this always be NO?

alazier:
should definitely test this.

Realm/RLMObjectBase.mm
segiddins: no, it'll be `YES` if and only if `self == object` (because that's the `NSObject` implementation)

tgoyne: i.e. it'll always be NO because we would have already returned YES if that was the case.

segiddins: not necessarily, since we only check `isEqualToObject:` when there's a primary key property

tgoyne: Oh, I misread it and missed that the `self == object` check a few lines up wasn't part of this method...

jpsim:
I've added tests and a changelog entry for this. Anything else? /cc @segiddins @tgoyne

segiddins:
:+1:

PR #1551, opened 2015-02-26
Parallelize the pull request test running

tgoyne:
This does a few things:

1. Refactors `build.sh` to make the `-debug` versions of commands just set a variable and call the normal version, to eliminate a bunch of duplication and to make it easier to programmatically call either the debug or release versions.
2. Splits `verify` into sub-tasks which each build `Realm.framework` in one configuration and then run all tests relevant to that framework.
3. Makes `build.sh osx` copy the built result to `build/osx` and switches the browser and examples to looking there, to make it easier to use the browser with a debug build of the framework.

https://ci.tightdb.com/job/cocoa_pr/ takes advantage of this to run each sub-task on a different slave in parallel. Net result is a reduction from 12-15 minutes to 5-6 minutes to run all of the tests and a slightly better UI for determining what exactly failed.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -631,6 +631,7 @@ EOF
         ;;
 
     *)
+        echo "Unknown command '$COMMAND'"
         usage
         exit 1
         ;;
--- a/build.sh
+++ b/build.sh
@@ -334,8 +334,8 @@ case "$COMMAND" in
 
     "verify-osx")
         CONFIGURATION="$CONFIGURATION" sh build.sh test-osx
-        CONFIGURATION="$CONFIGURATION" sh build.sh test-browse
-        CONFIGURATION="$CONFIGURATION" sh build.sh examples-os
+        CONFIGURATION="$CONFIGURATION" sh build.sh test-browser
+        CONFIGURATION="$CONFIGURATION" sh build.sh examples-osx
 
         (
             cd examples/osx/objc/build/DerivedData/RealmExamples/Build/Products/$CONFIGURATION
--- a/build.sh
+++ b/build.sh
@@ -257,6 +257,9 @@ case "$COMMAND" in
 
     "osx")
         xcrealm "-scheme OSX -configuration $CONFIGURATION"
+        rm -rf build/osx
+        mkdir build/osx
+        cp -R build/DerivedData/Realm/Build/Products/$CONFIGURATION/Realm.framework build/osx
         exit 0
         ;;
 
--- a/build.sh
+++ b/build.sh
@@ -3,7 +3,7 @@
 ##################################################################################
 # Custom build tool for Realm Objective C binding.
 #
-# (C) Copyright 2011-2014 by realm.io.
+# (C) Copyright 2011-2015 by realm.io.
 ##################################################################################
 
 # Warning: pipefail is not a POSIX compatible option, but on OS X it works just fine.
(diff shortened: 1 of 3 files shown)


build.sh
```diff
@@ -184,6 +184,14 @@ COMMAND="$1"
 XCMODE="$2"
 : ${XCMODE:=xcodebuild} # must be one of: xcodebuild (default), xcpretty, xctool
 
+# Use Debug config if command ends with -debug, otherwise default to Release
+case "$COMMAND" in
+    *-debug)
```
segiddins: should we also set `ONLY_ACTIVE_ARCHITECTURES=NO` here?

alazier:
Thanks so much for doing this. Will take a look in detail later tonight.

alazier:
Any idea if this will fix the problem we have been having with `testBeginWriteTransactionsRefreshesRealm`?

alazier:
Looks like there was a different test failure when running this on ci: `testCannotHaveReadOnlyAndReadWriteRealmsAtSamePathAtSameTime`

alazier:
The build.sh changes on the swift branch will conflict with this pretty badly. We also change the way we deal with configuration completely on that branch (it is the first argument instead of XCMODE). Might make sense to apply these changes on that branch instead of to master - that is assuming we are finishing the new apis up in the next week.

tgoyne:
The releasability_pipeline failures don't involve anything changed here.

Merging these changes with the ones on al-swift looks pretty trivial. I'll pull over the changes that aren't directly Swift-related.

build.sh
tgoyne: Why?

segiddins: So the output of `sh build.sh ios-debug` will be usable on both architectures?

tgoyne: It already is.

segiddins: :+1:

build.sh
```diff
@@ -32,33 +32,33 @@ Usage: sh $0 command [argument]
...
-  test-browser [xcmode]:   tests the Realm Browser OSX app
-  verify [xcmode]:         cleans, removes docs/output/, then runs docs, test-all, examples & browser
+  clean:                   clean up/remove all generated files
+  build: [settings]        builds iOS and OS X frameworks
+  ios: [settings]          builds iOS frameworks
+  ios-dynamic: [settings]  builds iOS dynamic frameworks
+  ios-static: [settings]   builds a fat iOS static framework
+  osx: [settings]          builds OS X framework
+  test-ios: [settings]     tests iOS framework
+  test-ios-devices:        tests iOS on all attached iOS devices
```
jpsim: Shouldn't this accept `[settings]` as well? Minor point: `[settings]` should come before the colon (e.g. `build [settings]: ...`).

build.sh
```diff
@@ -181,9 +202,17 @@ download_core() {
 }
 
 COMMAND="$1"
-XCMODE="$2"
-: ${XCMODE:=xcodebuild} # must be one of: xcodebuild (default), xcpretty, xctool
+BUILD_SETTINGS="$2"
 
+# Use Debug config if command ends with -debug, otherwise default to Release
```
jpsim: Is support for this worth keeping? Did you do this so there would be minimal changes required to CI?

tgoyne: No, it's just because I find it less annoying to use.

jpsim: Sure, it's grown on me.

build.sh
```diff
@@ -274,78 +290,91 @@ case "$COMMAND" in
...
+        sh build.sh verify-docs
+        sh build.sh verify-osx
+        sh build.sh verify-osx-debug
+        sh build.sh verify-ios
+        sh build.sh verify-ios-debug
+        sh build.sh verify-ios-device
+        ;;
+
+    "verify-osx")
+        CONFIGURATION="$CONFIGURATION" sh build.sh test-osx
```
jpsim: Setting `CONFIGURATION` to its existing value actually does something? If so, we should do it everywhere else we call `sh build.sh`.

jpsim:
This is super great, @tgoyne. :+1:

PR #1556, opened 2015-02-26
Fix race conditions in tests

tgoyne:
Consistently use dispatch_sync(queue, ^{}); to wait for a previously async thing on the queue as the other approaches used didn't guarantee that the block would be released before the test ended, resulting in RLMRealm instances sometimes being deallocated on the wrong thread and throwing an exception.

Might be why tests are sometimes breaking on master.

@alazier

### DIFF
--- a/Realm/Tests/ArrayPropertyTests.m
+++ b/Realm/Tests/ArrayPropertyTests.m
@@ -18,8 +18,6 @@
 
 #import "RLMTestCase.h"
 
-#import <libkern/OSAtomic.h>
-
 @implementation DogArrayObject
 @end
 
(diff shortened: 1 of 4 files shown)


alazier:
:+1:

alazier:
Lets hope this actually fixes the ci issues. Think it has a good chance.

segiddins:
According to the docs on `dispatch_group_async`, `This function performs a Block_copy and Block_release on behalf of the caller.`.

alazier:
I think the `dispatch_group_async` was fine, but isn't needed for only one async method since we do a `dispatch_sync` at the end.

tgoyne:
The dispatch_group_async was the only one I actually saw problems from. I'm not entire sure how, but the deallocation of the group (which happened on one of the libdispatch control queues) was leading to an RLMRealm being deallocated on that thread.

PR #1558, opened 2015-02-27
Set the deployment targets to OS X 10.9 / iOS 7.0 for the Project, while...

danielpovlsen:
... letting the individual Targets inherit. The iOS 8 Framework and iOS 8 Tests have deployment target set to 8.0.

I do this to be able to run the tests on our iOS 7 device here at the CPH office.

@alazier @tgoyne, please review. :-)


segiddins:
This looks fine to me

jpsim:
Thanks @danielpovlsen!

PR #1559, opened 2015-02-27
Fix unsafe uses of __unsafe_unretained

tgoyne:
Fix a place where a newly created object was being assigned to a __uu pointer, and make all __uu pointers const to ensure that the same thing isn't being done elsewhere.

@alazier

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@ x.x.x Release notes (yyyy-MM-dd)
 * An exception will now be thrown when calling `-beginWriteTransaction` from within a notification
   triggered by calling `-beginWriteTransaction` elsewhere.
 * When calling `delete:` we now verify that the object being deleted is persisted in the target Realm.
+* Fix crash when calling `createOrUpdate:inRealm` with nested linked objects.
 
 0.90.6 Release notes (2015-02-20)
 =============================================================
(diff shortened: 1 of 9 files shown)


segiddins:
Is there a compiler warning we can enable to help us prevent this happening in the future?

tgoyne:
Not that I know of, but only ever using const __uu pointers will make it impossible for this exact issue to occur (there are plenty of other ways to run into problems with non-owning pointers).

danielpovlsen:
Did you find other places where new objects were assigned to a __uu pointer?

tgoyne:
No.

danielpovlsen:
If __uu is *much* faster than the default (__strong), wouldn't it make sense to work around the issue for `RLMAddLinkedObject()`? How critical is this function?

danielpovlsen:
I mean, we get back this object, just to figure out the row index, right?

tgoyne:
```
- (void)testAssignLink {
    RLMRealm *realm = RLMRealm.defaultRealm;

    CircleObject *circle = [CircleObject createInRealm:realm withObject:@[@"", NSNull.null]];

    [self measureBlock:^{
        for (int i = 0; i < 50000; ++i) {
            circle.next = circle;
        }
        [realm cancelWriteTransaction];
    }];
}
```

This hits the worst-case for this change and is slowed down from 525ms to 563ms on my 4S.

alazier:
We can add the optimization back in. This code is changed anyways on the swift branch - I can re-make these changes on this branch.

alazier:
Looks good to me :+1:

PR #1562, opened 2015-03-02
Add some tests for encryption stuff

### LINKED ISSUE #1533: Test encryption apis
Now that this is actually possible we need to thoroughly test our encryption apis.

tgoyne:
Adds some tests to verify that all of the key handling stuff does the right thing and a few places where it wasn't, removes the check for an attached debugger, and adds an option to use encrypted realms for all of the tests. Currently doesn't run the tests in encrypted mode on CI since #1551 is needed to make that not take an unreasonable amount of time.

Closes #1533.

@alazier


Realm.xcodeproj/project.pbxproj
```diff
@@ -85,6 +85,10 @@
 		3F512DAE19F9A1F300746016 /* SwiftUnicodeTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = E891759A197A1B600068ACC6 /* SwiftUnicodeTests.swift */; };
 		3F512DAF19F9A1F300746016 /* TransactionTests.m in Sources */ = {isa = PBXBuildFile; fileRef = E81A1FD11955FE0100FDED82 /* TransactionTests.m */; settings = {COMPILER_FLAGS = "-fobjc-arc-exceptions"; }; };
 		3F512DB019F9A1F300746016 /* UnicodeTests.m in Sources */ = {isa = PBXBuildFile; fileRef = E8917597197A1B350068ACC6 /* UnicodeTests.m */; settings = {COMPILER_FLAGS = "-fobjc-arc-exceptions"; }; };
+		3F68999B1A9FB04800DFDE6A /* EncryptionTests.m in Sources */ = {isa = PBXBuildFile; fileRef = 3F68999A1A9FB04800DFDE6A /* EncryptionTests.m */; };
```
jpsim: Looks like only the `EncryptionTests.m` changes are relevant in here. It'd be nice to clean this up a bit.

tgoyne: The files were unsorted, so I fixed that when adding another file. I can split it into a separate commit if you really want.

jpsim: No, not at all, let's keep that in here. Carry on :)

jpsim:
Awesome, @tgoyne. You get my :+1:.

PR #1563, opened 2015-03-03
Added support for bulk updates to RLMResults

astigsen:
Warning, rusty programmer ahead. Review carefully. I needed to de-stress a bit, so I implemented a fix for #1172

 This allows you to to bulk updates on query results, so you can nice one-liners like this:

```obj-c
[[Emails objectsWhere:@"unread = YES"] setValue:@NO forKey:"unread"];
```

@realm/cocoa

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ x.x.x Release notes (yyyy-MM-dd)
 
 ### Enhancements
 
+* Added support for bulk operations with `setValue:forKey:` to RLMResults.
 * The bowser will no longer show objects that have no persisted properties.
 
 ### Bugfixes
--- a/Realm/RLMResults.h
+++ b/Realm/RLMResults.h
@@ -225,6 +225,16 @@
 
 - (id)objectAtIndexedSubscript:(NSUInteger)index;
 
+/**
+ Bulk update values by invoking `setValue:forKey:` on each of the array's items using the specified `value` and `key`.
+ 
+ @warning This method can only be called during a write transaction.
+ 
+ @param value The object value.
+ @param key The name of the property.
+ */
+- (void)setValue:(id)value forKey:(NSString *)key;
+
 #pragma mark -
 
 /**---------------------------------------------------------------------------------------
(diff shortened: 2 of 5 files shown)


segiddins:
I believe I already have an open pull request that implements this?

-Samuel E. Giddins

On Mar 2, 2015, at 5:02 PM, astigsen <notifications@github.com> wrote:

Warning, rusty programmer ahead. Review carefully. I needed to de-stress a bit, so I implemented a fix for #1172

This allows you to to bulk updates on query results, so you can nice one-liners like this:

[[Emails objectsWhere:@"unread = YES"] setValue:@NO forKey:"unread"];
@realm/cocoa

You can view, comment on, or merge this pull request online at:

  https://github.com/realm/realm-cocoa/pull/1563

Commit Summary

Added support for bulk updates to RLMResults
File Changes

M CHANGELOG.md (1)
M Realm/RLMResults.h (10)
M Realm/RLMResults.mm (161)
M Realm/Tests/RLMTestObjects.h (2)
M Realm/Tests/ResultsTests.m (193)
Patch Links:

https://github.com/realm/realm-cocoa/pull/1563.patch
https://github.com/realm/realm-cocoa/pull/1563.diff

astigsen:
> I believe I already have an open pull request that implements this?

That would be great. If you have a more efficient implementation, we can just use that. What PR is it?

segiddins:
https://github.com/realm/realm-cocoa/pull/1536

astigsen:
Nice. It seems like it would be best if we could merge the two PR's. Then you could avoid all the unnecessary row accessor creations and object conversions.

segiddins:
Ah we can't avoid that because there's no guarantee the user is attempting to set the value for a persisted property (it might be a computed property). Ari and I already agreed that directly hitting the database might be a premature optimization (and an increase in complexity). 

-Samuel E. Giddins

On Mar 2, 2015, at 5:24 PM, astigsen <notifications@github.com> wrote:

Nice. It seems like it would be best if we could merge the two PR's. Then you could avoid all the unnecessary row accessor creations and object conversions.

astigsen:
Ok, I just benchmarked them, and it seems like my approach is around two times as fast as yours. That is probably not enough to justify the extra complexity, so we can just go with yours.

If there are any parts of my PR that you find useful as optimizations, you can just reuse them in your PR.

PR #1569, opened 2015-03-04
[RLMArray] Add `isInvalidated` to RLMArray.

### LINKED ISSUE #1492: Add isInvalidated to RLMArray
Hello, 
what is the proper way to check, whether some RLMCollection is still valid?

jpsim:
Closes #1492. /cc @tgoyne


Realm/RLMArrayLinkView.mm
```diff
@@ -79,6 +79,10 @@ - (NSUInteger)count {
     return _backingLinkView->size();
 }
 
+- (BOOL)isInvalidated {
+    return !_backingLinkView->is_attached() || !RLMMatchesThread(_realm);
```
tgoyne: Checking if it's on the correct thread is inconsistent with RLMObject and is conflating two completely different things.

jpsim: I started writing my reasoning and ended up agreeing with you in the process :smile:. I removed that check.

tgoyne:
:+1:

PR #1570, opened 2015-03-05
[RLMArray] Test deleting an RLMArray member updates count and maintains appropriate validation

jpsim:
I wrote this in response to a user reporting that invalid RLMArray member objects would be included in the RLMArray count. AFAICT we're not currently testing this. /cc @tgoyne

### DIFF
--- a/Realm/Tests/ArrayPropertyTests.m
+++ b/Realm/Tests/ArrayPropertyTests.m
@@ -97,6 +97,20 @@
     [realm commitWriteTransaction];
 }
 
+- (void)testDeleteObjectInArrayProperty {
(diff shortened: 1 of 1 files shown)


tgoyne:
I can't find any tests that cover exactly this (there's a few that dance around similar things), so :+1:.

My blind guess about the user's problem is that he's getting the count before refreshing (`dispatch_async(dispatch_get_main_queue(), ^{ .. })` will happen before autorefresh).

jpsim:
> My blind guess about the user's problem is that he's getting the count before refreshing

Cool, that's what I told him was likely happening.

PR #1575, opened 2015-03-05
Fix some small test issues

tgoyne:
Remove the RLM prefix from some test cases for consistency with the rest of the tests and add the proper license header to one file.

@jpsim


jpsim:
:+1:

PR #1576, opened 2015-03-05
Remove unused variable

tgoyne:
$SRCROOT isn't being set correctly, but it's not actually used anywhere or exported to subshells so just remove it.

@jpsim


jpsim:
:+1:

PR #1581, opened 2015-03-06
[Swift] Throw when RLMObject subclass is nested in *any* Swift declaration

### LINKED ISSUE #1577: Throw an exception when declaring a nested RLMObject subclass in Swift
Adding the code below causes this Assertion failed: top_size>=3 [0, 3]

We should prevent/warn the user about what they are doing

````swift
    func application(application: UIApplication, didFinishLaunchingWithOptions launchOptions: [NSObject: AnyObject]?) -> Bool {
        
        class CardDeck: RLMObject {
            dynamic var name = ""
        }
        
        let deck = CardDeck()
        deck.name = "New Deck"
        let realm = RLMRealm.defaultRealm()
        realm.beginWriteTransaction()
        realm.addObject(deck)
        realm.commitWriteTransaction()
        
        return true
    }
````

### PR DESCRIPTION
**Pull request by jpsim**
Fixes #1577. /cc @tgoyne

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,8 @@ x.x.x Release notes (yyyy-MM-dd)
 * Fix crash when calling `createOrUpdate:inRealm` with nested linked objects.
 * Use the key from `+[RLMRealm setEncryptionKey:forRealmsAtPath:]` in
   `-writeCopyToPath:error:` and `+migrateRealmAtPath:`.
+* Improved error message when an `RLMObject` subclass is defined nested within
+  another Swift declaration.
 
 0.90.6 Release notes (2015-02-20)
 =============================================================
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by tgoyne** [comment]
retest this please

**Comment by tgoyne** [comment]
:+1: 

Initial failure to run the CI job was due to github's API returning a 404.

**Comment by segiddins** [comment]
Could you add corresponding tests to the object scheme initialization tests on the swift branch?

**Comment by jpsim** [comment]
> Could you add corresponding tests to the object scheme initialization tests on the swift branch?

@segiddins Do you mean the tests in #1518? How would you suggest I test this given that the exception is thrown at initialize time?

**Comment by segiddins** [comment]
You subclass `SwiftFakeObject` inside another declaration, and then try and get an object schema for that subclass, and assert that it throws an exception,

**Comment by jpsim** [comment]
Ok, we'll just need to move the check from `+[RLMSchema initialize]` to `+[RLMObjectSchema schemaForObjectClass:]`.

**Comment by pbrewczynski** [comment]
Could you explain why realm class could not be nested inside other class?

**Comment by tgoyne** [comment]
Given the following schema:
```
class A {
  class Inner : Object { ... }
}
class B {
  class Inner : Object { ... }
}
class C : Object {
 var prop: A.Inner? = nil
}
```

The objective-c runtime tells us that `C.prop` has type `Inner` with no indication as to *which* `Inner` class it is.

PR #1588, opened 2015-03-10
[build.sh] Silence the output from running the JSONImport output

segiddins:
\c @alazier @jpsim


tgoyne:
Deleting the realm file between runs would also eliminate the spam and make the test more useful.

jpsim:
:+1:

PR #1590, opened 2015-03-10
[RLMObjectSchema] +dynamicSchemaForRealm: now respects search indexes

jpsim:
/cc @tgoyne @alazier

### DIFF
--- a/Realm/RLMObjectSchema.mm
+++ b/Realm/RLMObjectSchema.mm
@@ -195,7 +195,7 @@
         RLMProperty *prop = [[RLMProperty alloc] initWithName:name
                                                          type:RLMPropertyType(table->get_column_type(col))
                                               objectClassName:nil
-						      indexed:table->has_search_index(col)];
+                                                      indexed:table->has_search_index(col)];
         prop.column = col;
         if (prop.type == RLMPropertyTypeObject || prop.type == RLMPropertyTypeArray) {
             // set link type for objects and arrays
@@ -272,8 +272,11 @@
     NSArray *otherProperties = objectSchema.properties;
     for (NSUInteger i = 0; i < _properties.count; i++) {
         RLMProperty *p1 = _properties[i], *p2 = otherProperties[i];
-	if (![p1 isEqualToProperty:p2] ||
-	    p1.column != p2.column) {
+        if (p1.type != p2.type ||
+            p1.column != p2.column ||
+            p1.isPrimary != p2.isPrimary ||
+            ![p1.name isEqualToString:p2.name] ||
+            !(p1.objectClassName == p2.objectClassName || [p1.objectClassName isEqualToString:p2.objectClassName])) {
             return NO;
         }
     }
--- a/Realm/RLMObjectSchema.mm
+++ b/Realm/RLMObjectSchema.mm
@@ -195,7 +195,7 @@
         RLMProperty *prop = [[RLMProperty alloc] initWithName:name
                                                          type:RLMPropertyType(table->get_column_type(col))
                                               objectClassName:nil
-                                                      indexed:NO];
+						      indexed:table->has_search_index(col)];
         prop.column = col;
         if (prop.type == RLMPropertyTypeObject || prop.type == RLMPropertyTypeArray) {
             // set link type for objects and arrays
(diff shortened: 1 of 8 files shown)


Realm/RLMProperty.mm
```diff
@@ -306,6 +306,7 @@ - (id)copyWithZone:(NSZone *)zone {
 
 - (BOOL)isEqualToProperty:(RLMProperty *)property {
     return _type == property->_type
+	&& _indexed == property->_indexed
```
segiddins: is this indentation off?

Realm/RLMObjectSchema.mm
```diff
@@ -285,11 +285,8 @@ - (BOOL)isEqualToObjectSchema:(RLMObjectSchema *)objectSchema {
     NSArray *otherProperties = objectSchema.properties;
     for (NSUInteger i = 0; i < _properties.count; i++) {
         RLMProperty *p1 = _properties[i], *p2 = otherProperties[i];
-        if (p1.type != p2.type ||
```
tgoyne: We don't want to be checking for `indexed` being equal in the place this is used (checking if we can use the same accessor classes).

tgoyne:
This needs a test that explicitly verifies that the `indexed` property is being read property.

Realm/RLMObjectSchema.mm
jpsim: But we do want to check for `isPrimary`? Can you explain why, I don't understand the distinction wrt accessors.

tgoyne: The setter for primary keys checks for uniqueness.

alazier:
I'm not sure why this pr is needed at all. What is the problem that you are trying to solve? I don't think indexed shouldn't be used to determine property or schema equality at all. Add/removing an index does not change the schema, and should not require migrations.

jpsim:
> Add/removing an index does not change the schema, and should not require migrations.

Shouldn't adding an index to a property add an index to that column and therefore require a migration?

The main goal of this PR is to make `+[RLMSchema dynamicSchemaForRealm:]` actually include the indexing settings present in the file in the schema. The other change is really just to simplify testing. I could leave the check for `indexed` out of `isEqualToProperty:` but it really belongs in there IMO.

tgoyne:
Adding a property to `indexedProperties` should make it add an index to existing Realm files (and I think it currently doesn't), but adding or removing a search index doesn't require any migration logic, so like adding new tables it should be allowed without a schema version bump.

alazier:
Another issue is that you can't tell if there is an index on property from a dynamic schema due to the property being primary, or if the user explicitly marked the property as indexed. The real solution to this problem is to actually have an indexedProperties metadata table to store this information.

tgoyne:
When does it matter that the user redundantly specified that the pk should be indexed?

jpsim:
Regardless of *why* the property has an index, its `indexed` property should be set.

alazier:
@tgoyne it doesn't matter until we make the dynamic api public. At that point if a user change the primary key we need to know whether the property should still be indexed or not. Right now I don't think there is not a scenario where this can occur. 

I'm fine with the change of populating the indexed property for dynamic schema so :+1:

tgoyne:
Even with the dynamic API I don't really see how it'd be an issue. If they give us an `RLMProperty` with `primary = NO` and `indexed = YES` then it'll stay indexed when we make it not a pk, and if it has `indexed = NO` the index should be removed. I don't think it makes sense to store some metadata that tries to cleverly override what the user tells us to do based on what we think they intended.

tgoyne:
The updated test still doesn't actually verify that the indexed property is read correctly or that `isEqualToProperty:` is correct; it just verifies that either neither or both of them are incorrect.

alazier:
I don't think it makes sense to worry about any of this right now as there are clearly higher priorities at the moment.

jpsim:
> I don't think it makes sense to worry about any of this right now as there are clearly higher priorities at the moment.

I really just needed this because I had to use the dynamic API to help debug a support issue. I'll address @tgoyne's point about adding a test for `isEqualToProperty:` and then hopefully we can get this in.

alazier:
That's fair enough. I'm trying to get all of the other swift stuff rebased/merged so don't have the time to debate this now. I'm fine with the current pr - isEqualToProperty isn't used internally so I'm not too worried about it, but @tgoyne is probably correct that it should be tested - we could also just abandon that change as well for the time being if it a lot of effort.

jpsim:
`isEqualToProperty:` now has some tests. I could spend more time covering more cases, but that doesn't seem worthwhile. I could keep this PR open and add more tests later if you prefer. @tgoyne @alazier

PR #1593, opened 2015-03-11
Make Realm collections conform to CollectionType

segiddins:
\c @jpsim @alazier

### DIFF
--- a/RealmSwift/List.swift
+++ b/RealmSwift/List.swift
@@ -196,8 +196,6 @@ public final class List<T: Object>: ListBase, ExtensibleCollectionType {
 
     // MARK: Collection Support
 
-    typealias Index = Int
-
     /// The position of the first element in a non-empty collection.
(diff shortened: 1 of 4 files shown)


segiddins:
We could also make `List` conform to `RangeReplaceableCollectionType`, if we want to.

jpsim:
Unless there are technical limitations from doing so, it'd be nice to have a `RealmCollectionType` protocol conforming to `CollectionType` and add other realm niceties.

segiddins:
Rebased

jpsim:
Nice! :+1: :+1:

alazier:
I think the ordering of methods in Results is kind of confusing - a bunch of protocol implementation interspersed with other methods. Since we also don't have explicit tests for this might be nice to have these methods in an extension or at the end of the class.

segiddins:
@alazier done

alazier:
That seems a lot more sensible. :+1:

PR #1612, opened 2015-03-12
Fix a crash in CFRunLoopSourceInvalidate

tgoyne:
It turns out that the documentation's recommendation to release run loop sources immediate after adding them to the run loop is a rather bad idea, as the run loop releases its sources when it's shut down, so we'd crash if the kqueue waiter outlived the source thread.

The test forces this to happen by holding on to the RLMRealm after the thread dies, but it can happen inconsistently without that.

@alazier

### DIFF
--- a/Realm/Tests/RealmTests.mm
+++ b/Realm/Tests/RealmTests.mm
@@ -1158,4 +1158,22 @@ extern "C" {
     }
 }
 
+- (void)testHoldRealmAfterSourceThreadIsDestroyed {
+    __block RLMRealm *realm;
+
+    // Using an NSThread to ensure the thread (and thus runloop) is actually destroyed
+    NSThread *thread = [[NSThread alloc] initWithTarget:self selector:@selector(runBlock:) object:^{
+        realm = [RLMRealm defaultRealm];
+    }];
+    [thread start];
+    while (!thread.isFinished)
+        usleep(100);
+
+    [realm path]; // ensure ARC releases the object after the thread has finished
+}
+
+- (void)runBlock:(void (^)())block {
+    block();
+}
+
 @end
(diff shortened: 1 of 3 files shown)


alazier:
Does this fix the issues with device tests we have been seeing?

tgoyne:
I have no idea. Two users reported crashes from this.

Realm/Tests/RealmTests.mm
```diff
@@ -1158,4 +1158,22 @@ - (void)testNotificationPipeBufferOverfull {
...
+
+    // Using an NSThread to ensure the thread (and thus runloop) is actually destroyed
+    NSThread *thread = [[NSThread alloc] initWithTarget:self selector:@selector(runBlock:) object:^{
+        realm = [RLMRealm defaultRealm];
+    }];
+    [thread start];
+    while (!thread.isFinished)
+        usleep(100);
+
+    [realm path]; // ensure ARC releases the object after the thread has finished
```
alazier: How does this ensure the object is released?

Could we not use a __weak variable to check this?

tgoyne: It ensures that the object *isn't* released before this point.

alazier:
The code changes look fine. I assume the test crashes without the fix?

tgoyne:
Yes, the test crashes in CFRunLoopSourceInvalidate.

Realm/Tests/RealmTests.mm
alazier: Well would a __weak variable work for that? The current code is a bit confusing, even with the comment.

tgoyne: No, a __weak variable would be the exact opposite of what is needed. The point is to have the RLMRealm outlive the NSThread it's created on.

alazier:
:+1:

PR #1628, opened 2015-03-13
Don't allow NUL bytes in indexed strings

tgoyne:
Might be enough to allow us to continue to index primary keys by default if we're confident that embedded nulls are the only problem, and even if we do make it opt-in we need this check until string indexes are fixed.

@alazier

### DIFF
--- a/Realm/Tests/ObjectTests.m
+++ b/Realm/Tests/ObjectTests.m
@@ -1009,6 +1009,9 @@ RLM_ARRAY_TYPE(PrimaryIntObject);
     IndexedObject *io = [IndexedObject createInDefaultRealmWithObject:@[@"", @0]];
     XCTAssertThrows(io.name = @"a\0b");
     XCTAssertThrows(([IndexedObject createInDefaultRealmWithObject:@[@"a\0b", @0]]));
+    io = [[IndexedObject alloc] init];
+    io.name = @"a\0b";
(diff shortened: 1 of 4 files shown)


alazier:
Are we sure string properties without indexes work properly with NULL bytes? It would be great if we had a test which was confirmed to crash when using indexes so we could confirm that NULL bytes work otherwise. If not we will need to do this check for all string properties.

Realm/Tests/ObjectTests.m
```diff
@@ -1002,6 +1002,27 @@ - (void)testIndex
     XCTAssertFalse(ageProperty.indexed, @"non-indexed property shouldn't have an index");
 }
 
+- (void)testIndexedPropertyRejectsNullBytesInString
+{
+    RLMRealm *realm = RLMRealm.defaultRealm;
+    [realm beginWriteTransaction];
+    IndexedObject *io = [IndexedObject createInDefaultRealmWithObject:@[@"", @0]];
+    XCTAssertThrows(io.name = @"a\0b");
+    XCTAssertThrows(([IndexedObject createInDefaultRealmWithObject:@[@"a\0b", @0]]));
```
segiddins: we should probably also test that an exception is thrown when adding a standalone object with an invalid string property

tgoyne:
Added this check in core instead.

PR #1630, opened 2015-03-13
Change Object.init(object:) to Object.init(value:)

### LINKED ISSUE #1585: Rename initWithObject/createWithObject to withValue
We need to make this change in objective-c to be consistent with the new swift apis and to make things more clear/less confusing. 

This change should be made once al-swift is merged to master, at which point in time we can fix Object.init(object:)

Old api calls should be deprecated for a few releases.

alazier:
Closes #1585 

@segiddins

### DIFF
--- a/examples/ios/swift-next/Backlink/AppDelegate.swift
+++ b/examples/ios/swift-next/Backlink/AppDelegate.swift
@@ -26,7 +26,7 @@ class Dog: Object {
     var owners: [Person] {
         // Realm doesn't persist this property because it only has a getter defined
         // Define "owners" as the inverse relationship to Person.dogs
-        return linkingObjects(Person.self, forProperty: "dogs") as [Person]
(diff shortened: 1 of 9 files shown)


examples/ios/swift-next/Backlink/AppDelegate.swift
```diff
@@ -26,7 +26,7 @@ class Dog: Object {
     var owners: [Person] {
         // Realm doesn't persist this property because it only has a getter defined
         // Define "owners" as the inverse relationship to Person.dogs
-        return linkingObjectsOfClass("Person", forProperty: "dogs") as [Person]
+        return linkingObjects(Person.self, forProperty: "dogs") as [Person]
```
segiddins: the `as [Person]` is unnecessary

segiddins:
:+1: other than comment

segiddins:
@alazier I think I fixed the CI errors on the examples, please take a look at my changes

PR #1639, opened 2015-03-16
New Swift apis

alazier:
This is finally ready for review.

@tgoyne @segiddins @jpsim


Realm.xcodeproj/xcshareddata/xcschemes/iOS Dynamic.xcscheme
```diff
@@ -32,13 +32,30 @@
...
                ReferencedContainer = "container:Realm.xcodeproj">
             </BuildableReference>
+            <SkippedTests>
+               <Test
+                  Identifier = "SwiftObjectInterfaceTests/testNSStringFromClassDemangledTopLevelClassNames()">
+               </Test>
+               <Test
+                  Identifier = "SwiftObjectInterfaceTests/testNestedClassNameMangling()">
+               </Test>
+            </SkippedTests>
```
tgoyne: Do these still need to be skipped?

Realm/RLMArray.h
```diff
@@ -300,8 +300,3 @@
 
 @end
 
-
-@interface RLMArray (Private)
-- (instancetype)initWithObjectClassName:(NSString *)objectClassName;
```
tgoyne: Doesn't this still need to be in the public header for swift stuff using the obj-c API?

Realm.xcodeproj/xcshareddata/xcschemes/OSX.xcscheme
```diff
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <Scheme
-   LastUpgradeVersion = "0600"
+   LastUpgradeVersion = "0610"
```
segiddins: should this be `0620`?

Realm/RLMArrayLinkView.mm
```diff
@@ -138,7 +141,7 @@ - (void)addObject:(RLMObject *)object {
     RLMLinkViewArrayValidateInWriteTransaction(self);
 
     RLMValidateObjectClass(object, self.objectClassName);
-    if (object.realm != self.realm) {
+    if (object->_realm != self.realm) {
```
tgoyne: I think using `_realm` in all the spots where we're comparing it to `object->_realm` would look a little better. Mixing ivar and property syntax for the same sort of thing looks a little off.

Realm/RLMArrayLinkView.mm
```diff
@@ -194,12 +200,13 @@ - (void)replaceObjectAtIndex:(NSUInteger)index withObject:(RLMObject *)object {
...
     // check attached for table and object
     RLMLinkViewArrayValidateAttached(self);
+    RLMValidateObjectClass(object, self.objectClassName);
     if (object->_realm && !object->_row.is_attached()) {
         @throw RLMException(@"RLMObject is no longer valid");
     }
 
     // check that object types align
-    if (![_objectClassName isEqualToString:object.objectSchema.className]) {
+    if (![_objectClassName isEqualToString:object->_objectSchema.className]) {
```
tgoyne: This check can be removed (it's done in RLMValidateObjectClass now).

tgoyne: Although note that the error message in RLMValidateObjectClass is wrong for this case.

Realm/RLMArrayLinkView.mm
```diff
@@ -64,8 +63,12 @@ static inline void RLMLinkViewArrayValidateInWriteTransaction(__unsafe_unretaine
         @throw RLMException(@"Can't mutate a persisted array outside of a write transaction.");
     }
 }
-static inline void RLMValidateObjectClass(__unsafe_unretained RLMObject *const obj, __unsafe_unretained NSString *const expected) {
-    NSString *objectClassName = obj.objectSchema.className;
+static inline void RLMValidateObjectClass(__unsafe_unretained RLMObjectBase *const obj, __unsafe_unretained NSString *const expected) {
+    if (!obj) {
+        @throw RLMException(@"Trying to add `nil` object");
```
segiddins: maybe it would also be helpful to say what `expected` is?

Realm/RLMArrayLinkView.mm
```diff
@@ -148,7 +151,10 @@ - (void)insertObject:(RLMObject *)object atIndex:(NSUInteger)index {
     RLMLinkViewArrayValidateInWriteTransaction(self);
 
     RLMValidateObjectClass(object, self.objectClassName);
-    if (object.realm != self.realm) {
+    if (index > _backingLinkView->size()) {
+        @throw RLMException(@"Trying to insert object at invalid index");
```
segiddins: this message could be better -- `NSArray` says `Terminating app due to uncaught exception 'NSRangeException', reason: '*** -[__NSArrayM insertObject:atIndex:]: index 200 beyond bounds [0 .. 0]'`

Realm/RLMArrayLinkView.mm
```diff
@@ -194,12 +200,13 @@ - (void)replaceObjectAtIndex:(NSUInteger)index withObject:(RLMObject *)object {
 - (NSUInteger)indexOfObject:(RLMObject *)object {
     // check attached for table and object
     RLMLinkViewArrayValidateAttached(self);
+    RLMValidateObjectClass(object, self.objectClassName);
```
segiddins: why was this added? We already check the tables for equality and return `NSNotFound`

Realm/RLMListBase.m
```diff
@@ -26,4 +28,9 @@ - (instancetype)initWithArray:(RLMArray *)array {
     }
     return self;
 }
-@end
\ No newline at end of file
+
+- (NSUInteger)countByEnumeratingWithState:(NSFastEnumerationState *)state objects:(id __unsafe_unretained [])buffer count:(NSUInteger)len {
+    return [__rlmArray countByEnumeratingWithState:state objects:buffer count:len];
+}
+
```
segiddins: should also forward `description`

Realm/RLMMigration.mm
```diff
@@ -142,6 +145,9 @@ - (void)execute:(RLMMigrationBlock)block {
 
         // verify uniqueness for any new unique columns before committing
         [self verifyPrimaryKeyUniqueness];
+
+        // reset schema to saved schema since it has been altered
+        RLMRealmSetSchema(_realm, savedSchema, true);
```
tgoyne: Was the lack of this an existing bug on master? It'd be nice to have a test that'd break on `isPrimary` not being reset properly if there were actually any effects other than the user-visible object schema being wrong.

tgoyne: Also while it's not needed right now, I think putting this above `verifyPrimaryKeyUniqueness` will be less likely to result in funny bugs if we change how `verifyPrimaryKeyUniqueness` works.

Realm/RLMObject.mm
```diff
@@ -16,12 +16,13 @@
 //
 ////////////////////////////////////////////////////////////////////////////
 
+#import "RLMRealm_Private.hpp"
 #import "RLMObject_Private.h"
```
tgoyne: `#import "RLMObject_Private.h"` should be before `#import "RLMRealm_Private.hpp"`.

Realm/RLMObjectBase.mm
```diff
@@ -159,32 +156,67 @@ - (NSArray *)linkingObjectsOfClass:(NSString *)className forProperty:(NSString *
...
-- (id)objectForKeyedSubscript:(NSString *)key {
-    if (_realm) {
-        return RLMDynamicGet(self, key);
+id RLMObjectBaseObjectForKeyedSubscript(RLMObjectBase *object, NSString *key) {
+    if (!object) {
+        return nil;
+    }
+
+    if (object->_realm) {
+        return RLMDynamicGet(object, key);
```
segiddins: @tgoyne is doing `RLMDynamicGet` here any better than `valueForKey:`?

Realm/RLMObject.mm
```diff
@@ -32,14 +33,14 @@
 @implementation RLMObject
 
 // These properties are synthesized in RLMObjectBase
-@dynamic objectSchema, realm, invalidated;
+@dynamic invalidated;
```
tgoyne: Only one of them now.

Realm/RLMObjectBase.mm
```diff
@@ -88,9 +81,9 @@ - (instancetype)initWithRealm:(__unsafe_unretained RLMRealm *const)realm
     if (self) {
         _realm = realm;
         _objectSchema = schema;
-        if (useDefaults) {
+        if (useDefaults && !schema.isSwiftClass) {
```
tgoyne: afaict `useDefaults` is only used in the standalone init case, so it seems simpler to just move all this to that method.

Realm/RLMObject_Private.h
```diff
@@ -28,17 +28,49 @@
     __unsafe_unretained RLMObjectSchema *_objectSchema;
 }
 
+// standalone initializer
+- (instancetype)initWithObject:(id)value schema:(RLMSchema *)schema;
+
+// liver accessor initializer
```
segiddins: `liver`?

Realm/RLMUtil.mm
```diff
@@ -171,7 +175,7 @@ id RLMValidatedObjectForProperty(id obj, RLMProperty *prop, RLMSchema *schema) {
             RLMObjectSchema *objSchema = schema[prop.objectClassName];
             RLMArray *objects = [[RLMArray alloc] initWithObjectClassName: objSchema.className standalone:YES];
             for (id el in obj) {
-                [objects addObject:[[objSchema.objectClass alloc] initWithObject:el]];
+                [objects addObject:[[objSchema.objectClass alloc] initWithObject:el schema:schema]];
```
segiddins: should the schema be `objcSchema`?

Realm/Realm.h
```diff
@@ -25,3 +25,5 @@
 #import <Realm/RLMRealm.h>
 #import <Realm/RLMResults.h>
 #import <Realm/RLMSchema.h>
+#import <Realm/RLMObjectSchema.h>
```
segiddins: were these purposefully removed from `Realm.h` before?

Realm/Tests/PropertyTests.m
```diff
@@ -1,44 +1,21 @@
-////////////////////////////////////////////////////////////////////////////
```
segiddins: revert the change to the licence header

segiddins: actually, I believe this whole file can be reverted?

Realm/Tests/RealmSwift/SwiftObjectInterfaceTests.swift
```diff
@@ -0,0 +1,109 @@
...
+////////////////////////////////////////////////////////////////////////////
+
+import XCTest
+import RealmSwift
+
+class SwiftDefaultObject: Object {
+    dynamic var intCol = 1
+    dynamic var boolCol = true
+
+    override class func defaultPropertyValues() -> [NSObject : AnyObject]! {
```
segiddins: this func no longer is used

Realm/RLMObjectBase.mm
```diff
@@ -133,24 +106,48 @@ + (RLMObjectSchema *)sharedSchema {
     return RLMSchema.sharedSchema[self.className];
 }
 
-- (NSArray *)linkingObjectsOfClass:(NSString *)className forProperty:(NSString *)property {
-    if (!_realm) {
+void RLMObjectBaseSetRealm(__unsafe_unretained RLMObjectBase *object, __unsafe_unretained RLMRealm *realm) {
```
tgoyne: I don't think we need these nil checks -- these functions appear to only ever be called on self or non-optional objects. Currently most of the C functions don't check for nil, and doing it inconsistently is kinda dangerous.

tgoyne: Annotating every pointer with `REALM_NULLABLE`/`REALM_NONNULL` and building with a 6.3 beta would be a good idea that'll probably catch at least one mismatch between our expectations and implementation.

segiddins: :+1: that's been on my list of things to do, but it will only check when a literal `nil` is passed through

Realm/RLMObjectBase.mm
```diff
@@ -159,32 +156,67 @@ - (NSArray *)linkingObjectsOfClass:(NSString *)className forProperty:(NSString *
...
+    // if not the correct types throw
+    if ((o1 && ![o1 isKindOfClass:RLMObjectBase.class]) || (o2 && ![o2 isKindOfClass:RLMObjectBase.class])) {
+        @throw RLMException(@"Can only compare objects of class RLMObjectBase");
+    }
+    // if identical object
+    if (o1 == o2) {
+        return YES;
+    }
+    // if one is nil
+    if (!!o1 != !!o2) {
```
tgoyne: `o1 == nil || o2 == nil` really isn't long enough to need a clever shortening and explanatory comment.

RealmSwift/Migration.swift
```diff
@@ -0,0 +1,233 @@
...
+public typealias MigrationObject = DynamicObject
+
+/**
+Provides both the old and new versions of an object in this Realm. Objects properties can only be
+accessed using subscripting.
+
+:param: oldObject Object in original `Realm` (read-only)
+:param: newObject Object in migrated `Realm` (read-write)
+*/
+public typealias MigrationObjectEnumerateBlock = (oldObject: MigrationObject, newObject: MigrationObject) -> Void
```
segiddins: shouldn't those objects be optional?

Realm/RLMObjectSchema.mm
```diff
@@ -105,7 +91,7 @@ + (instancetype)schemaForObjectClass:(Class)objectClass {
     Class cls = objectClass;
     Class superClass = class_getSuperclass(cls);
     NSArray *props = @[];
-    while (superClass != RLMObjectBase.class) {
+    while (superClass && superClass != RLMObjectBase.class) {
```
tgoyne: Why is this change needed? It only behaves differently if `RLMObjectBase.class` is nil.

RealmSwift/Migration.swift
```diff
@@ -0,0 +1,233 @@
...
+    :param: className The name of the `Object` class to create.
+    :param: object    The object used to populate the object. This can be any key/value coding
+                      compliant object, or a JSON object such as those returned from the methods in
+                      `NSJSONSerialization`, or an `Array` with one object for each persisted
+                      property. An exception will be thrown if any required properties are not
+                      present and no default is set.
+    
+    :returns: The created object.
+    */
+    public func create(className: String, value: AnyObject) -> MigrationObject {
```
segiddins: `value` should have a default value?

Realm/RLMObjectStore.mm
```diff
@@ -28,6 +28,7 @@
 #import "RLMRealm_Private.hpp"
 #import "RLMSchema_Private.h"
 #import "RLMUtil.hpp"
+#import "RLMSwiftSupport.h"
```
tgoyne: sort imports

Realm/RLMObjectSchema.mm
segiddins: Needed so we can get test schemas that throw exceptions (and `NSObject` in Cocoa has properties)

RealmSwift/Object.swift
```diff
@@ -0,0 +1,295 @@
...
+    WARNING: This is an internal initializer not intended for public use.
+    */
+    public override init(realm: RLMRealm, schema: RLMObjectSchema, defaultValues: Bool) {
+        super.init(realm: realm, schema: schema, defaultValues: defaultValues)
+    }
+
+    /**
+    WARNING: This is an internal initializer not intended for public use.
+    */
+    public override init(object: AnyObject, schema: RLMSchema) {
```
segiddins: didnt we just get rid of this method?

Realm/RLMUtil.mm
```diff
@@ -188,14 +192,10 @@ id RLMValidatedObjectForProperty(id obj, RLMProperty *prop, RLMSchema *schema) {
...
-    // for swift merge
-    // FIXME: for new apis only return swift initialized values
-    NSMutableDictionary *defaults = [NSMutableDictionary dictionaryWithDictionary:[objectSchema.objectClass defaultPropertyValues]];
+    NSMutableDictionary *defaults = [NSMutableDictionary dictionary];
     RLMObject *defaultObject = [[objectSchema.objectClass alloc] init];
     for (RLMProperty *prop in objectSchema.properties) {
-        if (!defaults[prop.name] && defaultObject[prop.name]) {
-            defaults[prop.name] = defaultObject[prop.name];
-        }
+        defaults[prop.name] = defaultObject[prop.name];
```
tgoyne: This change breaks existing Swift applications.

Realm/Realm.h
```diff
@@ -25,3 +25,5 @@
...
+#import <Realm/RLMProperty.h>
```
tgoyne: sort imports

RealmSwift/Object.swift
```diff
@@ -0,0 +1,295 @@
...
+        super.init(object: object, schema: schema)
+    }
+
+    /// Get RLMArray values when getting array properties
+    public override func valueForKey(key: String) -> AnyObject? {
```
segiddins: we really should have better docs for KVC :)

RealmSwift/Object.swift
```diff
@@ -0,0 +1,295 @@
...
+            if let value = value as? NSFastEnumeration {
+                list._rlmArray.removeAllObjects()
+                list._rlmArray.addObjects(value)
+            }
+            return
+        }
+        super.setValue(value, forKey: key)
+    }
+
+    public subscript(key: String) -> AnyObject? {
```
segiddins: docs

Realm/RLMObjectSchema.mm
tgoyne: I don't understand what that has to do with adding an extra check for nil.

Realm/RLMObjectBase.mm
tgoyne: skips some pointless indirection

Realm/Tests/ArrayPropertyTests.m
```diff
@@ -127,6 +127,23 @@ -(void)testInsertMultiple {
...
+    ArrayPropertyObject *obj = [ArrayPropertyObject createInRealm:realm withObject:@[@"arrayObject", @[], @[]]];
+    StringObject *child1 = [StringObject createInRealm:realm withObject:@[@"a"]];
+    StringObject *child2 = [[StringObject alloc] init];
+    child2.stringCol = @"b";
+    [obj.array addObject:child2];
+    XCTAssertThrows([obj.array insertObject:child1 atIndex:2]);
+    [realm commitWriteTransaction];
+
+    RLMArray *children = obj.array;
+    XCTAssertEqual(children.count, (NSUInteger)1);
```
tgoyne: `1U` rather than `(NSUInteger)1`.

Realm/RLMObjectSchema.mm
segiddins: SwiftFakeObjectWithURL <- SwiftFakeObject <- NSObject

With this change, only `SwiftFakeObjectWithURL`'s properties will be considered, exactly as if it were a subclass of `RealmSwift.Object`

tgoyne: Oh, I misread the conditional as doing the exact opposite of what it does.

Realm/Tests/RealmSwift/ResultsTests.swift
```diff
@@ -0,0 +1,235 @@
...
+        var str = ""
+        for obj in results {
+            str += obj.stringCol
+        }
+
+        XCTAssertEqual(str, "12")
+    }
+}
+
+class ResultsFromTableViewTests: ResultsTests {
```
tgoyne: This is actually `ResultsFromTableTests`.

Realm/Tests/RealmSwift/SwiftArrayTests.swift
```diff
@@ -0,0 +1,143 @@
+////////////////////////////////////////////////////////////////////////////
```
tgoyne: Delete these old test files? Everything in them should now be covered by better tests.

Realm/Tests/RealmTests.mm
```diff
@@ -1100,6 +1100,22 @@ - (void)testRefreshCreatesAReadTransaction
     XCTAssertEqual(2U, [IntObject allObjects].count);
 }
 
+- (void)testBadEncryptionKeys
```
tgoyne: remove these (they were moved to a different file on master)

Realm/module.modulemap
```diff
@@ -11,6 +11,7 @@ framework module Realm {
         header "RLMSchema_Private.h"
         header "RLMRealm_Private.h"
         header "RLMObject_Private.h"
+        header "RLMArray_Private.h"
```
tgoyne: can these be sorted?

RealmSwift/Object.swift
```diff
@@ -0,0 +1,295 @@
...
+                    if let list = listProperties[key] {
+                        return list
+                    }
+                    let list = List<DynamicObject>()
+                    listProperties[key] = list
+                    return list
+
+                }
+            }
+            return RLMObjectBaseObjectForKeyedSubscript(self, key)
```
segiddins: this can just return `super`?

RealmSwift/Object.swift
```diff
@@ -0,0 +1,295 @@
...
+            if mirror.valueType is RLMListBase.Type {
+                properties.append(reflection[i].0)
+            }
+        }
+
+        return properties
+    }
+
+    @objc private class func initializeListProperty(object: RLMObjectBase?, property: RLMProperty?, array: RLMArray?) {
+        let list = (object as Object)[property!.name]! as RLMListBase
```
segiddins: this really should not be force unwrapping -- optional binding should be used here

RealmSwift.xcodeproj/project.pbxproj
```diff
@@ -0,0 +1,1351 @@
...
+			target = C0E04FAA1AB2366200DCC8C0 /* TestHost */;
+			targetProxy = C0E04FD21AB2367A00DCC8C0 /* PBXContainerItemProxy */;
+		};
+/* End PBXTargetDependency section */
+
+/* Begin XCBuildConfiguration section */
+		0201B2791A54C68C004CFB6F /* Debug */ = {
+			isa = XCBuildConfiguration;
+			buildSettings = {
+				CLANG_CXX_LANGUAGE_STANDARD = "gnu++0x";
```
tgoyne: Shouldn't matter unless we add C++ files to this project, but this should be c++14.

RealmSwift.xcodeproj/project.pbxproj
```diff
@@ -0,0 +1,1351 @@
...
+			buildSettings = {
+				CLANG_ENABLE_MODULES = YES;
+				CODE_SIGN_IDENTITY = "iPhone Developer";
+				FRAMEWORK_SEARCH_PATHS = (
+					"$(SDKROOT)/Developer/Library/Frameworks",
+					"$(inherited)",
+					"$(CONFIGURATION_BUILD_DIR)/RealmSwift.framework/Frameworks",
+				);
+				INFOPLIST_FILE = "Realm/Tests/RealmTests-Info.plist";
+				IPHONEOS_DEPLOYMENT_TARGET = 8.1;
```
tgoyne: Should be 8.0 unless we actually need 8.1.

RealmSwift.xcodeproj/xcshareddata/xcbaselines/3F8DCA5619930F550008BD7F.xcbaseline/7F8339ED-0034-40BF-9189-6C569D2814B7.plist
```diff
@@ -0,0 +1,202 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
tgoyne: This file appears to be for the obj-c perf tests.

RealmSwift/List.swift
```diff
@@ -0,0 +1,310 @@
...
+
+    :returns: The index of the given object, or `nil` if the object is not in the list.
+    */
+    public func indexOf(predicate: NSPredicate) -> Int? {
+        return notFoundToNil(_rlmArray.indexOfObjectWithPredicate(predicate))
+    }
+
+    /**
+    Returns the index of the first object matching the given predicate,
+    or `nil` if the object is not in the list.
```
tgoyne: or `nil` if none match.

RealmSwift/List.swift
```diff
@@ -0,0 +1,310 @@
...
+
+    :param: predicateFormat The predicate format string, optionally followed by a variable number
```
tgoyne: Maybe "The `NSPredicate` format string"? There currently doesn't seem to be any explicit mention of what a "predicate" is anywhere.

RealmSwift/List.swift
```diff
@@ -0,0 +1,310 @@
...
+                            of arguments.
+
+    :returns: The index of the given object, or `nil` if the object is not in the list.
```
tgoyne: The index of the matching object, or `nil` if there are none in the list.

RealmSwift/List.swift
```diff
@@ -0,0 +1,310 @@
...
+                return self._rlmArray[i++] as? T
+            }
+        }
+    }
+
+    // MARK: Mutation
+
+    /**
+    Appends the given object to the end of the list.
+
```
tgoyne: Adds the object to the List's Realm if the object is standalone and the List is persisted in a Realm, and copies the object to the List's Realm if they are persisted in different Realms.

RealmSwift/List.swift
```diff
@@ -0,0 +1,310 @@
...
+    :param: index The index at which to remove the object.
+    */
+    public func remove(index: Int) {
+        throwForNegativeIndex(index)
+        _rlmArray.removeObjectAtIndex(UInt(index))
+    }
+
+    /**
+    Removes the given object from the list. Does not remove the object from the Realm.
```
tgoyne: Removes the first occurrence of the given object from the list.

RealmSwift/List.swift
```diff
@@ -0,0 +1,310 @@
...
+        for obj in SequenceOf<T>(objects) {
+            _rlmArray.addObject(unsafeBitCast(obj, RLMObject.self))
+        }
+    }
+
+    /**
+    Inserts the given object at the given index.
+
+    :warning: This method can only be called during a write transaction.
+    :warning: Throws an exception when called with an index smaller than zero or greater than the
```
tgoyne: greater than or equal

RealmSwift/List.swift
```diff
@@ -0,0 +1,310 @@
...
+    public func insert(object: T, atIndex index: Int) {
+        throwForNegativeIndex(index)
+        _rlmArray.insertObject(unsafeBitCast(object, RLMObject.self), atIndex: UInt(index))
+    }
+
+    /**
+    Removes the object at the given index from the list. Does not remove the object from the Realm.
+
+    :warning: This method can only be called during a write transaction.
+    :warning: Throws an exception when called with an index smaller than zero or greater than the
```
tgoyne: greater than or equal

RealmSwift/List.swift
```diff
@@ -0,0 +1,310 @@
...
+    :param: object An object.
+    */
+    public func remove(object: T) {
+        if let index = indexOf(object) {
+            remove(index)
+        }
+    }
+
+    /**
+    Removes the last object in the list. Does not remove the object from the Realm.
```
tgoyne: should mention what happens if it's already empty

RealmSwift/List.swift
```diff
@@ -0,0 +1,310 @@
...
+    */
+    public func removeAll() {
+        _rlmArray.removeAllObjects()
+    }
+
+    /**
+    Replaces an object at the given index with a new object.
+
+    :warning: This method can only be called during a write transaction.
+    :warning: Throws an exception when called with an index smaller than zero or greater than the
```
tgoyne: greater than or equal

RealmSwift/Migration.swift
tgoyne: They'd only be nil if you add/delete objects before enumerating existing objects, and I suspect that doesn't actually work.

RealmSwift/Migration.swift
```diff
@@ -0,0 +1,233 @@
...
+property is added to an existing object.
+
+You should call this method before accessing any `Realm` instances which
+require migration. After registering your migration block, Realm will call your
+block automatically as needed.
+
+:warning: Unsuccessful migrations will throw exceptions when the migration block is applied.
+          This will happen in the following cases:
+
+          - The migration block was run and returns a schema version which is not higher
```
tgoyne: Migration blocks no longer return a version. Obj-c docs are probably out of date too.

RealmSwift/Migration.swift
```diff
@@ -0,0 +1,233 @@
...
+/**
+Get the schema version for a Realm at a given path.
+:param: realmPath     Path to a Realm file.
+:param: encryptionKey Optional 64-byte encryption key for encrypted Realms.
+:param: error         If an error occurs, upon return contains an `NSError` object
+                      that describes the problem. If you are not interested in
+                      possible errors, omit the argument, or pass in `nil`.
+:returns: The version of the Realm at `realmPath` or `nil` if the version cannot be read.
+*/
+public func schemaVersionAtPath(realmPath: String, encryptionKey: NSData? = nil, error: NSErrorPointer = nil) -> UInt? {
```
tgoyne: Would `Array<Byte>` be more pleasant to work with for any reason than `NSData`?

RealmSwift/Migration.swift
```diff
@@ -0,0 +1,233 @@
...
+of a `Realm` instance.
+
+This object provides access to the previous and current `Schema`s for this migration.
+*/
+public final class Migration {
+
+    // MARK: Properties
+
+    /// The migration's old `Schema`, describing the `Realm` before applying a migration.
+    public var oldSchema: Schema { return Schema(rlmMigration.oldSchema) }
```
tgoyne: Maybe make all these wrapper getters `@lazy` rather than returning a new object every time?

RealmSwift/Migration.swift
```diff
@@ -0,0 +1,233 @@
...
+
+    // MARK: Altering Objects During a Migration
+
+    /**
+    Enumerates objects of a given type in this Realm, providing both the old and new versions of
+    each object. Object properties can be accessed using subscripting.
+
+    :warning: All objects returned are of a type specific to the current migration and should not be
+              casted to the normal object type. Instead you should access them as `Object`s and use
+              subscripting to access properties.
```
tgoyne: This paragraph doesn't really apply to Swift migrations.

RealmSwift/Object.swift
```diff
@@ -0,0 +1,295 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+////////////////////////////////////////////////////////////////////////////
+
+import Realm
+import Realm.Private
+
+/**
+In Realm you define your model classes by subclassing RLMObject and adding properties to be persisted.
```
tgoyne: RLMObject

RealmSwift/Object.swift
```diff
@@ -0,0 +1,295 @@
...
+    }
+
+    /**
+    WARNING: This is an internal initializer not intended for public use.
+    */
+    public override init(object: AnyObject, schema: RLMSchema) {
+        super.init(object: object, schema: schema)
+    }
+
+    /// Get RLMArray values when getting array properties
```
tgoyne: I don't really know what this sentence means.

RealmSwift/Object.swift
```diff
@@ -0,0 +1,295 @@
...
+                return object_getIvar(self, prop.swiftListIvar) as RLMListBase?
+            }
+        }
+        return nil
+    }
+}
+
+// MARK: Equatable
+
+/// Returns whether both objects are equal.
```
tgoyne: should say what it means for objects to be equal

RealmSwift/Migration.swift
segiddins: well also if you're going through and adding things to a new object class or also if you're pulling values out of a deleted object class

RealmSwift/Object.swift
```diff
@@ -0,0 +1,295 @@
...
+                    return list
+
+                }
+            }
+            return RLMObjectBaseObjectForKeyedSubscript(self, key)
+        }
+        set(value) {
+            if let prop = RLMObjectBaseObjectSchema(self)[key] {
+                if prop.type == .Array {
+                    throwRealmException("Setting List properties is unsupported. Instead you can add or remove objects from the current List.")
```
tgoyne: Kinda weird to not support `List` here when the base-class version does support it.

RealmSwift/Migration.swift
segiddins: There is a `RawByte` type, but the Swift stdlib seems to prefer `[UInt8]` in those situations

RealmSwift/Realm.swift
```diff
@@ -0,0 +1,562 @@
...
+memory (see `init(inMemoryIdentifier:)`).
+
+Realm instances are cached internally, and constructing equivalent Realm
+objects (with the same path or identifier) multiple times on a single thread
+within a single iteration of the run loop will normally return the same
+Realm object. If you specifically want to ensure a Realm object is
+destroyed (for example, if you wish to open a realm, check some property, and
+then possibly delete the realm file and re-open it), place the code which uses
+the realm within an `autoreleasepool {}` and ensure you have no other
+strong references to it.
```
tgoyne: This paragraph isn't really true. They'll have the same backing `RLMRealm`, but you don't get the same `Realm` object every time.

RealmSwift/Realm.swift
```diff
@@ -0,0 +1,562 @@
...
+    the existing object is updated with any changed values.
+
+    When added, all linked (child) objects referenced by this object will also be
+    added to the Realm if they are not already in it. If the object or any linked
+    objects already belong to a different Realm an exception will be thrown. Use one
+    of the `create` functions to insert a copy of a persisted object into a different
+    Realm.
+
+    The object to be added must be valid and cannot have been previously deleted
+    from a Realm (i.e. `invalidated`) must be false.
```
tgoyne: close paren is in the wrong place

RealmSwift/Realm.swift
```diff
@@ -0,0 +1,562 @@
...
+
+    :param: object Object to be added to this Realm.
+    :param: update If true will try to update existing objects with the same primary key.
+    */
+    public func add(object: Object, update: Bool = false) {
+        var options : RLMCreationOptions = .allZeros
+        if update == true {
```
tgoyne: `== true` is dumb

Realm/RLMArray.h
segiddins: :+1: I believe it does

RealmSwift/Realm.swift
```diff
@@ -0,0 +1,562 @@
...
+    public func objects<T: Object>(type: T.Type) -> Results<T> {
+        return Results<T>(RLMGetObjects(rlmRealm, T.className(), nil))
+    }
+
+    /**
+    Get an object with the given primary key.
+
+    Returns `nil` if no object exists with the given primary key.
+
+    This method requires that `primaryKey()` be overridden on the receiving subclass.
```
tgoyne: given subclass?

RealmSwift/Results.swift
```diff
@@ -0,0 +1,257 @@
...
+
+    :returns: The index of the given object, or `nil` if the object is not in the results.
+    */
+    public func indexOf(object: T) -> Int? {
+        return notFoundToNil(rlmResults.indexOfObject(unsafeBitCast(object, RLMObject.self)))
+    }
+
+    /**
+    Returns the index of the first object matching the given predicate,
+    or `nil` if the object is not in the results.
```
tgoyne: same comments as for the List version of these apply

RealmSwift/Results.swift
```diff
@@ -0,0 +1,257 @@
...
+/**
+`Results` is an auto-updating container type in Realm returned from object
+queries.
+
+Results can be queried with the same predicates as `List<T>` and you can chain queries to further
+filter query results.
+
+Results cannot be created directly.
+*/
+public final class Results<T: Object>: Printable {
```
tgoyne: Should probably have a protocol that `Results` and `List` implement like `RLMCollection`.

segiddins: :+1: we punted on this seeing as they both already conform to `CollectionType` and there is some wonkeyness with trying to say that the `Generator.Type: Object`

RealmSwift/Tests/MigrationTests.swift
```diff
@@ -0,0 +1,296 @@
...
+    return realmWithSingleClass(path, objectSchema)
+}
+
+class MigrationTests: TestCase {
+
+    // MARK Utility methods
+
+    // create realm at path and test version is 0
+    private func createAndTestRealmAtPath(realmPath: String) {
+        autoreleasepool { () -> () in
```
tgoyne: remove everything after { on this line

RealmSwift/Tests/MigrationTests.swift
```diff
@@ -0,0 +1,296 @@
...
+        XCTAssertEqual(Realm().objects(SwiftStringObject.self).count, 1)
+    }
+
+    // test getting/setting all property types
+    func testMigrationObject() {
+        autoreleasepool {
+            Realm().write {
+                var object = SwiftObject()
+                object.boolCol = true
+                object.objectCol = SwiftBoolObject(value:[true])
```
tgoyne: space after colon

Realm/RLMListBase.h
```diff
@@ -22,7 +22,7 @@
 
 // A base class for Swift generic Lists to make it possible to interact with
 // them from obj-c
-@interface RLMListBase : NSObject
+@interface RLMListBase : NSObject<NSFastEnumeration>
```
segiddins: space before the `<` ?

RealmSwift/Tests/ObjectAccessorTests.swift
```diff
@@ -0,0 +1,77 @@
...
+
+        object.floatCol = 20
+        XCTAssertEqual(object.floatCol, 20 as Float)
+        object.floatCol = 20.2
+        XCTAssertEqual(object.floatCol, 20.2 as Float)
+
+        object.doubleCol = 20
+        XCTAssertEqual(object.doubleCol, 20)
+        object.doubleCol = 20.2
+        XCTAssertEqual(object.doubleCol, 20.2)
```
tgoyne: A value that can be exactly represented as a double but not a float would be a nice thing to test on float and double properties.

RealmSwift/Tests/RealmTests.swift
```diff
@@ -0,0 +1,475 @@
...
+
+    func testCancelWrite() {
+        assertThrows(Realm().cancelWrite())
+        Realm().beginWrite()
+        Realm().create(SwiftStringObject.self, value:["1"])
+        Realm().cancelWrite()
+        XCTAssertEqual(Realm().objects(SwiftStringObject).count, 0)
+
+        Realm().write {
+            self.assertThrows(self.realmWithTestPath().cancelWrite())
```
tgoyne: should insert an object and check for zero afterwards again to verify it actually cancels the block write and doesn't just fail to throw

RealmSwift/Tests/ObjectAccessorTests.swift
```diff
@@ -0,0 +1,77 @@
...
+// limitations under the License.
+//
+////////////////////////////////////////////////////////////////////////////
+
+import XCTest
+import RealmSwift
+import Foundation
+
+class ObjectAccessorTests: TestCase {
+    func setAndTestAllProperties(object: SwiftObject) {
```
tgoyne: sized int properties should also be tested

build.sh
```diff
@@ -34,7 +34,6 @@ command:
   download-core:               downloads core library (binary version)
   clean:                       clean up/remove all generated files
   build [settings]:            builds iOS and OS X frameworks
```
tgoyne: remove [settings] from all the help text

build.sh
```diff
@@ -156,18 +160,20 @@ test_ios_devices() {
...
-    exit 0
+    return $failed
 }
 
 ######################################
 # Input Validation
 ######################################
 
-if [ "$#" -eq 0 -o "$#" -gt 2 ]; then
+if [ "$#" -eq 0 -o "$#" -gt 3 ]; then
```
tgoyne: missed this when removing BUILD_SETTINGS

Realm/Tests/RealmSwift/ListTests.swift
```diff
@@ -0,0 +1,352 @@
...
+}
+
+class ListStandaloneTests: ListTests {
+    override func createArray() -> SwiftArrayPropertyObject {
+        let array = SwiftArrayPropertyObject()
+        XCTAssertNil(array.realm)
+        return array
+    }
+
+    // Things not implemented in standalone
```
tgoyne: These should now check for the method throwing.

RealmSwift/Property.swift
```diff
@@ -0,0 +1,61 @@
...
+
+/**
+This class represents properties persisted to Realm in an ObjectSchema.
+
+When using Realm, Property objects allow performing migrations and
+introspecting the database's schema.
+
+These properties map to columns in the core database.
+*/
+public final class Property {
```
tgoyne: Should add `Printable` and a description property since the backing things now have them.

Realm.xcodeproj/xcshareddata/xcschemes/iOS Dynamic.xcscheme
alazier: These old tests should actually be deleted.

alazier: Actually these are fine now - though these were swift tests...

Realm/RLMArray.h
alazier: You are both wrong! This is no longer needed as we now import Realm.Private in RLMSupport.swift.

Realm.xcodeproj/xcshareddata/xcschemes/OSX.xcscheme
alazier: Not sure - what does this setting even do? Does it even matter?

tgoyne: It's used for the thing where Xcode nags you to update your project settings to avoid doing it when you've explicitly switched away from the new default. Should be no need to ever care about it.

Realm/RLMArrayLinkView.mm
alazier: I believe you wrote this @segiddins ...

Realm/RLMArrayLinkView.mm
alazier: Good catch

Realm.xcodeproj/xcshareddata/xcschemes/iOS Dynamic.xcscheme
alazier: I take it all back - these test only work for the 'iOS Tests' target. They will not work for the 'iOS Dynamic' target as they depend on the target name.

Realm/RLMObjectBase.mm
alazier: More importantly, you should note that `valueForKey:` does not work for dynamic objects.

Realm/RLMUtil.mm
alazier: the standalone initializer actually needs the full schema to validate nested objects.

Realm/Tests/PropertyTests.m
alazier: This whole directory is a merge error.

Realm/RLMObjectBase.mm
alazier: I gave the same feedback on the original pr so I agree.

Also interspersing c code in an objc implementation is pretty annoying. Will move these out.

Realm/RLMObjectBase.mm
alazier: I'll change the code, but keep the comment.

alazier: Also note they are not exactly the same - we are returning NO only if one is nil and the other is not. Your suggested code only works because we previously checked if the two objects are equal, which is true if they are both nil. In the case that these lines were reordered your code would be be incorrect.

RealmSwift/Migration.swift
alazier: I think @segiddins is correct. Will need to add tests for these scenarios

tgoyne: Yeah, after looking at the obj-c code there's some code paths that explicitly pass nil.

Realm/RLMObjectStore.mm
alazier: I think sorting alphabetically is silly. Should instead put the privates together and group other similar apis together.

RealmSwift/Object.swift
alazier: We got rid of a different one, the one which takes an objectSchema.

Realm/RLMObjectStore.mm
tgoyne: I don't really care what the ordering is as long as it's consistent between files, and currently they're all sorted alphabetically (except for the header for the current implementation file, which is always first).

RealmSwift/Object.swift
alazier: This was discussed in the previous pr. This will be of type object and it is a coding error in the binding which should crash otherwise.

RealmSwift/List.swift
alazier: we should throw if the object is not in the list

RealmSwift/Migration.swift
alazier: This is something we can revisit later.

Realm/Swift/RLMSupport.swift
```diff
@@ -17,6 +17,7 @@
 ////////////////////////////////////////////////////////////////////////////
 
 import Realm
+import Realm.Private
```
segiddins: This means we'll be exposing everything, such as the dynamic API and the `RLMObjectBase` functions, to everyone who uses this -- not sure I love that.

Realm/RLMArrayLinkView.mm
segiddins: doesnt mean it couldn't be better!

Realm/RLMListBase.m
segiddins: not necessary -- we implement `description` on `RLMListBase` in Swift.

RealmSwift/Migration.swift
segiddins: not possible to make computed properties lazy

Realm/RLMArrayLinkView.mm
```diff
@@ -194,15 +200,11 @@ - (void)replaceObjectAtIndex:(NSUInteger)index withObject:(RLMObject *)object {
 - (NSUInteger)indexOfObject:(RLMObject *)object {
     // check attached for table and object
     RLMLinkViewArrayValidateAttached(self);
+
     if (object->_realm && !object->_row.is_attached()) {
         @throw RLMException(@"RLMObject is no longer valid");
     }
 
-    // check that object types align
```
jpsim: why was this removed? it should be a programmer error if you're passing in objects that can't be contained by the RLMArray.

alazier: The next line checks the same thing more efficiently.

RealmSwift/List.swift
```diff
@@ -0,0 +1,298 @@
...
+    :param: index  The list index of the object to be replaced.
+    :param: object An object to replace at the specified index.
+    */
+    public func replace(index: Int, object: T) {
+        throwForNegativeIndex(index)
+        _rlmArray.replaceObjectAtIndex(UInt(index), withObject: unsafeBitCast(object, RLMObject.self))
+    }
+}
+
+extension List : ExtensibleCollectionType {
```
jpsim: we've kept the colon next to the type name for all our other declarations and should also do it here: `extension List : ExtensibleCollectionType`.

RealmSwift/List.swift
```diff
@@ -0,0 +1,298 @@
...
+
+    :returns: `Results` with elements sorted by the given sort descriptors.
+    */
+    public func sorted<S: SequenceType where S.Generator.Element == SortDescriptor>(sortDescriptors: S) -> Results<T> {
+        return Results<T>(_rlmArray.sortedResultsUsingDescriptors(map(sortDescriptors) { $0.rlmSortDescriptorValue }))
+    }
+
+    // MARK: Sequence Support
+
+    /// Returns a `GeneratorOf<T>` that yields successive elements in the list.
```
jpsim: Move this with the rest of the `ExtensibleCollectionType` support functions.

RealmSwift/Migration.swift
```diff
@@ -0,0 +1,220 @@
...
+
+:param: path          The path of the Realm to migrate.
+:param: encryptionKey Optional 64-byte encryption key for encrypted Realms.
+                      If the Realms at the given path are not encrypted, omit the argument or pass
+                      in `nil`.
+
+:returns: `nil` if the migration was successful, or an `NSError` object that describes the problem
+          that occured otherwise.
+*/
+public func migrateRealm(path: String, encryptionKey: NSData? = nil) -> NSError? {
```
jpsim: how about using the default realm path as a default parameter value? so users using the default realm can just call `migrationRealm()`

RealmSwift/List.swift
segiddins: but `SortDescriptor` is a realm thing, not an `ExtensibleCollectionType` thing

RealmSwift/Object.swift
```diff
@@ -0,0 +1,296 @@
...
+    }
+
+    /**
+    WARNING: This is an internal initializer not intended for public use.
+    */
+    public override init(object: AnyObject, schema: RLMSchema) {
+        super.init(object: object, schema: schema)
+    }
+
+    public override func valueForKey(key: String) -> AnyObject? {
```
jpsim: Add docs:

> /// WARNING: This is an internal function not intended for public use.

RealmSwift/Object.swift
```diff
@@ -0,0 +1,296 @@
...
+        if let list = listProperty(key) {
+            return list
+        }
+        return super.valueForKey(key)
+    }
+
+    public override func setValue(value: AnyObject?, forKey key: String) {
```
jpsim: Add docs:

> /// WARNING: This is an internal function not intended for public use.

RealmSwift/Object.swift
```diff
@@ -0,0 +1,296 @@
...
+                    throwRealmException("Setting List properties is unsupported. Instead you can add or remove objects from the current List.")
+                }
+            }
+            RLMObjectBaseSetObjectForKeyedSubscript(self, key, value)
+        }
+    }
+}
+
+
+// Internal class. Do not use directly.
```
jpsim: Use three slashes for this text to show up in docs.

RealmSwift/Realm.swift
```diff
@@ -0,0 +1,562 @@
...
+    Realm.
+
+    The object to be added must be valid and cannot have been previously deleted
+    from a Realm (i.e. `invalidated`) must be false.
+
+    :param: object Object to be added to this Realm.
+    :param: update If true will try to update existing objects with the same primary key.
+    */
+    public func add(object: Object, update: Bool = false) {
+        var options : RLMCreationOptions = .allZeros
```
jpsim:
```swift
if update && object.objectSchema.primaryKeyProperty == nil {
  throwRealmException("'\(object.objectSchema.className)' does not have a primary key and can not be updated")
}
RLMAddObjectToRealm(object, rlmRealm, update ? .UpdateOrCreate : .allZeros)
```

RealmSwift/Realm.swift
```diff
@@ -0,0 +1,562 @@
...
+                    thrown if any required properties are not present and no default is set.
+
+                    When passing in an `Array`, all properties must be present,
+                    valid and in the same order as the properties defined in the model.
+    :param: update  If true will try to update existing objects with the same primary key.
+
+    :returns: The created object.
+    */
+    public func create<T: Object>(type: T.Type, value: AnyObject = [:], update: Bool = false) -> T {
+        var options : RLMCreationOptions = .allZeros
```
jpsim: see `add(_:)` implementation recommendations.

RealmSwift/Realm.swift
```diff
@@ -0,0 +1,562 @@
...
+    :returns:             If an error occurs, returns an `NSError` object, otherwise `nil`.
+
+    :returns: Whether the realm was copied successfully.
+    */
+    public func writeCopyToPath(path: String, encryptionKey: NSData? = nil) -> NSError? {
+        var error: NSError?
+        if let encryptionKey = encryptionKey {
+            rlmRealm.writeCopyToPath(path, encryptionKey: encryptionKey, error: &error)
+        }
+        else {
```
jpsim: else is on same line as previous closing brace in the rest of our Swift code

RealmSwift/Results.swift
```diff
@@ -0,0 +1,257 @@
...
+    :returns: The average of the given property over all objects in the Results.
+    */
+    public func average<U: AddableType>(property: String) -> U {
+        return rlmResults.averageOfProperty(property) as AnyObject as U
+    }
+
+    // MARK: Sequence Support
+
+    /// Returns a `GeneratorOf<T>` that yields successive elements in the results.
+    public func generate() -> GeneratorOf<T> {
```
jpsim: move to `CollectionType` extension

jpsim:
I've reviewed this completely now, but I'd like to do another once over after you've addressed everyone's feedback in here.

RealmSwift/Object.swift
segiddins: disagree- KVC is a public API!

Realm.xcodeproj/xcshareddata/xcschemes/iOS Dynamic.xcscheme
segiddins: :+1: this is all good

RealmSwift/Tests/ObjectAccessorTests.swift
segiddins: already fixed

RealmSwift/Tests/RealmTests.swift
segiddins: did this

RealmSwift/Tests/ObjectAccessorTests.swift
segiddins: done

RealmSwift/Object.swift
alazier: We don't want this method/class to show up in the docs at all.

jpsim: As it stands now, since it's a public declaration, it'll show up in the jazzy-generated docs with no indication whatsoever that it's a class for internal use only. There's an open issue on jazzy to explicitly ignore certain tokens, but that's not done yet: https://github.com/realm/jazzy/issues/67.

Even jazzy aside, this class will show up in Xcode's autocomplete and without a documentation comment, there won't be any indication that it's a private class.

Realm/RLMArrayLinkView.mm
jpsim: We need to re-add the object check, and add a test for calling this method with an RLMObject from a different realm.

Issue #1595, opened 2015-03-11
Support Carthage prebuilt framework


l4u:
Please add `Realm-cocoa.framework.zip` in releases, so that Carthage can pull from prebuilt binaries instead of building it from source.

jpsim:
Thanks for the feature request, @l4u. We'll update this issue once we've updated our release scripts to do this.

timanglade:
For our first swift release, we will manually upload a prebuilt framework to our Github releases.

jpsim:
If you specify `github "realm/realm-cocoa" ~> 0.92` in your Cartfile, it'll download the prebuilt Realm Objective-C and Swift frameworks from our [v0.92.0 release](https://github.com/realm/realm-cocoa/releases/tag/v0.92.0).

PR #1645, opened 2015-03-16
Update packaging scripts for new Swift APIs

### LINKED ISSUE #1602: Fix release packaging
Things have broken with our build.sh/ci changes for swift. Some Things that are known to be broken:
- The dynamic framework is names differently
- We need to bring back the old swift examples (#1604)

jpsim:
Closes #1602.

- removed old Swift examples
- added source code to release package for building dynamic & swift frameworks
- added `ios-dynamic-fat` and `ios-swift-fat` targets to build.sh
- update build.sh usage

/cc @tgoyne @alazier @segiddins

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -19,6 +19,10 @@ set -e
 # You can override the xcmode used
 : ${XCMODE:=xcodebuild} # must be one of: xcodebuild (default), xcpretty, xctool
 
+# Whether or not to package RealmSwift
+# TODO: Remove all references when RealmSwift is released
+: ${PACKAGE_REALM_SWIFT:=false}
+
 PATH=/usr/local/bin:/usr/bin:/bin:/usr/libexec:$PATH
 
 if ! [ -z "${JENKINS_HOME}" ]; then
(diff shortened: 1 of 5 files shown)


tgoyne:
Has Apple stopped rejecting things that use fat dynamic frameworks?

Still need to include RLMSupport.swift in the release package.

Doesn't appear to be anything that verifies that the packaged source is actually usable.

build.sh
```diff
@@ -371,7 +391,6 @@ case "$COMMAND" in
         sh build.sh verify-ios-swift
         sh build.sh verify-ios-swift-debug
         sh build.sh verify-ios-device
-        sh build.sh verify-docs
```
segiddins: why is this removed?

build.sh
```diff
@@ -28,27 +28,39 @@ fi
 
 usage() {
 cat <<EOF
-Usage: sh $0 command [argument]
+Usage: sh $0 command [argument] [settings]
```
tgoyne: [settings] was removed

segiddins:
I thought we wanted to keep the ObjC Swift examples?

tgoyne:
We did.

build.sh
jpsim: it's a duplicate of the first item in verify

jpsim:
> I thought we wanted to keep the ObjC Swift examples?

Last we spoke about these, we wanted to keep the old Swift tests, but not the old examples.

tgoyne:
We need to keep the old samples until we fully release the new one.

jpsim:
> Has Apple stopped rejecting things that use fat dynamic frameworks?

No, they still reject them.

> Still need to include RLMSupport.swift in the release package.

I'll add that back in.

> Doesn't appear to be anything that verifies that the packaged source is actually usable.

We verify that the new Swift examples build from the release package, which builds the iOS dynamic and RealmSwift-iOS frameworks from the packaged source. We could also run `verify`.

jpsim:
After speaking with @alazier and @tgoyne, I'll be updating this so that we ship 2 prebuilt dynamic iOS frameworks (release for device & simulator architectures). This will avoid having to ship `build.sh`, `Realm.xcodeproj` and the Objective-C source, just to build the Swift API.

jpsim:
I've updated the packaging scripts to package RealmSwift and the swift-next examples. The scripts do as much of the new process as possible, short of including the files in the zip, to minimize bit rot.

Here's the zip structure when the new packaging is enabled:

```
realm-cocoa-0.91.1
├── LICENSE.txt
├── browser
├── docs.webloc
├── examples
│   ├── README.md
│   ├── ios
│   │   ├── objc
│   │   ├── rubymotion
│   │   ├── swift
│   │   └── swift-next
│   └── osx
│       └── objc
├── ios
│   └── Realm.framework
├── osx
│   └── Realm.framework
├── plugin
└── swift
    ├── RLMSupport.swift
    ├── RealmSwift
    ├── RealmSwift.xcodeproj
    └── dynamic_frameworks
        ├── Realm.framework
        ├── iphoneos
        │   └── Realm.framework
        └── iphonesimulator
            └── Realm.framework
```

Please review /cc @tgoyne @alazier @segiddins.

tgoyne:
Is swift/dynamic_frameworks/Realm.framework the OS X framework? It'd be a bit more consistent if it was in an osx subdir then.

jpsim:
> Is swift/dynamic_frameworks/Realm.framework the OS X framework?

It's the fat framework. We could place it in a `fat` directory.

tgoyne:
Ah, makes sense why you'd put it in the root then. Not sure whether or not a `fat` dir would be better.

jpsim:
I've completely redone this to be much simpler:

1. no longer modifying RealmSwift.xcodeproj
2. no longer including the fat dynamic iOS framework

build.sh
```diff
@@ -31,28 +35,39 @@ cat <<EOF
...
-  test-all:         	tests iOS and OS X frameworks with debug and release configurations
-  examples:         	builds all examples in examples/
-  browser:         	builds the Realm Browser OSX app
-  test-browser:     	tests the Realm Browser OSX app
-  verify:           	cleans, removes docs/output/, then runs docs, test-all, examples & browser
+  download-core:        downloads core library (binary version)
+  build:                builds all iOS  and OS X frameworks
+  ios-static:           builds fat iOS static framework
+  ios-dynamic:          builds iOS dynamic frameworks
+  ios-dynamic-fat:      builds fat iOS dynamic framework
```
alazier: I thought we were no longer doing ios fat

build.sh
```diff
@@ -31,28 +35,39 @@ cat <<EOF
...
+  test-osx:             tests OS X framework
+  test-osx-swift:       tests RealmSwift OS X framework
+  verify:               verifies docs, osx, osx-swift, ios-static, ios-dynamic, ios-swift, ios-device in both Debug and Release configurations
   docs:                 builds docs in docs/output
+  examples:             builds all examples
+  examples-ios:         builds all static iOS examples
+  examples-ios-swift:   builds all Swift iOS examples
+  examples-osx:         builds all OS X examples
+  browser:              builds the Realm Browser
+  test-browser:         tests the Realm Browser
```
alazier: move this next to the rest of test targets?

build.sh
jpsim: we're not *packaging* it, but it is useful to be able to build a fat version of it.

build.sh
jpsim: we have a "browser" section in build.sh.

build.sh
alazier: If we are not using it please don't add it. This is currently not in use and there is already too much in build.sh to have dead code hanging around.

alazier: It also seems dangerous to build a library that apple will reject when used in apps.

alazier:
If the old packaging still works lets please just merge this. That was really all we needed to get done here, and do not yet need packing of the new apis.

There is still more to do for the new apis such as package verification and some other stuff but we can dot his later.

build.sh
jpsim: Sure, I'll remove it. If we choose to do #1595, we'll need to re-add it, but we should have that conversation first.

jpsim:
I've diffed the release zip from master with the one produced by this and everything matches, so I'm merging this in.

PR #1649, opened 2015-03-17
create separate module for private dependencies used by RLMSupport.swift

alazier:
@segiddins

### DIFF
--- a/Realm/Swift/RLMSupport.swift
+++ b/Realm/Swift/RLMSupport.swift
@@ -17,7 +17,7 @@
 ////////////////////////////////////////////////////////////////////////////
 
 import Realm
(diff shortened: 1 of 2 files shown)


segiddins:
Let's rename `RLMArray_Private.h` and add documentation to the one method in it?

alazier:
I think this is fine the way it is for now. Not worth further effort at this point.

segiddins:
With this change, you'll have to `import Realm.SwiftSupport` everywhere you want to use `RLMArray(objectClassName:)`

alazier:
That shouldn't be the case - at least it wasn't when just using the Realm.Private module.

PR #1651, opened 2015-03-17
[Realm] Test that when cancelling inside a write block, the count of obj...

segiddins:
...ects doesn't change

\c @tgoyne @alazier

### DIFF
--- a/RealmSwift/Tests/RealmTests.swift
+++ b/RealmSwift/Tests/RealmTests.swift
@@ -152,8 +152,12 @@ class RealmTests: TestCase {
 
         Realm().write {
             self.assertThrows(self.realmWithTestPath().cancelWrite())
+            let object = Realm().create(SwiftStringObject)
             Realm().cancelWrite()
+            XCTAssertTrue(object.invalidated)
+            XCTAssertEqual(Realm().objects(SwiftStringObject).count, 0)
         }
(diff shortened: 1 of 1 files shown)


alazier:
Could also hold onto `object` and make sure it is invalidated.

segiddins:
You mean outside of the `write` block?

jpsim:
> Could also hold onto object and make sure it is invalidated.

There's no point in doing that both inside and outside the block since it's the same transaction.

The ObjC rollback tests are much more comprehensive: https://github.com/realm/realm-cocoa/blob/master/Realm/Tests/RealmTests.mm#L778-L931

Since those tests only use the public API, we'd gain a lot more useful test coverage by porting those here rather than starting over in an ad-hoc fashion.

segiddins:
> There's no point in doing that both inside and outside the block since it's the same transaction.

But that's an implementation detail :)

tgoyne:
One of the reasons why the Swift tests are more manageable and understandable is because they're more limited in scope. Only the Swift-specific things should be tested, which in this case means verifying that the transaction was cancelled, but not that transaction canceling actually works correctly in general, since that's already covered by the obj-c tests.

tgoyne:
:+1: to the current test.

PR #1652, opened 2015-03-17
[DynamicObject] Properly support subscripting of list properties

segiddins:
\c @alazier @tgoyne @jpsim

### DIFF
--- a/RealmSwift/Object.swift
+++ b/RealmSwift/Object.swift
@@ -219,33 +219,23 @@ public func == <T: Object>(lhs: T, rhs: T) -> Bool {
 
 /// Object interface which allows untyped getters and setters for Objects.
 public final class DynamicObject : Object {
-    private var listProperties = [String: List<DynamicObject>]()
+    private var listProperties = [String: Optional<List<DynamicObject>>]()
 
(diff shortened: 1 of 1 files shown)


RealmSwift/Object.swift
```diff
@@ -219,33 +219,23 @@ public func == <T: Object>(lhs: T, rhs: T) -> Bool {
...
-                    let list = List<DynamicObject>()
-                    listProperties[key] = list
-                    return list
-
-                }
-            }
-            return RLMObjectBaseObjectForKeyedSubscript(self, key)
+    private override func listProperty(key: String) -> RLMListBase? {
+        // get it or set it
+        if let list = listProperties[key] {
```
jpsim: We want to return if `listProperties` contains the key, not just if the result of the subscript isn't nil. As it is now, this optional binding will fail even when we've already determined that the given key isn't a List.

```swift
if contains(listProperties.keys, key) {
  return listProperties[key]
}
```

jpsim:
:+1:

segiddins:
Closing in favor of https://github.com/realm/realm-cocoa/pull/1674.

PR #1660, opened 2015-03-18
[RealmExamples] remove reference cycle in examples

### LINKED ISSUE #1658: Reference cycle in one of the examples
Hi! 

I think [one of the examples](https://github.com/realm/realm-cocoa/blob/master/examples/ios/swift/TableView/TableViewController.swift#L40) has a reference cycle.

    class TableViewController: UITableViewController {

        var notificationToken: RLMNotificationToken?

        override func viewDidLoad() {
            // Set realm notification block
            notificationToken = RLMRealm.defaultRealm().addNotificationBlock { note, realm in
                self.tableView.reloadData()
            }
            ...
        }
    }

The block above captures a strong reference to `self`, which is then stored in RLMNotificationToken:

    @interface RLMNotificationToken ()
    @property (nonatomic, copy) RLMNotificationBlock block;
    @end

And the TableViewController stores RLMNotificationToken in the `notificationToken` property.

So that creates a reference cycle: 
`self.notificationToken —>  RLMNotificationToken.block —> self`

If I've missed something, please let me know. I'd love to understand how it works :) Thanks!

jpsim:
Fixes #1658. /cc @alazier

### DIFF
--- a/examples/ios/swift/GroupedTableView/TableViewController.swift
+++ b/examples/ios/swift/GroupedTableView/TableViewController.swift
@@ -48,7 +48,7 @@ class TableViewController: UITableViewController {
         setupUI()
 
         // Set realm notification block
-        notificationToken = RLMRealm.defaultRealm().addNotificationBlock { note, realm in
+        notificationToken = RLMRealm.defaultRealm().addNotificationBlock { [unowned self] note, realm in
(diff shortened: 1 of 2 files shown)


segiddins:
:+1:

PR #1668, opened 2015-03-18
Re-add check for objects of wrong type in indexOfObject

alazier:
@segiddins @jpsim


jpsim:
:+1:

Realm/RLMArrayLinkView.mm
```diff
@@ -205,6 +205,11 @@ - (NSUInteger)indexOfObject:(RLMObject *)object {
         @throw RLMException(@"RLMObject is no longer valid");
     }
 
+    // check that object types align
+    if (![_objectClassName isEqualToString:object.objectSchema.className]) {
+        @throw RLMException(@"Object type does not match RLMArray");
```
segiddins: `[NSString stringWithFormat:@"Object of type (%@) does not match RLMArray type (%@)", object.objectSchema.className, _objectClassName]`

PR #1672, opened 2015-03-19
Remove the Swift- prefix from the new test objects and files

tgoyne:
Might happen to prevent the transient test failures due to not having multiple files named the same thing in different places, and makes the tests read better by eliminating some noise.


alazier:
The failures in the other pr were due to a code error which has now been fixed.

PR #1674, opened 2015-03-19
Test subscripting

### LINKED ISSUE #1669: Fix Object/DynamicObject subscripting and valueForKey/setValueForKey
Should call into RLMDynamicSet/Get directly where applicable. Requires full support for Lists and full test coverage.

alazier:
@segiddins @tgoyne 

Closes #1669 

Anything else we should be testing?


segiddins:
This looks pretty good to me, but we should also test setting a list property with an array

PR #1675, opened 2015-03-19
[RealmSwift] hide unfortunately public declarations that should be private

jpsim:
Uses jazzy's latest `nodoc` feature: https://github.com/realm/jazzy/pull/180. /cc @segiddins

### DIFF
--- a/RealmSwift/Object.swift
+++ b/RealmSwift/Object.swift
@@ -150,6 +150,7 @@ public class Object: RLMObjectBase, Equatable, Printable {
 
     /**
     WARNING: This is an internal initializer not intended for public use.
+    :nodoc:
     */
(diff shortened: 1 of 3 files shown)


RealmSwift/Object.swift
```diff
@@ -150,6 +150,7 @@ public class Object: RLMObjectBase, Equatable, Printable {
 
     /**
     WARNING: This is an internal initializer not intended for public use.
+    :nodoc:
```
segiddins: this style parses properly?

segiddins:
:+1:

RealmSwift/Object.swift
jpsim: Yes.

PR #1676, opened 2015-03-20
Allow Realm to be used within an app extension

pizthewiz:
Currently if one links to the Realm framework from an app extension, the compiler gives warning: _linking against dylib not safe for use in application extensions_

The [App Extension Programming Guide > Using an Embedded Framework to Share Code](https://developer.apple.com/library/prerelease/ios/documentation/General/Conceptual/ExtensibilityPG/ExtensionScenarios.html#//apple_ref/doc/uid/TP40014214-CH21-SW5) doc mentions that the framework can be built with the `APPLICATION_EXTENSION_API_ONLY ` project setting enabled to avoid this (though it is unclear if the current config would result in App Store rejection). The 3 framework and paired test targets build successfully when the project setting is enabled, so it would appear as though the framework uses any verboten API.

### DIFF
--- a/Realm.xcodeproj/project.pbxproj
+++ b/Realm.xcodeproj/project.pbxproj
@@ -1535,6 +1535,7 @@
 		29E3C7381A71C1C700B62C1D /* Debug */ = {
 			isa = XCBuildConfiguration;
 			buildSettings = {
+				APPLICATION_EXTENSION_API_ONLY = YES;
 				CODE_SIGN_IDENTITY = "iPhone Developer";
 				"CODE_SIGN_IDENTITY[sdk=iphoneos*]" = "iPhone Developer";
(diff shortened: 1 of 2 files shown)


segiddins:
We also need to do the same for RealmSwift.framework

pizthewiz:
70693f8 makes the change to the RealmSwift project, both targets compile and tests run successfully. 

What is the RealmSwift project? Is it to replace RLMSupport.swift?

segiddins:
This looks good to me, thanks @pizthewiz ! @jpsim, could you take a look?

jpsim:
Thanks for doing this, @pizthewiz!

alazier:
@pizthewiz - the RealmSwift project contains our new swift apis. There are a few kinks to work out so they are not yet officially supported but you are welcome to try them out and feedback is welcome.

pizthewiz:
Thanks @alazier, sounds as though it is worth a try!

PR #1680, opened 2015-03-21
Invoke test methods correctly

tgoyne:
`-[XCTest invokeTest]` does `[self setUp]; [self.invocation invoke]; [self tearDown]` so calling `[super invokeTest]` results in the `setUp` and `tearDown` methods being called twice. All of our current obj-c setup and teardown methods are idempotent so this happens to not break anything at the moment, and the new Swift tests already use `self.invocation.invoke()`.

@alazier

### DIFF
--- a/Realm/Tests/RLMTestCase.m
+++ b/Realm/Tests/RLMTestCase.m
@@ -112,7 +112,7 @@ static BOOL encryptTests() {
 - (void)invokeTest
 {
     @autoreleasepool { [self setUp]; }
-    @autoreleasepool { [super invokeTest]; }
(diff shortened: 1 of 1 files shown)


alazier:
:+1:

segiddins:
I'm not sure this change plays nicely with our performance tests.

tgoyne:
Added a hack to make the performance tests work.

segiddins:
:+1:

segiddins:
@alazier I don't think your commit does the same thing as @tgoyne's, because now the invokation is not wrapped in its own @autoreleasepool

alazier:
I know it doesn't and it doesn't need to be.

PR #1686, opened 2015-03-24
[RealmSwift] Port PerformanceTests from Obj-C

### LINKED ISSUE #1454: Implement performance tests for RealmSwift
Can likely just be a copy of the ObjC performance tests

segiddins:
Closes #1454.

\c @alazier @tgoyne


tgoyne:
Needs separate baselines for the Swift tests.

segiddins:
I'm looking into the hanging test case right now

segiddins:
Added baselines, please double-check them

tgoyne:
The RealmSwift baseline file shouldn't also have the baselines for the obj-c tests.

tgoyne:
`testEnumerateAndMutateAll` baseline of 0 seconds is suspicious.

RealmSwift/Tests/PerformanceTests.swift
```diff
@@ -0,0 +1,444 @@
...
+            for i in 0..<results.count {
+                let string = results[i].stringCol
+            }
+        }
+    }
+
+    func testEnumerateAndAccessArrayProperty() {
+        let realm = getStringObjects(5)
+        realm.beginWrite()
+        let arrayPropertyObject = realm.create(SwiftArrayPropertyObject.self, value: ["name", map(realm.objects(SwiftStringObject)) { $0 }, []])
```
jpsim: why is `map(realm.objects(SwiftStringObject)) { $0 }` necessary?

segiddins: because `Results` doesn't conform to `NSFastEnumeration`

segiddins:
@jpsim @tgoyne this is green, please take another look?

RealmSwift/Tests/TestCase.swift
```diff
@@ -33,6 +33,7 @@ class TestCase: XCTestCase {
         Realm.defaultPath = realmPathForFile("\(realmFilePrefix()).default.realm")
         NSFileManager.defaultManager().createDirectoryAtPath(realmPathForFile(""), withIntermediateDirectories: true, attributes: nil, error: nil)
 
+        internalImplementation().setNumberOfTestIterations(1)
```
tgoyne: This should instead do what we ended up doing for obj-c on master (just `autoreleasepool { super.invokeTests()` here and wrap the bodies of each `setUp` and `tearDown` method with  `autoreleasepool {}`).

segiddins: That actually isn't possible in Swift 1.1, since you can't ever call through to `super` inside a closure (which has been fixed in 1.2)

tgoyne: The call to super doesn't need to be in the autorelease pool.

segiddins: I can't get it to work...

segiddins:
The only failure is due to the standard deviation in ObjC `PerformanceTests.testCrossThreadSyncLatency`. Should I just remove the baseline for that benchmark?

tgoyne:
The standard deviation for that test when I run it locally on a device is <5%, so something funny is going on.

tgoyne:
Oh, it's only <5% if I run only that test, and >20% if I run all of them. I suspect state is leaking.

segiddins:
This is green now!

RealmSwift/Tests/PerformanceTests.swift
```diff
@@ -0,0 +1,462 @@
...
+            for stringObject in arrayPropertyObject.array {
+                let string = stringObject.stringCol
+            }
+        }
+    }
+
+    func testEnumerateAndAccessArrayPropertySlow() {
+        let realm = getStringObjects(5)
+        realm.beginWrite()
+    let arrayPropertyObject = realm.create(SwiftArrayPropertyObject.self, value: ["name", map(realm.objects(SwiftStringObject)) { $0 }, []])
```
jpsim: indentation

jpsim:
:+1:

segiddins:
I can't reproduce the failure that CI is reporting, which is a leaked default realm.

RealmSwift/Tests/ObjectSchemaTests.swift
```diff
@@ -24,7 +24,9 @@ class ObjectSchemaTests: TestCase {
 
     override func setUp() {
         super.setUp()
-        objectSchema = Realm().schema["SwiftObject"]
+        autoreleasepool {
+            self.objectSchema = Realm().schema["SwiftObject"]
```
alazier: Do we even need this as a member? As said in the past this pattern has caused issues and should be avoided if possible.

RealmSwift/Tests/PerformanceTests.swift
```diff
@@ -0,0 +1,462 @@
...
+    let realm = Realm(inMemoryIdentifier: factor.description)
+    realm.write {
+        for _ in 0..<(1000 * factor) {
+            realm.create(SwiftStringObject.self, value: ["a"])
+            realm.create(SwiftStringObject.self, value: ["b"])
+        }
+    }
+    return realm
+}
+private var smallRealm: Realm!
```
alazier: Doesn't look like these aren't even used. They should be removed.

RealmSwift/Tests/PerformanceTests.swift
```diff
@@ -0,0 +1,462 @@
...
+private var mediumRealm: Realm!
+private var largeRealm: Realm!
+
+class SwiftPerformanceTests: TestCase {
+    override class func setUp() {
```
alazier: We should avoid test specific code in setUp and tearDown whenever possible.

RealmSwift/Tests/ObjectSchemaTests.swift
segiddins: it's not "needed" per-se but it was a member before (this PR just moves setting it into the autoreleasepool) and it helps reduce code duplication

RealmSwift/Tests/PerformanceTests.swift
segiddins: they are used to keep those realms alive

RealmSwift/Tests/PerformanceTests.swift
segiddins: this is in the class setUp method and is used to avoid recreating the same realms (which go unmodified) for each test. This pattern was copied verbatim from the ObjC performance tests

RealmSwift/Tests/ObjectSchemaTests.swift
alazier: This is not meaningful code duplication, and it makes the rest of the code less readable to other developers.

RealmSwift/Tests/PerformanceTests.swift
tgoyne: Copies of the realms created in `setUp` are used in a bunch of the tests because inserting the objects into a realm for each test was responsible for the bulk of the test runtime (significantly more than the code actually being benchmarked in many of the tests).

alazier: The fact that this needs to be explained here rather than being clear from the code/comments says to me that we either need to change this or document it.

Is there a reason we can't simply create the Realm outside of `measureBlock`/`inMeasureBlock`?

tgoyne: The problem being solved isn't that the measurements were thrown off by the setup time; it's that running the tests on device took an extremely long time.

alazier: That's a fair argument, but this should be mentioned in the comments.

More importantly, I still don't see where these realms are used.

tgoyne: They aren't used directly. `getStringObjects()` returns an in-memory realm with the same identifier, and since they're in-memory there needs to be a strong reference lying around to keep the data alive.

The whole thing could use a refactoring (or at least a good explanation in the code).

alazier: That makes sense, but I don't see `createStringObjects` getting called anywhere else in this code, nor do I see uses of the member realm instances - not sure these realms are even used in the swift tests.

segiddins: Those realms are accessed via `getStringObjects`, just like in the ObjC tests

tgoyne: `createStringObjects()` is called in `setUp()` to create three realms with test objects, and strong references to those are stored in global variables to keep the in-memory realms alive. Each test which wants one of the pre-populated realms calls `getStringObjects()`, which re-opens the in-memory realm by creating a new Realm with the same identify (and actually gets the same RLMRealm from the cache), makes a copy of it, and returns a Realm referencing the copy.

The whole point is that `createStringObjects()` is only called once in the class `setUp()` function rather than in each test which needs a test realm, and the global variables are not accessed because they're used purely to keep the in-memory realms alive, with actual access to the data being done by creating new Realms with the same identifier.

alazier: Ok. We shoud get rid of getStringObjects though and just use the realm instances. Found one instance where we call getStringObjects with 10 as an arguments, which means that test effectively does nothing. Someone in the future will undoubtedly change these numbers again and run into the same issue.

RealmSwift/Tests/PerformanceTests.swift
```diff
@@ -0,0 +1,462 @@
...
+        measureBlock {
+            let list = arrayPropertyObject.array
+            for i in 0..<list.count {
+                let string = list[i].stringCol
+            }
+        }
+    }
+
+    func testEnumerateAndMutateAll() {
+        let realm =  getStringObjects(10)
```
alazier: does a realm with 10 even exist?

segiddins: no

RealmSwift/Tests/PerformanceTests.swift
alazier: Or change getStringObjects to `copyRealmToTestPath` and pass in the existing Realm.

tgoyne: Yes, I'm in full agreement that the current design is poor and it needs to be refactored. Explicitly getting a copy of the global realms sounds like a good approach (could skip the copy for non-mutating tests, but that sounds a little error-prone).

alazier:
Once we replace `getStringObjects` with `copyRealmToTestPath` or the equivalent fix :+1:

segiddins:
> Once we replace getStringObjects with copyRealmToTestPath or the equivalent fix :+1:

Done!

PR #1687, opened 2015-03-24
Fix cannot retrieve keychain item in release build.

kishikawakatsumi:
The Keychain API `SecItemCopyMatching()` with `Unmanaged#takeUnretainedValue()` cannot retrieve the value correctly in Release build (to be precise, fastest/smallest optimization [Os]).
`SecItemCopyMatching()` with `Unmanaged#takeUnretainedValue()` always return `nil` in release build. Maybe it's swift bug.

So this example project always crashes to cast nil value if build with release configuration.

```swift
return dataTypeRef?.takeUnretainedValue() as NSData // Crashes this line to cast nil value
```

To fix this issue use `withUnsafeMutablePointer()` function instead `takeUnretainedValue()`.

See also http://stackoverflow.com/questions/24145838/querying-ios-keychain-using-swift/27721328#27721328

Could you please review @alazier @jpsim ?

Version: Xcode 6.2, realm-cocoa 0.91.1 and master

### DIFF
--- a/examples/ios/swift-next/Encryption/ViewController.swift
+++ b/examples/ios/swift-next/Encryption/ViewController.swift
@@ -102,6 +102,8 @@ class ViewController: UIViewController {
             kSecReturnData: true
         ]
 
+        // To avoid Swift optimization bug, should use withUnsafeMutablePointer() function to retrieve the keychain item
+        // See also: http://stackoverflow.com/questions/24145838/querying-ios-keychain-using-swift/27721328#27721328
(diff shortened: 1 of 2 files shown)


jpsim:
Please add a comment to the examples explaining why this is needed, perhaps linking to your StackOverflow answer.

tgoyne:
:+1:

alazier:
:+1:

segiddins:
@kishikawakatsumi :+1:

PR #1694, opened 2015-03-26
Forcibly deallocate leaked RLMRealms in Swift tests

tgoyne:
Leaking RLMRealms in assertThrows eventually results in us hitting GCD's thread limit, and then dispatch_async() stops actually doing anything and we don't get any change notifications for newly opened Realms, resulting in the tests added in https://github.com/realm/realm-cocoa/pull/1686 hanging.

We also were somewhat close to hitting the per-process open file limit on iOS (which is 256 by default).

@alazier @segiddins


RealmSwift/Tests/TestUtils.mm
```diff
@@ -33,3 +36,23 @@ void RLMAssertThrows(XCTestCase *self, dispatch_block_t block, NSString *message
         [self recordFailureWithDescription:message inFile:fileName atLine:lineNumber expected:NO];
     }
 }
+
+void RLMDeallocateRealm(NSString *path) {
```
segiddins: `RLMDeallocateRealmAtPath`?

segiddins:
Can you please explain how leaking the Realms was also causing a GCD threading issues?

tgoyne:
Notifications are done by listening on a pipe on a separate thread.

segiddins:
:+1: it was only a matter of time before we had a forceable `dealloc` somewhere in the repo :)

alazier:
We should only need to do this when using assertThrows. Doing this elsewhere seems dangerous.

alazier:
We are also assuming that only Realm objects are getting leaked. If there are other object leaked which point to a Realm wont this cause issues?

alazier:
Also, wouldn't calling `invalidate` on any leaked realms also address the issues by stopping notifications? This seems like it would be a safer approach.

tgoyne:
> We should only need to do this when using assertThrows. Doing this elsewhere seems dangerous.

Wouldn't be hard to track if a given test uses `assertThrows`. Aborting if there's an RLMRealm still alive when a test which doesn't use `assertThrows` would probably be useful.

I think all of the scenarios that could lead to this doing bad things would be pretty clearly bugs in our tests, and with the debug malloc it should consistently crash on double-frees, so I don't think there's any great risk of them going unnoticed.

> We are also assuming that only Realm objects are getting leaked. If there are other object leaked which point to a Realm wont this cause issues?

`RLMRealm` (and things owned by it) is the only thing with non-memory resources which need to be released.

> Also, wouldn't calling `invalidate` on any leaked realms also address the issues by stopping notifications?

No, `invalidate` doesn't do that.

alazier:
Should calling `invalidate` stop notifications for that instance?

tgoyne:
Waiting to create the notifier until the read transaction begins and destroying it when the read transaction ends is probably a good idea (I didn't do it because in a previous version of the notifications code it got way too complicated, but I think it'd be fine with the current design), but isn't enough to make leaking `RLMRealm`s fine. At 5(!) files per `RLMRealm` we can only have 51 `RLMRealm` instances open at a time.

tgoyne:
Er, 51 unique paths for `RLMRealm`s open at a time. Multiple `RLMRealm`s for the same file on different threads don't count against that limit (and the actual limit's a little higher since some of the write log files won't actually be opened sometimes).

RealmSwift/Tests/TestUtils.mm
```diff
@@ -33,3 +36,23 @@ void RLMAssertThrows(XCTestCase *self, dispatch_block_t block, NSString *message
...
+    __weak RLMRealm *realm;
+
+    @autoreleasepool {
+        realm = RLMGetThreadLocalCachedRealmForPath(path);
+        if (!realm) {
```
alazier: Could get rid of this block as it is checked in the next loop.

tgoyne:
Switched to asserting that the default realm wasn't leaked for tests that don't call `assertThrows` rather than deallocating it.

alazier:
looks good :+1:

PR #1701, opened 2015-03-30
[Project] Add Realm.xcodeproj as subproject to RealmSwift.xcodeproj

mrackwitz:
This has the following advantages:
* allows sane debugging of RealmSwift, because you can step into the underlying Objective-C++ binding
* makes the progress of building the target dependencies observable
* enables to use more support of Xcode to cache & clean builds

The first commit was created programmatically.
```ruby
require 'xcodeproj'
project = Xcodeproj::Project.open('RealmSwift.xcodeproj')
project.generate_available_uuid_list(project.objects_by_uuid.count) #optimization
objects_by_old_uuid = project.objects_by_uuid.clone
project.objects_by_uuid.each do |_, object|
   object.instance_variable_set "@uuid", project.generate_uuid
end
project.targets.each do |target|
  target.dependencies.each do |dep|
    dep.target_proxy.container_portal = objects_by_old_uuid[dep.target_proxy.container_portal].uuid
  end
end
project.save
```

### DIFF
--- a/RealmSwift.xcodeproj/project.pbxproj
+++ b/RealmSwift.xcodeproj/project.pbxproj
@@ -6,31 +6,6 @@
 	objectVersion = 46;
 	objects = {
 
-/* Begin PBXAggregateTarget section */
-		24CBD82935847103D0D36207 /* Realm OSX */ = {
-			isa = PBXAggregateTarget;
-			buildConfigurationList = 9C8C2B9D1C8548E0D3D70BF8 /* Build configuration list for PBXAggregateTarget "Realm OSX" */;
-			buildPhases = (
-				19AF9027433D8DF3244F941B /* ShellScript */,
-			);
-			dependencies = (
-			);
-			name = "Realm OSX";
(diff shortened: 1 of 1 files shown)


alazier:
Things up are currently set up as is because we will be distributing RealmSwift.xcodeproj in the release package. We do not want to ship Realm.xcodeproj as well.

mrackwitz:
Right! That would fail at the mmoent, because Xcode would miss the explicit target dependencies.
We could strip out those and the file reference to the project (but not required) from RealmSwift.xcodeproj in the build script using `xcodeproj`. The build settings and shell script paths for using pre-built frameworks are kept and compatible with that approach.

jpsim:
> We could strip out those and the file reference to the project (but not required) from RealmSwift.xcodeproj in the build script using `xcodeproj`.

That's what I did originally in #1645, but after some convincing by others, realized that there was a cleaner way, which didn't involve modifying the project file, so that's what we're doing now.

PR #1711, opened 2015-03-31
Avoid using private implementation details in the Swift tests

tgoyne:
Explicitly wrap things creating Realms in `setUp` methods in `autoreleasepool` blocks to avoid the need to wrap all calls to `setUp` explicitly, and then just call `super.invokeTest()` in an autorelease pool rather than trying to replicate all of the work it does. Requires an obj-c helper to work around not being able to use `super` in a block in swift 1.1.

@segiddins @jpsim

### DIFF
--- a/RealmSwift/Tests/ListTests.swift
+++ b/RealmSwift/Tests/ListTests.swift
@@ -39,15 +39,13 @@ class ListTests: TestCase {
         arrayObject = createArray()
         array = arrayObject.array
 
-        autoreleasepool {
-            let realm = self.realmWithTestPath()
-            realm.write {
-                realm.add(self.str1)
-                realm.add(self.str2)
-            }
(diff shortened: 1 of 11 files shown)


RealmSwift/Tests/ListTests.swift
```diff
@@ -39,13 +39,15 @@ class ListTests: TestCase {
         arrayObject = createArray()
         array = arrayObject.array
 
-        let realm = realmWithTestPath()
-        realm.write {
-            realm.add(self.str1)
-            realm.add(self.str2)
+        autoreleasepool {
```
segiddins: this doesnt need to be here

segiddins:
:+1:

jpsim:
:+1:

Issue #1707, opened 2015-03-30
Test untested RealmSwift functionality


jpsim:
The following RealmSwift functionality is currently untested:

- [x] NSNull for object
- [x] NSNull for list
- [x] literals with standalone objects
- [x] literals with existing objects
- [x] literals with existing lists
- [x] Object & DynamicObject subscripts
- [x] DynamicObject in general

Let's check off the items as we add these tests.

PR #1713, opened 2015-03-31
Properly support KVC during migrations

### LINKED ISSUE #1708: Add support for KVC for RealmSwift's DynamicObject

segiddins:
Closes #1708.
Part of #1707.

\c @tgoyne @jpsim

### DIFF
--- a/Realm/RLMObject.mm
+++ b/Realm/RLMObject.mm
@@ -16,6 +16,7 @@
 //
 ////////////////////////////////////////////////////////////////////////////
 
+#import "RLMAccessor.h"
 #import "RLMObject_Private.h"
 #import "RLMObjectSchema_Private.hpp"
 #import "RLMObjectStore.h"
@@ -150,6 +151,18 @@
 
 + (NSArray *)ignoredProperties {
     return nil;
+}
+
+@end
+
+@implementation RLMDynamicObject
+
+- (id)valueForKey:(NSString *)key {
+    return RLMDynamicGet(self, key);
+}
+
+- (void)setValue:(id)value forKey:(NSString *)key {
+    RLMDynamicValidatedSet(self, key, value);
 }
 
 @end
--- a/Realm/RLMObject.mm
+++ b/Realm/RLMObject.mm
@@ -157,11 +157,11 @@
 
 @implementation RLMDynamicObject
 
-- (id)valueForKey:(NSString *)key {
+- (id)valueForUndefinedKey:(NSString *)key {
     return RLMDynamicGet(self, key);
 }
 
-- (void)setValue:(id)value forKey:(NSString *)key {
+- (void)setValue:(id)value forUndefinedKey:(NSString *)key {
     RLMDynamicValidatedSet(self, key, value);
 }
 
--- a/Realm/RLMObjectSchema.mm
+++ b/Realm/RLMObjectSchema.mm
@@ -221,7 +221,7 @@
         }
     }
 
-    // for dynamic schema use vanilla RLMObject accessor classes
+    // for dynamic schema use vanilla RLMDynamicObject accessor classes
     schema.objectClass = RLMObject.class;
     schema.accessorClass = RLMDynamicObject.class;
     schema.standaloneClass = RLMObject.class;
--- a/Realm/RLMObjectSchema.mm
+++ b/Realm/RLMObjectSchema.mm
@@ -84,7 +84,7 @@
     }
     schema.className = className;
     schema.objectClass = objectClass;
-    schema.accessorClass = RLMObject.class;
+    schema.accessorClass = RLMDynamicObject.class;
     schema.isSwiftClass = isSwift;
 
     // create array of RLMProperties, inserting properties of superclasses first
@@ -223,7 +223,7 @@
 
     // for dynamic schema use vanilla RLMObject accessor classes
     schema.objectClass = RLMObject.class;
-    schema.accessorClass = RLMObject.class;
+    schema.accessorClass = RLMDynamicObject.class;
     schema.standaloneClass = RLMObject.class;
 
     return schema;
--- a/RealmSwift/Object.swift
+++ b/RealmSwift/Object.swift
@@ -250,6 +250,16 @@ public final class DynamicObject : Object {
         }
         return nil
     }
+
+    /// :nodoc:
+    public override func valueForKey(key: String) -> AnyObject? {
+        return self[key]
+    }
+
+    /// :nodoc:
+    public override func setValue(value: AnyObject?, forKey key: String) {
+        self[key] = value
+    }
 }
 
 /// :nodoc:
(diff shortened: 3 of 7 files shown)


Realm/RLMObjectSchema.mm
```diff
@@ -223,7 +223,7 @@ +(instancetype)schemaFromTableForClassName:(NSString *)className realm:(RLMRealm
 
     // for dynamic schema use vanilla RLMObject accessor classes
```
tgoyne: Comment needs to be updated.

Realm/RLMObject.mm
```diff
@@ -153,3 +154,15 @@ + (NSArray *)ignoredProperties {
 }
 
 @end
+
+@implementation RLMDynamicObject
+
+- (id)valueForKey:(NSString *)key {
```
alazier: This won't work for valid properties of RLMObject such as `realm` and `objectSchema`. It probably should. One way to do this cleanly is to instead implement `valueForUndefinedKey`

alazier: Same thing goes for `setValue:forKey:`

RealmSwift/Object.swift
```diff
@@ -250,6 +250,16 @@ public final class DynamicObject : Object {
         }
         return nil
     }
+
+    /// :nodoc:
+    public override func valueForKey(key: String) -> AnyObject? {
```
alazier: This breaks `realm`, `objectSchema`, `invalidated`, and 'description` (same as for the objc version)

RealmSwift/Tests/ObjectTests.swift
```diff
@@ -179,15 +179,82 @@ class ObjectTests: TestCase {
...
+        list.removeAll();
+        setter(object, list, "arrayCol")
+        XCTAssertEqual((getter(object, "arrayCol") as List<DynamicObject>).count, 0)
+
+        setter(object, [boolObject], "arrayCol")
+        XCTAssertEqual((getter(object, "arrayCol") as List<DynamicObject>).count, 1)
+        XCTAssertEqual((getter(object, "arrayCol") as List<DynamicObject>).first!, boolObject)
+    }
+
+    private func withMigrationObject(block: ((MigrationObject, Migration) -> ())) {
```
alazier: It isn't clear to me what this function does from the name alone. Either need a comment or a better name.

RealmSwift/Tests/ObjectTests.swift
```diff
@@ -179,15 +179,83 @@ class ObjectTests: TestCase {
...
+        setter(object, list, "arrayCol")
+        XCTAssertEqual((getter(object, "arrayCol") as List<DynamicObject>).count, 0)
+
+        setter(object, [boolObject], "arrayCol")
+        XCTAssertEqual((getter(object, "arrayCol") as List<DynamicObject>).count, 1)
+        XCTAssertEqual((getter(object, "arrayCol") as List<DynamicObject>).first!, boolObject)
+    }
+
+    /// Yields a read-write migration `SwiftObject` to the given block
+    private func withMigrationObject(block: ((MigrationObject, Migration) -> ())) {
```
alazier: perhaps rename `performWithMigrationObject`

alazier:
We should test `valueForKey` works for common properties. Should also test that it still throws for invalid properties.

alazier:
We also need to update the changelog to say migration blocks now return DynamicObjects. Should also read through the objc docs as well to make sure we update places we incorrectly mention `RLMObject` where it should now be `RLMDynamicObject`

alazier:
We may also want to consider changing `RLMObjectMigrationBlock` to return `RLMDynamicObject`. Not sure this is worth it though as it will most likely break existing users.

segiddins:
As it stands right now, RLMDynamicObject isn't public. 

-Samuel E. Giddins

On Apr 2, 2015, at 7:10 PM, Ari Lazier <notifications@github.com> wrote:

We may also want to consider changing RLMObjectMigrationBlock to return RLMDynamicObject. Not sure this is worth it though as it will most likely break existing users.

tgoyne:
IMO `RLMDynamicObject` should remain purely an implementation detail that does not appear anywhere in the API or documentation until we open up the dynamic API.

RealmSwift/Tests/ObjectTests.swift
segiddins: the `with` is idiomatic swift naming, see `withCString`, `withUnsafePointer`, etc

segiddins:
@alazier added those tests

RealmSwift/Tests/ObjectTests.swift
```diff
@@ -179,15 +179,83 @@ class ObjectTests: TestCase {
...
+            let realm = self.realmWithTestPath()
+            realm.write {
+                _ = realm.create(SwiftObject)
+            }
+        }
+        autoreleasepool {
+            var enumerated = false
+            setSchemaVersion(1, self.testRealmPath()) { migration, _ in
+                migration.enumerate(SwiftObject.className()) { oldObject, newObject in
+                    if let newObject = newObject {
```
alazier: We don't test `oldObject` and we need to

segiddins: the reason I currently don't is because there's no way to set the value  on an `oldObject`, and thus it's very hard to know what to assert things are equal to

alazier: The issue is that your new code written doesn't actually get executed using `newObject` - we only create a dynamic realm and dynamic objects for the old realm, as the new realm is not dynamic and uses the normal accessors. 

We should at least test getting values from the old object. Alternatively, we could choose to put this off until later when we have dynamic api support outside the context of migrations which would make this more testable.

segiddins:
@jpsim what do I need to do to get this merged?

RealmSwift/Object.swift
```diff
@@ -250,6 +250,16 @@ public final class DynamicObject : Object {
         }
         return nil
     }
+
+    /// :nodoc:
+    public override func valueForUndefinedKey(key: String) -> AnyObject? {
```
jpsim: Are these overrides necessary?

segiddins: Yes, since these are dynamic objects and therefore won't have accessor methods.

jpsim:
Rebase this and :+1:

RealmSwift/Object.swift
JadenGeller: Why is `valueForUndefinedKey` used instead of `valueForKey`?

segiddins: It was either performance or so that KVC accessors for other properties would continue to work

JadenGeller: Calling super would allow other properties to still work, and afaik you could call super before looking up the property in the Realm so you can get the same performance, no?

PR #1715, opened 2015-04-01
Add tests for RLMArray delete methods

kishikawakatsumi:
Add 2 tests for `RLMArray#removeLastObject` and `RLMArray#removeAllObjects `.

Because there are no tests for the standalone array object's delete methods, although there are tests for array object after persistence.

I found that `removeLastObject ` and `removeAllObjects` in methods of RLMArray are not covered by the unit tests. There are some test cases that calls same name methods, but they are for the array after persistence (not for stand alone objet, they does not call RLMArray's methods).

Code coverage of RLMArray.mm increased (+6.5%) from 83.7 to 89.2% when merged this PR.

Could you please review? @jpsim @segiddins @alazier

### DIFF
--- a/Realm/Tests/ArrayPropertyTests.m
+++ b/Realm/Tests/ArrayPropertyTests.m
@@ -234,7 +234,7 @@
     XCTAssertEqualObjects(array.array[0], stringObj1, @"Objects should be equal");
     XCTAssertEqualObjects(array.array[1], stringObj2, @"Objects should be equal");
     XCTAssertEqualObjects(array.array[2], stringObj3, @"Objects should be equal");
-    XCTAssertEqual(array.array.count, 3U, @"Should have three elements in string array");
+    XCTAssertEqual(array.array.count, 3U, @"Should have 3 elements in string array");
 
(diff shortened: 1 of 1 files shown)


Realm/Tests/ArrayPropertyTests.m
```diff
@@ -207,6 +207,59 @@ -(void)testStandalone {
...
+
+    XCTAssertEqualObjects(array.array[0], stringObj1, @"Objects should be equal");
+    XCTAssertEqual(array.array.count, 1U, @"Should have 1 elements in string array");
+
+    [array.array removeLastObject];
+
+    XCTAssertEqual(array.array.count, 0U, @"Should have 0 elements in string array");
+
+    [array.intArray removeAllObjects];
+    XCTAssertEqual(array.intArray.count, 0U, @"Should have 0 elements in string array");
```
segiddins: `int array`?

segiddins:
:+1:

Realm/Tests/ArrayPropertyTests.m
kishikawakatsumi: Fixed, thanks.

tgoyne:
:+1:

PR #1716, opened 2015-04-01
Clean up watch extension example project

kishikawakatsumi:
#### Clean up the garbage of the deleted file :put_litter_in_its_place:

I found that deleted file remains in the navigator area. 

####  Fix code signing entitlements file name for case-sensitive file system

The first character of `Extension.entitlements` filename is upper case, but written in lowercase for code signing entitlements setting.

CC: @jpsim @segiddins


segiddins:
:+1: the CI failure looks unrelated

tgoyne:
:+1:

PR #1720, opened 2015-04-01
[List][Results] Use NSFastGenerator for generate()

segiddins:
\c @tgoyne @jpsim


tgoyne:
Does give a bit of a speed boost and simplifies the code, so :+1: .

jpsim:
Sweet! :+1:

PR #1721, opened 2015-04-01
[List] Implement description with max depth

segiddins:
\c @jpsim @tgoyne

### DIFF
--- a/Realm/RLMObject_Private.h
+++ b/Realm/RLMObject_Private.h
@@ -61,6 +61,8 @@ FOUNDATION_EXTERN BOOL RLMObjectBaseAreEqual(RLMObjectBase *o1, RLMObjectBase *o
 // Get ObjectUil class for objc or swift
 FOUNDATION_EXTERN Class RLMObjectUtilClass(BOOL isSwift);
 
+FOUNDATION_EXTERN const NSUInteger RLMDescriptionMaxDepth;
+
 @class RLMProperty, RLMArray;
 @interface RLMObjectUtil : NSObject
 
--- a/RealmSwift/List.swift
+++ b/RealmSwift/List.swift
@@ -32,7 +32,7 @@ public class ListBase: RLMListBase, Printable {
 
     @objc private func descriptionWithMaxDepth(depth: UInt) -> String {
         let type = "List<\(_rlmArray.objectClassName)>"
-        return gsub("RLMArray <0x[a-z0-9]+>", type, _rlmArray.description) ?? type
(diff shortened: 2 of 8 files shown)


RealmSwift/List.swift
```diff
@@ -27,6 +27,10 @@ public class ListBase: RLMListBase, Printable {
     // generic class.
     /// Returns a human-readable description of the objects contained in the list.
     @objc public override var description: String {
+        return descriptionWithMaxDepth(RLMDescriptionMaxDepth)
+    }
+
+    @objc private func descriptionWithMaxDepth(depth: UInt) -> String {
         let type = "List<\(_rlmArray.objectClassName)>"
         return gsub("RLMArray <0x[a-z0-9]+>", type, _rlmArray.description) ?? type
```
jpsim: You'll need to pass `depth` to `_rlmArray.descriptionWithMaxDepth(_:)`.

jpsim:
Why is this at all needed? Making a constant for `RLMDescriptionMaxDepth` is nice, but AFAICT this PR doesn't "fix" anything.

segiddins:
It does, since the `List` description is used for Array properties on object. Without this fix, it would use `description` instead of `descriptionWithMaxDepth`, leaving open the possibility of infinite recursion.

jpsim:
Alright, I now understand what this is doing. Could you test the description of a `Results` with embedded `List`s? Right now it's just testing with `SwiftStringObject`.

segiddins:
Added a test that covers recursive objects

jpsim:
:+1:

segiddins:
@tgoyne this look OK to you?

Realm/RLMObject_Private.h
```diff
@@ -61,6 +61,8 @@ FOUNDATION_EXTERN BOOL RLMObjectBaseAreEqual(RLMObjectBase *o1, RLMObjectBase *o
 // Get ObjectUil class for objc or swift
 FOUNDATION_EXTERN Class RLMObjectUtilClass(BOOL isSwift);
 
+FOUNDATION_EXTERN const NSUInteger RLMDescriptionMaxDepth;
```
tgoyne: Should just be a `static const` variable rather than an `extern` one. The point of externing string constants is to ensure that they can be compared with pointer equality.

segiddins: According to https://twitter.com/jckarter/status/583725475777748992, using `static const` declarations from Swift doesn't work (and is a known bug).

PR #1723, opened 2015-04-02
[RLMRealm] added +migrationRequiredAtPath:encryptionKey:error:

### LINKED ISSUE #1692: Allow configuring a Realm file to be deleted when a migration is required

```objc

        }];
    }
}];

```

```objc

}

@try {
}
    
        RLMCleanupFiles(realmPath);
    }
    else {
        @throw; // rethrow
    }
}
```

jpsim:
Fixes #1692. /cc @alazier @tgoyne @segiddins


Realm/RLMRealm.mm
```diff
@@ -766,6 +766,32 @@ + (NSUInteger)schemaVersionAtPath:(NSString *)realmPath encryptionKey:(NSData *)
...
+        key = validatedKey(key) ?: keyForPath(realmPath);
+        realm = [[RLMRealm alloc] initWithPath:realmPath key:key readOnly:YES inMemory:NO dynamic:YES error:&error];
+        if (error) {
+            RLMSetErrorOrThrow(error, outError);
+            // the realm could not be created, which means it cannot be migrated.
+            return NO;
+        }
+    }
+    for (RLMObjectSchema *objectSchema in [[RLMSchema sharedSchema] objectSchema]) {
+        RLMObjectSchema *tableSchema = [RLMObjectSchema schemaFromTableForClassName:objectSchema.className realm:realm];
```
segiddins: make you you test this works when `tableSchema == nil`

RealmSwift/Realm.swift
```diff
@@ -512,11 +511,26 @@ public final class Realm {
...
+    
+    :param: path          Path to a Realm file.
+    :param: encryptionKey Optional 64-byte encryption key.
+    :param: error         If an error occurs, upon return contains an `NSError` object
+                          that describes the problem. If you are not interested in
+                          possible errors, omit the argument, or pass in `nil`.
+
+    :returns:             Whether or not the schema of the Realm at `path` matches the current schema.
+    */
+    public class func migrationRequired(atPath path: String = defaultPath, encryptionKey: NSData? = nil, error: NSErrorPointer = nil) -> Bool {
```
segiddins: probably belongs in `Migration.swift` as a free function?

RealmSwift/Tests/MigrationTests.swift
```diff
@@ -32,14 +32,14 @@ private func realmWithSingleClass(path: String, objectSchema: RLMObjectSchema) -
     return realmWithCustomSchema(path, schema)
 }
 
-private func realmWithSingleClassProperties(path: String, className: String, properties: [AnyObject]) -> RLMRealm {
+internal func realmWithSingleClassProperties(path: String, className: String, properties: [AnyObject]) -> RLMRealm {
```
segiddins: if this is going to be internal, move it to the util file?

RealmSwift/Tests/MigrationTests.swift
```diff
@@ -51,7 +51,9 @@ class MigrationTests: TestCase {
     }
 
     // migrate realm at path and ensure migration
-    private func migrateAndTestRealm(realmPath: String, shouldRun: Bool = true, schemaVersion: UInt = 1, autoMigration: Bool = false, block: MigrationBlock? = nil) {
+    private func migrateAndTestRealm(realmPath: String, isNecessary: Bool = false, shouldRun: Bool = true, schemaVersion: UInt = 1, autoMigration: Bool = false, block: MigrationBlock? = nil) {
+        var error: NSError? = nil
+        XCTAssertEqual(Realm.migrationRequired(atPath: realmPath, error: &error), isNecessary)
```
segiddins: check there's an error iff `isNecessary`

Realm/RLMRealm.mm
```diff
@@ -766,6 +766,32 @@ + (NSUInteger)schemaVersionAtPath:(NSString *)realmPath encryptionKey:(NSData *)
...
 
++ (BOOL)migrationRequiredAtPath:(NSString *)realmPath encryptionKey:(NSData *)key error:(NSError **)outError {
+    RLMRealm *realm = RLMGetThreadLocalCachedRealmForPath(realmPath);
+    if (realm) {
+        // we've already successfully opened a realm, so a migration isn't necessary.
+        return NO;
+    } else {
+        NSError *error;
+        key = validatedKey(key) ?: keyForPath(realmPath);
+        realm = [[RLMRealm alloc] initWithPath:realmPath key:key readOnly:YES inMemory:NO dynamic:YES error:&error];
```
segiddins: should all of this be wrapped in an `autoreleasepool`?

Realm/RLMRealm.mm
```diff
@@ -766,6 +766,32 @@ + (NSUInteger)schemaVersionAtPath:(NSString *)realmPath encryptionKey:(NSData *)
...
+        if (error) {
+            RLMSetErrorOrThrow(error, outError);
```
segiddins: do we really want to return an error if the file doesn't exist? should return `NO` in my opinion

jpsim: It's a programmer error if you ask if a path with no realm file requires a migration. We *do* return `NO` if an error pointer is passed in.

Realm/RLMRealm.mm
jpsim: No. The realm will be released at the end of the method, so it's safe to delete the file after calling this, as long as there are no other active realms. Although we should probably add a `+[RLMRealm isRealmOpenAtPath:]` method too, because otherwise it will be difficult for users to know for sure if they can delete the file.

RealmSwift/Realm.swift
jpsim: Yes.

RealmSwift/Tests/MigrationTests.swift
jpsim: Sure.

RealmSwift/Tests/MigrationTests.swift
jpsim: 1. that's not always the case since you'll get all the errors that could happen when opening a realm (e.g. file doesn't exist, etc.)
2. those errors are already tested elsewhere

Realm/RLMRealm.h
```diff
@@ -551,6 +551,19 @@ typedef void (^RLMMigrationBlock)(RLMMigration *migration, NSUInteger oldSchemaV
 + (NSUInteger)schemaVersionAtPath:(NSString *)realmPath encryptionKey:(NSData *)key error:(NSError **)error;
 
 /**
+ Returns whether or not the realm at the given path can be accessed without performing a migration.
+
+ @param realmPath   Path to a Realm file.
+ @param key         64-byte encryption key.
```
tgoyne: "64-byte encryption key or nil for unencrypted Realms."

It's not optional most of the places where there's a key parameter, so it needs to call out that this one is.

Realm/RLMRealm.mm
```diff
@@ -766,6 +766,32 @@ + (NSUInteger)schemaVersionAtPath:(NSString *)realmPath encryptionKey:(NSData *)
     return RLMRealmSchemaVersion(realm);
 }
 
++ (BOOL)migrationRequiredAtPath:(NSString *)realmPath encryptionKey:(NSData *)key error:(NSError **)outError {
+    RLMRealm *realm = RLMGetThreadLocalCachedRealmForPath(realmPath);
```
tgoyne: `RLMGetAnyCachedRealmForPath`

Realm/RLMRealm.mm
tgoyne: It doesn't. Should just continue if `tableSchema` is nil (since creating missing tables doesn't require a migration).

Realm/RLMRealm.mm
tgoyne: Why? If there's no Realm file there, then opening a Realm file at that path doesn't require a migration. Making the user explicitly check for a file there first is unfriendly and race-conditiony.

Realm/RLMRealm.mm
jpsim: Ah, we should definitely do this in `schemaVersionAtPath:` too then.

tgoyne: No, schemaVersionAtPath actually reads from the realm so it needs one for the current thread.

Realm/RLMRealm.mm
jpsim: So do you think we should ignore just that one error, and set or throw as we are now otherwise?

Realm/RLMRealm.mm
jpsim: `migrationRequiredAtPath:` also reads from the realm (`+[RLMObjectSchema schemaFromTableForClassName:realm:]`), so both should continue to use `RLMGetThreadLocalCachedRealmForPath()`.

tgoyne: It only reads from the realm if it opened it. This part is only checking if the realm is already open, and if it is it returns without doing anything further.

Realm/RLMRealm.mm
tgoyne: Ignore anything which wouldn't result in an error when actually opening the Realm (which I think if just file not found, but I'm not certain).

alazier:
I don't think this is the correct solution for #1692

The user of this issue should still be setting a schema version, and we should provide a mechanism to clear the Realm in case the migration is required and run. With this pr you are encouraging the user manually delete files which is very error prone, and you are also teaching them that ignoring schema version is ok in some cases (which it really isn't).

Instead of this pr we should instead add a mechanism to clear the realm safely, and this user should update the schema version when they make schema changes.

jpsim:
I like the idea of a safe realm deletion method, which would take care of checking if there are any open db connections, perhaps even closing them, before deleting all the relevant files. I think we need that regardless of what we do.

However, I do think letting users to know if their realm files require a migration will give them much more flexibility to migrate, delete or otherwise make better informed decisions about what to do with their realms at runtime.

I understand that we want to minimize the odds of accidental deletion, which is why I'd rather not add `deleteRealmBeforeOpening` or `deleteRealmIfMigrationNeeded` options like the java team is considering. But I do think that this PR requires users to be very deliberate about how their realm file is deleted, without requiring to increase a schema version on every change, which can definitely hinder development.

segiddins:
I don't think incrementing a schema version is really that much of a burden, it's literally changing one constant somewhere in your app.

segiddins:
Let's revisit & try and get this merged?

jpsim:
This has been updated and is ready for review again /cc @segiddins @tgoyne @bdash

jpsim:
Closing in favor of https://github.com/realm/realm-cocoa/issues/1584#issuecomment-99160451.

natanrolnik:
@jpsim is this method public? I can't find it to check if my realm needs a migration or not.

jpsim:
@natanrolnik, no this PR was closed without merging. We opted to go with https://github.com/realm/realm-cocoa/issues/1584#issuecomment-99160451 instead, although that has yet to be implemented.

Issue #1684, opened 2015-03-23
row accessors can be deallocated on the incorrect thread


alazier:
It is possible for a row to get deallocated on a different thread from which it is created. This can cause issues as structures used to track accessors for a shared_group can in this case be changed on an incorrect thread possibly causing issues. We need to somehow ensure that row accessors are deallocated on the correct thread, even if the parent objc accessor is deallocated on a different thread.

tgoyne:
We could just call `RLMCheckThread` in `dealloc` since you aren't supposed to pass objects to other threads at all currently. For handover we'll need locks around everything that touches the row accessor linked list (which is not very much).

segiddins:
@tgoyne should this go in the accessor's dealloc or the realm's?

tgoyne:
Accessors being deallocated on the wrong thread is the potentially problematic case.

PR #1727, opened 2015-04-03
[ObjectBase] Throw an exception when an attached accessor is deallocated...

segiddins:
... on the wrong thread

Closes https://github.com/realm/realm-cocoa/issues/1684.

\c @alazier @tgoyne


segiddins:
Got a verbal :+1: from tg and jp.

Issue #1498, opened 2015-02-13
Pass the realm to the block in transactionWithBlock:


CraigSiemens:
Currently when calling `transactionWithBlock:`, it requires you to create a realm variable that you can use in the block.

```objc
RLMRealm *realm = [RLMRealm defaultRealm];
[realm transactionWithBlock:^{
    [realm addObject:foo];
}];
```

It would be nice if the realm was passed into the block for you instead of requiring the user to create a variable every time they need it. The ideal solution would look like this.

```objc
[[RLMRealm defaultRealm] transactionWithBlock:^(RLMRealm *realm){
    [realm addObject:foo];
}];
```

segiddins:
We've actually been discussing this change this week, thanks for the suggestion!

astigsen:
We should probably consider a class method:

```obj-c
[RLMRealm transactionInDefaultRealmWithBlock:^(RLMRealm *realm){
    [realm addObject:foo];
}];
```

yoshyosh:
I like that @astigsen, with the autocomplete it will be easy for users to recall

alazier:
It also makes autocomplete less useful in general as we continue to add more and more methods. Is it worth it? I'm not sure. In any case I see no reason not to pass the Realm to the transaction block other than breaking current code, so we would probably have to have two variants...

yoshyosh:
Agreed, but I think this is the most used function out of all. 3 lines of code to add an object to your DB is very impressive

anlaital:
I like @astigsen's idea. I'm a huge fan of autocomplete and this would not cause me any issues since autocomplete remembers the methods you've used most recently. :+1:

segiddins:
We need to (eventually) decide if this is something we want to do, as it's a major breaking change with no real way to deprecate.

jpsim:
The extra class method is a step backwards since it only works with the default realm, but I could see us implementing the original suggestion:

```objc
[[RLMRealm defaultRealm] transactionWithBlock:^(RLMRealm *realm){
    [realm addObject:foo];
}];
```

But it's not worth it when saving a single-line assignment, only in *some* scenarios. Especially in Swift, where usage would look more like this:

```swift
Realm().write { realm in
  realm.add(foo)
}
// or
let realm = Realm()
realm.write { _ in
  realm.add(foo)
}
```

[...]

PR #1730, opened 2015-04-05
Add transaction block helper for default Realm

fcanas:
This hopefully addresses #1498.

Passing the realm as an argument in `-transactionWithBlock:` would break a great deal of code without making a separate method with a different signature. It also seemed unnecessary since the receiver is likely readily available for capturing in the block.

### DIFF
--- a/Realm/RLMRealm.h
+++ b/Realm/RLMRealm.h
@@ -300,6 +300,11 @@ typedef void(^RLMNotificationBlock)(NSString *notification, RLMRealm *realm);
 - (void)transactionWithBlock:(void(^)(void))block;
 
 /**
+ Helper to perform a block within a write transaction in the default Realm.
+ */
++ (void)transactionInDefaultRealmWithBlock:(void(^)(RLMRealm *))block;
+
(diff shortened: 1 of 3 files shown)


segiddins:
I don't particularly like the idea of putting more class methods on `RLMRealm` that use the notion of a `defaultRealm`, but it's up to the rest of the team if we merge this.

Thanks for the pull request, @fcanas!

fcanas:
@segiddins I hear you.

I don't know if this fits with your architecture or style, but I'd be happy to refactor methods that use the default realm into a category on `RLMRealm`.

jpsim:
Sorry we've been neglecting this, @fcanas. I guess the reason why is that we're a bit uneasy about the current proposal. I could be more in favor of passing the `RLMRealm` instance as a parameter to the block:

```objc
[[RLMRealm defaultRealm] transactionWithBlock:^(RLMRealm *realm){
  // mutating operations here...
}];
```

jpsim:
Closing this due to inactivity.

Issue #1737, opened 2015-04-07
Migration block documentation issues


anlaital:
The following snippet is taken directly from the documentation  (http://realm.io/docs/cocoa/0.91.1/#migrations):

```
// Notice setSchemaVersion is set to 1, this is always set manually. It must be
// higher than the previous version (oldSchemaVersion) or an RLMException is thrown
[RLMRealm setSchemaVersion:1
            forRealmAtPath:[RLMRealm defaultRealmPath] 
        withMigrationBlock:^(RLMMigration *migration, NSUInteger oldSchemaVersion) {
  // We haven’t migrated anything yet, so oldSchemaVersion == 0
  if (oldSchemaVersion < 1) {
    // Nothing to do!
    // Realm will automatically detect new properties and removed properties
    // And will update the schema on disk automatically
  }
}];
```

Firstly, ```It must be higher than the previous version (oldSchemaVersion) or an RLMException is thrown``` is incorrect, as no exception will be thrown if you set the schema version to be equal to what it is on disk.

Secondly, ```oldSchemaVersion``` is of type NSUInteger. The default value of this is ```RLMNotVersioned``` (-1, not part of the public API). ```We haven’t migrated anything yet, so oldSchemaVersion == 0``` is also therefore incorrect.

tgoyne:
The migration block isn't supposed to be getting called at all when the Realm is first created (which is the only time it's RLMNotVersioned; the version is set to zero on init if you don't set a schema version), but it appears that at some point "needs a migration" and "needs the schema version to be set" got lumped together.

anlaital:
It would appear that my migration block is in fact being called when the database is first created. However, with oldSchemaVersion being -1 (NSUIntegerMax) all the linear migrations simply fall through.

lbanders:
I see the same thing. We are just started on adding Realm.IO, so could be doing something wrong :) Here is the code I use for creating the RLMRealm. 
<pre>
    static func RealmFromPath(realmPath: String) -> RLMRealm? {
        RLMRealm.setSchemaVersion(CURRENT_DATABASE_VERSION, forRealmAtPath: realmPath, withMigrationBlock: migrationBlock)

        return RLMRealm(path: realmPath, readOnly: false, error: nil)
    }
</pre>

During this process the migration i started from a database version set to RLMNotVersioned but was expected to be 0.

segiddins:
I believe we've changed it so that the block isn't called when the realm is first created?

[...]

PR #1739, opened 2015-04-07
Don't call migration blocks when first initializing a Realm file

tgoyne:
It doesn't really make sense to do so since there's nothing to migrate, and the old version is an undocumented value that makes checking schema versions awkward. See #1737.

@alazier

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,14 +1,14 @@
 x.xx.x Release notes (yyyy-MM-dd)
 =============================================================
 
+### API breaking changes
+
+* Migration blocks are no longer called when a Realm file is first created.
+
 ### Enhancements
 
 * `Int8` properties defined in Swift are now treated as integers, rather than
   booleans.
-
-### Bugfixes
-
-* Migration blocks are no longer called when a Realm file is first created.
 
 0.91.3 Release notes (2015-04-17)
 =============================================================
@@ -20,6 +20,8 @@ x.xx.x Release notes (yyyy-MM-dd)
 
 0.91.2 Release notes (2015-04-16)
 =============================================================
+
+* Migration blocks are no longer called when a Realm file is first created.
 
 ### Enhancements
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@ x.xx.x Release notes (yyyy-MM-dd)
 * `Int8` properties defined in Swift are now treated as integers, rather than
   booleans.
 
+### Bugfixes
+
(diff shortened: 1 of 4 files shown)


tgoyne:
I had to change what `-[RealmSwift_iOS_Tests.MigrationTests testCreate]` is testing because calling `enumerate()` after adding/removing objects doesn't work in a remotely sensible way if the table exists in both schemas (it'll either crash, fail to enumerate all objects, or give non-matching objects).

RealmSwift/Tests/MigrationTests.swift
```diff
@@ -170,33 +174,35 @@ class MigrationTests: TestCase {
...
-                }
-                count++
-            })
-            XCTAssertEqual(count, 3)
         })
 
         XCTAssertEqual(Realm().objects(SwiftStringObject.self).count, 3)
+
+        var i = 0
+        for obj in Realm().objects(SwiftStringObject) {
```
alazier: Since we only have once conditional for each element might be less code if we just remove the loop and have the 3 `XCTAssertEqual`s

CHANGELOG.md
```diff
@@ -9,6 +9,7 @@ x.x.x Release notes (yyyy-MM-dd)
 
 * `+[RLMSchema dynamicSchemaForRealm:]` now respects search indexes.
 * `+[RLMProperty isEqualToProperty:]` now checks for equal `indexed` properties.
+* Migration blocks are no longer called when a Realm file is first created.
```
alazier: I think we actually want to put this in breaking changes. Some users may have relied on the previous behavior.

alazier:
The one case this pr seems detrimental is if a users wants to pre-populate a realm when updating to a particular schema version. We do not require there to be data in the db for users to add new objects during migration which seems like a perfectly valid use case. Would be nice if we could continue to support this use case in a clean way.

tgoyne:
I feel like populating a DB with seed data in a migration block is sort of a misuse of the concept, and if it's something we want to support then we'd probably want to use a different name. It's also not something that can be cleanly done today (object insertions done in a `if (oldVersion < 5) { ... }` block won't run on a new realm file and they have to explicitly check for -1).

I think the best way to handle that sort of scenario would be to ship a pre-populated realm file with the app, and either copy it to the R/W path if the realm file doesn't already exist, or copy the objects over from the bundled realm in the migration block.

jpsim:
:+1:. Can you rebase this @tgoyne?

segiddins:
:+1:

PR #1740, opened 2015-04-07
Allow Int8 object properties defined in Swift

### LINKED ISSUE #1329: Swift: Create Int property when Int8 is used in a Swift model
As @tgoyne stated in [this mailing list post](https://groups.google.com/forum/?utm_medium=email&utm_source=footer#!msg/realm-cocoa/gWvJ_uxY5kk/HnlHDFCwLGYJ):

> Objective-C's BOOL is unfortunately a typedef for char on 32-bit platforms rather than a proper boolean type, so we assume all 8-bit ints are boolean properties.

But we could still throw an exception if a model is declared in Swift using an Int8. The exception message could suggest that these properties be declared with an `Int` instead.

### LINKED ISSUE #1763: Make an enum for accessor codes
It was unclear to me that they didn't map 1-to-1 with objc type codes, so we should probably make an enum so they're less cryptic

segiddins:
Closes #1329.
Closes #1763.

\c @tgoyne @alazier

### DIFF
--- a/Realm/RLMAccessor.mm
+++ b/Realm/RLMAccessor.mm
@@ -111,7 +111,7 @@ static inline bool RLMGetBool(__unsafe_unretained RLMObjectBase *const obj, NSUI
     RLMVerifyAttached(obj);
     return obj->_row.get_bool(colIndex);
 }
-static inline void RLMSetValue(__unsafe_unretained RLMObjectBase *const obj, NSUInteger colIndex, bool val) {
+static inline void RLMSetValue(__unsafe_unretained RLMObjectBase *const obj, NSUInteger colIndex, BOOL val) {
     RLMVerifyInWriteTransaction(obj);
     obj->_row.set_bool(colIndex, val);
 }
@@ -423,7 +423,7 @@ static IMP RLMAccessorSetter(RLMProperty *prop, RLMAccessorCode accessorCode) {
         case RLMAccessorCodeLongLong: return RLMMakeSetter<long long>(colIndex, prop.isPrimary);
         case RLMAccessorCodeFloat: return RLMMakeSetter<float>(colIndex, prop.isPrimary);
         case RLMAccessorCodeDouble: return RLMMakeSetter<double>(colIndex, prop.isPrimary);
-        case RLMAccessorCodeBool: return RLMMakeSetter<bool>(colIndex, prop.isPrimary);
+        case RLMAccessorCodeBool: return RLMMakeSetter<BOOL>(colIndex, prop.isPrimary);
         case RLMAccessorCodeString: return RLMMakeSetter<NSString *>(colIndex, prop.isPrimary);
         case RLMAccessorCodeDate: return RLMMakeSetter<NSDate *>(colIndex, prop.isPrimary);
         case RLMAccessorCodeData: return RLMMakeSetter<NSData *>(colIndex, prop.isPrimary);
@@ -672,7 +672,7 @@ void RLMDynamicSet(__unsafe_unretained RLMObjectBase *const obj, __unsafe_unreta
             RLMSetValue(obj, col, [val doubleValue]);
             break;
         case RLMAccessorCodeBool:
-            RLMSetValue(obj, col, (bool)[val boolValue]);
+            RLMSetValue(obj, col, [val boolValue]);
             break;
         case RLMAccessorCodeString:
             if (options & RLMCreationOptionsEnforceUnique) {
--- a/Realm/RLMAccessor.mm
+++ b/Realm/RLMAccessor.mm
@@ -410,7 +410,7 @@ static IMP RLMAccessorSetter(RLMProperty *prop, char accessorCode) {
         case 'f': return RLMMakeSetter<float>(colIndex, prop.isPrimary);
         case 'd': return RLMMakeSetter<double>(colIndex, prop.isPrimary);
         case 'B': return RLMMakeSetter<bool>(colIndex, prop.isPrimary);
-        case 'c': return RLMMakeSetter<BOOL, bool>(colIndex, prop.isPrimary);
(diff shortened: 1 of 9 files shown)


tgoyne:
`RLMDynamicGet` also needs to be modified, and need tests that hit all the dynamic get/set code paths.

RealmSwift/Tests/ObjectAccessorTests.swift
```diff
@@ -89,17 +90,39 @@ class ObjectAccessorTests: TestCase {
...
             let obj = SwiftAllIntSizesObject()
 
-            obj.int16 = v16
-            XCTAssertEqual(obj.int16, v16)
-            obj.int32 = v32
-            XCTAssertEqual(obj.int32, v32)
-            obj.int64 = v64
-            XCTAssertEqual(obj.int64, v64)
+            let testObject: Void -> Void = {
+                func setAndTestValue<O: Object, T: Equatable>(object: O, #value: T, #getter: (O) ->(T), #setter: (O, T) -> ()) {
```
tgoyne: This function is pointless. The code using it is not shorter or simpler than just writing the tests directly.

RealmSwift/Tests/ObjectAccessorTests.swift
```diff
@@ -89,17 +90,39 @@ class ObjectAccessorTests: TestCase {
...
+                }
+
+                obj.objectSchema.properties.map { $0.name }.map { obj[$0] = 0 }
+
+                setAndTestValue(obj, value: v8, getter: { Int8($0["int8"]! as Int) }, setter: { $0["int8"] = Int($1) as NSNumber })
+                setAndTestValue(obj, value: v16, getter: { Int16($0["int16"]! as Int) }, setter: { $0["int16"] = Int($1) as NSNumber })
+                setAndTestValue(obj, value: v32, getter: { Int32($0["int32"]! as Int) }, setter: { $0["int32"] = Int($1) as NSNumber })
+                setAndTestValue(obj, value: v64, getter: { Int64($0["int64"]! as Int) }, setter: { $0["int64"] = Int($1) as NSNumber })
+
+                setAndTestValue(obj, value: v8, getter: { Int8($0.valueForKey("int8")! as Int) }, setter: { $0.setValue(Int($1) as NSNumber, forKey: "int8") })
```
tgoyne: Need to repeat the `obj.objectSchema.properties.map { $0.name }.map { obj[$0] = 0 }` between each set of tests.

alazier:
I think we are taking the wrong approach here. We should be creating a new accessor code that is returned from `accessorCodeForType`. Then we can more easily define int8 specific behavior rather than having a conditional every time we have a bool. We do this for other types that collapse down to a single objc code (all objc non-primitives) and this would end up being much cleaner if it works out.

segiddins:
@alazier the issue is that we switch on the type code in a lot of places, instead of using our `RLMPropertyType` enum

alazier:
There are many places where we should be using the accessor code, and not the type. accessorCodes are more specific than the realm type, and map specifically to the underlying ivar type being used. We do the same thing in objc to differentiate between different int types. Conceptually this seems to me to be exactly the case where a new accesorCode should be created.

segiddins:
This has now been cleaned up by introducing an `RLMAccessorCode` enum to decouple those value from objc type codes.

tgoyne:
Combining handling for `BOOL` and `bool` doesn't work. For the getter it's fine, but for the setter passing a `BOOL` to a method expecting a `bool` results in undefined behavior if the `BOOL` is anything other than 0 or 1.

Realm/RLMAccessor.mm
```diff
@@ -507,26 +527,40 @@ static IMP RLMAccessorStandaloneSetter(RLMProperty *prop, char accessorCode) {
...
                 // throw for all primitive types
                 case RLMPropertyTypeBool:
                 case RLMPropertyTypeDouble:
                 case RLMPropertyTypeFloat:
                 case RLMPropertyTypeInt:
-                    @throw RLMException(@"Invalid type for objc typecode");
+                    break;
             }
+        case 'c':
+            switch (rlmType) {
```
alazier: An if/else might be nicer here. Nested switches are kind of nasty.

Realm/RLMAccessor.mm
```diff
@@ -29,6 +29,23 @@
 
 #import <objc/runtime.h>
 
+typedef NS_ENUM(char, RLMAccessorCode) {
+    RLMAccessorCodeChar,
```
alazier: This naming is misleading since we don't support char in objc or swift.

segiddins: We do: an Int8 is actually a char (fixed length types only exist in the eyes of the compiler)

-Samuel E. Giddins

On Apr 20, 2015, at 9:21 AM, Ari Lazier <notifications@github.com> wrote:

In Realm/RLMAccessor.mm:

> @@ -29,6 +29,23 @@
>  
>  #import <objc/runtime.h>
>  
> +typedef NS_ENUM(char, RLMAccessorCode) {
> +    RLMAccessorCodeChar,
This naming is misleading since we don't support char in objc or swift.

alazier: We don't support char in objc - and I don't think we can since BOOL is represented as char.

Realm/RLMAccessor.mm
```diff
@@ -648,19 +682,19 @@ void RLMDynamicSet(__unsafe_unretained RLMObjectBase *const obj, __unsafe_unreta
...
             break;
-        case 't':
+        case RLMAccessorCodeArray:
             RLMSetValue(obj, col, (RLMArray *)val, options);
             break;
-        case '@':
+        case RLMAccessorCodeAny:
             RLMSetValue(obj, col, val);
             break;
         default:
```
tgoyne: Should be able to remove the `default:` cases from most of the places switching on the enum.

segiddins:
@tgoyne does this need anything else?

jpsim:
:+1: This looks good to me. The last CI failure was a single performance test failure in Release mode, but I doubt that the changes in here had any effect on that.

segiddins:
This has now been rebased.

tgoyne:
Need changelog entry

:+1:

PR #1751, opened 2015-04-13
Update to core 0.90.0

tgoyne:


mrackwitz:
@tgoyne: Did you checked whether the renaming of realm-core's header dir from `tightdb` to `realm` is not causing any issues, especially regarding the distribution over CocoaPods? Because by default OSX's file-system is case-insensitive and by that changeset, we loose the separation of the directories for the headers of realm-core and realm-cocoa, we had before. I'm not aware of any direct issues with that, but it could have potentially unexpected side-effects.

tgoyne:
There's no overlap in names since all of the binding file names are prefixed with RLM and none of the core ones are. Things may break horribly if files are moved between case-sensitive and case-insensitive file systems, though.

Issue #1752, opened 2015-04-13
Browser crashes opening uninitialized Realm


segiddins:
It should instead handle the error and show an alert.

```
Application Specific Information:
*** Terminating app due to uncaught exception 'RLMException', reason: 'Cannot open an uninitialized realm in read-only mode'
terminating with uncaught exception of type NSException
abort() called

Application Specific Backtrace 1:
0 CoreFoundation 0x00007fff896d703c __exceptionPreprocess + 172
1 libobjc.A.dylib 0x00007fff883bc76e objc_exception_throw + 43
2 Realm 0x000000010da7d85e +[RLMRealm realmWithPath:key:readOnly:inMemory:dynamic:schema:error:] + 2674
3 Realm Browser 0x000000010da0a8df -[RLMRealmNode connect:] + 95
4 Realm Browser 0x000000010d9f4839 __50-[RLMDocument initWithContentsOfURL:ofType:error:]_block_invoke + 73
5 libdispatch.dylib 0x00007fff85d64c13 _dispatch_client_callout + 8
6 libdispatch.dylib 0x00007fff85d7204e _dispatch_barrier_sync_f_slow_invoke + 412
7 libdispatch.dylib 0x00007fff85d64c13 _dispatch_client_callout + 8
8 libdispatch.dylib 0x00007fff85d70cbf _dispatch_main_queue_callback_4CF + 861
9 CoreFoundation 0x00007fff8962a3f9 __CFRUNLOOP_IS_SERVICING_THE_MAIN_DISPATCH_QUEUE__ + 9
10 CoreFoundation 0x00007fff895e568f __CFRunLoopRun + 2159
11 CoreFoundation 0x00007fff895e4bd8 CFRunLoopRunSpecific + 296
12 HIToolbox 0x00007fff89a3356f RunCurrentEventLoopInMode + 235
13 HIToolbox 0x00007fff89a332ea ReceiveNextEventCommon + 431
14 HIToolbox 0x00007fff89a3312b _BlockUntilNextEventMatchingListInModeWithFilter + 71
15 AppKit 0x00007fff8e8659bb _DPSNextEvent + 978
16 AppKit 0x00007fff8e864f68 -[NSApplication nextEventMatchingMask:untilDate:inMode:dequeue:] + 346
17 AppKit 0x00007fff8e85abf3 -[NSApplication run] + 594
18 AppKit 0x00007fff8e7d7354 NSApplicationMain + 1832
[...]

PR #1754, opened 2015-04-13
[RLMRealm] Return an error when attempting to open a read-only uninitial...

segiddins:
...ized realm

Closes https://github.com/realm/realm-cocoa/issues/1752.

\c @tgoyne @kishikawakatsumi


tgoyne:
Need to `return nil` after calling `RLMSetErrorOrThrow`.

segiddins:
> Need to return nil after calling RLMSetErrorOrThrow.

Done

tgoyne:
:+1:

Issue #1735, opened 2015-04-07
Indexing is broken


anlaital:
If you define an indexed property on an RLMObject other than its primary key by overriding ```+ [RLMObject indexedProperties]``` this will break indexing at some point in the future when objects are added, updated and deleted. 

The actual result of this is that RLMResults do not contain objects whose indices are broken if they are used in filtering clauses, such as [SomeRLMObject objectsWhere:@"someIndexedProperty = 'foo'"].

This is an extremely serious issue and you should probably communicate it with your users and tell them not to define indexed properties until this is fixed. We have now had several high-priority issues in production because of this failure.

I am also in possession of an encrypted database where indices are broken that can easily be used to reproduce this error. However, it will need some serious sanitation before I can send it to you guys. What is the best approach to clean up some highly sensitive data from the database before sending it to you?

rrrlasse:
Does any string contain 0-bytes in the indexed column?

tgoyne:
There shouldn't be any difference querying-wise between a primary key and a regular indexed column. The only differences are that PK columns are checked for uniqueness on insert and `objectForPrimaryKey:` hits a different codepath from other queries.

tgoyne:
The key part for sanitizing a Realm file is to use `writeCopyToPath:` at the end to get a new file with only the live data, as a Realm file will normally contain some old data that hasn't been overwritten yet. If the sensitive data isn't needed to reproduce the issue, one option would be to create a small program with your object schema but with all of the sensitive properties removed, open the Realm (with a no-op migration), and write a new copy.

anlaital:
@rrrlasse There are no 0-bytes in the strings, as they are IBANs.

@tgoyne I'll try to asterisk out the sensitive parts of the data and send you the database file along with the encryption key.

I should also note that the results are returned correctly when indexes are not hit, i.e. performing CONTAINS or BEGINSWITH on the indexed string column.

timanglade:
P.S. you can send us files to help@realm.io

anlaital:
I've sent you the minimal test project along with the encrypted database that's had its indices corrupted.

tgoyne:
The problem is that the matching row indexes for each value in the index are not kept in sorted order in `move_last_over` (i.e. when a row is deleted), and the querying code relies on them being sorted. Primary Keys of course only have one row for each value, so the row indexes can't be out of order.

Fixing the string index updating code shouldn't be difficult, but fixing any existing databases will require regenerating the index.

[...]

PR #1755, opened 2015-04-13
Add and remove indexes from existing columns during migrations

tgoyne:
I'm not sure if it really needs to require a migration, but making it not require a migration would require some nontrivial refactoring.

Partially addresses #1735.

@jpsim @alazier

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@ x.x.x Release notes (yyyy-MM-dd)
 * Sorting `RLMResults` is 2-5x faster (typically closer to 2x).
 * Refreshing `RLMRealm` after a write transaction which inserts or modifies
   strings or `NSData` is committed on another thread is significantly faster.
+* Indexes are now added and removed from existing properties during migrations,
+  rather than only when properties are first added.
 
 ### Bugfixes
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,8 +7,8 @@ x.x.x Release notes (yyyy-MM-dd)
 * Sorting `RLMResults` is 2-5x faster (typically closer to 2x).
 * Refreshing `RLMRealm` after a write transaction which inserts or modifies
   strings or `NSData` is committed on another thread is significantly faster.
-* Indexes are now added and removed from existing properties during migrations,
-  rather than only when properties are first added.
+* Indexes are now added and removed from existing properties when a Realm file
+  is opened, rather than only when properties are first added.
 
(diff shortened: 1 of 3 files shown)


tgoyne:
Switched to not requiring a migration/schema version bump because changing what sorts of schema changes require version bumps is dangerous (a user could update Realm, fail to test on an old database and then ship an update to their app without changing any of their code and then have it break on launch for all existing users).

The unexpected loading delay was my main concern about this, but it doesn't seem to be too bad. Adding an index to 100k random strings takes 800ms on a 4S and 200ms on an iPad Air, and removing it takes about a fifth of the time (for comparison, inserting the 100k strings takes well over ten times as long).

CHANGELOG.md
```diff
@@ -7,6 +7,8 @@ x.x.x Release notes (yyyy-MM-dd)
 * Sorting `RLMResults` is 2-5x faster (typically closer to 2x).
 * Refreshing `RLMRealm` after a write transaction which inserts or modifies
   strings or `NSData` is committed on another thread is significantly faster.
+* Indexes are now added and removed from existing properties during migrations,
```
jpsim: your latest commit makes this "upon accessing the realm" instead of "during migrations", no?

Realm/RLMObjectStore.mm
```diff
@@ -87,6 +87,34 @@ static void RLMVerifyAndAlignColumns(RLMObjectSchema *tableSchema, RLMObjectSche
...
+            if (prop.type == RLMPropertyTypeString && prop.indexed != table->has_search_index(prop.column)) {
+                if (!realm.inWriteTransaction) {
+                    [realm beginWriteTransaction];
+                    commitWriteTransaction = true;
+                }
+                if (prop.indexed) {
+                    table->add_search_index(prop.column);
+                }
+                else {
+                    table->remove_search_index(prop.column);
```
segiddins: won't this not remove the index for databases that were created back when we supported int indices?

jpsim: This. I don't know why we're checking for `prop.type == RLMPropertyTypeString` here.

tgoyne: We never released a version with int indices enabled. It lasted about a day on master before getting reverted.

tgoyne: The check is there only because the old version of the code had a similar check, but it does appear to be pointless as indexing non-string columns is rejected earlier.

jpsim:
Looks good. :+1:

PR #1758, opened 2015-04-15
Synchronize access to the accessor linked list with a spinlock

segiddins:
Another take at #1684.

\c @tgoyne @alazier

### DIFF
--- a/Realm/RLMObjectBase.mm
+++ b/Realm/RLMObjectBase.mm
@@ -183,7 +183,7 @@ const NSUInteger RLMDescriptionMaxDepth = 5;
 
 - (void)dealloc {
     if (_row) {
-        RLMReleaseRow(_objectSchema, _row);
+        RLMReleaseRow(self, _objectSchema);
     }
 }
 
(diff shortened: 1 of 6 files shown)


tgoyne:
Needs to be something like `void RLMSetRow(RLMObjectBase *object, RLMObjectSchema *objectSchema, NSUInteger index)` with the assignment to `_row` within the lock. Copying or assigning a `Row` object also causes the accessor linked-list to be modified.

tgoyne:
There's at least one more place where `_row` is assigned to (in RLMArrayLinkView fast enumeration).

There's also more operations that need to be locked. `move_last_over` also mutates the linked list (and it being able to update row accessors is the entire point of the list), as does `advance_read` when applying a transaction which removed rows. The latter case needs to lock all of the object schemas which have active table references.

astigsen:
Why are the locks needed? Are there cases where the schema is being used by
multiple threads?​

tgoyne:
It's to handle the scenario where the user captures a reference to an object that's never used on the wrong thread, but is released and deallocated on the wrong thread.

PR #1761, opened 2015-04-16
Cascade Delete Relationships

LoganWright:
Hi @realm!

I thought I'd take a shot at implementing some cascade delete relationships and see what you think.

From what I can see, the implementation is working, but that doesn't always mean it's right.  If you guys have any issues for whatever reason, let me know and I'll see if I can adjust my approach.

Basically, you override `cascadeProperties` on your realm object like so to indicate cascade relationships:

```ObjC
+@implementation SomeRealmObject : RLMObject
+ (NSArray *)cascadeProperties {
    return @[@"someRealmProperty",
             @"someOtherRealmProperty"];
}
@end
```

Best,

Logan

### DIFF
--- a/Realm.xcodeproj/project.pbxproj
+++ b/Realm.xcodeproj/project.pbxproj
@@ -201,6 +201,10 @@
 		3FDE338D19C39A87003B7DBA /* RLMSupport.swift in Sources */ = {isa = PBXBuildFile; fileRef = E88C36FF19745E5500C9963D /* RLMSupport.swift */; };
 		3FE79FF819BA6A5900780C9A /* RLMSwiftSupport.h in Headers */ = {isa = PBXBuildFile; fileRef = 3FE79FF719BA6A5900780C9A /* RLMSwiftSupport.h */; };
 		3FE79FF919BA6A5900780C9A /* RLMSwiftSupport.h in Headers */ = {isa = PBXBuildFile; fileRef = 3FE79FF719BA6A5900780C9A /* RLMSwiftSupport.h */; };
+		8095FB2C1ADF368C00D6B930 /* CascadeTest.m in Sources */ = {isa = PBXBuildFile; fileRef = 8095FB291ADF360E00D6B930 /* CascadeTest.m */; };
+		8095FB2D1ADF368D00D6B930 /* CascadeTest.m in Sources */ = {isa = PBXBuildFile; fileRef = 8095FB291ADF360E00D6B930 /* CascadeTest.m */; };
+		8095FB2E1ADF368E00D6B930 /* CascadeTest.m in Sources */ = {isa = PBXBuildFile; fileRef = 8095FB291ADF360E00D6B930 /* CascadeTest.m */; };
+		8095FB2F1ADF368F00D6B930 /* CascadeTest.m in Sources */ = {isa = PBXBuildFile; fileRef = 8095FB291ADF360E00D6B930 /* CascadeTest.m */; };
 		E81A1F851955FC9300FDED82 /* RLMAccessor.h in Headers */ = {isa = PBXBuildFile; fileRef = E81A1F631955FC9300FDED82 /* RLMAccessor.h */; };
 		E81A1F861955FC9300FDED82 /* RLMAccessor.mm in Sources */ = {isa = PBXBuildFile; fileRef = E81A1F641955FC9300FDED82 /* RLMAccessor.mm */; };
 		E81A1F891955FC9300FDED82 /* RLMArray.h in Headers */ = {isa = PBXBuildFile; fileRef = E81A1F661955FC9300FDED82 /* RLMArray.h */; settings = {ATTRIBUTES = (Public, ); }; };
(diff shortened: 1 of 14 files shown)


LoganWright:
Jenkins seems to be hitting me with an error on build check: 

>'user' is missing the Overall/Read permission

jpsim:
Hi @LoganWright, thanks for submitting this PR!

Before we get too deep in review, could you please sign our [CLA](https://github.com/realm/realm-cocoa/blob/master/CONTRIBUTING.md#cla)?

Your approach seem to work well for the simple use case where deleting an object always means all of its properties listed in `+cascadeProperties` should also be deleted. I would prefer a solution that would only delete the object in a relationship if it's not linked to by any other object in the realm. That way, we could also support a scenario where multiple objects can link to another, and the linked-to object would only be deleted once it's not used by any other.

For example:

```objc
@interface Tag : RLMObject
@property NSString *name;
@property NSDate *createdAt;
@end

@interface Post : RLMObject
@property NSString *title;
@property NSString *content;
@property RLMArray<Tag> *tags;
@end
```

So multiple posts can share some of their tags, and querying for posts with specific tags is as simple as following the links. But whenever the last post with a specific tag is deleted, so is the tag, so there are never any free-hanging tags.

We've internally discussed building a concept of "delete always", "delete if last linked" and "delete never" into relationships, but I can't recall in which scenarios a "delete always" semantic would actually be preferable over a "delete if last linked".

LoganWright:
@jpsim Thanks for getting back to me, I just signed the CLA.

That sounds like a good plan, I'm thinking of the three states you mentioned as such:

### Delete Never

This is current behavior and should probably continue to be the default.  Don't override `cascadeProperties` and the objects are not deleted.

### Delete If Last Linked

If you override `cascadeProperties` this is the default cascade implementation.  This seems like it would be the most commonly desired use case.

### Delete Always

I also can't think of a good scenario for this, so I think this would need to be specified some how.  I haven't thought through how this should go from an api standpoint yet.  Might be considered just `delete` as opposed to `cascade`.

### Additions

I started sketching out the delete if last linked code because that seemed like the next logical step.  It currently covers finding whether or not other objects share the same `parent -> child` relationship, but it doesn't cover external links to the cascade property.  ie:

```ObjC
id one;

id a = a();
a.one = one;

id aPrime = a();
aPrime.one = one;

id b = b();
b.one = one;
```

If I deleted `a` in the above example with `one` marked cascade, it would notice that `aPrime` still has a link and would not delete it.  However, it would not notice the link to `b`. Do you have a suggestion about how to find additional link objects within realm?  If not, it might require some sort of user facing api to mark alternative links.

Thanks for taking a look at some of this, I'm looking forward to any input!

 - Logan

jpsim:
Looking further into helping you figure this out, I realized that this functionality was actually added into core in v0.86.0 (around November), but we never came around to exposing it through either the cocoa or java bindings.

The main advantages to using core's implementation of this are that 1) it's fairly robust and well tested, 2) it will nullify weak links when an object is deleted, 3) the logic is cross-platform and 4) it's stored in the file format so it's more portable.

So all that should really be needed to expose this in the binding is to set the proper `LinkType` whenever we call `realm::Table()->add_column_link()` and also update the link type using `realm::Descriptor()->set_link_type()` in the same place where we add/remove search indexes.

The relevant function signatures and declarations in core are as follow:

```cpp
namespace realm {
  enum LinkType {
    link_Strong,
    link_Weak
  };

  Table {
    std::size_t add_column_link(DataType type, StringData name, Table& target, LinkType = link_Weak);
  }

  Descriptor {
    void set_link_type(std::size_t column_ndx, LinkType);
  }
}
```

You can get a `Descriptor` for a `Table` by calling `table->get_descriptor().get()`.

Here's the documentation associated with `set_link_type`:

---

There are two kinds of links, 'weak' and 'strong'. A strong link is one that implies ownership, i.e., that the origin row (parent) owns the target row (child). Simply stated, this means that when the origin row (parent) is removed, so is the target row (child). If there are multiple strong links to a target row, the origin rows share ownership, and the target row is removed when the last owner disappears. Weak links do not
imply ownership, and will be nullified or removed when the target row disappears.

To put this in precise terms; when a strong link is broken, and the target row has no other strong links to it, the target row is removed. A row that is implicitly removed in this way, is said to be *cascade-removed*. When a weak link is broken, nothing is cascade-removed.

A link is considered broken if

 - the link is nullified, removed, or replaced by a different link (Row::nullify_link(), Row::set_link(), LinkView::remove_link(), LinkView::set_link(), LinkView::clear()), or if
 - the origin row is explicitly removed (Row::move_last_over(), Table::clear()), or if
 - the origin row is cascade-removed, or if
 - the origin column is removed from the table (Table::remove_column()), or if
 - the origin table is removed from the group.

Note that a link is *not* considered broken when it is replaced by a link to the same target row. I.e., no no rows will be cascade-removed due to such an operation.

When a row is explicitly removed (such as by Table::move_last_over()), all links to it are automatically removed or nullified. For single link columns (type_Link), links to the removed row are nullified. For link
list columns (type_LinkList), links to the removed row are removed from the list.

When a row is cascade-removed there can no longer be any strong links to it, but if there are any weak links, they will be removed or nullified.

It is important to understand that this cascade-removal scheme is too simplistic to enable detection and removal of orphaned link-cycles. In this respect, it suffers from the same limitations as a reference
counting scheme generally does.

It is also important to understand, that the possible presence of a link cycle can cause a row to be cascade-removed as a consequence of being modified. This happens, for example, if two rows, A and B, have strong links to each other, and there are no other strong links to either of them. In this case, if A->B is changed to A->C, then both A and B will be cascade-removed. This can lead to obscure bugs in some applications, such as in the following case:

    table.set_link(col_ndx_1, row_ndx, ...);
    table.set_int(col_ndx_2, row_ndx, ...); // Oops, `row_ndx` may no longer refer to the same row

To be safe, applications, that may encounter cycles, are advised to adopt the following pattern:

    Row row = table[row_ndx];
    row.set_link(col_ndx_1, ...);
    if (row)
        row.set_int(col_ndx_2, ...); // Ok, because we check whether the row has disappeared

LoganWright:
Thanks for pointing that out, it's great to hear the functionality is already built in at a lower level.

Do you guys have an idea of what type of api you'd like to have for this?  I am keeping my initial thought for interface right now while I try to get the functionality working, then I'll try to make a clean interface.

In `RLMObjectStore.mm` in `RLMCreateColumn`, I added the linktype to the call to look like this:

```
realm::LinkType linktype = prop.cascade ? link_Strong : link_Weak;
prop.column = table.add_column_link(realm::DataType(prop.type), prop.name.UTF8String, *linkTable, linktype);
```

This is the only instance of `add_column_link` I found in the project, and from what I can see in the subsequent functions use the link type passed in, but the objects are not getting removed on deletion of their parent.  

You mentioned above:

> also update the link type using realm::Descriptor()->set_link_type() in the same place where we add/remove search indexes.

The only instance in the project of search indexes (based on quick search) is in the same `RLMCreateColumn` function, but this seems to correspond to non-realm objects, and I'm not sure where to set the link type.

All api aside, is there something missing that is required to establish the appropriate links.

jpsim:
Hi @LoganWright, sorry for the delay in getting back to you. Upon talking to the person who implemented `LinkType` in realm's core, it seems that the functionality is as-yet incomplete.

I'll talk to him to see if we can complete that functionality soon, because that really is the best place to have this kind of logic.

LoganWright:
Hi @jpsim 

At least I'm not crazy that the relationships weren't functioning properly!

I fully agree that this is the best place to have the logic, I'd be happy to wrap it once it's implemented in the core, but it sounds like you guys aren't quite ready yet.  Is there any sort of ETA on that (I know you guys have been backed up on Swift support, multiple processes, etc.)

Would you prefer to leave this pr open in the interim, or shall I close it?

jpsim:
We really do appreciate the effort you've put into this PR, but for the time being I think we should close this. We'll try to ping you once core support is added.

PR #1767, opened 2015-04-17
Restore initWithObjectClassName to the public headers

### LINKED ISSUE #1765: Extra argument 'objectClassName' in call
I admit, I haven't been updating Realm for a few weeks now, but since today's "pod update" + building I got this:
``Extra argument 'objectClassName' in call`` on RLMArray

Here's my code:

	class Stop: RLMObject {
		dynamic var physicalStopCode = ""
		dynamic var stopCode = ""
		dynamic var stopName = ""
		dynamic var coordinates = Coordinates()
		dynamic var connections = RLMArray(objectClassName: Connection.className())  // <---- Error occurs right here
		dynamic var distance:Double = 0.0
		override class func primaryKey() -> String! {
			return "physicalStopCode"
		}
	}

Everything was working fine before but now on each occurrence of  ``RLMArray(objectClassName: MyClassNameHere.className())`` I'm getting this error. It seems that it may have been caused by the latest realm update. If not I apologise. In any case some help would appreciated.

I forgot to add that I'm on Xcode 6.3. Building for iOS 8.3. Cleaned build folder and purged DerivedData.

tgoyne:
The podspec doesn't create `Realm.Private` yet, so trying to import it in RLMSupport.swift doesn't work very well.

Fixes #1765.

@jpsim @segiddins

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,11 @@
+0.91.3 Release notes (2015-04-17)
+=============================================================
+
+### Bugfixes
+
+* Fix `Extra argument 'objectClassName' in call` errors when building via
+  CocoaPods.
+
 0.91.2 Release notes (2015-04-16)
 =============================================================
 
--- a/Realm/RLMArray.h
+++ b/Realm/RLMArray.h
@@ -300,3 +300,7 @@
 
 @end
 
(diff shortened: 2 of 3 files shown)


segiddins:
:+1:

Issue #1747, opened 2015-04-09
Fix dynamic framework instructions


yoshyosh:
In our docs we say Drag `Realm-dynamic.framework` & `Realm-dynamic-simulator.framework` from the ios/ directory... but `Realm-dynamic` and `Realm-dynamic-simulator.framework` is no where to be found based on the folder structure mentioned above. I know JP said these instructions will be fixed when the new Swift api's launch. This will be needed for the Obj-c api's even once the RealmSwift api is released

segiddins:
This is closed by https://github.com/realm/realm.io/pull/452, I believe.

PR #1772, opened 2015-04-21
Package dynamic iOS framework

jpsim:
CI still needs to be updated. This addresses #1747 /cc @segiddins @tgoyne 

Installation instructions are in https://github.com/realm/realm.io/pull/452.

### DIFF
--- a/scripts/strip-frameworks.sh
+++ b/scripts/strip-frameworks.sh
@@ -1,6 +1,24 @@
 #!/usr/bin/env bash
 
-# This script stripped all non-valid architectures from dynamic libraries in
+################################################################################
+#
+# Copyright 2015 Realm Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+################################################################################
+
+# This script strips all non-valid architectures from dynamic libraries in
 # the application's `Frameworks` directory.
 # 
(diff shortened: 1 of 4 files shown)


scripts/strip-frameworks.sh
```diff
@@ -0,0 +1,36 @@
+#!/usr/bin/env bash
```
segiddins: should probably include our license header?

segiddins:
Is it intentional that this script will strip more than just the Realm framework?

jpsim:
Yes, it is intentional that this script will strip more than just the Realm framework.

scripts/strip-frameworks.sh
jpsim: I don't know, should it? Our other scripts don't include it, so is that an omission?

scripts/strip-frameworks.sh
```diff
@@ -0,0 +1,54 @@
...
+  if ! [[ "$(file "$file")" == *"dynamically linked shared library"* ]]; then
+    continue
+  fi
+  # Get architectures for current file
+  archs="$(lipo -info "${file}" | rev | cut -d ':' -f1 | rev)"
+  stripped=""
+  for arch in $archs; do
+    if ! [[ "${VALID_ARCHS}" == *"$arch"* ]]; then
+      # Strip non-valid architectures in-place
+      lipo -remove "$arch" -output "$file" "$file" || exit_code=1
```
segiddins: would it be better to just fail-fast here?

jpsim: I'm not sure what you mean.

mrackwitz:
```diff
-…|| exit_code=1
+…|| exit 1
```

jpsim: Ah, I see. No, I'd prefer not because this way, we will continue to strip other frameworks, which is the intention.

segiddins: But since we'll `exit 1` at the end, the build will fail anyways?

jpsim: Alright, good point. Early failure it is.

segiddins:
This looks fine to me.

tgoyne:
:+1:

PR #1774, opened 2015-04-21
RealmSwift.podspec

### LINKED ISSUE #1705: Write RealmSwift.podspec

mrackwitz:
This closes #1705 and supersedes #1510.
This is blocked by the (full public) release of CocoaPods/CocoaPods#3420.

#### Local Testing

:warning: Those instructions are only thought for development on Realm. Please wait until the release of RealmSwift before trying to consume the podspec.

```shell
# Preconditions
cd realm-cocoa
git fetch
git checkout mr-swift-podspec

# `pod --version` ~> 0.37.0.beta.1
gem install --pre cocoapods

# Setup a local private repo
mkdir -p ~/.cocoapods/bare-repos
git init --bare ~/.cocoapods/bare-repos/realm-private
pod repo add realm-private ~/.cocoapods/bare-repos/realm-private

# Change the source of Realm.podspec to point to this branch
# otherwise the changes in build.sh wouldn’t be applied
sed -ie "s/:tag => \"v#{s.version}/:branch => \"mr-swift-podspec/" Realm.podspec

# Nuke the CocoaPods downloader cache
rm -rf ~/Library/Caches/CocoaPods/**/Realm

# Push Realm.podspec with modulemap to private local repo
pod ipc spec Realm.podspec > Realm.podspec.json
pod repo push --verbose realm-private Realm.podspec.json --allow-warnings

# Lint RealmSwift.podspec
pod ipc spec RealmSwift.podspec > RealmSwift.podspec.json
pod spec lint --verbose RealmSwift.podspec.json --sources=realm-private
```

### DIFF
--- a/Realm.podspec
+++ b/Realm.podspec
@@ -38,7 +38,7 @@ Pod::Spec.new do |s|
   s.header_mappings_dir     = 'include'
   s.xcconfig                = { 'CLANG_CXX_LANGUAGE_STANDARD' => 'compiler-default',
                                 'OTHER_CPLUSPLUSFLAGS' => '-std=c++1y $(inherited)' }
-  s.preserve_paths          = %w(build.sh)
+  s.preserve_paths          = %w(build.sh Realm/*.{h,hpp})
 
   s.ios.deployment_target   = '7.0'
   s.ios.vendored_library    = 'core/librealm-ios.a'
--- a/Realm.podspec
+++ b/Realm.podspec
@@ -38,7 +38,7 @@ Pod::Spec.new do |s|
   s.header_mappings_dir     = 'include'
   s.xcconfig                = { 'CLANG_CXX_LANGUAGE_STANDARD' => 'compiler-default',
                                 'OTHER_CPLUSPLUSFLAGS' => '-std=c++1y $(inherited)' }
-  s.preserve_paths          = %w(build.sh Realm/*.{h,hpp})
+  s.preserve_paths          = %w(build.sh)
 
   s.ios.deployment_target   = '7.0'
   s.ios.vendored_library    = 'core/librealm-ios.a'
--- a/Realm.podspec
+++ b/Realm.podspec
@@ -4,7 +4,7 @@ Pod::Spec.new do |s|
   s.version                 = `sh build.sh get-version`
   s.summary                 = 'Realm is a modern data framework & database for iOS & OS X.'
   s.description             = <<-DESC
-                              The Realm database, for Objective-C.
+                              The Realm database, for Objective-C. (If you want to use Realm from Swift, see the “RealmSwift” pod.)
 
                               Realm is a mobile database: a replacement for Core Data & SQLite. You can use it on iOS & OS X. Realm is not an ORM on top SQLite: instead it uses its own persistence engine, built for simplicity (& speed). Learn more and get help at https://realm.io
                               DESC
--- a/Realm.podspec
+++ b/Realm.podspec
@@ -49,6 +49,6 @@ Pod::Spec.new do |s|
   s.subspec 'Headers' do |s|
     s.source_files          = 'include/**/*.{h,hpp}'
     s.public_header_files   = public_header_files
-    s.private_header_files  = 'include/Realm/*{ListBase,ObjectStore,Private,Dynamic}.h'
+    s.private_header_files  = 'include/Realm/*{RealmUtil,ListBase,ObjectStore,Private,Dynamic}.h'
   end
 end
--- a/Realm.podspec
+++ b/Realm.podspec
@@ -1,5 +1,6 @@
 Pod::Spec.new do |s|
   s.name                    = 'Realm'
+  s.module_map              = 'Realm/module.modulemap'
   s.version                 = `sh build.sh get-version`
   s.summary                 = 'Realm is a modern data framework & database for iOS & OS X.'
   s.description             = <<-DESC
@@ -16,25 +17,24 @@ Pod::Spec.new do |s|
   s.documentation_url       = "https://realm.io/docs/objc/#{s.version}"
   s.license                 = { :type => 'Apache 2.0', :file => 'LICENSE' }
 
-  public_header_files = 'include/Realm/RLMArray.h',
-                        'include/Realm/RLMCollection.h',
-                        'include/Realm/RLMConstants.h',
-                        'include/Realm/RLMListBase.h',
-                        'include/Realm/RLMMigration.h',
-                        'include/Realm/RLMObject.h',
-                        'include/Realm/RLMObjectBase.h',
-                        'include/Realm/RLMObjectSchema.h',
-                        'include/Realm/RLMPlatform.h',
-                        'include/Realm/RLMProperty.h',
-                        'include/Realm/RLMRealm.h',
-                        'include/Realm/RLMResults.h',
-                        'include/Realm/RLMSchema.h',
-                        'include/Realm/Realm.h'
+  public_header_files       = 'include/Realm/RLMArray.h',
+                              'include/Realm/RLMCollection.h',
+                              'include/Realm/RLMConstants.h',
+                              'include/Realm/RLMListBase.h',
+                              'include/Realm/RLMMigration.h',
+                              'include/Realm/RLMObject.h',
+                              'include/Realm/RLMObjectBase.h',
+                              'include/Realm/RLMObjectSchema.h',
+                              'include/Realm/RLMPlatform.h',
+                              'include/Realm/RLMProperty.h',
+                              'include/Realm/RLMRealm.h',
+                              'include/Realm/RLMResults.h',
+                              'include/Realm/RLMSchema.h',
+                              'include/Realm/Realm.h'
 
   s.compiler_flags          = "-DREALM_HAVE_CONFIG -DREALM_COCOA_VERSION='@\"#{s.version}\"'"
   s.prepare_command         = 'sh build.sh cocoapods-setup'
-  s.public_header_files     = public_header_files
-  s.source_files            = 'Realm/*.{h,m,mm,hpp}', 'include/**/*.hpp'
+  s.source_files            = 'Realm/*.{m,mm}'
   s.header_mappings_dir     = 'include'
   s.xcconfig                = { 'CLANG_CXX_LANGUAGE_STANDARD' => 'compiler-default',
                                 'OTHER_CPLUSPLUSFLAGS' => '-std=c++1y $(inherited)' }
@@ -47,7 +47,8 @@ Pod::Spec.new do |s|
   s.osx.vendored_library    = 'core/librealm.a'
 
   s.subspec 'Headers' do |s|
-    s.source_files          = 'include/**/*.h'
+    s.source_files          = 'include/**/*.{h,hpp}'
     s.public_header_files   = public_header_files
+    s.private_header_files  = 'include/Realm/*{ListBase,ObjectStore,Private,Dynamic}.h'
   end
 end
--- a/Realm.podspec
+++ b/Realm.podspec
@@ -1,6 +1,5 @@
 Pod::Spec.new do |s|
   s.name                    = 'Realm'
-  s.module_map              = 'Realm/module.modulemap'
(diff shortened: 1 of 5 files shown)


Realm.podspec
```diff
@@ -52,7 +52,8 @@ Pod::Spec.new do |s|
   s.osx.vendored_library    = 'core/librealm.a'
 
   s.subspec 'Headers' do |s|
-    s.source_files          = 'include/**/*.h'
+    s.source_files          = 'include/**/*.{h,hpp}'
```
tgoyne: I still would like an explanation for this change.

mrackwitz: That's more or less a limitation of CocoaPods. It's the only possible way to configure the `*.hpp` headers in the podspec to have the visibility **project**. It's not possible to set that directly. (Intentionally, to avoid conflicting configurations) Otherwise we would need to duplicate the public and private header file definitions on the root podspec, this will make the podspec more complex and cause that the headers are added twice in the project / file ref tree, which is unfortunate for users inspecting the target. /c @segiddins

tgoyne: Why do the `*.hpp` files need to have `project` visibility?  They should be private.

segiddins: @tgoyne right now in the xcodeproject they're project headers, not private

tgoyne: That's just because it's the default and there was no reason to change it.

segiddins: If they're private, that means they'll ship inside the build framework (as will all the headers in core they depend on), I believe.

Realm.podspec
```diff
@@ -21,29 +22,28 @@ Pod::Spec.new do |s|
...
   s.compiler_flags          = "-DREALM_HAVE_CONFIG -DREALM_SWIFT=0 -DREALM_COCOA_VERSION='@\"#{s.version}\"'"
   s.prepare_command         = 'sh build.sh cocoapods-setup'
-  s.public_header_files     = public_header_files
-  s.source_files            = 'Realm/*.{h,m,mm,hpp}', 'include/**/*.hpp'
+  s.source_files            = 'Realm/*.{m,mm}'
   s.header_mappings_dir     = 'include'
   s.xcconfig                = { 'CLANG_CXX_LANGUAGE_STANDARD' => 'compiler-default',
                                 'OTHER_CPLUSPLUSFLAGS' => '-std=c++1y $(inherited)' }
-  s.preserve_paths          = %w(build.sh)
+  s.preserve_paths          = %w(build.sh Realm/*.{h,hpp})
```
segiddins: Since the header files are globe from `include`, I don't see why they need to be added to the `preserve_paths`?

RealmSwift.podspec
```diff
@@ -0,0 +1,31 @@
+Pod::Spec.new do |s|
+  s.name                    = 'RealmSwift'
+  s.version                 = `sh build.sh get-version`
+  s.summary                 = 'Realm is a modern data framework & database for iOS & OSX.'
+  s.description             = <<-DESC
+                              Realm is a modern data framework & database for iOS & OSX. You can use it purely in memory — or persist to disk with extraordinary performance.
+
+                              Realm’s data structures look like NSObjects and NSArrays, but provide additional features such as: querying, relationships & graphs, thread safety, and more.
```
segiddins: this should probably be updated to be swift-specific?

RealmSwift.podspec
```diff
@@ -0,0 +1,31 @@
...
+                              DESC
+  s.homepage                = "http://realm.io"
+  s.source                  = { :git => 'https://github.com/Realm/realm-cocoa.git', :tag => "v#{s.version}" }
+  s.author                  = { 'Realm' => 'help@realm.io' }
+  s.requires_arc            = true
+  s.social_media_url        = 'https://twitter.com/realm'
+  s.documentation_url       = "http://realm.io/docs/cocoa/#{s.version}"
+  s.license                 = { :type => 'Apache 2.0', :file => 'LICENSE' }
+
+  s.compiler_flags          = "-DREALM_SWIFT=1"
```
segiddins: Is `REALM_SWIFT` actually used anywhere?

Realm.podspec
mrackwitz: That's a workaround. See 3d3c02e33905afccdea4f6d4b7aa7443e9ce915e.

RealmSwift.podspec
```diff
@@ -0,0 +1,31 @@
...
+
+  s.dependency 'Realm', "= #{s.version}"
+
+  s.source_files = 'RealmSwift/*.swift'
+
+  s.ios.deployment_target   = '8.0'
+  s.osx.deployment_target   = '10.10'
```
segiddins: Doesn't swift work on 10.9?

RealmSwift.podspec
mrackwitz: That's how I interpreted @jpsim here: https://github.com/realm/realm-cocoa/pull/1510#discussion_r25736978. But I would also prefer to not add it, before it's not in use.

Realm.podspec
segiddins: What's it working around?

mrackwitz: The issue with the download cache: the pod is cleaned before the headers are copied by the `prepare_command` into `include`, so they are deleted before they would be available there.

segiddins: Please submit a PR fixing that in CocoaPods so we don't need this here?

mrackwitz: It was still an ongoing discussion about how to approach that when I opened the PR, but yeah this commit / line is subject to change.

segiddins:
@mrackwitz can you please verify that this is still working in swift 1.2?

RealmSwift.podspec
jpsim: `REALM_SWIFT` is no longer used anywhere, so you can remove that from the podspec. I'll make a PR removing it from everywhere else.

mrackwitz: :tada:

jpsim: #1791

RealmSwift.podspec
jpsim: @mrackwitz why is this 10.10 and not 10.9?

Realm.podspec
jpsim: What's the latest news on this? Can this be fixed in CocoaPods?

RealmSwift.podspec
mrackwitz: by accident, fixed.

RealmSwift.podspec
mrackwitz: any ideas on the exact copy?

tgoyne: Maybe just "...look like regular objects and arrays, but...".

jpsim: That description was never updated since the original public release of realm-cocoa and should be reworded. I'll work with @timanglade on an update for both podspecs.

We'll also want to mention both podpspecs in each description and say when one should be used over the other.

jpsim: How about these?

**Realm.podspec**

This pod is the official Objective-C API for Realm: a modern data framework &
database for iOS & OSX. You can use it purely in memory — or persist to disk
with extraordinary performance.

If you're looking for the Swift API, use `pod "RealmSwift"`.

Realm’s data structures look like regular objects and arrays, but provide
additional features such as: querying, relationships & graphs, thread safety,
encryption, multi-process sharing and more.

Realm is not built on SQLite. Instead, a custom C++ core is used to provide
memory-efficient access to your data by using Realm objects, which usually
consume less RAM than native objects. The core also provides an optional
persistence layer that can automatically save and retrieve your objects from
disk.

Realm offers extraordinary performance compared to SQLite and other persistence
solutions, and is currently powering apps on over 100 million devices.

**RealmSwift.podspec**

RealmSwift is the official Swift API for Realm: a modern data framework &
database for iOS & OSX. You can use it purely in memory — or persist to disk
with extraordinary performance.

If you're looking for the Objective-C API, use `pod "Realm"`.

Realm’s data structures look like regular objects and arrays, but provide
additional features such as: querying, relationships & graphs, thread safety,
encryption, multi-process sharing and more.

Realm is not built on SQLite. Instead, a custom C++ core is used to provide
memory-efficient access to your data by using Realm objects, which usually
consume less RAM than native objects. The core also provides an optional
persistence layer that can automatically save and retrieve your objects from
disk.

Realm offers extraordinary performance compared to SQLite and other persistence
solutions, and is currently powering apps on over 100 million devices.

jpsim: Take 2, much shorter:

**Realm.podspec**

The Realm database, for Objective-C. (If you want to use Realm from Swift, see the “RealmSwift” pod.)

Realm is a mobile database: a replacement for Core Data & SQLite. You can use it on iOS & OSX. Realm is not an ORM on top SQLite: instead it uses its own persistence engine, built for simplicity (& speed). Learn more and get help at https://realm.io

**RealmSwift.podspec**

The Realm database, for Swift. (If you want to use Realm from Objective-C, see the “Realm” pod.)

Realm is a mobile database: a replacement for Core Data & SQLite. You can use it on iOS & OSX. Realm is not an ORM on top SQLite: instead it uses its own persistence engine, built for simplicity (& speed). Learn more and get help at https://realm.io

timanglade: :+1:

Realm.podspec
segiddins: A fix for this will be merged into CocoaPods tonight.

segiddins: Fixed in CocoaPods, @mrackwitz could you please update the podspec?

mrackwitz:
I can confirm that this works with RealmSwift 1.2. But CocoaPods 0.37 has a bug in the downloader cache, which blocks this for now.

jpsim:
Can you recommend a work-around? A link to the CocoaPods bug in question would be nice too.

mrackwitz:
The only possible workaround is disabling the cache via setting the environment var `COCOAPODS_SKIP_CACHE`. This must be also done for `pod spec lint`.

jpsim:
Having a working podspec is a pretty important requirement to releasing Realm Swift. What can be done to resolve this? I can't find the relevant CocoaPods issue.

jpsim:
> I'm working on CocoaPods/CocoaPods#3498 to fix that.

Ok, let's wait for that before releasing Realm Swift then.

segiddins:
After rebasing (and using Marius' PR branch to lint), `Realm.podspec` passes lint, but `RealmSwift.podspec` yields a warning:

    - WARN  |  <module-includes>:1:1: warning: umbrella header for module 'Realm' does not include header 'Target Support Files/Pods-Realm/Pods-Realm-umbrella.h'

Otherwise, everything works!

segiddins:
After https://github.com/CocoaPods/CocoaPods/pull/3509 and https://github.com/CocoaPods/CocoaPods/pull/3498 are merged and released, this PR is good to merge and ship.

jpsim:
:+1: great work @mrackwitz and @segiddins!

segiddins:
With https://github.com/CocoaPods/CocoaPods/releases/tag/0.37.1, I think this is good to go!

\c @mrackwitz @tgoyne

Realm.podspec
```diff
@@ -1,9 +1,10 @@
 Pod::Spec.new do |s|
   s.name                    = 'Realm'
+  s.module_map              = 'Realm/module.modulemap'
```
tgoyne: I'd prefer to have this with the other build settings stuff rather than the metadata stuff.

tgoyne:
:+1: assuming it works and stuff.

segiddins:
Just confirmed that it appears to be working!

marcvanolmen:
I just upgraded to Cocoapod 0.37.1 to try this because the manual approach didn't work. 

But I got the following error:

/Users/marcvano/dev/cameo/Cameo2/Pods/RealmSwift/RealmSwift/List.swift:21:8: error: no such module 'Realm.Private'
import Realm.Private

segiddins:
@marcvanolmen you need to also do the following in your podfile:

```ruby
pod 'Realm', git: 'https://github.com/realm/realm-cocoa.git'
```

marcvanolmen:
Ok thanks that worked also I was able to use RealmSwift now

PR #1778, opened 2015-04-22
Added `-[RLMRealm compact]`

jpsim:
Keeping this private for the time being because it's easy to misuse. I could have added checks to make sure there are no other `RLMRealm` instances, but since core will do that anyway, I opted to not duplicate that check. /cc @tgoyne @segiddins

### DIFF
--- a/Realm/RLMRealm.mm
+++ b/Realm/RLMRealm.mm
@@ -584,6 +584,48 @@ static void CheckReadWrite(RLMRealm *realm, NSString *msg=@"Cannot write to a re
     }
 }
 
+/**
+ Replaces all string columns in this Realm with a string enumeration column and compacts the
+ database file.
+ 
+ Cannot be called from a write transaction.
+
+ Compaction will not occur if other `RLMRealm` instances exist.
+ 
+ While compaction is in progress, attempts by other threads or processes to open the database will
+ wait.
+ 
+ Be warned that resource requirements for compaction is proportional to the amount of live data in
+ the database.
+ 
+ Compaction works by writing the database contents to a temporary database file and then replacing
+ the database with the temporary one. The name of the temporary file is formed by appending
+ `.tmp_compaction_space` to the name of the database.
+
+ @return YES if the compaction succeeded.
+ */
+- (BOOL)compact
+{
+    RLMCheckThread(self);
+    BOOL compactSucceeded = NO;
+    if (!_inWriteTransaction) {
+        try {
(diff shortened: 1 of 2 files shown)


Realm/RLMRealm.mm
```diff
@@ -584,6 +584,48 @@ - (void)invalidate {
...
+ 
+ Compaction works by writing the database contents to a temporary database file and then replacing
+ the database with the temporary one. The name of the temporary file is formed by appending
+ `.tmp_compaction_space` to the name of the database.
+
+ @return YES if the compaction succeeded.
+ */
+- (BOOL)compact
+{
+    RLMCheckThread(self);
```
segiddins: seems like we also need to validate that we're not invalidated?

Realm/Tests/RealmTests.mm
```diff
@@ -1172,6 +1173,24 @@ - (void)testHoldRealmAfterSourceThreadIsDestroyed {
     [realm path]; // ensure ARC releases the object after the thread has finished
 }
 
+- (void)testCompact
+{
+    RLMRealm *realm = self.realmWithTestPath;
+    [realm transactionWithBlock:^{
+        [StringObject createInRealm:realm withObject:@[@"A"]];
+        [StringObject createInRealm:realm withObject:@[@"A"]];
```
segiddins: maybe add more objects inside a loop?

segiddins:
> I could have added checks to make sure there are no other RLMRealm instances, but since core will do that anyway, I opted to not duplicate that check.

That might lead to users thinking they've compacted successfully, when it just silently failed.

Realm/Tests/RealmTests.mm
jpsim: why would we want to test *more* than the minimal case? If it compacts in this minimal case, it will in larger cases too.

tgoyne:
It returns false if it didn't compact due to the file being open on other threads.

Realm/RLMRealm.mm
tgoyne: Realms don't get invalidated (calling `invalidate` just puts it back into the state it was in before the first read).

segiddins: thats what I meant. Since we call `end_read` and `begin_read` later on, shouldn't we ensure that we're already in a read transaction?

tgoyne: Accessing the tables begins the read transaction if we're not already in one.

tgoyne:
:+1:

PR #1784, opened 2015-04-23
[Docs] re-enable jazzy in build-docs.sh because it now supports Swift 1.2

jpsim:
Jazzy 0.2.0 is installed on all CI workers.


jpsim:
device tests failed for an unrelated reason. merging.

PR #1788, opened 2015-04-23
Ignore dynamic objects from the sharedSchema

### LINKED ISSUE #1783: Don't make a table for RLMDynamicObject / DynamicObject / MigrationObject

segiddins:
Closes #1783.

\c @jpsim  @tgoyne

### DIFF
--- a/RealmSwift/Object.swift
+++ b/RealmSwift/Object.swift
@@ -260,6 +260,10 @@ public final class DynamicObject : Object {
     public override func setValue(value: AnyObject?, forUndefinedKey key: String) {
         self[key] = value
     }
+
+    @objc private class func shouldPersistToRealm() -> Bool {
+        return false;
+    }
 }
 
 /// :nodoc:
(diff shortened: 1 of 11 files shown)


RealmSwift/Object.swift
```diff
@@ -260,6 +260,10 @@ public final class DynamicObject : Object {
     public override func setValue(value: AnyObject?, forUndefinedKey key: String) {
         self[key] = value
     }
+
+    @objc private class func shouldPersistToRealm() -> Bool {
+        return false;
```
tgoyne: This looks blatantly wrong. Is it somehow not?

segiddins: Why would we ever want to save DynamicObject?

-Samuel E. Giddins

On Apr 23, 2015, at 5:32 PM, Thomas Goyne <notifications@github.com> wrote:

In RealmSwift/Object.swift:

> @@ -260,6 +260,10 @@ public final class DynamicObject : Object {
>      public override func setValue(value: AnyObject?, forUndefinedKey key: String) {
>          self[key] = value
>      }
> +
> +    @objc private class func shouldPersistToRealm() -> Bool {
> +        return false;
This looks blatantly wrong. Is it somehow not?

tgoyne: Oh, I missed that it was in DynamicObject and not Object...

jpsim:
Since we don't specify Realm for any other schema-related methods (`ignoredProperties`/`primaryKey`/`defaultPropertyValues`), I'd prefer if this one was just called `shouldPersist`. Otherwise :+1:

tgoyne:
:+1:

segiddins:
@jpsim the reason I made this method more verbose is that it's not exposed publicly and I wanted to limit the chances of a name collision.

jpsim:
:+1:

Issue #628, opened 2014-07-21
BETA! support for NULL string & binary properties


heardrwt:
This would allow for the storage of prop = [NSData data] and prop = nil;

ishaq:
hope it gets integrated soon

alazier:
We have someone actively working on this, but we need to add support for each type individually in core, so this work will take quite a bit of time to complete.

yasirmturk:
sanitizing json is a load on performance and memory

timanglade:
I know @rrrlasse is working on this in core at the moment, but it would be good to add a subsection about this + workaround to the [Current Limitations](realm.io/docs/cocoa#current-limitations) section of our docs right away. /cc @jpsim @segiddins

checkbutton:
@timanglade @jpsim Are there any updates on this issue? Missing support for null values broke our app so badly, that we had to switch back to Core Data. :-(

jpsim:
> @timanglade @jpsim Are there any updates on this issue? Missing support for null values broke our app so badly, that we had to switch back to Core Data. :-(

We hear you, null support is a top priority for us. We're adding in null support type by type in Realm's core database engine. Strings are done, and progress on other types is advancing steadily. We'll expose that in the higher level language bindings as soon as it's ready.

In the meantime, it really shouldn't be necessary to completely move away from Realm for this limitation. Our current suggested workaround should be flexible enough to support any scenario: wrap properties that should be nullable in an RLMObject subclass:

```objc
@interface NullableInt : RLMObject
@property NSInteger intValue;
@end
```

timanglade:
Let me just pile up here on what @jpsim is saying and add that yes, we’ve
had someone working on this full-time for weeks now, but not allowing any
nullable fields was such a core design decision, that it affected
everything from how we store data on disk, to how queries are handled, and
so this hasn’t been an easy fix at all. It is close to being done though,
and we’ll try to release support for the most obvious types (string and
data) asap. Our apologies again for it taking so long.

yasirmturk:
+1 @checkbutton same story

mootoh:
@jpsim, That workaround looks good but that does not allow properties to be populated properly from JSON object. Suppose we have the following JSON received from API server:

```
{
  "name": "Me",
  "age": 30
}
```

and hope to populate it in Realm:
```
@interface NullableInt : RLMObject
@property NSInteger intValue;
@end

@interface Client : RLMObject
@property NullableInt age;
@end
```

[...]

PR #1798, opened 2015-04-25
Add support for nullable NSString columns

segiddins:
This is the first step toward addressing #628.

\c @tgoyne @jpsim


Realm/RLMObjectBase.mm
```diff
@@ -65,7 +65,12 @@ - (instancetype)initWithObject:(id)value schema:(RLMSchema *)schema {
         array = RLMValidatedArrayForObjectSchema(array, _objectSchema, schema);
         NSArray *properties = _objectSchema.properties;
         for (NSUInteger i = 0; i < array.count; i++) {
-            [self setValue:array[i] forKeyPath:[properties[i] name]];
+            id propertyValue = array[i];
+            // strip out NSNull before passing values to standalone setters
+            if (propertyValue == NSNull.null) {
```
tgoyne: `if (x == NSNull.null) { x = nil; }` appears enough places that it could use a helper function.

Realm/RLMObjectSchema.mm
```diff
@@ -123,6 +123,13 @@ + (instancetype)schemaForObjectClass:(Class)objectClass {
         }
     }
 
+    for (RLMProperty *prop in schema.properties) {
+        RLMPropertyType type = prop.type;
+        if (prop.optional && type != RLMPropertyTypeObject && type != RLMPropertyTypeArray && type != RLMPropertyTypeString) {
+            @throw RLMException([NSString stringWithFormat:@"Only 'string', 'array', and 'object' properties can be made optional, and property '%@' is of type '%@'.", prop.name, RLMTypeToString(type)]);
```
tgoyne: Optional arrays aren't supported by core at the moment and I don't think there's really plans to change that.

Realm/RLMObjectStore.mm
```diff
@@ -125,9 +133,9 @@ static void RLMCreateColumn(RLMRealm *realm, realm::Table &table, RLMProperty *p
             prop.column = table.add_column_link(realm::DataType(prop.type), prop.name.UTF8String, *linkTable);
             break;
         }
-        default:
-            prop.column = table.add_column(realm::DataType(prop.type), prop.name.UTF8String);
-            break;
+        default: {
```
tgoyne: Change seems unnecessary.

Realm/RLMObjectStore.mm
```diff
@@ -495,8 +536,12 @@ void RLMAddObjectToRealm(RLMObjectBase *object, RLMRealm *realm, RLMCreationOpti
         for (NSUInteger i = 0; i < array.count; i++) {
             RLMProperty *prop = props[i];
             // skip primary key when updating since it doesn't change
```
tgoyne: Comment needs to move down.

Realm/RLMUtil.hpp
```diff
@@ -111,9 +111,9 @@ static inline NSString *RLMTypeToString(RLMPropertyType type) {
 static inline NSString * RLMStringDataToNSString(realm::StringData stringData) {
     static_assert(sizeof(NSUInteger) >= sizeof(size_t),
                   "Need runtime overflow check for size_t to NSUInteger conversion");
-    return [[NSString alloc] initWithBytes:stringData.data()
-                                    length:stringData.size()
-                                  encoding:NSUTF8StringEncoding];
+    return stringData.is_null() ? nil : [[NSString alloc] initWithBytes:stringData.data()
```
tgoyne: Line is excessively long.

Realm/RLMObjectSchema.mm
segiddins: only reason I have that as `optional` is because it's a property type whose setters will accept `nil`, though I get that's kinda weird and confusing

Realm/RLMObjectStore.mm
segiddins: which change in particular?

segiddins:
@bdash if you could give this a quick glance, that'd be great

Realm/RLMObjectSchema.mm
jpsim: This should be in a `RLMPropertyTypeIsNullable()` function so it's easy to support more types in the future.

Realm/RLMObjectStore.mm
```diff
@@ -202,10 +210,36 @@ static bool RLMRealmGetTables(RLMRealm *realm, RLMSchema *targetSchema) {
 static bool RLMPropertyHasChanged(RLMProperty *p1, RLMProperty *p2) {
     return p2 == nil
         || p1.type != p2.type
+        || p1.optional != p2.optional
         || ![p1.name isEqualToString:p2.name]
         || (p1.objectClassName != p2.objectClassName && ![p1.objectClassName isEqualToString:p2.objectClassName]);
 }
 
+static bool RLMPropertyCanBeMigrated(RLMProperty *oldProperty, RLMProperty *newProperty) {
```
jpsim: `RLMPropertyCanBeAutoMigratedToOptional`? Longer but more informative.

segiddins: I was envisioning we could add other auto-migrations, like Date -> Double, Int -> Float, etc?

jpsim: I don't think auto-updating to optional is semantically equivalent to automatically converting Date -> Double or Int -> Float.

Realm/RLMObjectStore.mm
```diff
@@ -202,10 +210,36 @@ static bool RLMRealmGetTables(RLMRealm *realm, RLMSchema *targetSchema) {
...
+        return false;
+    }
+    if (oldProperty.type == newProperty.type && !oldProperty.optional && newProperty.optional && [oldProperty.name isEqualToString:newProperty.name]) {
+        return true;
+    }
+    return false;
+}
+
+
+static void RLMMigratePropertyToProperty(RLMProperty *oldProperty, RLMProperty *newProperty, Table *table) {
```
jpsim: Wouldn't this be more `RLMCopyPropertyToProperty`?

jpsim:
My two main concerns are:

1. we need a mechanism through which to make a property optional or required in Objective-C.
2. what happens when a `String?` property is declared in Swift? (as opposed to `NSString?`)

segiddins:
You can't make a dynamic optional swift string

-Samuel E. Giddins

On May 11, 2015, at 10:10 PM, JP Simard <notifications@github.com> wrote:

My two main concerns are:

we need a mechanism through which to make a property optional or required in Objective-C.
what happens when a String? property is declared in Swift? (as opposed to NSString?)

jpsim:
Ok, well at least it'll be caught at compile-time. That's fine. we still need a way to specify optional/required properties in Objective-C.

segiddins:
This should be ready for a review so we can release a beta, but right now it depends upon https://github.com/realm/realm-core/pull/855 to pass.

Realm/RLMObjectSchema.mm
```diff
@@ -131,6 +131,17 @@ + (instancetype)schemaForObjectClass:(Class)objectClass {
         }
     }
 
+    for (RLMProperty *prop in schema.properties) {
+        RLMPropertyType type = prop.type;
+        if (prop.optional && !RLMPropertyTypeIsNullable(type)) {
+            NSString *error = [NSString stringWithFormat:@"Only 'string', 'array', and 'object' properties can be made optional, and property '%@' is of type '%@'.", prop.name, RLMTypeToString(type)];
```
tgoyne: Array properties can't be optional.

Realm/RLMObjectSchema.mm
```diff
@@ -188,6 +199,24 @@ + (NSArray *)propertiesForClass:(Class)objectClass isSwift:(bool)isSwiftClass {
         }
     }
 
+    if (NSArray *optionalProperties = [objectUtil getOptionalPropertyNames:swiftObjectInstance]) {
+        for (RLMProperty *property in propArray) {
+            property.optional = [optionalProperties containsObject:property.name] ||
+                                property.type == RLMPropertyTypeObject || property.type == RLMPropertyTypeArray; // remove if/when core supports required link columns
```
tgoyne: As above.

Realm/RLMObjectStore.mm
tgoyne: Adding the curly braces and removing the break.

tgoyne: i.e. all of it

Realm/Tests/RLMTestCase.m
```diff
@@ -29,6 +29,11 @@ + (instancetype)realmWithPath:(NSString *)path
 + (void)resetRealmState;
 @end
 
+__attribute((constructor))
```
tgoyne: Add a comment about why this is done.

tgoyne:
Looks fine for shipping a beta.

segiddins:
The only failure here is in an interprocess test exceeding its timeout... I think this is passing! 🎉

Realm/RLMObjectBase.mm
```diff
@@ -92,7 +92,8 @@ - (instancetype)initWithValue:(id)value schema:(RLMSchema *)schema {
             @throw RLMException(@"Invalid array input. Number of array elements does not match number of properties.");
         }
         for (NSUInteger i = 0; i < array.count; i++) {
-            [self setValue:RLMValidatedObjectForProperty(array[i], properties[i], schema) forKeyPath:[properties[i] name]];
+            id propertyValue = RLMValidatedObjectForProperty(array[i], properties[i], schema);
```
jpsim: is there a reason you're not calling `RLMNSNullToNil()` as part of `RLMValidatedObjectForProperty()`?

Realm/RLMObjectSchema.mm
```diff
@@ -188,6 +199,24 @@ + (NSArray *)propertiesForClass:(Class)objectClass isSwift:(bool)isSwiftClass {
         }
     }
 
+    if (NSArray *optionalProperties = [objectUtil getOptionalPropertyNames:swiftObjectInstance]) {
```
jpsim: shouldn't this only be done for Swift classes?

Realm/RLMObjectSchema.mm
```diff
@@ -188,6 +199,24 @@ + (NSArray *)propertiesForClass:(Class)objectClass isSwift:(bool)isSwiftClass {
...
+        for (RLMProperty *property in propArray) {
+            property.optional = [optionalProperties containsObject:property.name] ||
+                                property.type == RLMPropertyTypeObject; // remove if/when core supports required link columns
+        }
+    }
+    if (NSArray *requiredProperties = [objectUtil requiredPropertiesForClass:objectClass]) {
```
jpsim: Likewise, this could be limited to Objective-C classes.

examples/ios/objc/Simple/AppDelegate.m
```diff
@@ -79,8 +79,11 @@ - (BOOL)application:(UIApplication *)application didFinishLaunchingWithOptions:(
 
     [realm beginWriteTransaction];
     [realm addObject:person];
+    person.name = nil;
     [realm commitWriteTransaction];
 
+    NSLog(@"Person: %@", person);
```
jpsim: kind of strange to be printing a person with a nil name.

Realm/RLMObjectBase.mm
segiddins: because `RLMValidatedObjectForProperty` can handle `NSNull`

Realm/RLMObjectSchema.mm
segiddins: it essentially is, through the different `objectUtil` classes. I prefer this approach to switching explicitly on `isSwift`

examples/ios/objc/Simple/AppDelegate.m
segiddins: yeah... I was just so excited to print out a `nil` though!

Realm/RLMObjectSchema.mm
jpsim: Except we're already running code in a `isSwiftClass` conditional above, and you could use an `else` statement for the `requiredPropertiesForClass:` just after. You could avoid the stub implementations of `getOptionalPropertyNames` in ObjC and `requiredPropertiesForClass` in Swift.

segiddins: I'd prefer to get rid of the above conditional, and I don't mind the stub implementations, but I'm willing to change it -- we already have some stub implementations in the objc util class

jpsim: I'd prefer removing stub implementations when they're not used, and definitely think it's clearer to conditionally call the appropriate helpers. But this conversation certainly shouldn't prevent a null beta release.

tgoyne:
`RLMBinaryDataForNSData[NSData data])` results in a null `BinaryData` because `-bytes` returns 0. Doing something like `if (data && !data.length) return realm::BinaryData((__bridge void *)data, 0);` in `RLMBinaryDataForNSData` is a simple solution, but it's kinda dumb.

segiddins:
This has now been rebased to use the C++ object store

segiddins:
(I'm waiting to move over the nullability features to the c++ object store until #2094 is merged)

segiddins:
This has been feature-flagged (and thus actual null support disabled), but should be mergable to master!

segiddins:
@tgoyne @jpsim @bdash I'd like to get this reviewed and merged into master, especially since the core upgrade/file format change is no longer coupled with this getting merged.

Realm/ObjectStore/object_store.cpp
```diff
@@ -29,6 +29,34 @@
 using namespace realm;
 using namespace std;
 
+template <typename T>
+static inline
+T get_value(TableRef table, size_t row, size_t column);
```
tgoyne: I don't see any reason for these to be templates rather than just overloaded functions.

tgoyne: Sorry, that's just for set. Get does need to be templated.

Realm/RLMObject.h
```diff
@@ -250,6 +250,14 @@
  */
 + (NSArray *)ignoredProperties;
 
+/**
+ Implement to return an array of property names that should be required. These properties will not allow storing
```
tgoyne: Implement to return an array of property names that should not allow storing `nil`.

By default, all properties of a type that support storing `nil` are considered optional properties. To require that an object in a Realm always have a non-`nil` value for a property, add the name of the property to the array returned from this method.

Currently only String, Data, and Object properties support storing `nil`, and all other properties are implicitly treated as if they were required properties. Support for for types will come in the future.

Realm/RLMUtil.hpp
```diff
@@ -113,8 +126,13 @@ static inline realm::StringData RLMStringDataWithNSString(__unsafe_unretained NS
 }
 
 // Binary convertion utilities
+static inline NSData *RLMBinaryDataToNSData(realm::BinaryData binaryData) {
+    return binaryData ? [NSData dataWithBytes:binaryData.data() length:binaryData.size()] : nil;
+}
+
 static inline realm::BinaryData RLMBinaryDataForNSData(__unsafe_unretained NSData *const data) {
-    return realm::BinaryData(static_cast<const char *>(data.bytes), data.length);
+    auto bytes = static_cast<const char *>(data.bytes) ?: static_cast<char *>((__bridge void *)data);
```
tgoyne: Needs a comment saying wtf this is doing.

Realm/Tests/LinkTests.m
```diff
@@ -123,12 +123,10 @@ - (void)testInvalidLinks
...
 
-    OwnerObject *owner = [[OwnerObject alloc] init];
-    owner.name = @"Tim";
-    owner.dog = [[DogObject alloc] init];
-
+    LinkToAllTypesObject *linkObject = [[LinkToAllTypesObject alloc] init];
+    linkObject.allTypesCol = [[AllTypesObject alloc] init];
     [realm beginWriteTransaction];
-    XCTAssertThrows([realm addObject:owner], @"dogName not set on linked object");
+    XCTAssertThrows([realm addObject:linkObject], @"dogName not set on linked object");
```
tgoyne: Error message is wrong.

Realm/ObjectStore/object_store.cpp
segiddins: done

tgoyne:
:+1: 

Only CI failure is the pre-existing testNotificationsForChangesWhileSuspended one.

Realm/RLMObjectSchema.mm
```diff
@@ -132,6 +132,21 @@ + (instancetype)schemaForObjectClass:(Class)objectClass {
...
+    for (RLMProperty *prop in schema.properties) {
+        RLMPropertyType type = prop.type;
+        if (prop.optional && !RLMPropertyTypeIsNullable(type)) {
+#ifdef REALM_ENABLE_NULL
+            NSString *error = [NSString stringWithFormat:@"Only 'string', 'binary', and 'object' properties can be made optional, and property '%@' is of type '%@'.", prop.name, RLMTypeToString(type)];
+#else
+            NSString *error = [NSString stringWithFormat:@"Only 'object' properties can be made optional, and property '%@' is of type '%@'.", prop.name, RLMTypeToString(type)];
+#endif
+            if (prop.type == RLMPropertyTypeAny && isSwift) {
+                error = [error stringByAppendingString:@"\nIf this is a 'String?' property, it must be declared as 'NSString?' instead."];
```
jpsim: Just out of curiosity, does this still work with Swift 2.0?

segiddins: I'd presume so? I'll double check once this is merged and the `swift-2.0` branch is rebased

Realm/RLMUtil.hpp
```diff
@@ -100,9 +108,14 @@ static inline NSString *RLMTypeToString(RLMPropertyType type) {
 static inline NSString * RLMStringDataToNSString(realm::StringData stringData) {
     static_assert(sizeof(NSUInteger) >= sizeof(size_t),
                   "Need runtime overflow check for size_t to NSUInteger conversion");
-    return [[NSString alloc] initWithBytes:stringData.data()
-                                    length:stringData.size()
-                                  encoding:NSUTF8StringEncoding];
+    if (stringData.is_null()) {
```
jpsim: how did this even pass CI given that build.sh doesn't specify a null-enabled version of core?

Realm/RLMUtil.hpp
```diff
@@ -113,8 +126,16 @@ static inline realm::StringData RLMStringDataWithNSString(__unsafe_unretained NS
...
 // Binary convertion utilities
+static inline NSData *RLMBinaryDataToNSData(realm::BinaryData binaryData) {
+    return binaryData ? [NSData dataWithBytes:binaryData.data() length:binaryData.size()] : nil;
+}
+
 static inline realm::BinaryData RLMBinaryDataForNSData(__unsafe_unretained NSData *const data) {
-    return realm::BinaryData(static_cast<const char *>(data.bytes), data.length);
+    // this is necessary to ensure that the empty NSData isn't treated by core as the null realm::BinaryData
+    // because data.bytes == 0 when data.length == 0
+    // the casting bit ensure that we create a data with a non-null pointer
```
jpsim: ensures

Realm/RLMUtil.hpp
segiddins: because null string data has been around for over a year, apparently :)

RealmSwift/Tests/ObjectTests.swift
```diff
@@ -42,7 +42,9 @@ class ObjectTests: TestCase {
 
         let queue = dispatch_queue_create("background", DISPATCH_QUEUE_SERIAL)
         dispatch_async(queue) {
-            XCTAssertNotEqual(Realm(), persisted.realm!)
+            autoreleasepool {
```
jpsim: why is this autoreleasepool necessary where it previously wasn't?

examples/ios/objc/Simple/AppDelegate.m
jpsim: This will crash at runtime with null disabled.

RealmSwift/Tests/ObjectTests.swift
segiddins: I _believe_ it's because we are capturing `persisted.realm` in the block, but I'm not entirely sure

jpsim:
:+1:

segiddins:
🎉🎉🎂

yoshyosh:
wow nice ! :+1:

PR #1802, opened 2015-04-27
Eliminate some redundant work during migrations

tgoyne:
`RLMProperty` doesn't override `isEqual:`, so the PK table was always updated for each table with a PK even if it didn't actually change. Doesn't appear to have caused any actual problems, but it made debugging things more confusing.


segiddins:
:+1:

jpsim:
:+1:

PR #1809, opened 2015-04-29
[WIP] Move RealmSwift to Realm.xcodeproj

### LINKED ISSUE #1790: Support Carthage for RealmSwift
Because Carthage only builds framework schemes from the first Xcode project it finds in a repo, it builds the frameworks in `Realm.xcodeproj` but not `RealmSwift.xcodeproj`. There are a few ways we could support Carthage for RealmSwift:

## 1. Merge RealmSwift and Realm into a single Xcode project.

I believe this to be the simplest option and won't really have any adverse effects on our development. The biggest downsides to this are that:

1. including the RealmSwift source and Xcode project in a release zip will be more complicated
2. there will be more schemes to scroll through in Xcode

Although at this point 1) shouldn't matter too much because I think we should only be including prebuilt binaries in our release zip (see discussion here: https://github.com/Carthage/Carthage/issues/433).

## 2. Move RealmSwift into a separate repo.

Since RealmSwift is really just an alternative API to Realm.framework, and that it relies on many of the same parts of the repo, keeping both in a single repo is much simpler and better. So I don't think this is a viable option.

## 3. Add support for multiple projects to Carthage.

This was my original intention, but after thinking about it a bit, it probably wouldn't be ideal in the general case. Consider the different ways Carthage could be modified to build frameworks from multiple Xcode projects:

a. Build all framework schemes from all Xcode projects in a repo. This will build many more frameworks than are required (all development and test resources)
b. Build all framework schemes from an Xcode workspace, if present. This has the same problem as a).
c. Add the ability to specify which frameworks should be built in a Cartfile. Since Cartfile syntax should be simple, backwards-compatible, parseable and have smart defaults, I don't think this is viable either.

jpsim:
Having a single Xcode project will enable Carthage support for Realm Swift (fixes #1790).

This PR also fixes ##1771 (just needed a `LD_DYLIB_INSTALL_NAME`).

I tried setting up `RealmSwift iOS` to use a different DerivedData directory, but nothing worked well, so I have a working but non-ideal solution in `pre-ios-swift`. If you have a more elegant solution in mind, I'm all ears.

/cc @mrackwitz @tgoyne @segiddins


MarcoSero:
I tried integrating this with Carthage but xcodebuild is failing building RealmSwift.
Since I've just started using both Realm and Carthage, this might be something I'm doing wrong. 
I've uploaded xcodebuild's log [here](https://www.dropbox.com/s/n6mjfd1achu4toe/xcodebuild.log?dl=0).

jpsim:
I haven't been able to avoid or even fully work around the auto-linking issue, so I'm shelving this for the time being because there are more important tasks for the time being.

The issue is that moving `RealmSwift.framework` to `Realm.xcodeproj` makes all build artifacts share the same derived data, intermediate and built products. This means that if the static `Realm.framework` is built before `RealmSwift.framework`, it'll get picked up and attempt to link that, which is not what we want.

I was hoping to just be able to set `SYMROOT` to a different directory, but nothing has worked 100% so far.

So I'll revisit this later when I get a better idea of how to fix it.

jpsim:
Closing since we have pre-built binaries that Carthage can use.

jarsen:
@jpsim I see there is a .framework bundle in the 0.92.2 release, but when I actually `carthage update` I only find the normal Realm.framework in the iOS build folder, so it doesn't appear to be copying the pre-build binary. My cartfile matches the documentation. Am I missing a step?

jpsim:
@jarsen are you using Carthage 0.7.3? what are the exact contents of your Cartfile? Are you using `--use-binaries` when invoking `carthage` from the command line?

jarsen:
Looks like I needed to update Carthage :tired_face:. Thanks @jpsim .

PR #1810, opened 2015-04-29
[WIP] Allow reusing nested objects

### LINKED ISSUE #1362: Support object reuse/linking by primary key in createOrUpdate methods - remove RLMValidatedDictionary and RLMValidatedArray
When calling one of the createOrUpdate variants, all object properties are required to reuse existing nested objects. We should support reusing nested objects with only the primary key specified.

### LINKED ISSUE #1800: Create a circular object graph directly in one call
Hi.
I have a crash.

Info:
we have a several level structure.
Main model has several one-to-many and one-to-one relationships. Some of these models have also some relationships by their own.

Goal:
Copy the object with all its dependencies into a new realm.

Setup:
We have two realms, the default one, and the custom one.
We add the model to the default one, and then we need to copy it to the custom one.

```
RLMRealm *customRealm = [ModelClass customRealm];
[customRealm transactionWithBlock:^{
  [ModelClass createInRealm: customRealm withObject:model];
}];
```

Realm version: 0.91.1
Xcode: 6.3

Stack trace:
https://gist.github.com/OlegAnghelov/21949e98688f769825b1
Code crash:
[screenshot 2015-04-23 18 26 03](https://cloud.githubusercontent.com/assets/457094/7344836/2193d600-ecd8-11e4-8ad6-186be1b5ae16.png)

Or is there anything I am doing incorrectly in this situation? I can't find any hints in the documentation.

segiddins:
Closes #1362.
Closes #1800.

I'm just opening this up because I'm at a loss on how to refactor this to make it not so ugly, and would appreciate input from the team. This is very much a work in progress, I'd just like some design help at this point.

\c @jpsim @kishikawakatsumi @TimOliver

### DIFF
--- a/Realm/RLMAccessor.h
+++ b/Realm/RLMAccessor.h
@@ -38,7 +38,7 @@ id RLMDynamicGet(RLMObjectBase *obj, NSString *propName);
 
 // by property/column
 void RLMDynamicSet(RLMObjectBase *obj, RLMProperty *prop, id val, RLMCreationOptions options);
-void _RLMDynamicSet(RLMObjectBase *obj, RLMProperty *prop, id val, RLMCreationOptions options, NSMapTable *mapping);
+void _RLMDynamicSet(RLMObjectBase *obj, RLMProperty *prop, id val, RLMCreationOptions options, NSMapTable *valueToAccessorMapping);
 
 //
 // Class modification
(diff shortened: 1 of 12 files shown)


tgoyne:
One potentially less ugly option would be to store the persisted version of the object as an associated object on the original, but that might perform even more terribly (and might have lifetime problems?). Other than that I don't have any ideas that don't involve completely reworking object insertion.

PR #1816, opened 2015-05-02
Test that, in migrations, arrays yield DynamicObjects

### LINKED ISSUE #1331: RLMArray accessors should be of type RLMObject when using the dynamic interface
Right now RLMArray always returns accessors classes, even when using the dynamic interface during migrations. This is incorrect and could lead to a crash if a user tried to use outdated accessor methods during a migration instead of using the dynamic interface.

segiddins:
Closes #1331.

\c @jpsim


jpsim:
:+1:

tgoyne:
:+1:

PR #1829, opened 2015-05-04
Add a helper type alias for __unsafe_unretained

tgoyne:
No functional difference, but I was annoyed with how verbose `__unsafe_unretained T * const` was, and figured it'd be trivial to make it a bit better and remove the risk of forgetting the `const`. `unretained` doesn't need the `RLM` prefix because it doesn't result in an exported symbol (or a symbol at all), so it'd only clash with other things we import ourselves.

@jpsim @segiddins

### DIFF
--- a/Realm/RLMAccessor.mm
+++ b/Realm/RLMAccessor.mm
@@ -47,7 +47,7 @@ typedef NS_ENUM(char, RLMAccessorCode) {
 };
 
 // verify attached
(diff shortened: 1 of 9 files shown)


segiddins:
:+1:

jpsim:
Sure, :+1:. Although I personally would prefer having the `RLM` prefix for this too, just to visually convey that it's our own function, but I don't particularly care.

PR #1832, opened 2015-05-04
[RLMRealm] Make inWriteTransaction public

### LINKED ISSUE #1541: Make inWriteTransaction public
I have the following category on `RLMObject` in my app:

```objective-c
- (void)app_transactionWithBlock:(void(^)(RLMRealm *realm, id model))block
{
    NSAssert(block, @"Block is required.");
    if (!realm.inWriteTransaction) {
        [realm transactionWithBlock:^{
            block(realm, self);
        }];
    } else {
        block(realm, self);
    }
}
```

It would have been very handy if `inWriteTransaction` was public. Thanks!

segiddins:
Closes #1541.

\c @jpsim

### DIFF
--- a/RealmSwift/Tests/RealmTests.swift
+++ b/RealmSwift/Tests/RealmTests.swift
@@ -174,6 +174,10 @@ class RealmTests: TestCase {
             realm.cancelWrite()
             XCTAssertFalse(realm.inWriteTransaction)
         }
+
+        realm.beginWrite()
+        realm.invalidate()
+        XCTAssertFalse(realm.inWriteTransaction)
     }
 
     func testAddSingleObject() {
--- a/RealmSwift/Tests/RealmTests.swift
+++ b/RealmSwift/Tests/RealmTests.swift
@@ -163,6 +163,19 @@ class RealmTests: TestCase {
         XCTAssertEqual(Realm().objects(SwiftStringObject).count, 0)
     }
 
+    func testInWriteTransaction() {
+        let realm = Realm()
+        XCTAssertFalse(realm.inWriteTransaction)
+        realm.beginWrite()
+        XCTAssertTrue(realm.inWriteTransaction)
+        realm.cancelWrite()
+        realm.write {
+            XCTAssertTrue(realm.inWriteTransaction)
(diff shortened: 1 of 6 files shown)


RealmSwift/Tests/RealmTests.swift
```diff
@@ -163,6 +163,19 @@ class RealmTests: TestCase {
         XCTAssertEqual(Realm().objects(SwiftStringObject).count, 0)
     }
 
+    func testInWriteTransaction() {
```
jpsim: Please test `Realm().invalidate()`

jpsim:
Please add an Objective-C test.

I think the documentation should warn readers about the potential misuse or anti-patterns that this new property enables. That's the whole reason why we never made this public in the first place.

segiddins:
> I think the documentation should warn readers about the potential misuse or anti-patterns that this new property enables. That's the whole reason why we never made this public in the first place.

What exactly would those be, though?

jpsim:
Perhaps this:

```objc
/**
Indicates if this Realm is currently in a write transaction.

@warning Wrapping mutating operations in a write transaction if this property returns `NO`
         may cause a large number of write transactions to be created, which could negatively
         impact Realm's performance. Always prefer performing multiple mutations in a single
         transaction when possible.
*/
```

RealmSwift/Tests/RealmTests.swift
segiddins: done

jpsim:
:+1: CI failure is unrelated. I'll spend some time once Swift is released taming CI's flakiness.

astigsen:
Are there use cases for this method that goes beyond just being a
workaround for the lack of nested transactions? If not, we might want to
consider if we want to add methods that we know will have to be deprecated
later when the real feature is added.

tgoyne:
Mostly just bad ones, but there's clearly no chance of getting nested transactions for 1.0.

segiddins:
Updated.

jpsim:
:+1:

tgoyne:
:+1:

segiddins:
CI failure is totally unrelated.

PR #1833, opened 2015-05-04
[RLMObjectStore] Allow properly updating linked objects with primary keys

### LINKED ISSUE #1362: Support object reuse/linking by primary key in createOrUpdate methods - remove RLMValidatedDictionary and RLMValidatedArray

segiddins:
Closes #1362.
Successor to https://github.com/realm/realm-cocoa/pull/1810, but with out the cyclical object graph support.

Basically, this PR fixes the issue by fetching a persisted object when trying to update a link in `createOrUpdate`

\c @tgoyne @jpsim @bdash

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ x.x.x Release notes (yyyy-MM-dd)
   REALM_DISABLE_ENCRYPTION=YES in your application's environment variables to
   have requests to open an encrypted Realm treated as a request for an
   unencrypted Realm.
+* Linked objects are properly updated in `createOrUpdateInRealm:withValue:`.
 
 0.92.1 Release notes (2015-05-06)
 =============================================================
--- a/Realm/Tests/ObjectTests.m
+++ b/Realm/Tests/ObjectTests.m
@@ -989,32 +989,6 @@ RLM_ARRAY_TYPE(PrimaryEmployeeObject);
     [realm commitWriteTransaction];
 }
 
-//- (void)testCreateInRealmWithCircularObject
-//{
-//    RLMRealm *realm = [self realmWithTestPath];
-//
-//    CircleObject *object = [[CircleObject alloc] init];
-//    object.data = @"data";
-//    object.next = [[CircleObject alloc] initWithValue:@[@"other data", object]];
-//
-//    [realm beginWriteTransaction];
-//    [CircleObject createInRealm:realm withValue:object];
-//    [realm commitWriteTransaction];
-//}
-
-//- (void)testInitCircularObject
-//{
-//    RLMRealm *realm = [self realmWithTestPath];
-//
-//    NSMutableDictionary *value = [NSMutableDictionary dictionaryWithObject:@"data" forKey:@"data"];
-//    value[@"next"] = @{@"data": @"other data", @"next": value};
-//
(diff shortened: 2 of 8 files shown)


RealmSwift/Tests/ObjectCreationTests.swift
```diff
@@ -261,7 +261,7 @@ class ObjectCreationTests: TestCase {
         XCTAssertEqual(stringObjects.count, 1)
         let persistedObject = object.object!
 
-        XCTAssertEqual(persistedObject.intCol, 12)
+        XCTAssertEqual(persistedObject.intCol, 11)
```
tgoyne: Why should this now be the old value?

Realm/Tests/ObjectTests.m
```diff
@@ -930,6 +989,32 @@ - (void)testCreateInRealmWithMissingValue
     [realm commitWriteTransaction];
 }
 
+//- (void)testCreateInRealmWithCircularObject
```
tgoyne: Don't add commented-out stuff.

RealmSwift/Tests/ObjectCreationTests.swift
segiddins: because it's also the new value -- `standalone` is also passed into the `create` call

tgoyne: Oh, I see what it's doing now. The tests would be more understandable if testing for what happens with conflicting values was done in a separate test that tests only that.

segiddins: I don't think it's purposefully testing conflicting values -- that's why I made the comment earlier about it being semi-deterministic (it depends on the order of the column indices)

tgoyne: Then it shouldn't have conflicting values.

tgoyne:
CI failure was due to insufficient free space on the slave.

Needs changelog, but otherwise :+1:

bdash:
:+1: assuming the CI failure is still unrelated.

CHANGELOG.md
```diff
@@ -20,6 +20,10 @@ x.xx.x Release notes (yyyy-MM-dd)
 * `Int8` properties defined in Swift are now treated as integers, rather than
   booleans.
 
+### Bugfixes
+
+* Linked objects are properly updated in `createOrUpdateInRealm:withValue`.
```
jpsim: `createOrUpdateInRealm:withValue:` missing `:` at the end

Realm/Tests/ObjectTests.m
```diff
@@ -196,6 +196,29 @@ @interface DataObject : RLMObject
...
+    return @"name";
+}
+@end
+
+RLM_ARRAY_TYPE(PrimaryEmployeeObject);
+
+@interface PrimaryCompanyObject : RLMObject
+@property NSString *name;
+@property RLMArray<PrimaryEmployeeObject> *employees;
+@property PrimaryEmployeeObject *intern;
```
jpsim: lol

jpsim:
Changelog entry needs to be bumped up, but otherwise :+1:. Nice work!

PR #1840, opened 2015-05-05
[Podspec] update podspec for 0.92.0

jpsim:
This is needed to push 0.92.0 to trunk because the `documentation_url` needs to point to `https://realm.io/docs/objc/#{s.version}`. /cc @segiddins


segiddins:
:+1:

PR #1842, opened 2015-05-05
[Swift] import Foundation to make jazzy generate the proper USRs

jpsim:
This makes a number of USRs which are currently being mis-generated (`s:F10RealmSwift12migrateRealmERR`) into its proper form (`s:F10RealmSwift12migrateRealmFTSS13encryptionKeyGSqCSo6NSData__GSqCSo7NSError_ `).

This affects jazzy's URLs.

I've made an issue to fix this in jazzy as well: https://github.com/jpsim/SourceKitten/issues/48


segiddins:
I don't think every single one of those files requires the import...

RealmSwift/Aliases.swift
```diff
@@ -16,6 +16,7 @@
 //
 ////////////////////////////////////////////////////////////////////////////
 
+import Foundation
```
segiddins: no foundation is used in this file

jpsim: No, but the types being aliased are Objective-C types, which affects the USR.

PR #1851, opened 2015-05-06
Generate unique object IDs for objects in the RealmSwift Xcode project.

bdash:
The RealmSwift Xcode project started life as a copy of the Realm Xcode project,
causing many objects to have the same object IDs. This causes Xcode to crash if
you attempt to add both Xcode projects to the same workspace.

/cc @segiddins @jpsim


segiddins:
:+1:

PR #1852, opened 2015-05-06
Avoid force unwrapping in the Swift examples

jpsim:
We shouldn't be recommending that users do this.

The times in which values must not be nil, I'm using asserts instead. /cc @segiddins @tgoyne

### DIFF
--- a/examples/ios/swift/Encryption/ViewController.swift
+++ b/examples/ios/swift/Encryption/ViewController.swift
@@ -90,10 +90,7 @@ class ViewController: UIViewController {
     func getKey() -> NSData {
         // Identifier for our keychain entry - should be unique for your application
         let keychainIdentifier = "io.Realm.EncryptionExampleKey"
-        let optionalKeychainIdentifierData = keychainIdentifier.dataUsingEncoding(NSUTF8StringEncoding, allowLossyConversion: false)
-
-        assert(optionalKeychainIdentifierData != nil)
-        let keychainIdentifierData = optionalKeychainIdentifierData!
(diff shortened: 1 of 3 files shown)


segiddins:
:+1:

tgoyne:
This seems pointless. Asserting before force unwrapping doesn't give any extra safety, and doesn't even improve the error in debug builds much. If we want to have our sample code handle nils in places where they shouldn't be possible, then it needs to actually do that.

tgoyne:
Well, most of the changes are good. Just the assert-then-force-unwrap pattern is pointless.

jpsim:
Alright, I'll revert those. Sam had the same comment.

examples/ios/swift/Encryption/ViewController.swift
```diff
@@ -78,26 +74,28 @@ class ViewController: UIViewController {
...
     func log(text: String) {
-        self.textView.text = self.textView.text + text + "\n\n"
+        textView.text = textView.text + text + "\n\n"
     }
 
     func getKey() -> NSData {
         // Identifier for our keychain entry - should be unique for your application
         let keychainIdentifier = "io.Realm.EncryptionExampleKey"
-        let keychainIdentifierData = keychainIdentifier.dataUsingEncoding(NSUTF8StringEncoding, allowLossyConversion: false)
+        let optionalKeychainIdentifierData = keychainIdentifier.dataUsingEncoding(NSUTF8StringEncoding, allowLossyConversion: false)!
```
tgoyne: Rename needs to be reverted too?

jpsim: How right you are! Done.

PR #1853, opened 2015-05-06
Disable debugging encrypted realms

### LINKED ISSUE #1625: LLDB hangs when debugging an application that is using an encrypted Realm.
LLDB hangs in Xcode 6.2 when stepping (or continuing) through code in a project that has accessed an encrypted Realm in any way. The hang occurs even when you are not stepping over Realm statements. This is very easy to reproduce and happens every time, thus making debugging applications using encrypted Realms impossible.

I have also sent you a project to reproduce this issue with.

bdash:
Prevent debugging of an application using an encrypted Realm to work around
frequent LLDB hangs. Until the underlying issue is addressed you may set
REALM_DISABLE_ENCRYPTION=YES in your application's environment variables to
have requests to open an encrypted Realm treated as a request for an
unencrypted Realm.

Fixes #1625.

/cc @jpsim @tgoyne @segiddins

### DIFF
--- a/Realm/RLMRealm.h
+++ b/Realm/RLMRealm.h
@@ -102,7 +102,9 @@
  Encrypted Realms currently cannot be opened while lldb is attached to the
  process since lldb often hangs in this situation. See issue #1625 for
  further discussion. Attempting to open an encrypted Realm with lldb attached
- will result in an EXC_BAD_ACCESS.
+ will result in an EXC_BAD_ACCESS. Running your application with
+ REALM_DISABLE_ENCRYPTION=YES set in your environment will result in Realm
+ treating requests to open an encrypted Realm as requesting an unencrypted Realm.
 
  @param path        Path to the file you want the data saved in.
  @param key         64-byte key to use to encrypt the data.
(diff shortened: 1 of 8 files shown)


tgoyne:
It might be better to skip the encryption tests entirely if a debugger is attached. I've had to disable them quite a few times to be able to debug tests which are run after them. CI will still run them, and if a developer needs to run them locally they can comment out that check.

bdash:
Xcode appears to always run the tests under debugserver, even when using "xcodebuild test" like CI does.

Realm/RLMRealm.h
```diff
@@ -99,6 +99,13 @@
  The on-disk storage for encrypted Realms are encrypted using AES256+HMAC-SHA2,
  but otherwise they behave like normal persisted Realms.
 
+ Encrypted Realms currently cannot be opened while lldb is attached to the
```
tgoyne: This bit isn't true anymore (it was true when we used a signal handler), and should instead say something about debugging with encrypted realms leading to lldb hanging. A link to the bug discussing would be nice too.

bdash:
@tgoyne I revised the changes to address the issues you mentioned.

tgoyne:
:+1:

PR #1856, opened 2015-05-06
When a type error is rasied, include information about the problemati…

### LINKED ISSUE #1827: Better crash logging caused by invalid fields
Hi guys!
Thanks again for your great work.

I have a small issue. 
We got a crash report, with the following exception:
`Terminating app due to uncaught exception 'Invalid value', reason: 'object must be of type string'`

Usually when this happens while developing, we put some breakpoints in the Realm pod and see what key failed.
In this case we cannot do this.
Can you guys give the log some more information? What was the field, what was the value? The more the better...
What do you think?

bdash:
…c object in the exception.

Fixes #1827.

/cc @jpsim

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ x.x.x Release notes (yyyy-MM-dd)
 ### Enhancements
 
 * `-[RLMRealm inWriteTransaction]` is now public.
+* Exceptions raised when incorrect object types are used with predicates now contain more detailed information.
 
 ### Bugfixes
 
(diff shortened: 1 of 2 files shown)


jpsim:
There's more useful information we could add in there: "Expected object of type Employee for property employees, but received: <Person 0xdeadbeef>". And actually, wouldn't it make more sense to print the type of the value rather than the value itself?

bdash:
It's possible to print only the value's class, but the value itself seems like it could be useful information.

segiddins:
All of the values should be printed in single quotes.

segiddins:
:+1:

tgoyne:
:+1: 

changelog

PR #1866, opened 2015-05-07
[Scripts] Make strip-frameworks.sh not executable

### LINKED ISSUE #1865: Unable to submit to AppStore due to strip-frameworks.sh signature
Hi all. 
I'm, trying to Archive and submit an app that is using Realm as a dynamic framework (due to an issue with cocoapods, frameworks and extensions it's my only option right now).  

iTunes validation fails with this message: 
**The binary at path [Gulps.app/Frameworks/Realm.framework/strip-frameworks.sh] contains an invalid signature**
screen shot 2015-05-07 at 18 40 30

I'm currently using 0.92.1 with Xcode 6.3.1.

segiddins:
Potentially closes #1865.
The installation instructions will need to be updated to explicitly call `bash` for the script, though.

\c @jpsim


jpsim:
This seems fine, but I'd like us to double check validating an archive for the App Store before going any further.

segiddins:
Since @TimOliver has an app using Realm in the app store, maybe he can check?

PR #1872, opened 2015-05-08
Improve the exception messages when sorting on an invalid column name

### LINKED ISSUE #1861: Improve error message try to sort on multi-level keypaths
I've been using RLMResults as my data model for my tableview. One of my models has an optional property (date). I need to sort my RLMResults array using this property.

Here's my Goal model:

```swift
class DateObject: RLMObject {
    dynamic var date: NSDate = NSDate()
}

class Goal: RLMObject {
    dynamic var user = User()
    dynamic var name = ""
    dynamic var desc = ""
    ...
    dynamic var targetDate: DateObject?
}
```

Here's my RLMResults (without sort):

```swift
var bodyGoals:RLMResults {
        get {
            var predicate = NSPredicate(format: "user.email = %@ AND ...", user!.email, ...)   
            return Goal.objectsWithPredicate(predicate)
        }
    }
```

This is how I was trying to sort:

```swift
return Goal.objectsWithPredicate(predicate).sortedResultsUsingProperty("targetDate.date", ascending: false)
```

Which gives me the error below (which makes sense):

```swift
Terminating app due to uncaught exception 'Invalid sort column', reason: 'Column named '(null)' not found.'
```

What options do I have now? What would be the best approach to solve this? Is there something like sortedArrayUsingComparator to RLMResults?

bdash:
Improve the exception message when sorting on an invalid column name, and introduce an explicit exception indicating that sorting on key paths is not supported.

Fixes #1861.

/cc @segiddins @jpsim

### DIFF
--- a/Realm/RLMQueryUtil.mm
+++ b/Realm/RLMQueryUtil.mm
@@ -656,6 +656,7 @@ void update_query_with_predicate(NSPredicate *predicate, RLMSchema *schema,
 
 RLMProperty *RLMValidatedPropertyForSort(RLMObjectSchema *schema, NSString *propName) {
     // validate
+    RLMPrecondition(![propName containsString:@"."], @"Invalid sort property", @"Sorting on key paths is not supported.");
     RLMProperty *prop = schema[propName];
     RLMPrecondition(prop, @"Invalid sort property", @"Property named '%@' not found on object of type '%@'.", propName, schema.className);
 
--- a/Realm/RLMQueryUtil.mm
+++ b/Realm/RLMQueryUtil.mm
@@ -656,9 +656,9 @@ void update_query_with_predicate(NSPredicate *predicate, RLMSchema *schema,
 
 RLMProperty *RLMValidatedPropertyForSort(RLMObjectSchema *schema, NSString *propName) {
     // validate
-    RLMPrecondition(![propName containsString:@"."], @"Invalid sort property", @"Sorting on key paths is not supported.");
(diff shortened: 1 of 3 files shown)


Realm/RLMQueryUtil.mm
```diff
@@ -650,14 +650,15 @@ void update_query_with_predicate(NSPredicate *predicate, RLMSchema *schema,
...
         // invalid predicate type
         @throw RLMPredicateException(@"Invalid predicate",
-                                     @"Only support compound, comparison and constant predicates");
+                                     @"Only support compound, comparison, and constant predicates");
     }
 }
 
 RLMProperty *RLMValidatedPropertyForSort(RLMObjectSchema *schema, NSString *propName) {
     // validate
+    RLMPrecondition(![propName containsString:@"."], @"Invalid sort property", @"Sorting on key paths is not supported.");
```
segiddins: should probably include the property name somewhere in there

segiddins:
:+1:

Realm/RLMQueryUtil.mm
bdash: I can't see how to fit it into a sentence in a way that makes it clear that the specific key path that was passed isn't the issue. Any suggestions?

tgoyne: "Cannot sort by '%@': sorting on key paths is not supported."

Could reword the other errors similarly.

bdash: I like it! New commit pushed.

tgoyne:
:+1:

Issue #1876, opened 2015-05-08
List accessors aren't properly initialized in Swift fast enumeration


github2016a:
hi guys ,i'm using RealmSwift 0.92
i have a class
```swift
class MediaSentenceWord: Object {
   let librarywords = List<LibraryWord>()
}
```
```
var results = realm.objects(MediaSentenceWord)
for i in 0...results.count-1{
       println(libResults[i].librarywords.count) // every thing ok
}
```
but if i write this ,there is a problem:
```
for item in results{
        println(item.librarywords.count) // always 0  !!!!!
}
```

jpsim:
We can confirm that this is an issue with Realm Swift preventing relationships from properly resolving during Swift's fast enumeration. We'll have a fix out shortly. Thanks for reporting this.

PR #1878, opened 2015-05-08
Cannot retrieve relation objects in FastEnumeration

kishikawakatsumi:
Related: https://github.com/realm/realm-cocoa/issues/1876

In the loop using the FastEnumeration, it is not possible to obtain the relation objects correctly. 

**Environments:**
Realm-Cocoa 0.92.1 and master

iOS 8.3
iPhone 6 Plus
Xcode 6.3.1 (6D1002)

**Reproducible code:**
```swift
let dog1 = Dog()
dog1.name = "dog1"

let dog2 = Dog()
dog2.name = "dog2"

let person = Person()
person.name = "person1"
person.dogs.append(dog1)
person.dogs.append(dog2)

let realm = Realm()
realm.write { () -> Void in
    realm.add(person)
}

let people = realm.objects(Person)

// It is OK using first, last, or index access 
if let person = people.first {
    println("Count: \(person.dogs.count)") // 2 [OK]
}

// If using `for ... in` loop, failed to retrieve dogs objects 
for person in people {
    println("Count: \(person.dogs.count)") // 0 [NG]
}
```

### DIFF
--- a/RealmSwift/List.swift
+++ b/RealmSwift/List.swift
@@ -309,9 +309,13 @@ extension List: ExtensibleCollectionType {
 
     /// Returns a `GeneratorOf<T>` that yields successive elements in the list.
     public func generate() -> GeneratorOf<T> {
-        let base = NSFastGenerator(_rlmArray)
+        var i: UInt = 0
         return GeneratorOf<T>() {
-            return base.next() as! T?
+            if (i >= self._rlmArray.count) {
+                return .None
+            } else {
+                return self._rlmArray[i++] as? T
+            }
         }
     }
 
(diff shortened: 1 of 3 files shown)


kishikawakatsumi:
I found that it is possible to avoid the issue if do not use NSFastGenerator. But I do not know that way is best. Could you please tell me what you think?

@segiddins @jpsim @tgoyne

segiddins:
Closing in favor of https://github.com/realm/realm-cocoa/issues/1882, since it allows us to keep using the NSFastGenerator, which is a nice performance boost.

PR #1888, opened 2015-05-08
[ObjectSchemaInitialization] Allow declaring swift list properties as…

### LINKED ISSUE #1844: Using duplicate property definitions in models should throw a more descriptive exception message
I can't instantiate Realm at all, I just get this assertion:

let r = Realm() // It fails already here!

Assertion failed: (props.count == [NSSet setWithArray:[props valueForKey:@"name"]].count), function +[RLMObjectSchema schemaForObjectClass:], file /Users/realmpro/workspace/Package iOS Swift/tightdb_objc/Realm/RLMObjectSchema.mm, line 102.

segiddins:
… dynamic var

Closes #1844.

\c @tgoyne

### DIFF
--- a/Realm/RLMObjectSchema.mm
+++ b/Realm/RLMObjectSchema.mm
@@ -179,6 +179,12 @@
             Ivar ivar = class_getInstanceVariable(objectClass, propName.UTF8String);
             id value = object_getIvar(swiftObjectInstance, ivar);
             NSString *className = [value _rlmArray].objectClassName;
+            NSUInteger existing = [propArray indexOfObjectPassingTest:^BOOL(RLMProperty *obj, __unused NSUInteger idx, __unused BOOL *stop) {
+                return [obj.name isEqualToString:propName];
+            }];
+            if (existing != NSNotFound) {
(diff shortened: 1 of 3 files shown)


tgoyne:
:+1:

jpsim:
This should really have a changelog entry and the docs/examples/samples should be updated to reflect this.

Even though the old dynamic-less approach still works, we should be nudging users towards the unified dynamic approach to keep things A) easy to understand and B) easier for us to deprecate the old dynamic-less approach.

segiddins:
I'm not really comfortable with saying this is 100% supported, as the Xcode/Swift release notes never said it officially should work.

jpsim:
I've posted on the Apple developer forums to evaluate whether or not this is expected behavior: https://devforums.apple.com/thread/270405

tgoyne:
I'd somewhat prefer to wait to actively tell people to use `dynamic` until we're sure that we can switch to the less hacky dynamic-required approach.

jpsim:
Ok, let's wait before recommending a switch.

jpsim:
:+1:

Issue #1894, opened 2015-05-10
containsString: unrecognized selector sent to iOS 7


buh:
Hi!
In class `RLMQueryUtil` used selector `containsString` that appeared in iOS 8, but in your docs Prerequisites: iOS >= 7. Do you still support/test iOS 7?
Thank you!

kishikawakatsumi:
Thank you reporting it. We fixed this at https://github.com/realm/realm-cocoa/pull/1896.

PR #1896, opened 2015-05-10
Use rangeOfString: instead containsString:

kishikawakatsumi:
Ralated: https://github.com/realm/realm-cocoa/issues/1894

Use `rangeOfString:` instead `containsString:` for support iOS 7 because `containsString:` was introduced in iOS 8. 

Could you please review @segiddins @bdash?


segiddins:
The CI failure looks unrelated. 👍🏻

bdash:
:+1:

Hopefully Xcode's version of clang will pick up [`-Wpartial-availability`](http://llvm.org/viewvc/llvm-project?view=revision&revision=232750) soon so we can catch these issues at build time.

PR #1898, opened 2015-05-10
One of the tests of iOS static lib runs on iOS 7

kishikawakatsumi:
I think that it is better at least one test runs on iOS 7 so far, what do you think?
(`ios-dynamic` and `ios-swift` support only iOS 8 or later, but `ios-static` needs support iOS 7.)

@segiddins @jpsim

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -330,7 +330,7 @@ case "$COMMAND" in
 
     "test-ios-static")
         xcrealm "-scheme iOS -configuration $CONFIGURATION -sdk iphonesimulator -destination 'name=iPhone 6' test"
-        xcrealm "-scheme iOS -configuration $CONFIGURATION -sdk iphonesimulator -destination 'name=iPhone 4S' test"
+        xcrealm "-scheme iOS -configuration $CONFIGURATION -sdk iphonesimulator -destination 'name=iPhone 4S,OS=7.1' test"
         exit 0
         ;;
 
--- a/build.sh
+++ b/build.sh
@@ -44,6 +44,7 @@ command:
   test:                 tests all iOS and OS X frameworks
   test-all:             tests all iOS and OS X frameworks in both Debug and Release configurations
   test-ios-static:      tests static iOS framework on 32-bit and 64-bit simulators
+  test-ios7-static:     tests static iOS framework on 32-bit and 64-bit iOS 7 simulators
   test-ios-dynamic:     tests dynamic iOS framework on 32-bit and 64-bit simulators
   test-ios-swift:       tests RealmSwift iOS framework on 32-bit and 64-bit simulators
   test-ios-devices:     tests dynamic and Swift iOS frameworks on all attached iOS devices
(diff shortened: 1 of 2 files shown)


kishikawakatsumi:
If you accept this PR, please add iOS 7 simulator to Xcode devices in CI machine.

bdash:
Running tests for iOS 7 simulator is great, but I think it'd be better to perform separate iOS 7 and 8 testing. This would let us cover both 32- and 64-bit for each, and would make it clearer whether the failures are specific to iOS 7 or 32-bit.

kishikawakatsumi:
@bdash Agreed. Then I edited iPhone 4s (32bit) simulator tested in both iOS 7 and 8. To be exact, it is should be testing the combination of 64bit machine and iOS 7. But it can be omitted because to isolate the problem it is enough if there is the combination of the three above.

jpsim:
@kishikawakatsumi we'll have to update all CI workers to have iOS 7 simulators installed before we can merge this. In an effort to minimize the time it takes to run CI on PRs, could you extract into `ios-static` and `ios-static-ios7` and test both 32 and 64 bit simulators.

kishikawakatsumi:
> could you extract into ios-static and ios-static-ios7 and test both 32 and 64 bit simulators?

Sure!

segiddins:
retest this please

segiddins:
review this please?

Realm/Tests/MigrationTests.mm
```diff
@@ -136,6 +136,7 @@ - (void)testPerRealmMigration {
     [RLMRealm setSchemaVersion:2 forRealmAtPath:RLMTestRealmPath() withMigrationBlock:^(__unused RLMMigration *migration,
                                                                                         uint64_t oldSchemaVersion) {
         XCTAssertEqual(0U, oldSchemaVersion);
+        [RLMRealm defaultRealm];
```
jpsim: this test should pass without accessing the default realm here.

build.sh
```diff
@@ -330,7 +332,13 @@ case "$COMMAND" in
 
     "test-ios-static")
         xcrealm "-scheme iOS -configuration $CONFIGURATION -sdk iphonesimulator -destination 'name=iPhone 6' test"
-        xcrealm "-scheme iOS -configuration $CONFIGURATION -sdk iphonesimulator -destination 'name=iPhone 4S' test"
+        xcrealm "-scheme iOS -configuration $CONFIGURATION -sdk iphonesimulator -destination 'name=iPhone 4S,OS=7.1' test"
```
jpsim: extraneous `,OS=7.1`.

jpsim:
Could you please 1) fix `test-ios-static` and 2) ensure `test-ios7-static` runs and passes reliably on CI?

segiddins:
Only failure now is iOS device!

jpsim:
:+1:

PR #1908, opened 2015-05-11
[Results] Return nil for average(_:) when count == 0

### LINKED ISSUE #1902: Different return for min, max, sum, average in RealmSwift
Big kudos on the Swift release! It's really impressive and feels much more native now.

I don't understand why `min` and `max` on `Results` return optionals while `sum` and `average` returns non-optionals?

```swift
    public func max<U: MinMaxType>(property: String) -> U? {
        return rlmResults.maxOfProperty(property) as! U?
    }
    public func sum<U: AddableType>(property: String) -> U {
        return rlmResults.sumOfProperty(property) as AnyObject as! U
    }
```

The documentation for `max` states:
```swift
// The maximum value for the property amongst objects in the Results, or `nil` if the Results is empty.
```
Shouldn't this also be true for `sum` and `average`?

segiddins:
Closes #1902 

\c @tgoyne @jpsim

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,8 @@ x.x.x Release notes (yyyy-MM-dd)
 
 ### API breaking changes
 
-* None.
+* `Results.average(_:)` now returns an optional, which is `nil` if and only if the results
+  set is empty.
 
 ### Enhancements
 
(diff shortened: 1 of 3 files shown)


tgoyne:
:+1:

dbworku:
Looks good. :+1:

jpsim:
:+1:

CHANGELOG.md
```diff
@@ -3,7 +3,8 @@
 
 ### API breaking changes
 
-* None.
+* `Results.average(_:)` now returns an optional, which is `nil` if and only if the results
```
segiddins: wrong section

PR #1927, opened 2015-05-12
Revert "Add a helper type alias for __unsafe_unretained"

tgoyne:
It turns out that this doesn't actually fully work. It appears to only suppress ARC inserting retain/release calls around places where the pointer is just passed around, but not when ivars are read.

This reverts commit f43997154e49b086be3ce69035ba50d48bc3a98e.


jpsim:
:sad_trombone: but :+1:

PR #1948, opened 2015-05-15
[WIP] Cache linked objects for persisted objects

tgoyne:
This is significantly incomplete, but I wanted to make sure no one had objections to the basic idea before I spend too much time on it.

For KVO support on keypaths, I need to keep the observed linked-to objects alive for as long as the source object (i.e. when observing `foo.bar.baz`, `foo.bar` needs to continue to exist as long as `foo` does (or until it's changed)). The simplest way to do this is to just store the linked-to object in the ivar for the link property, and update it if needed in the property getter.

Outside of KVO this also gives a bit of a performance improvement when the same property is read repeatedly, and might make ignored properties work more sensibly. Currently if you do `obj.link.ignoredProperty = 5` and then immediately read `obj.link.ignoredProperty`, you'll get 0 (or whatever the default is), because you get different objects each time you read from `link`. With this change it'll work correctly, but I'm worried that there could be confusion with the related case of two different accessors for the same row, where maybe a user would expect both accessors to return the same linked object?

Any thoughts on if this matters, or any other potential problems from this?

@jpsim @bdash @segiddins

### DIFF
--- a/Realm/RLMAccessor.mm
+++ b/Realm/RLMAccessor.mm
@@ -64,6 +64,43 @@ static inline void RLMVerifyInWriteTransaction(__unsafe_unretained RLMObjectBase
     }
 }
 
+template<typename FunctionType>
+static FunctionType RLMGetImpForSelector(Class cls, SEL selector) {
+    return reinterpret_cast<FunctionType>(method_getImplementation(class_getInstanceMethod(cls, selector)));
+}
+
+// Helper for calling superclass methods outside the context of a method
+template<typename T>
+class RLMSuperImpl {
+public:
+    RLMSuperImpl() { }
+
+    RLMSuperImpl(__unsafe_unretained RLMProperty *const prop, Class superClass)
+    : getterSel(prop.getterSel)
+    , getter(RLMGetImpForSelector<T (*)(id, SEL)>(superClass, getterSel))
+    , setterSel(prop.setterSel)
+    , setter(RLMGetImpForSelector<void (*)(id, SEL, T)>(superClass, setterSel))
+    {
+    }
+
+
+    T get(__unsafe_unretained id const obj) const {
+        return getter ? getter(obj, getterSel) : T{};
+    }
+
+    void set(__unsafe_unretained id const obj, const T value) const {
+        if (setter) {
+            setter(obj, setterSel, value);
+        }
+    }
+
+private:
+    SEL getterSel;
+    T (*getter)(id, SEL) = nullptr;
+    SEL setterSel;
+    void (*setter)(id, SEL, T) = nullptr;
+};
+
 // long getter/setter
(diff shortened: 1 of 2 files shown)


segiddins:
@tgoyne any thoughts on how this will interact with `RLMAddObjectToRealm` clearing out those ivars?

tgoyne:
It relies on that to work, as otherwise it'd have to handle the case where the ivars have non-accessor objects.

segiddins:
Wouldn't this cause a retain cycle with a circular object graph?

tgoyne:
I don't think so. Given an object which links to itself, each of the objects in the chain `obj.next.next` would still be different objects backed by the same row. Each part of the chain would return the same object each time, but it's still not trying to deduplicate accessors for a row.

alazier:
Have you considered following the pattern used with RLMNotificationToken for kvo? If users hold onto a notification token which referenced the object being observed would that make these changes unnecessary?

tgoyne:
I'm adding support for standard KVO, not implementing a new API.

tgoyne:
And to clarify: this change is not *required* for KVO. It merely simplifies things a bit and I was wondering if people though it was a good idea in general.

segiddins:
I don't really love it, since it means there's different behavior for certain property types at the accessor level. That being said, I don't see any huge downsides to it.

It does seem to be breaking some tests, however, but I don't know how incidental those failures are.

alazier:
I think it might be worth considering alternative apis/modifications. Vanilla kvo can be very fragile at times and requires you to explicitly hold onto objects being observed and to explicitly remove all observers. I don't think people would complain if you improved upon this. fb has a library that makes some improvements on the basic design: https://github.com/facebook/KVOController

segiddins:
KVOController is purely a layer _on top of_ KVO -- it in no way replaces the KVO primitives

tgoyne:
> it means there's different behavior for certain property types at the accessor level

Do you just mean in terms of the implementation? API wise I don't see how it makes anything more different than it is now.

segiddins:
Yes, I mean in terms of the implementation. Again, it's fine.

alazier:
You are missing the point of my last comment @segiddins 

To restate - there are flaws and major inconveniences when using kvo as evidenced by the need for more reliable/easy to use apis built on top. If we can just provide a better set of apis which solve the common issues when using kvo then I think that would be a s sufficient if not a better approach.

In fact, I don't think it's necessary to support kvo for multilevel keypaths at all, and would rather see us only support observing of persisted properties on RLMObjects. This would obviate the need to hold onto accessors internally for multi-level keypaths and I imagine would simplify other aspects of the implementation.

In general I would rather see progress towards supporting changeset notifications which is the overwhelming top requested features rather than perfecting multilevel keypath kvo which no users if any have asked for afaik.

tgoyne:
I wrote some tests for this and concluded that it introduces weirdly inconsistent behavior that can't be made consistent without running into issues with ref cycles.

PR #1961, opened 2015-05-19
Add a less slow implementation of indexOfObjectWithPredicate

tgoyne:
I don't remember what prompted this, but the commit's been sitting around unmerged for a few months and it seems mildly useful.

### DIFF
--- a/Realm/RLMArray.mm
+++ b/Realm/RLMArray.mm
@@ -175,6 +175,12 @@ static void RLMValidateMatchingObjectType(RLMArray *array, RLMObject *object) {
     [_backingArray setValue:value forKey:key];
 }
 
+- (NSUInteger)indexOfObjectWithPredicate:(__unused NSPredicate *)predicate {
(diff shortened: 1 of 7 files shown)


Realm/RLMArray.mm
```diff
@@ -176,6 +176,12 @@ - (void)setValue:(id)value forKey:(NSString *)key {
     [_backingArray setValue:value forKey:key];
 }
 
+- (NSUInteger)indexOfObjectWithPredicate:(__unused NSPredicate *)predicate {
```
bdash: Why __unused?

bdash:
The style guidelines indicate that the opening brace for methods should be on its own line. The new methods in this patch don't follow this, but neither do a lot of the existing code.

:+1:

segiddins:
Seems some RealmSwift tests need to be updated? Otherwise, :+1:

bmunkholm:
The style guideline must be followed. If other parts of the code isn't then it just has to be fixed. (we should have a styleguide linter... )

tgoyne:
I'd prefer to just update the style guidelines to match the codebase over reformatting everything.

Issue #1965, opened 2015-05-20
Unable to inherit from RLMRealm


cfeckardt:
As a developer I am unable to subclass RLMRealm because the convenience methods call [RLMRealm alloc] instead of [self alloc], causing the initializer to instantiate the base class.

segiddins:
Why are you trying to subclass RLMRealm? It's not really intended to be subclassed.

cfeckardt:
I'm looking to hook into the lifecycle of RLMObjects and I want to trigger specific events as individual objects change in the realm, in a more granual way than RLMNotification currently allows me to do.

segiddins:
@cfeckardt a lot of Realm's internals are implemented in terms of functions not inside `RLMRealm`, so even overriding `RLMRealm` methods wouldn't guarantee that you'd be able to hook into Realm in the places you'd like. We're actually working on implementing fine-grained notifications at the moment, you can track that at https://github.com/realm/realm-cocoa/issues/601.

cfeckardt:
There are a number of reasons why one might want to hook in to the public methods that I could think of. Running validations on objects before they are written to the database comes to mind. This patch would make that possible by inheriting from RLMObject (setValue) and RLMRealm (commitWriteTransaction)

segiddins:
@cfeckardt I understand the desire to hook into Realm objects' life cycle events, but we believe the best way to do so will be when we expose fine-grained notifications that can be used a building blocks for things like this. Since `RLMRealm` (and many parts of `RLMObject`) are not built to be subclassed, you'd run into lots of _really_ weird behavior where your overridden methods would never be called when you expected them to be, since those methods are really just an interface to a lot of Realm's underlying functionality. We'd also run into issues with our caching of realm instances, since we could be returning an instance of the wrong class.

I hope this explains a bit why this isn't something we plan on supporting directly, but if there are specific use cases you're interested in, let us know and maybe we can thing of other ways of implementing them!

cfeckardt:
Consider this use-case:

```objective-c
// In my RLMRealmSubclass
- (void) addObject:(RLMObject *) obj {
  if([obj isKindOfClass:[RLMObjectSubclass class]]) {
    [self.pendingChanges addObj:object];
  }
  [super addObject:obj];
}
// Others ...

- (void) commitWriteTransaction {
  BOOL valid = YES;
  for(RLMObjectSubclass x in pendingChanges) {
    valid &= [x valid];
  }

  if(valid) {
[...]

PR #1966, opened 2015-05-20
Call subclass alloc, instead of RLMRealm alloc

cfeckardt:
Instantiate a subclass of RLMRealm if it exists instead of RLMRealm w…hen using convenience constructors.

Solves #1965


segiddins:
As we discussed in https://github.com/realm/realm-cocoa/issues/1965, this  isn't something we're quite ready to do yet, so I'm going to close this PR and we can continue the discussion on that issue -- thanks for submitting it, though!

PR #1967, opened 2015-05-20
Give RLMNotVersioned a consistent value in 32- and 64-bit

### LINKED ISSUE #1957: RLMNotVersioned has different values in 32- and 64-bit
`RLMNotVersioned` is initialized like so:

```objectivec
    const NSUInteger RLMNotVersioned = (NSUInteger)-1;
```
This results in `RLMNotVersioned` having the value 4294967295 in 32-bit, but 18446744073709551615 in 64-bit. This could result in incorrect behavior when moving a Realm that was created on 32-bit device to a 64-bit device.

bdash:
Fixes #1957.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,19 @@
+x.x.x Release notes (yyyy-mm-DD)
+=============================================================
+
+### API breaking changes
+
+* Schema versions are now represented as uint64_t (Objective-C) and UInt64 (Swift) so that they have
+  the same representation on all architectures.
+
+### Enhancements
+
+* None.
+
+### Bugfixes
+
+* None.
+
 0.92.3 Release notes (2015-05-13)
 =============================================================
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,7 @@ x.x.x Release notes (yyyy-mm-DD)
 
 ### API breaking changes
 
-* Schema versions are now represented as uint64_t (Objective-C) and UInt64 (Swift) so that they have
+* Schema versions are now represented as `uint64_t` (Objective-C) and `UInt64` (Swift) so that they have
   the same representation on all architectures.
 
 ### Enhancements
--- a/Realm/RLMSchema.mm
+++ b/Realm/RLMSchema.mm
@@ -163,15 +163,15 @@ static NSMutableDictionary *s_localNameToClass;
     return schema;
 }
 
-NSUInteger RLMRealmSchemaVersion(RLMRealm *realm) {
+uint64_t RLMRealmSchemaVersion(RLMRealm *realm) {
     realm::TableRef table = realm.group->get_table(c_metadataTableName);
     if (!table || table->get_column_count() == 0) {
         return RLMNotVersioned;
     }
-    return NSUInteger(table->get_int(c_versionColumnIndex, 0));
+    return table->get_int(c_versionColumnIndex, 0);
 }
 
-void RLMRealmSetSchemaVersion(RLMRealm *realm, NSUInteger version) {
+void RLMRealmSetSchemaVersion(RLMRealm *realm, uint64_t version) {
     realm::TableRef table = realm.group->get_or_add_table(c_metadataTableName);
     table->set_int(c_versionColumnIndex, 0, version);
 }
--- a/Realm/RLMSchema.mm
+++ b/Realm/RLMSchema.mm
@@ -39,8 +39,7 @@ const size_t c_primaryKeyObjectClassColumnIndex =  0;
 const char * const c_primaryKeyPropertyNameColumnName = "pk_property";
 const size_t c_primaryKeyPropertyNameColumnIndex =  1;
 
-const NSUInteger RLMNotVersioned = (NSUInteger)-1;
-
+const uint64_t RLMNotVersioned = std::numeric_limits<uint64_t>::max();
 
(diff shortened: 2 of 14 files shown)


Realm/RLMSchema.mm
```diff
@@ -163,15 +162,24 @@ + (instancetype)dynamicSchemaFromRealm:(RLMRealm *)realm {
...
     realm::TableRef table = realm.group->get_table(c_metadataTableName);
     if (!table || table->get_column_count() == 0) {
         return RLMNotVersioned;
     }
-    return NSUInteger(table->get_int(c_versionColumnIndex, 0));
+
+    uint64_t version = table->get_int(c_versionColumnIndex, 0);
+
+    const uint64_t old64BitRLMNotVersioned = std::numeric_limits<uint64_t>::max();
+    if (version == old64BitRLMNotVersioned) {
```
tgoyne: This can't actually happen (barring a *very* strange bug, the schema version is always set to an actual version before the initial transaction is committed).

bdash:
Updated to address @tgoyne's feedback.

/cc @simonask

CHANGELOG.md
```diff
@@ -1,3 +1,19 @@
+x.x.x Release notes (yyyy-mm-DD)
+=============================================================
+
+### API breaking changes
+
+* Schema versions are now represented as uint64_t (Objective-C) and UInt64 (Swift) so that they have
```
segiddins: Backticks around type names

simonask:
If the actual value in `RLMNotVersioned` is never actually persisted in the Realm file, this should be good. Otherwise, we should probably also check for the old 32-bit value. Theoretically, you could end up with an app that crashed before getting to the initial schema-creating transaction, but after persisting `0xffffffff` as the schema version number, which then has Realm updated with this change, and then suddenly thinks that 2^32-1 is a valid version number.

I'm unsure if that scenario is actually realistic, or if it's something we want to even care about maintaining pre-1.0. :-)

tgoyne:
An app which crashes before committing the initial transaction will not have persisted a schema version at all, because that's set during the initial transaction.

kspangsege:
@tgoyne, it seems to me that the value of `RLMNotVersioned` is persisted the the Realm file. If it was not, then it would not have have caused a problem for us in connection with sync. As a temporary workaround, we compare changesets from the initial schema creating transactions, and can only proceed if they are strict equal. Before @simonask fix, they were not equal accoss a 32/64 bit barrier.

What did I miss?

tgoyne:
The schema version is set to RLMNotVersioned, the tables are created, and then the schema version is set to the actual version all within a single write transaction. It shows up in the transaction log, but a write transaction is never committed while the version is RLMNotVersioned.

kspangsege:
Ah, alright :-)

CHANGELOG.md
jpsim: yes please.

Realm/Tests/MigrationTests.mm
```diff
@@ -815,11 +815,21 @@ - (void)testEnumeratedObjectsDuringMigration {
...
                 forRealmAtPath:RLMTestRealmPath()
-            withMigrationBlock:^(__unused RLMMigration *migration, __unused NSUInteger oldSchemaVersion) {
+            withMigrationBlock:^(__unused RLMMigration *migration, __unused uint64_t oldSchemaVersion) {
                 XCTFail(@"Migration block should not have been called");
             }];
     XCTAssertNoThrow([self realmWithTestPath]);
 }
 
-@end
+- (void)testRLMNotVersionedHasCorrectValue
```
jpsim: What are we actually testing here?

RealmSwift/Tests/TestCase.swift
```diff
@@ -53,6 +53,10 @@ class TestCase: RLMAutoreleasePoolTestCase {
         RLMAssertThrows(self, { _ = block() } as dispatch_block_t, message, fileName, lineNumber);
     }
 
+    func assertNil<T>(@autoclosure block: () -> T?, _ message: String? = nil, fileName: String = __FILE__, lineNumber: UInt = __LINE__) {
```
jpsim: How is this different than `XCTAssertNil`?

RealmSwift/Tests/TestCase.swift
```diff
@@ -53,6 +53,10 @@ class TestCase: RLMAutoreleasePoolTestCase {
...
+        XCTAssert(block() == nil, message ?? "", file: fileName, line: lineNumber);
```
jpsim: we've been avoiding semicolons in Swift so far.

jpsim:
Other than my feedback, :+1:.

RealmSwift/Tests/TestCase.swift
bdash: `XCTAssertNil` requires the expression be of type `AnyObject?`. `UInt64?` isn't.

Realm/Tests/MigrationTests.mm
bdash: That `RLMNotVersioned` has a fixed, consistent value on all platforms that run the test.

RealmSwift/Tests/TestCase.swift
jpsim: I think `XCTAssert(value == nil)` would be simpler instead, but it's your call.

tgoyne:
:+1:

PR #1968, opened 2015-05-21
Remove all objects optimization

mamaral:
It appears that the implementation of many `RLMArray` methods such as `objectAtIndex:`  and `removeObjectAtIndex:` simply call those methods on the underlying `NSMutableArray`, but I noticed `RLMArray`'s implementation of `removeAllObjects` didn't take advantage of this and instead calls a series of methods essentially iterating through the array and removing the last object until it's empty. I *presume* calling `NSMutableArray`'s  `removeAllObjects` directly, rather than the loop/if-statement/method-calls, would be more efficient, unless I'm missing something?

### DIFF
--- a/Realm/RLMArray.mm
+++ b/Realm/RLMArray.mm
@@ -83,7 +83,8 @@
     }
 }
 
-- (void)removeAllObjects {
+- (void)removeAllObjects
+{
     [_backingArray removeAllObjects];
 }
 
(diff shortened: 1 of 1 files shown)


tgoyne:
Thanks! I'm not sure why the CI job failed (since the server is offline at the moment), but `verify` passes for me locally with both a 32-bit and 64-bit devices and I'm not sure how it could break...

mamaral:
Awesome!

PR #1969, opened 2015-05-21
Don't copy data or use standalone objects during validation and object creation

alazier:
These changes are needed for the JS binding as it requires support for object creation from containers when using the dynamic interface.

We now no longer used standalone objects when creating persisted objects. We do validation of each property as objects are being created which means we no longer need to do an extra allocation and initialization of standalone objects when creating nested objects. So this should speed things up in theory.

A number of small bugs were fixed and the tests were updated accordingly. There was one issue where objects would be updated even without using one of the update apis. update and standalone promotion flags are now propagated to all nested objects even through objects without a primary key.

### DIFF
--- a/Realm/RLMAccessor.mm
+++ b/Realm/RLMAccessor.mm
@@ -654,7 +654,7 @@ void RLMDynamicValidatedSet(RLMObjectBase *obj, NSString *propName, id val) {
 }
 
 void RLMDynamicSet(__unsafe_unretained RLMObjectBase *const obj, __unsafe_unretained RLMProperty *const prop,
-                   __unsafe_unretained id val, RLMCreationOptions creationOptions) {
+                   __unsafe_unretained id const val, RLMCreationOptions creationOptions) {
     NSUInteger col = prop.column;
     switch (accessorCodeForType(prop.objcType, prop.type)) {
         case RLMAccessorCodeByte:
--- a/Realm/RLMAccessor.mm
+++ b/Realm/RLMAccessor.mm
@@ -193,19 +193,20 @@ static inline RLMObjectBase *RLMAddLinkedObjectOrLiteral(__unsafe_unretained RLM
         @throw RLMException(@"Adding a deleted or invalidated object to a Realm is not permitted");
     }
 
-    if (!link->_realm && (creationOptions & RLMCreationOptionsPromoteStandalone)) {
-        // standalone
-        RLMAddObjectToRealm(link, realm, creationOptions & RLMCreationOptionsCreateOrUpdate);
+    if (link->_realm == realm) {
         return link;
     }
-    if (link->_realm != realm) {
-        // copy from another realm or un-promoted standalone
-        return RLMCreateObjectInRealmWithValue(realm, className, link, creationOptions & RLMCreationOptionsCreateOrUpdate);
+
+    if (creationOptions & RLMCreationOptionsPromoteStandalone) {
+        if (!link->_realm) {
+            RLMAddObjectToRealm(link, realm, creationOptions & RLMCreationOptionsCreateOrUpdate);
+            return link;
+        }
+        @throw RLMException(@"Can not add objects from a differnt Realm");
     }
-    else {
-        // already persisted in this realm
-        return link;
-    }
+
+    // copy from another realm or copy from standalone
+    return RLMCreateObjectInRealmWithValue(realm, className, link, creationOptions & RLMCreationOptionsCreateOrUpdate);
 }
 
 // link getter/setter
--- a/Realm/RLMAccessor.mm
+++ b/Realm/RLMAccessor.mm
@@ -202,7 +202,7 @@ static inline RLMObjectBase *RLMGetLinkedObjectForValue(__unsafe_unretained RLMR
             RLMAddObjectToRealm(link, realm, creationOptions & RLMCreationOptionsCreateOrUpdate);
             return link;
         }
-        @throw RLMException(@"Can not add objects from a differnt Realm");
+        @throw RLMException(@"Can not add objects from a different Realm");
     }
 
     // copy from another realm or copy from standalone
--- a/Realm/RLMAccessor.mm
+++ b/Realm/RLMAccessor.mm
@@ -179,19 +179,44 @@ static inline void RLMSetValue(__unsafe_unretained RLMObjectBase *const obj, NSU
     }
 }
 
-static inline size_t RLMAddLinkedObject(RLMObjectBase *link,
-                                        __unsafe_unretained RLMRealm *const realm,
-                                        bool createOrUpdate) {
+static inline RLMObjectBase *RLMAddLinkedObjectOrLiteral(__unsafe_unretained RLMRealm *const realm,
+                                                         NSString *className,
+                                                         id obj,
+                                                         bool createOrUpdate,
+                                                         bool promoteStandalone) {
+    RLMObjectBase *link = RLMDynamicCast<RLMObjectBase>(obj);
+    if (!link || ![link->_objectSchema.className isEqualToString:className]) {
+        // literal
+        return RLMCreateObjectInRealmWithValue(realm, className, obj, createOrUpdate);
+    }
+    if (!link->_realm && promoteStandalone) {
+        // standalone
+        RLMAddObjectToRealm(link, realm, createOrUpdate);
+        return link;
+    }
+    if (link->_realm != realm) {
+        // copy from another realm or un-promoted standalone
+        return RLMCreateObjectInRealmWithValue(realm, className, link, createOrUpdate);
+    }
+    else {
+        // already persisted in this realm
+        return link;
+    }
+}
+
+static inline size_t RLMCopyOrPromoteLinkedObject(__unsafe_unretained RLMRealm *const realm,
+                                                  RLMObjectBase *link) {
     if (link.isInvalidated) {
         @throw RLMException(@"Adding a deleted or invalidated object to a Realm is not permitted");
     }
-    else if (!link->_realm) {
+
+    if (!link->_realm) {
         // standalone
-        RLMAddObjectToRealm(link, realm, createOrUpdate);
+        RLMAddObjectToRealm(link, realm, false);
     }
     else if (link->_realm != realm) {
         // copy from another realm
-        link = RLMCreateObjectInRealmWithValue(realm, link->_objectSchema.className, link, createOrUpdate);
+        link = RLMCreateObjectInRealmWithValue(realm, link->_objectSchema.className, link, false);
     }
     return link->_row.get_index();
 }
@@ -208,11 +233,10 @@ static inline RLMObjectBase *RLMGetLink(__unsafe_unretained RLMObjectBase *const
 }
 
 static inline void RLMSetValue(__unsafe_unretained RLMObjectBase *const obj, NSUInteger colIndex,
-                               __unsafe_unretained RLMObjectBase *const val, bool createOrUpdate = false) {
+                               __unsafe_unretained RLMObjectBase *const val) {
     RLMVerifyInWriteTransaction(obj);
 
-    if (!val || (id)val == NSNull.null) {
-        // if null
+    if (!val) {
         obj->_row.nullify_link(colIndex);
     }
     else {
@@ -224,8 +248,7 @@ static inline void RLMSetValue(__unsafe_unretained RLMObjectBase *const obj, NSU
                                 valSchema.className, [objSchema.properties[colIndex] objectClassName]];
             @throw RLMException(reason);
         }
-
-        obj->_row.set_link(colIndex, RLMAddLinkedObject(val, obj->_realm, createOrUpdate));
+        obj->_row.set_link(colIndex, RLMCopyOrPromoteLinkedObject(obj->_realm, val));
     }
 }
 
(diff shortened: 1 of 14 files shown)


bdash:
IMO it'd be preferable to keep using an enum rather than a bool, even though you've eliminated most of the members. `RLMCreationOptionsUpdateOrCreate` is much more descriptive at a call site than `true`.

alazier:
I agree having `RLMCreationOptions` is better than a bool. My main issue with this was that the options we had were confusing and were set and mutated in odd ways - removing them made cleaning up the code a lot easier. I'm happy to add this back but we will only support two options `RLMCreationOptionsUpdateOrCreate` and `RLMCreationOptionsPromoteStandalone`.

segiddins:
I personally don't like referring to `literals`, since all we care about those objects is that they're KVC-compliant

alazier:
Not quite true. We also support array literals.

bdash:
Literals doesn't seem like the appropriate term. At the point we're dealing with the objects its irrelevant whether they were literals or created in some other fashion.

segiddins:
Since Jenkins doesn't seem to have run on this, I'm also concerned that this might break our recent improvements with object re-use.

alazier:
All the tests pass. I'm not sure what the goal of the object re-use pr was but we should be moving away from creating standalone objects internally, except in the case where we are actually initializing standalone objects. This pr accomplishes that in all cases where we are creating persisted objects.

Feel free to suggest alternate terminology. I can't think of anything better offhand.

My main goal here is unblock the JS work. The current functionality in this pr does this and also fixes a few other bugs with the current code. Don't think I am going to spend much more time on this so it would be helpful to figure out what else needs to be done to get this merged.

Realm/RLMAccessor.mm
```diff
@@ -179,25 +179,34 @@ static inline void RLMSetValue(__unsafe_unretained RLMObjectBase *const obj, NSU
     }
 }
 
-static inline size_t RLMAddLinkedObject(RLMObjectBase *link,
-                                        __unsafe_unretained RLMRealm *const realm,
-                                        RLMCreationOptions options) {
+static inline RLMObjectBase *RLMAddLinkedObjectOrLiteral(__unsafe_unretained RLMRealm *const realm,
+                                                         NSString *className,
```
segiddins: pretty sure this can be made `__unsafe_unretained`

Realm/RLMObjectBase.mm
```diff
@@ -58,6 +59,73 @@ - (instancetype)init {
...
+            RLMObjectSchema *objSchema = schema[prop.objectClassName];
+            RLMArray *objects = [[RLMArray alloc] initWithObjectClassName: objSchema.className standalone:YES];
+            for (id el in obj) {
+                [objects addObject:RLMValidatedRealmObject(el, schema, objSchema)];
+            }
+            return objects;
+        }
+
+        // if not a literal throw
+        NSString *message = [NSString stringWithFormat:@"Invalid value '%@' for property '%@'", obj ?: @"nil", prop.name];
```
segiddins: no need for the `?:` here

tgoyne:
"Literal" is dumb and wrong but in lieu of a similarly concise term I'd prefer to keep using it over "KVC compliant object or array with values in the correct order".

Realm/RLMAccessor.mm
```diff
@@ -179,25 +179,34 @@ static inline void RLMSetValue(__unsafe_unretained RLMObjectBase *const obj, NSU
...
+    if (link->_realm == realm) {
+        return link;
+    }
+
+    if (creationOptions & RLMCreationOptionsPromoteStandalone) {
+        if (!link->_realm) {
+            RLMAddObjectToRealm(link, realm, creationOptions & RLMCreationOptionsCreateOrUpdate);
+            return link;
         }
+        @throw RLMException(@"Can not add objects from a differnt Realm");
```
segiddins: `different`

Realm/RLMObjectStore.mm
```diff
@@ -463,8 +468,91 @@ void RLMAddObjectToRealm(RLMObjectBase *object, RLMRealm *realm, RLMCreationOpti
     RLMInitializeSwiftListAccessor(object);
 }
 
-RLMObjectBase *RLMCreateObjectInRealmWithValue(RLMRealm *realm, NSString *className, id value, RLMCreationOptions options) {
-    if (options & RLMCreationOptionsUpdateOrCreate && RLMIsObjectSubclass([value class])) {
+static void RLMValidateNestedObject(id obj, NSString *className, RLMSchema *schema, bool validateNested, bool allowMissing) {
+    if (obj != nil && obj != NSNull.null) {
+        if (RLMObjectBase *objBase = RLMDynamicCast<RLMObjectBase>(obj)) {
+            RLMObjectSchema *objectSchema = objBase->_objectSchema;
```
segiddins: `__unsafe_unretained const`?

Realm/RLMObjectStore.mm
```diff
@@ -463,8 +468,91 @@ void RLMAddObjectToRealm(RLMObjectBase *object, RLMRealm *realm, RLMCreationOpti
...
+            break;
+        case RLMPropertyTypeObject:
+            if (validateNested) {
+                RLMValidateNestedObject(obj, prop.objectClassName, schema, validateNested, allowMissing);
+            }
+            break;
+        case RLMPropertyTypeArray: {
+            if (validateNested) {
+                if (obj != nil && obj != NSNull.null) {
+                    id<NSFastEnumeration> array = obj;
```
segiddins: this should be a dynamic cast and throw an exception if it fails

Realm/Tests/ObjectTests.m
```diff
@@ -205,12 +205,20 @@ + (NSString *)primaryKey {
 }
 @end
 
+@interface EmployeeWrapper : RLMObject
```
segiddins: `LinkToPrimaryEmployeeObject`?

Realm/RLMObjectStore.mm
alazier: I don't think it is necessary to add this everywhere. It is an optimization and should be used in code paths which need optimization.

Realm/RLMObjectStore.mm
alazier: dynamic cast doesn't work with protocols. Can validate that this is enumerable though.

Realm/RLMObjectBase.mm
```diff
@@ -58,26 +59,56 @@ - (instancetype)init {
     return self;
 }
 
+static id RLMValidatedObjectForProperty(id obj, RLMProperty *prop, RLMSchema *schema) {
+    if (!RLMIsObjectValidForProperty(obj, prop)) {
```
jpsim: I'd prefer an early return here.

Realm/RLMAccessor.mm
```diff
@@ -179,25 +179,34 @@ static inline void RLMSetValue(__unsafe_unretained RLMObjectBase *const obj, NSU
...
 }
 
-static inline size_t RLMAddLinkedObject(RLMObjectBase *link,
-                                        __unsafe_unretained RLMRealm *const realm,
-                                        RLMCreationOptions options) {
+static inline RLMObjectBase *RLMGetLinkedObjectForValue(__unsafe_unretained RLMRealm *const realm,
+                                                        __unsafe_unretained NSString *const className,
+                                                        __unsafe_unretained id const value,
+                                                        RLMCreationOptions creationOptions) {
+    RLMObjectBase *link = RLMDynamicCast<RLMObjectBase>(value);
```
segiddins: this should also be `__unsafe_unretained const`?

Realm/RLMAccessor.mm
```diff
@@ -644,12 +650,11 @@ void RLMDynamicValidatedSet(RLMObjectBase *obj, NSString *propName, id val) {
...
     }
-    RLMDynamicSet(obj, prop, val, prop.isPrimary ? RLMCreationOptionsEnforceUnique : 0);
+    RLMDynamicSet(obj, prop, val, RLMCreationOptionsPromoteStandalone);
 }
 
-void RLMDynamicSet(__unsafe_unretained RLMObjectBase *const obj,
-                   __unsafe_unretained RLMProperty *const prop,
-                   __unsafe_unretained id val, RLMCreationOptions options) {
+void RLMDynamicSet(__unsafe_unretained RLMObjectBase *const obj, __unsafe_unretained RLMProperty *const prop,
+                   __unsafe_unretained id val, RLMCreationOptions creationOptions) {
```
segiddins: `val` needs to be `const`

Realm/RLMObjectBase.mm
```diff
@@ -58,26 +59,56 @@ - (instancetype)init {
...
+        if (array.count != properties.count) {
+            @throw RLMException(@"Invalid array input. Number of array elements does not match number of properties.");
+        }
         for (NSUInteger i = 0; i < array.count; i++) {
-            [self setValue:array[i] forKeyPath:[properties[i] name]];
+            [self setValue:RLMValidatedObjectForProperty(array[i], properties[i], schema) forKeyPath:[properties[i] name]];
         }
     }
     else {
         // assume our object is an NSDictionary or a an object with kvc properties
```
jpsim: "a an"

Realm/RLMObjectStore.mm
```diff
@@ -435,10 +432,13 @@ void RLMAddObjectToRealm(__unsafe_unretained RLMObjectBase *const object,
 
     // get or create row
     bool created;
-    auto primaryGetter = [=](__unsafe_unretained RLMProperty *const p) {
-        return [object valueForKey:p.getterName];
-    };
-    object->_row = (*schema.table)[RLMCreateOrGetRowForObject(schema, primaryGetter, options, created)];
+    auto primaryGetter = [=](RLMProperty *p) { return [object valueForKey:p.getterName]; };
```
jpsim: what required this change to remove the `__unsafe_unretained`?

Realm/RLMAccessor.mm
tgoyne: Local variables generally don't get pointless retain/releases inserted, so only if profiling indicates a problem.

Realm/RLMObjectStore.mm
alazier: rebase error

Realm/RLMObjectBase.mm
```diff
@@ -58,26 +59,56 @@ - (instancetype)init {
...
-        NSDictionary *dict = RLMValidatedDictionaryForObjectSchema(value, _objectSchema, schema);
-        for (NSString *name in dict) {
-            id val = dict[name];
-            // strip out NSNull before passing values to standalone setters
-            if (val == NSNull.null) {
-                val = nil;
+        NSDictionary *defaultValues = nil;
```
jpsim: This code is quite similar to what's in `RLMValidateValueForObjectSchema`, could you write a function?

jpsim:
Please add a CHANGELOG entry for the update issue you've fixed.

Realm/RLMAccessor.mm
alazier: In general I think we should only be adding optimizations where the profiler indicates there is an issue...

Realm/RLMObjectBase.mm
alazier: Don't think it is worth it in this case. The contents of the loop are slightly different so we would either need a block/lambda, or and out paramater for the cached defaults which is awkward.

alazier:
anything else?

Realm/RLMObjectStore.mm
jpsim: ok, then please revert this change.

jpsim:
Other than the remaining rebase error, you get my :+1:

Realm/RLMAccessor.mm
```diff
@@ -179,25 +179,34 @@ static inline void RLMSetValue(__unsafe_unretained RLMObjectBase *const obj, NSU
     }
 }
 
-static inline size_t RLMAddLinkedObject(RLMObjectBase *link,
-                                        __unsafe_unretained RLMRealm *const realm,
-                                        RLMCreationOptions options) {
+static inline RLMObjectBase *RLMGetLinkedObjectForValue(__unsafe_unretained RLMRealm *const realm,
```
tgoyne: This probably needs to be marked as `NS_RETURNS_RETAINED` to avoid autoreleasing everything (`RLMAddLinkedObject` avoided the autorelease by returning the row index rather than the object).

tgoyne: i.e. add `static inline RLMObjectBase *RLMGetLinkedObjectForValue(RLMRealm *realm, NSString *className, id value, RLMCreationOptions creationOptions) NS_RETURNS_RETAINED;` on the line before (GCC doesn't support marking the definition directly so clang gives a warning).

segiddins:
Seems good :+1:

Realm/RLMAccessor.mm
alazier: Does it matter since it's inlined?

tgoyne: Marking it as inline doesn't guarantee that it actually will be inlined.

Realm/RLMObjectStore.mm
alazier: I made this change in the other spot RLMCreateOrGetRowForObject was called...

Will make a new pr with this fix.

Issue #1970, opened 2015-05-21
autoreleasepool for dispatch_async now necessary in new version of Realm / RealmSwift


alexanderjarvis:
Hey there!

The most recent versions of Realm introduced some new functionality called "Interprocess communication" by means of a notifier that internally uses a `kqueue`.

I have found that in our app this notifier doesn't go away unless we wrap everything in an `autoreleasepool` (so that the Realm is `deinit` and also the notifier is stopped).

The problem is quite severe as basically the thread usage continues to climb (as we are using `dispatch_async`) until the point when no new network requests can be made by the app.

Wrapping stuff in `autoreleasepool` inside the `dispatch_async` seems to fix the issue. This was not an issue in previous versions of Realm however (we also migrated to RealmSwift at the same time).

Some guidance on how to properly use Realm and GCD would be helpful as at this stage I've tried various things including keeping a Realm instance around for a specific serial queue but found that the queue is not always equal to the thread so although my test case would work, it would fail when running as an app.

We don't actually care for the notifier as we call `realm.refresh()` on the main thread realm manually when there is a write transaction anyway (by means of callback). Being able to create a Realm inside `dispatch_async` that doesn't also create a notifier could be a solution for this pattern.

In addition to this, if we are creating new Realm instances in `dispatch_async` blocks where of course sometimes there are cached instances if on the same pthread, how do we effectively limit the number of concurrent operations without using a serial queue?

bdash:
Is the key issue that you're hitting that after creating Realms on various dispatch queues, threads become tied up and you're unable to make new network requests?

As for guidance on using Realm with GCD:

* Realm uses thread confinement for its objects, so you need to ensure that all work with a given object is performed on the same thread. Since GCD doesn't guarantee that all blocks submitted to a given serial queue will execute on the same thread, you'll need to get a fresh Realm instance inside each block.

* Wrapping work in autorelease pools is beneficial since GCD makes no guarantee as to how timely its own autorelease pools will be drained. Keeping Realm instances alive in undrained autorelease pools will result in higher than necessary footprint, and other related issues.

tgoyne:
#1975 will significantly mitigate this, but you can still run into issues with autoreleased realm objects if you have your own long blocking operations on background queues -- the autorelease pool for a GCD worker thread is drained only when the queue is empty, and the resulting long-lived RLMRealm will pin the data for its version, resulting in the realm file growing with each write transaction.

bigfish24:
:+1:!

[...]

PR #1975, opened 2015-05-21
Have RLMNotifier create a dedicated thread to listen for notifications…

bdash:
… rather than doing so on a global concurrent queue.

There is a relatively low cap on the number of worker threads that can
service the global concurrent queues. Using a dedicated thread to listen
for notifications avoids tying up a worker thread for the lifetime of
each RLMRealm instance, which prevents starving other GCD clients when a
large number of RLMRealm instances are live.

Addresses an issue mentioned in #1970.

### DIFF
--- a/Realm/Tests/RealmTests.mm
+++ b/Realm/Tests/RealmTests.mm
@@ -1240,6 +1240,22 @@ extern "C" {
     XCTAssertGreaterThan(fileSizeBefore, fileSizeAfter);
 }
 
+- (void)testCanCreate100RealmsWithoutBreakingGCD
+{
+    NSMutableArray *realms = [NSMutableArray array];
+    for (int i = 0; i < 100; ++i) {
+        NSString *realmFileName = [NSString stringWithFormat:@"test.%d.realm", i];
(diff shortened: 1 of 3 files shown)


tgoyne:
Oh hey, I was just about to say we should do this, although for slightly different reasons (a long running GCD task ensures that the autorelease pool for that thread never drains).

tgoyne:
:+1:

Realm/Tests/RealmTests.mm
```diff
@@ -1240,6 +1240,22 @@ - (void)testCompact
     XCTAssertGreaterThan(fileSizeBefore, fileSizeAfter);
 }
 
+- (void)testCanCreate100RealmsWithoutBreakingGCD
+{
+    NSMutableArray *realms = [NSMutableArray array];
+    for (int i = 0; i < 100; ++i) {
+        NSString *realmFileName = [NSString stringWithFormat:@"test.%d.realm", i];
```
jpsim: It looks like these files will never be cleaned up after the test runs.

bdash: Will fix.

PR #1988, opened 2015-05-23
RLMResults containsObject

mamaral:
It would be nice to have this quick and easy method to determine if an object exists in an instance of RLMResults. Thoughts?

### DIFF
--- a/Realm/RLMResults.h
+++ b/Realm/RLMResults.h
@@ -97,6 +97,13 @@
  *  ---------------------------------------------------------------------------------------
  */
 /**
+ Returns a Boolean value that indicates whether a given object is present in this RLMResults.
+
+ @param object  An object (of the same type as returned from the objectClassName selector).
+ */
+- (BOOL)containsObject:(RLMObject *)object;
+
+/**
  Gets the index of an object.
 
  Returns NSNotFound if the object is not found in this RLMResults.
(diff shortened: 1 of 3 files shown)


jpsim:
Thanks for the PR! Though I'm not convinced we should add a method to our API that provides syntactic sugar for something so simple.

If we do, for symmetry we'd have to provide `containsObject:` on `RLMResults`, `RLMArray` and have equivalents in Swift, which at that point means adding 4 methods + docs + tests.

IF you find yourself checking for membership often, I'd suggest adding a category on RLMResults to provide `containsObject:` for your app.

mamaral:
TBH, I'll likely use it *mostly* for unit testing my realm helper classes, to make sure my own convenience methods are working properly, but I'd imagine there's obviously other uses as well. I'd be happy to add those other methods, tests, and docs, or if you think it's not practical (you may be right) no biggie! Let me know!

mamaral:
Also, LMK if that Travis failure was due to my change or not.

jpsim:
That CI failure was unrelated to your change (simulator timing out).

If you want to unit test your Realm helper classes, you should be able to do that from your Xcode project.

It's not that adding these sets of methods (plus support material) isn't practical, it's that every new addition we make to our public API increases the amount users have to know and read through to fully understand Realm. A small change like this may not add *too* much to the API's weight, but changes like these add up quickly.

So I'll be closing this, but I hope you contribute to Realm again in the future!

mamaral:
Totally understand - thanks!

PR #1996, opened 2015-05-25
[Results] `Results` now conforms to `CVarArgType`.

### LINKED ISSUE #1995: Make Results and List conform to CVarArgType
Currently, the following code fails to compile:

```swift
Realm().objects(Company).filter("ANY employees IN %@", realm.objects(Employee))
```

Because `Results` doesn't conform to `CVarArgType`. The simplest workaround for now is to convert the `Results` into an `Array` like so:

```swift
Realm().objects(Company).filter("ANY employees IN %@", map(realm.objects(Employee), { $0 }))
```

But the proper solution would be to make `Results` and `List` conform to `CVarArgType`.

jpsim:
Fixes #1995. `List` already conforms to `CVarArgType` because it inherits from `NSObject`. /cc @tgoyne @bdash @segiddins

### DIFF
--- a/RealmSwift/Results.swift
+++ b/RealmSwift/Results.swift
@@ -291,6 +291,6 @@ extension Results: CVarArgType {
     /// Transform `self` into a series of machine words that can be
     /// appropriately interpreted by C varargs
     public func encode() -> [Word] {
-        return (map(self, { $0 }) as NSArray).encode()
+        return rlmResults.encode()
     }
 }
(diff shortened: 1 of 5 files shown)


RealmSwift/Results.swift
```diff
@@ -284,3 +284,13 @@ extension Results: CollectionType {
...
     public var endIndex: Int { return count }
 }
+
+// MARK: Variadic Arguments Support
+
+extension Results: CVarArgType {
+    /// Transform `self` into a series of machine words that can be
+    /// appropriately interpreted by C varargs
+    public func encode() -> [Word] {
+        return (map(self, { $0 }) as NSArray).encode()
```
segiddins: Why not return `rlmResults.encode()`

jpsim: Because that would just be way too simple! /jk, will do. it hadn't occurred to me.

tgoyne:
:+1:

Issue #1999, opened 2015-05-25
RealmSwift reset relationship list


krzyzanowskim:
Hi.

I have a relationship that I want to reset with new list, and I can't make is happen. What is the proper way to reset relationship ?

this is what I tried
```swift
class MyModel: Object {
    var childs =  List<Child>()

    func change() {
       realm?.write {
            // I want to replace childs relationship list. Nothing I tries is working
            // self.realm?.delete(self.childs)
            // self.chats.removeAll()
            self.childs = List<Child>()
            self.childs.extend(newChildrens)
       }
    }
}

```

jpsim:
I should preface this answer with a bit of context. The state of dynamic dispatch in Swift is quite limited and ripe with sharp corners. To make things work in Realm, we've had to make some compromises in this area, especially when it comes to `List` properties.

Since generic class properties cannot be represented in the Objective-C runtime, you shouldn't declare your `List` properties as `dynamic`. You're already doing this, so this part is ok. :+1: 

However, since non-dynamic properties don't have setters which can be overridden at runtime, Realm can't catch when you're trying to set that property and perform the equivalent operation in the underlying database.

So when you do `self.childs = List<Child>()` in your code above, Realm is never notified of what you did. Swift :cry:.

So this is why our documentation only ever shows `List` properties declared as `let`. Because setters wouldn't work :disappointed:.

So the recommended way to update a `List` property is to call [`List.removeAll()`](https://realm.io/docs/swift/latest/api/Classes/List.html#/s:FC10RealmSwift4List9removeAllU__FGS0_Q__FT_T_), followed by [`List.extend(_:)`](https://realm.io/docs/swift/latest/api/Classes/List.html#/s:FC10RealmSwift4List6extendU__FGS0_Q__USs12SequenceType_USs13GeneratorType__FQ_T_).

We'd really love to see Swift improve its dynamic capabilities and have filed numerous radars to help nudge future language developments in this direction and we hope you'll do the same if you're unsatisfied with the status quo.

krzyzanowskim:
there is issue with this sequence:

```swift
list.removeAll()
list.extend(newList)
```

let's say I already have record in db. by calling `extended()` realm engine will try to insert it again to the table, and if I have setup with primary key... it wont be possible. So I tried something like this

```swift
realm.write {
    realm.delete(list)
    list.removeAll()
    list.extend(newList)
}
```
Can't set primary key property 'mykey' to existing value '35;75'.

[...]

PR #2000, opened 2015-05-26
[RLMArray] call addOrUpdateObject if an object has a primary key in -[RLMArray addObject:]

jpsim:
Prompted by #1999. I couldn't think of a reason why we shouldn't be doing this, and that functionality hasn't changed since originally introducing it (before primary keys were even supported).

/cc @tgoyne @bdash @segiddins

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,20 @@
+x.x.x Release notes (yyyy-MM-dd)
+=============================================================
+
+### API breaking changes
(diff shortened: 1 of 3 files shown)


tgoyne:
`addObject:` updating the value of an existing object in the Realm is really weird and different from `-[RLMRealm addObject:]`.

jpsim:
Ok, agreed.

PR #2006, opened 2015-05-26
Improve error reporting when trying to index unindexable properties

### LINKED ISSUE #1997: Crash moving from 0.90.5 to 0.92.3
Hi,

I recently updated my realm base to 0.92.3 and am getting a crash now just loading the app.

Schema:
```
GTAmenity {
	remoteID {
		type = int;
		objectClassName = (null);
		indexed = NO;
		isPrimary = YES;
	}
	synchronizedAt {
		type = date;
		objectClassName = (null);
		indexed = NO;
		isPrimary = NO;
	}
	modifiedAt {
		type = date;
		objectClassName = (null);
		indexed = NO;
		isPrimary = NO;
	}
	company {
		type = string;
		objectClassName = (null);
		indexed = NO;
		isPrimary = NO;
	}
	address {
		type = string;
		objectClassName = (null);
		indexed = NO;
		isPrimary = NO;
	}
	zip {
		type = string;
		objectClassName = (null);
		indexed = NO;
		isPrimary = NO;
	}
	url {
		type = string;
		objectClassName = (null);
		indexed = NO;
		isPrimary = NO;
	}
	phone {
		type = string;
		objectClassName = (null);
		indexed = NO;
		isPrimary = NO;
	}
	city {
		type = object;
		objectClassName = GTCity;
		indexed = NO;
		isPrimary = NO;
	}
	photos {
		type = array;
		objectClassName = GTOptionalString;
		indexed = NO;
		isPrimary = NO;
	}
	latitude {
		type = double;
		objectClassName = (null);
		indexed = YES;
		isPrimary = NO;
	}
	longitude {
		type = double;
		objectClassName = (null);
		indexed = YES;
		isPrimary = NO;
	}
	trailType {
		type = int;
		objectClassName = (null);
		indexed = NO;
		isPrimary = NO;
	}
	distance {
		type = double;
		objectClassName = (null);
		indexed = NO;
		isPrimary = NO;
	}
	reviewCount {
		type = int;
		objectClassName = (null);
		indexed = NO;
		isPrimary = NO;
	}
	averageRating {
		type = double;
		objectClassName = (null);
		indexed = NO;
		isPrimary = NO;
	}
	category {
		type = object;
		objectClassName = GTAmenityCategory;
		indexed = NO;
		isPrimary = NO;
	}
	sponsorType {
		type = int;
		objectClassName = (null);
		indexed = NO;
		isPrimary = NO;
	}
}
 ```

screenshot 2015-05-25 14 03 10

with the offending property:
```
latitude {
	type = double;
	objectClassName = (null);
	indexed = YES;
	isPrimary = NO;
}```

crashes with:
```WARNING: An RLMRealm instance was deallocated during a write transaction and all pending changes have been rolled back. Make sure to retain a reference to the RLMRealm for the duration of the write transaction.
libc++abi.dylib: terminating with uncaught exception of type realm::LogicError: Illegal combination
```
Note: the code is siimply calling `[RLMRealm defaultRealm];` in AppDelegate

no code has changed in since the base was 0.90.5. What has changed, or what do I need to change in order to get this to work?

Thanks

-Matt

tgoyne:
Rollback the write transaction and throw an obj-c exception with a useful error message rather than just letting the c++ exception bubble out.

Closes #1997.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,8 +14,8 @@ x.x.x Release notes (yyyy-MM-dd)
 * Swift: Made `SortDescriptor` conform to the `Equatable` and
   `StringLiteralConvertible` protocols.
 * Int primary keys are once again automatically indexed.
-* Improve error reporting when indexing is request for a property that can't be
-  indexed.
+* Improve error reporting when attempting to mark a property of a type that
+  cannot be indexed as indexed.
 
 ### Bugfixes
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,8 @@ x.x.x Release notes (yyyy-MM-dd)
 * Swift: Made `SortDescriptor` conform to the `Equatable` and
   `StringLiteralConvertible` protocols.
 * Int primary keys are once again automatically indexed.
+* Improve error reporting when indexing is request for a property that can't be
+  indexed.
 
 ### Bugfixes
 
--- a/Realm/RLMObjectStore.mm
+++ b/Realm/RLMObjectStore.mm
@@ -92,17 +92,30 @@ static void RLMRealmUpdateIndexes(RLMRealm *realm) {
     for (RLMObjectSchema *objectSchema in realm.schema.objectSchema) {
         realm::Table *table = objectSchema.table;
         for (RLMProperty *prop in objectSchema.properties) {
-            if (prop.indexed != table->has_search_index(prop.column)) {
-                if (!realm.inWriteTransaction) {
-                    [realm beginWriteTransaction];
-                    commitWriteTransaction = true;
-                }
-                if (prop.indexed) {
+            if (prop.indexed == table->has_search_index(prop.column)) {
+                continue;
+            }
+
+            if (!realm.inWriteTransaction) {
+                [realm beginWriteTransaction];
+                commitWriteTransaction = true;
+            }
+            if (prop.indexed) {
+                try {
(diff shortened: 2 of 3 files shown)


CHANGELOG.md
```diff
@@ -12,6 +12,8 @@ x.x.x Release notes (yyyy-MM-dd)
   and `List.filter(_:...)`.
 * Swift: Made `SortDescriptor` conform to the `Equatable` and
   `StringLiteralConvertible` protocols.
+* Improve error reporting when indexing is request for a property that can't be
```
jpsim: Wording could be improved here. Maybe:

> Improved error reporting when attempting to mark a property as indexed with a type that cannot be indexed.

jpsim:
:+1:

segiddins:
Any reason not to do this validation when constructing the schema, instead of when adding the index?

tgoyne:
Doing it while constructing the schema requires hardcoding which types can be indexed rather than just supporting whatever core supports.

segiddins:
Since we do that anyways with primary key column types, and that's the way it appears we'll be doing checking for nullability, I think it makes sense to hardcode it -- it will also help ensure we have tests around all the types we support being indexed.

tgoyne:
Primary keys aren't exposing core functionality, so it does not make sense to defer to core for what types can be primary keys (all types technically could be).

I would prefer not to have a hardcoded list of types that can be nullable in the binding other than in error messages/documentation if it is possible.

PR #2007, opened 2015-05-26
Re-enable automatically indexing int primary keys

tgoyne:
It turns out that we've been unintentionally allowing indexing int properties since 0.92.0 due to missing error checking, so just go ahead and make it explicitly supported.


jpsim:
Looks good to me. :+1:

Surprisingly, there are no references to the supported indexable types in our long-form docs, so there's nothing to update there.

segiddins:
:+1:

PR #1942, opened 2015-05-14
Make the RealmSwift binary submittable to the iOS app store

segiddins:
Closes #1907.

\c @jpsim @tgoyne @bdash

PR #2008, opened 2015-05-26
Reduce target duplication in RealmSwift's Xcode project

bdash:
The framework targets for iOS and OS X are folded into a single target since they're virtually identical. The test targets for iOS, iOS devices, and OS X are folded into a single target with a small amount of conditional logic to preserve the small behavioral differences between the targets.

/cc @jpsim @segiddins @tgoyne


segiddins:
This will be a nightmare to reconcile with https://github.com/realm/realm-cocoa/pull/1942, which I'm actively working on as we speak

bdash:
@segiddins I'll handle resolving the conflicts that arise once that's merged in.

bdash:
Rebased on top of #1942.

segiddins:
failure seems unrelated

jpsim:
Nice work :+1:

PR #2009, opened 2015-05-27
Fix the dynamic framework installation examples

segiddins:
\c @jpsim


jpsim:
:+1: I aborted the CI build as it wouldn't check for this anyway.

PR #2011, opened 2015-05-27
Allow key paths on either side of operators in cases where it clearly makes sense

### LINKED ISSUE #1990: Support keypaths on either side of NSPredicate expressions
I test with the following code in Swift and Realm 0.92.3 . 
Only the predicate with variable as the left hand expression can work correctly.
```swift
import RealmSwift
class Cat: Object {
  dynamic var long:Double = 0.0
}
func testTmp(){
    let realm = Realm()
    realm.write {
      realm.create(Cat.self, value: ["long":4.0], update: false)
      realm.create(Cat.self, value: ["long":3.0], update: false)
    }
    XCTAssertEqual(realm.objects(Cat.self).filter(NSPredicate(format: "long > %@",NSNumber(double: 0.0))).count, 2) // pass 2==2
    XCTAssertEqual(realm.objects(Cat.self).filter(NSPredicate(format: "%@  < long",NSNumber(double: 0.0))).count, 2) // fail 0 != 2
  }
```

bdash:
This takes care of the basic cases of the key path being on the right hand side of the operator. There are some operators that this does not yet support (e.g., 'foo' CONTAINS stringKeyPath). It may require improvements in core to support them.

I'd appreciate feedback on whether anyone can see a nicer approach than using templates like I've done in this approach.

Fixes #1990.

### DIFF
--- a/Realm/RLMQueryUtil.mm
+++ b/Realm/RLMQueryUtil.mm
@@ -347,6 +347,8 @@ void add_link_constraint_to_query(realm::Query & query,
 }
 
 void add_link_constraint_to_query(realm::Query& query, NSPredicateOperatorType operatorType, RLMObject *obj, NSUInteger column) {
+    // Link constraints only support the equal-to and not-equal-to operators. The order of operands
+    // is not important for those comaprisons so we can delegate to the other implementation.
     add_link_constraint_to_query(query, operatorType, column, obj);
 }
 
--- a/Realm/RLMQueryUtil.mm
+++ b/Realm/RLMQueryUtil.mm
@@ -348,7 +348,7 @@ void add_link_constraint_to_query(realm::Query & query,
 
 void add_link_constraint_to_query(realm::Query& query, NSPredicateOperatorType operatorType, RLMObject *obj, NSUInteger column) {
     // Link constraints only support the equal-to and not-equal-to operators. The order of operands
-    // is not important for those comaprisons so we can delegate to the other implementation.
+    // is not important for those comparisons so we can delegate to the other implementation.
     add_link_constraint_to_query(query, operatorType, column, obj);
 }
 
--- a/Realm/RLMQueryUtil.mm
+++ b/Realm/RLMQueryUtil.mm
@@ -379,61 +379,81 @@ void process_or_group(Query &query, id array, Func&& func) {
     query.end_group();
 }
 
-template <typename T, typename TableGetter,
-    typename std::enable_if_t<!std::is_same<T, realm::DateTime>::value>* = nullptr>
-realm::Columns<T> valueOfTypeForQuery(TableGetter&& table, NSUInteger idx)
-{
-    return table()->template column<T>(idx);
-}
+template <typename RequestedType, typename TableGetter>
+struct ColumnOfTypeHelper {
+    static realm::Columns<RequestedType> convert(TableGetter&& table, NSUInteger idx)
+    {
+        return table()->template column<RequestedType>(idx);
+    }
+};
 
-template <typename T, typename TableGetter,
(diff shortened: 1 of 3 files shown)


Realm/RLMQueryUtil.mm
```diff
@@ -174,6 +171,23 @@ void add_string_constraint_to_query(realm::Query &query,
...
+                                    NSString *value,
+                                    Columns<String>&& column) {
+    switch (operatorType) {
+        case NSEqualToPredicateOperatorType:
+        case NSNotEqualToPredicateOperatorType:
+            add_string_constraint_to_query(query, operatorType, predicateOptions, std::move(column), value);
+            break;
+        default:
+            @throw RLMPredicateException(@"Invalid operator type",
+                                         @"Operator type %lu is not supported for string type with key path on right side of operator",
```
segiddins: unrelated, but we should turn those type numbers into strings

tgoyne:
I think swapping the lhs and rhs and inverting the comparison operator would work for everything that core supports currently. If you're interested in also adding support for things like `"str".begins_with(col)` to core (which I think would be pretty straightforward as far as the query stuff goes) you'll need to figure out what the API for that should look like to determine what the obj-c side needs to do.

Couldn't `valueOfTypeForQuery` be done with a static function in a partially-specialized struct? `enable_if` tends to do not-so-nice things to compile times.

bdash:
I went with this approach rather than inverting the comparison operator since looks like it'll be necessary to support the extra string operations.

I'll rework this to use a partially-specialized struct. `enable_if` was just the first thing that came to mind after remembering that partially-specializing function templates was not an option.

bdash:
This is now at a reviewable point.

Realm/RLMQueryUtil.mm
```diff
@@ -282,6 +346,10 @@ void add_link_constraint_to_query(realm::Query & query,
     }
 }
 
+void add_link_constraint_to_query(realm::Query& query, NSPredicateOperatorType operatorType, RLMObject *obj, NSUInteger column) {
```
tgoyne: I had to look at the code a bit to figure out why this is needed, so a brief comment would be useful.

Realm/RLMQueryUtil.mm
```diff
@@ -457,8 +605,13 @@ void update_query_with_value_expression(RLMSchema *schema,
     }
 
     validate_property_value(prop, value, @"Expected object of type %@ for property '%@' on object of type '%@', but received: %@", desc, keyPath);
-    add_constraint_to_query(query, prop.type, pred.predicateOperatorType,
-                            pred.options, indexes, index, value);
+    if (pred.leftExpression.expressionType == NSKeyPathExpressionType) {
+        add_constraint_to_query(query, prop.type, pred.predicateOperatorType,
+                                pred.options, indexes, index, value);
+    } else {
```
jpsim: shouldn't this check that `pred.rightExpression.expressionType == NSKeyPathExpressionType`?

tgoyne: There's a check earlier that at least one side is a keypath expression.

tgoyne:
:+1:

jpsim:
:+1:

Realm/RLMQueryUtil.mm
```diff
@@ -90,54 +90,87 @@ void set_table() override {}
     const Table* get_table() override { return nullptr; }
 };
 
+NSString *operatorName(NSPredicateOperatorType operatorType)
```
segiddins: static

segiddins:
this C++ is over my head, so I'm probably not qualified to give a :+1:

Realm/RLMQueryUtil.mm
bdash: This is inside an unnamed namespace so it already has internal linkage.

PR #2026, opened 2015-05-30
Fix Boken Docs Links

mamaral:
Just noticed a couple links on the README were broken.


segiddins:
:+1:

jpsim:
Thanks for the corrections, @mamaral!

PR #2033, opened 2015-06-01
[RLMRealm] Create in-memory realms in a temporary directory

### LINKED ISSUE #2032: Write In-Memory realms to a tmpdir
See https://github.com/realm/realm-cocoa/issues/2029

segiddins:
Closes #2032.

\c @jpsim @bdash

### DIFF
--- a/Realm/RLMRealm.mm
+++ b/Realm/RLMRealm.mm
@@ -294,6 +294,17 @@ static NSString * const c_defaultRealmFileName = @"default.realm";
     return [path stringByAppendingPathComponent:fileName];
 }
 
++ (NSString *)writeableTemporaryPathForFile:(NSString*)fileName
+{
+    NSString *path = NSTemporaryDirectory();
+    [[NSFileManager defaultManager] createDirectoryAtPath:path
+                              withIntermediateDirectories:YES
+                                               attributes:nil
+                                                    error:nil];
+    fileName = [NSString stringWithFormat:@"%d-%@", NSProcessInfo.processInfo.processIdentifier, fileName];
+    return [path stringByAppendingPathComponent:fileName];
+}
+
 + (instancetype)defaultRealm
 {
     return [RLMRealm realmWithPath:[RLMRealm defaultRealmPath] readOnly:NO error:nil];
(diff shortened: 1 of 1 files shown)


Realm/RLMRealm.mm
```diff
@@ -294,6 +294,17 @@ + (NSString *)writeablePathForFile:(NSString*)fileName
     return [path stringByAppendingPathComponent:fileName];
 }
 
++ (NSString *)writeableTemporaryPathForFile:(NSString*)fileName
```
bdash: `NSString*` -> `NSString *`.

`fileName` doesn't feel like the best name for the argument since it's not used as the file name. It's really an opaque identifier.

Realm/RLMRealm.mm
```diff
@@ -294,6 +294,17 @@ + (NSString *)writeablePathForFile:(NSString*)fileName
...
+{
+    NSString *path = NSTemporaryDirectory();
+    [[NSFileManager defaultManager] createDirectoryAtPath:path
+                              withIntermediateDirectories:YES
+                                               attributes:nil
+                                                    error:nil];
+    fileName = [NSString stringWithFormat:@"%d-%@", NSProcessInfo.processInfo.processIdentifier, fileName];
```
bdash: Is there a reason we're doing our own thing rather than using `mkstemp` or similar?

segiddins: I didn't know about `mkstemp` :)

tgoyne: Needs to give a consistent filename for the identifier so that an in-memory realm can be opened in multiple threads or processes (which this change currently breaks).

segiddins: Opening an in-memory realm in multiple processes is something we support?

tgoyne: Yes. There's even tests for it.

segiddins: For some reason, this change doesn't seem to break any tests though :/

tgoyne: The OS X tests hang when I run them locally.

segiddins:
Closing because this change doesn't work.

PR #2039, opened 2015-06-02
[build.sh] Just use the build directory for github releases

segiddins:
\c @jpsim


jpsim:
:+1: Thanks, I should have caught this.

PR #2042, opened 2015-06-03
[Swift][List] Removed unused List.init(_ rlmArray:) initializer

jpsim:
/cc @tgoyne @segiddins @bdash turns out we weren't using this, as detected by @kishikawakatsumi's SwiftCov tool :smile:.


segiddins:
:+1: should be fine since `List` has no default values

PR #2049, opened 2015-06-03
[build.sh] Only generate test coverage files in debug builds of OS X framework

jpsim:
Since we package the same `Realm.framework` that we test. /cc @tgoyne @segiddins @bdash


segiddins:
:+1:

kishikawakatsumi:
:+1:

PR #2050, opened 2015-06-03
Make Realm objects Key-Value Observing compliant

tgoyne:
Still lots of refactoring and cleanup to do, but this mostly works. Currently the main thing that's reviewable is the tests. Requires a core build from the tg-kvo branch.

Stuff still to do, in roughly the order I plan to do it:

- [x] Get core changes merged (https://github.com/realm/realm-core/pull/862, https://github.com/realm/realm-core/pull/867, https://github.com/realm/realm-core/pull/908)
- [x] Extract most of the notification stuff to a new source file
- [x] Finish removing the overridden methods on standalone classes and make the RLMObjectBase versions handle both standalone and persisted
- [x] Make all the notification and transaction log handling stuff not a horrible O(N^4) hackjob mess
- [x] Fix the remaining failing/disabled tests
- [x] Add basic Swift tests (`List<>` probably needs some special logic somewhere)
- [x] Perf testing: verify no significant regressions when not using KVO
- [x] Perf testing: verify KVO is not unusably slow (making it *fast* can come later)
- [x] Documentation (a link to the foundation docs plus a list of differences and notes)
- [x] Try to run ReactiveCocoa/ReactKit tests on Realm objects
- [x] A very minimal RAC3 sample in objc/swift


Realm/RLMArray.mm
```diff
@@ -149,9 +212,15 @@ - (NSUInteger)indexOfObject:(RLMObject *)object {
...
+- (void)removeAllObjects {
+    changeArray(self, NSKeyValueChangeRemoval, NSMakeRange(0, _backingArray.count), ^{
+        [_backingArray removeAllObjects];
+    });
+}
+
 - (void)deleteObjectsFromRealm {
     for (RLMObject *obj in _backingArray) {
-        RLMDeleteObjectFromRealm(obj, _realm);
+        RLMDeleteObjectFromRealm(obj, obj->_realm);
```
jpsim: this is a change in behavior

Realm/RLMArray.mm
```diff
@@ -168,6 +237,9 @@ - (RLMResults *)objectsWhere:(NSString *)predicateFormat args:(va_list)args
 }
 
 - (id)valueForKey:(NSString *)key {
+    if (!_backingArray) {
+        return @[];
```
jpsim: shouldn't this return `nil`? Why an empty array?

Realm/RLMAccessor.h
```diff
@@ -18,10 +18,16 @@
 
 #import <Foundation/Foundation.h>
 
-@class RLMObjectSchema, RLMProperty, RLMObjectBase;
+@class RLMObjectSchema, RLMProperty, RLMObjectBase, RLMRealm;
 
 typedef NSUInteger RLMCreationOptions;
 
+namespace realm {
```
segiddins: This need to be in a `.hpp` header.

Realm/RLMArray.mm
tgoyne: The change here is just that `_backingArray` is created lazily for standalone arrays (for reasons I don't remember; that may end up getting changed back if there's no longer a good reason for it), so a `nil` backing array needs to behave like an empty one.

Realm/RLMArray.mm
```diff
@@ -30,21 +30,55 @@
...
+        _parentObject = object;
+        _key = key;
     }
     return self;
 }
 
+// used by legacy models
 - (instancetype)initWithObjectClassName:(NSString *)objectClassName {
-    return [self initWithObjectClassName:objectClassName standalone:YES];
+    self = [super init];
```
segiddins: any reason not to call thru to `initWithObjectClassName:parentObject:key:` here?

Realm/RLMArrayLinkView.mm
```diff
@@ -74,9 +77,67 @@ static inline void RLMValidateObjectClass(__unsafe_unretained RLMObjectBase *con
...
+    }
+    else {
+        f();
+    }
+}
+
 //
 // public method implementations
 //
+- (RLMRealm *)realm {
```
segiddins: what's wrong with the synthesized getter?

Realm/RLMArray.mm
jpsim: cool

Realm/RLMArrayLinkView.mm
```diff
@@ -86,6 +147,19 @@ - (BOOL)isInvalidated {
     return !_backingLinkView->is_attached();
 }
 
+// These two methods take advance of that LinkViews are interned, so there's
```
segiddins: `advantage`

Realm/RLMArrayLinkView.mm
```diff
@@ -278,4 +395,18 @@ - (NSUInteger)indexOfObjectWithPredicate:(NSPredicate *)predicate {
     return RLMConvertNotFound(query.find());
 }
 
+- (NSArray *)objectsAtIndexes:(__unused NSIndexSet *)indexes {
```
segiddins: this doesn't seem correct?

Realm/RLMArrayLinkView.mm
tgoyne: This is overriding the base class getter, which always returns nil.

segiddins: And is there any reason the base class can’t use the synthesized getter?

Realm/RLMArrayLinkView.mm
tgoyne: actually implementing this method is just an optimization (array kvo requires that it exist and either return nil or the correct thing)

Realm/RLMArrayLinkView.mm
tgoyne: That adds a pointless extra ivar.

Realm/RLMObjectStore.mm
```diff
@@ -678,19 +696,46 @@ void RLMDeleteObjectFromRealm(__unsafe_unretained RLMObjectBase *const object,
...
 
     // set realm to nil
     object->_realm = nil;
 }
 
+void RLMClearTable(RLMObjectSchema *objectSchema) {
+    for (auto info : objectSchema->_observedObjects) {
+        for_each(info, [&](__unsafe_unretained auto o) {
+            [o willChangeValueForKey:@"invalidated"];
+            for (RLMProperty *prop in objectSchema.properties)
```
segiddins: why are you sending notifications for all the properties?

tgoyne: Link properties need to change to null before the accessor is invalidated so that observations over the key path can get cleaned up. Other property types don't strictly need to, but it's more consistent and matches the normal kvo behavior when you're observing a keypath and an intermediate property is set to nil, which is the closest analogy to an object being deleted.

Realm/Tests/KVOTests.mm
```diff
@@ -0,0 +1,1278 @@
...
+// Validate that `recorder` has at least one notification, and that the first
+// notification is the expected one
+#define AssertChanged(recorder, from, to) do { \
+    if (NSDictionary *note = AssertNotification((recorder))) { \
+        XCTAssertEqualObjects(@(NSKeyValueChangeSetting), note[NSKeyValueChangeKindKey]); \
+        XCTAssertEqualObjects((from), note[NSKeyValueChangeOldKey]); \
+        XCTAssertEqualObjects((to), note[NSKeyValueChangeNewKey]); \
+    } \
+    else { \
+        return; \
```
segiddins: won't this mean returning for the calling function?

tgoyne: yes. once there's one missing notification generally everything after it fails in increasing odd ways resulting in lots of error spam that makes it harder to find the actual problem.

segiddins: In that case, should there be an `XCTFail` there?

tgoyne: there's an error message from AssertNotification already

Realm/Tests/KVOTests.mm
```diff
@@ -0,0 +1,1278 @@
...
+- (void)testOptionsInitial {
+    KVOObject *obj = [self createObject];
+
+    {
+        KVORecorder r(self, obj, @"int32Col", 0);
+        XCTAssertEqual(0U, r.size());
+    }
+    {
+        KVORecorder r(self, obj, @"int32Col", NSKeyValueObservingOptionInitial);
+        r.pop_front();
```
segiddins: what's this testing?

tgoyne: that there is exactly one notification (pop_front asserts that there's at least one and KVORecorder's destructor asserts that all notifications were consumed)

Realm/Tests/KVOTests.mm
```diff
@@ -0,0 +1,1278 @@
...
+@end
+@implementation KVOMultipleAccessorsTests
+- (id)observableForObject:(id)value {
+    if (RLMObject *obj = RLMDynamicCast<RLMObject>(value)) {
+        RLMObject *copy = [[obj.objectSchema.accessorClass alloc] initWithRealm:obj.realm schema:obj.objectSchema];
+        copy->_row = obj->_row;
+        return copy;
+    }
+    else if (RLMArray *array = RLMDynamicCast<RLMArray>(value)) {
+        return array;
```
segiddins: couldn't this, for a link view, return a new `RLMArrayLinkView` backed by the same linklist?

tgoyne: That turned out to be shockingly awkward to implement and is unnecessary for what's being tested (there's nothing that you can observe directly on RLMArray which changes due to you doing something directly to an RLMArray object, so the whole notion of observing one object while modifying another doesn't really apply).

Realm/RLMArray.mm
tgoyne: The previous behavior was to crash unless the first object in the array was a persisted object, and throw an exception in that case. This changes it to actually work.

tgoyne:
This now passes all tests and has reasonable performance in most scenarios ("reasonable" meaning that the time spent in KVO is either insignificant or dominated by Foundation), and I don't have any more major implementation changes planned.

Realm/RLMArray.h
```diff
@@ -17,7 +17,7 @@
 ////////////////////////////////////////////////////////////////////////////
 
 #import <Foundation/Foundation.h>
-#import <Realm/RLMCollection.h>
+#import "RLMCollection.h"
```
segiddins: this doesn't seem right, since `RLMCollection` is a public header?

Realm/RLMAccessor.h
```diff
@@ -18,7 +18,7 @@
 
 #import <Foundation/Foundation.h>
 
-@class RLMObjectSchema, RLMProperty, RLMObjectBase;
+@class RLMObjectSchema, RLMProperty, RLMObjectBase, RLMRealm, RLMProperty;
```
segiddins: adding `RLMRealm` here seems to be superfluous

Realm/RLMAccessor.mm
```diff
@@ -28,6 +29,7 @@
 #import "RLMUtil.hpp"
 
 #import <objc/runtime.h>
+#import <realm/descriptor.hpp>
```
segiddins: unused?

Realm/RLMArray.h
```diff
@@ -32,6 +32,18 @@
 
  RLMArrays cannot be created directly. RLMArray properties on RLMObjects are
  lazily created when accessed, or can be obtained by querying a Realm.
+
+ ### Key-Value Observing
+
+ RLMArray supports array key-value observing on `LMArray properties on RLMObject
```
segiddins: `RLMArray`

Realm/RLMArray.mm
segiddins: This needs to check if the key is `invalidated` ?

Realm/RLMObservation.mm
```diff
@@ -0,0 +1,698 @@
...
+#import "RLMRealm_Private.hpp"
+#import "RLMSchema.h"
+
+#import <realm/lang_bind_helper.hpp>
+
+using namespace realm;
+
+namespace {
+    template<typename Iterator>
+    struct IteratorPair {
```
segiddins: any reason this can't be a `std::pair`?

tgoyne: `begin` and `end` need to be in the same namespace as the range type and adding overloads to namespace `std` is forbidden.

Realm/RLMAccessor.mm
tgoyne: No, it's used.

Realm/Tests/KVOTests.mm
```diff
@@ -0,0 +1,1233 @@
...
+    }
+}
+
+- (void)testArrayDiffs {
+    KVOLinkObject2 *obj = [self createLinkObject];
+    KVORecorder r(self, obj, @"array");
+
+    id mutator = [obj mutableArrayValueForKey:@"array"];
+
+#define AssertIndexChange(kind, indexes) do { \
```
segiddins: `undef` this if it's not used after this test?

segiddins:
(We talked about this in the office, but we might need tests around `RLMDynamicSet`?)

jpsim:
it seems like Realm Swift's `KVOTests.testAddToRealmAfterAddingObservers` consistently fails?

Realm/RLMAccessor.mm
```diff
@@ -391,36 +399,52 @@ static IMP RLMAccessorGetter(RLMProperty *prop, RLMAccessorCode accessorCode, NS
     }
 }
 
+template<typename Function>
+static void RLMWrapSetter(__unsafe_unretained RLMObjectBase *const obj, __unsafe_unretained NSString *const name, Function&& f) {
+    RLMObservationInfo *info = RLMGetObservationInfo(obj->_observationInfo, obj->_row.get_index(), obj->_objectSchema);
+    if (info) {
```
bdash: This could be:

    if (RLMObservationInfo info = …) {

Realm/RLMArray.mm
```diff
@@ -32,19 +32,39 @@ @implementation RLMArray {
...
+        f();
+        [parent didChange:kind valuesAtIndexes:is forKey:ar->_key];
+    }
+    else {
+        f();
+    }
+}
+
+static void changeArray(__unsafe_unretained RLMArray *const ar, NSKeyValueChange kind, NSUInteger index, dispatch_block_t f) {
+    changeArray(ar, kind, [NSIndexSet indexSetWithIndex:index], f);
```
bdash: In RLMArrayLinkView.mm you avoided creating these `NSIndexSet` instances until they were used. Is there a reason it is done eagerly here while being lazy in RLMArrayLinkView?

Realm/RLMListBase.mm
```diff
@@ -0,0 +1,60 @@
+////////////////////////////////////////////////////////////////////////////
+//
+// Copyright 2014 Realm Inc.
```
bdash: 2015?

Realm/RLMObject.h
```diff
@@ -60,8 +60,27 @@ RLM_ASSUME_NONNULL_BEGIN
  ### Relationships
  
  See our [Cocoa guide](http://realm.io/docs/cocoa/latest) for more details.
- */
 
+ ### Key-Value Observing
+
+ All RLMObject properties (including properties you create in subclasses) are
```
bdash: RLMObject -> \`RLMObject\` here, and in a few other places in this comment.

Realm/RLMObservation.hpp
```diff
@@ -0,0 +1,143 @@
...
+
+public:
+    // storage for the observationInfo property on RLMObjectBase
+    void *kvoInfo = nullptr;
+};
+
+// Get the the observation info chain for the given row
+// Will simply return info if it's non-null, and will search ojectSchema's array
+// for a matching one otherwise, and return null if there are none
+RLMObservationInfo *RLMGetObservationInfo(std::unique_ptr<RLMObservationInfo> const& info,
```
bdash: Why take a `const std::unique_ptr&` rather than taking a raw pointer? Does `RLMGetObservationInfo` for some reason care that the caller owns the `RLMObservationInfo` that's being passed as an argument?

Realm/RLMObservation.mm
```diff
@@ -0,0 +1,734 @@
...
+    else {
+        forEach([=](__unsafe_unretained auto o) {
+            [o didChangeValueForKey:key];
+        });
+    }
+}
+
+void RLMObservationInfo::prepareForInvalidation() {
+    REALM_ASSERT_DEBUG(objectSchema);
+    for (auto info = this; info; info = info->next)
```
bdash: What takes care of marking the previous items in the linked list as invalidated?

Realm/RLMObservation.mm
```diff
@@ -0,0 +1,734 @@
...
+
+void RLMTrackDeletions(__unsafe_unretained RLMRealm *const realm, dispatch_block_t block) {
+    struct change {
+        RLMObservationInfo *info;
+        __unsafe_unretained NSString *property;
+        NSMutableIndexSet *indexes;
+    };
+
+    std::vector<change> changes;
+    std::vector<RLMObservationInfo *> invalidated;
```
bdash: `changes` and `invalidated` are only used if objects are observed, so their declarations can be moved after that check.

Realm/RLMObservation.mm
```diff
@@ -0,0 +1,734 @@
...
+                RLMProperty *prop = observer->getObjectSchema().properties[link.origin_col_ndx];
+                NSString *name = prop.name;
+                if (prop.type != RLMPropertyTypeArray) {
+                    changes.push_back({observer, name});
+                    continue;
+                }
+
+                auto linkview = observer->getRow().get_linklist(prop.column);
+                change *c = nullptr;
+                for (auto& ac : changes) {
```
bdash: I'm not clear what `ac` stands for here.

Realm/RLMArray.mm
tgoyne: When I first wrote it willChange/didChange were always called so there was no point, and I forgot to update that.

Realm/RLMObservation.hpp
tgoyne: It makes the calling code a little less verbose, but it's not really worth it for such a trivial gain.

Realm/RLMObservation.mm
tgoyne: It's only ever called on the head of the list (since it's called on the item obtained from the array on the object schema). Not sure if it'd be better to document that that's a requirement (and assert !prev) or make it work despite being unneeded.

Realm/RLMObservation.mm
tgoyne: For good reason: it's an abbreviation of a variable name that no longer exists.

Realm/RLMObservation.mm
bdash: Documenting and asserting seems fine.

tgoyne:
Now actually passes on CI.

Realm/Tests/KVOTests.mm
```diff
@@ -0,0 +1,1461 @@
...
+
+    ~KVORecorder() {
+        id self = _observer;
+        @try {
+            [_obj removeObserver:_observer forKeyPath:_keyPath context:this];
+        }
+        @catch (NSException *e) {
+            XCTFail(@"%@", e.description);
+        }
+//        assert(_notifications.count == 0);
```
jpsim: can get rid of this

Realm/RLMObservation.hpp
```diff
@@ -0,0 +1,145 @@
...
+// for a matching one otherwise, and return null if there are none
+RLMObservationInfo *RLMGetObservationInfo(RLMObservationInfo *info, size_t row, RLMObjectSchema *objectSchema);
+
+// Call the appropriate SharedGroup member function, with change notifications
+void RLMAdvanceRead(realm::SharedGroup &sg, realm::History &history, RLMSchema *schema);
+void RLMRollbackAndContinueAsRead(realm::SharedGroup &sg, realm::History &history, RLMSchema *schema);
+void RLMPromoteToWrite(realm::SharedGroup &sg, realm::History &history, RLMSchema *schema);
+
+// delete all objects from a single table with change notifications
+void RLMClearTable(RLMObjectSchema *realm);
```
jpsim: Parameter name should be `objectSchema` here.

Realm/RLMObservation.mm
```diff
@@ -0,0 +1,732 @@
...
+                    row.get_table()->get_index_in_group(),
+                    row.get_index(),
+                    info});
+            }
+        }
+        sort(begin(observers), end(observers));
+    }
+
+    // Send didChange notifications to all observers marked as needing them
+    // Loop in reverse order to avoid O(N^2) behavior in Foundation
```
jpsim: I don't understand this. Where would the O(N^2) behavior come from if this was iterating forward?

Realm/Tests/KVOTests.mm
```diff
@@ -0,0 +1,1461 @@
...
+        return array;
+    }
+    else {
+        XCTFail(@"unsupported type");
+        return nil;
+    }
+}
+
+- (void)testIgnoredProperty {
+    // ignored properties do not notify other accessors for the same row
```
jpsim: Missing test?

jpsim:
This is absolutely fantastic. Great work! :clap: 

This will need to be rebased on master and the Swift changes will need to be applied to both Swift versions.

And there are some useful docs additions in the Objective-C API that would benefit from also being on the Swift API.

Once that's done, I'll be happy to see this merged :smile:.

PS: I was surprised that you had a RAC 3 example in Objective-C but not in Swift, although that can certainly be done in another PR.

examples/ios/objc/RACTableView/TableViewController.m
```diff
@@ -0,0 +1,254 @@
...
+         NSIndexSet *indexes = change[NSKeyValueChangeIndexesKey];
+
+         if (indexes) {
+             NSInteger section = [self.parent.groups indexOfObject:group];
+             NSArray *paths = [self indexSetToIndexPathArray:indexes section:section];
+             if (kind == NSKeyValueChangeInsertion) {
+                 [self.tableView insertRowsAtIndexPaths:paths withRowAnimation:UITableViewRowAnimationAutomatic];
+             }
+             else if (kind == NSKeyValueChangeRemoval) {
+                 [self.tableView insertRowsAtIndexPaths:paths withRowAnimation:UITableViewRowAnimationAutomatic];
```
bigfish24: This should be`deleteRowsAtIndexPath:`

Realm/RLMObservation.mm
tgoyne: The fact that it's O(N^2) is based on observation, not reasoning. It's not too hard to think of reasons why it would be (e.g. if `willChange:` pushes things onto the end of an array and `didChange` pops, since front popping is O(N) and back popping in O(1)), but I don't think why Foundation behaves how it does is really all that relevant.

Realm/Tests/KVOTests.mm
tgoyne: No, it's removing a test defined in the parent class that doesn't apply to this subclass.

Realm/RLMObservation.mm
jpsim: You're right it's not that relevant, but the comment confused me into thinking it was actually an important distinction.

Realm/Tests/KVOTests.mm
jpsim: k, just making sure

tgoyne:
The RAC example is RAC2 since RAC3 doesn't have UIKit bindings yet and thus it's hard to write a concise example that does anything interesting.

tgoyne:
This is now rebased.

jpsim:
:+1: :tada: :clap:

PR #2056, opened 2015-06-05
[RLMSchema] Throw an exception if there are duplicate class names

### LINKED ISSUE #2054: Error if the same class name is found twice
See if this is possible when and `RLMObjectBase` class is linked to the same target twice.

segiddins:
This should be more informative than dynamic casting in swift unit
tests silently failing, and will also handle the case where the two class
definitions aren't identical.

Closes #2054.

\c @jpsim @bdash


tgoyne:
:+1:

jpsim:
yup! :+1:

PR #2057, opened 2015-06-05
Refactor common object store code to c++

alazier:
First pass at creation a common code base that can be shared across bindings. Functionally not much has changed, but there are few small differences from previous behaviors:
- `isIndexed` is now ignored for primary key properties - primary key properties will always have an index created
- manually running a migration returns any issues encountered as an NSError as the return value. Before we threw exceptions even though the `performMigration` returns an error.

@tgoyne @jpsim @bdash


segiddins:
I believe the podspec will need to be updated?

Realm/ObjectStore/object_store.cpp
```diff
@@ -0,0 +1,395 @@
...
+        if (!primary_prop) {
+            continue;
+        }
+
+        realm::TableRef table = table_for_object_type(group, object_schema.name);
+        if (table->get_distinct_view(primary_prop->table_column).size() != table->size()) {
+            throw ObjectStoreException(ObjectStoreException::RealmDuplicatePrimaryKeyValue, {{"object_type", object_schema.name}, {"property_name", primary_prop->name}});
+        }
+    }
+}
```
segiddins: missing newline

alazier:
performance issues heave been resolved.

jpsim:
@alazier Sam thinks this may require podspec changes. Did you confirm that your changes work with our current podspec?

alazier:
So there are issues with the podspc. Fixed some of them but there are still some issues with the new headers. Would be great if @segiddins or someone else who actually knows cocoapods could fix rather than having me waste more time trying to figure this out. There should also be a test run by ci which verifies the podspec is in order for future changes.

Also feedback on the actual code would be appreciated.

segiddins:
Podspec should work now!

tgoyne:
Review status: 0 of 23 files reviewed, 24 unresolved discussions, some commit checks failed.

---

<sup>**[Realm/ObjectStore/object_schema.cpp, line 32 \[r4\]](https://reviewable.io:443/reviews/realm/realm-cocoa/2057#-JrTkNE8l_4wqCRoaikR)** ([raw file](https://github.com/realm/realm-cocoa/blob/29a556744b862092bdce2d9ede21496243d63dd3/Realm/ObjectStore/object_schema.cpp#L32)):</sup>
properties.reserve(count);

---

<sup>**[Realm/ObjectStore/object_schema.cpp, line 46 \[r4\]](https://reviewable.io:443/reviews/realm/realm-cocoa/2057#-JrTkKM_hElv8lLChFaq)** ([raw file](https://github.com/realm/realm-cocoa/blob/29a556744b862092bdce2d9ede21496243d63dd3/Realm/ObjectStore/object_schema.cpp#L46)):</sup>
This is a no-op

---

<sup>**[Realm/ObjectStore/object_schema.cpp, line 48 \[r4\]](https://reviewable.io:443/reviews/realm/realm-cocoa/2057#-JrTjzkdxzOgAzzXGzEH)** ([raw file](https://github.com/realm/realm-cocoa/blob/29a556744b862092bdce2d9ede21496243d63dd3/Realm/ObjectStore/object_schema.cpp#L48)):</sup>
properties.push_back(std::move(property));

---

<sup>**[Realm/ObjectStore/object_schema.cpp, line 53 \[r4\]](https://reviewable.io:443/reviews/realm/realm-cocoa/2057#-JrTkQbiPYdw2ysmTyOD)** ([raw file](https://github.com/realm/realm-cocoa/blob/29a556744b862092bdce2d9ede21496243d63dd3/Realm/ObjectStore/object_schema.cpp#L53)):</sup>
This isn't an iterator.

---

<sup>**[Realm/ObjectStore/object_schema.cpp, line 57 \[r4\]](https://reviewable.io:443/reviews/realm/realm-cocoa/2057#-JrTkZ35I1o26CF2h65d)** ([raw file](https://github.com/realm/realm-cocoa/blob/29a556744b862092bdce2d9ede21496243d63dd3/Realm/ObjectStore/object_schema.cpp#L57)):</sup>
Unimportant, but can just be:

    std::string error = "No property matching primary key '" + primary_key + "'";
    throw ObjectStoreValidationException({error}, name);

---

<sup>**[Realm/ObjectStore/object_schema.hpp, line 19 \[r4\]](https://reviewable.io:443/reviews/realm/realm-cocoa/2057#-JrTl-TTLpq-QrVfM0HV)** ([raw file](https://github.com/realm/realm-cocoa/blob/29a556744b862092bdce2d9ede21496243d63dd3/Realm/ObjectStore/object_schema.hpp#L19)):</sup>
REALM_OBJECT_SCHEMA_HPP

---

<sup>**[Realm/ObjectStore/object_schema.hpp, line 27 \[r4\]](https://reviewable.io:443/reviews/realm/realm-cocoa/2057#-JrTlJE4dEjAeQkof2eG)** ([raw file](https://github.com/realm/realm-cocoa/blob/29a556744b862092bdce2d9ede21496243d63dd3/Realm/ObjectStore/object_schema.hpp#L27)):</sup>
These should just be forward declarations.

---

<sup>**[Realm/ObjectStore/object_store.cpp, line 113 \[r4\]](https://reviewable.io:443/reviews/realm/realm-cocoa/2057#-JrTm5K5D6iJS-Tx9PTI)** ([raw file](https://github.com/realm/realm-cocoa/blob/29a556744b862092bdce2d9ede21496243d63dd3/Realm/ObjectStore/object_store.cpp#L113)):</sup>
This does an out-of-bounds read if table_name is a prefix of c_object_table_name_prefix (compare() doesn't check bounds).

---

<sup>**[Realm/ObjectStore/object_store.cpp, line 136 \[r4\]](https://reviewable.io:443/reviews/realm/realm-cocoa/2057#-JrTn6-kXVT1uexDmMSP)** ([raw file](https://github.com/realm/realm-cocoa/blob/29a556744b862092bdce2d9ede21496243d63dd3/Realm/ObjectStore/object_store.cpp#L136)):</sup>
Spaces around colon.

---

<sup>**[Realm/ObjectStore/object_store.cpp, line 243 \[r4\]](https://reviewable.io:443/reviews/realm/realm-cocoa/2057#-JrTnWUvhV_gFkY1uQGt)** ([raw file](https://github.com/realm/realm-cocoa/blob/29a556744b862092bdce2d9ede21496243d63dd3/Realm/ObjectStore/object_store.cpp#L243)):</sup>
No parens around returned thing

---

<sup>**[Realm/ObjectStore/object_store.cpp, line 246 \[r4\]](https://reviewable.io:443/reviews/realm/realm-cocoa/2057#-JrTnb85aJZP5Ssu4jC6)** ([raw file](https://github.com/realm/realm-cocoa/blob/29a556744b862092bdce2d9ede21496243d63dd3/Realm/ObjectStore/object_store.cpp#L246)):</sup>
Not an iterator

---

<sup>**[Realm/ObjectStore/object_store.cpp, line 256 \[r4\]](https://reviewable.io:443/reviews/realm/realm-cocoa/2057#-JrTniued7raFWwvWXBO)** ([raw file](https://github.com/realm/realm-cocoa/blob/29a556744b862092bdce2d9ede21496243d63dd3/Realm/ObjectStore/object_store.cpp#L256)):</sup>
First check appears to be redundant

---

<sup>**[Realm/ObjectStore/object_store.cpp, line 257 \[r4\]](https://reviewable.io:443/reviews/realm/realm-cocoa/2057#-JrToDFzC99HiZY2TCe0)** ([raw file](https://github.com/realm/realm-cocoa/blob/29a556744b862092bdce2d9ede21496243d63dd3/Realm/ObjectStore/object_store.cpp#L257)):</sup>
Most of these realm::ObjectStore::func calls don't need to be fully qualified

---

<sup>**[Realm/ObjectStore/object_store.cpp, line 291 \[r4\]](https://reviewable.io:443/reviews/realm/realm-cocoa/2057#-JrToVA3w-FsvLU_pi6a)** ([raw file](https://github.com/realm/realm-cocoa/blob/29a556744b862092bdce2d9ede21496243d63dd3/Realm/ObjectStore/object_store.cpp#L291)):</sup>
||

---

<sup>**[Realm/ObjectStore/object_store.cpp, line 326 \[r4\]](https://reviewable.io:443/reviews/realm/realm-cocoa/2057#-JrTod2ppXpU96XgGXcN)** ([raw file](https://github.com/realm/realm-cocoa/blob/29a556744b862092bdce2d9ede21496243d63dd3/Realm/ObjectStore/object_store.cpp#L326)):</sup>
schema.emplace_back(group, std::move(object_type))

---

<sup>**[Realm/ObjectStore/object_store.hpp, line 19 \[r4\]](https://reviewable.io:443/reviews/realm/realm-cocoa/2057#-JrTou8_uLu2DzkYY2NN)** ([raw file](https://github.com/realm/realm-cocoa/blob/29a556744b862092bdce2d9ede21496243d63dd3/Realm/ObjectStore/object_store.hpp#L19)):</sup>
REALM_OBJECT_STORE_HPP

---

<sup>**[Realm/ObjectStore/object_store.hpp, line 23 \[r4\]](https://reviewable.io:443/reviews/realm/realm-cocoa/2057#-JrTowVS74oqvvf6QNBQ)** ([raw file](https://github.com/realm/realm-cocoa/blob/29a556744b862092bdce2d9ede21496243d63dd3/Realm/ObjectStore/object_store.hpp#L23)):</sup>
Forward-declare stuff instead

---

<sup>**[Realm/ObjectStore/object_store.hpp, line 100 \[r4\]](https://reviewable.io:443/reviews/realm/realm-cocoa/2057#-JrTpMwJddEsDoKCf8eL)** ([raw file](https://github.com/realm/realm-cocoa/blob/29a556744b862092bdce2d9ede21496243d63dd3/Realm/ObjectStore/object_store.hpp#L100)):</sup>
Either override `what()` or don't inherit from std::exception

---

<sup>**[Realm/ObjectStore/object_store.hpp, line 109 \[r4\]](https://reviewable.io:443/reviews/realm/realm-cocoa/2057#-JrTp6l-yj3QNezY74cY)** ([raw file](https://github.com/realm/realm-cocoa/blob/29a556744b862092bdce2d9ede21496243d63dd3/Realm/ObjectStore/object_store.hpp#L109)):</sup>
garbage whitespace

---

<sup>**[Realm/ObjectStore/object_store.hpp, line 113 \[r4\]](https://reviewable.io:443/reviews/realm/realm-cocoa/2057#-JrTpCkfvemJZgV4RxWH)** ([raw file](https://github.com/realm/realm-cocoa/blob/29a556744b862092bdce2d9ede21496243d63dd3/Realm/ObjectStore/object_store.hpp#L113)):</sup>
const ref

---

<sup>**[Realm/ObjectStore/property.cpp, line 19 \[r4\]](https://reviewable.io:443/reviews/realm/realm-cocoa/2057#-JrTpVE97pJZyTp3RZr5)** ([raw file](https://github.com/realm/realm-cocoa/blob/29a556744b862092bdce2d9ede21496243d63dd3/Realm/ObjectStore/property.cpp#L19)):</sup>
Pointless file?

---

<sup>**[Realm/ObjectStore/property.hpp, line 48 \[r4\]](https://reviewable.io:443/reviews/realm/realm-cocoa/2057#-JrTph_BkEKxP8-DccjH)** ([raw file](https://github.com/realm/realm-cocoa/blob/29a556744b862092bdce2d9ede21496243d63dd3/Realm/ObjectStore/property.hpp#L48)):</sup>
No functional difference, but should probably be a struct since it's just a lump of public data.

---

<sup>**[Realm/ObjectStore/property.hpp, line 57 \[r4\]](https://reviewable.io:443/reviews/realm/realm-cocoa/2057#-JrTpeK2F0ISD02OTO0u)** ([raw file](https://github.com/realm/realm-cocoa/blob/29a556744b862092bdce2d9ede21496243d63dd3/Realm/ObjectStore/property.hpp#L57)):</sup>
||

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/realm/realm-cocoa/2057)

tgoyne:
Oh, that works far shittierly than I'd hoped.

tgoyne:
I can re-do the comments via the GH UI if you'd prefer (I'd hoped for it to make inline github comments...). Haven't looked at the obj-c code yet.

alazier:
No need to re-do the comments.

Realm/RLMMigration.mm
```diff
@@ -157,20 +132,16 @@ - (BOOL)deleteDataForClassName:(NSString *)name {
...
     }
 
     if ([_realm.schema schemaForClassName:name]) {
-        _realm.group->get_table(table)->clear();
+        table->clear();
     }
     else {
-        _realm.group->remove_table(table);
-
-        if (RLMRealmPrimaryKeyForObjectClass(_realm, name)) {
```
tgoyne: Was the cleanup of the primary key info moved somewhere else? I don't see it anywhere.

Realm/RLMObjectSchema.mm
```diff
@@ -316,17 +265,58 @@ - (void)setTable:(realm::Table *)table {
...
+    objectSchema.name = _className.UTF8String;
+    objectSchema.primary_key = _primaryKeyProperty ? _primaryKeyProperty.name.UTF8String : "";
+    for (RLMProperty *prop in _properties) {
+        Property p;
+        p.name = prop.name.UTF8String;
+        p.type = (PropertyType)prop.type;
+        p.object_type = prop.objectClassName ? prop.objectClassName.UTF8String : "";
+        p.is_indexed = prop.indexed;
+        p.is_primary = (prop == _primaryKeyProperty);
+        objectSchema.properties.push_back(p);
```
tgoyne: `push_back(std::move(p))`

Realm/RLMObjectSchema.mm
```diff
@@ -316,17 +265,58 @@ - (void)setTable:(realm::Table *)table {
...
-                                         NSString *className) {
-    NSString *tableName = RLMTableNameForClass(className);
-    return realm.group->get_table(tableName.UTF8String);
++ (instancetype)objectSchemaForObjectStoreSchema:(realm::ObjectSchema &)objectSchema {
+    RLMObjectSchema *schema = [RLMObjectSchema new];
+    schema.className = @(objectSchema.name.c_str());
+
+    // create array of RLMProperties
+    NSMutableArray *propArray = [NSMutableArray arrayWithCapacity:objectSchema.properties.size()];
+    for (size_t col = 0; col < objectSchema.properties.size(); col++) {
```
tgoyne: `for (Property &prop : objectSchema.properties) {`

Realm/RLMObjectStore.mm
```diff
@@ -339,13 +167,13 @@ static bool RLMMigrationRequired(RLMRealm *realm, uint64_t newVersion, uint64_t
         else {
             [realm cancelWriteTransaction];
         }
-    }
-    @catch (NSException *) {
+    } catch (ObjectStoreException e) {
```
tgoyne: catch by const reference or it does stupid things

Realm/RLMMigration.mm
alazier: Will need to add this back in I guess. We can remove entries for any tables for deleted object types whether or not this method is called.

alazier:
@tgoyne I think I have addressed all of your comments other than implementing `what()` for the exception classes. The issue there is that we really want to return a formatted string, so we need to allocate memory to store each message in the exception. If this is acceptable I will make this change, otherwise will simply not derive from std::exception.

alazier:
@tgoyne I've implemented `what()` for the exception classes. I think that covers all of the feedback thus far. Want to take another look?

Realm/ObjectStore/property.hpp
```diff
@@ -0,0 +1,86 @@
...
+        bool is_primary;
+        bool is_indexed;
+
+        size_t table_column;
+        bool requires_index() { return is_primary || is_indexed; }
+    };
+
+    static inline const char *string_for_property_type(PropertyType type) {
+        switch (type) {
+            case PropertyTypeString:
```
kneth: The returned strings should be dependent on which binding, we are working on. I know that `#if defined(...)` is an ugly way of doing it but at least it is easy.

Realm/ObjectStore/object_store.cpp
```diff
@@ -0,0 +1,418 @@
...
+
+const uint64_t ObjectStore::NotVersioned = numeric_limits<uint64_t>::max();
+
+bool ObjectStore::has_metadata_tables(realm::Group *group) {
+    return group->get_table(c_primaryKeyTableName) && group->get_table(c_metadataTableName);
+}
+
+bool ObjectStore::create_metadata_tables(realm::Group *group) {
+    bool changed = false;
+    realm::TableRef table = group->get_or_add_table(c_primaryKeyTableName);
```
kneth: Is the `realm::` really needed?

Realm/ObjectStore/object_store.cpp
```diff
@@ -0,0 +1,418 @@
...
+            table->remove(row);
+        }
+    }
+    else {
+        table->set_string(c_primaryKeyPropertyNameColumnIndex, row, primary_key);
+    }
+}
+
+string ObjectStore::object_type_for_table_name(const string &table_name) {
+    if (table_name.size() >= 6 && table_name.compare(0, 6, c_object_table_name_prefix) == 0) {
```
kneth: Wouldn't it be safer to declare a constant to be the length of `c_object_table_name_prefix` - in case we change it.

Realm/ObjectStore/property.hpp
tgoyne: This PR is explicitly not handling things not needed to switch the obj-c code over to it to minimize the time we have two implementations of the exact same thing to keep up to date.

Realm/ObjectStore/object_store.cpp
```diff
@@ -0,0 +1,418 @@
...
+
+    // second pass adds/removes columns for out of date tables
+    for (auto& target_object_schema : to_update) {
+        TableRef table = table_for_object_type(group, target_object_schema->name);
+        ObjectSchema current_schema(group, target_object_schema->name, table.get());
+        vector<Property> &target_props = target_object_schema->properties;
+
+        // add missing columns
+        for (auto& target_prop : target_props) {
+            auto current_prop = current_schema.property_for_name(target_prop.name);
```
kneth: A column name must be less than 64 characters (I believe core will truncate). I think we should validate `target_prop.name`.

Realm/ObjectStore/object_store.cpp
```diff
@@ -0,0 +1,418 @@
...
+            // there is no primary key, so if there was one nil out
+            set_primary_key_for_object(group, target_object_schema->name, "");
+            changed = true;
+        }
+    }
+
+    // remove primary key entries for deleted types
+    // FIXME - delete actual tables once we have proper testing
+    ObjectStore::Schema schema;
+    for (int i = (int)group->size() - 1; i >= 0 ; i--) {
```
kneth: Why is `change` to set to `true`?

Realm/ObjectStore/object_schema.cpp
```diff
@@ -0,0 +1,70 @@
+////////////////////////////////////////////////////////////////////////////
+//
+// Copyright 2014 Realm Inc.
```
cmelchior: 2015

Realm/ObjectStore/object_store.cpp
```diff
@@ -0,0 +1,418 @@
...
+
+    validate_primary_column_uniqueness(group, schema);
+
+    set_schema_version(group, version);
+    return true;
+}
+
+ObjectStore::Schema ObjectStore::schema_from_group(Group *group) {
+    ObjectStore::Schema schema;
+    for (size_t i = 0; i < group->size(); i++) {
```
kneth: In line 281, you cast to `int` but not here?

Realm/ObjectStore/object_schema.hpp
```diff
@@ -0,0 +1,50 @@
+////////////////////////////////////////////////////////////////////////////
+//
+// Copyright 2014 Realm Inc.
```
cmelchior: 2015

kneth:
I haven't reviewed the Objective C pieces (read them as an example on how to use the object store). I like the object store parts, and I think we can use them easily in the Java binding too.

Realm/ObjectStore/object_schema.hpp
```diff
@@ -0,0 +1,50 @@
...
+    class Group;
+    class Table;
+
+    class ObjectSchema {
+    public:
+        ObjectSchema() {}
+
+        // create object schema from existing table
+        // if no table is provided it is looked up in the group
+        ObjectSchema(Group *group, const std::string &name, Table *table = nullptr);
```
cmelchior: Wouldn't it be better to split this into two as it covers two use cases: 
1) with string, but no table and 
2) with table but then you don't need the name as you can look that up in the group?

Realm/ObjectStore/object_store.cpp
```diff
@@ -0,0 +1,418 @@
+////////////////////////////////////////////////////////////////////////////
+//
+// Copyright 2014 Realm Inc.
```
cmelchior: 2015

Realm/ObjectStore/object_store.cpp
```diff
@@ -0,0 +1,418 @@
...
+    realm::TableRef table = group->get_table(c_metadataTableName);
+    if (!table || table->get_column_count() == 0) {
+        return realm::ObjectStore::NotVersioned;
+    }
+    return table->get_int(c_versionColumnIndex, 0);
+}
+
+void ObjectStore::set_schema_version(realm::Group *group, uint64_t version) {
+    realm::TableRef table = group->get_or_add_table(c_metadataTableName);
+    table->set_int(c_versionColumnIndex, 0, version);
```
cmelchior: I would make the row index a constant too

Realm/ObjectStore/object_store.cpp
```diff
@@ -0,0 +1,418 @@
...
+
+    return changed;
+}
+
+uint64_t ObjectStore::get_schema_version(realm::Group *group) {
+    realm::TableRef table = group->get_table(c_metadataTableName);
+    if (!table || table->get_column_count() == 0) {
+        return realm::ObjectStore::NotVersioned;
+    }
+    return table->get_int(c_versionColumnIndex, 0);
```
cmelchior: I would make the row index a constant too

Realm/ObjectStore/object_store.cpp
cmelchior: 6 is kinda magic, I assume it is "class_" but a constant would be better

Realm/ObjectStore/object_store.cpp
```diff
@@ -0,0 +1,418 @@
...
+            }
+        }
+    }
+
+    return changed;
+}
+
+bool ObjectStore::is_migration_required(realm::Group *group, uint64_t new_version) {
+    uint64_t old_version = get_schema_version(group);
+    if (old_version > new_version && old_version != NotVersioned) {
```
cmelchior: Is this enough to detect a schema mismatch? I would assume that you need to validate all tables as a user might accidentally set the same version even though tables has changed?

Realm/ObjectStore/object_store.hpp
```diff
@@ -0,0 +1,112 @@
...
+        static uint64_t get_schema_version(Group *group);
+
+        // checks if a migration is required for a given schema version
+        static bool is_migration_required(realm::Group *group, uint64_t new_version);
+
+        // verify a target schema against its table, setting the table_column property on each schema object
+        // updates the column mapping on the target_schema
+        // if no table is provided it is fetched from the group
+        // returns array of validation errors
+        static std::vector<std::string> validate_schema(Group *group, ObjectSchema &target_schema, Table *cached_table = nullptr);
```
cmelchior: I assume better support for localised errors has been postponed?

Realm/ObjectStore/object_store_exceptions.cpp
```diff
@@ -0,0 +1,45 @@
+////////////////////////////////////////////////////////////////////////////
+//
+// Copyright 2014 Realm Inc.
```
cmelchior: 2015

Realm/ObjectStore/object_store_exceptions.cpp
```diff
@@ -0,0 +1,45 @@
...
+
+#include "object_store_exceptions.hpp"
+
+using namespace realm;
+using namespace std;
+
+ObjectStoreException::ObjectStoreException(Kind kind, Dict dict) : m_kind(kind), m_dict(dict) {
+    switch (m_kind) {
+        case Kind::RealmVersionGreaterThanSchemaVersion:
+            m_what = "Schema version less than last set version in Realm";
```
cmelchior: Returning the two different versions would be nice

Realm/ObjectStore/object_store_exceptions.hpp
```diff
@@ -0,0 +1,66 @@
+////////////////////////////////////////////////////////////////////////////
+//
+// Copyright 2014 Realm Inc.
```
cmelchior: 2015

Realm/ObjectStore/object_store.cpp
tgoyne: Core throws an exception now if it's too long.

Realm/ObjectStore/object_store.cpp
```diff
@@ -0,0 +1,418 @@
...
+            }
+        }
+        else if (current_schema.primary_key.length()) {
+            // there is no primary key, so if there was one nil out
+            set_primary_key_for_object(group, target_object_schema->name, "");
+            changed = true;
+        }
+    }
+
+    // remove primary key entries for deleted types
```
tgoyne: This is not the correct logic. We decided not to ever automatically remove tables to continue to support adding tables without bumping the schema version. The PK table entries should continue to be removed at the same time as the table is removed.

Realm/ObjectStore/object_store.cpp
tgoyne: That one's looping backwards.

Realm/ObjectStore/object_store.hpp
tgoyne: Yes. To avoid merge hell we want to minimize the amount of time that we have ported and unmerged versions of the current obj-c code.

Realm/ObjectStore/property.hpp
alazier: After this pr there will be another with allows binding authors to supply their own strings for some or all error messages.

Realm/ObjectStore/object_store.cpp
alazier: It's not needed and will be removed.

Realm/ObjectStore/object_store.cpp
alazier: Anytime something is written to the db changed needs to be set to true. Callers of this method can cancel the write transaction if no changes are made. If changes are made even to the pk table the transaction needs to be committed.

Realm/ObjectStore/object_schema.hpp
alazier: I agree this is confusing. This was added as an optimization when we already had the table available. Turns out passing in the table isn't really any faster than just getting the table from the group, so I will just remove the last argument and fetch the table every time.

Realm/ObjectStore/object_store.cpp
alazier: renaming to `is_schema_at_version`

Realm/ObjectStore/object_store.cpp
alazier: We aren't actually removing the table here and we don't have to in the future (although I think we should, and I don't see how it's related to adding new tables). 

In any case this new code is a no-op unless tables are removed, and the only difference from the current functionality should be the time at which the primary key entry is removed.

tgoyne: This removes the primary key for tables which still exist that are not present in the object schema., which is not even remotely the same thing as removing the primary key when a table is removed.  Even if it did the more sensible thing of removing entries in the pk table which mention tables that don't exist, this runs before the user gets the chance to call `deleteDataForClassName:`, so the stale data will persist until the next time the schema is initialized (and thus require an extra write transaction then).

tgoyne: We want to allow multiple things with different subsets of the schema to be able to share a realm file as long as they have the same schema version and their overlapping subsets are compatible. This currently works fine, but removing extra tables breaks it.

tgoyne: I don't remember if that was the actually reason why we decided not to automatically remove tables, though. IIRC there were a few issues that we could think of.

segiddins: That was definitely one of the reasons against automatically removing tables.

alazier: Can you come up with an of example where someone would actually need support for schema subsets/sharing realms/etc? All of this sounds pretty complicated to explain/maintain especially when its usefulness is not obvious.

tgoyne: The only thing we've ever had to do to keep it working is not make pointless changes to break it, and that's still the case here. That's about as far from complicated to maintain as you can get.

tgoyne: Like if you really want to waste another day arguing about this then I guess we can do that, but it shouldn't be an incidental change slipped into a rewrite.

alazier: I made this change in response to your earlier feedback. Rather than complaining about the fixes it may be more constructive to propose a solution that you would be happy with. My only requirement is that the pk table is not exposed to users and should be abstracted behind the object store apis.

cmelchior: Just to note. We have just merged custom schema support for Android. However we are still requiring a complete match between the custom schema definition and the schema found in the actual Realm file.

tgoyne: The simple solution is just leave the existing logic alone. The better solution would be to move all of the logic in `deleteDataForClassName:` into `ObjectStore`, again unchanged other than the cosmetic things required.

alazier: @cmelchior do you expect the object store validation to ensure a complete match (I assume you would)? If so we can maybe add an option to ignore extra tables in the Realm file (which seems to be what objc expects).

I assume since you require a strict match you also expect tables for removed types to be deleted? We could add an option for this as well if we need differing behavior. Note that if we don't behave the same across bindings objc could produce Realm files which would be rejected by the andriod binding during validation. Ideally both bindings would behave identically to avoid these issues altogether.

alazier:
I added a new method per @tgoyne 's suggestion to remove the table and clear the primary key to preserve the previous behavior in the objc binding. @cmelchior if there is other stuff we need to support the java binding we can add that to a future pr. As soon as we finish this I can get started adding the ability to specify custom error messages and whatever else you guys need. We should also probably discuss ways to reconcile differences in behavior across bindings where possible but there is no need for that discussion to block this pr.

Realm/ObjectStore/object_store.cpp
cmelchior: Yes that is the idea, since having leftover tables would be due to a developer error.

cmelchior:
So far this looks pretty good also from a Java perspective. Right now we have some migration code to fix a mistake in the primary key table, but that will be deprecated before implementing this anyway, and I agree that merging this sooner than later is probably a good idea.

tgoyne:
`verify` passes for me locally (other than some pre-existing issues with device tests), so CI issues appear to actually be just CI being broken.

LGTM now, but hold off on merging until 0.93.2 is released (which I guess will be tomorrow now since I assume all the CPH people are asleep).

Realm/ObjectStore/object_store.cpp
bmunkholm: It's a clear requirement that Realm files must be compatible across bindings. That's among the reasons we make this common layer besides save time for developing new bindings.

Realm/ObjectStore/object_schema.cpp
```diff
@@ -0,0 +1,67 @@
...
+#include "object_schema.hpp"
+#include "object_store.hpp"
+
+#include <realm/group.hpp>
+#include <realm/table.hpp>
+
+using namespace realm;
+using namespace std;
+
+ObjectSchema::ObjectSchema(Group *group, const std::string &name) : name(name) {
```
bmunkholm: Our C++ code standard requires the { to be on the next line. (so does the Obj-C coding standard, although that's not being enforced).

Realm/ObjectStore/object_store.cpp
```diff
@@ -0,0 +1,414 @@
...
+}
+
+static inline bool property_has_changed(Property &p1, Property &p2) {
+    return p1.type != p2.type || p1.name != p2.name || p1.object_type != p2.object_type;
+}
+
+// set references to tables on targetSchema and create/update any missing or out-of-date tables
+// if update existing is true, updates existing tables, otherwise validates existing tables
+// NOTE: must be called from within write transaction
+bool ObjectStore::create_tables(Group *group, ObjectStore::Schema &target_schema, bool update_existing) {
```
bmunkholm: This method is a good role model example of commenting. I really appreciate the short but precise one-line comments ahead of a small code section. :+1:  Some of the other methods could be more like this :-)

Realm/Tests/RealmTests.mm
```diff
@@ -20,6 +20,9 @@
 
 #import "RLMObjectSchema_Private.hpp"
 #import "RLMRealm_Dynamic.h"
+#import "RLMRealm_Private.hpp"
```
segiddins: these look unnecessary?

Realm/Tests/PerformanceTests.m
```diff
@@ -18,8 +18,6 @@
 
 #import "RLMTestCase.h"
 
-#if !DEBUG
```
segiddins: I think this is unintentional?

alazier: sure is. good catch.

Realm/ObjectStore/object_schema.cpp
bmunkholm: @alazier Once this is moved out of the Cocoa repo to it's own, please adhere to the C++ coding style guideline for { placement on a separate line for method body.

PR #2061, opened 2015-06-08
Make RLMArray and RLMResults generic

tgoyne:
Not really very useful at the moment since you have to declare properties as `@property RLMArray<DogObject *><DogObject> *dogs` to get the typed version, and doing things like `RLMResults<DogObject *> dogs = [EmployeeObject allObjects]` doesn't give a warning.

### DIFF
--- a/Realm/RLMCollection.h
+++ b/Realm/RLMCollection.h
@@ -22,6 +22,15 @@ RLM_ASSUME_NONNULL_BEGIN
 
 @class RLMRealm, RLMResults, RLMObject;
 
+#if __has_extension(objc_generics)
+#define RLM_GENERIC_COLLECTION <RLMObjectType>
+#define RLMObjectArgument RLMObjectType
+#else
+#define RLM_GENERIC_COLLECTION
+typedef id RLMObjectType;
+typedef RLMObject * RLMObjectArgument;
+#endif
+
 @protocol RLMCollection <NSFastEnumeration>
 
 @required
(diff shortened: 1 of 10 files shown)


segiddins:
@tgoyne I actually liked your idea of making the inline macro

Realm/RLMCollection.h
```diff
@@ -18,6 +18,15 @@
 
 @class RLMRealm, RLMResults, RLMObject;
 
+#if __has_extension(objc_generics)
+#define RLM_GENERIC_COLLECTION <RLMObjectType>
```
segiddins: isn't there some way to specify that `RLMObjectType` is `__kindof RLMObject` ?

jpsim:
I think we could merge this once we add a changelog entry and maybe point to `RLM_GENERIC_ARRAY` to our `RLMArray` docs. We'll need to add realm.io docs too. Anything else?

Realm/RLMCollection.h
```diff
@@ -22,6 +22,17 @@ RLM_ASSUME_NONNULL_BEGIN
 
 @class RLMRealm, RLMResults, RLMObject;
 
+#if __has_extension(objc_generics)
```
segiddins: would this belong better in `RLMDefines.h` ?

Realm/RLMCollection.h
jpsim: no.

Realm/RLMCollection.h
jpsim: yes.

jpsim:
This is ready to review /cc @tgoyne @segiddins @bdash. Accompanying docs PR: https://github.com/realm/realm.io/pull/742

segiddins:
Is there no way to provide this type info for the places where we create `RLMResults` ?

tgoyne:
`RLMResults<instancetype>` isn't supported. I keep meaning to file a radar for it as the reasons why `Array<Self>` isn't supported don't really apply.

segiddins:
In that case, :+1:

jpsim:
:+1:

PR #2063, opened 2015-06-09
Make backwards-compatible changes for Xcode 7

tgoyne:
This makes all of the obj-c stuff build and pass tests in both Xcode 6 and 7. Swift stuff obviously doesn't build in Xcode 7, but this includes all of the changes needed for it that don't break 6 to cut down on the diff a bit.

### DIFF
--- a/Realm/Swift/RLMSupport.swift
+++ b/Realm/Swift/RLMSupport.swift
@@ -29,18 +29,27 @@ extension RLMObject {
     }
 }
 
+public class RLMGenerator: GeneratorType {
+    private var i: UInt = 0
+    private let collection: RLMCollection
+
+    init(collection: RLMCollection) {
+        self.collection = collection
+    }
+
+    public func next() -> RLMObject? {
+        if i >= collection.count {
+            return .None
+        } else {
+            return collection[i++] as? RLMObject
+        }
+    }
+}
+
 extension RLMArray: SequenceType {
     // Support Sequence-style enumeration
-    public func generate() -> GeneratorOf<RLMObject> {
-        var i: UInt  = 0
-
-        return GeneratorOf<RLMObject> {
-            if (i >= self.count) {
-                return .None
-            } else {
-                return self[i++] as? RLMObject
-            }
-        }
+    public func generate() -> RLMGenerator {
+        return RLMGenerator(collection: self)
     }
 
     // Swift query convenience functions
@@ -55,16 +64,8 @@ extension RLMArray: SequenceType {
 
 extension RLMResults: SequenceType {
     // Support Sequence-style enumeration
-    public func generate() -> GeneratorOf<RLMObject> {
-        var i: UInt  = 0
-
-        return GeneratorOf<RLMObject> {
-            if (i >= self.count) {
-                return .None
-            } else {
-                return self[i++] as? RLMObject
-            }
-        }
+    public func generate() -> RLMGenerator {
+        return RLMGenerator(collection: self)
     }
 
     // Swift query convenience functions
--- a/Realm/Tests/RLMMultiProcessTestCase.m
+++ b/Realm/Tests/RLMMultiProcessTestCase.m
@@ -58,10 +58,8 @@
 
 - (void)setUp {
     self.isParent = !getenv("RLMProcessIsChild");
-
-    NSProcessInfo *info = NSProcessInfo.processInfo;
-    self.xctestPath = info.arguments[0];
-    self.testsPath = [info.arguments lastObject];
+    self.xctestPath = NSProcessInfo.processInfo.arguments[0];
+    self.testsPath = [NSBundle bundleForClass:[self class]].bundlePath;
 
     [super setUp];
 }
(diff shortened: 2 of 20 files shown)


Realm/Tests/RLMMultiProcessTestCase.m
```diff
@@ -77,6 +75,7 @@ - (NSTask *)childTask {
     NSString *testName = [NSString stringWithFormat:@"%@/%@", self.className, self.testName];
     NSMutableDictionary *env = [NSProcessInfo.processInfo.environment mutableCopy];
     env[@"RLMProcessIsChild"] = @"true";
+    [env removeObjectForKey:@"XCTestConfigurationFilePath"];
```
bdash: Why is this necessary?

tgoyne: Xcode 7 switched from passing all the undocumented options for things like how to communicate with the IDE from the command line to a config file, with the path specific in the env variable. We don't want the child process to use these settings because multiple xctest processes communicating with the IDE doesn't work at all.

Realm/Swift/RLMSupport.swift
```diff
@@ -29,18 +29,29 @@ extension RLMObject {
...
-        var i: UInt  = 0
-
-        return GeneratorOf<RLMObject> {
-            if (i >= self.count) {
-                return .None
-            } else {
-                return self[i++] as? RLMObject
-            }
-        }
+        return GeneratorOf(RLMGenerator(collection: self))
```
bdash: Why is this necessary?

Realm/Swift/RLMSupport.swift
```diff
@@ -29,18 +29,29 @@ extension RLMObject {
...
 
+public class RLMGenerator: GeneratorType {
+    private var i: UInt = 0
+    private let collection: RLMCollection
+
+    init(collection: RLMCollection) {
+        self.collection = collection
+    }
+
+    typealias Element = RLMObject
```
segiddins: I think this can be inferred?

RealmSwift/Tests/ObjectCreationTests.swift
```diff
@@ -441,7 +441,7 @@ class ObjectCreationTests: TestCase {
         Realm().commitWrite()
         switch type {
             case .Bool:     return ["invalid", 2 as Int, 1.1 as Float, 11.1 as Double]
-            case .Int:      return ["invalid", true, false, 1.1 as Float, 11.1 as Double]
```
segiddins: why this change?

Realm/Swift/RLMSupport.swift
jpsim: It allows reusing RLMGenerator for both RLMArray and RLMResults, I guess.

segiddins: But why not just return `RLMGenerator`?

RealmSwift/Tests/ObjectCreationTests.swift
```diff
@@ -441,7 +441,7 @@ class ObjectCreationTests: TestCase {
...
+            case .Int:      return ["invalid", 1.1 as Float, 11.1 as Double]
```
jpsim: So because of your changes to `nsnumber_is_like_integer`, Boolean values can now be assigned to `Int` properties? Why?

Realm/Swift/RLMSupport.swift
jpsim: Right, and then the `Element` type in `RLMGenerator` should probably be inferred.

tgoyne: GeneratorOf went away in Swift 2.0, and there doesn't appear to be any equivalent generator from a block in the standard library, so it needs an explicit generator type (which turns out to be less code anyway...). Returning the type-erased generator isn't really necessary; it's just much more natural in 2.0 because it's a protocol (AnyGenerator) rather than a struct wrapping the inner generator.

Realm/Swift/RLMSupport.swift
tgoyne: It wasn't.

RealmSwift/Tests/ObjectCreationTests.swift
tgoyne: In obj-c assigning a boolean value to an int property is perfectly legal and sensible, and NSNumber started actually returning a boolean type from objcType when given boolean boxed in obj-c.

Realm/Swift/RLMSupport.swift
jpsim: This works:

```swift
public class RLMGenerator: GeneratorType {
    private var i: UInt = 0
    private let collection: RLMCollection

    init(collection: RLMCollection) {
        self.collection = collection
    }

    public func next() -> RLMObject? {
        if i >= collection.count {
            return .None
        } else {
            return collection[i++] as? RLMObject
        }
    }
}

extension RLMArray: SequenceType {
    // Support Sequence-style enumeration
    public func generate() -> RLMGenerator {
        return RLMGenerator(collection: self)
    }
}
```

segiddins:
This looks reasonably sane

segiddins:
:+1:

RealmSwift/Tests/ObjectCreationTests.swift
jpsim: Shouldn't we now add `true` and `false` to `validValuesForSwiftObjectType` for Int?

RealmSwift/Tests/PerformanceTests.swift
```diff
@@ -19,8 +19,6 @@
 import XCTest
 import RealmSwift
 
-#if !DEBUG && os(iOS)
```
jpsim: :tada:

jpsim:
:+1:

PR #2067, opened 2015-06-10
Fix a memory leak when constructing standalone Swift objects

### LINKED ISSUE #2064: NSDate properties are leaked
While running an over night test with on our app I noticed a memory leak of NSDate and associated string objects. I managed to reproduce it with the following program:

```
import Foundation
import RealmSwift

class Foo: Object {
    dynamic var date = NSDate()
}

for i in 1...10000000 {
    Foo()
}
```

Ran it with: ```swift -F frameworks/ -framework Realm -framework RealmSwift test.swift```
Running leaks(1) on it shows:
```
Leak: 0x7ffe88fb4ac0  size=16  zone: DefaultMallocZone_0x10f363000   __NSDate  ObjC  CoreFoundation  2015-06-10 10:53:01 +0200
Leak: 0x7ffe88fb4ad0  size=16  zone: DefaultMallocZone_0x10f363000   __NSDate  ObjC  CoreFoundation  2015-06-10 10:53:01 +0200
Leak: 0x7ffe88fb4ae0  size=16  zone: DefaultMallocZone_0x10f363000   __NSDate  ObjC  CoreFoundation  2015-06-10 10:53:01 +0200
...
```

Edit: It does not leak if the NSDate is initialized via timeIntervalSince1970, only leaks with default initialiser.

tgoyne:
Calling `initWithRealm:schema:` from `init` results in the Swift property initialization code running twice and not properly deallocating the values created during the first run, so don't do that (it didn't save any code anyway).

Closes #2064.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,8 @@ x.x.x Release notes (yyyy-MM-dd)
 * Fixed an issue where the packaged OS X Realm.framework was built with
   `GCC_GENERATE_TEST_COVERAGE_FILES` and `GCC_INSTRUMENT_PROGRAM_FLOW_ARCS`
   enabled.
+* Fix a memory leak when constructing standalone Swift objects with NSDate
+  properties.
 
 0.93.1 Release notes (2015-05-29)
 =============================================================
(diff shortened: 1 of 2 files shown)


bdash:
:+1:

segiddins:
👍🏻

PR #2068, opened 2015-06-11
Set the test host for RealmSwift iOS tests so that they can run on-device

tgoyne:
Running the swift tests on-device (and thus the perf test CI job) was broken by #2008.

### DIFF
--- a/RealmSwift.xcodeproj/project.pbxproj
+++ b/RealmSwift.xcodeproj/project.pbxproj
@@ -372,6 +372,9 @@
 					026B0EFB1A780414005E26C8 = {
 						CreatedOnToolsVersion = 6.1;
 					};
+					02B822551A532100007F28A5 = {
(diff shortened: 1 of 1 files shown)


bdash:
`TEST_HOST` should already be set for iOS device builds in [RealmSwift/Configuration/Tests.xcconfig](https://github.com/realm/realm-cocoa/blob/784274fe641361e7292402b3d46c4aad2a8eb211/RealmSwift/Configuration/Tests.xcconfig#L15). Is the value wrong (your patch includes Contents/MacOS in the path, which AFAIK isn't necessary for iOS apps), or is it not taking effect for some reason?

tgoyne:
The executable actually is ending up at `TestHost.app/Contents/MacOS/TestHost` when built for iOS, so that's probably the root cause of the problem. I have no idea what needs to change to fix that, though.

bdash:
I opened PR #2074 with a similar fix that also avoids breaking the Mac tests.

PR #2073, opened 2015-06-11
Validate that the value for an array property is enumerable even if validateNested is false

tgoyne:
Not doing this resulted in an unrecognized selector exception being thrown (and NSLogged) rather than our invalid property value exception being thrown.

### DIFF
--- a/Realm/RLMArray.mm
+++ b/Realm/RLMArray.mm
@@ -107,7 +107,17 @@ static void RLMValidateMatchingObjectType(RLMArray *array, RLMObject *object) {
     }
 }
 
+static void RLMValidateArrayBounds(__unsafe_unretained RLMArray *const ar,
+                                   NSUInteger index, bool allowOnePastEnd=false) {
+    NSUInteger max = ar->_backingArray.count + allowOnePastEnd;
+    if (index >= max) {
+        @throw RLMException([NSString stringWithFormat:@"Index %llu is out of bounds (must be less than %llu).",
+                             (unsigned long long)index, (unsigned long long)max]);
+    }
+}
+
 - (id)objectAtIndex:(NSUInteger)index {
+    RLMValidateArrayBounds(self, index);
     return [_backingArray objectAtIndex:index];
 }
 
@@ -125,15 +135,18 @@ static void RLMValidateMatchingObjectType(RLMArray *array, RLMObject *object) {
 
 - (void)insertObject:(RLMObject *)anObject atIndex:(NSUInteger)index {
     RLMValidateMatchingObjectType(self, anObject);
+    RLMValidateArrayBounds(self, index, true);
     [_backingArray insertObject:anObject atIndex:index];
 }
 
 - (void)removeObjectAtIndex:(NSUInteger)index {
+    RLMValidateArrayBounds(self, index);
     [_backingArray removeObjectAtIndex:index];
 }
 
 - (void)replaceObjectAtIndex:(NSUInteger)index withObject:(id)anObject {
     RLMValidateMatchingObjectType(self, anObject);
+    RLMValidateArrayBounds(self, index);
     [_backingArray replaceObjectAtIndex:index withObject:anObject];
 }
 
--- a/Realm/RLMObjectBase.mm
+++ b/Realm/RLMObjectBase.mm
@@ -95,17 +95,7 @@ static id RLMValidatedObjectForProperty(id obj, RLMProperty *prop, RLMSchema *sc
         // assume our object is an NSDictionary or an object with kvc properties
         NSDictionary *defaultValues = nil;
         for (RLMProperty *prop in properties) {
-            id obj;
-            @try {
-                obj = [value valueForKey:prop.name];
-            }
-            @catch (NSException *e) {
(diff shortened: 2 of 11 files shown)


bdash:
Can this be tested?

tgoyne:
I made it so that the RealmSwift checks which assert that something throws all check the name of the exception thrown, which covers this (it was already checking that an exception was thrown in this case; just not that it was the correct one).

RealmSwift/Tests/ListTests.swift
```diff
@@ -148,7 +148,7 @@ class ListTests: TestCase {
         XCTAssertEqual(str2, array[0])
         XCTAssertEqual(str1, array[1])
 
-        assertThrows(self.array[200])
+        assertThrows(self.array[200], named: nil)
```
bdash: Are we using `named: nil` to indicate that `assertThrows` should not check the exception name? It would be clearer to use the version without the name argument for that purpose.

Realm/RLMObjectBase.mm
```diff
@@ -94,7 +94,17 @@ - (instancetype)initWithValue:(id)value schema:(RLMSchema *)schema {
         // assume our object is an NSDictionary or an object with kvc properties
         NSDictionary *defaultValues = nil;
         for (RLMProperty *prop in properties) {
-            id obj = [value valueForKey:prop.name];
+            id obj;
+            @try {
+                obj = [value valueForKey:prop.name];
```
segiddins: how will this affect performance on 32-bit?

Realm/RLMObjectStore.mm
```diff
@@ -638,7 +638,18 @@ static void RLMValidateValueForObjectSchema(__unsafe_unretained id const value,
         // populate
         NSDictionary *defaultValues = nil;
         for (RLMProperty *prop in objectSchema.properties) {
-            id propValue = [value valueForKey:prop.name];
+            id propValue;
+            @try {
+                propValue = [value valueForKey:prop.name];
+            }
+            @catch (NSException *e) {
+                if ([e.name isEqualToString:NSUndefinedKeyException]) {
```
segiddins: since this is duplicated, can we make this into a helper function?

RealmSwift/Tests/ListTests.swift
tgoyne: The vast majority of cases expect the name to be RLMException, so making that case the least verbose seems preferable to me. Maybe something like `assertThrowsWithoutCheckingName()`?

Realm/RLMObjectBase.mm
tgoyne: It shouldn't (much), since there's a lot of work done in the loop body.

RealmSwift/Tests/ListTests.swift
bdash: Why don't we care which exception is thrown in this case? Are there multiple valid exceptions that could be thrown here? I can see some benefit to having all uses of `assertThrows` check the name if that's possible. Otherwise an alternative name for the non-checking case does seem clearer than passing a nil name.

RealmSwift/Tests/TestCase.swift
```diff
@@ -50,7 +50,12 @@ class TestCase: RLMAutoreleasePoolTestCase {
 
     func assertThrows<T>(@autoclosure(escaping) block: () -> T, _ message: String? = nil, fileName: String = __FILE__, lineNumber: UInt = __LINE__) {
         exceptionThrown = true
-        RLMAssertThrows(self, { _ = block() } as dispatch_block_t, message, fileName, lineNumber)
+        RLMAssertThrows(self, { _ = block() } as dispatch_block_t, "RLMException", message, fileName, lineNumber)
```
segiddins: why not just make `RLMException` the default argument?

segiddins:
@tgoyne I just pushed a pair of commits addressing my comments.

segiddins:
@tgoyne I'm happy to rebase this and merge it if you're happy.

tgoyne:
I still want to eliminate the thing where it passes `nil` to skip the exception name check since that's kinda weird, and was going to do so by eliminating the root problem and adding some redundant checking to standalone RLMArray so that it throws the same exceptions as persisted ones.

tgoyne:
Added more bounds checking and make it more consisted so that RLMArray and RLMArrayLinkView throw the same exceptions (which hopefully are more useful than RLMArrayLinkView's old exceptions), and then removed the ability to not check the exception name in the Swift tests.

Realm/RLMArray.mm
```diff
@@ -107,7 +107,17 @@ static void RLMValidateMatchingObjectType(RLMArray *array, RLMObject *object) {
     }
 }
 
+static void RLMValidateArrayBounds(__unsafe_unretained RLMArray *const ar,
+                                   NSUInteger index, bool allowOnePastEnd=false) {
+    NSUInteger max = ar->_backingArray.count + allowOnePastEnd;
```
segiddins: Just double checking: adding the `bool` here is well-defined?

tgoyne: Yes, `bool` is always 0 or 1 (unlike `BOOL`, which is zero or non-zero).

segiddins:
:+1:

PR #2074, opened 2015-06-11
Fix running the Swift tests on iOS devices

bdash:
My changes in #2008 broke them. Xcode requires that `TEST_HOST` be set for iOS device tests, but the conditional way in which it was being set in Tests.xcconfig was resulting in Xcode thinking it was unset.

To address this `TEST_HOST` is now set unconditionallly. This requires some minor changes to allow TestHost.app to build and run on OS X.

### DIFF
--- a/RealmSwift-swift2.0/Tests/PerformanceTests.swift
+++ b/RealmSwift-swift2.0/Tests/PerformanceTests.swift
@@ -85,7 +85,16 @@ class SwiftPerformanceTests: TestCase {
     }
 
     private func copyRealmToTestPath(realm: Realm) -> Realm {
-        try! NSFileManager.defaultManager().removeItemAtPath(testRealmPath())
+        do {
+            try NSFileManager.defaultManager().removeItemAtPath(testRealmPath())
+        }
+        catch let error as NSError {
+            XCTAssertTrue(error.domain == NSCocoaErrorDomain && error.code == 4)
+        }
+        catch {
+            fatalError("Unexpected error: \(error)")
+        }
+
         try! realm.writeCopyToPath(testRealmPath())
         return realmWithTestPath()
     }
--- a/RealmSwift-swift2.0/Tests/PerformanceTests.swift
+++ b/RealmSwift-swift2.0/Tests/PerformanceTests.swift
@@ -412,17 +412,19 @@ class SwiftPerformanceTests: TestCase {
             let queue = dispatch_queue_create("background", nil)
             let semaphore = dispatch_semaphore_create(0)
             dispatch_async(queue) {
-                let realm = inMemoryRealm("test")
-                let object = realm.objects(SwiftIntObject).first!
-                var stop = false
-                let token = realm.addNotificationBlock { _, _ in
-                    stop = object.intCol == stopValue
+                autoreleasepool {
+                    let realm = inMemoryRealm("test")
+                    let object = realm.objects(SwiftIntObject).first!
+                    var stop = false
+                    let token = realm.addNotificationBlock { _, _ in
+                        stop = object.intCol == stopValue
+                    }
+                    dispatch_semaphore_signal(semaphore)
+                    while !stop {
+                        NSRunLoop.currentRunLoop().runMode(NSDefaultRunLoopMode, beforeDate: NSDate.distantFuture())
+                    }
+                    realm.removeNotification(token)
                 }
-                dispatch_semaphore_signal(semaphore)
-                while !stop {
-                    NSRunLoop.currentRunLoop().runMode(NSDefaultRunLoopMode, beforeDate: NSDate.distantFuture())
-                }
-                realm.removeNotification(token)
             }
 
             dispatch_semaphore_wait(semaphore, DISPATCH_TIME_FOREVER)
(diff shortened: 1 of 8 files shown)


tgoyne:
This won't let me codesign TestHost because it's trying to sign it as 'Mac Developer', and naturally that signing identity isn't part of the provisioning profile for the device.

bdash:
Ugh. I saw that here at one point but thought I'd addressed it. I'll look into it.

bdash:
I was able to reproduce the "Mac Developer" issue on another machine this morning. I saw it after using Xcode's "Fix This" to address a complaint about an outdated provisioning profile. When looking at the Xcode project file I noticed that as part of Xcode's "fixing" it had inserted some references to the "Mac Developer" identity in the Xcode project. Reverting the project file let things build successfully. @tgoyne, any idea if Xcode had made changes to the project file when you were testing this PR?

tgoyne:
That does appear to have been the problem. I also had to delete the TestHost app from the devices, but after doing so I'm able to run both the obj-c and swift tests on device and alternate between them with no issues.

:+1:

jpsim:
:+1:

segiddins:
:+1:

jpsim:
For whatever reason, this PR causes CI's simulator to frequently (though not always) time out when booting, usually when running the `verify-ios-swift`.

Although I struggle to see how the changes made here would have that effect, I haven't seen this issue with any other PR. So I'm hesitant to merge this, given that we've been making progress lately towards making CI more reliable and consistent.

Any ideas about how we could avoid this? What's odd is that the simulator is booted when the CI VM is launched, before any jobs are run.

RealmSwift/Configuration/Tests.xcconfig
```diff
@@ -12,4 +12,4 @@ LD_RUNPATH_SEARCH_PATHS[sdk=macosx*] = $(inherited) @executable_path/../Framewor
 EXCLUDED_SOURCE_FILE_NAMES[sdk=iphone*] = build/osx/*;
 EXCLUDED_SOURCE_FILE_NAMES[sdk=macosx*] = build/ios-dynamic/*;
 
-TEST_HOST[sdk=iphoneos*] = $(BUILT_PRODUCTS_DIR)/TestHost.app/TestHost;
+TEST_HOST = $(BUILT_PRODUCTS_DIR)/TestHost.app/Contents/MacOS/TestHost;
```
segiddins: maybe this should be conditional?

segiddins:
retest this please

RealmSwift/Configuration/Tests.xcconfig
tgoyne: Xcode doesn't support that.

bdash:
Retest this please

bdash:
This is currently failing on CI for the ios-swift tests. The only output it's giving is `** TEST FAILED **`, with no further information.

jpsim:
What does the xcodebuild log show? If it's proving too problematic to read, we could just skip xcpretty altogether

bdash:
I did a run without `xcpretty` earlier this evening. [It doesn't have any extra information](https://ci.realm.io/view/3.%20ObjectiveC/job/objc_pr/lastFailedBuild/configuration=Debug,target=ios-swift/consoleText) that I could see. I may need to look around on the VM while the tests are running in order to make progress.

jpsim:
@bdash `build.sh verify-ios-swift` is failing for me locally too, so you shouldn't need CI to investigate this further.

bdash:
@jpsim If it were failing locally for me I'd not be trying to investigate this via CI…

jpsim:
Marking this as WIP until we can solve the problems with `ios-swift` that this causes on CI.

RealmSwift-swift2.0/Tests/PerformanceTests.swift
```diff
@@ -85,7 +85,16 @@ class SwiftPerformanceTests: TestCase {
     }
 
     private func copyRealmToTestPath(realm: Realm) -> Realm {
-        try! NSFileManager.defaultManager().removeItemAtPath(testRealmPath())
+        do {
+            try NSFileManager.defaultManager().removeItemAtPath(testRealmPath())
+        }
+        catch let error as NSError {
+            XCTAssertTrue(error.domain == NSCocoaErrorDomain && error.code == 4)
```
segiddins: what does this error code correspond to?

tgoyne: I don't know, and I didn't feel it was worth spending the time figuring it out until the bigger problem of things not working at all is solved.

segiddins: cool (cursory googling seems to imply that it somewhat corresponds to `ENOENT`)

segiddins:
This actually seems to have passed the device tests, the only failures had to do with touching the test target failing?

tgoyne:
My changes got it back to the previous state of device tests working but simulator tests never working.

segiddins:
But it looks like the swift 1.2 debug simulator test passed?

jpsim:
retest this please

bdash:
Rebased on top of current master. I'm having Xcode code signing issues at present so haven't been able to test this locally. Fingers crossed that CI is happier now than it has been in the past. If not, I'll investigate further.

jpsim:
Looks like `ios-swift` is still broken with this change. Also, you could ignore Swift 1.2 as these tests will should be run with Xcode 7.

jpsim:
Nice work digging into this. Odds are that running `build test` rather than `test` simply gives the simulator more time to boot.

bdash:
I don't think it has to do with giving the simulator more time to boot. For one, the overall time taken for `build test` is virtually the same as `test`. Secondly, when reproducing locally the simulator isn't launched until the last possible moment when using `build test`, so it actually reduces the amount of lead time available for the simulator boot. The key seems to be that `build test` reduces the amount of time that `test` spends doing work prior to executing the first test.

PR #2075, opened 2015-06-11
Throw an exception rather than asserting when an invalidated object is added to an RLMArray

tgoyne:


Realm/RLMObject_Private.hpp
```diff
@@ -30,3 +32,21 @@
...
+
+// throw an exception if the object can't be modified for any reason
+static inline void RLMVerifyInWriteTransaction(__unsafe_unretained RLMObjectBase *const obj) {
+    // first verify is attached
+    RLMVerifyAttached(obj);
+
+    if (!obj->_realm->_inWriteTransaction) {
+        @throw RLMException(@"Attempting to modify object outside of a write transaction - call beginWriteTransaction on an RLMRealm instance first.");
+    }
+}
```
jpsim: missing a newline here

jpsim:
:+1:

PR #2077, opened 2015-06-12
Update to core 0.89.6

tgoyne:


jpsim:
@tgoyne what do you make of those performance test failures?

tgoyne:
Perf getting worse for insertion is expected, since they're growing the file.

PR #2079, opened 2015-06-12
[RLMSchema] Dont copy the class list

### LINKED ISSUE #2021: Memory issues in notification center widget
I have been experiencing issues with Realm when used in an iOS notification center widget. Widgets have strict limits on memory usage. If they exceed 4096 pages (or 16 MB) of memory, they are terminated by Jetsam.

If I initialize an `RLMRealm` (393 KB) that is in the app group container, the widget is killed due to memory pressure. If I don't initialize it, the widget loads. I am not allocating any `RLMObject` subclasses or initializing any UI. The process does have other objects in memory, but not enough that it should kill the process. Using the realm as readonly does not seem to help.

I know Realm takes advantage of memory mapping, which means that if the process is under memory pressure it should sync any modified pages back to disk and remove them from memory. I see a 5 MB increase in virtual memory usage when a realm is initialized.

Any idea why I might be experiencing this? Does Realm lock any pages into memory?

segiddins:
Closes #2021.

\c @conradev @tgoyne @bdash

### DIFF
--- a/Realm/RLMSchema.mm
+++ b/Realm/RLMSchema.mm
@@ -82,8 +82,9 @@ static NSMutableDictionary *s_localNameToClass;
     NSMutableArray *schemaArray = [NSMutableArray array];
     RLMSchema *schema = [[RLMSchema alloc] init];
 
-    unsigned int numClasses;
-    Class *classes = objc_copyClassList(&numClasses);
+    unsigned int numClasses = objc_getClassList(nullptr, 0);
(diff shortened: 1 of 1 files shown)


tgoyne:
I'm pretty sure that's exactly what `objc_copyClassList` does.

tgoyne:
[Looking at the source](http://www.opensource.apple.com/source/objc4/objc4-646/runtime/objc-runtime-new.mm) it is equivalent to objc_copyClassList.

segiddins:
Oops, my bad. I was sure that it did something different...

PR #2069, opened 2015-06-11
Port to Xcode 7/Swift 2.0

tgoyne:
Closes #2062.

To use this with CocoaPods:
```ruby
# Podfile
pod 'Realm', :git => 'https://github.com/realm/realm-cocoa', :branch => 'swift-2.0'

# Uncomment the following line if you're using RealmSwift
# Note that the Realm pod must be explicitly listed as above even if you're using RealmSwift
# pod 'RealmSwift', :git => 'https://github.com/realm/realm-cocoa', :branch => 'swift-2.0'
```

If you aren't using CocoaPods, clone the repository, checkout the swift-2.0 branch, and then run `sh build.sh ios-static`, `sh build.sh ios-dynamic`, `sh build.sh ios-swift`, `sh build.sh osx`, or `sh build.sh osx-swift` depending on which framework you need. The built framework will be placed in the `build` directory.

PR #2085, opened 2015-06-15
[README] Add an overview section with long-running branches

mrackwitz:
/c @jpsim @timanglade

### DIFF
--- a/README.md
+++ b/README.md
@@ -57,11 +57,11 @@ Here is an overview of some long-running branches we're currently using in devel
 _:warning: Feedback is very welcome for unreleased version, but please aware that we can't provide support to the same
 extent as we do for the official released version._
 
- Branch | Xcode Version | Swift Version | Associated PR for reference
---------|---------------|---------------|-
- [`master`](/realm/realm-cocoa/tree/master)  | 6.3.2 | Swift 1.2 (swiftlang-602.0.53.1 clang-602.0.53) | -
- [`swift-2.0`](/realm/realm-cocoa/tree/swift-2.0) | 7.0.0-beta1 | Swift 2.0 (swiftlang-700.0.38.1 clang-700.0.53) | [#2069](/realm/realm-cocoa/pull/2069)
- [`null-string-beta-2`](/realm/realm-cocoa/tree/null-string-beta-2) | same as `master` | same as `master` | [#628](/realm/realm-cocoa/issues/628)
+| Branch                                                             | Xcode Version    | Swift Version                                   | Associated PR for reference           |
+|:-------------------------------------------------------------------|:-----------------|:------------------------------------------------|:--------------------------------------|
+| [`master`](/realm/realm-cocoa/tree/master)                         | 6.3.2            | Swift 1.2 (swiftlang-602.0.53.1 clang-602.0.53) | -                                     |
+| [`swift-2.0`](/realm/realm-cocoa/tree/swift-2.0)                   | 7.0.0-beta1      | Swift 2.0 (swiftlang-700.0.38.1 clang-700.0.53) | [#2069](/realm/realm-cocoa/pull/2069) |
+| [`null-string-beta-2`](/realm/realm-cocoa/tree/null-string-beta-2) | same as `master` | same as `master`                                | [#628](/realm/realm-cocoa/issues/628) |
 
 If you want to use non-released version of Realm e.g. from the `swift-2.0` branch,
 you can still use your preferred dependency manager. We don't provide prebuilt binaries.
--- a/README.md
+++ b/README.md
@@ -73,8 +73,8 @@ pod 'RealmSwift', :git => 'https://github.com/realm/realm-cocoa.git', :branch =>
 ```Cartfile
 github "realm/realm-cocoa" "null-string-beta-2"
 ```
-Using Realm Swift via Carthage relies on prebuilt binaries currently as [Carthage doesn't offer a
-way to specify the desired scheme to build at the moment](https://github.com/carthage/Carthage/issues/395). This doesn't affect the availability of the regular releases released Realm Swift via Carthage._
+Only regular releases of Realm Swift are available via Carthage. Realm Swift relies on prebuilt binaries currently as [Carthage doesn't offer a
+way to specify the desired scheme to build at the moment](https://github.com/carthage/Carthage/issues/395).
 
 ## Contributing
 
--- a/README.md
+++ b/README.md
@@ -57,10 +57,11 @@ Here is an overview of some long-running branches we're currently using in devel
 _:warning: Feedback is very welcome for unreleased version, but please aware that we can't provide support to the same
 extent as we do for the official released version._
 
- Branch      | Xcode Version | Swift Version | Associated PR for reference
--------------|---------------|---------------|-
- `master`    | 6.3.2         | Swift 1.2 (swiftlang-602.0.53.1 clang-602.0.53) | -
- `swift-2.0` | 7.0.0-beta1   | Swift 2.0 (swiftlang-700.0.38.1 clang-700.0.53) | [#2069](https://github.com/realm/realm-cocoa/pull/2069)
+ Branch | Xcode Version | Swift Version | Associated PR for reference
+--------|---------------|---------------|-
+ [`master`](/realm/realm-cocoa/tree/master)  | 6.3.2 | Swift 1.2 (swiftlang-602.0.53.1 clang-602.0.53) | -
+ [`swift-2.0`](/realm/realm-cocoa/tree/swift-2.0) | 7.0.0-beta1 | Swift 2.0 (swiftlang-700.0.38.1 clang-700.0.53) | [#2069](/realm/realm-cocoa/pull/2069)
+ [`null-string-beta-2`](/realm/realm-cocoa/tree/null-string-beta-2) | same as `master` | same as `master` | [#628](/realm/realm-cocoa/issues/628)
 
 If you want to use non-released version of Realm e.g. from the `swift-2.0` branch,
 you can still use your preferred dependency manager. We don't provide prebuilt binaries.
--- a/README.md
+++ b/README.md
@@ -81,6 +81,10 @@ pod 'RealmSwift', :git => 'https://github.com/realm/realm-cocoa.git', :branch =>
 github "realm/realm-cocoa" "swift-2.0"
 ```
 
+_**Attention**: RealmSwift via Carthage relies on prebuilt binaries currently as [Carthage doesn't offer yet any
+possibility to specify the desired scheme to build](https://github.com/carthage/Carthage/issues/395).  
+This doesn't effect the availability of the regularly released RealmSwift via Carthage._
+
 ## Contributing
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -49,6 +49,35 @@ Run `sh build.sh help` to see all the actions you can perform (build ios/osx, ge
...
+| `master`    | 6.3.2         | Swift 1.2 (swiftlang-602.0.53.1 clang-602.0.53) | -
+| `swift-2.0` | 7.0.0-beta1   | Swift 2.0 (swiftlang-700.0.38.1 clang-700.0.53) | #2069
+
+If you want to use non-released version of Realm e.g. from the `swift-2.0` branch,
+you can still use your preferred dependency manager. We don't provide prebuilt binaries.
+
+### CocoaPods
+
+```ruby
+pod 'RealmSwift', :git => 'https://github.com/realm/realm-cocoa.git', :branch => 'swift-2.0'
```
segiddins: also needs to specify the custom `pod 'Realm'`

tgoyne:
Should note that Carthage can't be used for RealmSwift on a branch (since it only supports installing it from the prebuilt binary).

tgoyne:
:+1:

jpsim:
We should include the null beta in this section too (although the wording will have to change, as it won't strictly be about Apple's beta tools).

README.md
```diff
@@ -49,6 +49,42 @@ Run `sh build.sh help` to see all the actions you can perform (build ios/osx, ge
 
 Executing the examples under the `examples/` folder, requires that you have built the `Realm.framework`.
 
+
+## Branches
+
+Here is an overview of long-running branches, we use in development, which you can use when you want to use Apple's
+latest beta tool releases.
+
+_:warning: Feedback is very welcome for non-released version, but please aware that we can't provide support in the same
```
jpsim: `in the same extent` -> `to the same extent`

README.md
```diff
@@ -49,6 +49,42 @@ Run `sh build.sh help` to see all the actions you can perform (build ios/osx, ge
...
+extent as we do for the official released version._
```
jpsim: `as we do` -> `that we do`

README.md
```diff
@@ -49,6 +49,42 @@ Run `sh build.sh help` to see all the actions you can perform (build ios/osx, ge
...
+
+| Branch      | Xcode Version | Swift Version | Associated PR for reference
+| ------------|---------------|---------------|-
+| `master`    | 6.3.2         | Swift 1.2 (swiftlang-602.0.53.1 clang-602.0.53) | -
+| `swift-2.0` | 7.0.0-beta1   | Swift 2.0 (swiftlang-700.0.38.1 clang-700.0.53) | #2069
```
jpsim: Could you link `#2069`?

README.md
```diff
@@ -49,6 +49,42 @@ Run `sh build.sh help` to see all the actions you can perform (build ios/osx, ge
...
+pod 'RealmSwift', :git => 'https://github.com/realm/realm-cocoa.git', :branch => 'swift-2.0'
+```
+
+### Carthage
+
+```Cartfile
+github "realm/realm-cocoa" "swift-2.0"
+```
+
+_**Attention**: RealmSwift via Carthage relies on prebuilt binaries currently as [Carthage doesn't offer yet any
```
jpsim: `RealmSwift` -> `Realm Swift`

README.md
```diff
@@ -49,6 +49,42 @@ Run `sh build.sh help` to see all the actions you can perform (build ios/osx, ge
...
+possibility to specify the desired scheme to build](https://github.com/carthage/Carthage/issues/395).  
+This doesn't effect the availability of the regularly released RealmSwift via Carthage._
```
jpsim: `effect` -> `affect`, `RealmSwift` -> `Realm Swift`

README.md
```diff
@@ -49,6 +49,42 @@ Run `sh build.sh help` to see all the actions you can perform (build ios/osx, ge
 
 Executing the examples under the `examples/` folder, requires that you have built the `Realm.framework`.
 
+
+## Branches
+
+Here is an overview of long-running branches, we use in development, which you can use when you want to use Apple's
```
segiddins: `Here is an overview of some long-running branches we're currently using in development, which you may wish to use` ?

jpsim:
Thanks for doing this, @mrackwitz! Ping me again once you've addressed my feedback and I'll give it another final look over.

README.md
segiddins: `unreleased versions`

mrackwitz:
>We should include the null beta in this section too (although the wording will have to change, as it won't strictly be about Apple's beta tools).
That's now also addressed.

README.md
```diff
@@ -49,6 +49,42 @@ Run `sh build.sh help` to see all the actions you can perform (build ios/osx, ge
...
+# Attention: Realm's custom branch must be also specified explicitly
+# in your Podfile even if you want to use only Realm Swift directly.
+pod 'Realm',      :git => 'https://github.com/realm/realm-cocoa.git', :branch => 'swift-2.0'
+pod 'RealmSwift', :git => 'https://github.com/realm/realm-cocoa.git', :branch => 'swift-2.0'
+```
+
+### Carthage
+
+```Cartfile
+github "realm/realm-cocoa" "swift-2.0"
```
jpsim: maybe `github "realm/realm-cocoa" "null-string-beta-2"` would be a better example, given Carthage's limitations with Realm Swift.

README.md
```diff
@@ -49,6 +49,42 @@ Run `sh build.sh help` to see all the actions you can perform (build ios/osx, ge
...
 
+
+## Branches
+
+Here is an overview of some long-running branches we're currently using in development, which you may wish to use.
+
+_:warning: Feedback is very welcome for unreleased version, but please aware that we can't provide support to the same
+extent as we do for the official released version._
+
+ Branch | Xcode Version | Swift Version | Associated PR for reference
```
timanglade: Can we format the text so that the chart looks clean in text mode too, please?

README.md
```diff
@@ -49,6 +49,42 @@ Run `sh build.sh help` to see all the actions you can perform (build ios/osx, ge
 
 Executing the examples under the `examples/` folder, requires that you have built the `Realm.framework`.
 
+
+## Branches
+
+Here is an overview of some long-running branches we're currently using in development, which you may wish to use.
+
+_:warning: Feedback is very welcome for unreleased version, but please aware that we can't provide support to the same
```
timanglade: Is this really necessary? Can we wait until we see what kind of requests we get?

mrackwitz: It was more about making sure to direct the feedback directly to the PR and making aware to avoid shipping a prerelease to the AppStore.

README.md
```diff
@@ -49,6 +49,42 @@ Run `sh build.sh help` to see all the actions you can perform (build ios/osx, ge
...
+ Branch | Xcode Version | Swift Version | Associated PR for reference
+--------|---------------|---------------|-
+ [`master`](/realm/realm-cocoa/tree/master)  | 6.3.2 | Swift 1.2 (swiftlang-602.0.53.1 clang-602.0.53) | -
+ [`swift-2.0`](/realm/realm-cocoa/tree/swift-2.0) | 7.0.0-beta1 | Swift 2.0 (swiftlang-700.0.38.1 clang-700.0.53) | [#2069](/realm/realm-cocoa/pull/2069)
+ [`null-string-beta-2`](/realm/realm-cocoa/tree/null-string-beta-2) | same as `master` | same as `master` | [#628](/realm/realm-cocoa/issues/628)
+
+If you want to use non-released version of Realm e.g. from the `swift-2.0` branch,
+you can still use your preferred dependency manager. We don't provide prebuilt binaries.
+
+### CocoaPods
```
timanglade: Not sure we need those CocoaPods/Carthage subsections. Maybe just make that a column of your handy table?

If we run out of horizontal space in that table, might be better to switch the table into a nested list? e.g

```
### Swift 2.0

* Xcode Version: …
* PR: …
* podfile: `…`
* Carffile: `…`
```

README.md
timanglade: Still feels extraneous to me. If you still want to keep it please rewrite the sentence — it’s not proper english at the moment: 

_Warning: we welcome feedback on these branches — please comment on their PRs whenever possible. We strongly recommend against using those branches to ship production applications._

README.md
mrackwitz: I tried it out with a nested list, but as there are pitfalls with both dependency managers with custom branches, it would end up in a lot of duplicated hints and warnings.

timanglade:
I don’t believe this should be included in our README. These instructions should be in our docs. (Our README doesn’t even have installation instructions anymore, so it’s confusing to have installation instructions for long-running branches).

The README should just have a reference to the fact we do offer Swift 2.0 support, with a link to the docs.

README.md
```diff
@@ -49,6 +49,33 @@ Run `sh build.sh help` to see all the actions you can perform (build ios/osx, ge
...
+pod 'Realm',      :git => 'https://github.com/realm/realm-cocoa.git', :branch => 'swift-2.0'
+pod 'RealmSwift', :git => 'https://github.com/realm/realm-cocoa.git', :branch => 'swift-2.0'
+```
+
+### Using Long-running Branches in Carthage
+
+```Cartfile
+github "realm/realm-cocoa" "null-string-beta-2"
+```
+Using Realm Swift via Carthage relies on prebuilt binaries currently as [Carthage doesn't offer a
```
timanglade: Is this really necessary?

tgoyne: We get questions about why Carthage can't be used for things like the null string beta on a regular basis.

timanglade: OK this is note is very confusing then. I read it as meaning “in the back this uses pre-built binary but don’t worry regular releases can be used via Carthage too” which seemed like we were just disclosing an implementation detail. Is that not true?

segiddins: It's an implementation detail _of Carthage_, not of ours

timanglade: Or is that warning intended to indicate you can’t use the null-beta via a Cartfile?

jpsim: No. It means that only official releases of Realm Swift can be installed via Carthage, since Carthage will pull prebuilt binaries for those (which include Realm Swift), but will build from source otherwise (which won't build Realm Swift).

README.md
```diff
@@ -49,6 +49,33 @@ Run `sh build.sh help` to see all the actions you can perform (build ios/osx, ge
...
+pod 'Realm',      :git => 'https://github.com/realm/realm-cocoa.git', :branch => 'swift-2.0'
+pod 'RealmSwift', :git => 'https://github.com/realm/realm-cocoa.git', :branch => 'swift-2.0'
+```
+
+### Using Long-running Branches in Carthage
+
+```Cartfile
+github "realm/realm-cocoa" "null-string-beta-2"
+```
+Only regular releases of Realm Swift are available via Carthage. Realm Swift relies on prebuilt binaries currently as [Carthage doesn't offer a
```
mrackwitz: I fixed a typo and tried to clearify the limitations to address the [discussion, which happened here](https://github.com/realm/realm-cocoa/pull/2085#discussion_r33381541).

jpsim:
thanks Marius, :+1:

PR #2086, opened 2015-06-15
Lower the stack size for notification lister threads

tgoyne:
The notifier listener threads need very little stack, so lower their stack size to the minimum allowed by the OS to cut down on the amount of address space used a bit.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,18 @@
+x.xx.x Release notes (yyyy-MM-dd)
+=============================================================
+
+### API breaking changes
+
+* None.
+
+### Enhancements
+
+* Reduce the amount of memory used by RLMRealm notification listener threads.
+
+### Bugfixes
+
+* None.
+
 0.93.2 Release notes (2015-06-12)
 =============================================================
 
--- a/Realm/RLMRealmUtil.mm
+++ b/Realm/RLMRealmUtil.mm
@@ -211,6 +211,7 @@ public:
 
         NSThread *thread = [[NSThread alloc] initWithTarget:self selector:@selector(listen) object:nil];
         // Use the minimum allowed stack size, as we need very little in our listener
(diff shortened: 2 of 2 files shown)


Realm/RLMRealmUtil.mm
```diff
@@ -209,13 +209,16 @@ - (instancetype)initWithRealm:(RLMRealm *)realm error:(NSError **)error {
         _shutdownReadFd = pipeFd[0];
         _shutdownWriteFd = pipeFd[1];
 
-        [NSThread detachNewThreadSelector:@selector(listen) toTarget:self withObject:nil];
+        NSThread *thread = [[NSThread alloc] initWithTarget:self selector:@selector(listen) object:nil];
+        // Use the minimum allowed stack size, as we need very little in our listener
```
jpsim: I had to dig around a bit to confirm that 16KB is indeed the minimum allowed value, so maybe include a link to https://developer.apple.com/library/ios/documentation/Cocoa/Conceptual/Multithreading/CreatingThreads/CreatingThreads.html#//apple_ref/doc/uid/10000057i-CH15-SW7 in your comment (or more appropriate source if you have one).

CHANGELOG.md
```diff
@@ -1,3 +1,18 @@
+x.xx.x Release notes (yyyy-MM-dd)
+=============================================================
+
+### API breaking changes
+
+* None.
+
+### Enhancements
+
+* Reduce the amount of memory used by RLMRealm notification listener threads.
```
jpsim: Small change, but maybe replace "RLMRealm" with "Realm" to make more language agnostic?

jpsim:
Otherwise :+1:

CHANGELOG.md
tgoyne: The threads are always named "RLMRealm notification listener".

segiddins:
:+1:

tgoyne:
Only CI failure was swift device tests not running.

PR #2087, opened 2015-06-16
Revert "One of the tests of iOS static lib runs on iOS 7"

jpsim:
Reverts realm/realm-cocoa#1898. Looks like there are still issues with this PR (see https://ci.realm.io/job/objc_ios/958/). /cc @segiddins


segiddins:
I thought this was fixed...

jpsim:
So did I, but evidently not so I'm reverting this.

PR #2094, opened 2015-06-16
Support custom error messages/format strings

alazier:
@cmelchior @tgoyne

### DIFF
--- a/Realm/ObjectStore/object_store_exceptions.hpp
+++ b/Realm/ObjectStore/object_store_exceptions.hpp
@@ -23,13 +23,14 @@
 #include <map>
 #include <string>
 
+#define INFO_KEY(key) InfoKey InfoKey##key = "InfoKey" #key;
+
 namespace realm {
     class Property;
 
     class ObjectStoreException : public std::exception {
     public:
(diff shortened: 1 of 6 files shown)


Realm/ObjectStore/object_store_exceptions.hpp
```diff
@@ -23,43 +23,77 @@
 #include <map>
 #include <string>
 
+#define INFO_KEY(key) InfoKey InfoKey##key = "InfoKey" #key;
```
segiddins: this should definitely be `undef`ed after it's used

cmelchior:
I think this looks like an okay approach. No matter how we do it, we will end up with a lot of strings that has been managed somehow, and this way the strings are very readable.

jpsim:
I actually like this overall for consolidating exception messages, keeping things a bit more typed, and moving towards more shared code in the bindings. Well done, @alazier :+1:.

jpsim:
I added an object store style exception for property optionality change. @alazier can you please confirm this was done the right way?

jpsim:
Changed our minds about this design. Closing.

PR #2097, opened 2015-06-17
Disabled testCrossThreadSyncLatency until we can fix it. Currently hangs.

jpsim:
This should be a priority to fix, but I'm disabling it for now so we can run device tests on CI.


segiddins:
I can't seem to get this to fail locally, so I'm unsure how to fix it?

jpsim:
Debug with the device currently plugged into CI?

PR #2109, opened 2015-06-22
Fix property column alignment when copying from the object store

segiddins:
\c @alazier @tgoyne

### DIFF
--- a/Realm/RLMObjectSchema.mm
+++ b/Realm/RLMObjectSchema.mm
@@ -311,6 +311,7 @@
         p.is_indexed = prop.indexed;
         p.is_primary = (prop == _primaryKeyProperty);
         p.is_nullable = prop.optional;
+        p.table_column = prop.column;
         objectSchema.properties.push_back(std::move(p));
     }
     return objectSchema;
@@ -328,6 +329,7 @@
                                                   objectClassName:prop.object_type.length() ? @(prop.object_type.c_str()) : nil
                                                           indexed:prop.is_indexed
                                                          optional:prop.is_nullable];
+        property.column = prop.table_column;
         property.isPrimary = (prop.name == objectSchema.primary_key);
         [propArray addObject:property];
     }
--- a/Realm/Tests/MigrationTests.mm
+++ b/Realm/Tests/MigrationTests.mm
@@ -31,6 +31,16 @@
 #import "object_store.hpp"
 #import <realm/table.hpp>
 
+static void RLMAssertRealmSchemaMatchesTable(id self, RLMRealm *realm) {
+    for (RLMObjectSchema *objectSchema in realm.schema.objectSchema) {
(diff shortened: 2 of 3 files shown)


segiddins:
(test failure is unrelated)

segiddins:
Unless there are any objections, I'd really like to merge this, as migrations are currently broken (see the migration example)

jpsim:
Please just add a regression unit test and changelog entry since it is a bug fix.

segiddins:
> since it is a bug fix
It's only fixing a bug introduced in the C++ object store, so this has never shipped (that was merged post=-0.93.2)

jpsim:
Ok, then a changelog entry isn't necessary, but please add a regression test nonetheless.

Realm/RLMObjectSchema.mm
```diff
@@ -276,6 +276,7 @@ - (void)setTable:(realm::Table *)table {
         p.object_type = prop.objectClassName ? prop.objectClassName.UTF8String : "";
         p.is_indexed = prop.indexed;
         p.is_primary = (prop == _primaryKeyProperty);
+        p.table_column = prop.column;
```
alazier: I don't think this change does anything - everywhere `objectStoreCopy` is used, the column mappings are updated afterwards. This also seems possibly dangerous since we only ever update ObjectSchema instances with column mappings from the table, where here we are copying the mapping in the other direction.

segiddins:
test added

Realm/RLMObjectSchema.mm
segiddins: sure, for now it does nothing, but it does more correctly create the `objectStoreCopy` if you ever were to use a fully-initialized `RLMProperty` as the source

Realm/Tests/MigrationTests.mm
```diff
@@ -322,10 +339,39 @@ - (void)testRemoveAndAddProperty {
...
     realm = [self realmWithTestPath];
+    RLMAssertRealmSchemaMatchesTable(self, realm);
     MigrationObject *mig1 = [MigrationObject allObjectsInRealm:realm][1];
     XCTAssertThrows(mig1[@"oldIntCol"], @"Deleted column should no longer be accessible.");
 }
 
+- (void)testMigrationProperlySetsPropertyColumns {
+    RLMObjectSchema *objectSchema = [RLMObjectSchema schemaForObjectClass:MigrationObject.class];
+    objectSchema.properties = @[
+                                [[RLMProperty alloc] initWithName:@"firstName" type:RLMPropertyTypeString objectClassName:nil indexed:false],
```
jpsim: these RLMProperty's are missing `optional:NO`

segiddins:
Fixed

segiddins:
Closing because https://github.com/realm/realm-cocoa/issues/2142 was merged.

PR #2114, opened 2015-06-23
Remove -[RLMArray deleteObjectsFromRealm]

tgoyne:
The standalone version of this wasn't actually reachable via the public API, and if you did manage to call it, it'd either do nothing, throw an exception, or crash.


segiddins:
Can we add a fallback that checks for `NSFastEnumeration`?

tgoyne:
Should just replace the NSArray version with that.

segiddins:
:+1:

PR #2122, opened 2015-06-24
Make swift tests more stable with regard to leaking realms

### LINKED ISSUE #2120: Swift's testRefresh(), testAutorefresh() and testRealm() occasionally fail
By leaking realms: `RealmSwift/Tests/TestCase.swift:44: error: -[RealmSwift_Tests.RealmTests testRefresh] : XCTAssertNil failed: "Optional(<RLMRealm: 0x103f1d110>)"`. See https://ci.realm.io/job/objc_pr/1009/configuration=Debug,target=osx-swift/console.

segiddins:
Closes #2120.

\c @jpsim @tgoyne

### DIFF
--- a/RealmSwift/Tests/RealmTests.swift
+++ b/RealmSwift/Tests/RealmTests.swift
@@ -395,15 +395,12 @@ class RealmTests: TestCase {
             notificationFired.fulfill()
         }
 
-        let queue = dispatch_queue_create("background", nil)
-        dispatch_async(queue) {
+        dispatchSyncNewThread {
             Realm().write {
                 Realm().create(SwiftStringObject.self, value: ["string"])
-                return
             }
         }
-        waitForExpectationsWithTimeout(2, handler: nil)
-        dispatch_sync(queue) {}
+        waitForExpectationsWithTimeout(1, handler: nil)
         Realm().removeNotification(token)
 
         // get object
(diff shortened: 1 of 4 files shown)


RealmSwift/Tests/RealmTests.swift
```diff
@@ -489,11 +493,13 @@ class RealmTests: TestCase {
...
 
         let fired = expectationWithDescription("fired")
-        dispatch_async(dispatch_queue_create("background", nil)) {
+        let queue = dispatch_queue_create("background", nil)
+        dispatch_async(queue) {
             let otherThreadRealm = Realm()
             XCTAssertFalse(realm == otherThreadRealm)
             fired.fulfill()
         }
         waitForExpectationsWithTimeout(2, handler: nil)
```
tgoyne: The expectation here is pointless with the addition of the `dispatch_sync()` (in the other places it's needed to give the notification block a chance to run).

tgoyne:
The pattern of dispatch_async() then immediately dispatch_sync() the same queue occurs enough that it's probably worth adding a `callOnDifferentThreadAndWait()` helper (preferably with a better name).

RealmSwift/Tests/RealmTests.swift
segiddins: not pointless - it's necessary to check that the expectation has been fulfilled

tgoyne: That's not a thing that needs to be done. The code was previously just using the expectation to wait for the background block to finish.

segiddins: (you're right that it's pointless in this case, but it wasn't in the other 2)

tgoyne: Yes, that's exactly what I said.

jpsim:
Looks good!

PR #2125, opened 2015-06-25
[Results] Make sorting and filtering not eagerly create intermediary views

### LINKED ISSUE #1160: Skip sorting when calling count on an RLMResults that hasn't been constructed yet

segiddins:
Closes #1160 

\c @tgoyne

### DIFF
--- a/Realm/RLMResults.mm
+++ b/Realm/RLMResults.mm
@@ -287,7 +287,7 @@ static std::unique_ptr<RowIndexes::Sorter> RLMSorterFromDescriptors(RLMObjectSch
 }
 
 - (RLMResults *)objectsWithPredicate:(NSPredicate *)predicate {
-    RLMResultsValidate(self);
+    RLMCheckThread(_realm);
 
     // copy array and apply new predicate creating a new query and view
     auto query = [self cloneQuery];
@@ -303,7 +303,7 @@ static std::unique_ptr<RowIndexes::Sorter> RLMSorterFromDescriptors(RLMObjectSch
 }
 
 - (RLMResults *)sortedResultsUsingDescriptors:(NSArray *)properties {
-    RLMResultsValidate(self);
+    RLMCheckThread(_realm);
 
     auto query = [self cloneQuery];
     auto sorter = RLMSorterFromDescriptors(_objectSchema, properties);
--- a/Realm/RLMResults.mm
+++ b/Realm/RLMResults.mm
@@ -30,6 +30,7 @@
 
 #import <objc/runtime.h>
 #import <realm/table_view.hpp>
+#import <realm/views.hpp>
 
 //
 // RLMResults implementation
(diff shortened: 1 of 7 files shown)


jpsim:
Please add a changelog entry and a test confirming that the tableview isn't materialized when checking the count of a sorted RLMResults.

Realm/RLMResults.mm
```diff
@@ -64,7 +65,7 @@ + (instancetype)resultsWithObjectClassName:(NSString *)objectClassName
     ar->_objectClassName = objectClassName;
     ar->_viewCreated = NO;
     ar->_backingQuery = move(query);
-    ar->_sortOrder = sorter;
+    ar->_sortOrder = std::move(sorter);
```
tgoyne: `move` doesn't do anything here (can't move from a const reference)

Realm/RLMResults.mm
```diff
@@ -119,6 +120,14 @@ static inline void RLMResultsValidateInWriteTransaction(__unsafe_unretained RLMR
     }
 }
 
+static std::unique_ptr<RowIndexes::Sorter> RLMSorterFromDescriptors(RLMObjectSchema *schema, NSArray *descriptors)
```
tgoyne: This shouldn't be heap-allocating the sorter.

Realm/RLMResults.mm
```diff
@@ -278,14 +287,14 @@ - (RLMResults *)objectsWhere:(NSString *)predicateFormat args:(va_list)args {
...
-    RLMResultsValidate(self);
+    RLMCheckThread(_realm);
 
     // copy array and apply new predicate creating a new query and view
     auto query = [self cloneQuery];
     RLMUpdateQueryWithPredicate(query.get(), predicate, _realm.schema, _realm.schema[self.objectClassName]);
     return [RLMResults resultsWithObjectClassName:self.objectClassName
                                             query:move(query)
-                                             sort:_backingView.m_sorting_predicate
+                                             sort:std::move(_sortOrder)
```
tgoyne: Don't use `move` here. The Sorter needs to be copied into the new `RLMResults`, not moved from the current one into the new one.

tgoyne:
:+1: 

CI failure appears to just be a CI issue.

PR #2127, opened 2015-06-25
Added RLM_GENERIC_ARRAY() macro

jpsim:
Unfortunately this generates warnings in our unit tests where we test incompatible models, but I've posted https://forums.developer.apple.com/message/16886 to see if there's a way to ignore that. /cc @tgoyne


tgoyne:
Could just add casts to the array's type in the tests.

jpsim:
> Could just add casts to the array's type in the tests.

So obvious, I didn't think of it. Thanks.

tgoyne:
:+1:

PR #2128, opened 2015-06-25
Audit the Objective-C API for nullability

### LINKED ISSUE #1476: Explore adding __nonnull attributes in ObjC
Relies on Xcode 6.3?

bdash:
Add nullability annotations to the Objective-C API to provide enhanced compiler warnings and bridging to Swift.

Closes #1476.

/cc @jpsim @tgoyne @segiddins


Realm/RLMRealm.h
```diff
@@ -169,7 +171,7 @@
 /**
  The RLMSchema used by this RLMRealm.
  */
-@property (nonatomic, readonly) RLMSchema *schema;
+@property (nonatomic, readonly, null_unspecified) RLMSchema *schema;
```
segiddins: pretty sure this isn't nullable?

segiddins:
I think we'll need to conditionally define macros for these things, in case a user is using an older version of Xcode?

bdash:
This all builds fine on OS X 10.10.3 with Xcode 6.3. Do we support older versions of Xcode than that?

Realm/RLMRealm.h
bdash: The commit message for this change (d80e58bd262e6fc9464da5d6541077a7e62d0d22) explains why this is `null_unspecified` rather than `nonnull`.

Realm/RLMObject.h
```diff
@@ -307,7 +309,7 @@
  @return    An object of the subclass type or nil if an object with the given primary key does not exist.
  @see       -primaryKey
  */
-+ (instancetype)objectForPrimaryKey:(id)primaryKey;
++ (nullable instancetype)objectForPrimaryKey:(id)primaryKey;
```
tgoyne: PK is nullable once we enable null strings.

Realm/RLMRealm.h
```diff
@@ -91,7 +93,7 @@
 
  @return An `RLMRealm` instance.
  */
-+ (instancetype)realmWithPath:(NSString *)path readOnly:(BOOL)readonly error:(NSError **)error;
++ (nullable instancetype)realmWithPath:(NSString *)path readOnly:(BOOL)readonly error:(NSError **)error;
```
tgoyne: `error` is nullable.

segiddins:
@bdash well, now there's nothing preventing Realm from being usable in older versions of Xcode, and I'd rather not add an incompatibility when it can be avoided?

tgoyne:
All the non-swift stuff currently works fine in anything 6+.

Realm/RLMRealm.h
bdash: Per [WWDC 2015 session 401](https://developer.apple.com/videos/wwdc/2015/?id=401) (at 18m35s), `NSError**` parameters are assumed to be nullable for both levels within audited regions.

Realm/RLMObject.h
bdash: I assume exactly one object would be allowed to have a null value for the primary key? I'm struggling to think of a scenario in which that would be useful.

tgoyne: Probably not useful, no, but forbidding it when it'd technically work seems weird.

bdash:
Retest this please

Realm/RLMSchema_Private.h
```diff
@@ -32,19 +35,21 @@ extern "C" {
 @property (nonatomic, readwrite, copy) NSArray *objectSchema;
 
 // schema based on runtime objects
-+(instancetype)sharedSchema;
++ (instancetype)sharedSchema;
```
tgoyne: sharedSchema returns nil during initialization of the shared schema (and do we rely on that in `-[RLMObjectBase init]`).

tgoyne:
:+1:

Realm/RLMMigration.h
```diff
@@ -17,12 +17,15 @@
...
+#import <Realm/RLMDefines.h>
+
+RLM_ASSUME_NONNULL_BEGIN
 
 @class RLMSchema;
 @class RLMArray;
 @class RLMObject;
 
-typedef void (^RLMObjectMigrationBlock)(RLMObject *oldObject, RLMObject *newObject);
+typedef void (^RLMObjectMigrationBlock)(RLMObject * __nullable oldObject,  RLMObject * __nullable newObject);
```
segiddins: extra space before `RLMObject`

segiddins:
:+1:

PR #2133, opened 2015-06-26
use leading newline when adding lldb load script

jpsim:
See https://github.com/realm/realm-cocoa/commit/ca5c60981a76838836318bc2afd1601b7d2b3ccd#commitcomment-11883596

/cc @ikonst @tgoyne


tgoyne:
:+1:

Issue #2099, opened 2015-06-18
WatchOS support


MarcoSero:
I don't think there's anything to track the support of watchOS, so I thought of opening this.
#2062 seems to be mostly for Swift 2.

segiddins:
@MarcoSero is there anything in particular realm doesn't support for watchos? I believe @kishikawakatsumi looked into this earlier?

MarcoSero:
From what I can see in the Swift 2 branch, Realm doesn't have a watchOS target and RealmSwift doesn't list watchOS in the list of supported architectures.

segiddins:
We have watchOS support in 0.94.

PR #2139, opened 2015-06-30
Support building framework for watchOS architecture

kishikawakatsumi:
To link the watchOS application or framework and Realm.framework, Realm.framework must support armv7k architecture.

Add a build script and configuration that generates Realm.framework and RealmSwfit.framework including armv7k architecture to support the watch devices in this PR.

It needs to separate the binaries from for iOS binaries because i386 binary for iOS simulator and i386 binary for watchOS simulator could not coexist in the same fat binary.

I think we will have to discuss naming issue:
(e.g.)
`Realm.framework` => `Realm-iOS.framework`
`Realm.framework` => `Realm-WatchOS.framework`

@tgoyne @jpsim @segiddins

https://github.com/realm/realm-cocoa/issues/2099
https://github.com/realm/realm-core/pull/913


tgoyne:
Tagging the framework directory name with the platform doesn't work. We tried to do that for the iOS dynamic framework, and it turned out to require doing things like `#import <Realm-dynamic/Realm-dynamic.h>` for full module support.

What's actually different between the iOS and watchOS versions of the i386 slice?

kishikawakatsumi:
> What's actually different between the iOS and watchOS versions of the i386 slice?

They are completely same when without `-mios-simulator-version-min` or `-mwatchos-simulator-version-min` options.

But iOS simulator slice can not be linked to the framework for watchOS simulator due to following error in my environment.

```
ld: in /Users/katsumi/Desktop/Realm/librealm.a(column-iPhoneSimulator.o), building for watchOS simulator, but linking in object file built for OSX, for architecture i386
clang: error: linker command failed with exit code 1 (use -v to see invocation)
```

As far as I've tried, to link to the slice of each simulator, these options was required:
For iOS simulator: `-mios-simulator-version-min=8.0`
For watchOS simulator: `-mwatchos-simulator-version-min=2.0`

This is why I thought it was necessary to separate the binaries.

tgoyne:
Does specifying both `-mios-simulator-version-min=8.0 -mwatchos-simulator-version-min=2.0` for a single build work? If not we may want to radar that.

kishikawakatsumi:
It is not possible to specify these options at the same time. I've tried it.

jpsim:
@bigfish24 was able to run a watchOS 2 app using Realm natively on the watch, installed via CocoaPods, so therefore without an armv7k architecture for core, using Xcode 7 Beta 2. Maybe there's something wrong with your configuration, @kishikawakatsumi?

bigfish24:
@jpsim yeah that was a mistake on my side. My initial test back during WWDC led to the same conclusion and I hastily rebuilt the project this morning, mixing up the targets, which lead to me thinking it worked.

jpsim:
Ok, it looks like we'll probably need this PR then.

segiddins:
We also should add explicit watchos support to the podspec, although that requires CocoaPods >= 0.38.0.beta.1.

kishikawakatsumi:
The problems are:

##### To build watchOS 2 app/framework, requires `armv7k` architecture
- To support `armv7k` architecture in Realm/RealmSwift framework, the core library also needs to support `armv7k` architecture

##### A possibility need to separate the framework into two
- In Xcode 7, `-mios-simulator-version-min=8.0` option is required to link the simulator slice

If not, the following error occurred
```
ld: in /Users/katsumi/Desktop/LinkTest/librealm.a(test-iPhoneSimulator.o), building for iOS simulator, but linking in object file built for OSX, for architecture x86_64
clang: error: linker command failed with exit code 1 (use -v to see invocation)
```
- Linking to the watch simulator slice, it also needs to be built with `-mwatchos-simulator-version-min=2.0` option
- i386 slice for iOS simulator and i386 slice for watchOS simulator could not coexist in the same fat binary because they are same architecture.
- It is not possible to specify two or more `-m*-version-min=*` options at the same time (Compiler bug? I'll file radar.)
- This is why I thought it needed to separate the framework into two. (if ignore the simulator, it can remain to be one framework)

kishikawakatsumi:
Added `watchos` and `watchos-swift` task to build.sh. They will generate following artifacts that they are Realm.framework and RealmSwift.framework for i386 and armv7k architectures.

```
build
├── DerivedData
├── ios
│   ├── Realm.framework
│   ├── RealmSwift.framework
│   └── swift
│       ├── Realm.framework
│       └── RealmSwift.framework
├── ios-dynamic
│   └── Realm.framework
├── osx
│   ├── Realm.framework
│   └── RealmSwift.framework
└── watchos
    ├── Realm.framework      # Added
    └── RealmSwift.framework # Added
```

Did not add `test-*` tasks because Xcode 7 does not support testing on watchOS yet.

jpsim:
@kishikawakatsumi have you filed a radar for this yet? it would be so much easier if we could share the same prebuilt framework between iOS & watchOS, but we should put pressure on Apple to do this before Xcode 7 GM hits.

kishikawakatsumi:
@jpsim http://www.openradar.me/21724015
If you find any problems, please let me know.

jpsim:
Thank you very much for filing that, @kishikawakatsumi!

jpsim:
Ideally, we'd be able to use a single universal target to cut down on the build setting duplication because it's really getting out of hand with this third platform to support. The biggest hurdle to doing so AFAICT is https://github.com/Carthage/Carthage/issues/340.

kishikawakatsumi:
@segiddins @tgoyne @jpsim Apple closed [the radar](http://www.openradar.me/21724015) due to following comment:

> Apple Developer Relations11-Jul-2015 05:30 AM

> There are no plans to address this based on the following:

> You want one slice of the binary to be watchOS and the other to be iOS. E.g., v7k watchOS and armv7+arm64 iOS. Then when linking, the right one gets pulled out, similar to having a fat library contain both a simulator and a device slice.

> This is not something we want to support.

> We are now closing this report.

> If you have questions about the resolution, or if this is still a critical issue for you, then please update your bug report with that information.

> Please be sure to regularly check new Apple releases for any updates that might affect this issue.

Do you have any ideas?

jpsim:
> Do you have any ideas?

Apple is making a conscious choice here to not support this, so we have to continue with our current direction of having separate binaries for watchOS. :disappointed:

tgoyne:
Recreating the PR to change the target branch.

PR #2142, opened 2015-06-30
Eliminate a redundant call to RLMRealmSetSchemaAndAlign when migrations occur

tgoyne:
I think this is a slightly more correct version of https://github.com/realm/realm-cocoa/pull/2109. This still doesn't fix the root problem of that a lot of ridiculous things are done with the schema in the migrations codepath, but it at least avoids doing the specific bad thing that causes breakage (call RLMRealmSetSchemaAndAlign() on a single target schema multiple times) rather than trying to clean up afterwards.


segiddins:
:+1:

jpsim:
This does look like a more proper fix. :+1:

PR #2144, opened 2015-06-30
Enable documentation warnings and fix a few

tgoyne:
Turns on warnings for invalid documentation comments, removes a few bogus ones, and marks core headers as system headers to suppress warnings from them.


Realm.xcodeproj/project.pbxproj
```diff
@@ -1954,6 +1955,11 @@
 				MACOSX_DEPLOYMENT_TARGET = 10.9;
 				ONLY_ACTIVE_ARCH = YES;
 				OTHER_CFLAGS = "";
+				OTHER_CPLUSPLUSFLAGS = (
+					"$(OTHER_CFLAGS)",
+					"-isystem",
+					core/include,
+				);
```
bdash: Do we still need `core/include` in `HEADER_SEARCH_PATHS` after doing this?

segiddins:
Do those project file changes need to be mirrored in the podspec?

jpsim:
:+1:

tgoyne:
> Do those project file changes need to be mirrored in the podspec?

No point in turning on more warnings for the podspec, and at least in released versions the core headers should always be warning-clean on default settings.

segiddins:
:+1:

PR #1393, opened 2015-01-22
Add more files to private_header_files

tgoyne:
Closes #1388.

PR #2146, opened 2015-07-01
[Podspec] Add RLMDefines.h to public header files

jpsim:
Fixes an issue introduced by #2128 preventing Realm from building via CocoaPods. /cc @segiddins @bdash


bdash:
:+1:

Is there documentation somewhere about why this information needs to be duplicated for CocoaPods?

jpsim:
That was done in #1393 to address #1388.

segiddins:
CocoaPods podspec docs are available at https://guides.cocoapods.org/syntax/podspec.html#public_header_files

PR #2154, opened 2015-07-02
Guarantee 64 bit integer precision for dates even on 32-bit

### LINKED ISSUE #2152: NSDates in the far future are not stored correctly
(Using Realm 0.93.2)

I'm trying to store an expiration date on a subclass of `RLMObject`. So the default `expiresAt` property (*cough* since `nil` in unsupported *cough*) must be a date in the far future. I saw `[NSDate distantFuture]` is not supported in the documentation. So using [DateTools](https://github.com/MatthewYork/DateTools/tree/master/DateTools), I made the default value 500 years in the future.

```
+ (NSDictionary *)defaultPropertyValues{
	return @{
			 @"expiresAt" : [[NSDate date] dateByAddingYears:500]
			 };
}
```

But after it has been committed to the Realm:

```
expiresAt = 1901-12-13 20:45:52 +0000;
```

Huh. Well, that's fine, maybe 100 years in the future?

```
expiresAt = 1901-12-13 20:45:52 +0000;
```

So how many years in the future is supported? 25 doesn't work. 20 does. This is a huge limitation. I knew about no second precision, but this is not documented at all. Are there other limitations in your implementation of storing dates that are not documented? Why don't you store lossless dates?

segiddins:
Closes #2152.
Depends upon https://github.com/realm/realm-core/pull/922.

\c @tgoyne

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,7 +14,8 @@ x.xx.x Release notes (yyyy-MM-dd)
 
 ### Bugfixes
 
-* None.
+* Fix incorrect results when querying for < or <= on ints which
+  requires 64 bits to represent with a CPU that supports SSE 4.2.
 
 0.93.2 Release notes (2015-06-12)
 =============================================================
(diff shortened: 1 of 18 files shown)


segiddins:
@tgoyne this has now been merged into core

tgoyne:
Looks fine, but needs a core release before it's mergeable.

segiddins:
That's why I pointed this to the `ks-core-api-changes` branch

jpsim:
Shouldn't this use the same type as defined in core (`int_fast64_t`)?

segiddins:
@jpsim I don't think it really matters, given `int64_t` will be defined on every arch we build for

jpsim:
It matters in keeping the types consistent. Do you have a compelling reason to use `int64_t`?

tgoyne:
Needs to be recreated targeting master.

PR #2167, opened 2015-07-06
Add a test for correct updating of sorted Results after insertion

### LINKED ISSUE #2166: After updating database, Results sorted incorrectly if sort is applied before filter
Running: RealmSwift v.93.2

I discovered this after running some Unit Tests on my own code. What I did was grab a Results object by applying a sort and then applying a filter:

    let models = realm.objects(ModelA)
      .sorted("doubleField", ascending: true)
      .filter("%K CONTAINS %@", "stringField", par)

Initially, there are no objects in the database, so `models` has nothing in it.  Once I write objects to the database, the objects are filtered correctly but they are not sorted correctly.

If I change the order of chain it works properly:

    let models = realm.objects(ModelA)
      .filter("%K CONTAINS %@", "stringField", par)
      .sorted("doubleField", ascending: true)

Note: This *only* happens after I update the database.  If I first write all the objects to the database, the Results object will be correctly sorted/filtered in both cases (no matter how I order the chain).  But once I update the database the sort is incorrect (if it's before the filter in the chain).

Full Disclosure:  Here's my model (for testing):

```
public class ModelA: Object {
  public dynamic var id: String = ""
  public dynamic var stringField: String? = ""
  public dynamic var dateField: NSDate? = NSDate()
  public dynamic var dataField: NSData? = NSData()
  public dynamic var int64Field: Int64 = 0
  public dynamic var boolField: Bool = false
  public dynamic var doubleField: Double = 0
  public dynamic var floatField: Float = 0
  
  public override static func primaryKey() -> String? {
    return "id"
  }
}
```

I'm also generating my ModelA objects using randomized values and write them:
```
  public func newModelA() -> ModelA {
    let model = ModelA()
    model.id = newID()
    model.stringField = randomStringWithLength(1) as String
    model.dateField = NSDate()
    model.int64Field = Int64(arc4random())
    model.boolField = Bool(Int(arc4random_uniform(1)))
    model.doubleField = Double(arc4random()) / Double(arc4random())
    model.floatField = Float(arc4random()) / Float(arc4random())
    return model
  }

  public func generateModelAs(no: Int) -> [ModelA] {
    var models:[ModelA] = []
    for i in 0..<no {
      models.append(newModelA())
    }
    return models;
  }

  public func writeModels(models: [ModelA]) {
    realm.beginWrite()
    self.realm.add(models, update: true)
    realm.commitWrite()
  }

```
I retrieve the Results like so:
```
  public func getAllModelsFirstSortedThenFilteredOnDouble(par:String) -> Results<ModelA> {
    let models = realm.objects(ModelA)
      .sorted("doubleField", ascending: true)
      .filter("%K CONTAINS %@", "stringField", par)
    
    return models
  }
```
I run the following test:
```
  func testSortedThenFiltered() {
    let models = persistence!.generateModelAs(1000)
    let queried = persistence!.getAllModelsFirstSortedThenFilteredOnDouble("a")
    
    self.persistence!.writeModels(models)

    var last: ModelA? = nil
    for object in queried {
      if let lastObject = last {
        XCTAssert(object.doubleField >= lastObject.doubleField, "Object Double '\(object.doubleField)' !>= Last '\(lastObject.doubleField)'")
      }
      XCTAssertEqual(object.stringField!, "a", "Object should have string field == a")
      last = object
    }
  }
```

Note: `persistence` is where the models are generated and retrieved.

tgoyne:
This passes on master but was broken in 0.93.2 and didn't have any tests covering it. Closes #2166.

### DIFF
--- a/Realm/Tests/ResultsTests.m
+++ b/Realm/Tests/ResultsTests.m
@@ -460,6 +460,38 @@
     XCTAssertFalse((bool)object_getIvar(filtered, ivar));
 }
 
+- (void)testRerunningSortedQuery {
+    RLMRealm *realm = [RLMRealm defaultRealm];
+
+    RLMResults *sortedAge = [[EmployeeObject allObjects] sortedResultsUsingProperty:@"age" ascending:YES];
+    [sortedAge lastObject]; // Force creation of the TableView
+    RLMResults *sortedName = [sortedAge sortedResultsUsingProperty:@"name" ascending:NO];
+    [sortedName lastObject]; // Force creation of the TableView
+    RLMResults *filtered = [sortedName objectsWhere:@"age > 20"];
+    [filtered lastObject]; // Force creation of the TableView
+
+    [realm beginWriteTransaction];
+    [EmployeeObject createInRealm:realm withValue:@{@"name": @"A",  @"age": @20, @"hired": @YES}];
+    [EmployeeObject createInRealm:realm withValue:@{@"name": @"B", @"age": @30, @"hired": @NO}];
+    [EmployeeObject createInRealm:realm withValue:@{@"name": @"C", @"age": @40, @"hired": @YES}];
+    [realm commitWriteTransaction];
+
+    XCTAssertEqual(3U, sortedAge.count);
+    XCTAssertEqual(3U, sortedName.count);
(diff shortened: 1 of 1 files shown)


segiddins:
:+1:

Issue #2157, opened 2015-07-04
Realm instance persisting unexpectedly


ezwu:
Hello,

I'm developing an app that uses Realm in conjunction with user accounts. I've recently added this code for migration:
```
RLMRealm setSchemaVersion:1
                        forRealmAtPath:[GGRealmUtils realmPath]
                    withMigrationBlock:^(RLMMigration *migration, uint64_t oldSchemaVersion) {
        //migration code
}
[RLMRealm realmWithPath: realmPath]]
```
The app itself is user account based, so each realmPath is unique to the userId, and I can't just put this migration code in appdelegate since if a user logs out on their phone and has someone else log in, I would need to migrate for the new person as well, so I put it in my login controller and trigger it whenever someone logs in using the associated realmPath.

The issue I am having arises from writing to the user realm a second time while they are logged in, basically. Upon log in I make my first write transaction as follows:
```
// Instantiate realm object
RLMRealm *realm = [self realmForCurrentUser];
	
    [realm beginWriteTransaction];
    // write some objects using [realm addObject:]
    [realm commitWriteTransaction];
```
And at a certain point in the app after the user has done something I need to make another write transaction, so I run a variation on that same code again. 

However, through testing and logging I've seen that the first realm instance gets deleted by ARC automatically as soon as the scope of my write transaction ends, but the second time I do it it persists for some reason and ARC doesn't clean it up, even though I don't have any references to it in my code. And so, if the user tries to log out and log back in, the migration code will through an exception:
```Cannot set schema version for Realms that are already open.```
because the realm instance is still cached.

Is there a method I can use to manually clear the realm cache? I'm very certain that I don't have any references to the realm (my code is pretty much the exact same for both the first and second write, minus the different objects to write), so why is the realm instance not being deleted from the cache?

segiddins:
@ezwu there's no publicly exposed way to manually clear the Realm cache, but it's really not something you should ever have to do. If there's already a Realm open, I suspect something still has a strong reference to it, and we'd be happy to have a look at help you debug that, either here or privately at help@realm.io.

If I could offer a suggestion, however, I'd advise just calling `setSchemaVersion:ForRealmAtPath:withMigrationBlock:` for every `.realm` file in the directory you store the Realm files in at program launch -- that way, there would be no possibility of trying to set the schema version for a Realm that is already open.

ezwu:
[...]

PR #2168, opened 2015-07-06
[RLMRealm] Allow setting the same encryption key / schema version whe…

segiddins:
…n a realm is open

Closes https://github.com/realm/realm-cocoa/issues/2157.

\c @tgoyne @jpsim

### DIFF
--- a/Realm/RLMRealm.mm
+++ b/Realm/RLMRealm.mm
@@ -458,14 +458,16 @@ static id RLMAutorelease(id value) {
 }
 
 + (void)setEncryptionKey:(NSData *)key forRealmsAtPath:(NSString *)path {
-    if (RLMGetAnyCachedRealmForPath(path)) {
-        NSData *existingKey = keyForPath(path);
-        if (!(existingKey == key || [existingKey isEqual:key])) {
-            @throw RLMException(@"Cannot set encryption key for Realms that are already open.");
+    @synchronized (s_keysPerPath) {
+        if (RLMGetAnyCachedRealmForPath(path)) {
+            NSData *existingKey = keyForPath(path);
+            if (!(existingKey == key || [existingKey isEqual:key])) {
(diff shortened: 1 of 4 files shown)


tgoyne:
Test failure isn't a false-positive.

segiddins:
Should be fixed now

Realm/RLMRealm.mm
```diff
@@ -792,7 +795,7 @@ + (void)setDefaultRealmSchemaVersion:(uint64_t)version withMigrationBlock:(RLMMi
 }
 
 + (void)setSchemaVersion:(uint64_t)version forRealmAtPath:(NSString *)realmPath withMigrationBlock:(RLMMigrationBlock)block {
-    if (RLMGetAnyCachedRealmForPath(realmPath)) {
+    if (RLMGetAnyCachedRealmForPath(realmPath) && schemaVersionForPath(realmPath) != version) {
```
tgoyne: This check needs to be within the `@synchronized` block below to avoid a race condition (that'd require really weird code to hit).

segiddins: Does the same apply to the other check as well?

tgoyne: No. The race is that the result of `schemaVersionForPath()` could change after it's checked if it's not in the synchronized block, but neither `version` nor `NotVersioned` can change.

segiddins: I meant the check in set encryption key

tgoyne: Yes.

segiddins:
@tgoyne this should be slightly less race-y now.

tgoyne:
:+1:

segiddins:
Ping @jpsim ?

jpsim:
needs a changelog entry, but otherwise :+1:

segiddins:
CHANGELOG added

segiddins:
test failure is unrelated. merging.

Issue #2159, opened 2015-07-06
Crashing during a write transaction makes other processes "hang" indefinitely


conradev:
![screen shot 2015-07-04 at 7 32 04 pm](https://cloud.githubusercontent.com/assets/170649/8514307/19cf1020-2341-11e5-8d4b-536d7efef6a7.png)

If our WatchKit extension crashes while a write transaction is open, the lock is never released. Our main app then hangs. Is this a known/expected issue?

segiddins:
I believe this is to be expected, as we're using a `pthread_mutex` under the hood, which is not robust against crashing processes.

tgoyne:
Yes, this is expected. POSIX has robust mutexes that support recovering from a process crashing while the mutex is locked, but Apple hasn't implemented them. We've looked into some workarounds, but we haven't found any particularly reliable options. The mutex is recreated when a Realm file is opened and no other threads/processes have it opened already.

jpsim:
#2169 should resolve this.

tgoyne:
Only very partially. There's a lot of ways for a process to crash other than an uncaught exception.

bdash:
@tgoyne, do you recall what workarounds we looked at for this?

tgoyne:
The ideas I remember are storing the pid of the locking process and doing ~something~ if a process with that pid doesn't exist, using `flock()` for the write lock rather than a mutex, and using a named pipe rather than a mutex.

jpsim:
Moved to P2 because we don't think we can resolve this in the short term, along the same level of priorities as the rest of our P1 issues.

conradev:
That's unfortunate – an enable-able workaround would be an ideal solution for our use case. I'm not sure how hard it would be to get an alternate locking system to work reliably, but I would definitely enable "slightly slower, but more robust" locking.

If I decide to kill Workflow from the app switcher or Jetsam decides to kill one of our extensions and the process happens to be modifying the database at the time, I have to reboot my device in order to use Workflow. Granted, it's rare, but we do get support emails about it.

bdash:
Out of the options @tgoyne mentions, `flock` seems most directly suited to this purpose. I can't see any obvious downsides to it, except the potential for worse performance compared with our current use of mutexes.

conradev:
[...]

PR #2169, opened 2015-07-06
Install an uncaught exception handler to close write transactions whe…

segiddins:
…n possible

Might help out in a few situations where https://github.com/realm/realm-cocoa/issues/2159 is an issue.

\c @tgoyne @bdash

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,7 +14,8 @@ x.xx.x Release notes (yyyy-MM-dd)
 
 ### Bugfixes
 
-* None.
+* Processes crashing due to an uncaught exception during a write transaction will
+  no longer cause other processes using the same Realm to hang indefinitely.
 
 0.93.2 Release notes (2015-06-12)
 =============================================================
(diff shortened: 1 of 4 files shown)


jpsim:
I can't tell why you'd only want to cancel the write transactions for the thread the exception was thrown from rather than from all cached realms...

Overall, this PR makes me very uneasy and feels like much more of a hack than addressing the root cause (which may be unfeasible to address given the state of mutexes on Apple platforms).

tgoyne:
We'd prefer to cancel all of them, but trying to cancel the ones for other threads has a good chance of causing another crash that hides the original problem.

jpsim:
It also worries me that this is hard to test, but I don't think there's much overhead in installing this handler, and the process is already (hopefully) like to be terminated by the point this is hit, so I think this is safe to do. :+1: 

You could add a changelog "bugfix" entry along the lines of:

> Processes crashing due to an uncaught exception during a write transaction will longer cause other processes using the same Realm to hang indefinitely.

segiddins:
CHANGELOG added

jpsim:
:+1:

segiddins:
retest this please

tgoyne:
:+1:

PR #2170, opened 2015-07-06
Use a frozen copy of RLMResults/RLMArray for fast enumeration

### LINKED ISSUE #1152: RLMResults and RLMArray should throw if mutated during enumeration
Right now the behavior is undefined and can result in a crash

tgoyne:
Makes enumerating and modifying actually work, at the cost of having to create a copy of the TableView, which can be somewhat expensive. In theory it should be possible to use the KVO machinery to lazily do the copy only when needed. Alternatively, we could require that users call `freeze` (or maybe some better name) on the results if they want to be able to enumerate-and-mutate.

Needs a new core release and similar changes for standalone RLMArray before it's mergeable.

Closes #1152.

### DIFF
--- a/Realm/RLMResults.mm
+++ b/Realm/RLMResults.mm
@@ -33,6 +33,62 @@
 
 using namespace realm;
 
+static const int RLMEnumerationBufferSize = 16;
+
+@implementation RLMFastEnumerator {
+    id _strongBuffer[RLMEnumerationBufferSize];
+    realm::TableView _tableView;
+    RLMRealm *_realm;
+    RLMObjectSchema *_objectSchema;
+
+}
+
+- (instancetype)initWithTableView:(realm::TableView)tableView realm:(RLMRealm *)realm objectSchema:(RLMObjectSchema *)objectSchema {
+    self = [super init];
+    if (self) {
+        _tableView = std::move(tableView);
+        _realm = realm;
+        _objectSchema = objectSchema;
+    }
+    return self;
+}
+
+- (NSUInteger)countByEnumeratingWithState:(NSFastEnumerationState *)state
+                                    count:(NSUInteger)len {
+    RLMCheckThread(_realm);
+    if (!_tableView.is_attached()) {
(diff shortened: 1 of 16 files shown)


segiddins:
This needs a rebase now

segiddins:
I added frozen enumeration for standalone arrays

Realm/RLMResults.mm
```diff
@@ -33,6 +33,103 @@
...
+- (void)detach {
+    _tableView = [_collection tableView];
+    _collection = nil;
+}
+
+- (NSUInteger)countByEnumeratingWithState:(NSFastEnumerationState *)state
+                                    count:(NSUInteger)len {
+    RLMCheckThread(_realm);
+    if (!_tableView.is_attached() && !_collection) {
+        @throw RLMException(@"RLMResults is no longer valid");
```
segiddins: Couldn't this also be an `RLMArray` now?

segiddins:
:+1: other than the one comment

segiddins:
Would be nice if we could allow the following test case to pass as well

```diff
diff --git a/RealmSwift-swift2.0/Tests/ResultsTests.swift b/RealmSwift-swift2.0/Tests/ResultsTests.swift
index 1295ea1..1d0d067 100644
--- a/RealmSwift-swift2.0/Tests/ResultsTests.swift
+++ b/RealmSwift-swift2.0/Tests/ResultsTests.swift
@@ -284,6 +284,16 @@ class ResultsTests: TestCase {
         XCTAssertEqual(str, "12")
     }
 
+    func testGeneratorLifetime() {
+        let generator = results.generate()
+        autoreleasepool {
+            XCTAssertEqual(generator.next()?.stringCol, "1")
+        }
+        // Should not segfault
+        XCTAssertEqual(generator.next()?.stringCol, "2")
+        XCTAssertNil(generator.next())
+    }
+
     func testArrayAggregateWithSwiftObjectDoesntThrow() {
         let results = getAggregateableResults()
```

tgoyne:
Will address #2384 in a separate PR.

PR #2171, opened 2015-07-07
[Realm Browser Plugin] Open relevant realm file on Realm Browser

kittinunf:
A week ago, I have created an Xcode plugin to open Realm Browser with relevant Realm File. Little did I know that you guys have created already this plugin to open Realm Browser but not with file.

Thus, by recommendation of @jpsim in the issue (https://github.com/kittinunf/RealmBrowser-Plugin/issues/1), I think it is a nice idea to extend the functionality of Realm Browser plugin itself by combining my implementation into yours. This PR is to serve this purpose, and ultimately deprecate my plugin once it is merged. 

I intentionally break things into a multiple commits just for reviewer to easier review my code. 

All comments are appreciated. :sunny:


TimOliver:
Hi there Kittinun!

Awesome! Thank you so much for this! As someone who debugs Realm files in a production app, this sort of thing is incredibly useful for me!

That being said, I think I found a bit of a problem, although it might be an issue on Realm Browser's side that I could potentially take care of.

I just tried using the plugin on my own copy of Xcode, and this happened:
realmbrowseroverload

One of my apps uses Realm in a pretty crazy way. It has a standard default.realm file where all of the main data is kept, and then for each user-generated file in the app, it has a subsequent .realm file that stores cached information about each file. When I opened the Browser via the plugin, it looks like it was instructed to open every single .realm file in my Simulator device's directory, including all the cache files.

In any case, it might be best if I include some safeguards in Realm Browser for these kinds cases (Maybe a prompt that appears when a certain number of Realm files are opened at once)

/cc @jpsim What do you think?

kittinunf:
@TimOliver, I wouldn't think of that edge case. 

Yeah, you are very right. I way I am able to grasp the Realm file is to recursively find all of `.realm` file extension and open them all. It was kind of blind-fully because I don't have any info about how Realm Browser is opening the file or what file(s) the Browser is(are) actively opening. 

It works well and effectively in my use case (I have 2 apps in my Simulator with 2 default Realm files in 2 folders), but it seems like it might be not the case for you. :scream: 

I am very happy to add more guard on my side if there is anything I can do.

TimOliver:
@kittinunf Fair enough. I've been told it's not a normal way to use Realm (But it still works very well!). :)

At the very least, do you know if it's possible to open up the `.realm` files for the app that the Simulator is currently displaying (That would mitigate it a fair amount)? If not, I might do a bit of digging to have a look, but if it's not possible then it can't be helped.

No worries! Let's see what JP says, but I get the feeling that this should be a safeguard feature of the Browser itself to prevent too many files from opening at once.

plugin/RealmPlugin/RLMPSimulatorManager.m
```diff
@@ -0,0 +1,95 @@
...
+            }
+        }];
+    }
+    
+    return bootedDeviceUUID;
+}
+
++ (NSString *)readDeviceData {
+    // Set parameters to get device detail
+    NSString *path = @"/usr/bin/xcrun";
```
jpsim: This could potentially get the iOS Simulator directories from a different version of Xcode than what's currently running if another version is set via `xcode-select`

jpsim:
@kittinunf thanks for the PR :sparkles:! I look forward to getting this into the official Realm plugin.

@TimOliver I'd add a prompt to the browser when it's asked to open more than 5 files simultaneously (`Are you sure you want to open 'X' Realm files at once?`) to address this in the very short term.

plugin/RealmPlugin/RLMPSimulatorManager.m
kittinunf: Ah that's true. What would you think it should be our plugin could handle? Let's say we have 2 versions of Xcode installed at `/Applications/Xcode.app/Contents/Developer` and `/Applications/Xcode Beta.app/Contents/Developer`, one way I could think of is to set `xcode-select -switch ` to current running Xcode path. 

I might feel a bit weird as a user, when the plugin switches default path automatically behind our back? Perhaps, I could save the existing one first, then use `xcode-select -switch` then switch it back later.

plugin/RealmPlugin/NSFileManager+GlobAdditions.h
```diff
@@ -0,0 +1,34 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+////////////////////////////////////////////////////////////////////////////
+
+#import <Foundation/Foundation.h>
+
+@interface NSFileManager (GlobAdditions)
```
segiddins: Is there any reason this category (and the others) couldn't be written as functions instead of category methods? Adding categories on classes you don't own is inherently unsafe, and I prefer to stay away from that whenever possible.

kittinunf: No. They are just only conveniences. 

They are written as categories because the existing interfaces are too clumsy to work with and they are just extensions of their respective classes. I thought it would be nice to group things in their responsibilities so I wrote it that way.

If you feel strongly against this, I can change them into static functions in `RLMPRealmPlugin` itself.

kittinunf: Fix by 2b53e16

plugin/RealmPlugin/RLMPSimulatorManager.m
jpsim: since we're using private Xcode plugin APIs here anyway, could you not find out the path of the currently running Xcode app to get the relevant `simctl` program directly rather than through `xcrun`?

kittinunf: Address by c470530ca912438d6e23a98cd15ff8cba4041489

jpsim: Thanks!

jpsim:
I'll let @TimOliver continue to review this, but it looks fine to me @kittinunf. Thanks for your hard work!

jpsim:
@TimOliver thoughts?

TimOliver:
I'll test it with my Browser changes once more to be sure!

TimOliver:
Yep! It looks good on this end!

One final thing I might recommend is that even though I added some checks on the Browser side to ensure not too many files are opened at once, it might also be good to add the same check here in advance. That can wait until after the merge though.

In any case, I'm happy for the merge to go ahead!

jpsim:
Thank you so much for this PR, @kittinunf ! I hope you'll contribute to realm again in the future :smile:.

kittinunf:
@jpsim, I am more than happy to contribute. As I am a Realm fan myself, we use Realm in our iOS and Android (for Realm-Java, it is kind of love and hate along the process, haha) so a chance to contribute back to Realm eventually benefits ourself and Realm community. :dancers:

timanglade:
@kittinunf :) Is there anything in particular we can do better on the Java side? It’s been a long road, but we have some exciting stuff coming up there in the immediate future including _much_ improved Realm Objects (and of course null support, async queries/transactions and better notifications). Anything else tripping you up?

kittinunf:
@TimOliver I can open another simple PR to take care of that. It could be as simple as an Alert Confirmation with yes/no buttons. If yes, open with that many files, and no for doing nothing. How does that sound to you? If it sounds good, I will craft it over the weekend.

@timanglade, we have been fighting with lack of null support, unable to implement common interface on RealmObject, and gson serializing/deserializing with Realm Proxy class. We tried so hard, because we like to Realm with our Android so much and don't want to give up on it just yet. Don't take me wrong though, you guys makes an awesome DB.

TimOliver:
@kittinunf Yeah! If you can spare the time to do that, that would be a great help! Thank you so much for that! :)

TimOliver:
@kittinunf Actually, you know what? There might be a smarter way to go about doing this.

My check for multiple files in Realm Browser is in the `[NSApplication application:openFiles:]` method, but in my last set of tests, it looks like the Plugin isn't passing the files to the Browser through that delegate, more that it's most likely calling 'openFile' multiple times. 

Instead of doubling up on the logic for checking if multiple files are being passed, it might be better to ensure that the Realm Plugin is passing multiple Realm files to Realm Browser via the `openFiles:` delegate method.

Do you know anything about that? If not, then I can easily spend some time on this.

kittinunf:
@TimOliver Ah I guess, it might be my fault because I always use `NSWorkspaceLaunchNewInstance` when open file. Perhaps I need to use `NSWorkspaceLaunchDefault` or `0`, then it should help, I guess?

TimOliver:
@kittinunf Yeah, I was looking into it as well and it looks like there's a lot of different ways to go about doing this.

Apparently `[NSWorkspace openURLs:withAppBundleIdentifier:options:additionalEventParamDescriptor:launchIdentifiers:]` might do the trick. Shall I have a look into that one? :)

timanglade:
@kittinunf Understood! We feel the same honestly. You’ll really like the new object stuff we are working on now :)

kittinunf:
@TimOliver let's discuss on another thread. I will open up another PR to take care of this.

PR #2175, opened 2015-07-07
Set up Google Analytics for Swift API docs

jpsim:
All CI workers have been updated to jazzy 0.2.2, so this should be supported by CI immediately. /cc @yoshyosh @timanglade


timanglade:
lgtm :+1:

PR #2177, opened 2015-07-08
Add `-[RLMResults limit:]` to return an `RLMResults` that contains at most the given number of objects

bdash:
Depends on https://github.com/realm/realm-core/pull/933.

/cc @tgoyne @jpsim


tgoyne:
It should be possible to do this without the eager materialization. I think all you need is a version of `Query::find_all()` which doesn't actually call `find_all()` (i.e. just something that exposes the private `TableView(Table& parent, Query& query, size_t start, size_t end, size_t limit)` constructor).

bdash:
I've not had a chance to revisit this. I'm going to close the PR since the approach I took has the drawback of eagerly materializing results, and a version without may look quite different.

PR #2199, opened 2015-07-13
Add support for location-based bounding box queries

bdash:
A few outstanding points:
* Does it make sense to add these methods directly on `RLMObject` / `RLMResults`? Since internally they just build `NSPredicate` they could be exposed as functions that produce `NSPredicate`. Alternatively, they could live in separate location-related categories on `RLMObject` / `RLMResults`?
* Is it reasonable to use CoreLocation's coordinate type in our API, or should we define our own?
* What should a Swift API look like? Querying in our Swift API is currently via a single `filter` method. There's no obvious extrapolation like there is with the Objective-C API.

### DIFF
--- a/Realm/RLMLocationTypes.h
+++ b/Realm/RLMLocationTypes.h
@@ -1,0 +1,27 @@
+////////////////////////////////////////////////////////////////////////////
+//
+// Copyright 2015 Realm Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+////////////////////////////////////////////////////////////////////////////
+
+#import <CoreLocation/CoreLocation.h>
+
+/**
+ A bounding box is rectangle that is represented by coordinates for opposing corners.
+ */
+typedef struct {
+    CLLocationCoordinate2D corner1;
+    CLLocationCoordinate2D corner2;
+} RLMBoundingBox;
--- a/Realm/RLMLocationTypes.h
+++ b/Realm/RLMLocationTypes.h
@@ -16,12 +16,24 @@
 //
 ////////////////////////////////////////////////////////////////////////////
 
-#import <CoreLocation/CoreLocation.h>
+/**
+ RLMDegrees represents an angle in degrees. It is used to represent latitudes and longitudes,
+ with positive numbers representing North and East, and negative numbers representing South and West.
+ */
+typedef double RLMDegrees;
+
+/**
+ RLMCoordinate2D represents a geographical coordinate.
+ */
+typedef struct {
+    RLMDegrees latitude;
+    RLMDegrees longitude;
+} RLMCoordinate2D;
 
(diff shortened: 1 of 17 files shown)


segiddins:
I don't think these need to be available on `RLMObject` -- filtering based on `allObjects` should be sufficient.

I think a swift API could look something like `filter(boundingBox:latitudeProperty:LongitudeProperty)` ?

Realm/RLMLocationTypes.h
```diff
@@ -0,0 +1,27 @@
...
+
+#import <CoreLocation/CoreLocation.h>
+
+/**
+ A bounding box is rectangle that is represented by coordinates for opposing corners.
+ */
+typedef struct {
+    CLLocationCoordinate2D corner1;
+    CLLocationCoordinate2D corner2;
+} RLMBoundingBox;
```
segiddins: Do we need a matching `RLMBoundingBoxMake()` ?

segiddins:
I'm also not entirely convinced those should be exceptions rather than returning errors?

tgoyne:
My main concern about using the CoreLocation coordinate types is that since it's part of the umbrella header it drags in a lot of extra stuff for anyone including Realm, and I think if modules are enabled it'll even result in CL being linked into the user's app. I'm not sure if that's actually a problem in practice, but I'd expect a few complaints about it.

bdash:
@tgoyne, for Objective-C projects, auto-linking doesn't pull in CoreLocation.framework unless you make use of symbols it defines. Merely using types declared in its headers doesn't result in it being pulled in. Things don't work quite so well for Swift. Realm APIs mentioning `CLLocationCoordinate2D` doesn't result in CoreLocation.framework being pulled in directly, but it does result in CoreLocation's Swift overlay, `libswiftCoreLocation.dylib`, being pulled in, which in turn pulls in CoreLocation.framework.

It would be easy to declare our own coordinate type instead. The downside is that those users that choose to use our location-based APIs will have to convert between types when doing so. It's probably worth it to avoid pulling in CoreLocation unnecessarily for all clients using Swift.

bdash:
Rebased onto master with the updated version of core. This is now ready for review.

Realm/RLMObject.h
```diff
@@ -311,6 +312,17 @@ RLM_ASSUME_NONNULL_BEGIN
...
+/**
+ Get objects whose location is within the given bounding box.
+
+ @param box                The bounding box to find objects within.
+ @param latitudeProperty   The property of the object to interpret as the latitude.
+ @param longitudeProperty  The property of the object to interpret as the longitude.
+
+ @return                   An RLMResults of objects of the subclass type in the default Realm whose locations are within the bounding box.
+ */
++ (RLMResults *)objectsWithinBoundingBox:(RLMBoundingBox)box latitudeProperty:(NSString *)latitudeProperty longitudeProperty:(NSString *)longitudeProperty;
```
segiddins: I personally don't think these methods are necessary -- should be sufficient to support querying against `allObjects`

bdash: I added these matching methods on `RLMObject` for sake of discoverability. I don't feel strongly about having them here though.

tgoyne: It'd be weird and inconsistent to not have them on RLMObject.

jpsim:
I'd rather this be a utility to generate an `NSPredicate`, which would cut down on the number of API additions we need to make in both bindings.

bdash:
While this particular query can be represented as an `NSPredicate`, related queries such as the objects-within-distance-of-point cannot be. I think it'd be weird for these two related types of queries to be exposed in different ways.

bdash:
I'm going to close this PR for now since it doesn't make sense to add this without broader support for location types. When we attack that, we can revisit this branch.

PR #2202, opened 2015-07-14
Make the Swift version switchable on master

segiddins:


segiddins:
This is some infrastructure work necessary to allow merging Swift 2 into master alongside 1.2, so please do review it! \c @tgoyne @jpsim @bdash @mrackwitz

bdash:
Can you provide a high-level description of your approach so I can understand what you're trying to do rather than trying to reverse-engineer it from the diff?

segiddins:
Sure! The gist of this approach is that, everywhere where we need different files based upon the swift version, we'll have directories for the different versions, and symlink the version we're currently interested in to the "canonical" location. That way, we don't need to duplicate things like the Xcodeproject, which really doesn't want to be duplicated.

Realm.xcodeproj/project.pbxproj
```diff
@@ -1235,6 +1277,20 @@
...
+			files = (
+			);
+			inputPaths = (
+			);
+			name = "Switch Swift Version";
+			outputPaths = (
+			);
+			runOnlyForDeploymentPostprocessing = 0;
+			shellPath = /bin/sh;
+			shellScript = "if [[ (-L Realm/Tests/Swift) || (! -d Realm/Tests/Swift) ]]; then\n    version=\"$(xcrun swift --version | grep -io --color=never -E 'Apple Swift version ([^\\b ]*)' | cut -f 4 -d ' ')\"\n    rm -f Realm/Tests/Swift\n    (cd Realm/Tests && ln -s \"Swift$version\" Swift)\nfi";
```
bdash: You can extract the Swift version with `sed` (or `perl`) rather than combining `grep` and `cut`:

    sed -ne 's/^Apple Swift version \([^\b ]*\).*/\1/p'
    perl -nle '/Apple Swift version ([^\b ]*)/ && print $1'

To me it's also a little more obvious what they're doing.

RealmSwift
```diff
@@ -0,0 +1 @@
+RealmSwift-swift1.2
```
bdash: Is it necessary to check the symlinks in? The `Tests` one in particular looks like it'll be recreated before it's ever used, but perhaps this one already needs to exist before the Xcode project can be used?

segiddins: Yup, this needs to be here for Xcode to be able to load the config and thus get to ever running the switch script.

build.sh
```diff
@@ -220,6 +220,22 @@ case "$COMMAND" in
...
 export CONFIGURATION
 
+: ${REALM_SWIFT_VERSION:=1.2}
+case "$REALM_SWIFT_VERSION" in
+    "1.2")
+        DEVELOPER_DIR="/Applications/Xcode.app/Contents/Developer/"
+        REALM_SWIFT_VERSION=
+        ;;
+    "2.0")
+        DEVELOPER_DIR="/Applications/Xcode-beta.app/Contents/Developer/"
```
bdash: We talked on Slack about the problems with hard-coding paths to Xcode here. We'll need to revisit this when Xcode 7 GMs.

segiddins: Taken care of

Realm.xcodeproj/project.pbxproj
```diff
@@ -1781,6 +1852,26 @@
 			};
 			name = Release;
 		};
+		C04B4BD81B55C47600FAE58E /* Debug */ = {
+			isa = XCBuildConfiguration;
+			buildSettings = {
+				IPHONEOS_DEPLOYMENT_TARGET = 8.0;
+				PRODUCT_NAME = "$(TARGET_NAME)";
+				SDKROOT = "";
+				SUPPORTED_PLATFORMS = "macosx iphoneos iphonesimulator";
```
bdash: Does this need to include watchos / watchsimulator?

segiddins: Not yet, since there's no watchOS on this branch.

build.sh
```diff
@@ -220,6 +220,30 @@ case "$COMMAND" in
 esac
 export CONFIGURATION
 
+xcode_for_swift() {
```
bdash: Does this do the right thing if multiple Xcode 7 betas are found? It doesn't appear to do anything clever like preferring the most recent version, so it will just pick whichever is first in Spotlight's results. Do we care?

build.sh
```diff
@@ -220,6 +220,30 @@ case "$COMMAND" in
...
+            break;
+        fi
+    done
+}
+
+: ${REALM_SWIFT_VERSION:=1.2}
+: ${DEVELOPER_DIR:="$(xcode_for_swift "$REALM_SWIFT_VERSION")"}
+export DEVELOPER_DIR
+if [[ "$REALM_SWIFT_VERSION" = "1.2" ]]; then
+   REALM_SWIFT_VERSION=
```
bdash: Deos `build.sh` use two space or four space indentation? This change contains a mix.

bdash:
This seems fine to me assuming that CI is happy with it.

build.sh
segiddins: It doesn't yet, and I don't know that my `bash` skills are up to doing that.

build.sh
segiddins: There's already a mix of 2 and 4, I'm happy to make it consistent, just don't know which way that should be.

segiddins:
retest this please

segiddins:
Failures are unrelated.

build.sh
```diff
@@ -260,6 +284,19 @@ case "$COMMAND" in
...
 
     ######################################
+    # Swift versioning
+    ######################################
+    "set-swift-version")
+        version="$2"
+        if [[ -z "$version" ]]; then
+            version="$(xcrun swift --version 2>/dev/null | sed -ne 's/^Apple Swift version \([^\b ]*\).*/\1/p')"
+        fi
+        rm -f RealmSwift && ln -s "RealmSwift-swift$version" RealmSwift && (git update-index --assume-unchanged RealmSwift || true)
```
tgoyne: There's no need for this to be all on one line and the end result is somewhat unreadable.

tgoyne: Also needs a comment that briefly says what `git update-index` is using since I'm guessing it's not a command that most are familiar with.

tgoyne:
The automatic switching does in fact work correctly with the Swift 2.0 stuff in place. It's actually pretty seamless other than that it doesn't switch until you hit build.

PR #2208, opened 2015-07-15
Wrap dispatch_async tests in an autoreleasepool

segiddins:
Should hopefully help with the swift tests failing due to leaked realm instances


tgoyne:
:+1:

Issue #1291, opened 2015-01-02
Dynamic interface support


Kirow:
Does it possible to make two realm files with different data structure in single project? 
i.e. smth like SQLite with attach feature.

For example it will be very useful when I have large catalog of items on server, and i will upload only this catalog file with single table (no need to parse and save to db locally) and use it with my local db which have different structure.

alazier:
Yes this makes sense, but only using a dynamic interface. With the object interface there would be no way to specify two differing schema as your table structure is defined at compile time. With a dynamic interface you can specify schema at runtime.

The good new is there is already internal support for a dynamic interface in Realm but the apis are currently private. We eventually plan to make these or similar apis public (after working through a few issues). The browser currently uses these apis if you want to take a look but at the moment they are liable to change.

PR #2213, opened 2015-07-16
Partially Expose Dynamic API

bigfish24:
Adjusted podspec to include RLMRealm_Dynamic.h as public and moved RLMRealm_Dynamic.h to public headers in Xcode project.

Closes https://github.com/realm/realm-cocoa/issues/1291.


bigfish24:
@jpsim and @segiddins does this look ok in terms of making dynamic api more accessible?

Realm.podspec
```diff
@@ -30,7 +30,8 @@ Pod::Spec.new do |s|
                               'include/Realm/RLMRealm.h',
                               'include/Realm/RLMResults.h',
                               'include/Realm/RLMSchema.h',
-                              'include/Realm/Realm.h'
+                              'include/Realm/Realm.h',
+			      'include/Realm/RLMRealm_Dynamic.h'
```
segiddins: alignment

segiddins: also, do we maybe only want to make this public in a subspec?

mrackwitz:
The discussed change to the modulemap should also go into this PR.

Realm.podspec
bigfish24: I don't think so because would rather not force dev to adjust their install of Realm to use a component...

mrackwitz: It would be automatically available as a transitive dependency.

bigfish24: ah ok, but I guess why would you ever install the subspec only? I guess that is the point of breaking it out right?

mrackwitz: I think, we should be fine when our custom module map explicitly defines already a submodule for it. So no need for a subspec imo.

bigfish24: @mrackwitz ok

bigfish24: @segiddins I think the alignment is just github web view... whole file doesn't show that

segiddins: @bigfish24 you used tabs, please switch to spaces

bigfish24: @segiddins done

segiddins:
@bigfish24 you'll now need to add some `import Realm.Dynamic` statements

bigfish24:
@segiddins where?

segiddins:
wherever it's needed to get the build to pass?

bigfish24:
@segiddins builds fine on my machine?

segiddins:
Needs a rebase and to also change Swift 2

bigfish24:
@jpsim you said you had issues with opening this up. Would you have any issues with just releasing the dynamic retrieval methods? I would prefer to use this with components than objc reflection.

jpsim:
My main concerns are:

1. We need to better document the full dynamic RLMRealm constructor. For example, what parameter values are valid (e.g. for `customSchema`).
2. This API is Objective-C only. Doing this right would entail having an equivalent Swift API.
3. I'm not sure your changes to `build-docs.sh` are entirely correct (it seems like `RLMUtil.h` will now be documented). I still don't think we should generate HTML docs for the dynamic API.

scripts/build-docs.sh
```diff
@@ -21,12 +21,10 @@ appledoc \
     --no-warn-invalid-crossref \
     --no-warn-undocumented-object \
     --no-warn-undocumented-member \
-    --ignore "Realm/RLMRealm_Dynamic.h" \
-    --ignore "Realm/RLMArrayAccessor.h" \
-    --ignore "Realm/RLMArrayAccessor.mm" \
-    --ignore "Realm/RLMQueryUtil.h" \
-    --ignore "Realm/RLMUtil.h" \
-    --ignore "Realm/RLMRealm_Dynamic.h" \
+    --ignore ".m" \
```
jpsim: can't we do `--ignore ".m*"`?

scripts/build-docs.sh
```diff
@@ -21,12 +21,10 @@ appledoc \
...
+    --ignore ".mm" \
+    --ignore "Realm/RLMObjectStore.h" \
+    --ignore "Realm/RLMCollection.h" \
```
jpsim: Why are we now ignoring `RLMCollection.h`? This is part of the public API.

scripts/build-docs.sh
```diff
@@ -21,12 +21,10 @@ appledoc \
     --no-warn-invalid-crossref \
     --no-warn-undocumented-object \
     --no-warn-undocumented-member \
-    --ignore "Realm/RLMRealm_Dynamic.h" \
-    --ignore "Realm/RLMArrayAccessor.h" \
-    --ignore "Realm/RLMArrayAccessor.mm" \
-    --ignore "Realm/RLMQueryUtil.h" \
-    --ignore "Realm/RLMUtil.h" \
```
jpsim: This should be ignored.

scripts/build-docs.sh
```diff
@@ -21,12 +21,10 @@ appledoc \
     --no-warn-invalid-crossref \
     --no-warn-undocumented-object \
     --no-warn-undocumented-member \
-    --ignore "Realm/RLMRealm_Dynamic.h" \
-    --ignore "Realm/RLMArrayAccessor.h" \
-    --ignore "Realm/RLMArrayAccessor.mm" \
-    --ignore "Realm/RLMQueryUtil.h" \
```
jpsim: this should be ignored.

scripts/build-docs.sh
```diff
@@ -21,12 +21,10 @@ appledoc \
     --no-warn-invalid-crossref \
     --no-warn-undocumented-object \
     --no-warn-undocumented-member \
-    --ignore "Realm/RLMRealm_Dynamic.h" \
-    --ignore "Realm/RLMArrayAccessor.h" \
```
jpsim: this should be ignored.

scripts/build-docs.sh
bigfish24: sure I can simplify that

jpsim: I'm not 100% sure appledoc supports wildcards, so please confirm it works first.

scripts/build-docs.sh
bigfish24: I wasn't sure on that. The current API docs have it public but only setValue and valueForKey is actually documented by appledoc. I added the proper documentation for appledoc to generate, but didn't know if we wanted to actually include it. Given that it was incomplete, I assumed it might have been an oversight to have it public.

jpsim: No, we should keep that public IMO.

bigfish24: ok, I will add that in

scripts/build-docs.sh
bigfish24: also `RLMUtil.h` isn't even a header anymore? I can go in and add ignore statements to any files that shouldn't be in there. Reason I removed is that none of the files that were being ignored actually have syntax that will generate an appledoc.

bigfish24: Also, I am not against hiding the dynamic methods from the generated appledoc, but the thought process was since they are now more public they should be in there but with sufficient warnings (came out of conversant with Tim).

jpsim: no need to ignore non-existent files then.

jpsim: > since they are now more public they should be in there but with sufficient warnings (came out of conversant with Tim).

Sure, but the methods get added inline with the public ones, in alphabetical order, so it's not exactly well delineated.

bigfish24: Ok, well do you want me to add the files that are in repo but shouldn't generate docs, even though they don't generate docs because the comment syntax isn't appledoc compatible?

jpsim: I think that's safer, yes. We'll inevitably want to document a private API at some point, so I'd rather us be explicit about what's ignored. Just mirror what's a private header in the podspec.

bigfish24: ok will do

bigfish24:
ok, @jpsim I fixed the build script per your requests.

I also removed the full constructor from `RLMRealm_Dynamic.h` (it still resides in `RLMRealm_Private.h`), and then also moved the dynamic `RLMObjectSchema` method and `RLMProperty` method to their private headers.

The result is that the only methods that are being exposed via `RLMRealm_Dynamic.h` are the dynamic retrieval and dynamic create method.

Finally, on the swift side of things, I added three methods for Realm:
```
dynamicObjects(className: String) -> Results<DynamicObject>
```
```
dynamicObjectForPrimaryKey(className: String, key: AnyObject) -> DynamicObject?
```
```
dynamicCreate(className: String, value: AnyObject = [:], update: Bool = false) -> DynamicObject
```
which means the Swift API mirrors the Objc API.

I don't think it is possible though to move these methods into a separate module since Clang modules only support C/Objc...

bigfish24:
@jpsim and @segiddins per our conversation, I added a header: `RLMObjectBase_Dynamic.h` that exposes the `RLMRealm` and `RLMObjectSchema` at the object base level for use with components. This header is included in the same way `RLMRealm_Dynamic.h` is. Since we weren't creating docs for `RLMObjectBase.h` before, I made sure to ignore both this and the dynamic header in the docs script. 

If components act on `RLMObjectBase`, assuming it comes from Swift or Objc, they should be able to perform read/writes through key value coding. The only thing missing would be `linkingObjectFromClass:` but not sure there is a use yet for that, but we could push that down to `RLMObjectBase` too.

segiddins:
@bigfish24 `RLMObjectBaseLinkingObjectsOfClass` already handles backlinks

bigfish24:
ok @segiddins I removed the methods and simple exposed the functions that we are already utilizing, but only those that are relevant to current components.

Realm.podspec
```diff
@@ -30,7 +30,8 @@ Pod::Spec.new do |s|
                               'include/Realm/RLMRealm.h',
                               'include/Realm/RLMResults.h',
                               'include/Realm/RLMSchema.h',
-                              'include/Realm/Realm.h'
+                              'include/Realm/Realm.h',
+                              'include/Realm/RLMRealm_Dynamic.h',
```
segiddins: missing `RLMObjectBase_Dynamic.h`

bigfish24: ah, must have missed that with my clean up. Will fix.

bigfish24:
retest this please

bigfish24:
retest this please

Realm/RLMRealm_Dynamic.h
```diff
@@ -78,45 +89,44 @@
...
-
- @return    An initialized instance of RLMProperty.
- */
-- (instancetype)initWithName:(NSString *)name
-                        type:(RLMPropertyType)type
-             objectClassName:(NSString *)objectClassName
-                     indexed:(BOOL)indexed
-                    optional:(BOOL)optional;
-@end
+@end
```
jpsim: missing newline

jpsim:
From today's meeting: please ignore the dynamic API from jazzy and appledoc, otherwise :+1:.

Realm/RLMObject.h
```diff
@@ -203,12 +203,12 @@ RLM_ASSUME_NONNULL_BEGIN
 /**
  The Realm in which this object is persisted. Returns nil for standalone objects.
  */
-@property (nonatomic, readonly, nullable) RLMRealm *realm;
+@property (nonatomic, readonly, nullable, getter = realm) RLMRealm *realm;
```
jpsim: why do you have an explicit getter here?

Realm/RLMObject.h
```diff
@@ -203,12 +203,12 @@ RLM_ASSUME_NONNULL_BEGIN
...
 
 /**
  The ObjectSchema which lists the persisted properties for this object.
  */
-@property (nonatomic, readonly) RLMObjectSchema *objectSchema;
+@property (nonatomic, readonly, getter = objectSchema) RLMObjectSchema *objectSchema;
```
jpsim: why do you have an explicit getter here?

RealmSwift-swift2.0/Realm.swift
```diff
@@ -286,6 +286,37 @@ public final class Realm {
...
+    
+    :param: className   The class name of the object to create.
+    :param: value       The value used to populate the object. This can be any key/value coding compliant
+    object, or a JSON dictionary such as those returned from the methods in `NSJSONSerialization`,
+    or an `Array` with one object for each persisted property. An exception will be
+    thrown if any required properties are not present and no default is set.
+    
+    When passing in an `Array`, all properties must be present,
+    valid and in the same order as the properties defined in the model.
+    :param: update      If true will try to update existing objects with the same primary key.
```
jpsim: Documentation comments in these Swift 2 files need to use Swift 2's documentation format: http://ericasadun.com/2015/06/14/swift-header-documentation-in-xcode-7/

Realm/RLMObject.h
bigfish24: That's an oversight on my part. I saw the explicit getters for `invalidated` and `deletedFromRealm` and for some reason thought at the time I needed to add it (but the ivar isn't named differently...). What is missing is that I wanted to add `realm` and `objectSchema` properties to `RLMObjectBase` since that was the point of moving the getters to there. I will correct.

Realm/RLMProperty_Private.h
```diff
@@ -60,3 +60,26 @@ FOUNDATION_EXTERN BOOL RLMPropertyTypeIsNullable(RLMPropertyType propertyType);
 
 @end
 
+@interface RLMProperty (Dynamic)
+/**
+ This method is useful only in specialized circumstances, for example, in conjunction with
+ [RLMObjectSchema initWithClassName:objectClass:properties:]. If you are simply building an
```
segiddins: `+[RLMObjectSchema`

RealmSwift-swift1.2/Realm.swift
```diff
@@ -291,6 +291,39 @@ public final class Realm {
...
+    
+    When passing in an `Array`, all properties must be present,
+    valid and in the same order as the properties defined in the model.
+    :param: update      If true will try to update existing objects with the same primary key.
+    
+    :returns: The created object.
+    
+    :nodoc:
+    */
+    public func dynamicCreate(className: String, value: AnyObject = [:], update: Bool = false) -> DynamicObject {
```
segiddins: is it necessary to call this `dynamicCreate` just `create` should be fine, since it has a different signature

RealmSwift-swift1.2/Realm.swift
```diff
@@ -358,6 +391,25 @@ public final class Realm {
...
+    
+    :warning: This method is useful only in specialized circumstances.
+    
+    :param: className  The class name of the objects to be returned.
+    
+    :returns: All objects for the given class name as dynamic objects
+    
+    :nodoc:
+    */
+    public func dynamicObjects(className: String) -> Results<DynamicObject> {
```
segiddins: same comment as above -- could just be `objects`

RealmSwift-swift1.2/Realm.swift
```diff
@@ -376,7 +428,32 @@ public final class Realm {
...
+    :warning: This method is useful only in specialized circumstances.
+    
+    :param: className  The class name of the object to be returned.
+    :param: key        The primary key of the desired object.
+    
+    :returns: An object of type `DynamicObject` or `nil` if an object with the given primary key does not exist.
+    
+    :nodoc:
+    */
+    public func dynamicObjectForPrimaryKey(className: String, key: AnyObject) -> DynamicObject? {
```
segiddins: same comment as above -- could just be `objectForPrimaryKey`

bigfish24: Would like to hear everyone's input on this, not opposed to change but this does make difference more clear.

RealmSwift-swift1.2/Realm.swift
jpsim: I'd rather keep the "dynamic" prefixes to these methods to discourage accidental usage of the dynamic API.

RealmSwift-swift1.2/Realm.swift
mrackwitz: I think what @segiddins proposes here is more the idiomatic Swift way.

RealmSwift-swift1.2/Realm.swift
bigfish24: I agree I would rather keep it because it then makes the method distinct in the API docs.

bigfish24:
@jpsim @segiddins can we merge this into master? I rebased to include the changes from KVO.

Realm/RLMRealm_Private.h
```diff
@@ -47,6 +47,34 @@ FOUNDATION_EXTERN void RLMRealmAddPathSettingsToConfiguration(RLMRealmConfigurat
...
+ possible errors, pass in NULL.
+ 
+ @return An `RLMRealm` instance.
+ 
+ @see RLMRealm defaultRealm
+ @see RLMRealm realmWithPath:
+ @see RLMRealm realmWithPath:readOnly:error:
+ @see RLMRealm realmWithPath:encryptionKey:readOnly:error:
+ */
 + (instancetype)realmWithPath:(NSString *)path
```
mrackwitz: I don't think we should publicly expose _this_ method. We should fully commit on the new approach based on `RLMConfiguration`.

bigfish24: it isn't public... it is in the private module.

segiddins:
@bigfish24 I think this is mostly fine, however there are merge conflicts

bigfish24:
@segiddins merge conflicts fixed

segiddins:
Only thing left to test is that dynamic objects are actually returned from the functions that are supposed to return them. (I disagree with the `dynamic` prefix from a pure swift api perspective, but I understand wanting to make them stand out...)

bigfish24:
@segiddins take a look at the latest commit. I added two tests for `dynamicCreate` and `dynamicObjectForPrimaryKey` where `unsafeBitCast` is used. The tests verify subscripting is working on the returned object.

RealmSwift-swift1.2/Realm.swift
tgoyne: Definitely needs the prefix. If it has the same name I can guarantee that people will accidentally autocomplete their way into using the dynamic versions. It'd be okay if it required importing `RealmSwift.Dynamic`, but I suspect getting that to work would be another sizable project.

tgoyne:
This looks fine for the specific goal of making the dynamic API usable from more places without making it a fully public and officially supported thing. More changes will be needed to make it public-for-real, but those don't need to happen now.

:+1:

segiddins:
:+1:

jpsim:
:+1:

PR #2215, opened 2015-07-16
Support building framework for watchOS architecture

tgoyne:
Given a core build with bitcode this can now produce a (seemingly) fully working release package, but there's not watchOS sample so there may still be broken things...

See https://github.com/realm/realm-cocoa/pull/2139 for previous conversation.


tgoyne:
Now uses a core build with bitcode so it should actually work.

segiddins:
Only failure is unrelated

tgoyne:
test-package-release now runs to completion

segiddins:
:+1:

PR #2216, opened 2015-07-16
Disable syncing to disk for everything but running on device

tgoyne:
Should save about three minutes of worker time on the CI PR runs.

### DIFF
--- a/Realm/Tests/RLMTestCase.m
+++ b/Realm/Tests/RLMTestCase.m
@@ -87,12 +87,12 @@ static BOOL encryptTests() {
 
 @implementation RLMTestCase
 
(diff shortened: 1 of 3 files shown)


segiddins:
:+1:

segiddins:
Needs a rebase and to also change Swift 2

segiddins:
This should be good to go now, \c @jpsim

tgoyne:
Only test failure was due to no devices plugged in to the CI machine.

jpsim:
In practice, this change surprisingly seemed to be slowing down some performance tests on device. It'd be good to re-enable device tests on CI so we can look further into that.

PR #2252, opened 2015-07-20
Move common Realm functionality to new class realm::Realm

alazier:
Closes https://github.com/realm/realm-cocoa/issues/2479.


alazier:
@segiddins @simonask

Realm/ObjectStore/object_store_exceptions.hpp
```diff
@@ -24,42 +24,78 @@
...
+            ObjectSchemaNewPrimaryKey,              // ObjectType, PrimaryKey
+            ObjectSchemaChangedOptionalProperty,
+            ObjectSchemaNewOptionalProperty,
+            ObjectStoreValidationFailure,           // ObjectType, vector<ObjectStoreException>
         };
-        typedef std::map<std::string, std::string> Dict;
 
-        ObjectStoreException(Kind kind, Dict dict = Dict());
+        typedef const std::string InfoKey;
+        typedef std::map<InfoKey, std::string> Info;
```
simonask: Consider using `using` instead of `typedef` (easier to grep for).

```c++
using InfoKey = const std::string;
using Info = std::map<InfoKey, std::string>;
```

Also note that the "key" template parameter passed to `std::map` is turned `const` anyway, so unless you explicitly want to forbid modifications to an `InfoKey` *before* it gets to the map, it could be just `std::string` without the `const`.

Realm/RLMRealm.mm
```diff
@@ -829,16 +741,25 @@ + (uint64_t)schemaVersionAtPath:(NSString *)realmPath error:(NSError **)error {
...
-        if (error) {
-            RLMSetErrorOrThrow(error, outError);
-            return RLMNotVersioned;
-        }
+    if (realm) {
+        return realm->_realm->config().schema_version;
     }
 
-    return realm::ObjectStore::get_schema_version(realm.group);
+    NSError *error;
```
segiddins: This seems to be unused?

Realm/RLMRealm.mm
```diff
@@ -829,16 +741,25 @@ + (uint64_t)schemaVersionAtPath:(NSString *)realmPath error:(NSError **)error {
...
+        config.path = realmPath.UTF8String;
+        config.encryption_key = key ? static_cast<const char *>(key.bytes) : StringData();
+        uint64_t version = Realm::get_shared_realm(config)->config().schema_version;
+        if (version == realm::ObjectStore::NotVersioned) {
+            RLMSetErrorOrThrow([NSError errorWithDomain:RLMErrorDomain code:RLMErrorFail userInfo:@{NSLocalizedDescriptionKey:@"Cannot open an uninitialized realm in read-only mode"}], outError);
+        }
+        return version;
+    }
+    catch (std::exception *exp) {
+        RLMSetErrorOrThrow(error, outError);
```
segiddins: Shouldn't this be creating the error based upon the caught exception?

Realm/Tests/RLMTestCase.m
```diff
@@ -151,7 +157,7 @@ - (void)waitForNotification:(NSString *)expectedNote realm:(RLMRealm *)realm blo
     XCTestExpectation *notificationFired = [self expectationWithDescription:@"notification fired"];
     RLMNotificationToken *token = [realm addNotificationBlock:^(__unused NSString *note, RLMRealm *realm) {
```
segiddins: `note` shouldn't be marked as `__unused`

Realm/RLMRealm.mm
```diff
@@ -381,8 +378,69 @@ + (instancetype)realmWithPath:(NSString *)path
...
+    config.read_only = readonly;
+    config.in_memory = inMemory;
+    config.encryption_key = key ? static_cast<const char *>(key.bytes) : StringData();
+
+    __weak RLMRealm *weakRealm = realm;
+    config.migration_function = [=](__unused Group *group, ObjectStore::Schema &schema) {
+        RLMRealm *strongRealm = weakRealm;
+        if (!strongRealm) {
+            return;
+        }
```
simonask: I'm unsure if this is guaranteed to work. Underneath, the C++ lambda gets turned into a struct with a `__weak RLMRealm*` member, but I'm unsure if the `__weak` qualifier is actually applied to that member, or if it even can be applied to it. Does `__weak` generally work for struct members, or only for stack variables?

Realm/ObjectStore/object_store_exceptions.hpp
```diff
@@ -24,42 +24,78 @@
 #include <string>
 
 namespace realm {
+    class Property;
 
     class ObjectStoreException : public std::exception {
     public:
         enum class Kind {
```
simonask: I'm wondering what the rationale is behind using this approach rather than defining separate exception error types deriving from a common `ObjectStoreException`, which could then also just define extra information as members instead of going through a (potentially expensive) `std::map`. Especially considering that these exceptions will have to be translated into Java exceptions at some point, where it just becomes overhead. :)

Realm/ObjectStore/object_store_exceptions.cpp
```diff
@@ -17,29 +17,112 @@
...
+        errors_string += string("\n- ") + error.what();
     }
+    return errors_string;
 }
+
+std::string ObjectStoreException::populate_format_string(const std::string & format_string) const {
+    string out_string, current(format_string);
+    smatch sm;
+    regex re("\\{(\\w+)\\}");
+    while(regex_search(current, sm, re)) {
```
simonask: This does a lot of copying. It can be made copy-free by using the iterator-based version of `std::regex_search` in conjunction with `StringData` to represent the remaining substring of the `format_string`. Probably not a performance priority, but maybe add a comment about it.

Realm/ObjectStore/object_store_exceptions.cpp
```diff
@@ -17,29 +17,112 @@
...
-        m_what += "\n- " + error;
+        errors_string += string("\n- ") + error.what();
     }
+    return errors_string;
 }
+
+std::string ObjectStoreException::populate_format_string(const std::string & format_string) const {
+    string out_string, current(format_string);
+    smatch sm;
+    regex re("\\{(\\w+)\\}");
```
simonask: Please verify that `std::regex` is working with this expression on all platforms. I've had trouble in the past with GCC, but I'm especially worried about MSVC. :)

Realm/ObjectStore/object_store_exceptions.cpp
```diff
@@ -17,29 +17,112 @@
...
+    while(regex_search(current, sm, re)) {
+        out_string += sm.prefix();
+        const string &key = sm[1];
+        if (key == "ValidationErrors") {
+            out_string += validation_errors_string();
+        }
+        else {
+            out_string += m_info.at(key);
```
simonask: This throws an exception if `key` does not exist in `m_info`. I think the "correct" behavior in this case would be to actually leave the interpolation placeholder as is, so that specifying a key that doesn't exist "does nothing".

I.e., the format string "Hello, {KeyThatDoesNotExist}!" should simply yield the output "Hello, {KeyThatDoesNotExist}!", which is testable, but doesn't cause sudden crashes when we make a mistake.

Alternatively, all of this can be done away with if we switch to using actual specialized exception types that define members for all of this, in which case we would get a compile-time error instead. :)

Realm/ObjectStore/shared_realm.cpp
```diff
@@ -0,0 +1,366 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+////////////////////////////////////////////////////////////////////////////
+
+#include "shared_realm.hpp"
+
+#include <realm/commit_log.hpp>
+
+using namespace std;
```
simonask: Just as a heads up, we've mostly decided against doing this in Core, even in cpp files, because it makes it clearer where things come from to just always have the `std::` prefix.

Realm/ObjectStore/object_store_exceptions.hpp
alazier: This was mainly to avoid a bunch of boilerplate code and make it easy to add new exceptions. I agree not using map would improve efficiency but I don't think performance is a concern in the cases we would be constructing any of these exceptions.

Realm/RLMRealm.mm
alazier: I'm pretty sure this should work.

Realm/ObjectStore/object_store_exceptions.hpp
simonask: What I'm wondering, though, is how big the difference actually is in amount of boilerplate code, considering the type safety and locality of the alternative. But let's leave it for now and take it up when we start solidifying the API. :)

alazier:
retest this please

alazier:
Performance is now on par with what is on master. Would be good to start reviewing @tgoyne @bdash @segiddins

segiddins:
Looks like this needs a rebase?

alazier:
Just rebased. Unfortunately the kvo tests explode. If you guys could take a look would definitely appreciate it.

tgoyne:
Porting the more recent changes to RLMRealm to this will take some time.

Some assorted comments:

`make_shared()` could be used for `Realm` without making them externally constructible by doing the following:

```
private:
    struct PrivateTag {};
public:
    Realm(Config& config, PrivateTag);

...

    realm = make_shared<Realm>(config, PrivateTag{});
```

`ExternalNotificationFunction` doesn't need the `unique_ptr`; `std::function` can store `nullptr` directly.

`cacheTableAccessors` defeats the point of the table accessor lazy-fetching: avoiding looking up tables that are never referenced when doing stuff on background queues. I haven't profiled to see if this is actually why, but `testRealmCreationCached` is ~15% slower than master.

`enable_shared_from_this` is kinda gross. It looks like it should be possible to avoid it if `RLMRealm` also supported non-owning references to the `Realm`, since the `RLMRealm`s used in migrations don't need to stay alive as long as a user holds on a reference to them (and in fact I suspect that a user storing a reference to migration objects outside of the migration block would break things horribly anyway).

tgoyne:
There's still a lot of stuff related to this to do (e.g. getting rid of the separate RLMRealm cache and just using the objectstore one), but I've now covered everything I consider a blocker for merging it, so reviews needed.

tgoyne:
OS X failures unrelated; ios-device failure is a perf regression in RealmCreationCached.

Realm/ObjectStore/object_store.cpp
```diff
@@ -161,46 +136,97 @@ TableRef ObjectStore::table_for_object_type_create_if_needed(Group *group, const
...
+
+        auto more_errors = verify_object_schema(*matching_schema, object_schema, target_schema);
+        errors.insert(errors.end(), more_errors.begin(), more_errors.end());
+    }
+    if (errors.size()) {
+        throw SchemaValidationException(errors);
+    }
+}
+
+std::vector<ObjectSchemaValidationException> ObjectStore::verify_object_schema(ObjectSchema const& table_schema, ObjectSchema &target_schema, Schema &schema) {
```
bdash: It looks like `schema` can be a const reference.

Realm/ObjectStore/object_store.cpp
```diff
@@ -272,14 +283,29 @@ bool ObjectStore::create_tables(Group *group, ObjectStore::Schema &target_schema
...
+            }
+        }
 
         // add missing columns
         for (auto& target_prop : target_props) {
             auto current_prop = current_schema.property_for_name(target_prop.name);
 
             // add any new properties (new name or different type)
-            if (!current_prop || property_has_changed(*current_prop, target_prop)) {
+            if (!current_prop || current_prop->table_column == npos) {
```
bdash: The comment above appears looks like it needs updated for this change.

Realm/ObjectStore/object_store.cpp
```diff
@@ -346,35 +348,51 @@ bool ObjectStore::create_tables(Group *group, ObjectStore::Schema &target_schema
...
     if (old_version > version && old_version != NotVersioned) {
-        throw ObjectStoreException(ObjectStoreException::Kind::RealmVersionGreaterThanSchemaVersion,
-                                   {{"old_version", to_string(old_version)}, {"new_version", to_string(version)}});
+        throw InvalidSchemaVersionException(old_version, version);
     }
-    return old_version != version;
+    return old_version == version;
 }
 
+bool ObjectStore::needs_update(Schema const& old_schema, Schema& schema) {
```
bdash: `schema` can be const here.

Realm/ObjectStore/object_store.cpp
```diff
@@ -479,3 +476,106 @@ void ObjectStore::delete_data_for_object(Group *group, const StringData &object_
...
+{
+    m_what = "Primary key property '" + property.name + "' has duplicate values after migration.";
+};
+
+
+SchemaValidationException::SchemaValidationException(std::vector<ObjectSchemaValidationException> errors) :
+    m_validation_errors(errors)
+{
+    m_what ="Migration is required due to the following errors: ";
+    for (auto error : errors) {
```
bdash: `const auto&` to avoid unnecessary copying.

Realm/ObjectStore/object_store.cpp
```diff
@@ -479,3 +476,106 @@ void ObjectStore::delete_data_for_object(Group *group, const StringData &object_
...
 
+
+
+InvalidSchemaVersionException::InvalidSchemaVersionException(uint64_t old_version, uint64_t new_version) :
+    m_old_version(old_version), m_new_version(new_version)
+{
+    m_what = "Provided schema version " + std::to_string(old_version) + " is less than last set version " + std::to_string(new_version) + ".";
+}
+
+DuplicatePrimaryKeyValueException::DuplicatePrimaryKeyValueException(std::string object_type, Property const& property) :
```
bdash: The complex value types (`std::string`, `std::vector`) in these constructor arguments should probably all be taken as const references.

Realm/ObjectStore/object_store.hpp
```diff
@@ -109,6 +112,122 @@ namespace realm {
...
+      private:
+        std::string m_object_type;
+        Property m_property;
+    };
+
+    // Schema validation exceptions
+    class SchemaValidationException : public ObjectStoreException {
+      public:
+        SchemaValidationException(std::vector<ObjectSchemaValidationException> errors);
+        std::vector<ObjectSchemaValidationException> &validation_errors() { return m_validation_errors; }
```
bdash: Returning a mutable reference here is a bit surprising. Is there a need for it?

Realm/ObjectStore/object_store.hpp
```diff
@@ -109,6 +112,122 @@ namespace realm {
...
+        std::string m_what;
+    };
+
+    // Migration exceptions
+    class MigrationException : public ObjectStoreException {};
+
+    class InvalidSchemaVersionException : public MigrationException {
+      public:
+        InvalidSchemaVersionException(uint64_t old_version, uint64_t new_version);
+        uint64_t old_version() { return m_old_version; }
```
bdash: The getters on all of these exception classes should probably be marked const.

Realm/ObjectStore/shared_realm.cpp
```diff
@@ -0,0 +1,430 @@
...
+{
+    if (c.schema) {
+        schema = std::make_unique<Schema>(*c.schema);
+    }
+}
+
+Realm::Config& Realm::Config::operator=(realm::Realm::Config const& c)
+{
+    if (&c != this) {
+        *this = Config(c);
```
bdash: This confused me at first. It's delegating copy assignment to the copy constructor and move assignment. Interesting technique!

Realm/ObjectStore/shared_realm.cpp
```diff
@@ -0,0 +1,430 @@
...
+        throw InvalidTransactionException("Can't compact a Realm within a write transaction");
+    }
+
+    for (auto &object_schema : *m_config.schema) {
+        ObjectStore::table_for_object_type(read_group(), object_schema.name)->optimize();
+    }
+
+    m_shared_group->end_read();
+    success = m_shared_group->compact();
+    m_shared_group->begin_read();
```
bdash: There seems to be an implicit assumption here that `m_group` is non-null when this method is called. In the case where it's null, I think the following sequence of events will happen:
  1. `Realm::compact` begins a read transaction.
  2. The next call to `Realm::read_group` will notice that `m_group` is null and call through to `SharedGroup::begin_read()`.
  3. `SharedGroup::begin_read()` will throw since one of its preconditions is that it not be called when in a read transaction.

If we don't think it's possible to get here with a null `m_group` so documenting the assumption with an assertion is probably sufficient.

Realm/ObjectStore/shared_realm.cpp
```diff
@@ -0,0 +1,430 @@
...
+        read_group();
+    }
+
+    return true;
+}
+
+uint64_t Realm::get_schema_version(const realm::Realm::Config &config)
+{
+    auto existing_realm = s_global_cache.get_any_realm(config.path);
+    if (existing_realm) {
```
bdash: `if (auto existing_realm = …) {`

Realm/ObjectStore/shared_realm.hpp
```diff
@@ -0,0 +1,189 @@
...
+        bool update_schema(std::unique_ptr<Schema> schema, uint64_t version);
+
+        static uint64_t get_schema_version(Config const& config);
+
+        const Config &config() const { return m_config; }
+
+        void begin_transaction();
+        void commit_transaction();
+        void cancel_transaction();
+        bool is_in_transaction() { return m_in_transaction; }
```
bdash: `const`

Realm/ObjectStore/shared_realm.hpp
```diff
@@ -0,0 +1,189 @@
...
+
+        bool refresh();
+        void set_auto_refresh(bool auto_refresh) { m_auto_refresh = auto_refresh; }
+        bool auto_refresh() { return m_auto_refresh; }
```
bdash: `const`

Realm/RLMRealm.mm
```diff
@@ -360,123 +313,186 @@ static id RLMAutorelease(id value) {
...
++ (SharedRealm)openSharedRealm:(Realm::Config const&)config error:(NSError **)outError {
+    try {
+        return Realm::get_shared_realm(config);
+    }
+    catch (RealmFileException &ex) {
+        switch (ex.kind()) {
+            case RealmFileException::Kind::PermissionDenied: {
+                NSString *mode = config.read_only ? @"read" : @"read-write";
+                NSString *additionalMessage = [NSString stringWithFormat:@"Unable to open a realm at path '%@'. Please use a path where your app has %@ permissions.", @(config.path.c_str()), mode];
+                NSString *newMessage = [NSString stringWithFormat:@"%s\n%@", ex.what(), additionalMessage];
```
bdash: `SharedRealm` appears to construct this message in the same form. Does this result in a duplicated message?

Realm/RLMSchema.mm
```diff
@@ -236,4 +243,13 @@ - (NSString *)description {
     return [NSString stringWithFormat:@"Schema {\n%@}", objectSchemaString];
 }
 
+- (std::unique_ptr<Schema>)objectStoreCopy {
+    std::vector<realm::ObjectSchema> schema;
+    schema.reserve(_objectSchema.count);
+    for (RLMObjectSchema *objectSchema in _objectSchema) {
+        schema.push_back(objectSchema.objectStoreCopy);
+    }
+    return std::make_unique<realm::Schema>(std::move(schema));
```
bdash: Namespace qualification doesn't appear to be necessary within this function.

Realm/ObjectStore/object_store.cpp
```diff
@@ -161,46 +136,97 @@ TableRef ObjectStore::table_for_object_type_create_if_needed(Group *group, const
...
-        if (current_prop.is_primary != target_prop->is_primary) {
-            if (current_prop.is_primary) {
-                validation_errors.push_back("Property '" + current_prop.name + "' is no longer a primary key.");
-            }
-            else {
-                validation_errors.push_back("Property '" + current_prop.name + "' has been made a primary key.");
+
+        // check nullablity
+#if REALM_NULL_STRINGS == 1
+        if (current_prop.type == PropertyTypeArray && current_prop.is_nullable) {
```
segiddins: also mixed

Realm/ObjectStore/object_store.cpp
```diff
@@ -161,46 +136,97 @@ TableRef ObjectStore::table_for_object_type_create_if_needed(Group *group, const
...
+            exceptions.emplace_back(InvalidNullabilityException(table_schema.name, current_prop));
+        }
+#else
+        if (current_prop.type == PropertyTypeObject) {
```
segiddins: this needs to be outside the `REALM_NULL_STRINGS` check

Realm/ObjectStore/object_store.cpp
```diff
@@ -289,40 +315,16 @@ bool ObjectStore::create_tables(Group *group, ObjectStore::Schema &target_schema
...
-#else
-                        target_prop.table_column = table->add_column(DataType(target_prop.type), target_prop.name);
-#endif
+                        target_prop.table_column = table->add_column(DataType(target_prop.type),
+                                                                     target_prop.name,
+                                                                     target_prop.is_nullable);
                         break;
                 }
 
-                if (current_prop && property_can_be_migrated_to_nullable(*current_prop, target_prop)) {
```
segiddins: is this being removed intentionally? I thought we wanted the automatic migration?

Realm/ObjectStore/object_store.cpp
segiddins: this looks like it doesn't handle the set of properties changing, which will be needed if we want a `delete_if_migration_required` option

Realm/ObjectStore/object_store.cpp
```diff
@@ -479,3 +476,106 @@ void ObjectStore::delete_data_for_object(Group *group, const StringData &object_
...
+    m_what = "Property '" + property.name + "' is missing from latest object model.";
+}
+
+InvalidNullabilityException::InvalidNullabilityException(std::string object_type, Property const& property) :
+    ObjectSchemaPropertyException(object_type, property)
+{
+#if REALM_NULL_STRINGS == 1
+    m_what = "'Array' property '" + property.name + "' cannot be nullable";
+#else
+    if (property.type == PropertyTypeObject) {
```
segiddins: needs to be outside the conditional

Realm/ObjectStore/object_store.cpp
```diff
@@ -479,3 +476,106 @@ void ObjectStore::delete_data_for_object(Group *group, const StringData &object_
...
+    ObjectSchemaPropertyException(object_type, property)
+{
+    m_what = "Property '" + property.name + "' is missing from latest object model.";
+}
+
+InvalidNullabilityException::InvalidNullabilityException(std::string object_type, Property const& property) :
+    ObjectSchemaPropertyException(object_type, property)
+{
+#if REALM_NULL_STRINGS == 1
+    m_what = "'Array' property '" + property.name + "' cannot be nullable";
```
segiddins: mixed also cannot be nullable

Realm/ObjectStore/shared_realm.cpp
```diff
@@ -0,0 +1,440 @@
...
+            if (realm->config().in_memory != config.in_memory) {
+                throw MismatchedConfigException("Realm at path already opened with different inMemory settings.");
+            }
+            if (realm->config().encryption_key != config.encryption_key) {
+                throw MismatchedConfigException("Realm at path already opened with a different encryption key.");
+            }
+            if (realm->config().schema_version != config.schema_version && config.schema_version != ObjectStore::NotVersioned) {
+                throw MismatchedConfigException("Realm at path already opened with different schema version.");
+            }
+            // FIXME - enable schma comparison
```
segiddins: does this need to be addressed before a merge?

Realm/ObjectStore/transact_log_handler.cpp
```diff
@@ -0,0 +1,347 @@
...
+    // Doesn't change any data
+    bool optimize_table() { return true; }
+
+    // Used for subtables, which we currently don't support
+    bool select_descriptor(int, const size_t*) { return false; }
+};
+} // anonymous namespace
+
+namespace realm {
+namespace transaction {
```
segiddins: `write_transaction`?

Realm/RLMObjectStore.mm
```diff
@@ -517,10 +402,8 @@ static void RLMValidateValueForObjectSchema(__unsafe_unretained id const value,
                     RLMDynamicSet(object, prop, RLMNSNullToNil(propValue), creationOptions);
                 }
             }
-            else if (created &&!prop.optional) {
-                @throw RLMException(@"Missing property value",
-                                    @{@"Property name:" : prop.name ?: @"nil",
-                                      @"Value": propValue ? [propValue description] : @"nil"});
+            else if (created && !prop.optional) {
+                @throw RLMException([NSString stringWithFormat:@"Missing value for property '%@'", prop.name ?: @"nil"]);
```
segiddins: the `nil` coalescing is unnecessary

Realm/RLMRealm.mm
```diff
@@ -172,24 +176,16 @@ void RLMRealmAddPathSettingsToConfiguration(RLMRealmConfiguration *configuration
...
-    // Used for both
-    Group *_group;
-    BOOL _readOnly;
-    BOOL _inMemory;
-
-    NSData *_encryptionKey;
+    NSHashTable *_notificationHandlers;
 }
 
+@dynamic path;
```
segiddins: no need to mark as `@dynamic` since the getter for read-write properties is overriden

Realm/RLMRealm.mm
```diff
@@ -360,123 +313,186 @@ static id RLMAutorelease(id value) {
...
+            case RealmFileException::Kind::AccessError:
+                RLMSetErrorOrThrow(RLMMakeError(RLMErrorFileAccessError, ex), outError);
+                break;
+            default:
+                RLMSetErrorOrThrow(RLMMakeError(RLMErrorFail, ex), outError);
+                break;
+        }
+    }
+    catch (std::system_error const& ex) {
+        RLMSetErrorOrThrow([NSError errorWithDomain:NSPOSIXErrorDomain code:ex.code().value() userInfo:nil], outError);
```
segiddins: add an `RLMMakeError` overload for this?

Realm/RLMRealm.mm
```diff
@@ -360,123 +313,186 @@ static id RLMAutorelease(id value) {
...
         }
-        return RLMAutorelease(realm);
     }
 
-    NSData *key = configuration.encryptionKey ?: keyForPath(path);
-    realm = [[RLMRealm alloc] initWithPath:path key:key readOnly:readOnly inMemory:inMemory dynamic:dynamic error:error];
-    if (error && *error) {
-        return nil;
+    if (!configuration.encryptionKey) {
+        configuration.encryptionKey = keyForPath(path);
```
segiddins: should be set on `config` rather than `configuration` ?

Realm/RLMRealm.mm
```diff
@@ -360,123 +313,186 @@ static id RLMAutorelease(id value) {
...
     }
 
-    // we need to protect the realm cache and accessors cache
+    RLMRealm *realm = [RLMRealm new];
+    realm->_dynamic = dynamic;
+
+    config.migration_function = [=](SharedRealm old_realm, SharedRealm realm) {
+        RLMMigrationBlock userBlock = configuration.migrationBlock ?: migrationBlockForPath(path);
```
segiddins: we should avoid capturing the configuration in the lambda, and pull the properties in question off outside of it?

Realm/RLMRealm.mm
```diff
@@ -360,123 +313,186 @@ static id RLMAutorelease(id value) {
...
+            [[[RLMMigration alloc] initWithRealm:newRealm oldRealm:oldRealm] execute:userBlock];
+
+            oldRealm->_realm = nullptr;
+            newRealm->_realm = nullptr;
+        }
+    };
+
+    // protects the realm cache and accessors cache
     static id initLock = [NSObject new];
     @synchronized(initLock) {
```
segiddins: is this not synchronized in the Object Store Realm now?

Realm/RLMRealmUtil.mm
```diff
@@ -77,227 +85,35 @@ void RLMInstallUncaughtExceptionHandler() {
...
+    // This is owned by the realm, so it needs to not retain the realm
+    __unsafe_unretained RLMRealm *const _realm;
+};
+} // anonymous namespace
+
 
-- (void)notifyOtherRealms {
-    notifyFd(_notifyFd);
+std::unique_ptr<realm::RealmDelegate> RLMCreateRealmDelegate(RLMRealm *realm) {
+    return std::unique_ptr<realm::RealmDelegate>(new RLMNotificationHelper(realm));
```
segiddins: `std::make_unique`?

RealmSwift-swift1.2/Tests/TestUtils.h
```diff
@@ -27,3 +27,5 @@ FOUNDATION_EXTERN void RLMAssertThrows(XCTestCase *self,
 // Forcibly deallocate the RLMRealm for the given path on the main thread
 // Will cause crashes if it's alive for a reason other than being leaked by RLMAssertThrows
 FOUNDATION_EXTERN void RLMDeallocateRealm(NSString *path);
+
+FOUNDATION_EXTERN bool RLMHasCachedRealmForPath(NSString *path);
```
segiddins: trailing newline

RealmSwift-swift2.0/Tests/TestUtils.h
```diff
@@ -27,3 +27,5 @@ FOUNDATION_EXTERN void RLMAssertThrows(XCTestCase *self,
 // Forcibly deallocate the RLMRealm for the given path on the main thread
 // Will cause crashes if it's alive for a reason other than being leaked by RLMAssertThrows
 FOUNDATION_EXTERN void RLMDeallocateRealm(NSString *path);
+
+FOUNDATION_EXTERN bool RLMHasCachedRealmForPath(NSString *path);
```
segiddins: trailing newline

Realm/RLMRealmUtil.mm
simonask: `std::make_unique` is C++14, and we're only on C++11 unfortunately. :(

Realm/ObjectStore/object_store.cpp
tgoyne: This doesn't do anything useful for that (it checks if a write transaction is needed, not if it can be opened without a migration).

Realm/ObjectStore/shared_realm.cpp
tgoyne: No, current obj-c code doesn't check that either.

Realm/ObjectStore/transact_log_handler.cpp
tgoyne: It handles stuff related to both read and write transactions.

Realm/RLMRealm.mm
tgoyne: No, setting it on `configuration` updates `config` and avoids duplicating the conversion logic.

Realm/RLMRealm.mm
tgoyne: No.

Realm/RLMRealmUtil.mm
tgoyne: The type of the object being constructed and the type of the pointer being returned don't match (since the whole point is to avoid having to expose the definition of the concrete type and thus drag in extra headers), and `unique_ptr<Derived>` doesn't implicitly convert to `unique_ptr<Base>`.

Realm/ObjectStore/object_store.cpp
tgoyne: I still think it's functionality we want, but I was wondering if it'd be better done at some other point in the process and didn't get around to dealing with it.

Realm/RLMMigration.mm
```diff
@@ -106,20 +99,17 @@ - (void)enumerateObjects:(NSString *)className block:(RLMObjectMigrationBlock)bl
 
 - (void)execute:(RLMMigrationBlock)block {
     @autoreleasepool {
-        // copy old schema and reset after migration
```
jpsim: Why can this now be removed?

Realm/RLMRealm.mm
```diff
@@ -498,22 +515,29 @@ + (void)resetRealmState {
     clearMigrationCache();
     clearKeyCache();
     RLMClearRealmCache();
+    realm::Realm::s_global_cache.clear();
```
jpsim: Why is this not in `RLMClearRealmCache`?

Realm/ObjectStore/shared_realm.cpp
```diff
@@ -0,0 +1,446 @@
...
+#include <realm/commit_log.hpp>
+#include <realm/group_shared.hpp>
+
+#include <mutex>
+
+using namespace realm;
+
+RealmCache Realm::s_global_cache;
+
+Realm::Config::Config(const Config& c)
```
jpsim: This could conceivably also have `dynamic`.

Realm/Tests/InterprocessTests.m
```diff
@@ -173,7 +173,7 @@ - (void)testBidirectionalCommunication {
     }
 
     RLMNotificationToken *token = [realm addNotificationBlock:^(__unused NSString *note, __unused RLMRealm *realm) {
-        if (obj.intCol % 2 != self.isParent && obj.intCol < stopValue) {
+        if (obj.intCol % 2 == self.isParent && obj.intCol < stopValue) {
```
jpsim: Does this change actually make a difference? Smells fishy.

jpsim:
I'm definitely not the best person to review this, but I did anyway. There's of course a ton of code, so we should be extra careful, but it looks good to me. :+1:

Realm/ObjectStore/shared_realm.cpp
tgoyne: The dynamic schema is now always created (and used for the validation, so it's never wasted work), so for dynamic you just don't pass in a custom schema or call update_schema().

Realm/RLMMigration.mm
tgoyne: The migration is now run on an RLMRealm instance created specifically for the migration with a copy of the schema rather than on the RLMRealm that ends up being returned to the user and the post-migration validation doesn't use the RLMSchema, so there's no need to undo the changes to it.

Realm/Tests/InterprocessTests.m
tgoyne: The previous version didn't match the intent of the test, but happened to work due to a race condition. The new version also passes on master.

tgoyne:
This is now rebased onto master (plus #2680), and all tests should pass other than the ones testing the unimplemented null migration stuff.

bdash:
I think this is good to go now.

bdash:
By my reading, only 4df604ff...633470b are yet to be reviewed. The earlier changes have been reviewed both here and and separately for the object store.

tgoyne:
Null migration fixes lgtm.

jpsim:
Same opinion as last time, I'm definitely not the best person to review this, but what I've seen looks good to me. :+1:

jpsim:
I believe this can be merged.

PR #2263, opened 2015-07-21
Split "Set Swift Version" into iOS and OS X versions

jpsim:
This fixes an issue in `Realm.xcodeproj` where iOS destinations were available for OS X targets and vice-versa. /cc @segiddins @alazier


kishikawakatsumi:
:+1: 
I confirmed that the OS X build configuration is removed from each iOS targets.

segiddins:
:+1: until we move to a universal target

PR #2273, opened 2015-07-22
[List][RLMArray] Allow moving objects from one index to another

### LINKED ISSUE #2242: [RLMArray][List] Implement Move method
To support drag'n'drop reordering in lists we need a move method.

This is needed by sync as a move is semantically very different from a delete and insert. It is also important to avoid triggering cascading deletes when we start supporting that.
```obj-c
- (void)moveObjectFromIndex:(NSUInteger)from toIndex:(NSUInteger)to;
```

segiddins:
Closes #2242.

Depends upon https://github.com/realm/realm-core/pull/955.

### DIFF
--- a/Realm/RLMArray.h
+++ b/Realm/RLMArray.h
@@ -185,15 +185,15 @@ RLM_ASSUME_NONNULL_BEGIN
 - (void)replaceObjectAtIndex:(NSUInteger)index withObject:(RLMObjectArgument)anObject;
 
 /**
- Moves the object at from the given source index to the given destination index.
+ Moves the object at the given source index to the given destination index.
 
  Throws an exception when called with an index greater than or equal to the
  number of objects in this RLMArray.
 
  @warning This method can only be called during a write transaction.
 
- @param sourceIndex      The list index of the object to be replaced.
- @param destinationIndex An object to replace at the specified index.
+ @param sourceIndex      The index of the object to be moved.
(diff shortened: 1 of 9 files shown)


segiddins:
Failure is unrelated.

\c @jpsim

Realm/RLMArray.h
```diff
@@ -178,13 +178,26 @@ RLM_ASSUME_NONNULL_BEGIN
...
  @warning This method can only be called during a write transaction.
- 
+
  @param index       The array index of the object to be replaced.
  @param anObject    An object (of the same type as returned from the objectClassName selector).
  */
 - (void)replaceObjectAtIndex:(NSUInteger)index withObject:(RLMObjectArgument)anObject;
 
 /**
+ Moves the object at from the given source index to the given destination index.
```
jpsim: > Moves the object from the given source index to the given destination index.

Same change needs to be applied on all 3 versions of these docs.

Realm/RLMArray.h
```diff
@@ -178,13 +178,26 @@ RLM_ASSUME_NONNULL_BEGIN
...
+
+ Throws an exception when called with an index greater than or equal to the
```
jpsim: > Throws an exception when called with an index exceeding the bounds of this array.

jpsim:
:+1:

PR #2279, opened 2015-07-23
prelaunch-simulator before running ios-swift in package-ios-swift

jpsim:
Otherwise the "Package iOS Swift" job fails. /cc @segiddins


jpsim:
(aborted CI PR job)

segiddins:
:+1:

PR #2280, opened 2015-07-23
[Realm Browser Plugin] Add Sanity check to alert user when more than file is being opened

kittinunf:
This PR is a sequel from #2171. As suggested that, plugin should at least warn user in the case where 2 or more files are detected and about to open. This PR aims to fix this.

@TimOliver, I were trying several things, but I got no luck. :crying_cat_face: 

Using `[[NSWorkspace sharedWorkspace] launchApplicationAtURL:self.browserUrl options:0 configuration:configuration error:&error]`. or replacing with `options:NSWorkspaceLaunchDefault` seems to work, but it fails to open Realm File on the subsequent runs (meaning that closing the Realm file window without quitting the Realm Browser and go to File > Open Realm ... in order to open Browser with Realm file again)

I am not familiar enough with Apple Script so unfortunately I couldn't make the `[NSWorkspace openURLs:withAppBundleIdentifier:options:additionalEventParamDescriptor:launchIdentifiers:]` to work. 
:(

So, I decided that to the least having a check from plugin's standpoint is doable from my side. I went ahead and implemented those quickly in this PR. If you have a smarter way for this, I am all ears.

### DIFF
--- a/plugin/RealmPlugin/RLMPRealmPlugin.m
+++ b/plugin/RealmPlugin/RLMPRealmPlugin.m
@@ -147,12 +147,31 @@ static NSArray * RLMPGlobFilesAtDirectoryURLWithPredicate(NSFileManager *fileMan
     
     NSDictionary *configuration = @{ NSWorkspaceLaunchConfigurationArguments : arguments };
     
+    // Show confirmation alert if 2 or more files are detected
+    if (arguments.count > 1) {
+        NSAlert *alert = [[NSAlert alloc] init];
+        [alert setMessageText:@"Realm Browser"];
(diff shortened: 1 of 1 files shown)


TimOliver:
Hi @kittinunf! Thanks a lot for your follow-up pull request!

Ahh, okay then. It looks like we've ironically got just the wrong set of API combinations to make that possible. That's a bit of a shame. Sorry to have made you go away and research that! I might keep it in mind and see if there's a better way to do it (Even if it means some kind of direct communication between the plugin and the Browser).

I reviewed your code, and that looks like the best solution at present to me. That'll cover the portions of the multiple file handling operations that weren't getting covered on Realm Browser's side.

@jpsim I'm happy with this pull request. Shall we merge it in?

PR #2281, opened 2015-07-23
Delete dynamic framework before building each Swift version for package-ios-swift

jpsim:


jpsim:
`build.sh package-ios-swift` succeeds locally for me with this change and it makes sense to require the dynamic framework to be built with the same version of Xcode.

jpsim:
(Canceled the CI PR job because this only affects packaging, which isn't tested in PRs).

segiddins:
:+1:

PR #2287, opened 2015-07-24
Erase all available simulators when prelaunching

jpsim:
This should hopefully resolve issues on CI where the iOS Simulator doesn't boot when switching between Xcode 6 and Xcode 7. /cc @segiddins


segiddins:
:+1: once CI runs

PR #2291, opened 2015-07-24
[RLMTestCase] Ensure waitForNotification has an autoreleasepool

segiddins:
Might alleviate issues with testAutorefreshAfterBackgroundUpdate hanging

\c @jpsim

### DIFF
--- a/Realm/Tests/RLMTestCase.m
+++ b/Realm/Tests/RLMTestCase.m
@@ -149,15 +149,20 @@ static BOOL encryptTests() {
 
 - (void)waitForNotification:(NSString *)expectedNote realm:(RLMRealm *)realm block:(dispatch_block_t)block {
     XCTestExpectation *notificationFired = [self expectationWithDescription:@"notification fired"];
-    RLMNotificationToken *token = [realm addNotificationBlock:^(__unused NSString *note, RLMRealm *realm) {
+    RLMNotificationToken *token = [realm addNotificationBlock:^(NSString *note, RLMRealm *realm) {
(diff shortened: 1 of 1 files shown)


segiddins:
retest this please

jpsim:
Doesn't actually fix the hanging `testAutorefreshAfterBackgroundUpdate`, but I do believe these changes to be good anyway, so :+1:.

Realm/Tests/RLMTestCase.m
```diff
@@ -149,15 +149,20 @@ - (RLMRealm *)realmWithTestPathAndSchema:(RLMSchema *)schema {
 
 - (void)waitForNotification:(NSString *)expectedNote realm:(RLMRealm *)realm block:(dispatch_block_t)block {
     XCTestExpectation *notificationFired = [self expectationWithDescription:@"notification fired"];
-    RLMNotificationToken *token = [realm addNotificationBlock:^(__unused NSString *note, RLMRealm *realm) {
+    RLMNotificationToken *token = [realm addNotificationBlock:^(NSString *note, RLMRealm *realm) {
+        XCTAssertNotNil(note, @"Note should not be nil");
         XCTAssertNotNil(realm, @"Realm should not be nil");
-        if (note == expectedNote) {
+        if ([note isEqualToString:expectedNote]) {
```
tgoyne: This was deliberately checking pointer equality so that it would fail if the string constant for the notification is not used.

jpsim: Good point. Reverting in #2336.

PR #2296, opened 2015-07-25
updated Swift 2.0 doc comments to new syntax

jpsim:
I haven't added `- Throws:` yet and I can't tell if `- see:` and `- warning:` do anything special since option-clicking in Xcode 7 Beta 4 doesn't work (which is actually a known bug in the release notes).

### DIFF
--- a/RealmSwift-swift2.0/List.swift
+++ b/RealmSwift-swift2.0/List.swift
@@ -79,9 +79,9 @@ public final class List<T: Object>: ListBase {
     /**
     Returns the index of the given object, or `nil` if the object is not in the list.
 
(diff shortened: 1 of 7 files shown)


segiddins:
Pretty sure jazzy will at least handle `- see:` and `- warning:` specially, and add them to the table/dl?

jpsim:
Jazzy will detect `- warning:` as a special token, but not `- see:`, treating that as a normal markdown list item. Xcode 7 Beta 4 treats `warning` specially, but not `see`.

segiddins:
:+1:

PR #2299, opened 2015-07-27
Remove RLM_GENERIC_ARRAY macro

jpsim:
We determined that it was better to encourage users to use the full (albeit unsightly) representation rather than hiding it behind a macro. /cc @segiddins


segiddins:
The changes to the tests make it so we can no longer compile them in Xcode 6, I believe?

jpsim:
@segiddins I reintroduced the macro just for the tests.

segiddins:
@jpsim I'd prefer we actually put the macro _in_ the tests if we're going to use it. Right now, we're still shipping it

segiddins:
:+1:

PR #2300, opened 2015-07-27
[RLMResults][Results] Add handover to a target dispatch queue

### LINKED ISSUE #1097: Confine on dispatch queue rather than thread
tl;dr Realm uses thread ids and throws exceptions when they change.  Checking thread ID is not acceptable under GCD.

Realm docs [state](http://realm.io/docs/cocoa/0.87.1/#background-operations) objects are not gcd queue safe and when running under GCD:

> you must get an RLMRealm instance in each thread/dispatch_queue in which you want to read or write

I do this:

```
    // Local copy of the persisted task valid for the worker queue.
__block OFEPersistedTask* taskCopyOnWorkerQueue = nil;

NSString* taskID = aTask.taskID;
dispatch_sync(_workerQueue, ^{
    NSPredicate *pred = [NSPredicate predicateWithFormat:@"taskID = %@", taskID];
    RLMResults *results = [OFEPersistedTask objectsWithPredicate:pred];
    taskCopyOnWorkerQueue = [results firstObject];
}
```

To get a copy of my Task (RLMObject subclass) on the correct queue. Unfortunately, in `RLMVerifyAttached` Realm checks the thread ID through `RLMCheckThread` (i.e. not queue) which does not work under GCD.

A GCD queue can run its tasks on many different threads - thats kindof the point. From the [Apple docs](https://developer.apple.com/library/ios/documentation/General/Conceptual/ConcurrencyProgrammingGuide/OperationQueues/OperationQueues.html):

> The currently executing task runs on a distinct thread **(which can vary from task to task)**

My **emphasis**... So, even if I grab a RLMObject on my queue I still risk getting killed (actually, no risk, I do get killed) later for running on the wrong thread.

The only workaround I can think of is to store a primary key with __block scope and then perform another Realm fetch at the beginning of every task. That feels so suboptimal I don't think you can claim Realm is compatible with GCD _at all_.

segiddins:
Closes #1097.

### DIFF
--- a/Realm/RLMResults.h
+++ b/Realm/RLMResults.h
@@ -179,7 +179,7 @@ RLM_ASSUME_NONNULL_BEGIN
  @param queue The dispatch queue onto which the results should be delivered.
  @param block The block to be called on the given `queue` with the queue-local copy of the results.
  */
-- (void)deliverOnQueue:(dispatch_queue_t)queue block:(void (^)(RLMResults RLM_GENERIC_RETURN*))block;
+- (void)deliverOnQueue:(dispatch_queue_t)queue block:(void (^)( RLMResults RLM_GENERIC_RETURN * __nullable , NSError * __nullable ))block;
 
 #pragma mark -
 
--- a/Realm/RLMResults.h
+++ b/Realm/RLMResults.h
@@ -167,6 +167,24 @@ RLM_ASSUME_NONNULL_BEGIN
 
 
 /**---------------------------------------------------------------------------------------
+ *  @name Delivery
+ *  ---------------------------------------------------------------------------------------
+ */
+
+/**
+ Asynchronously yields a queue-local copy of the results onto the given dispatch queue.
+
+ @warning This method cannot be called during a write transaction, or when the containing realm is read-only.
+
+ @param queue The dispatch queue onto which the results should be delivered.
+ @param block The block to be called on the given `queue` with the queue-local copy of the results.
+ */
+- (void)deliverOnQueue:(dispatch_queue_t)queue block:(void (^)(RLMResults RLM_GENERIC_RETURN*))block;
+
+#pragma mark -
+
+
+/**---------------------------------------------------------------------------------------
  *  @name Aggregating Property Values
  *  ---------------------------------------------------------------------------------------
  */
--- a/Realm/RLMResults.mm
+++ b/Realm/RLMResults.mm
@@ -75,6 +75,7 @@ using namespace realm;
 
 + (instancetype)resultsWithObjectClassName:(NSString *)objectClassName
                                      query:(std::unique_ptr<realm::Query>)query
+                                      sort:(RowIndexes::Sorter const&)sorter
                                       view:(realm::TableView &&)view
                                      realm:(RLMRealm *)realm {
     RLMResults *ar = [[RLMResults alloc] initPrivate];
@@ -82,6 +83,7 @@ using namespace realm;
     ar->_viewCreated = YES;
     ar->_backingView = std::move(view);
     ar->_backingQuery = move(query);
+    ar->_sortOrder = sorter;
     ar->_realm = realm;
     ar->_objectSchema = realm.schema[objectClassName];
     return ar;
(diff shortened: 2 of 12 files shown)


Realm/RLMResults.mm
```diff
@@ -312,6 +314,109 @@ - (RLMResults *)sortedResultsUsingDescriptors:(NSArray *)properties {
...
+        BOOL readOnly;
+        BOOL inMemory;
+        BOOL dynamic;
+        RLMSchema *schema;
+    };
+
+    RLMRealm *createWithRealmCreation(RealmCreation realmCreation, SharedGroup::VersionID version) {
+        try {
+            NSError *realmCreationError = nil;
+            RLMRealm *realm = [RLMRealm realmWithPath:realmCreation.path key:realmCreation.key readOnly:realmCreation.readOnly inMemory:realmCreation.inMemory dynamic:realmCreation.dynamic schema:realmCreation.schema error:&realmCreationError];
```
jpsim: we probably want to pass `nil` for the error pointer here so that we throw if the realm can't be created.

Realm/RLMResults.mm
```diff
@@ -312,6 +314,109 @@ - (RLMResults *)sortedResultsUsingDescriptors:(NSArray *)properties {
     return [RLMResults resultsWithObjectClassName:self.objectClassName query:move(query) sort:sorter realm:_realm];
 }
 
+namespace {
+    struct RealmCreation {
```
jpsim: `RealmConfiguration` would be a more appropriate name

Realm/RLMResults.mm
segiddins: No, we specifically _dont_ want to throw since we're on a background thread

Realm/RLMResults.mm
segiddins: I was waiting until https://github.com/realm/realm-cocoa/issues/1584 to properly create a `RealmConfiguration`

RealmSwift-swift1.2/Results.swift
```diff
@@ -223,6 +223,22 @@ public final class Results<T: Object>: ResultsBase {
...
+
+    /**
+    Asynchronously yields a queue-local copy of the results onto the given dispatch queue.
+
+    :warning: This method cannot be called during a write transaction, or when the containing realm is read-only.
+
+    :param: queue The dispatch queue onto which the results should be delivered.
+    :param: block The block to be called on the given `queue` with the queue-local copy of the results.
+    */
+    public func deliver(onQueue queue: dispatch_queue_t = dispatch_get_main_queue(), block: Results<T> -> Void) {
```
jpsim: I think we should `@noescape` this block since it should only be run on the given queue?

segiddins: We can't, since the block _always_ escapes

RealmSwift-swift2.0/Results.swift
```diff
@@ -223,6 +223,22 @@ public final class Results<T: Object>: ResultsBase {
         return Results<T>(rlmResults.sortedResultsUsingDescriptors(sortDescriptors.map { $0.rlmSortDescriptorValue }))
     }
 
+    // MARK: Delivery
+
+    /**
+    Asynchronously yields a queue-local copy of the results onto the given dispatch queue.
+
+    :warning: This method cannot be called during a write transaction, or when the containing realm is read-only.
```
jpsim: Please use Swift 2.0 style documentation comments here.

jpsim:
The one thing that we're not doing here is error handling. I think it's worth the ability to pass in an NSError pointer in the public API and using `throws` in the Swift 2 case.

Realm/RLMResults.h
```diff
@@ -167,6 +167,24 @@ RLM_ASSUME_NONNULL_BEGIN
...
+
+/**
+ Asynchronously yields a queue-local copy of the results onto the given dispatch queue.
+
+ @warning This method cannot be called during a write transaction, or when the containing realm is read-only.
+
+ @param queue The dispatch queue onto which the results should be delivered.
+ @param block The block to be called on the given `queue` with the queue-local copy of the results.
+ */
+- (void)deliverOnQueue:(dispatch_queue_t)queue block:(void (^)( RLMResults RLM_GENERIC_RETURN * __nullable , NSError * __nullable ))block;
```
tgoyne: Maybe just `deliverOn:`?

There's some extra whitespace here.

Realm/RLMResults.mm
```diff
@@ -312,6 +314,120 @@ - (RLMResults *)sortedResultsUsingDescriptors:(NSArray *)properties {
...
+        std::unique_ptr<SharedGroup::Handover<Query>> queryHandover = sharedGroup->export_for_handover(*[self cloneQuery], ConstSourcePayload::Stay);
+        if (_viewCreated) {
+            auto tableViewHandoverPtr = sharedGroup->export_for_handover(_backingView, ConstSourcePayload::Stay).release();
+            deliverQuery(queryQueue, queryQueue, realmCreation, _objectClassName, queryHandover.release(), tableViewHandoverPtr, _sortOrder, block);
+        }
+        else {
+            queryOnBackgroundQueue(queryQueue, queue, realmCreation, _objectClassName, std::move(queryHandover), _sortOrder, block);
+        }
+    }
+    catch (std::exception const& exception) {
```
tgoyne: What exceptions is this catching?

segiddins: `export_for_handover` throws, in particular when a write transaction is open

segiddins:
Closing this because it's probably the wrong approach and a gross misuse of the Core APIs.

PR #2302, opened 2015-07-27
use temporary path for in-memory realms

### LINKED ISSUE #2032: Write In-Memory realms to a tmpdir

jpsim:
Closes #2032.

Otherwise, apps using in-memory realms risk violating Apple's data storage guidelines. /cc @segiddins

### DIFF
--- a/Realm/RLMRealm.mm
+++ b/Realm/RLMRealm.mm
@@ -276,7 +276,12 @@ static NSString * const c_defaultRealmFileName = @"default.realm";
     s_defaultRealmPath = defaultRealmPath;
 }
 
-+ (NSString *)writeablePathForFile:(NSString*)fileName
++ (NSString *)writeableTemporaryPathForFile:(NSString *)fileName
+{
+    return [NSTemporaryDirectory() stringByAppendingPathComponent:fileName];
+}
+
++ (NSString *)writeablePathForFile:(NSString *)fileName
 {
 #if TARGET_OS_IPHONE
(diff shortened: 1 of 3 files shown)


Realm/RLMRealm.mm
```diff
@@ -276,7 +276,12 @@ + (void)setDefaultRealmPath:(NSString *)defaultRealmPath {
     s_defaultRealmPath = defaultRealmPath;
 }
 
-+ (NSString *)writeablePathForFile:(NSString*)fileName
++ (NSString *)writeableTemporaryPathForFile:(NSString *)fileName
+{
+    return [NSTemporaryDirectory() stringByAppendingPathComponent:fileName];
```
segiddins: I believe using `
- URLForDirectory:inDomain:appropriateForURL:create:error:` is recommended over `NSTemporaryDirectory`

segiddins:
Add a test for this?

Realm/RLMRealm.mm
jpsim: Apple's docs don't say why `URLForDirectory` is preferable, but I suspect it's because it exposes more functionality, which we don't need in this case, so I'd say it's safe to use `NSTemporaryDirectory()` here.

jpsim:
> Add a test for this?

Done.

jpsim:
CI failure is due to a single performance test failure (37% stddev) which seems entirely unrelated.

segiddins:
:+1:

PR #2303, opened 2015-07-27
add favicon to html docs. fixes #2220

### LINKED ISSUE #2220: Favicon for api docs

jpsim:
Fixes #2220 /cc @yoshyosh

### DIFF
--- a/docs/templates/objc/html/document-template.html
+++ b/docs/templates/objc/html/document-template.html
@@ -1,6 +1,15 @@
 <!DOCTYPE HTML>
 <html>
 	<head>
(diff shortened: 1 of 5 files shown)


jpsim:
I opted to link to the favicons on realm.io so we wouldn't have to maintain a separate set here. @yoshyosh are these URLs stable?

yoshyosh:
I believe they should be, can you confirm @mrh-is ?

mrh-is:
They've been that way for a while, and we can make them stable going forward. If you like them somewhere else, I can move them very easily.

tl;dr: Yes, they are stable.

jpsim:
Cool! Thanks.

PR #2309, opened 2015-07-29
Update Core to 0.92.0

segiddins:
\c @jpsim


jpsim:
:+1: CI seems to be happy with this change, and I've confirmed that the core release notes don't have any user-facing ramifications.

PR #2317, opened 2015-07-29
[Examples][Installation] Test Carthage with Swift 2.0

jpsim:
/cc @segiddins


jpsim:
We should also be doing this for CocoaPods.

segiddins:
:+1:

PR #2319, opened 2015-07-30
Help prevent className from being overridden

jpsim:
I've seen users assume that they could override `+[RLMObject className]`/`Object.className` to return something other than the actual exact class name (e.g. to removing a prefix).

As far as I know, we don't state anywhere that overriding this doesn't actually do anything, so might as well just discourage it in Objective-C and prevent it from happening in Swift. /cc @segiddins @bigfish24

### DIFF
--- a/RealmSwift-swift2.0/Object.swift
+++ b/RealmSwift-swift2.0/Object.swift
@@ -103,8 +103,10 @@ public class Object: RLMObjectBase {
     /// Returns a human-readable description of this object.
     public override var description: String { return super.description }
 
+    /// Helper to return the class name for an Object subclass.
+    public final override var className: String { return super.className }
 
-    // MARK: Object customization
+    // MARK: Object Customization
 
     /**
     Override to designate a property as the primary key for an `Object` subclass. Only properties of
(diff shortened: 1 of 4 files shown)


RealmSwift-swift2.0/Object.swift
```diff
@@ -221,6 +204,35 @@ public class Object: RLMObjectBase {
         }
     }
 
+    // MARK: Equatable
+
+    /// Returns whether both objects are equal.
+    /// Objects are considered equal when they are both from the same Realm
+    /// and point to the same underlying object in the database.
+    public override func isEqual(object: AnyObject?) -> Bool {
```
segiddins: this method actually is from `NSObject`, not `Equatable`

jpsim: Yes, but we're only adding it for Equatable conformance, so the mark comment is still appropriate IMO.

jpsim:
This is ready.

segiddins:
:+1:

tgoyne:
Obj-c runtime stuff can check if a class overrides a method, so we could make this a runtime error in obj-c too.

jpsim:
Sam and I discussed that too, but opted against it since it doesn't affect Realm's behavior. I felt it was more of a documentation issue, with the added benefit that it's essentially free to do in Swift.

PR #2321, opened 2015-07-30
RealmCollectionType

jpsim:
/cc @segiddins


segiddins:
What about `realm`, `invalidated, and KVC?

jpsim:
* `realm` is optional on `List`
* `invalidated` can't happen to a `Results`
* `KVC` is defined on `NSObject`

segiddins:
- it should be fine to define it as an optional on the protocol but make it non-optional on `Results`
- it actually can, if the realm is invalidated
- true, but `RealmCollectionType` isn't limited to `NSObject` types, as far as I can see

jpsim:
> it should be fine to define it as an optional on the protocol but make it non-optional on `Results`

That doesn't work.

> it actually can, if the realm is invalidated

We don't expose that at the moment. Create a new issue if you think we should in both Objective-C and Swift results.

> true, but `RealmCollectionType` isn't limited to `NSObject` types, as far as I can see

Ugh, I can't make `RealmCollectionType` conform to `NSObject` or `@objc` because then it couldn't also conform to `CollectionType`. I could make a `KeyValueEncodable` protocol I suppose...

astigsen:
>> What about `realm`?
> `realm` is optional on `List`

We definitely need some way to access the realm from a `RealmCollectionType`. Otherwise you can't do any transactions on it without having to communicate the realm off-band.

Would it be enough to make the `realm` variable optional in `Results` as well?

jpsim:
> Would it be enough to make the realm variable optional in Results as well?

We could, but that would slightly weaken the interface on `Results`. I suppose that's a worthwhile tradeoff, though. I'll do that.

jpsim:
Done.

RealmSwift-swift1.2/RealmCollectionType.swift
```diff
@@ -0,0 +1,203 @@
...
+////////////////////////////////////////////////////////////////////////////
+
+import Foundation
+import Realm
+
+/**
+Encapsulates iteration state and interface for iteration over a
+`RealmCollectionType`.
+*/
+public class RLMGenerator<T: Object>: GeneratorType {
```
segiddins: `final`?

RealmSwift-swift1.2/RealmCollectionType.swift
```diff
@@ -0,0 +1,203 @@
...
+    */
+    func indexOf(predicateFormat: String, _ args: CVarArgType...) -> Int?
+
+
+    // MARK: Object Retrieval
+
+    /**
+    Returns the object at the given `index`.
+
+    :warning: You can only set an object during a write transaction.
```
segiddins: not true for standalone lists

RealmSwift-swift1.2/RealmCollectionType.swift
```diff
@@ -0,0 +1,203 @@
...
+}
+
+/**
+A homogenous collection of `Object`s which can be retrieved, filtered, sorted,
+and operated upon.
+*/
+public protocol RealmCollectionType: CollectionType {
+
+    /// Element type contained in this collection.
+    typealias Element: Object
```
segiddins: should the type alias be public?

RealmSwift-swift2.0/RealmCollectionType.swift
```diff
@@ -0,0 +1,203 @@
...
+    // MARK: Index Retrieval
+
+    /**
+    Returns the index of the given object, or `nil` if the object is not in the collection.
+
+    - parameter object: The object whose index is being queried.
+
+    - returns: The index of the given object, or `nil` if the object is not in the collection.
+    */
+    func indexOf(object: Element) -> Int?
```
segiddins: already implemented for equatable in swift 2, so this is redundant

RealmSwift-swift2.0/RealmCollectionType.swift
```diff
@@ -0,0 +1,203 @@
...
+    /**
+    Returns the object at the given `index`.
+
+    - warning: You can only set an object during a write transaction.
+
+    - parameter index: The index.
+
+    - returns: The object at the given `index`.
+    */
+    subscript(index: Int) -> Element { get }
```
segiddins: already on `CollectionType`

RealmSwift-swift2.0/RealmCollectionType.swift
```diff
@@ -0,0 +1,203 @@
...
+
+    /// Returns the first object in the collection, or `nil` if empty.
+    var first: Element? { get }
```
segiddins: already on CollectionType

RealmSwift-swift2.0/RealmCollectionType.swift
```diff
@@ -0,0 +1,203 @@
...
+
+    /// Returns the last object in the collection, or `nil` if empty.
+    var last: Element? { get }
```
segiddins: already on CollectionType

RealmSwift-swift1.2/RealmCollectionType.swift
jpsim: Sure.

RealmSwift-swift1.2/RealmCollectionType.swift
jpsim: I missed this.

RealmSwift-swift1.2/RealmCollectionType.swift
jpsim: I don't see how that's a problem.

RealmSwift-swift2.0/RealmCollectionType.swift
jpsim: It's not redundant as this is an optimized implementation.

jpsim: Sorry, you're right that it's redundant in the interface.

jpsim:
@segiddins addressed your feedback. I think the only things left to do are to test aggregate operations on `List`. Marking WIP until that's done.

segiddins:
@jpsim anything we can do to help get this in for the next release?

jpsim:
@tgoyne @segiddins this should now be ready for review.

segiddins:
I think some of the `List` tests are now redundant?

jpsim:
> I think some of the List tests are now redundant?

Good point, fixed in last commit.

RealmSwift-swift1.2/Results.swift
```diff
@@ -78,10 +78,16 @@ Results cannot be created directly.
...
 public final class Results<T: Object>: ResultsBase {
 
+    /// Element type contained in this collection.
+    typealias Element = T
+
     // MARK: Properties
 
     /// Returns the Realm these results are associated with.
-    public var realm: Realm { return Realm(rlmResults.realm) }
+    /// Despite returning an `Optional<Realm>` in order to conform to
```
tgoyne: Can it conform to the protocol by returning an implicitly-unwrapped optional?

jpsim: Unfortunately not.

segiddins: Filed as rdar://22426323

segiddins:
:+1:

RealmSwift-swift1.2/RealmCollectionType.swift
```diff
@@ -0,0 +1,655 @@
...
+    func average<U: AddableType>(property: String) -> U? { fatalError() }
+    subscript(index: Int) -> Element { fatalError() }
+    func generate() -> RLMGenerator<T> { fatalError() }
+    var startIndex: Int { fatalError() }
+    var endIndex: Int { fatalError() }
+    func valueForKey(key: String) -> AnyObject? { fatalError() }
+    func setValue(value: AnyObject?, forKey key: String) { fatalError() }
+}
+
+private class _AnyRealmCollection<C: RealmCollectionType>: _AnyRealmCollectionBase<C.Element> {
```
tgoyne: final

tgoyne:
:+1:

PR #2326, opened 2015-08-01
Add RealmConfiguration / RLMConfiguration

### LINKED ISSUE #1584: Add RLMConfiguration
Encapsulate all global state/config used to configure Realms so that we don't need separate config methods for every parameter.

segiddins:
Closes #1584


Realm/RLMConfiguration.h
```diff
@@ -0,0 +1,115 @@
...
+/// The current schema version.
+@property (nonatomic) NSUInteger schemaVersion;
+
+/// The block which migrates the Realm to the current version.
+@property (nonatomic, copy, nullable) RLMMigrationBlock migrationBlock;
+
+@end
+
+/**
+ Migration block used to migrate a Realm.
```
jpsim: Configuration not migration.

Realm/RLMConfiguration.h
```diff
@@ -0,0 +1,115 @@
...
+/// The path to the realm file.
+@property (nonatomic, copy, nullable) NSString *path;
+
+/// A string used to identify a particular in-memory Realm.
+@property (nonatomic, copy, nullable) NSString *inMemoryIdentifier;
+
+/// 64-byte key to use to encrypt the data.
+@property (nonatomic, copy, nullable) NSData *encryptionKey;
+
+/// Whether the Realm is read-only (must be used for read-only files).
```
jpsim: Must be `YES` for read-only files.

Realm/RLMConfiguration.h
```diff
@@ -0,0 +1,115 @@
...
+/**
+ An `RLMConfiguration` is used to describe the different options used to
+ create an `RLMRealm` instance.
+ */
+@interface RLMConfiguration : NSObject<NSCopying>
+
+/**
+ Returns the default configuration used to create realms.
+
+ @return defaultConfiguration The default realm configuration.
```
jpsim: Shouldn't name the return value.

Realm/RLMConfiguration.h
```diff
@@ -0,0 +1,115 @@
...
+
+/**
+ Sets the default configuration to the given `RLMConfiguration`.
+
+ @param configuration The new default realm configuration.
+ */
++ (void)setDefaultConfiguration:(nullable RLMConfiguration *)configuration;
+
+/**
+ Constructs a new `RLMConfiguration`, configured via the given configuration block.
```
jpsim: Should state that changes will be applied to the default configuration block.

Realm/RLMConfiguration.h
```diff
@@ -0,0 +1,115 @@
...
+/// 64-byte key to use to encrypt the data.
+@property (nonatomic, copy, readonly, nullable) NSData *encryptionKey;
+
+/// Whether the Realm is read-only (must be used for read-only files).
+@property (nonatomic, readonly) BOOL readOnly;
+
+/// The current schema version.
+@property (nonatomic, readonly) NSUInteger schemaVersion;
+
+/// The block which migrates the Realm to the current version.
```
jpsim: Current schema version.

Realm/RLMConfiguration.m
```diff
@@ -0,0 +1,159 @@
...
+    @"dynamic",
+    @"customSchema",
+};
+static const NSUInteger c_RLMConfigurationPropertiesCount = sizeof(c_RLMConfigurationProperties) / sizeof(NSString *);
+
+@interface RLMConfiguration ()
+
+@property (nonatomic, copy, readwrite) NSString *path;
+@property (nonatomic, copy, readwrite) NSString *inMemoryIdentifier;
+@property (nonatomic, copy, readwrite) NSData *encryptionKey;
```
jpsim: These two lines are half aligned.

Realm/RLMConfiguration.m
```diff
@@ -0,0 +1,159 @@
...
+@property (nonatomic, readwrite)       BOOL    readOnly;
+@property (nonatomic, copy, readwrite) NSString *fileProtectionAttributes;
+
+@property (nonatomic, readwrite)       NSUInteger schemaVersion;
+@property (nonatomic, copy, readwrite) RLMMigrationBlock migrationBlock;
+
+@end
+
+@interface RLMConfiguration ()<RLMConfigurator>
```
jpsim: Why the empty interface? You could move the RLMConfigurator conformance to the previous interface in this file.

Realm/RLMConfiguration.h
segiddins: ?

Realm/RLMConfiguration.m
```diff
@@ -0,0 +1,159 @@
...
+                                                        error:nil];
+    }
+#endif
+    return [path stringByAppendingPathComponent:fileName];
+}
+
++ (instancetype)defaultConfiguration {
+    static dispatch_once_t onceToken;
+    dispatch_once(&onceToken, ^{
+        if (!s_defaultConfiguration) {
```
jpsim: It'll be slightly more efficient to do this check before the dispatch once block.

Realm/RLMConfiguration.m
segiddins: only by coincidence

Realm/RLMConfiguration.m
segiddins: For some reason, the protocol conformance had to come after the class added the properties, but this was from back in April. I can double check

Realm/RLMConfiguration.m
segiddins: why? the dispatch_once block, after it's run the first time, is basically only a conditional

Realm/RLMConfiguration.h
jpsim: I was imagining this block acting like `copyWithChanges:` performed on the default configuration.

segiddins: But if you look at the implementation, it doesn't. It could.

Realm/RLMConfiguration.m
jpsim: In the case that it's set first, you get to skip that conditional. Slight, but still there.

Realm/RLMConfiguration.m
```diff
@@ -0,0 +1,159 @@
...
+    self = [super init];
+    if (self) {
+        self.path = [[self class] defaultRealmPath];
+    }
+
+    return self;
+}
+
+- (instancetype)copyWithZone:(__unused NSZone *)zone {
+    return self;
```
jpsim: You have to make an actual copy here, otherwise mutating the copy will mutate the original.

segiddins: The public interface is entirely immutable?

jpsim: Oh, I see what you're doing. This is fine. A comment could be useful here though.

Realm/RLMConfiguration.h
jpsim: Actually, the implementation applies changes to the original default configuration, which is even more confusing. It should certainly apply to the new default configuration.

Realm/RLMConfiguration.m
```diff
@@ -0,0 +1,159 @@
...
+}
+
+- (instancetype)copyWithChanges:(void (^)(id<RLMConfigurator>))block {
+    RLMConfiguration *configuration = [[[self class] alloc] init];
+    for (NSUInteger i = 0; i < c_RLMConfigurationPropertiesCount; i++) {
+        NSString *key = c_RLMConfigurationProperties[i];
```
jpsim: This could be so much simpler if calling copy with zone made an actual copy.

Realm/RLMRealm.mm
```diff
@@ -390,9 +381,9 @@ + (instancetype)realmWithPath:(NSString *)path
         return RLMAutorelease(realm);
     }
 
-    key = key ?: keyForPath(path);
-    realm = [[RLMRealm alloc] initWithPath:path key:key readOnly:readonly inMemory:inMemory dynamic:dynamic error:outError];
-    if (outError && *outError) {
+    NSData *key = configuration.encryptionKey ?: keyForPath(path);
```
jpsim: Does this key ever get validated? It would be worth considering moving that validation to RLMConfiguration.

Realm/RLMConfiguration.m
```diff
@@ -0,0 +1,159 @@
...
+        }
+    });
+    return s_defaultConfiguration;
+}
+
++ (void)setDefaultConfiguration:(RLMConfiguration *)configuration {
+    s_defaultConfiguration = configuration ?: [[RLMConfiguration alloc] init];
+}
+
++ (instancetype)configurationWithBlock:(void (^)(id<RLMConfigurator>))block {
```
jpsim: Make this block noescape?

RealmSwift-swift1.2/RealmConfiguration.swift
```diff
@@ -0,0 +1,97 @@
...
+    }
+
+    private static func fromRLMConfiguration(rlmConfiguration: RLMConfiguration) -> RealmConfiguration {
+        var configuration = RealmConfiguration()
+        configuration.path = rlmConfiguration.path
+        configuration.inMemoryIdentifier = rlmConfiguration.inMemoryIdentifier
+        configuration.encryptionKey = rlmConfiguration.encryptionKey
+        configuration.readOnly = rlmConfiguration.readOnly
+        configuration.schemaVersion = UInt64(rlmConfiguration.schemaVersion)
+        if let rlmMigrationBlock = rlmConfiguration.migrationBlock {
```
jpsim: Don't we already have a function for this conversion?

RealmSwift-swift1.2/RealmConfiguration.swift
```diff
@@ -0,0 +1,97 @@
...
+    }
+}
+
+// MARK: Printable
+
+extension RealmConfiguration: Printable {
+    /// Returns a human-readable description of the configuration.
+    public var description: String {
+        return gsub("\\ARLMConfiguration", "RealmConfiguration", rlmConfiguration.description) ?? ""
+    }
```
jpsim: `\\A`?

RealmSwift-swift2.0/Realm.swift
```diff
@@ -76,12 +76,22 @@ public final class Realm {
...
 
     /**
-    Obtains a Realm instance persisted at the specified file path. Defaults to
-    `Realm.defaultPath`
+    Obtains a Realm instance with the given configuration. Defaults to the default realm configuration.
+
+    - parameter configuration: The configuration to use when creating the Realm instance.
+    */
+    public convenience init(configuration: RealmConfiguration = RealmConfiguration.defaultConfiguration) throws {
+        print("configuration path: \(configuration.path)")
```
jpsim: Leftover from debugging?

jpsim:
This is moving in a great direction! Well done. Although I'd much prefer we deprecate the older APIs now rather than later. This will need tests and a few changelog entries too.

kishikawakatsumi:
Would you like to support `NSFileProtection` attributes and `NSURLIsExcludedFromBackupKey`?

jpsim:
@kishikawakatsumi this PR is just a first step meant to move our current API and functionality to this new interface. Adding new functionality is out of scope for this PR, but should be pretty simple to add afterwards.

segiddins:
@jpsim I explicitly didn't deprecate the existing APIs because the tests rely on them so heavily

segiddins:
@jpsim would you also like me to deprecate the existing apis for setting encryption keys and schema versions per path?

jpsim:
> would you also like me to deprecate the existing apis for setting encryption keys and schema versions per path?

Yes, it seems logical to me to minimize the number of ways to do the same thing, and I believe we're better off doing so sooner rather than later.

segiddins:
OK. Marking those as deprecated (and cleaning up the tests) will take a bit of work.

Realm/RLMRealm.mm
```diff
@@ -440,18 +431,18 @@ + (instancetype)realmWithPath:(NSString *)path
...
-    if (!readonly) {
-        realm.notifier = [[RLMNotifier alloc] initWithRealm:realm error:outError];
+    if (!readOnly) {
+        realm.notifier = [[RLMNotifier alloc] initWithRealm:realm error:error];
         if (!realm.notifier) {
             return nil;
         }
     }
-
+
```
tgoyne: whitespace

tgoyne:
I don't understand what the goal of `RLMConfigurator` is. It's not a pattern that our users will be familiar with, and AFAICT the only benefit we get from is having to make a copy of the configuration object when setting the default one. If we do actually need an immutable version, `RLMConfiguration`/`RLMMutableConfiguration` would be much more normal and would be a similar amount of code both in the implementation and the code using it.

segiddins:
@tgoyne the reason I went with that pattern is that it's the closest way possible of mimicking structs in Objective-C

tgoyne:
The equivalent of a struct in obj-c is a class with mutable data members that you explicitly copy rather than implicitly copy.

RealmSwift-swift1.2/RealmConfiguration.swift
segiddins: start of string

RealmSwift-swift1.2/RealmConfiguration.swift
segiddins: No, it goes the other way around, `MigrationBlock` -> `RLMMigrationBlock`

Realm/RLMConfiguration.m
segiddins: made the swift compiler crash

Realm/RLMRealm.mm
segiddins: yes, it gets validated at the same place it always has been

segiddins:
OK, changed to just be mutable. I'll add some tests if this design looks better?

Realm/RLMConfiguration.h
```diff
@@ -0,0 +1,64 @@
...
+#import <Foundation/Foundation.h>
+#import <Realm/RLMRealm.h>
+
+RLM_ASSUME_NONNULL_BEGIN
+
+/**
+ An `RLMConfiguration` is used to describe the different options used to
+ create an `RLMRealm` instance.
+ */
+@interface RLMConfiguration : NSObject<NSCopying>
```
tgoyne: There's a few things here that talk about the "default configuration", but nothing ever says what that is or what it's used for.

Realm needs to be capitalized.

Realm/RLMConfiguration.h
```diff
@@ -0,0 +1,64 @@
...
+ Sets the default configuration to the given `RLMConfiguration`.
+
+ @param configuration The new default realm configuration.
+ */
++ (void)setDefaultConfiguration:(nullable RLMConfiguration *)configuration;
+
+/// The path to the realm file.
+@property (nonatomic, copy, nullable) NSString *path;
+
+/// A string used to identify a particular in-memory Realm.
```
tgoyne: Needs to mention that it's mutually-exclusive with path.

segiddins:
Marked things as deprecated. Right now, we're ignoring deprecation warnings in the migration tests (71 of them), and there are some swift warnings due to deprecations, and I don't think there's any way to selectively disable those. Also, marking `defaultPath` as deprecated in Swift caused a bunch of issues, largely because `RealmConfiguration.defaultConfiguration.path` is optional. I also need to do some work to make the default configuration be used in Swift when initializing without any arguments.

segiddins:
I've now updated all of the examples to use undeprecated APIs.

jpsim:
CI is failing because `public func migrateRealm(configuration: RealmConfiguration = default) --> NSError?` isn't documented.

segiddins:
OK, I think this is ready for a set of eyes, @jpsim @jpsim @bdash .

Realm/RLMConfiguration.h
```diff
@@ -0,0 +1,65 @@
...
+ @return The default Realm configuration.
+ */
++ (instancetype)defaultConfiguration;
+
+/**
+ Sets the default configuration to the given `RLMConfiguration`.
+
+ @param configuration The new default Realm configuration.
+ */
++ (void)setDefaultConfiguration:(nullable RLMConfiguration *)configuration;
```
bdash: Is there a good reason for the argument to be nullable here?

segiddins: setting it to `nil` resets it

Realm/RLMRealm.h
```diff
@@ -59,12 +59,27 @@ RLM_ASSUME_NONNULL_BEGIN
  default Realm is persisted as default.realm under the Documents directory of
  your Application on iOS, and in your application's Application Support
  directory on OS X.
+ 
+ The default Realm is created using the default `RLMConfiguration`, which
+ can be changed via `-[RLMConfiguration setDefaultConfiguration]`.
```
bdash: `+[RLMConfiguration setDefaultConfiguration:]`.

Realm/RLMConfiguration.h
bdash: Why is that preferable to having users pass `[[RLMConfiguration alloc] init]` as the argument, or providing a separate `-resetDefaultConfiguration` method.

segiddins: just felt simpler to me, not necessarily better

bdash: I'm not sure that it's worth supporting a nil argument here. However, if we do choose to we should document what it does.

RealmSwift-swift1.2/RealmConfiguration.swift
```diff
@@ -0,0 +1,156 @@
...
+
+import Foundation
+import Realm
+import Realm.Private
+
+/**
+A `RealmConfiguration` is used to describe the different options used to
+create a `Realm` instance.
+*/
+public struct RealmConfiguration {
```
bdash: Why does the Swift version have a `Realm` prefix on the type name while the Objective-C version doesn't?

examples/ios/swift-2.0/Migration/AppDelegate.swift
```diff
@@ -109,8 +109,9 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
         if let v1Path = bundlePath("default-v1.realm"), v2Path = bundlePath("default-v2.realm") {
             let realmv1Path = defaultParentPath.stringByAppendingPathComponent("default-v1.realm")
             let realmv2Path = defaultParentPath.stringByAppendingPathComponent("default-v2.realm")
-            setSchemaVersion(3, realmPath: realmv1Path, migrationBlock: migrationBlock)
-            setSchemaVersion(3, realmPath: realmv2Path, migrationBlock: migrationBlock)
+
+            let realmv1Configuration = RealmConfiguration(path: realmv1Path, schemaVersion: 3, migrationBlock: migrationBlock)
+            let realmv2Configuration = RealmConfiguration(path: realmv1Path, schemaVersion: 3, migrationBlock: migrationBlock)
```
bdash: `realmv2Configuration` should be using `realmv2Path`.

bdash: The Swift 1.2 version has the same mistake.

RealmSwift-swift1.2/RealmConfiguration.swift
segiddins: Because the ObjC version has the `RLM` prefix, and just calling it `Configuration` felt weird

Realm/RLMConfiguration.h
segiddins: Done

segiddins:
I believe I've now addressed all of the outstanding comments?

RealmSwift-swift1.2/RealmConfiguration.swift
bdash: No other types in the Swift API are prefixed in this manner, despite several having very generic names (`Object`, `List`, `Property`).

segiddins: So are you suggesting it just be called `Configuration`? I think that's way too generic, personally.

bdash: Why is that too generic but `Object` and `List` aren't?

segiddins: How about a nested struct? `Realm.Configuration` ?

segiddins:
Would like to get this merged soon, since I've started work on adding some features based on this PR

RealmSwift-swift1.2/RealmConfiguration.swift
bdash: You haven't elaborated on why `Configuration` is too generic in the Swift API only, nor on it is special in that regard. Is `Configuration` too generic because it's possible we'll want to make some other object in RealmSwift configurable in the future? If so, why wouldn't we also give the Objective-C class a name that indicates that it is a configuration for a Realm (e.g., `RLMRealmConfiguration`). Is there something I'm missing that makes the need to qualify the typename specific to the Swift API? My impression is that either both should have the extra `Realm` qualification, or neither should.

segiddins: It's now nested in the `Realm` namespace, and I think I prefer it this way.

bdash: That still doesn't address the inconsistency between Swift and Objective-C…

segiddins:
@bdash could you please remind me what you see as the inconsistency?

bdash:
This is still relevant after the move from `RealmConfiguration` to `Realm.Configuration`:

>You haven't elaborated on why `Configuration` is too generic in the Swift API only, nor on why it is special in that regard. Is `Configuration` too generic because it's possible we'll want to make some other object in RealmSwift configurable in the future? If so, why wouldn't we also give the Objective-C class a name that indicates that it is a configuration for a Realm (e.g., `RLMRealmConfiguration`). Is there something I'm missing that makes the need to qualify the typename specific to the Swift API? My impression is that either both should have the extra Realm qualification, or neither should.

segiddins:
`Realm.Configuration` is not nearly as generic as just `Configuration` -- the type shows that it's something that configures Realm (as a framework). The `RLM` prefix also signifies this.

bdash:
Being a generic term doesn't seem like a strong argument for special treatment given that we declare types with names like `Object` and `List`. Why is `Configuration` different from those other types?

Your description of the `Realm` prefix sounds like it's duplicating the namespacing that's already provided by Swift's modules.

segiddins:
I personally dislike that we have things called `Object` and `List`, but at least they're incredibly common parts of our API that will basically be used wherever Realm is used in a user's app. The configuration will probably only be created once, and it is an ancillary, supporting class, not something crucial to using Realm.

tgoyne:
I suggested `Realm.Configuration` on the grounds that it exists only to be passed to `Realm`'s initializer (and setDefaultRealm), and so it feels natural to reduce its scope as much as possible and nest it within `Realm`. We can't do the same thing in obj-c, but I wouldn't be opposed to `RLMRealmConfiguration`.

segiddins:
Renamed.

bdash:
:+1:

segiddins:
CHANGELOG added

segiddins:
Test failures all look unrelated (`open() failed` or `testNotificationForChangesWhileSuspended`)

PR #2335, opened 2015-08-03
Fix crash when adding a property to a model without updating the schema version

### LINKED ISSUE #2320: Realm 0.94.0 EXC_BAD_ACCESS at launch until resetting Realm
Hi,

I'm using Realm in my iOS and OS X application (code shared) with Xcode 6.4, and everything was working well until the last update (0.94.0), since I updated, I'm getting a crash at launch, not everytime, but once I'm getting it, I won't be able to launch again the application until I reset Realm (by removing the application for example).

This is where I'm getting the crash:

screen shot 2015-07-30 at 19 51 12

The call stack: 
screen shot 2015-07-30 at 19 51 23

and the function calling Realm:

```
- (RLMResults *)loadCatalog {
	return [KFSong allObjects];
}
```

jpsim:
Fixes #2320. /cc @segiddins @tgoyne

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,8 @@ x.x.x Release notes (yyyy-MM-dd)
 
 ### Bugfixes
 
-* None.
+* Fix crash when adding a property to a model without updating the schema
+  version.
 
 0.94.0 Release notes (2015-07-29)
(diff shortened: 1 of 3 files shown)


segiddins:
:+1:

PR #2344, opened 2015-08-05
Remove generic specialization.

hadibadjian:
Testing on v0.94.0, defining an entity relationship does not require generic specialization <ObjectType*>.

Defining a property with generic specialization would cause a compiler error.

Tested on:
Xcode 6.4
iOS Deployment target 7.0
iOS Base SDK 8.4
Mac OS X Yosemite 10.10.4

Online [Objective-C documentation](https://realm.io/docs/objc/latest/#models) may need to be updated as well.

### DIFF
--- a/Realm/RLMArray.h
+++ b/Realm/RLMArray.h
@@ -38,7 +38,7 @@ RLM_ASSUME_NONNULL_BEGIN
 
      RLM_ARRAY_TYPE(ObjectType)
      ...
-     @property RLMArray<ObjectType *><ObjectType> *arrayOfObjectTypes;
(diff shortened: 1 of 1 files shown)


jpsim:
Hi @hadibadjian, thanks for the PR. You're right that providing a generic specialization is not required, but it is *supported* when building with Xcode 7 or later. We elaborate a bit more on how (and where) this works in the [Supported Types section](https://realm.io/docs/objc/latest/#supported-types) of our Objective-C documentation that you've linked to.

In light of this, I'll be closing this PR.

PR #2350, opened 2015-08-06
[WIP] Consolidate to a single Xcode project

segiddins:
Successor to https://github.com/realm/realm-cocoa/pull/2323.
~~Blocked by https://github.com/Carthage/Carthage/pull/622.~~


segiddins:
I'd love to get some eyes on this, so we can merge it if it passes, @bdash @tgoyne.

tgoyne:
Running tests on device from Xcode doesn't work.

bdash:
As we talked about on Slack, it looks like d178c85 will prevent the tests from running on devices from within Xcode. It also contains changes to make TestHost build on OS X despite never doing so.

bdash:
The Realm.framework target in the Xcode project appears to contain some configuration settings that are redundant given those in Realm.xcconfig, and some others that could be moved to the .xcconfig.

segiddins:
Closing because, sadly, I can't make this work with either iOS 7 or the device tests.

PR #2355, opened 2015-08-07
Remove uses of RowIndexes::Sorter

tgoyne:
It's not actually part of the public API and using it directly is no longer useful anyway.

@segiddins

### DIFF
--- a/Realm/RLMQueryUtil.hpp
+++ b/Realm/RLMQueryUtil.hpp
@@ -38,10 +38,10 @@ void RLMUpdateQueryWithPredicate(realm::Query *query, NSPredicate *predicate, RL
                                  RLMObjectSchema *objectSchema);
 
 // return column index - throw for invalid column name
-NSUInteger RLMValidatedColumnIndex(RLMObjectSchema *schema, NSString *columnName);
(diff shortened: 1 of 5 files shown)


Realm/RLMQueryUtil.hpp
```diff
@@ -38,10 +40,8 @@ void RLMUpdateQueryWithPredicate(realm::Query *query, NSPredicate *predicate, RL
 // return column index - throw for invalid column name
 NSUInteger RLMValidatedColumnIndex(RLMObjectSchema *schema, NSString *columnName);
 
-// populate columns and order with the values in properties and ascending, with validation
-void RLMGetColumnIndices(RLMObjectSchema *schema, NSArray *properties,
-                         std::vector<size_t> &columns, std::vector<bool> &order);
-
+// validate the array of RLMSortDescriptors and convert it to an RLMSortOrder
+RLMSortOrder RLMSortOrderFromDescriptors(RLMObjectSchema *schema, NSArray *descriptors);
```
segiddins: `objectSchema` ?

segiddins:
I think you can maybe delete a few includes of the header that defines `Sorter`, but otherwise, :+1:

PR #2356, opened 2015-08-07
[RLMResults] Allow checking for the index of standalone objects

### LINKED ISSUE #2354: `indexOf` is throwing "Object type does not match RLMResults" exception
I've written a persistence layer around RealmSwift.  Everything works fine except I can't seem to use `indexOf`.  When I do, it throws: `Object type does not match RLMResults`.  However, even when I look in the debugger, it shows I'm passing in the correct object type to `indexOf`.  Further details:

In order to "wrap" Realm, I had to create a translation layer to translate between the Realm objects and the "Model" objects I use in the rest of the app.

Example: 

    public protocol Translator {
      func translate(from: Object) -> Model
      func translate(from: Model) -> Object
    }

I created a wrapper around `Results`, I called `Dataset`.  The `Dataset` uses a `Translator` to translate between the objects returned by `Results` and the `Model` objects the rest of the app uses.

Because of the limitations of Swift, I could not directly wrap `Results<T>` (Swift generic types are invariant).  So instead, I _extended_ `Results` with a protocol to provide the functions I needed.  

    protocol RealmResults {
      func objectAt<T: Object>(index: Int) -> T
      var objectCount: Int { get }
      func indexOfObject<T: Object>(object: T) -> Int?
      func filterOn(predicate: NSPredicate) -> RealmResults
      func sortOn(sorts: [SortDescriptor]) -> RealmResults
    }

    extension Results : RealmResults {
      func objectAt<T : Object>(index: Int) -> T {
        return self[index] as! T
      }
      
      var objectCount: Int {
        return count
      }
      
      func indexOfObject<U : Object>(object: U) -> Int? {
        if let realmObject = object as? T {
          return indexOf(realmObject)
        }
        return nil
      }
      
      func filterOn(predicate: NSPredicate) -> RealmResults {
        return filter(predicate)
      }
      
      func sortOn(sorts: [SortDescriptor]) -> RealmResults {
        return sorted(sorts)
      }
    }

So `Dataset` wraps `RealmResults` and accesses `Results<T>` using the protocol.  This all works fine, except for `indexOfObject`.  It throws `Object type does not match RLMResults` even though I'm passing in the exact object that `Results` is expecting.  
screen shot 2015-08-07 at 1 08 34 pm

You can see from the debugger that Results is: `Results<MessageObject>` and the type I'm passing in is `MessageObject`.  Any idea why this may be happening?  Is it an issue or am I missing something?

segiddins:
Presumably closes #2354.

\c @tgoyne

### DIFF
--- a/Realm/RLMArray.mm
+++ b/Realm/RLMArray.mm
@@ -103,7 +103,8 @@
 
 static void RLMValidateMatchingObjectType(RLMArray *array, RLMObject *object) {
     if (!object || ![array->_objectClassName isEqualToString:object->_objectSchema.className]) {
-        @throw RLMException(@"Object type does not match RLMArray");
+        NSString *message = [NSString stringWithFormat:@"Object type '%@' does not match RLMArray type '%@'.", object->_objectSchema.className, array->_objectClassName];
+        @throw RLMException(message);
     }
 }
 
--- a/Realm/RLMResults.mm
+++ b/Realm/RLMResults.mm
@@ -246,6 +246,9 @@ static RowIndexes::Sorter RLMSorterFromDescriptors(RLMObjectSchema *schema, NSAr
     if (object.invalidated) {
         @throw RLMException(@"RLMObject is no longer valid");
     }
+    if (!object->_row) {
+        return NSNotFound;
+    }
 
     // check that object types align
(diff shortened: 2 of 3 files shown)


tgoyne:
:+1:

PR #2366, opened 2015-08-10
Fix bug casting T.Type in Realm.swift

isaacroldan:
Hello, i've been experiencing some issues using `objectForPrimaryKey` in swift.

It usually works, but under some circumstances it won't work saying that `"Object type 'Object' not persisted in Realm"`
This was caused by this casting `(T.self as Object.Type).className()`, although i was passing a valid T.Type to the func, it was taking `T.self` as `Object`.

I changed the casting by calling directly `type.className()`.
You had a FIXME comment saying to change it to `T.className()`, but that was failing for me also.

Let me know if this is a good solution for you! it definitely fix my issues :)

### DIFF
--- a/RealmSwift-swift2.0/Realm.swift
+++ b/RealmSwift-swift2.0/Realm.swift
@@ -279,8 +279,7 @@ public final class Realm {
     - returns: The created object.
     */
     public func create<T: Object>(type: T.Type, value: AnyObject = [:], update: Bool = false) -> T {
-        // FIXME: use T.className()
-        let className = (T.self as Object.Type).className()
+        let className = type.className()
         if update && schema[className]?.primaryKeyProperty == nil {
           throwRealmException("'\(className)' does not have a primary key and can not be updated")
         }
@@ -351,8 +350,7 @@ public final class Realm {
     - returns: All objects of the given type in Realm.
     */
     public func objects<T: Object>(type: T.Type) -> Results<T> {
-        // FIXME: use T.className()
-        return Results<T>(RLMGetObjects(rlmRealm, (T.self as Object.Type).className(), nil))
+        return Results<T>(RLMGetObjects(rlmRealm, type.className(), nil))
     }
 
     /**
@@ -370,8 +368,7 @@ public final class Realm {
     - returns: An object of type `type` or `nil` if an object with the given primary key does not exist.
     */
     public func objectForPrimaryKey<T: Object>(type: T.Type, key: AnyObject) -> T? {
-        // FIXME: use T.className()
-        return unsafeBitCast(RLMGetObject(rlmRealm, (T.self as Object.Type).className(), key), Optional<T>.self)
+        return unsafeBitCast(RLMGetObject(rlmRealm, type.className(), key), Optional<T>.self)
     }
 
 


segiddins:
@isaacroldan thanks for the PR, but it looks to me like it's causing compilation to fail:

```
/Users/realm/workspace/objc_pr/configuration/Debug/swift_version/2.0/target/osx-swift/tightdb_objc/RealmSwift/Realm.swift:282:25: error: ambiguous use of 'className'
        let className = type.className()
                        ^
Realm.RLMObjectBase:5:20: note: found this candidate
  @objc class func className() -> String
                   ^
Foundation.NSObject:5:20: note: found this candidate
  @objc class func className() -> String
                   ^
/Users/realm/workspace/objc_pr/configuration/Debug/swift_version/2.0/target/osx-swift/tightdb_objc/RealmSwift/Realm.swift:353:51: error: ambiguous use of 'className'
        return Results<T>(RLMGetObjects(rlmRealm, type.className(), nil))
                                                  ^
Realm.RLMObjectBase:5:20: note: found this candidate
  @objc class func className() -> String
                   ^
Foundation.NSObject:5:20: note: found this candidate
  @objc class func className() -> String
                   ^
/Users/realm/workspace/objc_pr/configuration/Debug/swift_version/2.0/target/osx-swift/tightdb_objc/RealmSwift/Realm.swift:371:53: error: ambiguous use of 'className'
        return unsafeBitCast(RLMGetObject(rlmRealm, type.className(), key), Optional<T>.self)
                                                    ^
Realm.RLMObjectBase:5:20: note: found this candidate
  @objc class func className() -> String
                   ^
Foundation.NSObject:5:20: note: found this candidate
  @objc class func className() -> String
                   ^


Testing failed:
	Ambiguous use of 'className'
	Ambiguous use of 'className'
	Ambiguous use of 'className'
** TEST FAILED **
```

Could you please share a case in which using `objectForPrimaryKey` isn't working for you ? Thanks!

isaacroldan:
weird, it compiled on my machine and passed the tests.
I'll try to post a reproducible case ;)

segiddins:
Thanks!

isaacroldan:
Ok, found an easy to reproduce case, it's related with optionals:

```objc
//basic object for demo
class Task: Object {
    dynamic var id = 0   
        return "id"
    }
}

//dumb function, given an object tries to fetch it again using it's primaryKey to see if it exists
func objectExists<T: Object>(object: T) -> Bool {
    let realm = try! Realm()
    if let _ = realm.objectForPrimaryKey(T.self, key: 100) { //hardcoded key, it´s not important
        return true
    }
    return false
}

 // Create a dummy object as explicit unwrap, this will cause the crash.
let myTask: Task! = Task() 
myTask.id = 100
if self.objectExists(myTask) {
   print("exists")
}

```

The thing is that in this case `objectForPrimaryKey`is receiving a `T` that is not a `Task` but a `Task!` and that makes that `(T.self as Object.Type).className()` returns `Object`.
If you execute the same code but defining the Task as non-optional it works.

Changing `(T.self as Object.Type).className()` for one of this two works fine for me:
- `type.className()`
- `(type as Object.Type).className()` 

Probably the second one will avoid the bugs you are having compiling. In any case, why does it compile and passes the test on my machine?, did you compile it differently? (i just executed the tests on the RealmSwift.xcodeproj)

Hope it helps!

segiddins:
Just out of curiosity, which version of Swift are you using? 1.2 or 2.0?

isaacroldan:
2.0 beta 4 (tested with beta5 and same result)

segiddins:
Hmmm, so looking at this further, and it just seems to be a limitation with Swift's type system, rather than anything realm-specific.

isaacroldan:
Agree, but maybe Realm could workaround that issue.
I've done a simple test in a Playground to demonstrate that this is a swift types issue

```swift
class Task: NSObject { }

let task1: Task = Task()
let task2: Task! = Task()

func test<T: NSObject>(object: T) {
    print(object.dynamicType)
    print(T.self as NSObject.Type)
    print(T.self)
}

test(task1)
//will print Task, Task, Task

test(task2) 
//will print Task, NSObject, NSObject
```

This means that accessing the objects `dynamicType` will always return a valid type while accessing the generic `T.self` doesn't guarantee to get a valid type.

If Realm uses the `type` passed in the function instead of accessing the generic to get it as it is doing right now, then the user could workaround this issue because the result will depend on how the user passes the type to the function.
But if Realm always uses the generic to access the type, the user won't be able to find a workaround to implicit unwraps, independently of how he passes the type to the function.

In any case, i assume this is not a very common use case, i've found this doing some tests for a Realm extension we've working on (and that we plan to opensource soon!) and were using Implicit Unwraps in the test cases.

segiddins:
@isaacroldan I think https://github.com/realm/realm-cocoa/issues/2374 should do the trick!

isaacroldan:
Yep, that should do it!

PR #2370, opened 2015-08-10
Handle leading blank lines in the core version check

tgoyne:
Newer versions of pandoc insert blank lines before the header, so only checking the first line results in always redownloading the core.

@segiddins


segiddins:
:+1:

PR #2374, opened 2015-08-10
[Realm] Support using IUOs in methods that use generics to get the cl…

segiddins:
…ass name

Closes #2366.

\c @tgoyne

### DIFF
--- a/RealmSwift-swift2.0/Realm.swift
+++ b/RealmSwift-swift2.0/Realm.swift
@@ -280,7 +280,7 @@ public final class Realm {
     */
     public func create<T: Object>(type: T.Type, value: AnyObject = [:], update: Bool = false) -> T {
         // FIXME: use T.className()
-        let className = (T.self as Object.Type).className()
+        let className = (type as Object.Type).className()
         if update && schema[className]?.primaryKeyProperty == nil {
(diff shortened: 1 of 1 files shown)


RealmSwift-swift2.0/Realm.swift
```diff
@@ -279,8 +279,7 @@ public final class Realm {
     - returns: The created object.
     */
     public func create<T: Object>(type: T.Type, value: AnyObject = [:], update: Bool = false) -> T {
-        // FIXME: use T.className()
-        let className = (T.self as Object.Type).className()
+        let className = (type as Object.Type).className()
```
jpsim: Is the `Object.Type` cast still necessary? You removed the fixme, but not the cast...

segiddins: yeah, the cast is still necessary. I don't see why we need the FIXME any more, as this fixes the issue with using the cast

jpsim: The fixme was there because I honestly don't understand why `type.className()` doesn't yield the same result as `(type as Object.Type).className()`.

segiddins: re-added the FIXME

tgoyne:
:+1:

PR #2378, opened 2015-08-11
Switch versions so that error makes sense

porsager:
I think the versions to be inserted in the error must have been mixed at some point.


segiddins:
@alazier could you please approve this, since it's in the object store code?

segiddins:
:+1:

PR #2381, opened 2015-08-11
Add verify-watchos CI target

tgoyne:
Only tests that it builds, since you currently can't run tests for watchOS targets. Should currently fail to build due to #2376.


tgoyne:
Successfully ran the new job and failed in the expected way.

PR #2382, opened 2015-08-11
Add a precompiled header for the Realm framework and obj-c tests

tgoyne:
Cuts compilation time of the framework plus tests from ~50 seconds to ~40 seconds on my machine. Using `REALM_EXTRA_BUILD_ARGUMENTS` to set `GCC_GENERATE_DEBUGGING_SYMBOLS=NO` cuts it further to ~30 seconds, which I plan to do for the Pull Request job only.


segiddins:
Can we just use the pch via the build script, and not in the xcpode project? that will help ensure we don't rely on it, and will stop Xcode from doing stupid things.

build.sh
```diff
@@ -69,6 +69,7 @@ environment variables:
   XCMODE: xcodebuild (default), xcpretty or xctool
   CONFIGURATION: Debug or Release (default)
   REALM_CORE_VERSION: version in x.y.z format or "current" to use local build
+  REALM_EXTRA_BUILD_ARGUMENTS: additional arguments to pass to the build tool
```
segiddins: is this only being set on CI, rather than in `build.sh` itself?

tgoyne: Yes. I was concerned that if it was always set for the `verify` targets then it'd be easier to accidentally end up with it being slipped into a releasability pipeline job, while setting things in the jenkins PR job can never effect the built products for the actual release.

segiddins: sounds good -- just hard to tell from the PR alone.

segiddins:
:+1: if the speedup is what you say

PR #2388, opened 2015-08-12
[RLMObjectSchema] Only suggest NSString? when null strings are enabled

### LINKED ISSUE #2379: Does RealmSwift 0.94.1 not support "String?" anymore?
I just updated RealmSwift (via CocoaPods) from 0.93.2 to 0.94.1 which crashes like this: 

```
Terminating app due to uncaught exception 'RLMException', reason: 'Only 'object' properties can be made optional, and property 'something' is of type 'any'.
If this is a 'String?' property, it must be declared as 'NSString?' instead.'
```
Not liking the idea to switch to "NSString?", I did that anyway just to see it. Which delivers just a slightly different error:
```
Terminating app due to uncaught exception 'RLMException', reason: 'Only 'object' properties can be made optional, and property 'something' is of type 'string'.'
```

I will probably find the solution the second after posting this, but since I saw a question on Stack too, I thought this could save time for others in that "worst case".

segiddins:
Closes #2379.


bdash:
:+1:

PR #2391, opened 2015-08-13
Fix more intermittent open() errors in tests

tgoyne:
open() failures in `commitWriteTransaction` are caused by the files getting deleted out from under the Realm, which can be caused by things sticking around on a background queue after a test finishes. This addresses some (but not all) of the places where that could occur.


tgoyne:
Seems to work. In three runs of the PR job the only thing to fail was ios-swift 2.0, which hit the "touch test bundle then don't run the tests at all" problem.

@segiddins

segiddins:
Looks excellent! :+1:

Issue #2325, opened 2015-07-31
Make public inMemory variable for Realm


bigfish24:
If the component needs to recreate a Realm instance (across threads for example), then it needs to store the `path`. However, if the Realm is an in-memory Realm, then the `lastPathComponent` is the identifier and the correct initializers for in-memory Realm must be used. The current API however, doesn't make public if a Realm is in-memory or not. The only way to handle this is for the component to have an API for in-memory Realm's so it knows specifically how to handle it.

jpsim:
The last time this was brought up, we determined that it would be preferable to expose this within the `RLMConfiguration`, which @segiddins said he'd work on today.

Let's keep this open until that work is done, but I don't think it's worth implementing outside RLMConfiguration if it's as close as we think it is.

NEVER-FADE-SOUL:
I have create an in-memory realm for a view controller, but I want to clear data when the view controller released. how can I do?

bdash:
Commenting on an issue that's been closed for 2+ years is kinda pointless. If you have a question about how to use Realm, please ask it on Stack Overflow. If you think you've found a bug or you have an enhancement request, please file a new issue on GitHub. Either way, you'll need to provide sufficient detail for people to be able to understand what you're trying to do if you want to get a helpful answer.

PR #2401, opened 2015-08-14
[RLMConfiguration] Allow specifying a set of classes

segiddins:
Closes https://github.com/realm/realm-cocoa/issues/2021.
Closes https://github.com/realm/realm-cocoa/issues/2325.

\c @jpsim @bdash @tgoyne

### DIFF
--- a/Realm/RLMAccessor.h
+++ b/Realm/RLMAccessor.h
@@ -57,5 +57,6 @@ void RLMReplaceClassNameMethod(Class accessorClass, NSString *className);
 
 // Replace sharedSchema method for the given class
 void RLMReplaceSharedSchemaMethod(Class accessorClass, RLMObjectSchema * __nullable schema);
+void RLMReplaceSharedSchemaMethodWithBlock(Class accessorClass, RLMObjectSchema *(^method)(Class));
 
 RLM_ASSUME_NONNULL_END
--- a/Realm/RLMAccessor.h
+++ b/Realm/RLMAccessor.h
@@ -56,6 +56,6 @@ void RLMDynamicSet(RLMObjectBase *obj, RLMProperty *prop, id val, RLMCreationOpt
 void RLMReplaceClassNameMethod(Class accessorClass, NSString *className);
 
 // Replace sharedSchema method for the given class
-void RLMReplaceSharedSchemaMethod(Class accessorClass,  RLMObjectSchema * __nullable schema);
+void RLMReplaceSharedSchemaMethod(Class accessorClass, RLMObjectSchema * __nullable schema);
 
 RLM_ASSUME_NONNULL_END
--- a/Realm/RLMAccessor.h
+++ b/Realm/RLMAccessor.h
@@ -56,6 +56,6 @@ void RLMDynamicSet(RLMObjectBase *obj, RLMProperty *prop, id val, RLMCreationOpt
 void RLMReplaceClassNameMethod(Class accessorClass, NSString *className);
 
 // Replace sharedSchema method for the given class
-void RLMReplaceSharedSchemaMethod(Class accessorClass, RLMObjectSchema *schema);
+void RLMReplaceSharedSchemaMethod(Class accessorClass,  RLMObjectSchema * __nullable schema);
 
 RLM_ASSUME_NONNULL_END
--- a/Realm/RLMAccessor.mm
+++ b/Realm/RLMAccessor.mm
@@ -579,7 +579,7 @@ static RLMAccessorCode accessorCodeForType(char objcTypeCode, RLMPropertyType rl
 static void RLMReplaceShouldIncludeInDefaultSchemaMethod(Class cls, bool shouldInclude) {
     Class metaClass = objc_getMetaClass(class_getName(cls));
     IMP imp = imp_implementationWithBlock(^(Class){ return shouldInclude; });
-    class_replaceMethod(metaClass, @selector(shouldIncludeInDefaultSchema), imp, "b@:");
+    class_addMethod(metaClass, @selector(shouldIncludeInDefaultSchema), imp, "b@:");
 }
 
 // implement the class method className on accessors to return the className of the
@@ -587,7 +587,7 @@ static void RLMReplaceShouldIncludeInDefaultSchemaMethod(Class cls, bool shouldI
 void RLMReplaceClassNameMethod(Class accessorClass, NSString *className) {
     Class metaClass = objc_getMetaClass(class_getName(accessorClass));
     IMP imp = imp_implementationWithBlock(^(Class){ return className; });
-    class_replaceMethod(metaClass, @selector(className), imp, "@@:");
+    class_addMethod(metaClass, @selector(className), imp, "@@:");
 }
 
 // implement the shared schema method
--- a/Realm/RLMAccessor.mm
+++ b/Realm/RLMAccessor.mm
@@ -576,19 +576,25 @@ static RLMAccessorCode accessorCodeForType(char objcTypeCode, RLMPropertyType rl
     }
 }
 
+static void RLMReplaceShouldIncludeInDefaultSchemaMethod(Class cls, bool shouldInclude) {
+    Class metaClass = objc_getMetaClass(class_getName(cls));
+    IMP imp = imp_implementationWithBlock(^(Class){ return shouldInclude; });
+    class_replaceMethod(metaClass, @selector(shouldIncludeInDefaultSchema), imp, "b@:");
+}
+
 // implement the class method className on accessors to return the className of the
 // base object
 void RLMReplaceClassNameMethod(Class accessorClass, NSString *className) {
     Class metaClass = objc_getMetaClass(class_getName(accessorClass));
-    IMP imp = imp_implementationWithBlock(^{ return className; });
-    class_addMethod(metaClass, @selector(className), imp, "@:");
+    IMP imp = imp_implementationWithBlock(^(Class){ return className; });
+    class_replaceMethod(metaClass, @selector(className), imp, "@@:");
 }
 
 // implement the shared schema method
 void RLMReplaceSharedSchemaMethod(Class accessorClass, RLMObjectSchema *schema) {
     Class metaClass = objc_getMetaClass(class_getName(accessorClass));
-    IMP imp = imp_implementationWithBlock(^{ return schema; });
-    class_replaceMethod(metaClass, @selector(sharedSchema), imp, "@:");
+    IMP imp = imp_implementationWithBlock(^(Class){ return schema; });
+    class_replaceMethod(metaClass, @selector(sharedSchema), imp, "@@:");
 }
 
 static Class RLMCreateAccessorClass(Class objectClass,
@@ -632,6 +638,7 @@ static Class RLMCreateAccessorClass(Class objectClass,
     
     // implement className for accessor to return base className
     RLMReplaceClassNameMethod(accClass, schema.className);
+    RLMReplaceShouldIncludeInDefaultSchemaMethod(accClass, false);
 
     return accClass;
 }
--- a/Realm/RLMAccessor.mm
+++ b/Realm/RLMAccessor.mm
@@ -597,6 +597,12 @@ void RLMReplaceSharedSchemaMethod(Class accessorClass, RLMObjectSchema *schema) 
     class_replaceMethod(metaClass, @selector(sharedSchema), imp, "@@:");
 }
 
+void RLMReplaceSharedSchemaMethodWithBlock(Class accessorClass, RLMObjectSchema *(^method)(Class)) {
+    Class metaClass = objc_getMetaClass(class_getName(accessorClass));
+    IMP imp = imp_implementationWithBlock(method);
+    class_replaceMethod(metaClass, @selector(sharedSchema), imp, "@@:");
+}
+
 static Class RLMCreateAccessorClass(Class objectClass,
                                     RLMObjectSchema *schema,
                                     NSString *accessorClassPrefix,
--- a/Realm/RLMObject.mm
+++ b/Realm/RLMObject.mm
@@ -40,7 +40,11 @@
 }
 
 - (instancetype)initWithValue:(id)value {
-    return [super initWithValue:value schema:RLMSchema.sharedSchema];
+    RLMSchema *schema = RLMSchema.partialSharedSchema;
+    if (![schema schemaForClassName:self.class.className]) {
+        schema = RLMSchema.sharedSchema;
+    }
+    return [super initWithValue:value schema:schema];
 }
 
 - (instancetype)initWithObject:(id)object {
@@ -181,7 +185,7 @@
 
 @implementation RLMDynamicObject
 
-+ (BOOL)shouldPersistToRealm {
++ (BOOL)shouldIncludeInDefaultSchema {
     return NO;
 }
 
--- a/Realm/RLMObject.mm
+++ b/Realm/RLMObject.mm
@@ -40,10 +40,8 @@
 }
 
 - (instancetype)initWithValue:(id)value {
+    [self.class sharedSchema]; // ensure this class' objectSchema is loaded in the partialSharedSchema
     RLMSchema *schema = RLMSchema.partialSharedSchema;
-    if (![schema schemaForClassName:self.class.className]) {
-        schema = RLMSchema.sharedSchema;
-    }
     return [super initWithValue:value schema:schema];
 }
 
--- a/Realm/RLMObjectBase.h
+++ b/Realm/RLMObjectBase.h
@@ -33,6 +33,7 @@ RLM_ASSUME_NONNULL_BEGIN
 
 + (NSString *)className;
 
+/// Returns whether the class is included in the default set of classes persisted in a Realm.
 + (BOOL)shouldIncludeInDefaultSchema;
 
 @end
--- a/Realm/RLMObjectBase.h
+++ b/Realm/RLMObjectBase.h
@@ -33,6 +33,8 @@ RLM_ASSUME_NONNULL_BEGIN
 
 + (NSString *)className;
 
++ (BOOL)shouldIncludeInDefaultSchema;
+
 @end
 
 RLM_ASSUME_NONNULL_END
--- a/Realm/RLMObjectBase.h
+++ b/Realm/RLMObjectBase.h
@@ -33,7 +33,7 @@ RLM_ASSUME_NONNULL_BEGIN
 
 + (NSString *)className;
 
-/// Returns whether the class is included in the default set of classes persisted in a Realm.
+// Returns whether the class is included in the default set of classes persisted in a Realm.
 + (BOOL)shouldIncludeInDefaultSchema;
 
 @end
--- a/Realm/RLMObjectStore.mm
+++ b/Realm/RLMObjectStore.mm
@@ -274,7 +274,13 @@ void RLMAddObjectToRealm(__unsafe_unretained RLMObjectBase *const object,
 
     // set the realm and schema
     NSString *objectClassName = object->_objectSchema.className;
-    RLMObjectSchema *schema = realm.schema[objectClassName];
+    RLMObjectSchema *schema = [realm.schema schemaForClassName:objectClassName];
+    if (!schema) {
+        NSString *message = [NSString stringWithFormat:@"Object type '%@' is not persisted in the Realm. "
+                                  @"If using a custom `objectClasses` array in your configuration, add `%@` to the list of `objectClasses`.",
+                                  objectClassName, objectClassName];
+        @throw RLMException(message);
+    }
     object->_objectSchema = schema;
     object->_realm = realm;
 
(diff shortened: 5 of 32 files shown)


RealmSwift-swift1.2/Schema.swift
```diff
@@ -44,6 +44,16 @@ public final class Schema: Printable {
 
     // MARK: Initializers
 
+    /**
+    Creates a `Schema` that contains only the given `objectTypes`.
+
+    :param: objectTypes An array of `Object` subclass types to include in the schema.
+    */
+    public init(objectTypes: [Object.Type]) {
+        self.rlmSchema = RLMSchema(objectClasses: objectTypes)
```
segiddins: this seems not to work in swift 1.2 -- `fatal error: value failed to bridge from Swift type to a Objective-C type`

Realm/RLMAccessor.mm
```diff
@@ -576,19 +576,25 @@ static RLMAccessorCode accessorCodeForType(char objcTypeCode, RLMPropertyType rl
...
+    IMP imp = imp_implementationWithBlock(^(Class){ return shouldInclude; });
+    class_replaceMethod(metaClass, @selector(shouldIncludeInDefaultSchema), imp, "b@:");
+}
+
 // implement the class method className on accessors to return the className of the
 // base object
 void RLMReplaceClassNameMethod(Class accessorClass, NSString *className) {
     Class metaClass = objc_getMetaClass(class_getName(accessorClass));
-    IMP imp = imp_implementationWithBlock(^{ return className; });
-    class_addMethod(metaClass, @selector(className), imp, "@:");
```
tgoyne: addMethod is the correct function for all of these functions. replaceMethod is significantly slower and is only needed if the class already has a definition of that method which needs to be replaced (as opposed to overriding a method from a parent class). RLMReplaceSharedSchemaMethod uses replaceMethod because opening a Realm file with a different column ordering used to modify the existing accessor classes rather than creating new ones, and it was never switched to addMethod when we fixed that.

Realm/RLMRealmConfiguration.h
```diff
@@ -60,6 +60,9 @@ RLM_ASSUME_NONNULL_BEGIN
 /// The block which migrates the Realm to the current version.
 @property (nonatomic, copy, nullable) RLMMigrationBlock migrationBlock;
 
+/// A custom schema to use for the Realm.
+@property (nonatomic, copy, nullable) RLMSchema *customSchema;
```
tgoyne: TA wants us to keep the word "schema" (and the general concept) out of our public API as much as possible, so this should probably be `NSArray<Class> *objectClasses;`. Setter should probably construct the RLMSchema object and store it in a private property as opposed to creating it when the Realm is opened to avoid having creating a bunch of Realm instances from a single config be slow.

Realm/RLMSchema.mm
```diff
@@ -39,10 +40,29 @@ @interface RLMSchema ()
...
 
 static RLMSchema *s_sharedSchema;
+static RLMSchema *s_partialSharedSchema;
 static NSMutableDictionary *s_localNameToClass;
 
 @implementation RLMSchema
 
++ (instancetype)schemaWithObjectClasses:(NSArray *)classes {
+    NSUInteger count = classes.count;
+    Class *classArray = static_cast<Class *>(malloc(count * sizeof(Class)));
```
tgoyne: This is never freed. Can just do `auto classArray = std::make_unique<Class[]>(count);`.

Realm/RLMSchema.mm
```diff
@@ -71,66 +91,92 @@ + (void)initialize {
...
+                    @throw RLMException(message);
+                }
+                continue;
+            }
+
+            s_localNameToClass[className] = cls;
+
+            RLMReplaceClassNameMethod(cls, className);
+            // override sharedSchema classs method to return nil to avoid topo-sort issues
+            RLMReplaceSharedSchemaMethod(cls, nil);
```
tgoyne: This doesn't look thread-safe. In the old code this was all run during `+load` so no one could be doing stuff with the class on another thread, but that's no longer the case.

Realm/RLMSchema.mm
```diff
@@ -71,66 +91,92 @@ + (void)initialize {
...
-    // process all RLMObject subclasses
-    for (Class cls in s_localNameToClass.allValues) {
-        RLMObjectSchema *schema = [RLMObjectSchema schemaForObjectClass:cls];
-        [schemaArray addObject:schema];
-
-        // override sharedSchema classs methods for performance
-        RLMReplaceSharedSchemaMethod(cls, schema);
+// schema based on runtime objects
++ (instancetype)sharedSchema {
+    if (!s_sharedSchema) {
```
tgoyne: This also isn't thread-safe.

Realm/RLMUtil.mm
```diff
@@ -280,7 +280,8 @@ static inline BOOL RLMIsSubclass(Class class1, Class class2) {
 }
 
 BOOL RLMIsObjectSubclass(Class klass) {
-    return RLMIsSubclass(class_getSuperclass(klass), RLMObjectBase.class);
+    static Class objectBaseClass = RLMObjectBase.class;
```
tgoyne: Is this change required for some reason? It's almost certainly slower than the old code.

Realm/RLMAccessor.mm
segiddins: Cool cool -- although `sharedSchema`, at least with these changes, needs to be `replaceMethod`

Realm/RLMRealmConfiguration.h
segiddins: I mean, `RLMSchema` is already public, so I feel that would be more confusing? But it should be easy enough to change

Realm/RLMSchema.mm
segiddins: Locked on `s_localNameToClass`

Realm/RLMSchema.mm
segiddins: I don't think it strictly needs to be, because all of the mutation work is already done inside a mutex, and I don't think we ever use pointer equality on the shared schema?

Realm/RLMUtil.mm
segiddins: no, I actually thought it would be faster? We use a static for the object base class elsewhere

tgoyne: A function-level static with a nontrivial initializer has the equivalent of a `dispatch_once()` wrapping the initialization. That's significantly cheaper than `NSClassFromString` (for the RealmSwift stuff), but probably not cheaper than a simple message send.

Realm/RLMRealmConfiguration.h
tgoyne: RLMSchema exists, but the guide never needs to mention it and the sort of user that could be confused by it will never encounter it.

Realm/RLMUtil.mm
segiddins: Ah gotcha. Might it make sense to have a global static for the object base class that's done inside `+initialize` somewhere?

Realm/RLMRealmConfiguration.h
segiddins: Sure. Not sure I agree with the decision, but I'll change it.

segiddins:
I'll try and get around to addressing the feedback on this tonight.

tgoyne:
One thing that needs to be tested (and probably fixed) is trying to do things with a table that doesn't exist in the given Realm. Reading should be fine since that's already supported for read-only Realms, but trying to add an object that isn't part of a Realm's schema needs to throw an exception.

tgoyne:
We may also want to make `shouldIncludeInDefaultSchema` semi-public. The use-case I have in mind is when writing a library which as an implementation detail uses Realm to store data. Currently this would still infect the application's schema with their classes.

segiddins:
@tgoyne that was precisely my intention with it :) It already is semi-public, in that it's in `RLMObjectBase.h`

segiddins:
Updated, ready for further review :)

Realm/RLMObjectBase.h
```diff
@@ -33,6 +33,9 @@ RLM_ASSUME_NONNULL_BEGIN
 
 + (NSString *)className;
 
+/// Returns whether the class is included in the default set of classes persisted in a Realm.
```
tgoyne: If we're treating this as an official customization point then the docs need to describe it for a user overriding it, not for someone calling it. It also needs to be documented in RLMObject and Object, not here (RLMObjectBase is an implementation detail that should not appear anywhere in the documentation).

segiddins:
@tgoyne I put it there precisely because I thought we only wanted it to be semi-public? I'm happy to document it on Object and RLMObject, though.

Realm/RLMSchema.mm
tgoyne: That doesn't help. All *uses* of the class would have to lock `s_localNameToClass`, not just things modifying it. On another thread someone could call `[ModelClass sharedSchema]` between this line and when the method is restored, and get nil.

Realm/RLMSchema.mm
tgoyne: Reads have to be synchronized, not just writes. For example, right now the assignment to `s_sharedSchema` could legally become visible before the assignment to `schema.objectSchema` because there is no memory barrier between them.

Realm/RLMSchema.mm
segiddins: No, because calling sharedSchema will cause us to hit this lock if the shared schema isn't loaded yet

Realm/RLMSchema.mm
segiddins: It sounds like the easiest thing to do is to put a `dispatch_once` here, but if I recall, you said that was avoided for reasons around exceptions?

tgoyne: std::call_once is basically the same thing as dispatch_once but supports throwing exceptions out of it.

segiddins: Done

RealmSwift-swift2.0/RealmConfiguration.swift
```diff
@@ -33,7 +33,7 @@ extension Realm {
         /// configuration is explicitly specified (i.e. `Realm()`).
         public static var defaultConfiguration: Configuration {
             get {
-                return fromRLMRealmConfiguration(RLMRealmConfiguration.defaultConfiguration())
+            return fromRLMRealmConfiguration(RLMRealmConfiguration.defaultConfiguration())
```
jpsim: indentation

RealmSwift-swift1.2/RealmConfiguration.swift
```diff
@@ -21,134 +21,162 @@ import Realm
...
+
+        /**
+        Initializes a `Configuration`, suitable for creating new `Realm` instances.
+
+        :param: path               The path to the realm file.
+        :param: inMemoryIdentifier A string used to identify a particular in-memory Realm.
+        :param: encryptionKey      64-byte key to use to encrypt the data.
+        :param: readOnly           Whether the Realm is read-only (must be true for read-only files).
+        :param: schemaVersion      The current schema version.
+        :param: migrationBlock     The block which migrates the Realm to the current version.
```
jpsim: missing `objectTypes` parameter

RealmSwift-swift2.0/RealmConfiguration.swift
```diff
@@ -59,13 +59,15 @@ extension Realm {
             encryptionKey: NSData? = nil,
             readOnly: Bool = false,
             schemaVersion: UInt64 = 0,
-            migrationBlock: MigrationBlock? = nil) {
+            migrationBlock: MigrationBlock? = nil,
+            objectTypes: [Object.Type]? = nil) {
```
jpsim: missing `objectTypes` parameter in comment doc

Realm/RLMAccessor.h
```diff
@@ -56,6 +56,6 @@ void RLMDynamicSet(RLMObjectBase *obj, RLMProperty *prop, id val, RLMCreationOpt
 void RLMReplaceClassNameMethod(Class accessorClass, NSString *className);
 
 // Replace sharedSchema method for the given class
-void RLMReplaceSharedSchemaMethod(Class accessorClass, RLMObjectSchema *schema);
+void RLMReplaceSharedSchemaMethod(Class accessorClass,  RLMObjectSchema * __nullable schema);
```
jpsim: bonus argument spacing

Realm/RLMObject.mm
```diff
@@ -40,7 +40,11 @@ - (instancetype)init {
 }
 
 - (instancetype)initWithValue:(id)value {
-    return [super initWithValue:value schema:RLMSchema.sharedSchema];
+    RLMSchema *schema = RLMSchema.partialSharedSchema;
+    if (![schema schemaForClassName:self.class.className]) {
```
jpsim: Shouldn't we be throwing an exception if the current class isn't in the partial shared schema?

Realm/RLMObjectBase.h
jpsim: I don't think this should be part of the public interface IMO. If a user wants custom schemas, let them use custom schemas. The whole point of default configurations is to avoid additional default configuration state like this.

jpsim:
Why not replace `[RLMSchema sharedSchema]` with `[[RLMConfiguration defaultConfiguration] schema]`?

segiddins:
@jpsim standalone objects

Realm/RLMObject.mm
jpsim: right, to support standalone objects, we need to use the shared schema, which leads me to my next question: why not keep this code as-is, if we're going to default to the shared schema anyway?

segiddins: Because we want to allow using standalone objects without requiring the shared schema be loaded (for extension memory purposes) when that class has already been 'registered'

Realm/RLMObjectBase.h
segiddins: @jpsim this is more for component makers, so if they use realm as an implementation detail, the end user won't have to specify class subsets.

Realm/RLMObject.mm
jpsim: But then we'd still load the entire shared schema for a single class that's not in the partial shared schema. Seems like a mechanism to cache object schemas would be preferable then.

segiddins: @jpsim they are cached? The idea is to make it _possible_ to avoid eagerly loading in every binary, because that can take up several megs

jpsim: They're only cached in the partial & full shared schemas. What I'm proposing is finer schema caching like a mutable (and private) `+[RLMSchema parsedSchemas]` so that instead of loading the entire shared schema if a single class is missing from the partial shared schema, the missing object schema is adding to the parsed schemas and *that*s passed in to `initWithValue:schema:`.

This would further minimize memory usage from loaded object schemas.

segiddins: @jpsim not entirely possible, because we need to ensure that classes referenced in link properties are loaded at the same time as the classes that reference them -- this is why classes are batch registered, rather than doing it one-by-one.

jpsim: Just checking if the current class is contained in the partial schema isn't sufficient to ensure its linked properties' classes are already included.

This is why it seems more reasonable to me to replace `[RLMSchema sharedSchema]` with `[[RLMConfiguration defaultConfiguration] schema]` and to require that all `RLMObject` classes that will be used be included in the default configuration's schema.

segiddins: @jpsim it _is_ sufficient to ensure the linked properties classes are already included, because when the classes are registered into the partial shared schema, all of their properties get parsed so the `RLMObjectSchema` can get created, and if a linked property's class were missing, an exception would be thrown.

jpsim: ok, then :+1:. thanks for explaining

Realm/RLMSchema.mm
```diff
@@ -71,66 +92,92 @@ + (void)initialize {
     }
     initialized = true;
 
-    NSMutableArray *schemaArray = [NSMutableArray array];
-    RLMSchema *schema = [[RLMSchema alloc] init];
+    s_localNameToClass = [NSMutableDictionary dictionary];
```
tgoyne: These can be initialized where they're declared and the `initialize` method can be removed entirely.

Realm/RLMSchema.mm
```diff
@@ -71,66 +92,92 @@ + (void)initialize {
...
+// schema based on runtime objects
++ (instancetype)sharedSchema {
+    static std::once_flag onceFlag;
+    std::call_once(onceFlag, [&](){
+        RLMSchema *schema = [[RLMSchema alloc] init];
 
-        // override sharedSchema classs methods for performance
-        RLMReplaceSharedSchemaMethod(cls, schema);
+        unsigned int numClasses;
+        std::unique_ptr<__unsafe_unretained Class[]> classes(objc_copyClassList(&numClasses));
```
tgoyne: This will use `delete[]` when it needs to use `free()`. `std::unique_ptr<__unsafe_unretained Class[], void (*)(void *)> classes(objc_copyClassList(&numClasses), &free);`

Realm/Tests/ObjectTests.m
```diff
@@ -683,6 +683,19 @@ - (void)testStringSizeLimits {
     [realm commitWriteTransaction];
 }
 
+- (void)testAddingObjectNotInSchemaThrows {
+    RLMRealmConfiguration *configuration = [RLMRealmConfiguration defaultConfiguration];
+    configuration.objectClasses = @[StringObject.class];
+    RLMRealm *realm = [RLMRealm realmWithConfiguration:configuration error:nil];
+
+    [realm beginWriteTransaction];
+    RLMAssertThrowsWithReasonMatching([realm addObject:[[IntObject alloc] initWithValue:@[@1]]], @"Object type 'IntObject' not persisted in Realm");
```
tgoyne: Maybe another sentence that suggests adding it to the configuration's objectClasses array?

Realm/RLMObjectBase.h
tgoyne: Either this needs to be made not a doc-comment, moved elsewhere, or RLMObjectBase.h needs to be addded to appledoc's ignore list, because RLMObjectBase should not appear in our docs.

Realm/Tests/ObjectTests.m
segiddins: But then it will get confusing when they just try and use an object property that isn't an `RLMObject` subclass (i.e. NSArray) -- the error then will have nothing to do with the `objectClasses` array.

tgoyne: That case could also use a more specific error message saying that it's not an RLMObject subclass. Can check for isKindOfClass:RLMObjectBase.class in the error case.

segiddins:
CHANGELOG added

Realm/RLMSchema.mm
```diff
@@ -64,73 +85,98 @@ - (void)setObjectSchema:(NSArray *)objectSchema {
...
-    s_sharedSchema = schema;
+        // set shared schema
+        s_sharedSchema = schema;
+    });
+    return s_sharedSchema;
 }
 
-// schema based on runtime objects
-+ (instancetype)sharedSchema {
++ (bool)isSharedSchemaInitialized {
```
tgoyne: This doesn't appear to be used anywhere.

Realm/RLMObjectStore.mm
```diff
@@ -274,7 +274,13 @@ void RLMAddObjectToRealm(__unsafe_unretained RLMObjectBase *const object,
 
     // set the realm and schema
     NSString *objectClassName = object->_objectSchema.className;
-    RLMObjectSchema *schema = realm.schema[objectClassName];
+    RLMObjectSchema *schema = [realm.schema schemaForClassName:objectClassName];
+    if (!schema) {
+        NSString *message = [NSString stringWithFormat:@"Object type '%@' is not persisted in the Realm. "
+                                  @"If using a custom `objectClasses` array in your configuration, add `%@` to the list of `objectClasses`.",
```
jpsim: this message should say `objectTypes` for Swift.

Realm/RLMObjectStore.mm
```diff
@@ -451,7 +457,13 @@ static void RLMValidateValueForObjectSchema(__unsafe_unretained id const value,
 
     // create the object
     RLMSchema *schema = realm.schema;
-    RLMObjectSchema *objectSchema = schema[className];
+    RLMObjectSchema *objectSchema = [realm.schema schemaForClassName:className];
+    if (!objectSchema) {
+        NSString *message = [NSString stringWithFormat:@"Object type '%@' is not persisted in the Realm. "
+                             @"If using a custom `objectClasses` array in your configuration, add `%@` to the list of `objectClasses`.",
```
jpsim: same

Realm/RLMSchema.mm
```diff
@@ -64,73 +85,94 @@ - (void)setObjectSchema:(NSArray *)objectSchema {
...
+                    NSString *message = [NSString stringWithFormat:@"RLMObject subclasses with the same name cannot be included twice in the same target. Please make sure '%@' is only linked once to your current target.", className];
+                    @throw RLMException(message);
+                }
+                continue;
+            }
+
+            s_localNameToClass[className] = cls;
+
+            RLMReplaceClassNameMethod(cls, className);
+            // override sharedSchema classs method to return nil to avoid topo-sort issues when on this thread
```
jpsim: classs

tgoyne:
:+1:

conradev:
This is great! Excited that the `sharedSchema` is now configurable.

PR #2409, opened 2015-08-17
Improve performance of adding standalone Swift objects

### LINKED ISSUE #2405: Performance: Swift API is a lot slower the ObjC API (?)
Hello,

I have been doing some basic comparison between Realm and SQLite, and in the meantime, I found that the Swift API is a lot slower in certain cases (query + insert?).
https://github.com/itchingpixels/realm-vs-sqlite

Am I doing something wrong? Are you guys aware of this?

Thanks a lot!

tgoyne:
Tweak how List properties are handled by KVC to avoid penalizing non-List properties, as they're the far more common case. Makes bulk-adding objects with a single int field nearly twice as fast overall.

The added test is just a sanity check to verify that we didn't actually need to be overriding the returned object for `dynamic` `List` properties, and is Swift 1.2-only because 2.0 switched back to forbidding them.

Fixes #2405.

### DIFF
--- a/Realm/RLMAccessor.mm
+++ b/Realm/RLMAccessor.mm
@@ -748,12 +748,12 @@ void RLMDynamicSet(__unsafe_unretained RLMObjectBase *const obj, __unsafe_unreta
     }
 }
 
-id RLMDynamicGet(__unsafe_unretained RLMObjectBase *obj, __unsafe_unretained NSString *propName) {
+RLMProperty *RLMValidatedGetProperty(__unsafe_unretained RLMObjectBase *const obj, __unsafe_unretained NSString *const propName) {
     RLMProperty *prop = obj->_objectSchema[propName];
     if (!prop) {
         @throw RLMException([NSString stringWithFormat:@"Invalid property name `%@` for class `%@`.", propName, obj->_objectSchema.className]);
     }
-    return RLMDynamicGet(obj, prop);
+    return prop;
 }
 
 id RLMDynamicGet(__unsafe_unretained RLMObjectBase *obj, __unsafe_unretained RLMProperty *prop) {
--- a/Realm/RLMObjectBase.mm
+++ b/Realm/RLMObjectBase.mm
@@ -20,9 +20,10 @@
 
 #import "RLMAccessor.h"
 #import "RLMArray_Private.hpp"
-#import "RLMObservation.hpp"
+#import "RLMListBase.h"
 #import "RLMObjectSchema_Private.hpp"
 #import "RLMObjectStore.h"
+#import "RLMObservation.hpp"
 #import "RLMProperty_Private.h"
 #import "RLMRealm_Private.hpp"
 #import "RLMSchema_Private.h"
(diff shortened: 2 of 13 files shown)


segiddins:
Is there any way to make the same improvements to subscripting? Also, do you have some numbers handy?

tgoyne:
With the benchmark in the linked issue I get 18 seconds to insert in obj-c, 34 seconds in Swift before this, and 17 seconds in Swift after this (there's a few seconds of variance so it's probably not actually *faster* than obj-c).

RealmSwift-swift1.2/Tests/SwiftTestObjects.swift
```diff
@@ -124,6 +124,10 @@ class SwiftListOfSwiftObject: Object {
     let array = List<SwiftObject>()
 }
 
+class SwiftDynamicListOfSwiftObject: Object {
+    dynamic let array = List<SwiftObject>()
```
segiddins: won't this need to be `dynamic var` ?

tgoyne: Why would it?

segiddins: I don't think you can have `dynamic let` properties

tgoyne: You can. Obj-c has readonly properties.

segiddins: OK cool :)

segiddins:
Seems good, but I think the object base subscripting could use the same kind of improvement.

tgoyne:
Subscript now has similar improvements. The changes outside of the subscript function are all related to making RLMAccessor.h usable from Swift, which was needed to avoid extra hashtable lookups in subscript.

Realm/RLMObjectBase.mm
```diff
@@ -329,7 +342,8 @@ id RLMObjectBaseObjectForKeyedSubscript(RLMObjectBase *object, NSString *key) {
     }
 
     if (object->_realm) {
-        return RLMDynamicGet(object, key);
+        return RLMDynamicGet(object, RLMValidatedGetProperty(object, key));
+        return nil;
```
segiddins: Dead code

Realm/RLMAccessor.mm
```diff
@@ -748,12 +748,12 @@ void RLMDynamicSet(__unsafe_unretained RLMObjectBase *const obj, __unsafe_unreta
     }
 }
 
-id RLMDynamicGet(__unsafe_unretained RLMObjectBase *obj, __unsafe_unretained NSString *propName) {
+RLMProperty *RLMValidatedGetProperty(RLMObjectBase *obj, NSString *propName) {
```
segiddins: any reason `obj` can't still be `__unsafe_unretained`?

segiddins:
:+1:

PR #2417, opened 2015-08-19
Have update_query_with_column_expression delegate to add_constraint_to_query

bdash:
/cc @tgoyne

### DIFF
--- a/Realm/RLMQueryUtil.mm
+++ b/Realm/RLMQueryUtil.mm
@@ -174,32 +174,32 @@ void add_bool_constraint_to_query(realm::Query &query, NSPredicateOperatorType o
     }
 }
 
+template <typename T>
 void add_string_constraint_to_query(realm::Query &query,
                                     NSPredicateOperatorType operatorType,
                                     NSComparisonPredicateOptions predicateOptions,
                                     Columns<String> &&column,
-                                    NSString *value) {
+                                    T value) {
     bool caseSensitive = !(predicateOptions & NSCaseInsensitivePredicateOption);
     bool diacriticInsensitive = (predicateOptions & NSDiacriticInsensitivePredicateOption);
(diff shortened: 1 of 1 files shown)


tgoyne:
To make sure I understand this correctly: this is taking advantage of the fact that `value_of_type_for_query` can return a `Columns` to remove the codepaths for comparing two columns and just use the one for comparing a column to a value? If so :+1:

Realm/RLMQueryUtil.mm
```diff
@@ -207,10 +207,18 @@ void add_string_constraint_to_query(realm::Query &query,
...
+                                    Columns<String> &&column,
+                                    id value) {
+    add_string_constraint_to_query(query, operatorType, predicateOptions, std::forward<Columns<String>>(column), RLMStringDataWithNSString(value));
+}
+
 void add_string_constraint_to_query(realm::Query& query,
                                     NSPredicateOperatorType operatorType,
                                     NSComparisonPredicateOptions predicateOptions,
-                                    NSString *value,
+                                    id value,
```
tgoyne: I find the fact that this change is needed somewhat unsettling, but looking at what breaks with it being `NSString *` I don't see an obvious solution.

bdash: It's not ideal, but it's not a big departure from what happens prior to this change. The code was already implicitly converting an `id` to `NSString *`. It now just happens at the point `RLMStringDataWithNSString` is called by `add_string_constraint_to_query`, rather than when `add_string_constraint_to_query` is called.

bdash:
> To make sure I understand this correctly: this is taking advantage of the fact that value_of_type_for_query can return a Columns to remove the codepaths for comparing two columns and just use the one for comparing a column to a value?

That's the one!

Issue #2416, opened 2015-08-19
RLMObject category init method not seen from Swift view controller


viktorasl:
Hello, I have a problem using category declared init method of RLMObject from Swift view controller. I have created a demo project (https://github.com/viktorasl/ObjcSwiftRealmModel) to demonstrate that. Mainly the steps are these:

* Swift: Model `SwiftModel: RLMObject`
* ObjC: Category `RLMObject+JSON` with `initWithJSONDictionary:` method
* Swift: `SwiftVC: UIViewController`
* `ObjcSwiftRealmModel-Bridging-Header` includes all required ObjC files

The problem: I cannot use `SwiftModel(JSONDictionary: [NSObject:AnyObject]())`

kishikawakatsumi:
I think it may be Swift compiler's bug because it reproduces without Realm like below:

```objc
@interface TestObj : NSObject
@end

@interface TestObj (test)
- (instancetype)initWithSomething:(NSObject *)obj;
- (void)test;
@end
```

```swift
class SwiftObj: TestObj {

}
```

```swift
let obj1 = TestObj() // OK
obj1.test() // OK

let obj2 = TestObj(something: nil) // OK
obj2.test() // OK

let sobj1 = SwiftObj() // OK
sobj1.test() // OK

let sobj2 = SwiftObj(something: nil) // NG
```

Further investigation with `type lookup`, it does not seem anything wrong.

```
(lldb) type lookup TestObj
@objc class TestObj : ObjectiveC.NSObject {
  @objc deinit  {
  }
  @objc init()
}
extension TestObj {
  @objc init!(something obj: ObjectiveC.NSObject!)
  @objc func test()
}
(lldb) type lookup SwiftObj
@objc class SwiftObj : __ObjC.TestObj {
  @objc deinit 
  @objc @objc init()
}
```

kishikawakatsumi:
By Further investigation, it might not be the compiler bug. It is fixed by adding `NS_DESIGNATED_INITIALIZER` to `init`, like below: 

```
@interface TestObj : NSObject
- (instancetype)init NS_DESIGNATED_INITIALIZER;
@end
```

It makes results of `type lookup` to:

```
(lldb) type lookup TestObj
@objc class TestObj : ObjectiveC.NSObject {
  @objc deinit  {
  }
  @objc init!()
}
extension TestObj {
  @objc convenience init!(something obj: ObjectiveC.NSObject!)
  @objc func test()
}
[...]

PR #2418, opened 2015-08-20
Add `NS_DESIGNATED_INITIALIZER` macro to `RLMObject`s init

kishikawakatsumi:
This fixes https://github.com/realm/realm-cocoa/issues/2416 . It makes to be able to find the custom init method that has been added in the category from Swift. I'm not sure that this change reasonable or not. I'd like to hear thoughts of others. @bdash @tgoyne @segiddins

### DIFF
--- a/Realm/RLMObject.h
+++ b/Realm/RLMObject.h
@@ -98,7 +98,7 @@ RLM_ASSUME_NONNULL_BEGIN
  
  @see [RLMRealm addObject:]:
  */
-- (instancetype)init;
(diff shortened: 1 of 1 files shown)


segiddins:
Should this also go on the other initializers as well?

tgoyne:
`initWithValue:` also needs to be marked as a designated init since it calls a super init.

tgoyne:
:+1:

segiddins:
:+1:

viktorasl:
Thanks!

PR #2420, opened 2015-08-21
[build.sh] Copy Realm.framework to OS X swift directory

segiddins:
\c @tgoyne


tgoyne:
clean_retrieve is the oddly-named function for doing that.

segiddins:
Should I use clean_retrieve, then?

jpsim:
> Should I use clean_retrieve, then?

yes. and then :+1:

segiddins:
Done -- it now uses `clean_retrieve`

PR #2429, opened 2015-08-25
[RLMRealm] Don't eagerly load the shared schema when accessing the co…

segiddins:
…figuration property

\c @tgoyne


segiddins:
Failure is unrelated

tgoyne:
:+1:

PR #2438, opened 2015-08-25
[Swift-2.0][Examples] update Swift 2 examples to do/catch instead of try!

jpsim:
In Swift 2, `try! ... removeItemAtPath` will assert if the file doesn't already exist and `try?` shows a warning in Xcode, so I'm using empty catch blocks instead. /cc @segiddins @tgoyne


segiddins:
:+1:

PR #2440, opened 2015-08-25
[README][LICENSE] added Crimea to list of prohibited regions

jpsim:
/cc @timanglade


timanglade:
lgtm :+1: 

(For interested parties: we’re adding this due to export restrictions around software that contains/calls encryption methods such as Realm.)

PR #2442, opened 2015-08-25
[RLMAccessor] Throw an exception when attempting to update persisted primary key property

segiddins:
\c @tgoyne @jpsim


tgoyne:
:+1:

jpsim:
:+1:

PR #2447, opened 2015-08-26
[Podspec] Enable WMO for RealmSwift

segiddins:
This require CocoaPods 0.38, but should fix releasability.

\c @jpsim @mrackwitz

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,12 +7,6 @@ x.x.x Release notes (yyyy-MM-dd)
 
 ### Enhancements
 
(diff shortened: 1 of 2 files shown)


tgoyne:
While this is probably a good change, I don't see how it could fix any problems other than by coincidence.

jpsim:
I think this warrants a changelog entry, even though we haven't done that in the past for installation-related changes.

segiddins:
CHANGELOG added

jpsim:
:+1:

Issue #1166, opened 2014-12-01
Support .@count and other aggregates in queries


laptobbe:
Would be great if you could filter in a query based on the count of a to many relationship. See the example below. Right now it gives the following error: Column name @count not found in table

````
[Company objectsWhere:@"ANY projects.@count > 0"];
````

ciryon:
What's the status here? I'm getting ```Property '@count' not found in object of type 'Bottle'``` when I try to do ```Realm().objects(Wine).filter("ANY bottles.@count > 0")```

tgoyne:
It's still not implemented.

kivannc:
Is there any workarounds for "ANY array.@count > 0" the query.

bryanbartow:
Not having this is killing me.  Anyone found a workaround?

RipRyness:
This is a workaround I used "ANY properties.id > 0"

rdougan:
This now appears to work in 0.96.1 thanks to the above commits by @bdash.

    realm.objects(Brand).filter("ANY items.@count == 0")

jpsim:
Yes, this is officially supported as of 0.96.0. Please see our release blogpost for details: https://realm.io/news/realm-objc-swift-0.96.0/

Gerst20051:
👍

StackHelp:
My app crashed when trying to filter with `@count`

`class PSMedia: Object {
    @objc dynamic var id = ""
    @objc dynamic var promotional_status = false
    var promotions = List<String>()
`

And here is that query that caused an issue

`realm.objects(PSMedia.self).filter("promotions.@count <= 5")`

PR #2449, opened 2015-08-26
Add support for querying based on the count of an RLMArray property.

bdash:
Also adds infrastructure to support `@min`, `@max`, `@sum` and `@avg`.

Work towards #1166.

Note that this requires a new release of core, and so will fail to build for now.

/cc @tgoyne @jpsim @segiddins

### DIFF
--- a/Realm/RLMQueryUtil.mm
+++ b/Realm/RLMQueryUtil.mm
@@ -126,13 +126,62 @@ NSString *operatorName(NSPredicateOperatorType operatorType)
     return [NSString stringWithFormat:@"unknown operator %lu", (unsigned long)operatorType];
 }
 
+class CollectionOperation {
+public:
+    enum Type {
+        Count,
+    };
+
+    CollectionOperation(Type type, RLMProperty *linkProperty, RLMProperty *property)
+        : m_type(type)
+        , m_link_property(linkProperty)
+        , m_property(property)
+    {
+        RLMPrecondition(m_link_property.type == RLMPropertyTypeArray, @"Invalid predicate", @"Collection operation can only be applied to a property of type RLMArray.");
+        switch (m_type) {
+            case Count:
+                RLMPrecondition(!m_property, @"Invalid predicate", @"Result of @count does not have any properties.");
+                break;
+        }
+    }
+
+    CollectionOperation(NSString *operationName, RLMProperty *linkProperty, RLMProperty *property = nil)
+        : CollectionOperation(type_for_name(operationName), linkProperty, property)
+    {
+    }
+
+    Type type() const { return m_type; }
+    NSUInteger linkColumnIndex() const { return m_link_property.column; }
+    NSUInteger columnIndex() const { return m_property.column; }
+
+    void validate_value(id value) const {
+        switch (m_type) {
+            case Count:
+                RLMPrecondition([value isKindOfClass:[NSNumber class]], @"Invalid operand", @"@count can only be compared with a numeric value.");
+                break;
+        }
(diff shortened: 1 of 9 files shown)


Realm/Tests/QueryTests.m
```diff
@@ -1827,4 +1827,24 @@ - (void)testQueryingOnLinkToNullableStringColumn {
...
+    ArrayPropertyObject *arr = [ArrayPropertyObject createInRealm:realm withValue:@[@"name", @[], @[]]];
+    [arr.array addObject:[StringObject createInRealm:realm withValue:@[@"value"]]];
+    [realm commitWriteTransaction];
+
+
+    XCTAssertEqual(1U, ([ArrayPropertyObject objectsWhere:@"array.@count > 0"].count));
+    XCTAssertEqual(1U, ([ArrayPropertyObject objectsWhere:@"0 < array.@count"].count));
+
+    // We do not yet handle collection operations with a keypath on the other side of the comparison.
+    XCTAssertThrows(([ArrayPropertyObject objectsWhere:@"array.@count != name"]));
```
segiddins: would be better to test on a key path that would yield the appropriate type?

jpsim:
`array.@count` is equivalent to `array[SIZE]`, so it'd be great to support that syntax as well.

bdash:
I'll revisit `array[SIZE]` later since it's represented completely differently from `@count` by `NSPredicate`. It's represented as an expression of type `NSFunctionExpressionType`, with the `SIZE` argument being represented by an undocumented `NSExpression` subclass and type.

Realm/RLMQueryUtil.mm
```diff
@@ -126,13 +126,62 @@ void set_table() override {}
...
+    }
+
+    Type type() const { return m_type; }
+    NSUInteger linkColumnIndex() const { return m_link_property.column; }
+    NSUInteger columnIndex() const { return m_property.column; }
+
+    void validate_value(id value) const {
+        switch (m_type) {
+            case Count:
+                RLMPrecondition([value isKindOfClass:[NSNumber class]], @"Invalid operand", @"@count can only be compared with a numeric value.");
```
tgoyne: Comparing the count to another column doesn't work? Not worth implementing if it requires code specifically for that, but I thought it'd just work.

bdash: Key path vs key path comparisons need more work in general since they are currently limited to single level keys. I'm hoping we'll be able to unify their handling with key path vs value comparisons since it'd be nice for them to support the same operators and to avoid writing similar code in multiple places when adding new functionality.

Realm/Tests/QueryTests.m
```diff
@@ -1827,4 +1827,24 @@ - (void)testQueryingOnLinkToNullableStringColumn {
...
+    RLMRealm *realm = [RLMRealm defaultRealm];
+    [realm beginWriteTransaction];
+
+    IntegerArrayPropertyObject *arr = [IntegerArrayPropertyObject createInRealm:realm withValue:@[ @1234, @[]]];
+    [arr.array addObject:[IntObject createInRealm:realm withValue:@[ @456 ]]];
+    [realm commitWriteTransaction];
+
+
+    XCTAssertEqual(1U, ([IntegerArrayPropertyObject objectsWhere:@"array.@count > 0"].count));
+    XCTAssertEqual(1U, ([IntegerArrayPropertyObject objectsWhere:@"0 < array.@count"].count));
```
segiddins: can you also add tests for the other comparison operators?

PR #2450, opened 2015-08-26
Fix tests with the current master version of core.

bdash:
`Table::column<T>` now checks that `T` matches the stored type of the column. This breaks with the existing implementation of `ColumnOfTypeHelper` that uses `Table::column<Int>` for `DateTime` columns. This was done to avoid compilation errors from query_expression.hpp that occurred due to `DateTime`'s `operator int64_t` being private. In core master, it's no longer private so our hack is no longer necessary. As a temporary measure we conditionally select between `DateTime` and `Int` for the column types to allow building with both the latest core release and the current core master.

/cc @tgoyne

### DIFF
--- a/Realm/RLMQueryUtil.mm
+++ b/Realm/RLMQueryUtil.mm
@@ -410,7 +410,7 @@ struct ColumnOfTypeHelper {
 // FIXME: This specialization can be removed when core v0.92.x is no longer used.
 template <typename TableGetter>
 struct ColumnOfTypeHelper<realm::DateTime, TableGetter> {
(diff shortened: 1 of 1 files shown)


tgoyne:
:+1: 

Unless I'm missing something it could be `using ResultType = std::conditional_t<std::is_convertible<DateTime, int64_t>::value, DateTime, Int>` rather than templating the function, which would IMO be a little more readable, but it doesn't really matter (especially for something that's inherently temporary).

bdash:
You're totally right. I was in the template argument mindset from earlier attempts at working around this.

bdash:
Test failure on CI is unrelated (`testNotificationsForChangesWhileSuspended`).

PR #2451, opened 2015-08-27
Allow to catch RLMError cases by ErrorType wrapper

mrackwitz:
Even though `RLMError` is accessible as Swift enumeration, it is based on integer as type and can only be based on a literal type in Objective-C. At the moment, you would need to go over `NSError` to catch  errors from Swift. This PR introduces a wrapper for improved user convenience, which allows to retrieve `ErrorType` compatible instances. They should be used exclusively for the cause of comparison in `do { … } catch Realm.Error.X { … }` statements.

#### ToDos:

* [x] ~~Adopt error handling in unit tests~~
* [x] Make use of them in unit tests
* [x] Get rid of error codes which may not happen over public API
* [x] Test all remaining error codes


jpsim:
I think you added `Error.swift` to the `Realm/` directory instead of just in `RealmSwift-swift2.0/`. Also, this would be the first time there's a file just for swift 2 and not 1.2, but the same Xcode project is used for both, so if we go with this, we'll need a stub `Error.swift` file for Swift 1.2 so Xcode can find it.

Why is Realm.Error is a struct with mutually exclusive static functions rather than an enum? Also, I don't think it's necessary to nest the Error type under Realm. Having it at the top-level would be preferable.

Realm/Error.swift
```diff
@@ -0,0 +1,115 @@
...
+        }
+
+    */
+    public struct Error : ErrorType {
+        /**
+        Implementation of hidden requirement by `ErrorType`.
+
+        - returns: the rawValue of a `RLMError` case.
+        */
+        public let _code: Int
```
segiddins: `:nodoc:` ?

mrackwitz:
>I think you added Error.swift to the Realm/ directory instead of just in RealmSwift-swift2.0/. Also, this would be the first time there's a file just for swift 2 and not 1.2, but the same Xcode project is used for both, so if we go with this, we'll need a stub Error.swift file for Swift 1.2 so Xcode can find it.

Thanks for pointing out, didn't quite catched all those implications.

>Why is Realm.Error is a struct with mutually exclusive static functions rather than an enum?

The solution is inspired by `NSCocoaError`. I don't see any advantage by using an enum. It would be nice, if we could use the defined `RLMError` codes as raw values, but Swift allows only literals.

>Also, I don't think it's necessary to nest the Error type under Realm. Having it at the top-level would be preferable.

I agree. I came up with the idea in a different namescape. :wink:

segiddins:
Seems fine to me. @mrackwitz, what makes this still a WIP?

segiddins:
:+1:

mrackwitz:
Rebased.
@jpsim: Is that generally good to go from your point of view?

RealmSwift-swift1.2/Error.swift
```diff
@@ -0,0 +1,19 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+////////////////////////////////////////////////////////////////////////////
+
+// Intentionally left empty.
```
jpsim: Clarify why

RealmSwift-swift2.0/Error.swift
```diff
@@ -0,0 +1,116 @@
...
+
+    let realm: Realm?
+    do {
+        realm = Realm()
+    } catch RealmSwift.Error.IncompatibleLockFile() {
+        print("Realm Browser app may be attached to Realm on device?")
+    }
+
+*/
+public enum Error : ErrorType {
```
jpsim: `Error: ErrorType`

RealmSwift-swift2.0/Error.swift
```diff
@@ -0,0 +1,116 @@
...
+    /**
+    Implementation of hidden requirement by `ErrorType`.
+
+    - returns: the rawValue of the underlying `rlmError`.
```
jpsim: `:nodoc:`

RealmSwift-swift2.0/Error.swift
```diff
@@ -0,0 +1,116 @@
...
+            return RLMError.FileNotFound
+        case .IncompatibleLockFile:
+            return RLMError.IncompatibleLockFile
+        }
+    }
+
+    /**
+    Implementation of hidden requirement by `ErrorType`.
+
+    - returns: `RLMErrorDomain`.
```
jpsim: `:nodoc:`

RealmSwift-swift2.0/Tests/RealmTests.swift
```diff
@@ -85,10 +85,15 @@ class RealmTests: TestCase {
...
             attributes: nil)
+
+        let expectedError = Error.FileAccessError
         do {
             _ = try Realm(path: Realm.defaultPath, readOnly: false)
             XCTFail("Realm creation should have failed")
+        } catch expectedError {
+            // Success!
         } catch {
+            XCTFail("Expected to catch <\(expectedError)>, but catched instead <\(error)>.")
```
jpsim: `but instead caught...`

bmunkholm:
Can the WIP be removed now @mrackwitz?

jpsim:
@bmunkholm no, there are a number of feedback items that remain to be addressed.

mrackwitz:
Should be ready now, would appreciate a quick look at the added changelog entry.

RealmSwift-swift2.0/Error.swift
```diff
@@ -0,0 +1,108 @@
...
+public enum Error: ErrorType {
+    /// :nodoc:
+    public var _code: Int {
+        return rlmError.rawValue
+    }
+
+    /**
+    - returns: the RLMError value, which can be used to derive the error's code.
+    */
+    internal var rlmError: RLMError {
```
jpsim: this can be `private`?

RealmSwift-swift2.0/Error.swift
```diff
@@ -0,0 +1,108 @@
...
+        }
+    }
+
+    /// :nodoc:
+    public var _domain: String {
+        return RLMErrorDomain
+    }
+
+    /**
+    - returns: error thrown by RLMRealm if no other specific error is returned when a realm is opened.
```
jpsim: Since this is Realm Swift, please change 'RLMRealm' references to 'Realm' instead.

RealmSwift-swift2.0/Error.swift
```diff
@@ -0,0 +1,108 @@
...
+    */
+    case Fail
+
+    /**
+    - returns: error thrown by RLMRealm for any I/O related exception scenarios when a realm is opened.
+    */
+    case FileAccessError
```
jpsim: None of the other cases end with 'Error', so this probably shouldn't either.

RealmSwift-swift2.0/Error.swift
```diff
@@ -0,0 +1,108 @@
...
+        return RLMErrorDomain
+    }
+
+    /**
+    - returns: error thrown by RLMRealm if no other specific error is returned when a realm is opened.
+    */
+    case Fail
+
+    /**
+    - returns: error thrown by RLMRealm for any I/O related exception scenarios when a realm is opened.
```
jpsim: For consistency with the rest of our Swift docs, this should probably just be a single-line doc (`///`) omitting the `- returns: `.

RealmSwift-swift2.0/Error.swift
```diff
@@ -0,0 +1,108 @@
...
+    case FilePermissionDenied
+
+    /**
+    - returns: error thrown by RLMRealm if no_create was specified and the file did already exist
+               when the realm is opened.
+    */
+    case FileExists
+
+    /**
+    - returns: error thrown by RLMRealm if no_create was specified and the file was not found
```
jpsim: `no_create` isn't an option in Swift, this will just confuse users.

RealmSwift-swift2.0/Error.swift
```diff
@@ -0,0 +1,108 @@
...
+    case FileAccessError
+
+    /**
+    - returns: error thrown by RLMRealm if the user does not have permission to open or create
+               the specified file in the specified access mode when the realm is opened.
+    */
+    case FilePermissionDenied
+
+    /**
+    - returns: error thrown by RLMRealm if no_create was specified and the file did already exist
```
jpsim: Same comment about `no_create` here.

RealmSwift-swift2.0/Error.swift
mrackwitz: `no_create` isn't even an option in Realm Objective-C and there doesn't seem to be any code path in both binding variants that cause that `no_create` is set, internally. So I wonder whether we really need that behavior or whether we could even remove the error codes.

`RLMErrorFileExists` can be thrown when a copy is written and the target file does already exists.

`RLMErrorFileNotFound` is not even used anywhere, while I'd expect it to be returned when the file for a read-only Realm doesn't exist.

/c @tgoyne

mrackwitz:
[The discussion here](https://github.com/realm/realm-cocoa/pull/2451#discussion_r41331736) is still pending.

RealmSwift-swift2.0/Error.swift
jpsim: If an error can never happen, then it shouldn't be exposed as part of the public API.

mrackwitz:
@jpsim @tgoyne @bdash: It would be helpful to get some quick review from your side for the five most recent commits on this branch, just to make sure that we are all on the same site on the approach for testing I've chosen here, before taking over the tests to Swift to actually test the new error enum in that way.

Realm/ObjectStore/shared_realm.cpp
```diff
@@ -84,6 +84,10 @@ Realm::Realm(Config config)
         throw RealmFileException(RealmFileException::Kind::Exists, ex.get_path(),
                                  "File at path '" + ex.get_path() + "' already exists.");
     }
+    catch (util::File::NotFound const& ex) {
+        throw RealmFileException(RealmFileException::Kind::NotFound, ex.get_path(),
+                                 "File at path '" + ex.get_path() + "' doesnot exists.");
```
jpsim: does not

Realm/RLMConstants.h
```diff
@@ -65,13 +65,13 @@ typedef NS_ENUM(NSInteger, RLMError) {
     /** Returned by RLMRealm if no other specific error is returned when a realm is opened. */
     RLMErrorFail                  = 1,
     /** Returned by RLMRealm for any I/O related exception scenarios when a realm is opened. */
-    RLMErrorFileAccessError       = 2,
+    RLMErrorFileAccess            = 2,
     /** Returned by RLMRealm if the user does not have permission to open or create
         the specified file in the specified access mode when the realm is opened. */
     RLMErrorFilePermissionDenied  = 3,
-    /** Returned by RLMRealm if no_create was specified and the file did already exist when the realm is opened. */
+    /** Returned by RLMRealm if the file did already exist when a copy should be written. */
```
jpsim: "the file already exists when a copy should be written."

Realm/RLMConstants.h
```diff
@@ -65,13 +65,13 @@ typedef NS_ENUM(NSInteger, RLMError) {
...
     RLMErrorFileExists            = 4,
-    /** Returned by RLMRealm if no_create was specified and the file was not found when the realm is opened. */
+    /** Returned by RLMRealm if file was not found when a realm is opened as read-only. */
```
jpsim: "if no file was found when a realm was opened as read-only."

CHANGELOG.md
```diff
@@ -92,6 +92,8 @@ Prebuilt frameworks are now built with Xcode 7.1.
   objects.
 * The `@count`, `@min`, `@max`, `@sum` and `@avg` collection operators are now
   supported in queries.
+* Swift: Added `Error` enum allowing to catch errors e.g. thrown on initializing
```
jpsim: this message needs to be moved up to the latest changelog segment.

PR #2456, opened 2015-08-27
Remove ancient workaround in migration tests

tgoyne:
Before we checked that the schema and accessors actually matched we had to ensure that accessor classes wouldn't get created for modified schemas in the migration tests (or other tests would end up using those classes), but that's long fixed.


jpsim:
:+1:

segiddins:
:+1:

Issue #2308, opened 2015-07-28
Improve Swift reflection code path


jpsim:
@segiddins made some improvements to this codepath in the null numbers branch, which would enable detecting `String?` as a property type.

jpsim:
This is no longer strictly necessary. Some improvements were made, and `String?` can now be detected as a property type.

Issue #2147, opened 2015-07-01
Nullable primitives in Swift


segiddins:
`dynamic var prop: Int` works, but `dynamic var prop: Int?` wont even compile
compiler complains that `Int?` is a type that can't be represented in ObjC

`dynamic var prop: RLMInt?` works (when `RLMInt` is an `@objc protocol` that
`NSNumber` conforms to) and we can use swift runtime reflection to figure out the 
proper type (protocols only seem to get type-erased when they're not being used 
as generic parameters). You can assign any 'number' type to these properties,
but you'd need to force downcast to get a value out. We cant make the protocols 
conform to their swift equivalents (i.e. `IntegerType`) because then it couldn't 
be `@objc` and thus couldn't be used as a `dynamic` property.
[This solution is the one most analogous to how I've successfully prototyped 
object-type primitives in ObjC]

using raw `dynamic var prop: NSNumber?` would work, except there's no good way to
figure out _which_ type of number to use during schema initialization. Swift has
no way to define a property as a type composed of a class and a protocol 
(ala `NSNumber<RLMInt> *`), so that approach is also out. 

If we required users to specify a non-nil default value for these properties,
we could require them to wrap those defaults in a function call
(`dynamic var prop: NSNumber? = RealmInt(5)`) that could return a subclass that 
we could use to identify the proper type, but that's gross, is easy to forget,
and doesnt allow for nil default values.

We could also use new classes, either a subclass of `NSNumber` or `NSObject`.
Either way, we'd be giving up on tagged pointer numbers, which would be a _massive_
performance loss. Either way, we could make assigning number literals work via
custom literal convertible protocols, but assigning computed values would only 
work if it were an `NSNumber` subclass. I'm not sure using swift's arithmetic 
operators would work with either or these approaches, although it's more likely
they would as an `NSNumber` subclass.

jpsim:
I feel like we're going further and further down the rabbit hole with these approaches and it may be worth exploring completely new options.

Here are two more bad options I could think of:

**Approach 1: Codegen (i.e. custom setter/getter for every property)**
```swift
import Foundation

class Object: NSObject {
    var _rlmObject = Dictionary<String, AnyObject?>()

    subscript(key: String) -> AnyObject? {
        get {
            return _rlmObject[key]!
        }
        set {
            _rlmObject[key] = newValue
        }
    }
}

class Model: Object {
[...]

PR #2458, opened 2015-08-27
Add support for nullable primitives

segiddins:
Closes https://github.com/realm/realm-cocoa/issues/2147.
Closes https://github.com/realm/realm-cocoa/issues/2308.

Requires a new core release because of https://github.com/realm/realm-core/pull/1063 and lots more test coverage, but the things I'm doing (particularly the reflection bits and the proxy optional object) are ugly enough to warrant suggestions even now.

- [x] ~~Return `RealmOptional` from `DynamicObject`~~
- [x] Test default values from `RealmOptional`
- [x] Test `NSNumber` default values
- [x] Test schema initialization with non-`RealmOptionalType` `RealmOptional`s
- [x] Test `NSNumber` schema initialization with incorrect protocols
- [x] Decide how to handle optional primitives using Realm Objective-C from Swift, if at all
- [x] Add query tests

### DIFF
--- a/Realm/RLMAccessor.mm
+++ b/Realm/RLMAccessor.mm
@@ -137,13 +137,21 @@ static inline void RLMSetValueUnique(__unsafe_unretained RLMObjectBase *const ob
 // date getter/setter
 static inline NSDate *RLMGetDate(__unsafe_unretained RLMObjectBase *const obj, NSUInteger colIndex) {
     RLMVerifyAttached(obj);
+    if (obj->_row.is_null(colIndex)) {
+        return nil;
+    }
     realm::DateTime dt = obj->_row.get_datetime(colIndex);
-    return [NSDate dateWithTimeIntervalSince1970:dt.get_datetime()];
+    return RLMDateTimeToNSDate(dt);
 }
 static inline void RLMSetValue(__unsafe_unretained RLMObjectBase *const obj, NSUInteger colIndex, __unsafe_unretained NSDate *const date) {
     RLMVerifyInWriteTransaction(obj);
-    int64_t time = date.timeIntervalSince1970;
-    obj->_row.set_datetime(colIndex, realm::DateTime(time));
+    if (date) {
+        realm::DateTime dt = RLMDateTimeForNSDate(date);
+        obj->_row.set_datetime(colIndex, dt);
+    }
+    else {
+        obj->_row.set_null(colIndex);
+    }
 }
 
 // data getter/setter
@@ -275,7 +283,7 @@ static inline id RLMGetAnyProperty(__unsafe_unretained RLMObjectBase *const obj,
         case RLMPropertyTypeBool:
             return @(mixed.get_bool());
         case RLMPropertyTypeDate:
-            return [NSDate dateWithTimeIntervalSince1970:mixed.get_datetime().get_datetime()];
+            return RLMDateTimeToNSDate(mixed.get_datetime());
         case RLMPropertyTypeData: {
             realm::BinaryData bd = mixed.get_binary();
             return RLMBinaryDataToNSData(bd);
@@ -304,7 +312,7 @@ static inline void RLMSetValue(__unsafe_unretained RLMObjectBase *const obj, NSU
         return;
     }
     if (NSDate *date = RLMDynamicCast<NSDate>(val)) {
-        obj->_row.set_mixed(col_ndx, realm::DateTime(int64_t([date timeIntervalSince1970])));
+        obj->_row.set_mixed(col_ndx, RLMDateTimeForNSDate(date));
         return;
     }
     if (NSData *data = RLMDynamicCast<NSData>(val)) {
--- a/Realm/RLMAccessor.mm
+++ b/Realm/RLMAccessor.mm
@@ -46,6 +46,11 @@ typedef NS_ENUM(char, RLMAccessorCode) {
     RLMAccessorCodeLink,
     RLMAccessorCodeArray,
     RLMAccessorCodeAny,
+
+    RLMAccessorCodeIntObject,
+    RLMAccessorCodeFloatObject,
+    RLMAccessorCodeDoubleObject,
+    RLMAccessorCodeBoolObject,
 };
 
 // long getter/setter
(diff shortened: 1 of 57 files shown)


RealmSwift-swift1.2/Optional.swift
```diff
@@ -0,0 +1,48 @@
...
+extension Int16: RealmOptionalType {}
+extension Int32: RealmOptionalType {}
+extension Int64: RealmOptionalType {}
+extension Float: RealmOptionalType {}
+extension Double: RealmOptionalType {}
+extension Bool: RealmOptionalType {}
+
+import Realm
+
+public final class RealmOptional<T: RealmOptionalType> : RLMOptionalBase {
```
jpsim: NilLiteralConvertible?

segiddins: ah knew I forgot something! Yeah, nil literal convertible _might_ be useful, but given that the type always needs to be given...

jpsim: it would definitely be useful when *using* the object, and would make the type definition nicer when it should be set to nil:

```swift
class Person: Object {
  dynamic var age: RealmOptional<Int>? = nil
}

// or when using the object:
person.age = nil
```
...not that age would ever be optional, but you get the idea.

segiddins: Not really, since `RealmOptional` is generic, it can't be dynamic -- we need it be a `let`, just like `List`, and mutations must happen through the `value` property.

as for declaring the object,
```swift
let age = RealmOptional<Int>()
```
is actually shorter than assigning the `nil`

jpsim: right, so maybe we'd be doing ourselves a disservice by implementing NilLiteralConvertible then.

segiddins: yeah, that's what I'm starting to think -- we want to make it clear that everything must be manipulated via an existing `RealmOptional`, not by creating new ones.

tgoyne:
Well that's unpleasant. Basic design and the obj-c stuff looks fine.

segiddins:
> Well that's unpleasant. 

Could you please elaborate?

Realm/RLMUtil.hpp
```diff
@@ -72,9 +74,12 @@ static inline T *RLMDynamicCast(__unsafe_unretained id obj) {
 
 template<typename T>
 static inline T *RLMNSNullToNil(T *obj) {
-    if (obj == NSNull.null) {
+    if (static_cast<id>(obj) == NSNull.null) {
         return nil;
     }
+    else if (RLMIsKindOfClass(object_getClass(obj), [RLMOptionalBase class])) {
+        return [static_cast<RLMOptionalBase *>(obj) underlyingValue];
```
bdash: This logic feels a little out of place given the function name.

Realm/RLMUtil.hpp
```diff
@@ -139,6 +144,16 @@ static inline realm::BinaryData RLMBinaryDataForNSData(__unsafe_unretained NSDat
...
 }
 
+// Date convertion utilities
+static inline NSDate *RLMDateTimeToNSDate(realm::DateTime dateTime) {
+    return [NSDate dateWithTimeIntervalSince1970:dateTime.get_datetime()];
+}
+
+static inline realm::DateTime RLMDateTimeForNSDate(__unsafe_unretained NSDate *const date) {
+    int64_t time = date.timeIntervalSince1970;
+    return realm::DateTime(time);
```
bdash: I don't think the local variable adds anything.

Realm/Tests/RLMTestObjects.h
```diff
@@ -24,6 +24,13 @@
 #define RLM_GENERIC_ARRAY(CLASS) RLMArray<CLASS>
 #endif
 
+@interface NSNumberObject : RLMObject
```
bdash: We shouldn't use the `NS` prefix on class names.

Realm/RLMUtil.hpp
segiddins: agreed, but I don't know of any other sane way to allow `RLMOptionalBase` to be able to proxy `nil` successfully

Realm/RLMUtil.hpp
segiddins: I liked it for the explicit conversion from `NSTimeInterval` to `int64_t`

Realm/Tests/RLMTestObjects.h
segiddins: Ah yeah, but this is in the tests :/

Realm/RLMUtil.hpp
bdash: A rename so the method doesn't seem specific to handling `NSNull` would address my concern.

Realm/RLMUtil.hpp
bdash: The conversion from `NSTimeInterval` to `int64_t` is still implicit. A cast would make it explicit.

Realm/RLMUtil.hpp
segiddins: this is done

jpsim: wouldn't `RLMValueOrNil()` be more appropriate?

Realm/RLMProperty.h
```diff
@@ -22,6 +22,18 @@
 
 RLM_ASSUME_NONNULL_BEGIN
 
+@protocol RLMInt
```
segiddins: where might be a more appropriate place to  put these declarations?

jpsim: `RLMDefines.h`?

segiddins: They're not really defines though...

jpsim: not `#define`s but they are protocol definitions

segiddins:
Added a checklist of TODOs in the PR comment, feel free to add more

segiddins:
All tests, minus those that depend upon https://github.com/realm/realm-core/issues/1137 getting fixed, pass.

Realm/RLMProperty.mm
```diff
@@ -162,11 +160,38 @@ -(BOOL)setTypeFromRawType {
                     @throw RLMException([NSString stringWithFormat:@"'%@' is not supported as an RLMArray object type. RLMArrays can only contain instances of RLMObject subclasses. See http://realm.io/docs/cocoa/#to-many for more information.", self.objectClassName]);
                 }
             }
+            else if (strncmp(code, numberPrefix, numberPrefixLen) == 0) {
+                // get number type from type string - @"NSNumber<objectClassName>"
+                _type = RLMPropertyTypeArray;
+                _objectClassName = [[NSString alloc] initWithBytes:code + numberPrefixLen
```
tgoyne: This is mimicking the array code in ways that doesn't make sense. Just check the suffix directly rather than creating an NSString and setting the ivars to nonsense values.

Realm/RLMProperty.mm
```diff
@@ -243,6 +268,28 @@ - (instancetype)initSwiftPropertyWithName:(NSString *)name
...
+                break;
+            case 'd':
+                _objcRawType = @"@\"NSNumber<RLMDouble>\"";
+                break;
+            case 'B':
+            case 'c':
+                _objcRawType = @"@\"NSNumber<RLMBool>\"";
+                break;
+            default:
+                @throw RLMException([NSString stringWithFormat:@"Can't persist NSNumber of type '%s', only integers, floats, doubles, and bools are currently supported.", numberType]);
```
tgoyne: comma splice (first comma should be a colon or period)

Realm/RLMProperty.mm
```diff
@@ -310,7 +357,26 @@ - (instancetype)initSwiftListPropertyWithName:(NSString *)name
...
+        return nil;
+    }
+
+    _name = name;
+    _type = propertyType;
+    _objcType = '@';
+    _swiftIvar = ivar;
+    _optional = true;
 
     // no obj-c property for generic lists, and thus no getter/setter names
```
tgoyne: comment is wrong

dimazen:
@segiddins Any updates on it? I found that implementation is almost ready except one hidden method in `RLMObjectBase`
```
+ (NSArray *)getOptionalPropertyNames:(__unused id)obj {
    return nil;
}
```
It always returns nil, however documentation already mentions optional data support :)

segiddins:
@dimazen that method is only used for Realm Swift's objects.

dimazen:
@segiddins sorry for dumb questions but:
- will you support optional attributes for ObjC 
- where can I read full description of how it'll be implemented? 
Thanks in advance!

segiddins:
@dimazen yes, it will. When this functionality is released, the release blog post will mention how this all works, and the guides will be updated to cover the new functionality.

Realm/RLMOptionalBase.h
```diff
@@ -0,0 +1,34 @@
...
+// limitations under the License.
+//
+////////////////////////////////////////////////////////////////////////////
+
+#import <Foundation/Foundation.h>
+#import <Realm/RLMConstants.h>
+
+@class RLMObjectBase, RLMProperty;
+
+@interface RLMOptionalBase : NSProxy
```
segiddins: does this need to be wrapped in a conditional for null being enabled?

Realm/RLMProperty.h
segiddins: These probably should be wrapped in a conditional for `REALM_ENABLE_NULL`

Realm/RLMUtil.mm
```diff
@@ -23,6 +23,7 @@
 #import "RLMObjectSchema_Private.hpp"
 #import "RLMObjectStore.h"
 #import "RLMObject_Private.hpp"
+#import "RLMOptionalBase.h"
```
segiddins: unecessary

Realm/Tests/ObjectTests.m
```diff
@@ -118,6 +118,16 @@ + (NSString *)primaryKey {
 }
 @end
 
+@interface PrimaryNullableIntObject : RLMObject
```
segiddins: `#if`

Realm/Tests/ObjectTests.m
```diff
@@ -1505,6 +1644,16 @@ - (void)testPrimaryKey {
     XCTAssertThrows([PrimaryInt64Object createInDefaultRealmWithValue:(@[@(1LL << 40)])], @"Duplicate primary key should throw");
     XCTAssertThrows(obj2.int64Col = 1LL << 41, @"Setting primary key should throw");
 
+    [PrimaryNullableIntObject createInDefaultRealmWithValue:@[@1]];
```
segiddins: `#ifdef REALM_ENABLE_NULL`

Realm/Tests/QueryTests.m
```diff
@@ -1825,6 +1881,10 @@ - (void)testQueryingOnLinkToNullableStringColumn {
     testWithStringClass([LinkStringObject class], [StringObject class]);
     testWithStringClass([LinkIndexedStringObject class], [IndexedStringObject class]);
 }
-#endif
+
+- (void)testSortingColumnsWithNull {
+
```
segiddins: implement

Realm/Tests/Swift1.2/SwiftObjectInterfaceTests.swift
```diff
@@ -44,6 +44,13 @@ class SwiftDefaultObject: RLMObject {
     }
 }
 
+class SwiftOptionalNumberObject: RLMObject {
```
segiddins: `#if REALM_ENABLE_NULL`

Realm/Tests/Swift2.0/SwiftObjectInterfaceTests.swift
```diff
@@ -44,6 +44,13 @@ class SwiftDefaultObject: RLMObject {
     }
 }
 
+class SwiftOptionalNumberObject: RLMObject {
```
segiddins: `#if REALM_ENABLE_NULL`

segiddins:
This is now building with an updated version of core, so I expect it to pass on CI

Realm/RLMAccessor.mm
```diff
@@ -258,6 +271,116 @@ static inline void RLMSetValue(__unsafe_unretained RLMObjectBase *const obj, NSU
...
+    size_t row;
+    if (intObject) {
+        longLongValue = intObject.longLongValue;
+        row = obj->_row.get_table()->find_first_int(colIndex, longLongValue);
+    }
+    else {
+        row = obj->_row.get_table()->find_first_null(colIndex);
+    }
+
+    if (row == obj->_row.get_index()) {
```
tgoyne: This pair of checks now appears in four places, so it's probably worth factoring out.

Realm/RLMAccessor.mm
```diff
@@ -258,6 +271,116 @@ static inline void RLMSetValue(__unsafe_unretained RLMObjectBase *const obj, NSU
...
+        return;
+    }
+    if (row != realm::not_found) {
+        NSString *reason = [NSString stringWithFormat:@"Can't set primary key property '%@' to existing value '%@'.", propName, intObject];
+        @throw RLMException(reason);
+    }
+
+    if (intObject) {
```
tgoyne: This could be a call to `RLMSetValue()`

tgoyne:
Needs KVO tests for obj-c and Swift. For a single NSNumber/RLMOptional property on `KVOObject` and corresponding checks in `testAllPropertyTypes` (and the KVC/dynamic versions) is probably sufficient since there's no KVO-relevant differences between them.

Realm/RLMProperty.mm
```diff
@@ -162,11 +160,35 @@ -(BOOL)setTypeFromRawType {
...
+                }
+                else if ([numberType isEqualToString:@"RLMDouble"]) {
+                    _type = RLMPropertyTypeDouble;
+                }
+                else if ([numberType isEqualToString:@"RLMBool"]) {
+                    _type = RLMPropertyTypeBool;
+                }
+                else {
+                    NSString *message = [NSString stringWithFormat:@"'%@' is not supported as an NSNumber object type. NSNumbers can only be RLMInt, RLMFloat, RLMDouble, and RLMBool at the moment. "
+                                                                   @"See http://realm.io/docs/cocoa/ for more information.", numberType];
```
jpsim: Link is out of date. https://realm.io/docs/objc/

jpsim:
Needs docs for `RealmOptional`. There are a few Swift classes declared `Class : Object` rather than keeping the colon on the left like the rest of our declarations.

Realm/RLMAccessor.mm
segiddins: more awkward than it's worth

Realm/RLMAccessor.mm
segiddins: does redundant `RLMVerifyInWriteTransaction`

segiddins:
retest this please

RealmSwift-swift1.2/Tests/SwiftTestObjects.swift
```diff
@@ -60,35 +60,60 @@ class SwiftObject: Object {
...
+    dynamic var optNSStringCol: NSString? = "A"
+    dynamic var optStringCol: String? = "B"
+    dynamic var optBinaryCol: NSData? = "C".dataUsingEncoding(NSUTF8StringEncoding)
+    dynamic var optDateCol: NSDate? = NSDate(timeIntervalSince1970: 10)
+    let optIntCol = RealmOptional<Int>()
+    let optFloatCol = RealmOptional<Float>()
+    let optDoubleCol = RealmOptional<Double>()
+    let optBoolCol = RealmOptional<Bool>()
+#endif
+    dynamic var optObjectCol: SwiftBoolObject? = SwiftBoolObject(value: [true])
```
tgoyne: This doesn't actually work because during schema init `SwiftOptionalDefaultValuesObject` could be initialized before `SwiftBoolObject`, resulting in `initWithValue:` failing. Might be able to make it work, since during schema init the linked objects don't actually need to be any sort of sensible state.

tgoyne:
Included in #2602.

PR #2465, opened 2015-08-28
Add SwiftLint target to RealmSwift.xcodeproj

jpsim:
Also addresses some SwiftLint style violations. I didn't make changes to Swift 1.2 files because this PR relies on https://github.com/realm/SwiftLint/pull/106, which is Swift 2.0 only.

To really make this useful, we should enable this on CI and fail if there are any violations, but at the moment we have file length, type body length and todo/fixme violations. I could uncomment the relevant lines from `.swiftlint.yml` but it seems pointless to use SwiftLint if only to ignore its output.

/cc @segiddins @tgoyne @bdash

### DIFF
--- a/.swiftlint.yml
+++ b/.swiftlint.yml
@@ -1,8 +1,8 @@
 # Realm Swift's SwiftLint configuration file
 # Uncomment the commented lines to avoid any violations
 
-excluded:
-  - RealmSwift/Tests
+included:
+  - RealmSwift
 disabled_rules:
   - force_cast
   - todo
--- a/.swiftlint.yml
+++ b/.swiftlint.yml
@@ -1,0 +1,8 @@
+excluded:
+  - RealmSwift/Tests
+disabled_rules:
+  - force_cast
+  - line_length
+  - file_length
+  - type_body_length
+  - todo
--- a/.swiftlint.yml
+++ b/.swiftlint.yml
@@ -1,8 +1,14 @@
+# Realm Swift's SwiftLint configuration file
+# Uncomment the commented lines to avoid any violations
+
 excluded:
   - RealmSwift/Tests
 disabled_rules:
   - force_cast
-  - line_length
-  - file_length
-  - type_body_length
-  - todo
+  # - todo
+line_length:
+  - 120
+# file_length:
+#   - 670
+# type_body_length:
+#   - 581
--- a/.swiftlint.yml
+++ b/.swiftlint.yml
@@ -2,7 +2,7 @@
 # Uncomment the commented lines to avoid any violations
 
 included:
-  - RealmSwift
+  - RealmSwift-swift2.0
 disabled_rules:
   - force_cast
   - todo
--- a/.swiftlint.yml
+++ b/.swiftlint.yml
@@ -5,10 +5,10 @@ excluded:
   - RealmSwift/Tests
 disabled_rules:
   - force_cast
-  # - todo
+  - todo
 line_length:
   - 120
-# file_length:
-#   - 670
-# type_body_length:
-#   - 581
+file_length:
+  - 670
+type_body_length:
+  - 581
--- a/.swiftlint.yml
+++ b/.swiftlint.yml
@@ -5,6 +5,7 @@ included:
   - RealmSwift-swift2.0
 disabled_rules:
   - force_cast
+  - force_try
   - todo
 line_length:
   - 120
--- a/.swiftlint.yml
+++ b/.swiftlint.yml
@@ -1,15 +1,10 @@
-# Realm Swift's SwiftLint configuration file
-# Uncomment the commented lines to avoid any violations
-
 included:
   - RealmSwift-swift2.0
+line_length: 120
 disabled_rules:
   - force_cast
   - force_try
   - todo
-line_length:
-  - 120
-file_length:
-  - 670
-type_body_length:
-  - 581
+  - function_body_length
+  - file_length
+  - type_body_length
--- a/RealmSwift-swift2.0/List.swift
+++ b/RealmSwift-swift2.0/List.swift
@@ -356,6 +356,7 @@ public final class List<T: Object>: ListBase {
         _rlmArray.replaceObjectAtIndex(UInt(index), withObject: unsafeBitCast(object, RLMObject.self))
     }
 
+    // swiftlint:disable variable_name
     /**
     Moves the object at the given source index to the given destination index.
 
@@ -367,6 +368,7 @@ public final class List<T: Object>: ListBase {
     - parameter to:    index to which the object at `from` should be moved.
     */
     public func move(from from: Int, to: Int) {
+        // swiftlint:enable variable_name
         throwForNegativeIndex(from)
         throwForNegativeIndex(to)
         _rlmArray.moveObjectAtIndex(UInt(from), toIndex: UInt(to))
--- a/RealmSwift-swift2.0/List.swift
+++ b/RealmSwift-swift2.0/List.swift
@@ -406,12 +406,12 @@ extension List: RealmCollectionType, RangeReplaceableCollectionType {
     */
     public func replaceRange<C: CollectionType where C.Generator.Element == T>(subRange: Range<Int>,
         with newElements: C) {
-        for _ in subRange {
-            removeAtIndex(subRange.startIndex)
-        }
-        for x in newElements.reverse() {
-            insert(x, atIndex: subRange.startIndex)
-        }
+            for _ in subRange {
+                removeAtIndex(subRange.startIndex)
+            }
+            for x in newElements.reverse() {
+                insert(x, atIndex: subRange.startIndex)
+            }
     }
 
     /// The position of the first element in a non-empty collection.
--- a/RealmSwift-swift2.0/List.swift
+++ b/RealmSwift-swift2.0/List.swift
@@ -405,13 +405,13 @@ extension List: RealmCollectionType, RangeReplaceableCollectionType {
     - parameter newElements: The new elements to be inserted into the List.
     */
     public func replaceRange<C: CollectionType where C.Generator.Element == T>(subRange: Range<Int>,
-        with newElements: C) {
-            for _ in subRange {
-                removeAtIndex(subRange.startIndex)
-            }
-            for x in newElements.reverse() {
-                insert(x, atIndex: subRange.startIndex)
-            }
+                                                                               with newElements: C) {
+        for _ in subRange {
+            removeAtIndex(subRange.startIndex)
+        }
+        for x in newElements.reverse() {
+            insert(x, atIndex: subRange.startIndex)
+        }
     }
 
     /// The position of the first element in a non-empty collection.
--- a/RealmSwift-swift2.0/List.swift
+++ b/RealmSwift-swift2.0/List.swift
@@ -356,7 +356,7 @@ public final class List<T: Object>: ListBase {
         _rlmArray.replaceObjectAtIndex(UInt(index), withObject: unsafeBitCast(object, RLMObject.self))
     }
 
-    // swiftlint:disable variable_name
+    // swiftlint:disable variable_name_min_length
     /**
     Moves the object at the given source index to the given destination index.
 
@@ -368,7 +368,7 @@ public final class List<T: Object>: ListBase {
     - parameter to:    index to which the object at `from` should be moved.
     */
     public func move(from from: Int, to: Int) {
-        // swiftlint:enable variable_name
+        // swiftlint:enable variable_name_min_length
         throwForNegativeIndex(from)
         throwForNegativeIndex(to)
         _rlmArray.moveObjectAtIndex(UInt(from), toIndex: UInt(to))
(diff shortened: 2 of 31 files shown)


RealmSwift-swift2.0/List.swift
```diff
@@ -69,7 +69,6 @@ public final class List<T: Object>: ListBase {
 
     /// Creates a `List` that holds objects of type `T`.
     public override init() {
-        // FIXME: use T.className()
         super.init(array: RLMArray(objectClassName: (T.self as Object.Type).className()))
```
bdash: Is this FIXME no longer applicable?

RealmSwift-swift2.0/Migration.swift
```diff
@@ -70,7 +70,9 @@ block automatically as needed.
 */
 @available(*, deprecated=1, message="Use Realm(configuration:error:)")
 public func setDefaultRealmSchemaVersion(schemaVersion: UInt64, migrationBlock: MigrationBlock) {
-    RLMRealmSetSchemaVersionForPath(schemaVersion, Realm.Configuration.defaultConfiguration.path, accessorMigrationBlock(migrationBlock))
+    RLMRealmSetSchemaVersionForPath(schemaVersion,
+        Realm.Configuration.defaultConfiguration.path,
+        accessorMigrationBlock(migrationBlock))
```
bdash: Is it necessary to split this over three lines? It looks like it'd fit on two.

RealmSwift-swift2.0/Realm.swift
```diff
@@ -121,7 +121,12 @@ public final class Realm {
     */
     @available(*, deprecated=1, message="Use Realm(configuration:)")
     public convenience init(path: String, readOnly: Bool, encryptionKey: NSData? = nil) throws {
-        let rlmRealm = try RLMRealm(path: path, key: encryptionKey, readOnly: readOnly, inMemory: false, dynamic: false, schema: nil)
+        let rlmRealm = try RLMRealm(path: path,
+            key: encryptionKey,
+            readOnly: readOnly,
+            inMemory: false,
+            dynamic: false,
+            schema: nil)
```
bdash: Turning one line into six doesn't feel like an improvement.

RealmSwift-swift2.0/Results.swift
```diff
@@ -132,7 +135,11 @@ public final class Results<T: Object>: ResultsBase {
     - returns: The index of the given object, or `nil` if no objects match.
     */
     public func indexOf(predicateFormat: String, _ args: AnyObject...) -> Int? {
-        return notFoundToNil(rlmResults.indexOfObjectWithPredicate(NSPredicate(format: predicateFormat, argumentArray: args)))
+        return notFoundToNil(
+            rlmResults.indexOfObjectWithPredicate(
+                NSPredicate(format: predicateFormat, argumentArray: args)
+            )
```
bdash: Why split the inner call over three lines?

RealmSwift-swift2.0/List.swift
jpsim: I had originally written this FIXME because I didn't understand why the `(T.self as Object.Type).className()` gymnastics were necessary. Now I understand it's because it's ambiguous (NSObject on OS X has className too). So yes, the FIXME is no longer necessary.

RealmSwift-swift2.0/Migration.swift
jpsim: since this PR is about style, might as well discuss it further here. my personal preference is to stay consistent within a function call or declaration: if placing one parameter on a newline, might as well do the same with all parameters.

RealmSwift-swift2.0/Results.swift
jpsim: easier to read because of all that nesting?

RealmSwift-swift2.0/List.swift
```diff
@@ -405,7 +404,8 @@ extension List: RealmCollectionType, RangeReplaceableCollectionType {
     - parameter subRange:    The range of elements to be replaced.
     - parameter newElements: The new elements to be inserted into the list.
     */
-    public func replaceRange<C: CollectionType where C.Generator.Element == T>(subRange: Range<Int>, with newElements: C) {
+    public func replaceRange<C: CollectionType where C.Generator.Element == T>(subRange: Range<Int>,
+        with newElements: C) {
```
tgoyne: The indentation makes this look like the first line of the function body.

tgoyne: This applies to a whole bunch of places.

RealmSwift-swift2.0/Results.swift
tgoyne: The dangling paren on its own line is weird and not at all consistent with any of the other code.

RealmSwift-swift2.0/List.swift
jpsim: how would you suggest we split it up?

tgoyne: The problem isn't where it's split, but rather that after splitting it's indented to the same place as the body of the function.

segiddins: have the start of arguments line up

jpsim:
first round of feedback addressed

.swiftlint.yml
```diff
@@ -0,0 +1,14 @@
+# Realm Swift's SwiftLint configuration file
+# Uncomment the commented lines to avoid any violations
+
+excluded:
+  - RealmSwift/Tests
```
segiddins: why are we excluding the tests?

jpsim:
Any further feedback? It'd be nice to turn this on in Jenkins

tgoyne:
There's still a bunch of places where function arguments look like part of the function body.

RealmSwift-swift2.0/ObjectSchema.swift
```diff
@@ -74,6 +74,6 @@ public final class ObjectSchema: CustomStringConvertible {
 extension ObjectSchema: Equatable {}
 
 /// Returns whether the two object schemas are equal.
-public func ==(lhs: ObjectSchema, rhs: ObjectSchema) -> Bool {
+public func == (lhs: ObjectSchema, rhs: ObjectSchema) -> Bool {
```
bdash: The style elsewhere appears to be to have no space between function name and opening parenthesis. Is the style intentionally different for operators, or is this being confused by a rule to have spaces around operators in other contexts?

segiddins: I believe it's intentional for operators?

jpsim: It's intentional, and based on GitHub's Swift style guide: https://github.com/github/swift-style-guide#use-whitespace-around-operator-definitions

jpsim:
> There's still a bunch of places where function arguments look like part of the function body.

I've now addressed these by indenting the bodies.

tgoyne:
That might be even worse.

jpsim:
Attempt no. 3.

RealmSwift-swift2.0/RealmCollectionType.swift
```diff
@@ -527,7 +563,10 @@ public final class AnyRealmCollection<T: Object>: RealmCollectionType {
 
     - returns: `Results` with elements sorted by the given sort descriptors.
     */
-    public func sorted<S: SequenceType where S.Generator.Element == SortDescriptor>(sortDescriptors: S) -> Results<Element> { return base.sorted(sortDescriptors) }
+    public func sorted<S: SequenceType where
+        S.Generator.Element == SortDescriptor>(sortDescriptors: S) -> Results<Element> {
```
segiddins: should be as indented as the `<`

RealmSwift-swift2.0/Results.swift
```diff
@@ -61,9 +61,12 @@ public class ResultsBase: NSObject, NSFastEnumeration {
...
     // MARK: Fast Enumeration
 
-    public func countByEnumeratingWithState(state: UnsafeMutablePointer<NSFastEnumerationState>, objects buffer: AutoreleasingUnsafeMutablePointer<AnyObject?>, count len: Int) -> Int {
-        let enumeration: NSFastEnumeration = rlmResults // FIXME: no idea why this is needed, but doesn't compile otherwise
-        return enumeration.countByEnumeratingWithState(state, objects: buffer, count: len)
+    public func countByEnumeratingWithState(state: UnsafeMutablePointer<NSFastEnumerationState>,
+                                            objects buffer: AutoreleasingUnsafeMutablePointer<AnyObject?>,
+                                            count len: Int) -> Int {
+        return Int(rlmResults.countByEnumeratingWithState(state,
+            objects: buffer,
```
segiddins: should be as indented as the `(`

jpsim:
Attempt no. 4.

segiddins:
Much better!

segiddins:
:+1: I'm in favor of all of these changes :)

jpsim:
@tgoyne @bdash I just updated this, could you please take another look?

jpsim:
Bump @tgoyne @bdash

RealmSwift-swift2.0/Tests/MigrationTests.swift
```diff
@@ -146,6 +149,7 @@ class MigrationTests: TestCase {
         }
     }
 
+    // swiftlint:disable function_body_length
```
bdash: Is there a good reason not to just disable this for the entirety of test files where this is an issue? I don't think that toggling it on and off in such code buys us anything unless we intend to fix these issues at some point.

RealmSwift-swift2.0/Tests/RealmCollectionTypeTests.swift
```diff
@@ -101,7 +101,9 @@ class RealmCollectionTypeTests: TestCase {
     }
 
     func testDescription() {
+        // swiftlint:disable line_length
```
bdash: After seeing a few of these, it makes me think there could be value in having something an annotation that applies only to the immediately following item (line, function, class, etc. depending on context), to avoid having to pair these in cases where they only apply to one item.

RealmSwift-swift2.0/Tests/MigrationTests.swift
jpsim: I think we certainly *should* address these at some point. Refactoring long tests to test more concisely is a nice goal.

RealmSwift-swift2.0/Tests/RealmCollectionTypeTests.swift
jpsim: Agreed.

.swiftlint.yml
```diff
@@ -0,0 +1,15 @@
+# Realm Swift's SwiftLint configuration file
+# Uncomment the commented lines to avoid any violations
```
bdash: I'm not sure what "Uncomment the commented lines" refers to. Is this part of a boilerplate header?

.swiftlint.yml
```diff
@@ -0,0 +1,15 @@
...
+disabled_rules:
+  - force_cast
+  - force_try
+  - todo
+line_length:
+  - 120
+file_length:
+  - 670
+type_body_length:
+  - 581
```
bdash: 670 and 581 are kinda strange numbers. Where'd they come from?

build.sh
```diff
@@ -668,6 +669,11 @@ case "$COMMAND" in
         exit 0
         ;;
 
+    "verify-swiftlint")
+        swiftlint lint --strict
+        exit 0
```
bdash: Why ignore swiftlint's exit status?

bdash:
Seems fine, I guess :+1:

tgoyne:
I dislike the idea of functions/files over some arbitrary number of lines being an error. Splitting up functions merely for the sake of splitting them up is not useful, and having to mark functions to acknowledge that they are long just to shut up the warning does not add any value.

RealmSwift-swift2.0/Tests/RealmCollectionTypeTests.swift
jpsim: Filed an issue on SwiftLint for this: https://github.com/realm/SwiftLint/issues/222. If you have any ideas on what the syntax would look like, please share.

.swiftlint.yml
jpsim: That's left over from a previous implementation. I'll remove.

build.sh
jpsim: That's a mistake. I'll address that.

jpsim: Actually it's fine, since `set -o pipefail`/`set -e` means that `exit 0` won't be reached if `swiftlint lint --strict` fails.

.swiftlint.yml
jpsim: They came from the existing file & type body lengths, so that the current sizes would pass, but exceeding that in the future would not. That reasoning was stupid, so I've now disabled those rules entirely.

jpsim:
> I dislike the idea of functions/files over some arbitrary number of lines being an error. Splitting up functions merely for the sake of splitting them up is not useful, and having to mark functions to acknowledge that they are long just to shut up the warning does not add any value.

I see your point and have disabled the type, file & function length rules.

PR #2466, opened 2015-08-29
Add KVO support for array move and exchange

tgoyne:
This were added to RLMArray after KVO was finished but before it was merged, and naturally didn't produce any merge conflicts, so they were missed.

### DIFF
--- a/Realm/RLMObservation.mm
+++ b/Realm/RLMObservation.mm
@@ -675,8 +675,30 @@ public:
         return true;
     }
 
-    // Will need to handle this once it's exposed in RLMArray
-    bool link_list_move(size_t, size_t) { return true; }
+    bool link_list_move(size_t from, size_t to) {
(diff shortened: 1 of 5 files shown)


Realm/RLMObservation.mm
```diff
@@ -675,8 +675,30 @@ bool link_list_clear(size_t old_size) {
         return true;
     }
 
-    // Will need to handle this once it's exposed in RLMArray
-    bool link_list_move(size_t, size_t) { return true; }
+    bool link_list_move(size_t from, size_t to) {
```
segiddins: what about `link_list_swap` -- is it already handled?

bdash:
:+1:

Realm/RLMObservation.mm
tgoyne: Yes, it generates two assignment replication instructions.

segiddins:
:+1:

PR #2469, opened 2015-08-31
Fix crash of GroupedTableView example in Swift 2.0

bugix:
Function arc4random returns an UInt32. If you get a value higher than Int.max, the Int(...) cast will crash.


tgoyne:
Thanks!

PR #2475, opened 2015-09-01
Update for changes in core's API

bdash:
1. `Expression::get_table` is now `const`.
2. `File::AccessError` now carries the path with it. Update how we call
   its constructor, and provide an overload of `RLMMakeError` that
   exposes the path as `NSFilePathErrorKey`.

Marking as a work in progress until we're ready to update to a new core release.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,7 +21,10 @@ x.x.x Release notes (yyyy-MM-dd)
 
 ### Bugfixes
 
-* None.
+* Fix assertion failure when inserting NSData between 8MB and 16MB in size.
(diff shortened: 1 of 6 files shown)


segiddins:
@bdash once we update, can't we get rid of some of the templating you add for changes in DateTime?

bdash:
Yes. The changes on this branch so far are only what were needed to get things building.

segiddins:
:+1: when the tests pass

bmunkholm:
@bdash: Still WIP?

tgoyne:
Included in #2602.

PR #2476, opened 2015-09-01
Add support for queries that perform comparisons on aggregates

### LINKED ISSUE #1166: Support .@count and other aggregates in queries

bdash:
This adds support for comparisons of the form `collection.@min.intCol > 10`, with support for `@min`, `@max`, `@sum` and `@avg`.

Fixes #1166.

Note that this requires a new release of core, and so will fail to build for now. The branch currently includes changes from #2449, and will until that is merged.

### DIFF
--- a/Realm/RLMProperty.mm
+++ b/Realm/RLMProperty.mm
@@ -38,6 +38,17 @@ BOOL RLMPropertyTypeIsNullable(RLMPropertyType propertyType) {
     }
 }
 
+BOOL RLMPropertyTypeIsNumeric(RLMPropertyType propertyType) {
+    switch (propertyType) {
+        case RLMPropertyTypeInt:
+        case RLMPropertyTypeFloat:
+        case RLMPropertyTypeDouble:
+            return YES;
+        default:
+            return NO;
+    }
+}
+
 @implementation RLMProperty {
     NSString *_objcRawType;
 }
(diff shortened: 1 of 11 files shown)


Realm/RLMProperty.mm
```diff
@@ -38,6 +38,17 @@ BOOL RLMPropertyTypeIsNullable(RLMPropertyType propertyType) {
     }
 }
 
+BOOL RLMPropertyTypeIsNumeric(RLMPropertyType propertyType) {
+    switch (propertyType) {
+        case RLMPropertyTypeInt:
+        case RLMPropertyTypeFloat:
+        case RLMPropertyTypeDouble:
+            return YES;
```
segiddins: iirc core supports min/max on DateTime

jpsim:
Realm Swift's List type has the aggregate functions implemented already (to conform to RealmCollectionType) by converting to a Results using `TRUEPREDICATE`. Could you update that to use this new, less jacket, approach? All the Swift tests should already be written.

jpsim:
s/jacket/hacky/

bdash:
@jpsim, I don't think this change helps with what List is doing.

jpsim:
You're right, my mistake.

bmunkholm:
@bdash: Still WIP?

tgoyne:
Included in #2602.

Issue #2119, opened 2015-06-24
testNotificationsForChangesWhileSuspended occasionally fails


jpsim:
See https://ci.realm.io/job/objc_pr/1031/configuration=Debug,target=osx/console

jpsim:
This only seems to occasionally fail on Debug OSX builds.

jpsim:
I was unable to reproduce this test failure, even when running ~40 times on the workers, even when their hosts were under heavy load.

segiddins:
@tgoyne observed this this morning on his machine -- appeared to be waiting for the process to finish when it was no longer alive

PR #2496, opened 2015-09-08
Throw when defining `Object` properties of an unsupported type

jpsim:
Fixes https://secure.helpscout.net/conversation/118264957/1804/ /cc @segiddins @tgoyne @bdash

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,8 @@ x.x.x Release notes (yyyy-MM-dd)
 * Fix poor performance when calling `-[RLMRealm deleteObjects:]` on an
   `RLMResults` which filtered the objects when there are other classes linking
   to the type of the deleted objects.
+* An exception is now thrown when defining `Object` properties of an unsupported
+  type.
 
 0.96.3 Release notes (2015-12-04)
 =============================================================
--- a/Realm/RLMProperty.mm
+++ b/Realm/RLMProperty.mm
@@ -267,12 +267,14 @@ BOOL RLMPropertyTypeIsNumeric(RLMPropertyType propertyType) {
         return nil;
     }
 
+    id propertyValue = [obj valueForKey:_name];
+
     // convert array types to objc variant
     if ([_objcRawType isEqualToString:@"@\"RLMArray\""]) {
-        _objcRawType = [NSString stringWithFormat:@"@\"RLMArray<%@>\"", [[obj valueForKey:_name] objectClassName]];
+        _objcRawType = [NSString stringWithFormat:@"@\"RLMArray<%@>\"", [propertyValue objectClassName]];
     }
     else if ([_objcRawType isEqualToString:@"@\"NSNumber\""]) {
-        const char *numberType = [[obj valueForKey:_name] objCType];
+        const char *numberType = [propertyValue objCType];
         switch (*numberType) {
             case 'i':
             case 'l':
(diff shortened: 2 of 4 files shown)


CHANGELOG.md
```diff
@@ -25,6 +25,8 @@ x.x.x Release notes (yyyy-MM-dd)
 
 * Setting the primary key property on persisted `RLMObject`s / `Object`s
   via subscripting or key-value coding will cause an exception to be thrown.
+* An exception is now thrown when defining `Object` properties of an unsupported
```
segiddins: This is a breaking change, since it completely disables mixed support

jpsim: Mixed is already not supported in Swift.

jpsim:
CI failures are unrelated (#2119 and a spurious perf test failure).

Realm/RLMProperty.mm
```diff
@@ -244,16 +244,22 @@ - (instancetype)initSwiftPropertyWithName:(NSString *)name
         _objcRawType = [NSString stringWithFormat:@"@\"RLMArray<%@>\"", [[obj valueForKey:_name] objectClassName]];
     }
 
-    if (![self setTypeFromRawType]) {
+    void (^throwForPropertyName)(NSString *) = ^(NSString *propertyName){
         NSString *reason = [NSString stringWithFormat:@"Can't persist property '%@' with incompatible type. "
-                            "Add to ignoredPropertyNames: method to ignore.", self.name];
+                            "Add to Object.ignoredProperties() class method to ignore.", propertyName];
```
tgoyne: IIRC this method is also used for Swift subclasses of RLMObject.

Realm/RLMProperty.mm
```diff
@@ -244,16 +244,22 @@ - (instancetype)initSwiftPropertyWithName:(NSString *)name
         _objcRawType = [NSString stringWithFormat:@"@\"RLMArray<%@>\"", [[obj valueForKey:_name] objectClassName]];
     }
 
-    if (![self setTypeFromRawType]) {
+    void (^throwForPropertyName)(NSString *) = ^(NSString *propertyName){
```
tgoyne: missing space before {

Realm/RLMProperty.mm
```diff
@@ -244,16 +244,22 @@ - (instancetype)initSwiftPropertyWithName:(NSString *)name
...
+
+    if (![self setTypeFromRawType]) {
+        throwForPropertyName(self.name);
     }
 
     // convert type for any swift property types (which are parsed as Any)
     if (_type == RLMPropertyTypeAny) {
         if ([[obj valueForKey:_name] isKindOfClass:[NSString class]]) {
             _type = RLMPropertyTypeString;
+        } else if (![[obj valueForKey:_name] isKindOfClass:[RLMListBase class]]) {
```
tgoyne: This check is kinda weird because it only works due to a Swift 1.2 bug and it'd break horribly if it was reachable with Swift 2.0. I think it'd be better to check `propArray` for `Any` properties in `+propertiesForClass:isSwift:` after the handling for list properties is done and there shouldn't be any Any properties left.

bmunkholm:
@jpsim: Anything more to do?

jpsim:
@bmunkholm yes, the feedback from @tgoyne still hasn't been addressed.

jpsim:
I've now updated this. Please review @tgoyne @bdash.

Realm/RLMProperty.mm
```diff
@@ -292,18 +294,23 @@ - (instancetype)initSwiftPropertyWithName:(NSString *)name
...
+                            "Add to Object.ignoredProperties() class method to ignore.", propertyName);
+    };
+
+    if (![self setTypeFromRawType]) {
+        throwForPropertyName(self.name);
     }
 
-    // convert type for any swift property types (which are parsed as Any)
-    if (_type == RLMPropertyTypeAny) {
+    if (_type == RLMPropertyTypeAny && [obj valueForKey:_name] != nil) {
```
tgoyne: Cache the results of `valueForKey:` rather than calling it a bunch of times.

Realm/RLMProperty.mm
```diff
@@ -292,18 +294,23 @@ - (instancetype)initSwiftPropertyWithName:(NSString *)name
...
         if ([[obj valueForKey:_name] isKindOfClass:[NSString class]]) {
+            // convert type for any swift property types (which are parsed as Any)
             _type = RLMPropertyTypeString;
+        } else if (![[obj valueForKey:_name] isKindOfClass:[RLMListBase class]] && ![[obj valueForKey:_name] isKindOfClass:[RLMOptionalBase class]]) {
```
tgoyne: It's weird to verify that the property is a list or optional, but leave it as Any, when a line above the type is being changed for Strings. Could use a comment as to why (I think it's because List/RealmOptional properties only get reported to obj-c with Swift 1.2 and not 2+ so we can't rely on that).

Realm/RLMProperty.mm
jpsim: wow. I totally glossed over how often this was being done.

Realm/RLMProperty.mm
jpsim: Agreed, it certainly warrants a comment. Done.

tgoyne:
:+1:

PR #2497, opened 2015-09-08
Return backlinks as RLMResults

### LINKED ISSUE #1508: Back links as `RLMResults` or `RLMArray`
If my whole project is set up to be manipulating data based on `RLMArray` and `RLMResults`, I need to be able to manipulate back linked objects from an inverse relationship as well. It seems very counter-intuitive to me to have the `linkingObjectsOfClass: forProperty:` method return an `NSArray`.

This is causing me problems in how I am able to work with the data I'm receiving from this query. Is there a way to convert this into `RLMResults`?

segiddins:
Closes #1508.

* [x] Depends upon https://github.com/realm/realm-core/pull/1084 making it into core, and then a new core release. 

I'm not sure if I've done the right things to construct the `Query` object, some feedback on that would be appreciated.

### DIFF
--- a/Realm/RLMObjectBase.mm
+++ b/Realm/RLMObjectBase.mm
@@ -30,6 +30,8 @@
 #import "RLMSwiftSupport.h"
 #import "RLMUtil.hpp"
 
+#import <realm/table_view.hpp>
+
 using namespace realm;
 
 const NSUInteger RLMDescriptionMaxDepth = 5;
@@ -304,7 +306,7 @@ RLMObjectSchema *RLMObjectBaseObjectSchema(__unsafe_unretained RLMObjectBase *ob
     return object ? object->_objectSchema : nil;
 }
 
-NSArray *RLMObjectBaseLinkingObjectsOfClass(RLMObjectBase *object, NSString *className, NSString *property) {
+RLMResults *RLMObjectBaseLinkingObjectsOfClass(RLMObjectBase *object, NSString *className, NSString *property) {
     if (!object) {
         return nil;
     }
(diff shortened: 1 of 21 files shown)


Realm/RLMObjectBase.mm
```diff
@@ -330,16 +332,14 @@ void RLMObjectBaseSetObjectSchema(__unsafe_unretained RLMObjectBase *object, __u
...
-    NSUInteger count = object->_row.get_backlink_count(*table, col);
-    NSMutableArray *links = [NSMutableArray arrayWithCapacity:count];
-    for (NSUInteger i = 0; i < count; i++) {
-        [links addObject:RLMCreateObjectAccessor(object->_realm, schema, object->_row.get_backlink(*table, col, i))];
-    }
-    return [links copy];
+    TableView tableView = object->_row.get_table()->get_backlink_view(object->_row.get_index(), table, prop.column);
+    TableView tv(tableView);
+    const Table& constTable = tv.get_parent();
+    auto query = std::make_unique<realm::Query>(constTable, &tv);
```
tgoyne: This query will only remain valid until you return from the function, as it just holds a reference to the stack-allocated `tv`. Query only takes ownership of the TableView if you use the overload that takes a unique_ptr.

There's also no real need to create the query here;  `-[RLMResults cloneQuery]` could instead handle the case where _backingQuery is null but _backingView is valid.

segiddins: Ah ok, I'll implement the latter.

bdash: This appears to be passing a pointer to a stack object into the `Query` constructor.

tgoyne:
Needs tests that query and sort the backlink results.

segiddins:
Tests added

segiddins:
@tgoyne @bdash please review

jpsim:
I think we'd gain a lot more "coverage" if you added a `RealmCollectionTypeTests` subclass where `getCollection()` and `getAggregateableCollection()` returned backlink-based results.

I must admit I don't understand the changes to `cloneQuery`, so I can't review that.

The rest looks good.

bdash:
Needs a changelog entry.

segiddins:
Changing the `RealmCollectionTypeTests` so everything can support backlinks is turning out to be a massive headache

bmunkholm:
Is this actually WIP?

jpsim:
@bmunkholm yes, this is not sufficiently tested IMO.

mrackwitz:
@bdash: Is that still required or can we close that as being superseded by newer concepts?

jpsim:
Closing as we're going in a different direction for this.

segiddins:
:sob: and there goes my last Realm PR

PR #2499, opened 2015-09-09
[Migration] added renameClass(_:to:).

### LINKED ISSUE #2491: Support renaming tables
As originally requested here: https://groups.google.com/forum/?utm_medium=email&utm_source=footer#!msg/realm-cocoa/sLd7S-jSX40/FeQSRPW0BwAJ

This could be exposed as `-[RLMMigration renameModelFrom:(NSString *)oldName to:(NSString *)newName]`, which would use core's [`Group::rename_table()`](https://github.com/realm/realm-core/blob/4422088866370dfefa70954a7cd103f0da686c5b/src/realm/group.hpp#L341) function.

jpsim:
Fixes #2491.

### DIFF
--- a/Realm/RLMMigration.h
+++ b/Realm/RLMMigration.h
@@ -101,6 +101,19 @@ typedef void (^RLMObjectMigrationBlock)(RLMObject * __nullable oldObject, RLMObj
  */
 - (BOOL)deleteDataForClassName:(NSString *)name;
 
+/**
+ Renames the underlying database table from the original name to the new name. The new class's
+ properties must be identical to the old class.
+
+ @param originalName The name of the original object class to rename.
+ @param newName      The new name of the object class after renaming.
+
+ @warning The rename occurs only in the new realm, which means that enumerating the old class name &
+          accessing the `oldObject` is equivalent to  enumerating the new class name & accessing the
+          `newObject`.
+ */
+- (void)renameClassFrom:(NSString *)originalName to:(NSString *)newName;
+
 @end
 
 RLM_ASSUME_NONNULL_END
--- a/Realm/RLMMigration.h
+++ b/Realm/RLMMigration.h
@@ -102,13 +102,13 @@ typedef void (^RLMObjectMigrationBlock)(RLMObject * __nullable oldObject, RLMObj
 - (BOOL)deleteDataForClassName:(NSString *)name;
 
 /**
- Renames the underlying database table from the original name to the new name. The new class's
- properties must be identical to the old class.
+ Renames the class in the underlying database file from the original name to the new name. The new
+ class's properties must be identical to the old class.
 
  @param originalName The name of the original object class to rename.
  @param newName      The new name of the object class after renaming.
 
- @warning The rename occurs only in the new realm, which means that enumerating the old class name &
+ @warning The rename occurs only in the new Realm, which means that enumerating the old class name &
           accessing the `oldObject` is equivalent to  enumerating the new class name & accessing the
           `newObject`.
  */
--- a/Realm/RLMMigration.mm
+++ b/Realm/RLMMigration.mm
@@ -165,7 +165,7 @@ using namespace realm;
     // Validate original table.
     realm::TableRef originalTable = realm::ObjectStore::table_for_object_type(_realm.group, originalName.UTF8String);
     if (!originalTable) {
-        NSString *message = [NSString stringWithFormat:@"Cannot rename '%@' class because it doesn't exist in the realm.", originalName];
+        NSString *message = [NSString stringWithFormat:@"Cannot rename '%@' class because it doesn't exist in the Realm file.", originalName];
         @throw RLMException(message);
     }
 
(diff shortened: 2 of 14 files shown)


jpsim:
Still missing Swift tests.

Realm/RLMMigration.mm
```diff
@@ -155,4 +155,54 @@ - (BOOL)deleteDataForClassName:(NSString *)name {
...
+    // Validate names.
+    if ([originalName isEqualToString:newName]) {
+        NSString *message = [NSString stringWithFormat:@"Cannot rename '%@' class to the same name.", originalName];
+        @throw RLMException(message);
+    }
+
+    // Validate original table.
+    realm::TableRef originalTable = realm::ObjectStore::table_for_object_type(_realm.group, originalName.UTF8String);
+    if (!originalTable) {
+        NSString *message = [NSString stringWithFormat:@"Cannot rename '%@' class because it doesn't exist in the realm.", originalName];
```
segiddins: `Realm`.

Realm/Tests/RLMTestCase.m
```diff
@@ -199,5 +202,51 @@ - (id)nonLiteralNil
...
+// TODO: add support for:
+//
+// - RLMPropertyTypeObject & RLMPropertyTypeArray
+// - indexedProperties, ignoredProperties, primaryKey
+- (Class)runtimeClassWithName:(NSString *)className properties:(NSArray *)properties {
+    Class objectClass = objc_allocateClassPair(RLMObject.class, className.UTF8String, 0);
+    for (RLMProperty *property in properties) {
+        const char *type = [NSString stringWithFormat:@"%c", property.objcType].UTF8String;
+        objc_property_attribute_t propertyAttributes[] = {
+            { "T",  type},
```
segiddins: spacing

jpsim:
Updating `testRenamingClass` to use dynamic schemas somehow broke the test and I don't know why. Let me know if you have an idea what could be causing this.

jpsim:
@tgoyne @bdash this is ready for review. A few points to note:

1. The Objective-C tests only pass when using runtime-generated classes. I tried moving to dynamic objects (adf1a6f6c2ca00bda69e81149a32384af4ccc19c) but that failed for reasons I couldn't figure out. I personally don't see how this would ever be an issue, but I may be missing something.
2. The Swift tests only run on Swift 2 in Debug mode because they rely on `@testable`. Funnily enough, dynamic objects work for the Swift tests.

jpsim:
CI failures are unrelated.

jpsim:
Bump @tgoyne @bdash

jpsim:
Bump @tgoyne @bdash

Realm/RLMMigration.h
```diff
@@ -101,6 +101,19 @@ typedef void (^RLMObjectMigrationBlock)(RLMObject * __nullable oldObject, RLMObj
  */
 - (BOOL)deleteDataForClassName:(NSString *)name;
 
+/**
+ Renames the underlying database table from the original name to the new name. The new class's
```
bdash: Reword this to avoid talking about tables?

Realm/RLMMigration.h
```diff
@@ -101,6 +101,19 @@ typedef void (^RLMObjectMigrationBlock)(RLMObject * __nullable oldObject, RLMObj
...
+ properties must be identical to the old class.
+
+ @param originalName The name of the original object class to rename.
+ @param newName      The new name of the object class after renaming.
+
+ @warning The rename occurs only in the new realm, which means that enumerating the old class name &
```
bdash: s/realm/Realm/

bdash:
The fact that the schemas must match seems to make this method difficult to use in cases where you may add a property to a class in the future. When you add a property you'll be forced to replace your use of rename with an alternative in order to support migrating from pre-rename schema versions.

tgoyne:
Rather than requiring that the properties match, it could rename the table and then update the target table to match the schema.

tgoyne:
The dynamic object version of the test is failing because `-[RLMObjectSchema initWithClassName:objectClass:properties:]` doesn't set `accessorClass`. After fixing that it passes for me.

jpsim:
Thanks for the tip about `accessorClass`, @tgoyne.

As for updating the target table to match the schema after performing the renaming, would that mostly just be a matter of calling `ObjectStore::create_tables()` with the `update_existing` argument set to `true` immediately after the rename?

magyarosibotond:
When will we get this update? I really need this functionality.

jpsim:
@magyarosibotond I'm still working on this functionality. I don't have a timeline on when I'll be able to complete this work.

magyarosibotond:
Cool! Can't wait to be ready! Any alternative solution until then?

jpsim:
> Cool! Can't wait to be ready! Any alternative solution until then?

You can copy over all the data from a previous class to a new class.

magyarosibotond:
Thanks! Worked like a charm!

codytwinton:
@jpsim Any updates here?

jpsim:
No

pedrofspinho:
@jpsim any deadline on this one? When can we have such feature? Thanks in advance.

ricardopereira:
Closed, not merged 🙁 Seems that this won't be released.

PR #2508, opened 2015-09-09
Fix for a race condition with RLMRealmConfiguration.

bigfish24:
Appeared in `ABFRealmGridController` where the `defaultConfiguration` was requested from 3 threads at the same time. The result was that one thread was attempting to copy `s_defaultConfiguration` while another was mutating the value.

`@synchronized` directive is used with` c_defaultRealmFileName` object as the token to lock from.


segiddins:
:+1:

segiddins:
CHANGELOG?

jpsim:
:+1:

Issue #1916, opened 2015-05-12
 'Object type 'Object' not persisted in Realm' when creating instances using generics


erkie:
An exception is thrown in the new RealmSwift `0.92.2` when trying to create an instance using a generic type (for example in a generic class or function).

## Expected results
It worked in `~> 0.91` so it must have been a bug introduced in `0.92`.

## Actual results
I have created a sample project with a README detailing how to steps and stack traces here:
https://github.com/erkie/ReamGenericsBug

## Offending code

```swift
import UIKit
import RealmSwift

class Post: Object {
    dynamic var title = "Default title"
    dynamic var body = "Default body"
}

class GenericExample<T: Object> {
    func create() -> T {
        return T()
    }
}

@UIApplicationMain
class AppDelegate: UIResponder, UIApplicationDelegate {
    var window: UIWindow?

    func application(application: UIApplication, didFinishLaunchingWithOptions launchOptions: [NSObject: AnyObject]?) -> Bool {
        
        let examplePostGenerator = GenericExample<Post>()
        
        // BLOWS UP HERE
        let myPost = examplePostGenerator.create()
        
        println("Title: \(myPost.title)")
        
        return true
    }
}
```

# Expected output:

`Title: Default title`

## Stack trace

    2015-05-12 08:06:39.130 RealmGenericsBug[52930:5659603] *** Terminating app due to uncaught exception 'RLMException', reason: 'Object type 'Object' not persisted in Realm'
    *** First throw call stack:
    (
        0   CoreFoundation                      0x000000010f958c65 __exceptionPreprocess + 165
        1   libobjc.A.dylib                     0x00000001114c3bb7 objc_exception_throw + 45
        2   Realm                               0x000000010f518653 -[RLMSchema objectForKeyedSubscript:] + 227
        3   Realm                               0x000000010f4ac7eb +[RLMObjectBase sharedSchema] + 123
        4   Realm                               0x000000010f4ab6e5 -[RLMObjectBase init] + 181
        5   RealmSwift                          0x000000010f7a6b61 _TFC10RealmSwift6ObjectcfMS0_FT_S0_ + 49
        6   RealmSwift                          0x000000010f7a6c64 _TFC10RealmSwift6ObjectCfMS0_FT_S0_ + 52
        7   RealmGenericsBug                    0x000000010f3e3c83 _TFC16RealmGenericsBug14GenericExample6createU__fGS0_Q__FT_Q_ + 35
[...]

PR #2514, opened 2015-09-10
Make `Object.init(value:)` a required initializer

kishikawakatsumi:
Related: https://github.com/realm/realm-cocoa/issues/1916

Invoking `init(value:)` with generics type will cause a crash due to Swift compiler cannot find appropriate initializer. (In Swift 2, the compiler treats it as compile error: `Constructing an object of class type 'T' with a metatype value must use a 'required' initializer`)

In Objective-C, both `- init` and `- initWithValue:` are specified `NS_DESIGNATED_INITIALIZER`. Assuming the same meaning it is a 'required' in Swift, `init(value:)` also should be specified `required`, I think. 

@tgoyne @segiddins @jpsim

### DIFF
--- a/RealmSwift-swift1.2/Object.swift
+++ b/RealmSwift-swift1.2/Object.swift
@@ -76,7 +76,7 @@ public class Object: RLMObjectBase, Equatable, Printable {
                     or an `Array` with one object for each persisted property. An exception will be
                     thrown if any required properties are not present and no default is set.
     */
-    public init(value: AnyObject) {
+    public required init(value: AnyObject) {
         self.dynamicType.sharedSchema() // ensure this class' objectSchema is loaded in the partialSharedSchema
         super.init(value: value, schema: RLMSchema.partialSharedSchema())
     }
--- a/RealmSwift-swift1.2/Tests/ObjectCreationTests.swift
+++ b/RealmSwift-swift1.2/Tests/ObjectCreationTests.swift
@@ -125,6 +125,15 @@ class ObjectCreationTests: TestCase {
         XCTAssertEqual(obj1.boolCol, obj2.boolCol, "object created via generic initializer should equal object created by calling initializer directly")
     }
 
+    func testGenericInitWithValue() {
+        func createObject<T: Object>(#value: AnyObject) -> T {
+            return T(value: value)
+        }
+        let obj1: SwiftBoolObject = createObject(value: ["boolCol": true])
+        let obj2 = SwiftBoolObject(value: ["boolCol": true])
+        XCTAssertEqual(obj1.boolCol, obj2.boolCol, "object created via generic initializer should equal object created by calling initializer directly")
+    }
+
     // MARK: Creation tests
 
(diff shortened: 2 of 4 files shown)


segiddins:
retest this please

jpsim:
@kishikawakatsumi doesn't this also require that if a custom initializer is added to an `Object` subclass, that `init()` *and* `init(value:)` also be implemented?

kishikawakatsumi:
@jpsim Yes. When this PR is merged, we must override `init()` and `init(value:)` to implement a custom initializer. Like below:

```
class Person: Object {
    dynamic var name = ""

    required init() {
        super.init()
    }

    required init(value: AnyObject) {
        super.init(value: value)
    }

    init(name: String) {
        super.init()
        self.name = name
    }
}
```

jpsim:
I'm not sure that's a worthwhile tradeoff, to be honest. It's already disappointing that we have to do this for `init()`, I'd rather not add yet another required initializer.

kishikawakatsumi:
Agreed. It might be fine keeping it because defining generic initializer of `init(value:)` occurs compilation error from Swift 2.0, not a runtime error.

I'm a bit concerning about consistency. Both `- init` and `- initWithValue:` are specified `NS_DESIGNATED_INITIALIZER` in Objective-C. It seems consistent that both initializers are also specified `required` in Swift.

gabro:
I'm facing the same issue. Previously (swift 1.2) I worked around the same issue by providing an extension to `Object`

```swift
extension Object {
  convenience init(v: AnyObject) {
    self.init(value: v)
  }
}
```

as suggested in https://github.com/realm/realm-cocoa/issues/1916#issuecomment-101524671

But now, I can't make my code compile any longer, because it requires a required initializer to construct the object.

This is roughly what I'm trying to accomplish.

```swift
func foo<T: Object>(value: [String: AnyObject]) -> T {
  return T(v: value)
}
```

This compiles and works (thanks to the extension above) in swift 1.2, but it doesn't in swift 2.0.

If we don't make `init(value: AnyObject)` a required initializer, how can we work around this issue?

gabro:
Ok, I found a possible workaround.

```swift
func foo<T: Object>(value: [String: AnyObject]) -> T {
  let t = T()
  t.setValuesForKeysWithDictionary(value)
  return t
}
```

Ugly, mutable, imperative, but it seems to be working.

gabro:
Ok, performing the allocation in Objective-C like suggested in https://github.com/realm/realm-cocoa/issues/1916#issuecomment-109256603 does the trick.

Still feels like a giant hack, but better than nothing.

jpsim:
We've decided not to move forward with this. Required initializers *require* subclasses to implement them even if just to forward them to the superclass implementation, so this is not ideal for most users.

I suggest that you file an issue at https://bugs.swift.org about the root problem here, which is (as far as I understand) not being able to invoke `Object.init(value:)` on the dynamic type of a subclass of `RealmSwit.Object`.

PR #2515, opened 2015-09-11
update build-docs.sh to support both Swift 1.2 and 2.0

jpsim:
/cc @segiddins @tgoyne


segiddins:
Why are the tests commented out?

jpsim:
> Why are the tests commented out?

They failed with "Unsupported Realm file format version". I commented them out to test to make sure that the rest of the CI flow works as expected with the new changes.

segiddins:
The docs changes are fine.

PR #2522, opened 2015-09-11
delete all realm files between Swift unit tests

jpsim:
Should fix the `RealmTests.testInitCustomClassList()` test case that's been failing lately /cc @segiddins


jpsim:
Fixed via other means.

PR #2526, opened 2015-09-13
Add Realm.isEmpty

### LINKED ISSUE #2501: Add Realm.isEmpty property
`List` and `Results` are missing the `isEmpty()` method as defined on the CollectionType Protocol.

It would also be really useful with an `isEmpty()` method on the Realm object, to show if it contains any objects at all.

jpsim:
Fixes #2501.

We've avoided doing this for a long time because it's fairly easy to do via existing means, but this is much more semantic and clear, so probably a good thing to do. /cc @tgoyne @bdash 

I didn't do this for `RLMCollection` despite it being available on `RealmCollectionType` (in Swift 2, by "inheritance"), because it's only there implicitly, and I think `isEmpty` provides the same level of clarity and semantic meaning as `.count == 0`.

### DIFF
--- a/Realm/ObjectStore/object_store.cpp
+++ b/Realm/ObjectStore/object_store.cpp
@@ -479,3 +479,16 @@ void ObjectStore::delete_data_for_object(Group *group, const StringData &object_
     }
 }
 
+bool ObjectStore::is_empty(const Group *group) {
+    for (size_t i = 0; i < group->size(); i++) {
+        ConstTableRef table = group->get_table(i);
+        string object_type = object_type_for_table_name(table->get_name());
+        if (!object_type.length()) {
+            continue;
+        }
+        if (!table->is_empty()) {
+            return false;
+        }
+    }
+    return true;
+}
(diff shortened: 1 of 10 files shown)


Realm/ObjectStore/object_store.cpp
```diff
@@ -479,3 +479,17 @@ void ObjectStore::delete_data_for_object(Group *group, const StringData &object_
     }
 }
 
+bool ObjectStore::is_empty(Group *group) {
+    Schema schema = schema_from_group(group);
```
tgoyne: Just looping over the tables (and only checking ones with the class prefix) would be much faster. As it is this will probably surprise people with how expensive it is.

tgoyne:
The given use-cases aren't really very compelling, but I can't think of any reasons why this would be a bad idea.

Realm/ObjectStore/object_store.cpp
jpsim: That was my first pass, but didn't think it would be wise to duplicate this logic. Considering that `schema_from_group` didn't do much more, I wouldn't expect it to be much more expensive, but I'll happily extract that out.

jpsim:
@tgoyne this is good to go

tgoyne:
:+1:

PR #2527, opened 2015-09-13
make Object conform to Hashable in Swift 1.2

### LINKED ISSUE #2519: 'Object' subclasses don't conform to Equatable protocol.
Xcode 7 GM, Swift 2.0. Using branch 'swift-2.0'.

When I started to migrate my project to swift 2.0, I have noticed that I unable to store 'Object' subclasses in collections that require it's elements to conform to Equatable protocol. And as result Hashable as well.

Here is the simplest example to reproduces it.
```swift
class A: Object {
}

let array = Set<A>()

let instance = A()
let index = [A]().indexOf(instance)
 ```

When you will try to compile that code you will get error on set initialization that type 'A' does not conform to protocol Hashable. And that there is no method 'indexOf' for supplied parameters (indexOf(element) requires element to conform to Equatable).

image

But on other case set of 'object' instances works fine.
```swift
let object = Set<Object>()
 ```

Could you help to find what can cause this issue?

jpsim:
Fixes #2519 for Swift 1.2. /cc @bdash 

Note: `Hashable` implies `Equatable` because it inherits from it.

### DIFF
--- a/RealmSwift-swift1.2/Object.swift
+++ b/RealmSwift-swift1.2/Object.swift
@@ -53,7 +53,7 @@ the `objects(_:)` instance method on `Realm`.
 
 See our [Cocoa guide](http://realm.io/docs/cocoa) for more details.
 */
-public class Object: RLMObjectBase, Equatable, Printable {
+public class Object: RLMObjectBase, Hashable, Printable {
 
     // MARK: Initializers
 


bdash:
Can we test this?

jpsim:
> Can we test this?

Yes, we can ensure that a test subclass responds successfully to a check for `Hashable` conformance, and try to actually call `hashValue` on it.

jpsim:
retest this please

jpsim:
Turns out this is not necessary for Swift 1.2 since the bug described in #2519 only applies to Xcode 7.0 (not present in 6.4 or 7.1).

PR #2536, opened 2015-09-16
Fix the sizes of the buffers used by RLMHashData.

bdash:
/cc @tgoyne @jpsim


tgoyne:
:+1:

jpsim:
Seems right to me. :+1:

PR #2555, opened 2015-09-22
Set default Swift version to 2.0

jpsim:
Now's the time to do this with Xcode 7 being officially out. I also updated the our web app to mirror changes in `master` in both `swift-1.2` and `swift-2.0` branches.

- [x] Update packaging format

### DIFF
--- a/Realm/Tests/Swift
+++ b/Realm/Tests/Swift
@@ -1,1 +1,1 @@
-Swift1.2
+Swift2.0
--- a/RealmSwift
+++ b/RealmSwift
@@ -1,1 +1,1 @@
-RealmSwift-swift1.2
+RealmSwift-swift2.0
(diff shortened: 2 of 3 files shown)


jpsim:
Turns out we're already packaging our Swift zips with subfolders for each Swift version (`swift-1.2`/`swift-2.0`) so the only packaging tasks required (that I could think of) was to copy the Swift 2.0 versions of the Carthage frameworks.

So this should be ready to be merged.

tgoyne:
:+1: 

swift-1.2 branch will need this reverted after it's been merged to that branch.

jpsim:
> swift-1.2 branch will need this reverted after it's been merged to that branch.

indeed.

Only CI failure was due to #2119. Merging.

PR #2560, opened 2015-09-23
[Doc] Explain when an object becomes invalidated

### LINKED ISSUE #2486: Documentation Request: What causes an object to be invalidated?
Currently there is no documentation to explain what would cause a `RLMObject` to be marked as `invalidated`. The two questions I have that would be nice if they were documented are:

* What causes `invalidated` to be `YES`?
* ~~How is `invalidated` related to `deletedFromRealm`, if at all?~~

Edit:
I just noticed `deletedFromRealm` is deprecated in the .h file but the documentation doesn't say that. Another possible update.

mrackwitz:
Fixes #2486.

### DIFF
--- a/Realm/RLMObject.h
+++ b/Realm/RLMObject.h
@@ -232,6 +232,9 @@ RLM_ASSUME_NONNULL_BEGIN
 
 /**
  Indicates if an object can no longer be accessed.
+ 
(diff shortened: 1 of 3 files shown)


jpsim:
Please also update `RealmSwift-swift1.2/Object.swift`, otherwise :+1:

jpsim:
Thanks @mrackwitz!

PR #2565, opened 2015-09-23
Fix issues with switching between simulators

tgoyne:
@jpsim


jpsim:
once this passes CI, :+1:

PR #2567, opened 2015-09-24
Fix bitcode generation for Release builds

conradev:
The current binary release of Realm for watchOS is unsuitable for submitting to the App Store.

It was compiled with `-fembed-bitcode-marker` and not `-fembed-bitcode`, meaning the bitcode section is size 1, or empty:

```
$ otool -l ./realm-objc-0.95.1/watchos/Realm.framework/Realm | grep -A 5 __bundle
  sectname __bundle
   segname __LLVM
      addr 0x00184000
      size 0x00000001
    offset 1589248
     align 2^0 (1)
```

This is because `BITCODE_GENERATION_MODE` is normally set to `marker` when building, and only set to `bitcode` when archiving. I assume this is for performance reasons.

This pull request embeds bitcode in all Release builds.


jpsim:
Thanks a lot for this, and for explaining what was wrong. Your thoroughness is appreciated.

If this fix is indeed correct (and it sure looks like it is), I imagine the same should be done for the `iOS Dynamic` scheme as well as the `RealmSwift.xcodeproj` xcconfig.

PR #2575, opened 2015-09-24
Actually enable bitcode for more things

tgoyne:
Enables bitcode generation rather than just the marker for release builds of Realm.framework (iOS static and dynamic) and all release builds of RealmSwift.framework (it appears to be ignored without issues for OS X).

@jpsim


jpsim:
:+1:

PR #2577, opened 2015-09-24
Resolve symlinks in cocoapods-setup by copying rather than moving

tgoyne:
Makes it work in the presence of multiple symlinks to the same thing.

@jpsim


jpsim:
yup :+1:

PR #2578, opened 2015-09-24
Fix swift 2.0 Carthage installation example

tgoyne:
The path to the Realm frameworks was out of date.

@jpsim

### DIFF
--- a/examples/installation/build.sh
+++ b/examples/installation/build.sh
@@ -30,57 +30,8 @@ EOF
 
 COMMAND="$1"
 
-prelaunch_simulator() {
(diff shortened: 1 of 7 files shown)


examples/installation/build.sh
```diff
@@ -31,29 +31,52 @@ EOF
 COMMAND="$1"
 
 prelaunch_simulator() {
```
jpsim: since this function is getting a bit complex, it would be worthwhile to keep in a shared script like with do for the Swift version

jpsim:
:+1:

PR #2537, opened 2015-09-16
Fix issues with partial shared schema init

tgoyne:
When the initial partial schema init is done with an explicit class list, duplicate object schemas for those classes would end up in the final shared schema, along with object schemas for the accessor classes. No new tests for any of this because we can't test it via our testing framework....

Closes #2534.

@jpsim @bdash

PR #2583, opened 2015-09-25
Fix 'unrecognized selector' exception when specifying class subsets.

### LINKED ISSUE #2494: Unrecognized selector sent to RLMAccessor class when registering classes in a custom schema
Hey guys,

Out of the sudden I have this weird exception:
     *** Terminating app due to uncaught exception 'NSInvalidArgumentException', reason: '+[RLMStandalone_MyClass requiredProperties]: unrecognized selector sent to class 0x7fa62b56e190'

I simply don't get how this issue appeared, but the bad thing is that it crashes on main & I have no stack-trace.

After a lot of debugging I found what caused this issue, I was creating my custom Realm.Configuration & this is how it looked like:
      
    let config = Realm.Configuration(path: path, objectTypes: [MyClass.self])

This issue appears after defining my object types in the realm, if I remove it, everything works.
What is the purpose of defining the objectTypes & what is the correct way to be used?

Thanks!

jpsim:
Fixes #2494. /cc @tgoyne @bdash

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,7 +21,8 @@ x.x.x Release notes (yyyy-MM-dd)
 
 ### Bugfixes
 
-* None.
+* Fix an issue where an 'unrecognized selector' exception would be thrown when
+  specifying `RLMRealmConfiguration.objectClasses`/`Realm.Configuration.objectTypes`.
 
 0.95.1 Release notes (2015-09-23)
 =============================================================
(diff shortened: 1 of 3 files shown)


tgoyne:
This breaks explicitly including a class which is not part of the default schema. #2537 fixes the same bug while still allowing that.

jpsim:
ah yes, let's go with that then

PR #2586, opened 2015-09-28
Include a watchOS framework to GitHub release zip

kishikawakatsumi:
Fix not to contain a watchOS framework to GitHub release zip for Carthage installation.


jpsim:
:+1:

PR #2592, opened 2015-09-28
Copy strip-frameworks.sh into watchOS framework

kishikawakatsumi:
We must remove i386 binary from watchOS framework. Otherwise,  errors will occur when submitting to the iTunes Connect.

screen shot 2015-09-29 at 01 41 45

screen shot 2015-09-29 at 01 41 55

Above errors are different from iOS errors, but it resolved by removing the i386 slice.


kishikawakatsumi:
### Changes:

- Remove unused "Link Binary with Libraries" and "Copy Bundle Resources"
- Add resource copy phase to copy `strip-frameworks.sh`


### Before

screen shot 2015-09-29 at 02 01 50

### After

screen shot 2015-09-29 at 02 01 29

/cc @jpsim @mrackwitz

tgoyne:
:+1: 

Setup instructions on the site will need to be updated too.

kishikawakatsumi:
@tgoyne OK, I created PR for updating documentation. https://github.com/realm/realm.io/pull/1035

jpsim:
Nice work @kishikawakatsumi! :+1: 

Although the CI failure is legitimate (file format failure) and should be addressed asap, it's not caused by this PR.

PR #2595, opened 2015-09-28
increase max std dev for testCommitWriteTransactionWithLocalNotification

jpsim:
I'm not sure this value is read from the baseline file, but if this works, it'll be preferable to device test failures for that specific test like happened here: https://ci.realm.io/job/objc_pr/1917/configuration=Release,swift_version=1.2,target=ios-device/console

Also removed an old baseline for an unused device.


jpsim:
Last CI failure was unrelated (operation failed on `sf_host3_1` again)

PR #2596, opened 2015-09-28
Clean up lingering in-memory realms for perf tests

tgoyne:
If the test application crashes while running a performance test, the in-memory realm file isn't deleted, and then is opened the next time the tests are run. This works out fine when it's the same file format version, but upgrading in-memory realms doesn't work correctly (and of course downgrading isn't supported at all).

@jpsim


jpsim:
This would ideally come with a `FIXME` to remove this once core handles this scenario more gracefully. Otherwise :+1:

tgoyne:
Added FIXME.

Fix in core: https://github.com/realm/realm-core/pull/1180

jpsim:
:+1:

PR #2599, opened 2015-09-29
Fix an assertion failure when deleting objects while enumerating

tgoyne:
Deleting objects after they're reached by the enumeration worked fine, but deleting them before they're reached wasn't handled correctly.

No changelog entry because 0.95.1 included a workaround (the proper fix required a core update).


jpsim:
:+1:

PR #2601, opened 2015-09-29
Support using initWithValue: in Swift property defaults

tgoyne:
If an object has a property linking to another object and that property is initialized using `initWithValue:`, that initializer may get run during schema initialization before the linked class's schema is initialized and thus throw an exception. We don't actually use any of the property values of linked objects during schema init, so just do nothing in `initWithValue:` if the class's schema has not been initialized.

Fixes intermittent test failures hit by seg-nsnumber.


jpsim:
Short of writing a whole set of multi-process tests for Realm Swift, this is very good. :+1:

Issue #2597, opened 2015-09-28
Update docs to encourage avoiding platform-dependent types


jpsim:
To avoid situations like #2594.

jpsim:
Done in https://github.com/realm/realm-cocoa/issues/2604 and https://github.com/realm/realm.io/pull/1123

PR #2604, opened 2015-09-29
Remove CGFloat from "Supported property types" section of docs

tgoyne:
While it technically works, it's a bad idea due to it differing in size between 32-bit and 64-bit. NSInteger is okay as our int columns aren't fixed-size.


jpsim:
:+1: partially addresses #2597, but docs should also be updated to explicitly discourage use of `CGFloat`.

PR #2611, opened 2015-09-30
Add validation for class subsets

tgoyne:
If the configurations objectTypes/objectClasses array contained objects which linked to objects not in the array, trying to open a Realm with the configuration would result in a crash. Fix this by validating that all links are valid when the class subset is set.


bdash:
:+1:

jpsim:
Nice! :+1:

PR #2614, opened 2015-10-01
Implement query support for null

tgoyne:
@bdash 

Main WIP part is that there's a lot more stuff to test, but I didn't want to wait too long on throwing this up in case it conflicts with the other query work.

### DIFF
--- a/Realm/RLMQueryUtil.mm
+++ b/Realm/RLMQueryUtil.mm
@@ -207,18 +207,10 @@ void add_string_constraint_to_query(realm::Query &query,
     }
 }
 
-void add_string_constraint_to_query(realm::Query &query,
-                                    NSPredicateOperatorType operatorType,
-                                    NSComparisonPredicateOptions predicateOptions,
-                                    Columns<String> &&column,
-                                    id value) {
-    add_string_constraint_to_query(query, operatorType, predicateOptions, std::forward<Columns<String>>(column), RLMStringDataWithNSString(value));
-}
-
 void add_string_constraint_to_query(realm::Query& query,
                                     NSPredicateOperatorType operatorType,
                                     NSComparisonPredicateOptions predicateOptions,
-                                    id value,
+                                    StringData value,
                                     Columns<String>&& column) {
     switch (operatorType) {
         case NSEqualToPredicateOperatorType:
@@ -254,11 +246,11 @@ void validate_and_extract_between_range(id value, RLMProperty *prop, id *from, i
                     @"NSArray objects must be of type %@ for BETWEEN operations", RLMTypeToString(prop.type));
 }
 
-template <typename... T>
+template <typename L, typename R>
 void add_constraint_to_query(realm::Query &query, RLMPropertyType type,
                              NSPredicateOperatorType operatorType,
                              NSComparisonPredicateOptions predicateOptions,
-                             const std::vector<NSUInteger>& linkColumns, T... values);
+                             const std::vector<NSUInteger>& linkColumns, L lhs, R rhs);
 
 void add_between_constraint_to_query(realm::Query &query, std::vector<NSUInteger> const& indexes, RLMProperty *prop, id value) {
     id from, to;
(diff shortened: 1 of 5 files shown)


Realm/RLMQueryUtil.mm
```diff
@@ -538,6 +517,43 @@ void add_constraint_to_query(realm::Query &query, RLMPropertyType type,
     }
 }
 
+void do_add_constraint_to_query(realm::Query&, RLMPropertyType, NSPredicateOperatorType,
+                                NSComparisonPredicateOptions, const std::vector<NSUInteger>&,
+                                id, realm::null)
+{
+    // This probably isn't actually reachable since there's earlier checks for
+    // it, but they're done before null is un-type-erased
+    @throw RLMPredicateException(@"Invalid predicate expressions",
```
bdash: This definitely isn't reachable due to the logic in `update_query_with_predicate`. I'm guessing this overload is necessary in order for `add_constraint_to_query` to compile?

Realm/RLMQueryUtil.mm
```diff
@@ -401,73 +405,48 @@ void process_or_group(Query &query, id array, Func&& func) {
...
-template <typename TableGetter>
-struct ValueOfTypeHelper<String, TableGetter> {
-    static id convert(TableGetter&&, id value)
-    {
-        return value;
-    }
+template <typename RequestedType, typename TableGetter>
+struct ValueOfTypeHelper {
+    static realm::null convert(TableGetter&&, realm::null) { return realm::null(); }
+    static auto convert(TableGetter&&, id value) { return ::convert<RequestedType>(value); }
```
bdash: This is much tidier!

bdash:
It'll conflict with the key-path vs key-path changes I'm working on, but I can work that out when we pick up a new enough version of core to support them.

:+1:

tgoyne:
Added basic tests for every supported operator on every nullable property type.

Realm/Tests/QueryTests.m
```diff
@@ -1929,4 +1922,269 @@ - (void)testSortingColumnsWithNull {
...
+        {@"doubleObj", @"1", @"0", @1, @0, true},
+        {@"string", @"'a'", @"''", @"a", @"", false, true},
+        {@"data", nil, nil, notMatchingData, matchingData},
+        {@"date", nil, nil, notMatchingDate, matchingDate, true},
+    };
+
+    // Assert that the query "prop op value" gives expectedCount results when
+    // assembled via string formatting
+#define RLMAssertCountWithString(expectedCount, op, prop, value) \
+    do { \
```
jpsim: why are these defines all wrapped in a `do {} while(0)`?

tgoyne: It makes the macro expand to the equivalent of a single statement that requires a trailing semicolon, which avoids a bunch of potential bugs from macro-expansion edge cases.

jpsim: fair enough, thanks!

bdash:
:+1:

Realm/RLMQueryUtil.mm
```diff
@@ -538,6 +519,43 @@ void add_constraint_to_query(realm::Query &query, RLMPropertyType type,
...
+void add_constraint_to_query(realm::Query &query, RLMPropertyType type,
+                             NSPredicateOperatorType operatorType,
+                             NSComparisonPredicateOptions predicateOptions,
+                             const std::vector<NSUInteger>& linkColumns, L lhs, R rhs)
+{
+    // The expression operators are only overloaded for realm::null on the rhs
+    RLMPrecondition(!is_nsnull(lhs), @"Unsupported operator",
+                    @"Nil is only supported on the right side of operators");
+
+    if (is_nsnull(rhs)) {
```
jpsim: this could be simplified a bit by using `auto validatedRHS = is_nsnull(rhs) ? realm::null() : rhs;` and passing the validated value to `do_add_constraint_to_query`.

tgoyne: That won't compile because `rhs` and `realm::null()` are unrelated types that can't be converted to each other. The whole point of this is to call different overloads (ultimately of the core query functions) depending on the runtime value of `rhs`.

jpsim: ah, yes makes sense

PR #2624, opened 2015-10-02
Added an objects method to allow querying a realm from class name as string

alexasfb:
- Added a method to allow querying objects from a Realm passing a class name as a String directly.

### DIFF
--- a/RealmSwift-swift2.0/Realm.swift
+++ b/RealmSwift-swift2.0/Realm.swift
@@ -405,7 +405,18 @@ public final class Realm {
     */
     public func objects<T: Object>(type: T.Type) -> Results<T> {
         // FIXME: use T.className()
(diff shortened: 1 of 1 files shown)


jpsim:
Turns out we already have a `Realm.dynamicObjects()` function that does exactly this! https://github.com/realm/realm-cocoa/blob/v0.95.2/RealmSwift-swift2.0/Realm.swift#L408-L425

It's not part of our HTML documentation because it's not relevant to the vast majority of Realm users out there, but it is an officially supported API. It's part of a larger set of dynamic APIs (along with [`Realm.dynamicCreate()`](https://github.com/realm/realm-cocoa/blob/v0.95.2/RealmSwift-swift2.0/Realm.swift#L309-L340), among others).

Issue #2585, opened 2015-09-27
Bitcode not embedded in iOS framework binaries


hibento:
Hello,

I'm using the latest release v0.95.2. The release notes states that bitcode is enabled, but in the corresponding sources bitcode isn't enabled for all targets / schemes. Could this be a problem?

In my Xcode Project (bitcode is enabled) I get this error:

    /Applications/Xcode.app/Contents/Developer/Toolchains/
     XcodeDefault.xctoolchain/usr/bin/bitcode_strip: 
    bitcode segment can't have symbols defined it its sections in:   
    /Users/hibento/Repositories/myProject/Carthage/
     Build/iOS/myFramework.framework/Frameworks/Realm.framework/Realm (for architecture arm64)
    error: bitcode_strip  /Users/hibento/Repositories/myProject/Carthage/Build/iOS/
     myFramework.framework/Frameworks/Realm.framework/Realm:       
    /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain
     /usr/bin/bitcode_strip exited with 1

jpsim:
Thanks for reporting this. You're right that not all platforms that we say have bitcode embedded currently do, despite our recent efforts to do so (see #2567 and #2575 among several others).

Running `otool -l core/librealm-*.a | grep LLVM` fails for all core libraries other than watchOS. From our 0.95.2 release zips, our watchOS binaries for Realm Objective-C and Realm Swift are the only ones that contain bitcode according to the above test.

hibento:
Are there any news regarding this issue? Could I provide you with more information? Currently we're unable to build / run Realm on our project and device.

tgoyne:
This should have no impact on your ability to run your project locally. Bitcode is only relevant for submitting to the app store.

hibento:
Okay, that´s true - but in our project with using 4 other frameworks, all of them with bitcode support. Correct if I'm wrong: But you either activate bitcode for all frameworks in your project or not at all. So to be able to build our project we have to change the settings for the remaining 4 frameworks. 

I would do this gladly, but since we're using some prebuilt frameworks that´s not an option.

jpsim:
> So to be able to build our project we have to change the settings for the remaining 4 frameworks.

That shouldn't be necessary for building & running, only when archiving.

hibento:
>> So to be able to build our project we have to change the settings for the remaining 4 frameworks.

> That shouldn't be necessary for building & running, only when archiving.

[...]

PR #2626, opened 2015-10-02
Bump core version to 0.92.3

jpsim:
and add release notes for 0.95.3. /cc @tgoyne @bdash


jpsim:
Closing because of https://github.com/realm/realm-cocoa/issues/2585#issuecomment-145109041.

PR #2627, opened 2015-10-02
Update core version to 0.92.4

jpsim:
This expects core's tarball to have the following structure:

```
LICENSE
include/
librealm-dbg.a
librealm-ios-bitcode-dbg.a
librealm-ios-bitcode.a
librealm-ios-dbg.a -> librealm-ios-bitcode-dbg.a
librealm-ios-no-bitcode-dbg.a
librealm-ios-no-bitcode.a
librealm-ios.a -> librealm-ios-bitcode.a
librealm-watchos-dbg.a
librealm-watchos.a
librealm.a
release_notes.txt
```

This automatically switches core bitcode symlink when building the "iOS Dynamic" target of Realm.xcodeproj.

Once this is merged, we'll need to update the `swift-1.2` branch's `Realm.podspec` to use `s.ios.vendored_library = librealm-ios-no-bitcode.a`. This way, we can point CocoaPods users of Realm Objective-C who can't update to Xcode 7 to point to the `swift-1.2` branch. This will also support CP users of Realm Swift.

To be extra safe, we should probably also be running `set-core-bitcode-symlink` when building the static iOS framework target (EDIT: Done).

/cc @tgoyne @bdash

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,6 +30,11 @@ x.x.x Release notes (yyyy-MM-dd)
 * Throw an exception when a class subset has objects with array or object
   properties of a type that are not part of the class subset.
 
+0.95.3 Release notes (2015-10-02)
+=============================================================
+
+* Compile iOS Simulator framework architectures with `-fembed-bitcode-marker`.
+
 0.95.2 Release notes (2015-09-24)
 =============================================================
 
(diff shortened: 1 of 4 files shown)


gintechsystems:
So how do we fix this to work with CocoaPods 0.39.0? I am getting an error saying RLMResults.h not found.

gintechsystems:
I was able to fix the issue by downgrading to 0.38.0 of CocoaPods.

PR #2628, opened 2015-10-03
Update release package to contain both Xcode 6 and 7 builds

tgoyne:
I suspect that the examples need updating as well, but everything else works. New package structure is `ios/xcode-6/Realm.framework` etc.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -735,7 +735,6 @@ case "$COMMAND" in
     "package-osx")
         cd tightdb_objc
         REALM_SWIFT_VERSION=2.0 sh build.sh test-osx
-        REALM_SWIFT_VERSION=2.0 sh build.sh osx
 
         cd build/DerivedData/Realm/Build/Products/Release
         zip --symlinks -r realm-framework-osx.zip Realm.framework
--- a/build.sh
+++ b/build.sh
@@ -582,7 +582,7 @@ case "$COMMAND" in
         ;;
 
     "examples-ios")
-        if [[ -f "examples/ios/objc" ]]; then
+        if [[ -d "examples/ios/objc" ]]; then
             project="examples/ios/objc/RealmExamples.xcodeproj"
         elif [[ "$REALM_SWIFT_VERSION" = 1.2 ]]; then
             project="examples/ios/xcode-6/objc/RealmExamples.xcodeproj"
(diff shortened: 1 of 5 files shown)


build.sh
```diff
@@ -694,24 +700,35 @@ case "$COMMAND" in
...
 
-        cd build/ios-dynamic
-        zip --symlinks -r realm-dynamic-framework-ios.zip Realm.framework
+        zip --symlinks -r build/ios-dynamic/realm-dynamic-framework-ios.zip xcode-6 xcode-7
         ;;
 
     "package-osx")
         cd tightdb_objc
         REALM_SWIFT_VERSION=2.0 sh build.sh test-osx
+        REALM_SWIFT_VERSION=2.0 sh build.sh osx
```
jpsim: why is this now necessary?

tgoyne: It isn't; I just did the wrong thing when reversing my change to call `osx` rather than `test-osx` to speed up the packaging test cycle a bit.

jpsim: Ok, then :+1: once this is removed.

tgoyne:
Running test-package-release locally to verify that everything works and will merge once that passes.

tgoyne:
Updated the packaged obj-c examples to support the new locations of Realm.framework. There's now two copies of each example in the release package (in xcode-6 and xcode-7 directories) to avoid having to do fancy things in the examples project.

jpsim:
Examples failed in `verify-ios-static`.

examples/installation/osx/swift-2.0/CarthageExample/CarthageExample.xcodeproj/project.pbxproj
```diff
@@ -44,8 +44,8 @@
 /* End PBXCopyFilesBuildPhase section */
 
 /* Begin PBXFileReference section */
-		E802DBE01B3CD5410047225A /* Realm.framework */ = {isa = PBXFileReference; lastKnownFileType = wrapper.framework; name = Realm.framework; path = Carthage/Build-Xcode-7.0/Mac/Realm.framework; sourceTree = "<group>"; };
```
jpsim: huh, I thought we had done this months ago...

jpsim:
:+1:

tgoyne:
Installation examples now pass locally, so hopefully everything works.

PR #2633, opened 2015-10-05
Automatically select the correct core library when installing via CocoaPods

tgoyne:
Rather than requiring that people using Xcode 6 use the swift-1.2 branch.

@jpsim


jpsim:
The reason I didn't do this initially was this this will only work when REALM_SWIFT_VERSION is set by the user, or they have the right Xcode version set in `xcode-select`. Whereas setting the branch in the Podfile will make sure the Podfile behaves the same way on different systems.

jpsim:
From our just-now conversation in the office, the downsides I stated are trumped by having our Xcode 6 users be able to target official releases of Realm rather than a master fork. So :+1:

PR #2647, opened 2015-10-07
Support passing in the branch to use to the CocoaPods installation samples

tgoyne:
Doesn't modify the swift-1.2 test because that actually requires using a different branch, but the swift-2.0 one will hopefully catch any issues.

### DIFF
--- a/examples/installation/ios/objc/CocoaPodsDynamicExample/Podfile
+++ b/examples/installation/ios/objc/CocoaPodsDynamicExample/Podfile
@@ -2,11 +2,12 @@
 # platform :ios, '6.0'
 
 use_frameworks!
+branch = ENV['sha']
 
-target 'CocoaPodsDynamicExample' do
-  pod 'Realm', :git => 'git@github.com:realm/realm-cocoa.git'
+target 'CocoaPodsExample' do
+  pod 'Realm', git: 'git@github.com:realm/realm-cocoa.git', branch: branch
 end
 
-target 'CocoaPodsDynamicExampleTests' do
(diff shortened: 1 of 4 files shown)


examples/installation/ios/objc/CocoaPodsDynamicExample/Podfile
```diff
@@ -2,11 +2,12 @@
 # platform :ios, '6.0'
 
 use_frameworks!
+branch = ENV['sha']
```
jpsim: `ranch = ENV['sha'] || "master"` so we can run locally?

jpsim:
:+1: but it'd be nice to fallback on `master`

examples/installation/ios/objc/CocoaPodsDynamicExample/Podfile
tgoyne: Passing `nil` makes it use the default branch (i.e. master).

jpsim: neat, good to go then

examples/installation/ios/objc/CocoaPodsDynamicExample/Podfile
```diff
@@ -2,11 +2,12 @@
...
 
-target 'CocoaPodsDynamicExample' do
-  pod 'Realm', :git => 'git@github.com:realm/realm-cocoa.git'
+target 'CocoaPodsExample' do
+  pod 'Realm', git: 'git@github.com:realm/realm-cocoa.git', branch: branch
 end
 
-target 'CocoaPodsDynamicExampleTests' do
```
jpsim: This change broke things.

PR #2649, opened 2015-10-07
Support assigning `Results` to `List` properties via KVC

### LINKED ISSUE #2634: Assign Results to List using subscript
```swift
class Conversation: Object {
    let messages = List<Message>()
}

conversation["messages"] = messages
```

This works fine if `messages` is a `RLMArray`, `RLMListBase`, or `NSArray` but does not work if it is a `Results`. It throws an exception in `RLMDynamicValidatedSet`.

As far as I can tell, `RLMDynamicSet` only requires a collection that conforms to `NSFastEnumeration` which `ResultsBase` and `RLMResults` do.

It whole be helpful if you could assign using `Results` or `RLMResults` too.

My current workaround is to create an `Array` from the `Results` before assigning

```swift
let messages = //some query
conversation["messages"] = Array(messages)
```

jpsim:
Fixes #2634.

Ideally we'd support this via RealmCollectionType as well, but I couldn't really get that to work. /cc @tgoyne @bdash

### DIFF
--- a/RealmSwift-swift1.2/Tests/RealmCollectionTypeTests.swift
+++ b/RealmSwift-swift1.2/Tests/RealmCollectionTypeTests.swift
@@ -292,6 +292,10 @@ class RealmCollectionTypeTests: TestCase {
         XCTAssertEqual(0, collection.count)
     }
 
+    func testAssignListProperty() {
+        fatalError("abstract")
+    }
+
     func testArrayAggregateWithSwiftObjectDoesntThrow() {
         let collection = getAggregateableCollection()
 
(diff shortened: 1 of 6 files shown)


tgoyne:
Would be good to have an obj-c test as well since the change looks like it should cover that, but otherwise :+1:.

jpsim:
objc test added.

jpsim:
Thoughts on this @bdash?

RealmSwift-swift1.2/Tests/RealmCollectionTypeTests.swift
```diff
@@ -302,9 +306,35 @@ class RealmCollectionTypeTests: TestCase {
...
+class ResultsTests: RealmCollectionTypeTests {
+    override class func defaultTestSuite() -> XCTestSuite {
+        // Don't run tests for the base class
+        if isEqual(ResultsTests) {
+            return XCTestSuite(name: "empty")
+        }
+        return super.defaultTestSuite()
+    }
+
+    func base() -> Results<SwiftStringObject> {
```
bdash: I'm not clear what the name "base" represents here. Base of what?

bdash:
:+1:

The Swift test changes here are another place where more detail in the commit message would have been appreciated.

jpsim:
Good points on providing more details, I'll make an effort to improve that in the future.

jpsim:
CI failure unrelated (#2119).

PR #2650, opened 2015-10-08
[Query] Make sure that linked objects are attached

mrackwitz:
Without that change, if you query for an object, which is linked to a just newly created object, which was never attached to the Realm, you get something arbitrary returned.

* [x] Changelog
* [x] Check whether object is in the *same* realm as the one being queried
* [x] ~~Check whether object is from the *same* class as the one being queried~~ (done already by `validate_property_value`)
* [x] Fix `RealmCollectionTypeTests.testArrayAggregateWithSwiftObjectDoesntThrow` Swift tests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,8 @@ x.x.x Release notes (yyyy-MM-dd)
 ### Bugfixes
 
 * Fix crash when trying to retrieve object instances via `dynamicObjects`.
+* Throw an exception when querying on a link providing objects, which are from a different Realm.
+* Return empty results when querying on a link providing an unattached object.
 * Fix crashes or incorrect results when calling `-[RLMRealm refresh]` during
   fast enumeration.
 * Add `Int8` support for `RealmOptional`, `MinMaxType` and `AddableType`.
--- a/Realm/RLMQueryUtil.mm
+++ b/Realm/RLMQueryUtil.mm
@@ -444,6 +444,11 @@ void add_link_constraint_to_query(realm::Query & query,
         return;
     }
 
+    // NOTE: This precondition assumes that the argument `obj` will be always originating from the
+    // queried table as verified before by `validate_property_value`
+    RLMPrecondition(query.get_table()->get_link_target(column).get() == obj->_row.get_table(),
+                    @"Invalid value origin", @"Object must be from the Realm being queried");
+
     query.links_to(column, obj->_row.get_index());
 }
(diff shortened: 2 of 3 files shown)


jpsim:
This makes all the `RealmCollectionTypeTests.testArrayAggregateWithSwiftObjectDoesntThrow` Swift tests fail.

tgoyne:
There also needs to be a check for the object being in the *same* realm as the one being queried.

I don't think this needs to be an error, though. Rather than throwing it could just insert a `FalseExpression` node.

mrackwitz:
>I don't think this needs to be an error, though. Rather than throwing it could just insert a `FalseExpression` node.

But you mean here that you would effectively return an empty result if the object isn't attached? I'm not sure if that's a good idea. It seems like it could be a programmer error in most cases.

tgoyne:
`indexOfObject:` already does that.

mrackwitz:
>`indexOfObject:` already does that.

Even though it would be a breaking semantic change, I'd lean towards changing the behavior of `indexOfObject:` to fail irrecoverably in that case as well.

tgoyne:
I don't think it's unreasonable to have a function which can operate on either standalone or persisted objects in the same way. Sort of the opposite, actually; ideally anything which doesn't involve a write transaction should rarely need to care if the objects it's given are persisted in a realm (or persisted in the *same* realm, for that matter). The semantics of searching for an object that happens to not be persisted in the realm being searched seems perfectly well-formed to me, so I see no reason to just arbitrarily make it an error.

mrackwitz:
I think there are different models, in which both behaviors are well-formed, while they are not in the other. If we want to stick with the latter, I'd still strongly suggest that we highlight the behavior in this specific case in the documentation of affected methods.

mrackwitz:
@tgoyne: How would you validate whether the realms are the same? In the function there is no reference to the `RLMRealm` accessible for which the query is built.

tgoyne:
`query.get_table()->get_link_target(column).get() == obj->_row.get_table()` would cover both the object being from a different realm and being the wrong class (which currently also doesn't appear to be checked anywhere?).

mrackwitz:
Yes, doesn't seem like that's checked so far. But good idea! Thanks :+1:

mrackwitz:
@tgoyne: This seems to be ready now to merge. Can you give it another review pass?

tgoyne:
:+1:

PR #2656, opened 2015-10-10
add performance test baselines for iPhone 5s & iPad Mini 2

jpsim:
These are now connected to CI so they need performance baselines. /cc @tgoyne


jpsim:
@tgoyne

build.sh
```diff
@@ -218,6 +220,16 @@ test_ios_devices() {
     configuration="$3"
     failed=0
     for device in "${serial_numbers[@]}"; do
+        if [[ "$device" = "07510d12c09c53e44b3682dd0367934f82cfab9c" ]]; then
```
tgoyne: Doesn't this result in the swift devices tests job just not running if the VM with the ios7 device gets picked? Removing it as on option using a tag for ios 8+ in jenkins seems better.

jpsim: You're right, I made this PR before we decided to parallelize the objc and swift device tests. I'll remove this change.

jpsim: Done.

PR #2665, opened 2015-10-13
Consolidate parameter validation in RLMRealmConfiguration

tgoyne:
Previously some things were validated on property set, some were validated in RLMRealm, and in RealmSwift some were validated on both. This makes RLMRealmConfiguration's setters validate everything that it can, Realm.Configuration validate nothing until it's converted to an RLMRealmConfiguration, and RLMRealm trusts that the passed-in config is valid.

This makes supporting the objectstore config object a bit easier and makes the tests more coherent.

### DIFF
--- a/Realm/RLMRealmConfiguration.mm
+++ b/Realm/RLMRealmConfiguration.mm
@@ -155,7 +155,7 @@ static NSString * const c_defaultRealmFileName = @"default.realm";
         @throw RLMException(@"In-memory identifier must not be empty");
     }
 
-    _inMemoryIdentifier = inMemoryIdentifier;
+    _inMemoryIdentifier = [inMemoryIdentifier copy];
     _path = nil;
 }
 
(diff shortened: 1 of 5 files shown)


jpsim:
This looks great! Much needed cleanup and validation. :+1:

Realm/RLMRealmConfiguration.mm
```diff
@@ -152,15 +151,21 @@ - (NSString *)description {
 }
 
 - (void)setInMemoryIdentifier:(NSString *)inMemoryIdentifier {
-    if ((_inMemoryIdentifier = inMemoryIdentifier)) {
-        _path = nil;
+    if (inMemoryIdentifier.length == 0) {
+        @throw RLMException(@"In-memory identifier must not be empty");
     }
+
+    _inMemoryIdentifier = inMemoryIdentifier;
```
bdash: Should we be copying `inMemoryIdentifier` in case we're passed a mutable string?

Realm/RLMRealmConfiguration.mm
```diff
@@ -152,15 +151,21 @@ - (NSString *)description {
...
 }
 
 - (void)setPath:(NSString *)path {
-    if ((_path = path)) {
-        _inMemoryIdentifier = nil;
+    if (path.length == 0) {
+        @throw RLMException(@"Realm path must not be empty");
     }
+
+    _path = path;
```
bdash: Ditto about copying in case we're passed a mutable string.

bdash:
:+1:

PR #2666, opened 2015-10-13
Have CocoaPods set APPLICATION_EXTENSION_API_ONLY=YES when building our frameworks

bdash:
This avoids linker warnings for application extensions that link against either framework.

### DIFF
--- a/Realm.podspec
+++ b/Realm.podspec
@@ -41,7 +41,8 @@ Pod::Spec.new do |s|
   s.source_files            = 'Realm/*.{m,mm}', 'Realm/ObjectStore/*.cpp'
   s.header_mappings_dir     = 'include'
   s.pod_target_xcconfig     = { 'CLANG_CXX_LANGUAGE_STANDARD' => 'compiler-default',
-                                'OTHER_CPLUSPLUSFLAGS' => '-std=c++1y $(inherited)' }
+                                'OTHER_CPLUSPLUSFLAGS' => '-std=c++1y $(inherited)',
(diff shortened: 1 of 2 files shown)


jpsim:
ah, nice catch! :+1:

bdash:
Failure on CI is unrelated.

kristofer:
I'll look forward to the next release.

segiddins:
CocoaPods should handle setting the build setting for pods linked against extension targets, so this should be unnecessary?

bdash:
@segiddins, Are you saying that CocoaPods changes the Xcode configuration settings of targets to set `APPLICATION_EXTENSION_API_ONLY=YES` based on its analysis of the types of user targets that link against pods? That seems like it'd handle some cases, but you'd still end up with linker warnings unless you rerun CocoaPods after adding a new extension target to your project. There doesn't seem to be any downside for Realm to always enable this setting given that we know Realm and RealmSwift to use only extension-safe APIs.

segiddins:
@bdash correct, although the fact that y'all thought it necessary would imply to me that there's a bug in CocoaPods.

jpsim:
> CocoaPods should handle setting the build setting for pods linked against extension targets, so this should be unnecessary?

I looked at CP's GitHub issues when reviewing this and that feature was still a work in progress. So even once that's merged in CP, this change will work with older versions of CP too.

segiddins:
It's implemented in cocoapods master, it's been in there for at least a month?

jpsim:
This is the PR I was referring to: CocoaPods/CocoaPods#4321, not that it matters because of my previous point about support for older CP versions.

PR #2667, opened 2015-10-13
[CONTRIBUTING] Propose to use a shell script to compile a list of relevant versions

mrackwitz:
tbd. Propose to use a shell script to compile a list of relevant versions and install paths.


jpsim:
I don't suspect that many people read this, but it certainly can't hurt. :+1:

Issue #2655, opened 2015-10-09
Exception: RLMException Missing property value


lucianocn:
`RLMException Missing property value`
It's a good idea to RLMException log also the property and the value that is causing the crash, because it's very hard to debug just with this information.

<img width="953" alt="screen shot 2015-10-09 at 20 51 46" src="https://cloud.githubusercontent.com/assets/3828136/10407889/a496a124-6ec7-11e5-84b8-85a7857b1f64.png">

```
Thread : Fatal Exception: RLMException
0  CoreFoundation                 0x0000000186424f5c __exceptionPreprocess
1  libobjc.A.dylib                0x000000019acfff80 objc_exception_throw
2  Realm                          0x0000000100503034 
3  Realm                          0x00000001004f9dd4 
4  Realm_JSON                     0x000000010072e9a4 
5  Get In Manager                 0x0000000100101418 __hidden#849_ (__hidden#1716_:188)
6  Socket_IO_Client_Swift         0x00000001007ea4e4 
7  Socket_IO_Client_Swift         0x00000001007ddb58 
8  Socket_IO_Client_Swift         0x00000001007ea020 
9  Socket_IO_Client_Swift         0x00000001007e8d10 
10 libdispatch.dylib              0x000000019b4ed7b0 _dispatch_call_block_and_release
11 libdispatch.dylib              0x000000019b4ed770 _dispatch_client_callout
12 libdispatch.dylib              0x000000019b4f2e20 _dispatch_main_queue_callback_4CF
13 CoreFoundation                 0x00000001863dc258 __CFRUNLOOP_IS_SERVICING_THE_MAIN_DISPATCH_QUEUE__
14 CoreFoundation                 0x00000001863da0c0 __CFRunLoopRun
15 CoreFoundation                 0x0000000186308dc0 CFRunLoopRunSpecific
16 GraphicsServices               0x0000000191284088 GSEventRunModal
17 UIKit                          0x000000018b9e2f44 UIApplicationMain
18 My App                         0x0000000100119148 main (__hidden#3215_:16)
19 libdyld.dylib                  0x000000019b51e8b8 start
```

jpsim:
Thanks for bringing this to our attention, @lucianocn. We were including the property and class names in the exception's `userInfo` property, but that wasn't very useful in this context. We've just merged a change that will print the property and class names in the exception message. You can expect this change to be available in our next major release.

lucianocn:
@jpsim thank you!

PR #2668, opened 2015-10-13
Print property and class names when assigning nil to a required property

jpsim:
Addresses #2655. /cc @tgoyne @bdash


tgoyne:
:+1:

bdash:
:+1:

PR #2669, opened 2015-10-13
Validate that the encryption key matches for cached realms

tgoyne:
This was previously handled by not allowing registering an encryption key for already-open paths, but that doesn't do anything about supplying a different encryption key in the RLMRealmConfiguration directly.


bdash:
:+1:

PR #2671, opened 2015-10-13
Use the default realm path in testInitCustomClassList()

tgoyne:
The tests do a good job of cleaning up the current default realm path, but not the default default realm path, so using the latter can result in files lingering between tests and causing problems. Should fix the file format unsupported errors on CI.


jpsim:
:+1:

PR #2675, opened 2015-10-14
Fix failing install with CocoaPods 0.39 to case-sensitive file system.

kishikawakatsumi:
This PR fixes the issue that fails to build when Realm is installed to case-sensitive file system using CocoaPods 0.39 due to following error:

```
...Pods/Realm/Realm/ObjectStore/object_store_exceptions.cpp:19:10: 'object_store_exceptions.hpp' file not found
```

screen shot 2015-10-14 at 19 13 51

Because two different directories (`Ream` and `realm`) are copied to `include` directory. Each directory has different headers.

```
include
├── Realm
│   ├── RLMAccessor.h
│   ├── RLMAnalytics.hpp
│   ├── ...
│   ├── object_store_exceptions.hpp
│   └── property.hpp
├── realm
│   ├── alloc.hpp
│   ├── alloc_slab.hpp
│   ├── ...
│   ├── version.hpp
│   └── views.hpp
└── realm.hpp
```

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -674,6 +674,7 @@ case "$COMMAND" in
           # CocoaPods doesn't support multiple header_mappings_dir, so combine
           # both sets of headers into a single directory
           rm -rf include
+          cp -R core/include/realm core/include/Realm
           cp -R core/include include
           mkdir -p include/Realm
           cp Realm/*.{h,hpp} include/Realm
--- a/build.sh
+++ b/build.sh
@@ -674,7 +674,9 @@ case "$COMMAND" in
           # CocoaPods doesn't support multiple header_mappings_dir, so combine
           # both sets of headers into a single directory
           rm -rf include
-          cp -R core/include/realm core/include/Realm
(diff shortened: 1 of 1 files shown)


tgoyne:
This will fail on case-insensitive filesystems.

kishikawakatsumi:
I see. Added check if a directory does or does not exist.

I did not understand in where the setup script is executed. It is executed in `~/Library/Caches/CocoaPods/` directory. So no error occurred when I execute `pod install` in the external disk.

mrackwitz:
You can disable the cache by setting the environment variable `COCOAPODS_SKIP_CACHE=YES`. This should do the job.

If that should work, I'd add a comment, because just from the code I don't understand what it is thought for.

kishikawakatsumi:
@mrackwitz Totally agreed. I addressed feedback.

jpsim:
Looks good to me. :+1:

jpsim:
I believe this can be merged.

PR #2680, opened 2015-10-14
Remove all deprecated things

tgoyne:
I ended up mostly rewriting the migration tests, since a lot of the tests simply didn't make any sense with the new way of doing things, but a large chunk of the migration functionality was sort of incidentally tested along the way in those tests (and nowhere else). The RLMRealm tests weren't as bad, but similarly required a lot of shuffling around due to tests for the removed functionality and tests for remaining things being entangled.

The changes to non-test code should all be straightforward and obvious (and the updated tests pass with none of the non-test changes other than the bug fix in migrateRealm:).

@jpsim @bdash

### DIFF
--- a/Realm/Tests/MigrationTests.mm
+++ b/Realm/Tests/MigrationTests.mm
@@ -39,7 +39,7 @@ static void RLMAssertRealmSchemaMatchesTable(id self, RLMRealm *realm) {
         Table *table = objectSchema.table;
         for (RLMProperty *property in objectSchema.properties) {
             XCTAssertEqual(property.column, table->get_column_index(RLMStringDataWithNSString(property.name)));
-            XCTAssertEqual(property.indexed, table->has_search_index(property.column));
+            XCTAssertEqual(property.indexed || property.isPrimary, table->has_search_index(property.column));
         }
     }
 }
@@ -101,9 +101,6 @@ RLM_ARRAY_TYPE(MigrationObject);
 @interface MigrationTests : RLMTestCase
 @end
 
-#pragma clang diagnostic push
-#pragma clang diagnostic ignored "-Wdeprecated-declarations"
-
 @implementation MigrationTests
 #pragma mark - Helper methods
 - (RLMSchema *)schemaWithObjects:(NSArray *)objects {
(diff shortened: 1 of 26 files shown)


jpsim:
Looks like Realm files aren't being properly cleaned up between tests in Swift 2 now. Would you like me to pick this up, @tgoyne?

tgoyne:
I'm guessing they're pre-existing files from other jobs.

jpsim:
> I'm guessing they're pre-existing files from other jobs.

No.

Realm/Tests/MigrationTests.mm
```diff
@@ -39,7 +39,7 @@ static void RLMAssertRealmSchemaMatchesTable(id self, RLMRealm *realm) {
         Table *table = objectSchema.table;
         for (RLMProperty *property in objectSchema.properties) {
             XCTAssertEqual(property.column, table->get_column_index(RLMStringDataWithNSString(property.name)));
-            XCTAssertEqual(property.indexed, table->has_search_index(property.column));
+            XCTAssertEqual(property.indexed || property.isPrimary, table->has_search_index(property.column));
```
jpsim: why this change?

tgoyne: property.indexed indicates if the property was explicitly marked as indexed by the user, and this function was previously not used on things with primary keys.

jpsim: :+1:

jpsim:
I fixed some issues in Swift 2's `TestCase.swift` and removed the older Objective-C deprecations. I think this is good to go. :+1: 

Feels great to remove dead-ish code!

tgoyne:
Why did you revert c67348b? As I said, the only reason it was initially failing was due to files from previous runs of the PR job, and not because the additional checks for files ending up in the wrong place were wrong.

jpsim:
No, the Swift 2 tests weren't just failing because of previous runs of the PR. I was getting the same errors when running cleanly both on CI and locally.

tgoyne:
They started passing on CI as soon as I added `~/Library/Application Support/io.realm.TestHost` to the set of directories deleted before running verify. Your local machine would also have lingering files in that directory from other things, since on master there are tests which use the wrong path.

jpsim:
I *had* deleted `~/Library/Application Support/io.realm.TestHost`...

tgoyne:
Then you should look into what's writing files to the incorrect path rather than just deleting the check for them.

jpsim:
You're right, the files were being cleaned correctly before I started diving into this. Sorry for being stubborn and wrong about this. I was convinced I had tested this with the TestHost files deleted, but I was wrong. I've removed my commit where I messed around with that stuff.

tgoyne:
It is kinda a problem that leftover files from other jobs could cause test failures, since for ios-device we can't do extra cleanup stuff between jobs. I guess it could delete all .realm files in +setUp?

jpsim:
> I guess it could delete all .realm files in +setUp?

done.

jpsim:
I believe this can be merged.

PR #2684, opened 2015-10-15
[Swift] Fix crash when using optional Int16/Int32/Int64 properties

### LINKED ISSUE #2681: Using optional values for Int64 will cause app to crash
I'm running Realm 0.96, trying the RealmOptional:

```swift
import Foundation
import RealmSwift

class Person : RealmSwift.Object {
    let id = RealmOptional<Int64>(nil)
    let spouseId = RealmOptional<Int64>(nil)
}

let person = Person()
person.id.value = 1 // CRASH
```

The setter try to case the value to AnyObject, where Int64 is not.

Sample project to reproduce the issue: https://www.dropbox.com/s/2xu277zo6y1rise/RealmOptionalSample.zip?dl=0

jpsim:
Fixes #2681. /cc @bdash @tgoyne

### DIFF
--- a/RealmSwift-swift1.2/Tests/ObjectAccessorTests.swift
+++ b/RealmSwift-swift1.2/Tests/ObjectAccessorTests.swift
@@ -269,7 +269,7 @@ class ObjectAccessorTests: TestCase {
         object.optInt16Col.value = Int16.max
         XCTAssertEqual(object.optInt16Col.value!, Int16.max)
         object.optInt16Col.value = nil
-        XCTAssertNil(object.optInt16Col.value)
+        XCTAssert(object.optInt16Col.value == nil)
 
         object.optInt32Col.value = Int32.min
         XCTAssertEqual(object.optInt32Col.value!, Int32.min)
@@ -278,7 +278,7 @@ class ObjectAccessorTests: TestCase {
         object.optInt32Col.value = Int32.max
         XCTAssertEqual(object.optInt32Col.value!, Int32.max)
         object.optInt32Col.value = nil
-        XCTAssertNil(object.optInt32Col.value)
+        XCTAssert(object.optInt32Col.value == nil)
 
         object.optInt64Col.value = Int64.min
         XCTAssertEqual(object.optInt64Col.value!, Int64.min)
@@ -287,7 +287,7 @@ class ObjectAccessorTests: TestCase {
         object.optInt64Col.value = Int64.max
         XCTAssertEqual(object.optInt64Col.value!, Int64.max)
         object.optInt64Col.value = nil
-        XCTAssertNil(object.optInt64Col.value)
+        XCTAssert(object.optInt64Col.value == nil)
 
         object.optFloatCol.value = -FLT_MAX
         XCTAssertEqual(object.optFloatCol.value!, -FLT_MAX)
--- a/RealmSwift-swift1.2/Tests/ObjectAccessorTests.swift
+++ b/RealmSwift-swift1.2/Tests/ObjectAccessorTests.swift
@@ -262,6 +262,33 @@ class ObjectAccessorTests: TestCase {
         object.optIntCol.value = nil
         XCTAssertNil(object.optIntCol.value)
 
+        object.optInt16Col.value = -1
+        XCTAssertEqual(object.optInt16Col.value!, -1)
+        object.optInt16Col.value = 0
+        XCTAssertEqual(object.optInt16Col.value!, 0)
+        object.optInt16Col.value = 1
+        XCTAssertEqual(object.optInt16Col.value!, 1)
+        object.optInt16Col.value = nil
+        XCTAssert(object.optInt16Col.value == nil)
+
+        object.optInt32Col.value = -1
+        XCTAssertEqual(object.optInt32Col.value!, -1)
+        object.optInt32Col.value = 0
+        XCTAssertEqual(object.optInt32Col.value!, 0)
+        object.optInt32Col.value = 1
+        XCTAssertEqual(object.optInt32Col.value!, 1)
+        object.optInt32Col.value = nil
+        XCTAssert(object.optInt32Col.value == nil)
+
+        object.optInt64Col.value = -1
+        XCTAssertEqual(object.optInt64Col.value!, -1)
+        object.optInt64Col.value = 0
+        XCTAssertEqual(object.optInt64Col.value!, 0)
+        object.optInt64Col.value = 1
+        XCTAssertEqual(object.optInt64Col.value!, 1)
+        object.optInt64Col.value = nil
+        XCTAssert(object.optInt64Col.value == nil)
+
         object.optFloatCol.value = 20
         XCTAssertEqual(object.optFloatCol.value!, 20 as Float)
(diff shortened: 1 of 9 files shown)


RealmSwift-swift1.2/Tests/ObjectAccessorTests.swift
```diff
@@ -262,6 +262,33 @@ class ObjectAccessorTests: TestCase {
...
+        XCTAssertEqual(object.optInt32Col.value!, 1)
+        object.optInt32Col.value = nil
+        XCTAssert(object.optInt32Col.value == nil)
+
+        object.optInt64Col.value = -1
+        XCTAssertEqual(object.optInt64Col.value!, -1)
+        object.optInt64Col.value = 0
+        XCTAssertEqual(object.optInt64Col.value!, 0)
+        object.optInt64Col.value = 1
+        XCTAssertEqual(object.optInt64Col.value!, 1)
```
mrackwitz: What about testing with `INT64_MAX` here instead of `1` to make sure that the boxing gone right?

tgoyne: Both min and max of each size would be a good idea (the smaller sizes are less likely to be wrong, but still could be...).

jpsim:
Feedback addressed.

bdash:
It'd be great if in the future commit messages contained a high-level explanation of how the changes achieve what they are doing, and why they're necessary. It'd make reviewing much easier.

jpsim:
> It'd be great if in the future commit messages contained a high-level explanation of how the changes achieve what they are doing, and why they're necessary. It'd make reviewing much easier.

Duly noted. I'll make more of an effort on that in the future. Anything in particular that you'd like clarified here for now? Otherwise, any further feedback?

bdash:
A few different questions:
1. Why was the old code crashing?
2. In what manner was it crashing? Was it throwing an exception, hitting an assertion failure, segfaulting, etc?
3. What is the new code doing differently that addresses the problem?

jpsim:
> Why was the old code crashing?

`RealmOptional` expects to be specialized with a type that's implicitly convertible to `AnyObject`, which works for `Int` but not its explicitly-sized variants.

```swift
Int(0) as AnyObject    // => 0
Int16(0) as? AnyObject // => nil
Int32(0) as? AnyObject // => nil
Int64(0) as? AnyObject // => nil
```

> In what manner was it crashing? Was it throwing an exception, hitting an assertion failure, segfaulting, etc?

It was triggering a Swift force-cast failure at runtime, because the value couldn't be force-casted to `AnyObject`.

> What is the new code doing differently that addresses the problem?

The new code is explicitly checking for the types which we know aren't force-castable to `AnyObject` in the setters and getters, and converting to a type that can be casted to `AnyObject` (in this case `NSNumber`).

jpsim:
Filed as [rdar://23172237](http://www.openradar.me/23172237).

bdash:
Thanks for the details, @jpsim. That makes the change much easier to follow.

A protocol-based approach could keep `RealmObject`s semantics from being obscured by the type conversions. Something like:

```diff
diff --git a/RealmSwift-swift2.0/Optional.swift b/RealmSwift-swift2.0/Optional.swift
index 8bf76ea..67291f3 100644
--- a/RealmSwift-swift2.0/Optional.swift
+++ b/RealmSwift-swift2.0/Optional.swift
@@ -18,8 +18,14 @@
 
 import Realm
 
+public protocol RealmOptionalTypeConversionsToAnyObject {
+    func asObject() -> AnyObject
+    static func fromObject(_: AnyObject?) -> Self?
+}
+
 /// Types that can be represented in a `RealmOptional`.
-public protocol RealmOptionalType {}
+public protocol RealmOptionalType: RealmOptionalTypeConversionsToAnyObject {}
+
 extension Int: RealmOptionalType {}
 extension Int16: RealmOptionalType {}
 extension Int32: RealmOptionalType {}
@@ -39,29 +45,10 @@ public final class RealmOptional<T: RealmOptionalType>: RLMOptionalBase {
     /// The value this optional represents.
     public var value: T? {
         get {
-            if T.self is Int16.Type {
-                return ((underlyingValue as! NSNumber?)?.longValue).map { Int16($0) } as! T?
-            } else if T.self is Int32.Type {
-                return ((underlyingValue as! NSNumber?)?.longValue).map { Int32($0) } as! T?
-            } else if T.self is Int64.Type {
-                return ((underlyingValue as! NSNumber?)?.longLongValue).map { Int64($0) } as! T?
-            }
-            return underlyingValue as! T?
+            return T.fromObject(underlyingValue)
         }
         set {
-            guard let unwrappedValue = newValue else {
-                underlyingValue = nil
-                return
-            }
-            if let int64Value = unwrappedValue as? Int64 {
-                underlyingValue = NSNumber(longLong: int64Value)
-            } else if let int32Value = unwrappedValue as? Int32 {
-                underlyingValue = NSNumber(long: Int(int32Value))
-            } else if let int16Value = unwrappedValue as? Int16 {
-                underlyingValue = NSNumber(long: Int(int16Value))
-            } else {
-                underlyingValue = unwrappedValue as! AnyObject
-            }
+            underlyingValue = newValue.map { $0.asObject() }
         }
     }
 
@@ -75,3 +62,23 @@ public final class RealmOptional<T: RealmOptionalType>: RLMOptionalBase {
         self.value = value
     }
 }
+
+public extension RealmOptionalTypeConversionsToAnyObject {
+    func asObject() -> AnyObject { return self as! AnyObject }
+    static func fromObject(object: AnyObject?) -> Self? { return object as! Self? }
+}
+
+extension Int16: RealmOptionalTypeConversionsToAnyObject {
+    public func asObject() -> AnyObject { return NSNumber(long: Int(self)) }
+    public static func fromObject(object: AnyObject?) -> Int16? { return object.map { Int16($0.longValue) } }
+}
+
+extension Int32: RealmOptionalTypeConversionsToAnyObject {
+    public func asObject() -> AnyObject { return NSNumber(long: Int(self)) }
+    public static func fromObject(object: AnyObject?) -> Int32? { return object.map { Int32($0.longValue) } }
+}
+
+extension Int64: RealmOptionalTypeConversionsToAnyObject {
+    public func asObject() -> AnyObject { return NSNumber(long: Int(self)) }
+    public static func fromObject(object: AnyObject?) -> Int64? { return object.map { Int64($0.longValue) } }
+}
```

It's more code, but a little more self-documenting. Thoughts?

jpsim:
I'd really rather not do that because that's really an implementation detail, but because `RealmOptionalType` is a public protocol, it would have to leak those details. We can prevent HTML documentation from being generated, but you get the idea that this really expands the scope. We then can't use our `SwiftObjectWithNonRealmOptionalType` test class because we can't provide public declarations of a generic type from a different module....

If your concern is with better documenting what this code is doing and why it's needed, I think we'd be better off making explicit, documented, private functions for `realmOptionalToAnyObject(_:)` and `anyObjectToRealmOptional(_:)`.

jpsim:
Hopefully that made things clearer.

bdash:
I see what you mean with regards to commingling implementation details with the public protocol. Your change achieves the same improvement in clarity without that downside. :+1:

Issue #2679, opened 2015-10-14
Parallelize CI PR job's device tests


jpsim:
We should be running objc & swift device tests in parallel, on any Jenkins worker labelled `ios_devices`.

jpsim:
Done in #2687.

PR #2687, opened 2015-10-16
[build.sh] split device tests

jpsim:
When this is ready to merge, we'll have to split the Jenkins job. We should merge #2656 before this to actually benefit from the parallelization. /cc @bdash @tgoyne

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -515,7 +515,8 @@ case "$COMMAND" in
         sh build.sh verify-ios-dynamic-debug
         sh build.sh verify-ios-swift
         sh build.sh verify-ios-swift-debug
-        sh build.sh verify-ios-device
+        sh build.sh verify-ios-device-objc
+        sh build.sh verify-ios-device-swift
         sh build.sh verify-watchos
         ;;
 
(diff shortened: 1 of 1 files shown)


jpsim:
Addresses #2679.

build.sh
```diff
@@ -466,13 +468,21 @@ case "$COMMAND" in
...
-            test_ios_devices xcrealmswift "RealmSwift" "$CONFIGURATION" || failed=1
-        fi
+        sh build.sh test-ios-devices-objc || failed=1
+        sh build.sh test-ios-devices-swift || failed=1
         exit $failed
         ;;
 
+    "test-ios-devices-objc")
+        test_ios_devices xcrealm "iOS Device Tests" "$CONFIGURATION" || exit 1
+        exit 0
```
bdash: The `|| exit 1` and `exit 0` could be combined into a single `exit $?`.

build.sh
```diff
@@ -466,13 +468,21 @@ case "$COMMAND" in
 
     "test-ios-devices")
         failed=0
-        test_ios_devices xcrealm "iOS Device Tests" "$CONFIGURATION" || failed=1
-        if [ $REALM_SWIFT_VERSION != '1.2' ]; then
-            test_ios_devices xcrealmswift "RealmSwift" "$CONFIGURATION" || failed=1
-        fi
+        sh build.sh test-ios-devices-objc || failed=1
+        sh build.sh test-ios-devices-swift || failed=1
         exit $failed
```
bdash: It could be marginally cleaner to write this as:

    failed=0
    trap "failed=1" ERR
    sh build.sh test-ios-devices-objc
    sh build.sh test-ios-devices-swift
    exit $failed

bdash:
:+1:

build.sh
jpsim: Good call. We could be doing this all over build.sh.

jpsim:
CI job passed "in spirit" (failed due to #2119).

Issue #2685, opened 2015-10-15
Provide a dynamic method to get List<DynamicObject> properties


tom-sparo:
Say I have a realm object:

```swift
class TestObject : Object {
    let otherObjects = List<TestObject>()
}
```

If I want to use key value coding:
```swift
let testObject = TestObject()
let aList = testObject.valueForKey("otherObjects")!
aList is List<Object> // false
aList is List<TestObject> // true
```
I have to know the exact type of that relationship.  This makes certain reflection-based approaches sort of difficult, for instance if I want to blindly walk over an object tree performing some action.

Is there a better way to get the KVC to behave, to get generics to behave, or to walk over an object tree?  I see that I can reflect on the objectSchema and get a string representation of the List'd field... but it's a String and I remember NSClassFromString being kind of a bear.  So if I could get the type in the List from the objectSchema that's another option.

Sorry this is sorta hacky!

Just for reference, is this an effect of List<> generics?  Because with normal arrays in swift this works:

```swift
class A: NSObject {}

class B: A {}

class Something: NSObject {
    let array: [B] = [B(), B()]
}

let arr = Something().valueForKey("array")
arr is Array<A> // true
arr is Array<B> // true
arr is Array<Int> // false
```

jpsim:
> Is there a better way to get the KVC to behave, to get generics to behave, or to walk over an object tree?

I think our dynamic API would be better suited for this kind of work. If you can represent the objects you'll be acting on as a query, you could then use [`Realm.dynamicObjects()`](https://github.com/realm/realm-cocoa/blob/v0.96.0/RealmSwift-swift2.0/Realm.swift#L411-L428).

Although we officially support our dynamic API and it is technically "public", we don't generally include it in most of our online documentation because it's really only useful in specific scenarios, but it sounds like yours applies.

If you need something we don't have in our dynamic API, please let us know and we'll gladly consider adding it. For example an `Object.dynamicList(listProperty: String) -> List<DynamicObject>` method in your case.

> Just for reference, is this an effect of List<> generics?

I don't know off the top of my head, but I can dig deeper into this if you'd like.

tom-sparo:
I did some more digging and it turns out that Array/Set/Dict all have different semantics when it comes to the `is` operator :rage: .  Their `is` operator does what looks like a protocol match (IMO the expected behavior), whereas the `is` operator on generics that we can write does a strict match.  How frustrating!  Hadn't noticed/forgot about it.

[...]

PR #2696, opened 2015-10-16
[Swift] Get list of DynamicObject

tom-sparo:
Hey @jpsim -- this pull request implements [#2685](https://github.com/realm/realm-cocoa/issues/2685)

Wasn't sure what style you all wanted for documentation/testing so I just guestimated based on code elsewhere (sorta shooting from the hip!)

As a user I can already get this functionality myself by using an unsafeBitCast on the result of KVC, but it's nice to wrap away these types of casts so the user doesn't feel like they're doing something wrong.

Let me know what you think!

### DIFF
--- a/RealmSwift-swift1.2/Object.swift
+++ b/RealmSwift-swift1.2/Object.swift
@@ -200,16 +200,16 @@ public class Object: RLMObjectBase, Equatable, Printable {
 
     Returns a List of DynamicObjects for a property name
 
-    - warning: This method is useful only in specialized circumstances
+    :warning: This method is useful only in specialized circumstances
 
-    - parameter propertyName: The name of the property to get a List<DynamicObject>
+    :param: propertyName The name of the property to get a List<DynamicObject>
 
-    - returns: A List of DynamicObjects
+    :returns: A List of DynamicObjects
 
     :nodoc:
     */
     public func dynamicList(propertyName: String) -> List<DynamicObject> {
-        return unsafeBitCast(self.listForProperty(RLMValidatedGetProperty(self, propertyName)), List<DynamicObject>.self)
+        return unsafeBitCast(listForProperty(RLMValidatedGetProperty(self, propertyName)), List<DynamicObject>.self)
     }
 
     // MARK: Private functions
--- a/RealmSwift-swift1.2/Object.swift
+++ b/RealmSwift-swift1.2/Object.swift
@@ -191,6 +191,27 @@ public class Object: RLMObjectBase, Equatable, Printable {
         }
     }
 
+    // MARK: Dynamic list
+
+    /**
+    This method is useful only in specialized circumstances, for example, when building
+    components that integrate with Realm. If you are simply building an app on Realm, it is
+    recommended to use instance variables or cast the KVC returns.
+
+    Returns a List of DynamicObjects for a property name
+
+    - warning: This method is useful only in specialized circumstances
+
+    - parameter propertyName: The name of the property to get a List<DynamicObject>
+
+    - returns: A List of DynamicObjects
+
+    :nodoc:
+    */
+    public func dynamicList(propertyName: String) -> List<DynamicObject> {
+        return unsafeBitCast(self.listForProperty(RLMValidatedGetProperty(self, propertyName)), List<DynamicObject>.self)
+    }
+
     // MARK: Private functions
 
     // FIXME: None of these functions should be exposed in the public interface.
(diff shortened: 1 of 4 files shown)


jpsim:
Thanks for the PR :sparkles:! It always feels awesome to see users contributing. Here are a few points of feedback:

1. All Swift changes should generally be applied to both Swift 1.2 and Swift 2.0 (there are two versioned RealmSwift directories in the root of the repo).
2. Tests should access members of the `List` rather than just check the empty case, since that's not comprehensive enough.

Thanks again.

tom-sparo:
Added in change to the Swift 1.2 branch and added in more robust testing -- let me know if you want more than that!

Also can't see what the failing checks are so if there are compile errors or test errors let me know so I can fix them.  I assumed there's no way for me to compile the repo to check myself because Core isn't open source, but if there's a way to compile without it I missed let me know so I can save you reading time.

jpsim:
We use an internal Jenkins setup for our CI, so we unfortunately can't easily share that for external contributors :cry:. This is something I'd love to address, but simply don't have the time.

But you can run the same gamut of tests we do by running `sh build.sh verify` assuming you have Xcode 6.4 (with iOS 7 simulators) and Xcode 7.0.1 installed on your system.

The whole set of tests take a really long time to run (we parallelize them on ~12 CI workers!), but you can usually get a good sense of whether or not things are broken by running the following small(er) tests:

```
./build.sh verify-osx
export REALM_SWIFT_VERSION=1.2
./build.sh verify-osx-swift
export REALM_SWIFT_VERSION=2.0
./build.sh verify-osx-swift
```

In your case, the Swift 1.2 tests are failing:

```
Testing failed:
	Consecutive statements on a line must be separated by ';'
	Consecutive statements on a line must be separated by ';'
	Use of unresolved identifier 'try'
	'List<SwiftStringObject>' does not have a member named 'appendContentsOf'
	Use of unresolved identifier 'try'
```

tom-sparo:
Ah great!  That's at least better than no testing at all.  Both my local tests (for swift 1.2 and swift 2.0) passed so I'll have to rely on you all again to paste errors :disappointed:

RealmSwift-swift1.2/Object.swift
```diff
@@ -191,6 +191,27 @@ public class Object: RLMObjectBase, Equatable, Printable {
...
+    - warning: This method is useful only in specialized circumstances
+
+    - parameter propertyName: The name of the property to get a List<DynamicObject>
+
+    - returns: A List of DynamicObjects
+
+    :nodoc:
+    */
+    public func dynamicList(propertyName: String) -> List<DynamicObject> {
+        return unsafeBitCast(self.listForProperty(RLMValidatedGetProperty(self, propertyName)), List<DynamicObject>.self)
```
jpsim: `self.` can probably be omitted here.

RealmSwift-swift1.2/Tests/ObjectTests.swift
```diff
@@ -284,4 +284,22 @@ class ObjectTests: TestCase {
             self.setAndTestAllTypes(setter, getter: getter, object: persistedObject)
         }
     }
+
+    func testDynamicList() {
+        let realm = Realm()
+        let arrayObject = SwiftArrayPropertyObject()
+        let str1 = SwiftStringObject()
+        let str2 = SwiftStringObject()
+        arrayObject.array.append(str1)
```
jpsim: `arrayObject.array.extend([str1, str2])`?

RealmSwift-swift1.2/Object.swift
```diff
@@ -191,6 +191,27 @@ public class Object: RLMObjectBase, Equatable, Printable {
         }
     }
 
+    // MARK: Dynamic list
+
+    /**
+    This method is useful only in specialized circumstances, for example, when building
```
jpsim: This comment is written in the Swift 2.0 documentation style. Please use `:warning:`, `:param: propertyName` and `:returns:`.

jpsim:
Thanks for the PR fixes, @tom-sparo! I think this could be merged after you address my last round of (nitpicky) feedback.

Our CI jobs are seeing some device test failures at the moment because we're adding some additional iOS devices and still working out some kinks, so we may merged this despite GitHub showing a red 'x'.

tom-sparo:
Fixed all feedback!

jpsim:
This looks all good to me, how about for you @bigfish24?

bigfish24:
:+1: thanks for the contribution @tom-sparo!

PR #2699, opened 2015-10-16
Make RLMException a little less awkward to use

tgoyne:
Remove the error-specific userInfo from all exceptions (there were only a few left) and move the relevant information to the exception message, since it's quite clear that no one actually looks at the userInfo.

Make RLMException take varargs and perform string formatting since nearly every usage of it was of the form `RLMException([NSString stringWithFormat:...])` (or something equivalent split over more lines), which was needlessly verbose.


bdash:
:+1:

jpsim:
Nice :+1:

Issue #2701, opened 2015-10-17
CocoaPods fails on first run


jpsim:
Since #2627, due to our dual bitcode scheme.

Issue #2678, opened 2015-10-14
Parallelize installation examples CI job


jpsim:
They currently take a very long time. We'll likely need to update `examples/installation/build.sh`'s `bootstrap` command to only bootstrap the specific example we're testing to get the most benefit out of this.

jpsim:
I did this a while ago

PR #2702, opened 2015-10-17
refactor installation examples to lazily bootstrap in order to parallelize

jpsim:
Should address #2678 but is blocked on us resolving #2701. /cc @bdash

### DIFF
--- a/examples/installation/build.sh
+++ b/examples/installation/build.sh
@@ -43,11 +43,11 @@ xctest() {
     LANG="$2"
     NAME="$3"
     DIRECTORY="$PLATFORM/$LANG/$NAME"
-    PROJ="$DIRECTORY/$NAME.xcodeproj"
-    WORK="$DIRECTORY/$NAME.xcworkspace"
-    CMD="-project $PROJ"
-    if [ -d $WORK ]; then
-        CMD="-workspace $WORK"
+    PROJECT="$DIRECTORY/$NAME.xcodeproj"
+    WORKSPACE="$DIRECTORY/$NAME.xcworkspace"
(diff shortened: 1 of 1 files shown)


examples/installation/build.sh
```diff
@@ -30,152 +28,121 @@ EOF
...
-xctest_osx() {
-    XCODE_COMMAND="$@"
-    xcodebuild $XCODE_COMMAND clean build test -sdk macosx || exit 1
+xctest() {
+    PLATFORM="$1"
+    LANG="$2"
+    NAME="$3"
+    DIRECTORY="$PLATFORM/$LANG/$NAME"
+    PROJ="$DIRECTORY/$NAME.xcodeproj"
+    WORK="$DIRECTORY/$NAME.xcworkspace"
```
bdash: `PROJECT` and `WORKSPACE` would be fine.

bdash:
Seems fine. Do you expect this to have issues on CI?

jpsim:
We still need to create a matrix job on CI for all of these to run in parallel. Other than that and #2706, I don't expect this to cause issues.

jpsim:
CI failures unrelated, filed a radar for the `xcodebuild` crash: http://www.openradar.me/23172400

bdash:
:+1:

It'd be good to confirm that the `xcodebuild` crash is transient and not a bug that this refactoring brings to light.

jpsim:
> It'd be good to confirm that the xcodebuild crash is transient and not a bug that this refactoring brings to light.

I can confirm that because the only file modified by this PR is never referenced in CI's PR jobs. Only in releasability.

PR #2707, opened 2015-10-19
Honor the schema version set in the configuration in `+[RLMRealm migrateRealm:]`

tgoyne:


bdash:
:+1:

jpsim:
This seems to have broken all the Swift migration tests.

bdash:
The deprecated `migrateRealm(path:encryptionKey:)` isn't setting a schema version on its configuration. All of the tests are using this legacy API rather than the new `migrateRealm(configuration:)`.

tgoyne:
Now passes all tests with a few more functional changes.

jpsim:
:+1:

PR #2709, opened 2015-10-19
Remove testNotificationsForChangesWhileSuspended

### LINKED ISSUE #2119: testNotificationsForChangesWhileSuspended occasionally fails

jpsim:
Because -[NSTask suspend] is asynchronous and there doesn't seem to be a
reliable way to check this programmatically, and this spuriously fails on CI on
a regular basis.

The intent of this test is also tested elsewhere in the interprocess tests.

Resolves #2119. /cc @tgoyne

### DIFF
--- a/Realm/Tests/InterprocessTests.m
+++ b/Realm/Tests/InterprocessTests.m
@@ -303,29 +303,4 @@
     }
 }
 
-- (void)testNotificationsForChangesWhileSuspended {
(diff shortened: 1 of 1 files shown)


tgoyne:
:+1:

PR #2714, opened 2015-10-21
[Podspec] add header files from ObjectStore's impl directory

jpsim:
Realm doesn't compile with CocoaPods without this (see https://ci.realm.io/job/Cocoa%20Installation%20Examples/12/). /cc @bdash


bdash:
:+1:

Should one of the CocoaPods tests be hooked up to objc_pr? We seem to miss things in this area until they're merged to master.

PR #2724, opened 2015-10-21
Enable and fix a few additional warnings

tgoyne:
While testing compilation of the objectstore stuff in GCC I hit a few warnings for class/struct mismatch, and there's really no reason to have that warning off.


bdash:
:+1:

Failure on CI appears to be unrelated.

PR #2733, opened 2015-10-22
Switch from Swift 2.0 to Swift 2.1

tgoyne:
7.1 still needs to be installed on the CI workers and there's a bunch of places there that need to be switched to 2.1.


bdash:
Everything is looking good for me locally. Sadly `verify-cocoapods` is failing in CI when doing `test-ios-swift-cocoapods`. From what I can see, it builds the Realm target but never attempts to build the RealmSwift target. This leads to failures when the example project attempts to `import RealmSwift`. I'm not sure how to proceed given that this succeeds locally.

tgoyne:
The CI machines have cocoapods 0.38.2 rather than 0.39.0, which seems like it could be relevant.

Issue #2732, opened 2015-10-22
Realm Swift 0.96.0-1: ignoredProperties no longer working


mlusas:
After updating to 0.96+, my ignoredProperties are no longer ignored, and are now included in the Realm database.

**After updating, I now receive an error of:**
Terminating app due to uncaught exception 'NSInvalidArgumentException', reason: '-[_NSZeroData setObject:]: unrecognized selector sent to instance <instanceID>

The instance refers to the NSData var I'm passing into my model.  Has anyone been able to fix this, and get ignored properties working again?  Here is my model for that object.  The commented out code are variations I tried to due documented variations in ignoring properties; none of these worked.

**Realm Model:**
```
class ImageFile: Object {
    dynamic var imageID:String = ""
    dynamic var imageFileID:String = ""
    dynamic var imageType:String = ""
    dynamic var selectedImageFile:Bool = false
    dynamic var folder:String = ""
    dynamic var filename:String = ""
    dynamic var scale:Int = 0
    dynamic var height:Float = 0
    dynamic var width:Float = 0
    dynamic var created:Int = 0
    dynamic var updated:Int = 0
    dynamic var expires:Int = 0
//    dynamic var imageData:NSData? = NSData()
    var imageData:NSData? = NSData()
    
    override static func primaryKey() -> String {
        return "imageFileID"
    }
    
    override static func indexedProperties() -> [String] {
        return [
            "imageID",
            "imageType",
            "selectedImageFile",
            "scale"
        ]
    }
    
//    override static func ignoredProperties() -> [String] {
//        return ["imageData"]
//    }
    
    override class func ignoredProperties() -> [String] {
        return ["imageData"]
    }
}
```

kishikawakatsumi:
@mlusas Thank you for reporting this. I've reproduced the same issue. It seems a bug. We'll have a fix out shortly.

mlusas:
@kishikawakatsumi Wonderful, thank you for the quick response.

kishikawakatsumi:
Fixed #2734

PR #2734, opened 2015-10-22
Fix an issue Swift optional properties no longer ignored

kishikawakatsumi:
This PR to fix the problem that the properties would be added to the schema even marked as ignored if the properties are optional.

It will happen only on Swift optional properties. 

Related issue: https://github.com/realm/realm-cocoa/issues/2732

### DIFF
--- a/Realm/RLMObjectSchema.mm
+++ b/Realm/RLMObjectSchema.mm
@@ -204,6 +204,9 @@ using namespace realm;
             property.optional = false;
         }
         [optionalProperties enumerateKeysAndObjectsUsingBlock:^(NSString *propertyName, NSNumber *propertyType, __unused BOOL *stop) {
+            if ([ignoredProperties containsObject:propertyName]) {
+                return;
+            }
             NSUInteger existing = [propArray indexOfObjectPassingTest:^BOOL(RLMProperty *obj, __unused NSUInteger idx, __unused BOOL *stop) {
                 return [obj.name isEqualToString:propertyName];
             }];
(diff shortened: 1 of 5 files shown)


kishikawakatsumi:
retest this please

Realm/RLMObjectSchema.mm
```diff
@@ -213,7 +213,7 @@ + (NSArray *)propertiesForClass:(Class)objectClass isSwift:(bool)isSwiftClass {
                 property.optional = true;
             }
             if (auto type = RLMCoerceToNil(propertyType)) {
-                if (existing == NSNotFound) {
+                if (existing == NSNotFound && ![ignoredProperties containsObject:propertyName]) {
```
bdash: Why did you choose to perform the check at this point? From my reading of this method, the natural point would be to do this check as the first thing inside the block we give to `-enumerateKeysAndObjectsUsingBlock:`, before we do any other work for this property. I believe the two locations will give the same behavior, but it seems confusing to do work related to this property before checking whether it is ignored.

kishikawakatsumi: It makes sense. I moved the check to the first of the block https://github.com/realm/realm-cocoa/commit/036cb4abee37d75de536f35e8c38a7c15f439485. Is this what you mean?

bdash:
:+1:

This seems like a good candidate for a 0.96.2 release.

mrackwitz:
:+1:

kishikawakatsumi:
CI failure is unrelated (documentation).

macvinod:
I have this kind of property how can i ignore this property
 var photos = Array<RLMFile>()

jpsim:
@macvinod please open a new issue. Also, `Array` is not an allowed property type for Realm models. See https://realm.io/docs/swift/latest/#models

Issue #2720, opened 2015-10-21
Terminating app due to uncaught exception 'NSInvalidArgumentException', reason: '-[Swift._SwiftDeferredNSArray objectClassName]: unrecognized selector sent to instance 0x608000037720'


bipinvaylu:
Exception logs: 
```
2015-10-21 14:55:34.914 ottodj[59456:1643376] *** Terminating app due to uncaught exception 'NSInvalidArgumentException', reason: '-[Swift._SwiftDeferredNSArray objectClassName]: unrecognized selector sent to instance 0x608000037720'
*** First throw call stack:
(
	0   CoreFoundation                      0x00007fff9268c03c __exceptionPreprocess + 172
	1   libobjc.A.dylib                     0x00007fff8879e76e objc_exception_throw + 43
	2   CoreFoundation                      0x00007fff9268f0ad -[NSObject(NSObject) doesNotRecognizeSelector:] + 205
	3   CoreFoundation                      0x00007fff925d4e24 ___forwarding___ + 1028
	4   CoreFoundation                      0x00007fff925d4998 _CF_forwarding_prep_0 + 120
	5   Realm                               0x00000001003c4952 -[RLMProperty initSwiftPropertyWithName:indexed:property:instance:] + 514
	6   Realm                               0x000000010038c266 +[RLMObjectSchema propertiesForClass:isSwift:] + 1174
	7   Realm                               0x000000010038a5e2 +[RLMObjectSchema schemaForObjectClass:] + 818
	8   Realm                               0x000000010049484f +[RLMSchema registerClasses:count:] + 1871
	9   Realm                               0x0000000100496625 _ZZ25+[RLMSchema sharedSchema]ENK3$_0clEv + 501
	10  Realm                               0x0000000100496425 _ZNSt3__117__call_once_proxyINS_5tupleIJZ25+[RLMSchema sharedSchema]E3$_0EEEEEvPv + 85
	11  libc++.1.dylib                      0x00007fff8b177332 _ZNSt3__111__call_onceERVmPvPFvS2_E + 117
	12  Realm                               0x0000000100494edd +[RLMSchema sharedSchema] + 413
	13  Realm                               0x000000010038680b +[RLMObjectBase sharedSchema] + 43
	14  Realm                               0x0000000100383f34 -[RLMObjectBase init] + 244
	15  Realm                               0x0000000100381bdb -[RLMObject init] + 59
[...]

PR #2743, opened 2015-10-24
[RLMProperty] Fail gracefully when `get<Key>` is defined for array properties

mrackwitz:
Makes it easier to figure out what's wrong in scenarios like described in #2720.
* [ ] Test
* [ ] Actual assert C function which catches exceptions and can be used from Swift

/c @tgoyne @jpsim

### DIFF
--- a/Realm/RLMProperty.mm
+++ b/Realm/RLMProperty.mm
@@ -267,7 +267,17 @@ BOOL RLMPropertyTypeIsNumeric(RLMPropertyType propertyType) {
 
     // convert array types to objc variant
     if ([_objcRawType isEqualToString:@"@\"RLMArray\""]) {
-        _objcRawType = [NSString stringWithFormat:@"@\"RLMArray<%@>\"", [[obj valueForKey:_name] objectClassName]];
+        id arrayValue = [obj valueForKey:_name];
+        if (![arrayValue isKindOfClass:[RLMArray class]]) {
+            NSString *hint = @"";
+            NSString *kvoFirstGuessGetterName = [NSString stringWithFormat:@"get%@", _name.capitalizedString];
+            bool hasCustomKvoGetter = [obj respondsToSelector:NSSelectorFromString(kvoFirstGuessGetterName)];
+            if (hasCustomKvoGetter) {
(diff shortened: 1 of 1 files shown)


bdash:
There are other cases in `-[RLMProperty initSwiftPropertyWithName:indexed:property:instance:]` that look susceptible to this problem.

I wonder if instead of using `-valueForKey:` we could use an approach that accesses only the property we're after, avoiding the risk of calling KVC methods that happen to collide with the property?

mrackwitz:
I wondered about the same question, but as we  use this on other places of the code as well to access the actual value, we would need to change them consequently, I guess. Otherwise it would just fail later.

Alternatively we would need further validations, which will at least increase the likelihood that if users override/intercept KVO accessors, that they still return correct value types. Instead we could also make them more strict and forbid implementations of those methods, which would help to avoid potential ambiguous code.

mrackwitz:
## Summary of Discussions

As this verification is very specific, we should think of whether we could do that in a more general fashion at a different place. There are different options, which I want to layout here, so that we could decide on whether we settle on one of them or don't do any verification like that at all. It all boils down to enforcing our modified `RLMObject` class contract in differentiation to the expectations given from Foundation/Cocoa/`NSObject`. For all options, it could likely make sense to reduce the performance impact by executing them only in the Debug configuration and so not affecting distributed builds of the user's apps.

### Variant A: Check type on initial / every value access

Instead of relying on KVC directly via `valueForKey:`, we could instead encapsulating retrieving the properties value in a method / function. This could then contain further type verification and assert when the type conflicts with the expectations.

### Variant B: Scan & forbid overwritten alternate KVC accessors

As Realm implements custom accessors on `RLMObject` subclasses, we could just completely forbid users to implement alternate KVC methods. (e.g. `is<Key>`, `get<Key>`, …)

## Questions

/c @jpsim @tgoyne @bdash 

* Do these variants make sense and how I described them? How can they be improved?
* Is there any other variant I may have forgotten here?

jpsim:
I'm not super clear on what kind of validation we should be doing, and how to achieve it. Thoughts @tgoyne @bdash?

mrackwitz:
Closing as we all seem to hesitate to introduce further complexity in that direction, given that this isn't among our most critical objectives currently.

PR #2749, opened 2015-10-26
Remove testRealmCreationUncached baseline from the new iOS devices

bdash:
`testRealmCreationUncached` is slow and has high variance, causing frequent failures on CI. Removing the baseline allows the test to be run locally while avoiding the sporadic failures on CI.

/cc @tgoyne


tgoyne:
:+1:

PR #2750, opened 2015-10-26
Fix the path to the core static library that CocoaPods uses

bdash:
It was set to `librealm-ios-no-bitcode.a` prior to `build.sh` picking up the automatic selection of bitcode or no-bitcode library based on the Xcode version. Now that `build.sh` contains that logic, `librealm-ios-no-bitcode.a` does not exist after `build.sh cocoapods-setup` completes.


tgoyne:
:+1:

jpsim:
I was hoping this would fix #2701 but that doesn't seem to be the case.

PR #2758, opened 2015-10-28
Add casts to suppress spurious warnings about values outside of the range of the enum

bdash:
The warnings are being emitted for two option enums that have no explicit "none" member, and for `NSCalendar`'s misuse of an option enum as an argument type where it expects the multiple options to be combined.

/cc @jpsim @tgoyne


tgoyne:
:+1:

PR #2759, opened 2015-10-28
Copy bcsymbolmap files to build products folder when archiving

kishikawakatsumi:
`*.bcsymbolmap` files are necessary to submit an app with including debug symbols. If there are no those files, following error occurs unless uncheck "include app symbols ..."

untitled

```
The archive did not contain 
<DVTFilePath:0x7f91ea526720:'...xcarchive/BCSymbolMaps/....bcsymbolmap'> as expected.
```

This PR makes it include the `*.bcsymbolmap` files to the each framework bundles. Those files are moved to an app bundle when archiving.

See also:
[Missing BCSymbolMap for AppStore Submission - Xcode7B5](https://forums.developer.apple.com/thread/14729)

/cc @jpsim @mrackwitz

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -101,6 +101,10 @@ xc() {
     fi
 }
 
+copy_bcsymbolmap() {
+    find "$1" -name '*.bcsymbolmap' -type f -exec cp {} "$2" \;
+}
+
 build_ios_combined() {
     local scheme="$1"
     local module_name="$2"
@@ -124,6 +128,9 @@ build_ios_combined() {
     if [ -d $iphonesimulator_path/Modules/$module_name.swiftmodule ]; then
       cp $iphonesimulator_path/Modules/$module_name.swiftmodule/* $iphoneos_path/Modules/$module_name.swiftmodule/
     fi
+    
+    # Copy *.bcsymbolmap to .framework for submitting app with bitcode
+    copy_bcsymbolmap "$build_products_path/$config-iphoneos$scope_suffix" "$iphoneos_path"
 
     # Retrieve build products
     clean_retrieve $iphoneos_path $out_path $product_name
@@ -154,6 +161,9 @@ build_watchos_combined() {
     if [ -d $watchsimulator_path/Modules/$module_name.swiftmodule ]; then
       cp $watchsimulator_path/Modules/$module_name.swiftmodule/* $watchos_path/Modules/$module_name.swiftmodule/
     fi
+    
+    # Copy *.bcsymbolmap
+    copy_bcsymbolmap "$build_products_path/$config-watchos$scope_suffix" "$watchos_path"
 
     # Retrieve build products
     clean_retrieve $watchos_path $out_path $product_name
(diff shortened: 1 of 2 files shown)


build.sh
```diff
@@ -111,6 +111,14 @@ xcrealmswift() {
     xc "-project $PROJECT $@"
 }
 
+copy_bcsymbolmap() {
```
tgoyne: Could just be `find "$1" -name '*.bcsymbolmap' -type f -exec cp {} "$2" \;`

kishikawakatsumi: Cool, thanks.

scripts/strip-frameworks.sh
```diff
@@ -60,3 +60,12 @@ for file in $(find . -type f -perm +111); do
     fi
   fi
 done
+
+if [ "$ACTION" = "install" ]; then
+  echo "Copy .bcsymbolmap files to .xcarchive"
+  mv Realm.framework/*.bcsymbolmap RealmSwift.framework/*.bcsymbolmap "${CONFIGURATION_BUILD_DIR}"
+else
+  # Delete *.bcsymbolmap files from framework bundle unless archiving
```
jpsim: this says archiving, but above you're checking for `"$ACTION" = "install"`. Are those actually the same?

jpsim:
Out of curiosity, how much larger does this make the binaries?

kishikawakatsumi:
@jpsim The size of each symbol files is 350KB for Realm and 150KB for RealmSwift.
For iOS has two architecture(armv7, arm64) and for OS X has no bitcode.
So total size of raw files is around 1.4MB. The symbol files are plain text files. So well compressed.

Compressed total size is around 200KB.

scripts/strip-frameworks.sh
kishikawakatsumi: Yes, `ACTION` variable is to be `build`, `clean`, `install`, `installhdrs`, and `installsrc`. I confirmed the variable will be `install` when archiving.

https://developer.apple.com/library/mac/documentation/DeveloperTools/Reference/XcodeBuildSettingRef/1-Build_Setting_Reference/build_setting_ref.html#//apple_ref/doc/uid/TP40003931-CH3-SW45

scripts/strip-frameworks.sh
```diff
@@ -60,3 +60,11 @@ for file in $(find . -type f -perm +111); do
     fi
   fi
 done
+
+if [ "$ACTION" = "install" ]; then
+  echo "Copy .bcsymbolmap files to .xcarchive"
+  find . -name '*.bcsymbolmap' -type f -exec mv {} "${CONFIGURATION_BUILD_DIR}" \;
```
jpsim: @kishikawakatsumi I've generalized this so `strip-frameworks.sh` works for any framework, like the rest of this script. Could you please test this?

jpsim:
@kishikawakatsumi feel free to merge this once you've confirmed that my latest change is correct.

kishikawakatsumi:
@jpsim Thanks. I've confirmed.

PR #2767, opened 2015-10-29
Improve the -[RLMRealm compact] test and fix a few bugs

tgoyne:
Compact wasn't leaving the Realm in a valid state or translating errors correctly. Also bumps the core version since one of the relevant bugs had to be fixed there.


bdash:
:+1:

PR #2781, opened 2015-11-02
Fix reporting of the Target OS Type when building with older iOS SDKs.

bdash:
`TARGET_OS_IOS` is new in iOS 9.0, so builds made with older iOS SDKs were reported as targeting OS X. Switch to using `TARGET_OS_IPHONE` instead.

/cc @jpsim @timanglade


kishikawakatsumi:
:+1:

PR #2783, opened 2015-11-02
Fix test-ios-static with XCMODE=xcpretty

bdash:
Update `xcrealm_work_around_rdar_23055637` to invoke the expected-to-fail build in a subshell so we can ignore its exit status in all cases.

This should also fix the releasability tests as they invoke `test-iso-static` with `XCMODE=xcpretty`.

/cc @jpsim @tgoyne


tgoyne:
:+1:

PR #2784, opened 2015-11-03
Add `@noescape` attribute to `write(_:)` method

kishikawakatsumi:
Because the closure will not be stored anywhere and executed synchronously.
Also, we can use an implicit `self`, saving us from typing `self`.

cc/ @jpsim @mrackwitz


jpsim:
Huh, we could never get this to compile under Xcode 6.x. I just re-ran this on CI, if it passes :+1:.

This should be accompanied by a changelog entry though. Nice work @kishikawakatsumi!

kishikawakatsumi:
Added changelog entry.

jpsim:
:+1:

PR #2787, opened 2015-11-03
fix type

prises0221:
fix type


jpsim:
Thanks for the fix, @prises0221.

PR #2791, opened 2015-11-04
Embed Realm.framework in Tests.xctest, and remove the unnecessary duplication of the TestHost target

bdash:
This is what I should have done instead of b1bf3bfa, but I didn't think of it at the time.


jpsim:
This just keeps getting better and better :+1:

PR #2793, opened 2015-11-04
Fold RealmSwift.xcodeproj into Realm.xcodeproj

bdash:
This should address two issues:

1. The lack of dependency tracking between the two projects that can lead to RealmSwift using a stale Realm.framework when you're making changes in both projects.
2. Carthage support for building RealmSwift from source.

Remaining work:
- [x] Ensure that tests pass on CI.
- [x] Test packaging via `build.sh test-package-release`.


bdash:
Some of the earlier CI failures were due to the CocoaPods specs GitHub repository being unavailable. Everything is working now though!

/cc @jpsim @tgoyne

build.sh
```diff
@@ -422,6 +409,7 @@ case "$COMMAND" in
         ;;
 
     "ios-swift")
+        sh build.sh ios-dynamic
```
jpsim: I'm assuming this is needed to build the fat framework?

build.sh
```diff
@@ -446,9 +435,10 @@ case "$COMMAND" in
         ;;
 
     "osx-swift")
-        xcrealmswift "-scheme 'RealmSwift' -configuration $CONFIGURATION build"
+        sh build.sh osx
```
jpsim: is the OS X Objective-C framework not built as a dependency of the Swift one?

build.sh
bdash: Building RealmSwift.framework takes care of building Realm.framework. This step is only necessary to copy the built framework to the expected location in `build/ios`.

build.sh
bdash: This is also just about getting the framework into the right place in `build`.

jpsim:
I'm impressed. This looks really great, @bdash. Thanks for the detailed commit messages along the way too!

kishikawakatsumi:
@bdash @jpsim Carthage still fails to build this branch. Because Carthage cannot support building static framework.

```
$ carthage update
*** Fetching realm-cocoa
*** Checking out realm-cocoa at "0d63ca9fee4043df79ece0fbf0b8c0f0eebc7cea"
*** xcodebuild output can be found in /var/folders/wr/f0q3fd5976nd8fwpwk717btr0000gn/T/carthage-xcodebuild.5KYQeJ.log
*** Building scheme "RealmSwift" in Realm.xcodeproj
*** Building scheme "Realm" in Realm.xcodeproj
*** Building scheme "Realm iOS static" in Realm.xcodeproj
Invalid architecture UUIDs: Could not parse UUIDs using dwarfdump from /Users/katsumi/Library/Developer/Xcode/DerivedData/Realm-cguenequhyeyovgplekzxgskrpqf/Build/Products/Release-iphonesimulator-static/Realm.framework/Realm
```

jpsim:
Carthage is supposed to skip static framework targets. Maybe we're using the wrong target type ID? e.g. `com.apple.product-type.framework.static`. See https://github.com/Carthage/Carthage/blob/35924f38cd20689c8af0a0b2ea97681bb1cac10f/Source/CarthageKit/Xcode.swift#L731-L734 and https://github.com/Carthage/Carthage/blob/35924f38cd20689c8af0a0b2ea97681bb1cac10f/Source/CarthageKit/Xcode.swift#L455-L456.

bdash:
@kishikawakatsumi, thanks for testing that. I'd tested Carthage building for Mac, but that skips the static framework due to the target not supporting OS X. I pushed 2f117e0601aa837cbbad340114065c4119c316b8 which should fix building for iOS with Carthage.

kishikawakatsumi:
@bdash Awesome. I confirmed to build succeeded.

tgoyne:
:+1: 

Changes all LGTM and I haven't hit any issues using it for dev today.

PR #2800, opened 2015-11-04
Added new API integration for disallowing automatic file format upgrades

TimOliver:
When adding the optional upgrade path feature to Realm Browser, it was necessary to expose the 'disallow file upgrade' option in the Core via the Cocoa binding. 

I tried to keep this new API private, and all internal modifications to be as unobtrusive as possible, but please let me know if you think there's anything that needs changing. 

/cc @jpsim @tgoyne


Realm/RLMRealm_Private.h
```diff
@@ -69,6 +69,44 @@ FOUNDATION_EXTERN NSData *RLMRealmValidatedEncryptionKey(NSData *key);
...
+ possible errors, pass in NULL.
+ 
+ @return An `RLMRealm` instance.
+ 
+ @see RLMRealm defaultRealm
+ @see RLMRealm realmWithPath:
+ @see RLMRealm realmWithPath:readOnly:error:
+ @see RLMRealm realmWithPath:encryptionKey:readOnly:error:
+ */
++ (instancetype)realmWithPath:(NSString *)path
```
bdash: Why add this method rather than exposing the `disableFormatUpgrade` property on `RLMRealmConfiguration`, and using `realmWithConfiguration:error:`?

TimOliver: Mainly convenience. It's actually just a wrapper method around a `RLMRealmConfiguration` object that handles all of those properties (Just like the method above it).

Is there a particular reason why it's not a good idea to add this in?

bdash: The other similar methods predate `RLMRealmConfiguration`. Now that it exists I don't think there's any reason to have this in Realm proper, when it can easily be a helper function in the browser code that uses it.

TimOliver: Fair enough. I'll remove this one and move the logic to the Browser then.

TimOliver:
Alright, I removed the `RLMRealm` method that I originally added and refactored the Browser to rely solely on `RLMRealmConfiguration` for that functionality now. At some point, we should probably do a review of the other private, pre-configuration APIs in there and see if they need deprecating.

Is there anything else this needs before merging?

jpsim:
You've improved this a lot, @TimOliver! Thanks. This looks good to me. Further thoughts @bdash?

Realm/RLMRealmConfiguration.mm
```diff
@@ -236,5 +236,16 @@ - (void)setCustomSchema:(RLMSchema *)customSchema {
     _config.schema = [_customSchema objectStoreCopy];
 }
 
+- (void)setDisableFormatUpgrade:(bool)disableFormatUpgrade
+{
+    _config.disable_format_upgrade = disableFormatUpgrade;
+    _config.cache = false;
```
bdash: In `-setDynamic:` we only disable caching if we're setting the value to the non-default value. I think we be consistent between these two methods.

bdash:
:+1:

Realm/RLMRealmConfiguration.mm
tgoyne: I don't think it actually needs to set `cache = false` at all, as setting this either does nothing or results in it throwing an exception, in which case the realm won't be cached anyway.

tgoyne:
This could use a test, but that requires a test file so it could end up being an annoying amount of work for something that's almost certainly correct, so eh. If you do want to add a test I'd just grab the smallest file from core's set of v2 realms and verify that it throws an exception when opened with this set.

:+1:

TimOliver:
Alrighty! If everyone's cool with this, I'll merge it in now and then update the Browser's podfile to stop pointing to this branch.

I agree with @tgoyne in that it might be best to have a test in place, if only to simply ensure this doesn't cause any issues down the line. But I'm not sure how we can keep an 'intentionally old' Realm file around in the CI in order to keep checking this, so it might be something I'll look at later, and put in a separate PR.

Thanks!

PR #2805, opened 2015-11-05
Use the path of the file which actually failed to open in error messages

tgoyne:
This was done before the object store changes and was missed when those were updated.


bdash:
:+1:

PR #2808, opened 2015-11-05
test Carthage installation examples from source

jpsim:
we couldn't do this before because we didn't support building Realm Swift from source, relying instead on testing the last prebuilt binaries uploaded to GitHub releases.

/cc @bdash

### DIFF
--- a/examples/installation/build.sh
+++ b/examples/installation/build.sh
@@ -57,8 +57,7 @@ xctest() {
     elif [[ $NAME == Carthage* ]]; then
         (
             cd "$DIRECTORY";
-            : ${sha:=master}
-            echo "github \"realm/realm-cocoa\" \"$sha\"" > Cartfile
+            echo "github \"realm/realm-cocoa\" \"${sha:-master}\"" > Cartfile
             carthage update
         )
     elif [[ $LANG == swift* ]]; then
--- a/examples/installation/build.sh
+++ b/examples/installation/build.sh
@@ -55,7 +55,12 @@ xctest() {
     if [[ $NAME == CocoaPods* ]]; then
         pod install --project-directory="$DIRECTORY"
     elif [[ $NAME == Carthage* ]]; then
-        (cd "$DIRECTORY"; carthage update)
+        (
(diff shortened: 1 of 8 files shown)


examples/installation/build.sh
```diff
@@ -55,7 +55,12 @@ xctest() {
     if [[ $NAME == CocoaPods* ]]; then
         pod install --project-directory="$DIRECTORY"
     elif [[ $NAME == Carthage* ]]; then
-        (cd "$DIRECTORY"; carthage update)
+        (
+            cd "$DIRECTORY";
+            : ${sha:=master}
+            echo "github \"realm/realm-cocoa\" \"$sha\"" > Cartfile
```
bdash: I'd have just used `${sha:-master}` when expanding it into the Cartfile rather than having the separate assign-if-unset, but that's just a matter of style.

bdash:
Do we have any tests that are checking GitHub releases after this?

jpsim:
> Do we have any tests that are checking GitHub releases after this?

Not at the moment, no. Ideally we'd have a job at the end of the `release-cocoa` Jenkins job that would verify all examples (usage & installation) pointing to the release, which we don't currently do at all for any of our release methods.

bdash:
When JP and I talked about this last week the conclusion that we reached is that it's important that we continue to test using Carthage with GitHub releases since that's still going to be the primary flow with official releases. JP said he'd add testing of GitHub releases into the release CI jobs as part of this PR.

jpsim:
@bdash `examples/installation/build.sh` now targets the last official release when generating the Cartfile if the `REALM_BUILD_USING_LATEST_RELEASE` environment variable is set.

This will allow us to test binary Carthage releases as part of the `release-cocoa` job. I'll implement that job in Jenkins once this is merged.

bdash:
:+1:

jpsim:
Jenkins' `release_cocoa` job now runs the `Cocoa Release Installation Examples` job at the end, which uses the GitHub release binaries to test Carthage installation & the realm.io binaries to test binary installation.

PR #2811, opened 2015-11-06
added 'import Foundation' to Swift template

samritchie:
Otherwise you’ll get “'dynamic' attribute used without importing module 'Foundation'”


jpsim:
Thanks @samritchie!

PR #2817, opened 2015-11-07
Generate both Objective-C and Swift docs using Jazzy

jpsim:
Requires https://github.com/realm/jazzy/pull/341. Output can be browsed here: https://static.realm.io/jazzy_demo/Realm/

/cc @tgoyne @bdash @timanglade


bdash:
I took a quick look at the output and noticed that most of the "Show on GitHub" links have incorrect line numbers. For instance, the documentation for [`-[RLMObject linkingObjectsOfClass:forProperty:]`](https://static.realm.io/jazzy_demo/Realm/Classes/RLMObject.html#/c:objc(cs)RLMObject(im)linkingObjectsOfClass:forProperty:) links to [line 381 of RLMObject.h](https://github.com/realm/realm-cocoa/blob/v0.96.2/Realm/RLMObject.h#L381).

jpsim:
That's because the GitHub link is being generated as if these were the docs at `v0.96.2`. If you replace that with `jp-jazzy-objc`, it'll lead to the right place: https://github.com/realm/realm-cocoa/blob/jp-jazzy-objc/Realm/RLMObject.h#L381

build.sh
```diff
@@ -601,9 +637,14 @@ case "$COMMAND" in
         ;;
 
     "verify-docs")
-        sh scripts/build-docs.sh
+        sh build.sh docs
+        if [ -s docs/objc_output/undocumented.txt ]; then
+          echo "Undocumented Realm Objective-C declarations:"
+          echo "$(cat docs/objc_output/undocumented.txt)"
```
tgoyne: Can just be `cat docs/objc_output/undocumented.txt`

build.sh
```diff
@@ -221,6 +221,42 @@ test_ios_devices() {
...
+
+    if [[ "$language" == "swift" ]]; then
+        : ${REALM_SWIFT_VERSION:=2.1}
+        ./build.sh set-swift-version
+        xcodebuild_arguments="-scheme,RealmSwift"
+        module="RealmSwift"
+        objc=""
+    fi
+
+    touch Realm/RLMPlatform.h # jazzy will fail if it can't find all public header files
```
tgoyne: Should delete this afterwards.

tgoyne:
:+1:

PR #2823, opened 2015-11-09
Fix crashes when manually refreshing during fast enumeration

tgoyne:
The enumerator was not being detached before the read transaction was advanced, which could lead to assertion failures or incorrect results.


jpsim:
:+1: nice find

tgoyne:
FIXME is addressed by https://github.com/realm/realm-core/pull/1304.

PR #2828, opened 2015-11-10
Fixes to Swift doc comments

samritchie:
Most of these are fairly minor formatting changes, fixes to param names & the like. I’ve also:

* Added warnings on the Realm methods that require write transactions - as these were on each of the mutating List methods.
* Fixed some Swift 2 style comments in Swift 1 & vice versa
* Removed the reference to the free objects() function as this seems to be gone?
* Added a 'background thread' warning on autoRefresh
* Where possible, changed references to 'links' to 'relationships' in line with the Java terminology
* Added the additional Int types. 

On that last one, I’ll need confirmation about the Int8 support, as it doesn’t seem to be complete - not sure if this is an error, or it’s not meant to be supported after all.

### DIFF
--- a/RealmSwift-swift1.2/Aliases.swift
+++ b/RealmSwift-swift1.2/Aliases.swift
@@ -42,7 +42,7 @@ See [Realm Models](http://realm.io/docs/swift/latest/#models).
 * Data
 * Date
 
-## Array/Linked object types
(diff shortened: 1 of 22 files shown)


jpsim:
Awesome changes, thanks!

PR #2829, opened 2015-11-10
add `Int8` support for `RealmOptional`, `MinMaxType` and `AddableType`.

jpsim:
/cc @bdash


bdash:
Presumably this was just an oversight?

bdash:
:+1:

jpsim:
> Presumably this was just an oversight?

Indeed it was. We initially didn't support Int8 at all, until #1740 addressed that, but we didn't fully integrate it.

PR #2831, opened 2015-11-10
Port most of RLMResults's functionality to realm::Results

tgoyne:
Everything but `-description` and the fast enumeration stuff is now a fairly thin wrapper around the objectstore Results class. I'd like to port `RLMFastEnumerator` to the object store in the future, but it's awkward to do so while not yet using the objectstore List class, and it's not needed for the async query stuff, which the rest of the stuff is a prereq for.

Shouldn't be any end-user visible functional changes.

### DIFF
--- a/Realm/ObjectStore/results.cpp
+++ b/Realm/ObjectStore/results.cpp
@@ -17,42 +17,288 @@
 ////////////////////////////////////////////////////////////////////////////
 
 #include "results.hpp"
-#import <stdexcept>
+
+#include <stdexcept>
 
 using namespace realm;
 
-Results::Results(SharedRealm &r, ObjectSchema &o, Query q, SortOrder s) :
-    realm(r), object_schema(o), backing_query(q), table_view(backing_query.find_all())
+Results::Results(SharedRealm r, Query q, SortOrder s)
+: m_realm(std::move(r))
+, m_query(std::move(q))
+, m_table(m_query.get_table().get())
+, m_sort(std::move(s))
+, m_mode(Mode::Query)
 {
-    setSort(std::move(s));
+}
+
+Results::Results(SharedRealm r, Table& table)
+: m_realm(std::move(r))
+, m_table(&table)
+, m_mode(Mode::Table)
+{
+}
+
+void Results::validate_read() const
+{
+    if (m_realm && !m_realm->check_thread())
+        throw Error::IncorrectThread;
+    if (m_table && !m_table->is_attached())
+        throw Error::Invalidated;
+}
+
+void Results::validate_write() const
+{
+    validate_read();
+    if (!m_realm || !m_realm->is_in_transaction())
+        throw Error::NotInWrite;
 }
 
 size_t Results::size()
 {
-    verify_attached();
-    return table_view.size();
-}
-
-void Results::setSort(SortOrder s)
-{
-    sort_order = std::make_unique<SortOrder>(std::move(s));
-    table_view.sort(sort_order->columnIndices, sort_order->ascending);
-}
-
-Row Results::get(std::size_t row_ndx)
-{
-    verify_attached();
-    if (row_ndx >= table_view.size()) {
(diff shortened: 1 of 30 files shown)


Realm/ObjectStore/results.cpp
```diff
@@ -0,0 +1,334 @@
...
+{
+    switch (m_mode) {
+        case Mode::Empty:
+            return;
+        case Mode::Table:
+            validate_write();
+            m_table->clear();
+            break;
+        case Mode::Query:
+            // Not using Query:remove() because bulding the tableview and
```
jpsim: s/bulding/building/

Realm/RLMObjectStore.mm
```diff
@@ -517,15 +519,21 @@ id RLMGetObject(RLMRealm *realm, NSString *objectClassName, id key) {
         return nil;
     }
 
-    return RLMCreateObjectAccessor(realm, objectSchema, row);
+    return RLMCreateObjectAccessor(realm, objectSchema, (*objectSchema.table)[row]);
```
jpsim: Isn't the whole point of keeping the `RLMCreateObjectAccessor` variant that takes an index to be able to just pass in the index? IMO we could just remove that variant.

Realm/RLMResults.mm
```diff
@@ -265,77 +243,57 @@ - (NSUInteger)indexOfObjectWhere:(NSString *)predicateFormat args:(va_list)args
...
-    // copy array and apply new predicate creating a new query and view
-    auto query = [self cloneQuery];
-    RLMUpdateQueryWithPredicate(query.get(), predicate, _realm.schema, _realm.schema[self.objectClassName]);
-    size_t index = query->find();
+    Query query = translateErrors([&] { return _results.get_query(); });
+    RLMUpdateQueryWithPredicate(&query, predicate, _realm.schema, _objectSchema);
+    size_t index = query.find();
     if (index == realm::not_found) {
-        return NSNotFound;
+        return (NSUInteger)NSNotFound;
```
jpsim: this cast shouldn't be necessary

Realm/ObjectStore/results.cpp
```diff
@@ -0,0 +1,334 @@
...
+}
+
+util::Optional<RowExpr> Results::first()
+{
+    validate_read();
+    switch (m_mode) {
+        case Mode::Empty:
+            return none;
+        case Mode::Table:
+            return m_table->size() == 0 ? util::none : util::Optional<RowExpr>(m_table->front());
```
bdash: You could use `util::make_optional` here to avoid restating the contained type.

Realm/RLMObjectStore.mm
tgoyne: I think I started removing that overload, then discovered that it was actually noticeably nicer in most of the places where it's used since the `objectSchema` is often an expensive expression that'd otherwise need to be cached. I'll switch this one back to just passing this index.

Realm/ObjectStore/results.cpp
```diff
@@ -0,0 +1,334 @@
...
+}
+
+Results Results::sort(realm::SortOrder&& sort) const
+{
+    return Results(m_realm, get_query(), std::move(sort));
+}
+
+Results Results::filter(Query&& q) const
+{
+    return Results(m_realm, get_query().and_query(q), get_sort());
```
bdash: `std::move(q)` will let `and_query` avoid a copy.

bdash:
:+1:

tgoyne:
The ios-device-swift job is consistently failing and I'm not really sure what's going on. It fails to copy the copy test bundle dSYM a whole bunch of times, runs "Test suite 'empty'" a whole bunch of times, and in between seems to run all of the actual tests and pass them all.

jpsim:
I'm also not sure why the device tests are failing, but they continue to pass consistently for other PRs.

PR #2833, opened 2015-11-11
Fix a typo in an error message

tgoyne:


jpsim:
:+1: you know you really don't need to await/solicit review for obvious typo fixes like this :wink:

PR #2842, opened 2015-11-12
Add support for running queries asynchronously

### LINKED ISSUE #3105: Support running queries asynchronously

tgoyne:
Things to do:

 - [x] Switch from deliverOn: to addNotificationBlock:
 - [x] Add addNotificationBlock: to RLMArray, RLMCollection, List, RealmCollection
 - [x] Don't create an AsyncQuery per notification block
 - [x] Background-update Results which don't have notification blocks
 - [x] Don't pointlessly update Results which aren't still being used
 - [x] Update samples
 - [x] Backport stuff to Swift 1.2
 - [x] Don't create background queries for Results outside of a runloop
 - [x] More testing
 - [x] Helgrind/tsan a stress-test of stuff

Fixes #3105


Realm/RLMResults.h
```diff
@@ -150,6 +150,67 @@ RLM_ASSUME_NONNULL_BEGIN
...
+ when opening the RLMRealm on the background worker thread or the destination
+ queue fails.
+
+ The RLMResults passed to the block is a normal auto-updating RLMResults object.
+ This means that if you perform a write transaction or explicitly call
+ -[RLMRealm refresh] from within the block, the query may be synchronously rerun
+ the next time you access the results.
+
+ @warning This method cannot be called during a write transaction, or when the
+          containing realm is read-only.
```
astigsen: How come that you cannot do async queries on read-only realms?

tgoyne: Handover is only implemented for SharedGroup since it uses the transaction version.

astigsen:
Would it be possible to add an example of how this API would be used by a UI component (like a TableView)?

Realm/RLMResults.h
astigsen: hmm, we should really get that fixed in core.

cmelchior:
Just to chip in with the lessons learned from the discussions on this on the Java side.
We originally had the API you propose here as well, but finally settled on a "promise"-based API, where an empty list is returned immediately:
```
RealmResults<Person> persons = realm.where(Person.class).findAllAsync()
assertTrue(persons.size == 0); // List is a normal list, but empty
assertFalse(persons.isLoaded()); // Boolean to check state
persons.load(); // Block until complete
assertTrue(persons.isLoaded()); 
Persons p = persons.get(0);
persons.addChangeListener(() -> {
  // Callback whenever the list is refreshed, also called when data is loaded the first time
})


```

This works nicely with UI elements as you just parse in the empty list immediately and the UI component can then register it's own changelistener and/or use `isLoaded()`. It also has really nice properties in reactive systems like RxJava,

tgoyne:
> assertTrue(persons.size == 0); // List is a normal list, but empty

Silently returning incorrect results rather than blocking or throwing sounds like a really bad idea. Other than that, `isLoaded()` looks like the only thing that `RLMResults` doesn't already have.

cmelchior:
It is not returning incorrect results, it is returning a "promise" that looks and behaves like a list. If you care about the difference you can check "isLoaded", but UI elements that just want to show "latest state", they just render it and don't have to know the difference between "loading", "empty result" or "results".

jpsim:
@cmelchior although I see the value of a promise-like API for async queries, especially for Android where `RealmResults` has explicit `findAll()` materializations in the synchronous case.

However, a promise-like API for cocoa doesn't work as well because it breaks the assumption that an `RLMResults` can always be accessed as if it was materialized and will always return an accurate view of the query in the Realm. By going with the block approach, the RLMResults that are handed back in the block retain this assumption. This way, users never have to check `isLoaded` (and to always know when they're dealing with synchronous or asynchronous results).

So I'm happy to stick with this block-based API.

Realm/RLMResults.h
```diff
@@ -150,6 +150,67 @@ RLM_ASSUME_NONNULL_BEGIN
...
+
+ @param queue The dispatch queue onto which the results should be delivered.
+ @param block The block to be called on the given `queue` with the queue-local copy of the results.
+ @return A token which must be held for as long as you want query results to be delivered.
+ */
+- (RLMNotificationToken *)deliverOn:(dispatch_queue_t)queue
+                              block:(void (^)(RLMResults RLM_GENERIC_RETURN *__nullable results, NSError *__nullable error))block;
```
jpsim: Under which circumstances can this block be invoked with a null `results`?

jpsim: Sorry, that's clearly covered in the documentation just above.

Realm/Tests/AsyncTests.m
```diff
@@ -0,0 +1,808 @@
...
+    }];
+    [self waitForExpectationsWithTimeout:2.0 handler:nil];
+
+    expectation = [self expectationWithDescription:@""];
+    ++expected;
+    [self dispatchAsyncAndWait:^{ [self createObject:1]; }];
+    [self waitForExpectationsWithTimeout:2.0 handler:nil];
+
+    expectation = [self expectationWithDescription:@""];
+    [self dispatchAsyncAndWait:^{ [self createObject:-1]; }];
```
jpsim: What kinds of changes will or will not trigger a redelivery? Will writing to a different column of an observed table trigger one?

Realm/Tests/AsyncTests.m
```diff
@@ -0,0 +1,808 @@
...
+    }];
+    [self waitForExpectationsWithTimeout:2.0 handler:nil];
+
+    [self waitForNotification:RLMRealmDidChangeNotification realm:RLMRealm.defaultRealm block:^{
+        RLMRealm *realm = [RLMRealm defaultRealm];
+        [realm transactionWithBlock:^{
+            [StringObject createInDefaultRealmWithValue:@[@""]];
+        }];
+    }];
+    [token stop];
```
jpsim: Shouldn't we pause and wait here before we stop listening for updates?

RealmSwift-swift2.0/Results.swift
```diff
@@ -285,6 +285,86 @@ public final class Results<T: Object>: ResultsBase {
...
+     parameter and a non-`nil` error. Currently the only errors that can occur
+     are when opening the Realm on the background worker thread or the
+     destination queue fails.
+
+     The Results passed to the block is a normal auto-updating Results object.
+     This means that if you perform a write transaction or explicitly call
+     realm.refresh() from within the block, the query may be synchronously
+     rerun the next time you access the results.
+
+     @warning This method cannot be called during a write transaction, or when
```
jpsim: all these swift async API docs are currently using the ObjC syntax

RealmSwift-swift2.0/Results.swift
```diff
@@ -285,6 +285,86 @@ public final class Results<T: Object>: ResultsBase {
...
+              the containing realm is read-only.
+
+     @param queue The dispatch queue onto which the results should be delivered.
+     @param block The block to be called on the given `queue` with the queue-local copy of the results.
+     @return A token which must be held for as long as you want query results to be delivered.
+     */
+    public func deliverOn(queue: dispatch_queue_t, block: (Results<T>?, NSError?) -> ()) -> NotificationToken {
+        weak var cachedResults: Results<T>? = nil
+        return rlmResults.deliverOn(queue) { results, error in
+            if let results = results {
```
jpsim: `guard`s would be nicer here

Realm/Tests/AsyncTests.m
tgoyne: Currently any write to the table or any table linking to/from it (or transitive links thereof) will cause a redelivery.

Realm/Tests/AsyncTests.m
tgoyne: There's nothing to wait for. This test relies on the fact that Realm notifications are sent after new results.

tgoyne:
One idea I've been considering is having a non-dispatching version that simply delivers to the current thread and looks much more like the RLMRealm notifications:

```
RLMResults *results = [...];
RLMNotificationToken *token = [results addNotificationBlock:^(NSError *error) {
    // Do stuff with the original `results`, which has been populated with the query data 
}];
```

The version that delivers to a different queue would then be something like:

```
RLMResults *results = [...];
RLMNotificationToken *token = [results addNotificationOnQueue:queue block:^(RLMResults *results, NSError *error) {
    // Have to used the passed-in results since it's on a different thread
}];
```

-----

With or without this, we could also asynchronously rerun *all* queries in the background. This could be done with fairly minimal overhead with a two-step automatic registration. When a Results's TableView is created, we'd schedule an event in the current runloop which would actually create the AsyncQuery. This would avoid running queries in the background for Results which are actually just temporary objects that will go away soon. We could also track if the background-updated version is ever actually used, and if not stop updating it, to avoid forever doing work for Results that have been leaked.

tgoyne:
I tried porting ABFRealmSearchViewController to this and it seems to work pretty well: https://github.com/tgoyne/ABFRealmSearchViewController/commit/1e158a988b063711ffc634353f8f355415fdca6b

Realm/ObjectStore/impl/apple/cached_realm.hpp
```diff
@@ -0,0 +1,65 @@
...
+
+    // Get a strong reference to the cached realm
+    std::shared_ptr<Realm> realm() const { return m_realm.lock(); }
+
+    // Does this CachedRealm store a Realm instance that should be used on the current thread?
+    bool is_cached_for_current_thread() const { return m_cache && m_thread_id == std::this_thread::get_id(); }
+
+    bool expired() const { return m_realm.expired(); }
+
+    // Asyncronously call notify() on the Realm on the appropriate thread
```
bdash: Asynchronously

Realm/ObjectStore/impl/realm_coordinator.cpp
```diff
@@ -0,0 +1,427 @@
...
+#include <realm/query.hpp>
+#include <realm/table_view.hpp>
+
+#include <cassert>
+
+using namespace realm;
+using namespace realm::_impl;
+
+static std::mutex s_coordinator_mutex;
+static std::map<std::string, std::weak_ptr<RealmCoordinator>> s_coordinators_per_path;
```
bdash: Did you consider using `std::unordered_map` here? I don't see any need for ordering here.

Realm/ObjectStore/impl/realm_coordinator.cpp
```diff
@@ -0,0 +1,427 @@
...
+    auto it = s_coordinators_per_path.find(path);
+    if (it != s_coordinators_per_path.end()) {
+        coordinator = it->second.lock();
+    }
+
+    if (!coordinator) {
+        s_coordinators_per_path[path] = coordinator = std::make_shared<RealmCoordinator>();
+    }
+
+    return coordinator;
```
bdash: This could be written more clearly as:

```c++
    auto& weak_coordinator = s_coordinators_per_path[path];
    if (auto coordinator = weak_coordinator.lock()) {
        return coordinator;
    }
    auto coordinator = std::make_shared<RealmCoordinator>();
    weak_coordinator = coordinator;
    return coordinator;
```

Realm/ObjectStore/impl/realm_coordinator.cpp
```diff
@@ -0,0 +1,427 @@
...
+            m_advancer_history = nullptr;
+        }
+    }
+    else if (m_queries.empty() && m_new_queries.empty()) {
+        // If this is the first query then we don't already have a read transaction
+        m_advancer_sg->begin_read(versionid);
+    }
+    else if (versionid < m_advancer_sg->get_version_of_current_transaction()) {
+        // Ensure we're holding a readlock on the oldest version we have a
+        // handover object for, as handover objects don't
```
bdash: This comment feels like it was cut off before it finished.

bigfish24:
@tgoyne thanks for your work on porting the search controller! Catching up here, I think I understand your position and how things would work with a UI component. Note: the search controller isn't demonstrating the issue though because @astigsen's request was to allow a UI component to be passed a `RLMResult` rather than create one internally. However, correct me if I am wrong, but it seems like with the current implementation/proposal the dev would do this:

``` objc
UIComponent *component = [UIComponent alloc] init];

// UIComponent is in "unloaded" state since it doesn't have a collection of objects

[Object objectsWhere:@"..." deliverOnMainThread:^(RLMResults *results, NSError *e) {
    [component updateObjects:results];
    // UIComponent is in "loaded" state since it has a collection of objects
}];

....
// Within UIComponent
- (void)updateObjects:(id<RLMCollection>) {
    // Component adds a notification block to update based off of query rerunning
    typeof(self) __weak weakSelf = self;
    self.token = [results addNotificationBlock:^(NSError *error) {
        [weakSelf.tableView reloadData];
    }];
}
```
I think this is reasonable, the only downside that I see is that the component is never automatically aware of when the query is running and thus it should be in a "loading" state.

tgoyne:
The pattern I have in mind is just:

```objc
UIComponent *component = [UIComponent alloc] init];
[component updateObjects:[Object objectsWhere:@"..."]];

....

// Within UIComponent
- (void)updateObjects:(id<RLMCollection>)objects {
    _objects = objects;

    // ... Switch to loading state ...

    // Add notification block for when the query is ready or has changed
    typeof(self) __weak weakSelf = self;
    self.token = [objects addNotificationBlock:^(__unused RLMResults *r, NSError *error) {
        // r is the same object as _results, so no need to actually use it in this case
        // ... switch out of loading state if needed ...
        [weakSelf.tableView reloadData];
    }];
}
```

jpsim:
@bigfish24 that's one way to do it, and if the component was designed specifically for async queries, it could take ownership of the results:

```objc
RLMResults *results = [Object objectsWhere:@"..."];
UIComponent *component = [[UIComponent alloc] initWithResults:results];

....
// Within UIComponent
- (instancetype)initWithResults:(RLMResults *)results {
    self = [super init];
    if (self) {
        __weak typeof(self) weakSelf = self;
        self.resultsToken = [results deliverOnMainThread:^(RLMResults *results, NSError *error) {
            if (self.results == nil) {
                self.results = results;
            }
            [weakSelf.tableView reloadData];
        }];
    }
    return self;
}
```

@tgoyne let me know if that's right

bigfish24:
@tgoyne @jpsim ah that makes sense to pass in the empty `RLMResults` and let the component take care of it. I like this... @astigsen any objections?

bigfish24:
I am assuming the `addNotificationBlock:` method will be in the `RLMCollection` protocol and apply to `RLMArray` too?

tgoyne:
Yes, I think it needs to exist on RLMArray despite being conceptually a bit weird there (although all the same weirdness applies to table-backed results with no query, which are also currently handled very suboptimally).

astigsen:
>  ah that makes sense to pass in the empty `RLMResults` and let the component take care of it. I like this... @astigsen any objections?

I like it as well. I do find it a bit confusing with the `RLMResults` essentially being "upgraded" to an async version, but maybe that can be made clearer through naming of the methods?

bigfish24:
@astigsen agreed the `addNotificationBlock:` method would need to be more clear that it will run the query if it hasn't yet, not sure if the name should elicit that or just the docs since this won't always be the case if you call it on a `RLMResults` in which the query already ran.

tgoyne:
`addNotificationBlock` will always asynchronously call the block with the initial results regardless of the state of the RLMResults. Abstracting away the idea of "has the query been run yet" is pretty much the entire point of the design of RLMResults.

kishikawakatsumi:
@tgoyne @jpsim 

I found a strange behavior. If I would overwrite the NotificationToken, second and subsequent notifications will not reach. The sample project is here https://github.com/kishikawakatsumi/AsyncQuerySample.

Reproducible code snippets are like the following:

```objc
- (IBAction)buttonAction:(id)sender {
//    [self.token stop];
    self.token = [[Book allObjects] addNotificationBlock:^(RLMResults * _Nullable results, NSError * _Nullable error) {
        NSLog(@"%@", error);
        NSLog(@"%@", results);
        // Never call this block after second notification.
    }];
}
```

If calling the `stop` method before overwriting new NotificationToken, the issue won't occur.

It occurs only on application. It has never seen on unit tests. I tried to write tests, but cannot. Is it related to runloop?

tgoyne:
@kishikawakatsumi:  06bf2a1 fixes that bug (and 507c8fa makes it so that the exception which was being thrown is logged rather than it just silently killing the thread).

kishikawakatsumi:
@tgoyne 

I found another crash. Reproducible sample project is here: https://github.com/kishikawakatsumi/AsyncQuerySample.

Launch apps, then push button, hits `EXC_BAD_ACCESS` on line 81 of `async_query.cpp`.

https://github.com/realm/realm-cocoa/pull/2842/files#diff-b64175b6d3ea73e6fb807024c62653b6R81

Maybe notification token is being deallocated before reaching callback causes it?

```
- (IBAction)buttonAction:(id)sender {
    self.token = [[Book objectsWhere:@"title CONTAINS %@", @"Realm"] addNotificationBlock:^(RLMResults * _Nullable results, NSError * _Nullable error) {
        NSLog(@"%@", error);
        NSLog(@"%@", results);
    }];
    self.token = [[Book objectsWhere:@"title CONTAINS %@", @"Realm"] addNotificationBlock:^(RLMResults * _Nullable results, NSError * _Nullable error) {
        NSLog(@"%@", error);
        NSLog(@"%@", results);
    }];
}
```

Realm/ObjectStore/impl/apple/cached_realm.hpp
```diff
@@ -0,0 +1,53 @@
...
+    CachedRealm(const std::shared_ptr<Realm>& realm, bool cache);
+    ~CachedRealm();
+
+    CachedRealm(CachedRealm&&);
+    CachedRealm& operator=(CachedRealm&&);
+
+    CachedRealm(const CachedRealm&) = delete;
+    CachedRealm& operator=(const CachedRealm&) = delete;
+
+    // Asyncronously call notify() on the Realm on the appropriate thread
```
jpsim: "Asynchronously" and "for the appropriate thread"?

Realm/Tests/AsyncTests.m
```diff
@@ -0,0 +1,823 @@
...
+    __block bool called = false;
+    XCTestExpectation *exp = [self expectationWithDescription:@""];
+    id token = [IntObject.allObjects addNotificationBlock:^(RLMResults *results, NSError *error) {
+        XCTAssertNil(results);
+        XCTAssertNotNil(error);
+        called = true;
+        [exp fulfill];
+    }];
+
+    // Block should still be called asyncronously
```
jpsim: asynchronously

Realm/ObjectStore/impl/apple/external_commit_helper.cpp
```diff
@@ -211,79 +130,57 @@ ExternalCommitHelper::ExternalCommitHelper(Realm* realm)
...
+    m_shutdown_write_fd = shutdown_pipe[1];
 
-    m_realms.push_back({realm, runloop, signal});
+    m_thread = std::async(std::launch::async, [=] {
+        try {
+            listen();
+        }
+        catch (std::exception const& e) {
+            fprintf(stderr, "uncaught exception in notifier thread: %s: %s\n", typeid(e).name(), e.what());
+            asl_log(nullptr, nullptr, ASL_LEVEL_ERR, "uncaught exception in notifier thread: %s: %s", typeid(e).name(), e.what());
```
jpsim: can you help me understand why the `fprintf` is necessary given that we're logging the same thing via `asl_log`?

Realm/ObjectStore/impl/cached_realm_base.hpp
```diff
@@ -0,0 +1,68 @@
...
+
+#include <memory>
+#include <thread>
+
+namespace realm {
+class Realm;
+
+namespace _impl {
+
+// CachedRealm stores a weak reference to a Realm instance, along with all of
```
jpsim: CachedRealmBase?

Realm/ObjectStore/results.cpp
```diff
@@ -146,6 +168,11 @@ void Results::update_tableview()
             m_mode = Mode::TableView;
             break;
         case Mode::TableView:
+            if (!m_background_query && m_realm->can_deliver_notifications() &&  !m_realm->is_in_transaction()) {
```
jpsim: Move the `!m_realm->is_in_transaction()` check up before `m_realm->can_deliver_notifications()` since it's much cheaper? Also, inconsistent spacing around `&&`.

Realm/ObjectStore/results.cpp
```diff
@@ -325,8 +353,69 @@ Results Results::filter(Query&& q) const
...
+    if (!m_background_query) {
+        m_background_query = std::make_shared<_impl::AsyncQuery>(*this);
+        _impl::RealmCoordinator::register_query(m_background_query);
+    }
+    return {m_background_query, m_background_query->add_callback(std::move(target))};
+}
+
+void Results::AsyncFriend::set_table_view(Results& results, realm::TableView &&tv)
+{
+    // If the previous TableView was never actually used, then generating new
```
jpsim: generate?

Realm/ObjectStore/results.cpp
```diff
@@ -325,8 +353,69 @@ Results Results::filter(Query&& q) const
...
+    // ones until the user actually uses the Results object again
+    if (results.m_mode == Mode::TableView) {
+        results.m_wants_background_updates = results.m_has_used_table_view;
+    }
+
+    results.m_table_view = std::move(tv);
+    results.m_mode = Mode::TableView;
+    results.m_has_used_table_view = false;
+    // needs https://github.com/realm/realm-core/pull/1392
```
jpsim: this is now in core as of 0.95.6

Realm/RLMArray.h
```diff
@@ -286,6 +286,25 @@ RLM_ASSUME_NONNULL_BEGIN
 /// :nodoc:
 - (void)setObject:(RLMObjectType)newValue atIndexedSubscript:(NSUInteger)index;
 
+#pragma mark - Notifications
+
+/**
+ Register a block to be called each time the RLMArray changes.
+
+ The block will be asynchronously called with the initial array, and then
+ called again after each writen transaction which changes the array or any
```
jpsim: written

Realm/RLMResults.h
```diff
@@ -156,6 +169,39 @@ RLM_ASSUME_NONNULL_BEGIN
  */
 - (RLMResults RLM_GENERIC_RETURN*)sortedResultsUsingDescriptors:(NSArray *)properties;
 
+#pragma mark - Notifications
+
+/**
+ Register a block to be called each time the RLMResults changes.
+
+ The block will be asynchronously called with the initial results, and then
+ called again after each writen transaction which causes the results to change.
```
jpsim: written

RealmSwift-swift1.2/Results.swift
```diff
@@ -286,6 +286,48 @@ public final class Results<T: Object>: ResultsBase {
...
         return rlmResults.averageOfProperty(property) as! U?
     }
+
+    // MARK: Notifications
+
+    /**
+     Register a block to be called each time the Results changes.
+
+     The block will be asynchronously called with the initial results, and then
+     called again after each writen transaction which causes the results to change.
```
jpsim: written

RealmSwift-swift2.0/Results.swift
```diff
@@ -290,6 +290,48 @@ public final class Results<T: Object>: ResultsBase {
...
         return rlmResults.averageOfProperty(property) as! U?
     }
+
+    // MARK: Notifications
+
+    /**
+     Register a block to be called each time the Results changes.
+
+     The block will be asynchronously called with the initial results, and then
+     called again after each writen transaction which causes the results to change.
```
jpsim: written

Realm/RLMResults.h
```diff
@@ -156,6 +169,39 @@ RLM_ASSUME_NONNULL_BEGIN
...
+ You must retain the returned token for as long as you want the results to
+ continue to be sent to the block. To stop receiving updates, call -stop on the
```
jpsim: '-stop' should be backticked

Realm/RLMArray.h
```diff
@@ -286,6 +286,25 @@ RLM_ASSUME_NONNULL_BEGIN
...
+ items contained in the array. You must retain the returned token for as long as
+ you want the block to continue to be called. To stop receiving updates, call
+ -stop on the token.
```
jpsim: '-stop' should be backticked

Realm/RLMArray.h
```diff
@@ -286,6 +286,25 @@ RLM_ASSUME_NONNULL_BEGIN
...
+
+ The error parameter will always be `nil`, and is present only for compatiblity
+ with the RLMResults version of this method, which can potentially fail.
+
+ @param block The block to be called each time the array changes
```
jpsim: missing terminating period

Realm/RLMCollection.h
```diff
@@ -169,6 +169,16 @@ RLM_ASSUME_NONNULL_BEGIN
  */
 - (void)setValue:(nullable id)value forKey:(NSString *)key;
 
+#pragma mark - Notifications
+
+/**
+ Register a block to be called each time the collection changes.
+
+ @param block The block to be called each time the collection changes
```
jpsim: missing terminating period

Realm/RLMRealm.h
```diff
@@ -505,9 +504,16 @@ typedef void (^RLMMigrationBlock)(RLMMigration *migration, uint64_t oldSchemaVer
...
- Notification token - holds onto the realm and the notification block
+ A token which is returned from methods which subscribe to changes to a Realm.
+
+ Change subscriptions in Realm return an RLMNotificationToken which can be used
+ to unsubscribe from the changes. You must store a strong reference to the token
+ for as long as you want to continue to receive notifications. When you wish to
+ stop, call the `-stop` method.
  */
 @interface RLMNotificationToken : NSObject
+/// Stop receiving notifications for the subcription that returned this token
```
jpsim: missing terminating period

Realm/RLMResults.h
```diff
@@ -156,6 +169,39 @@ RLM_ASSUME_NONNULL_BEGIN
...
+ happen. This will become more precise in future versions.
+
+ If an error occurs the block will be called with `nil` for the results
+ parameter and a non-`nil` error. Currently the only errors that can occur are
+ when opening the RLMRealm on the background worker thread or the destination
+ queue fails.
+
+ At the time when the block is called, the RLMResults object will be fully
+ evaluated and up-to-date, and as long as you do not perform a write transaction
+ on the same thread or explicitly call -[RLMRealm refresh], access it will never
```
jpsim: '-[RLMRealm refresh]' should be backticked

Realm/RLMResults.h
```diff
@@ -156,6 +169,39 @@ RLM_ASSUME_NONNULL_BEGIN
...
+ perform blocking work.
+
+ @warning This method cannot be called during a write transaction, or when the
+          containing realm is read-only.
+
+ @param block The block to be called with the evaluated results
```
jpsim: missing terminating period

RealmSwift-swift1.2/List.swift
```diff
@@ -386,6 +386,24 @@ public final class List<T: Object>: ListBase {
...
+    // MARK: Notifications
+
+    /**
+    Register a block to be called each time the List changes.
+
+    The block will be asynchronously called with the initial list, and then
+    called again after each write transaction which changes the list or any of
+    the items in the list. You must retain the returned token for as long as
+    you want the results to continue to be sent to the block. To stop receiving
+    updates, call stop() on the token.
```
jpsim: 'stop()' should be backticked

RealmSwift-swift1.2/List.swift
```diff
@@ -386,6 +386,24 @@ public final class List<T: Object>: ListBase {
...
+
+    :param: The block to be called each time the list changes
```
jpsim: missing terminating period

RealmSwift-swift1.2/RealmCollectionType.swift
```diff
@@ -652,4 +667,27 @@ public final class AnyRealmCollection<T: Object>: RealmCollectionType {
...
+    // MARK: Notifications
+
+    /**
+    Register a block to be called each time the collection changes.
+
+    The block will be asynchronously called with the initial collection, and
+    then called again after each write transaction which changes the collection
+    or any of the items in the collection. You must retain the returned token for
+    as long as you want updates to continue to be sent to the block. To stop
+    receiving updates, call stop() on the token.
```
jpsim: 'stop()' should be backticked

RealmSwift-swift1.2/RealmCollectionType.swift
```diff
@@ -652,4 +667,27 @@ public final class AnyRealmCollection<T: Object>: RealmCollectionType {
...
+
+    :param: block The block to be called each time the collection changes
```
jpsim: missing terminating period

Realm/ObjectStore/impl/apple/cached_realm.hpp
tgoyne: "on" is correct. A CachedRealm is associated with a specific Realm instance and it dispatches the notification to the thread for that Realm.

Realm/ObjectStore/impl/apple/external_commit_helper.cpp
tgoyne: Things logged to asl don't show up in the output window in Xcode.

jpsim: ah, I didn't know that. thanks

Realm/Tests/ArrayPropertyTests.m
```diff
@@ -854,4 +854,108 @@ - (void)testAssignArrayProperty {
     [realm commitWriteTransaction];
 }
 
+- (void)testNotificationSentInitially {
+    RLMRealm *realm = self.realmWithTestPath;
+    [realm beginWriteTransaction];
+    ArrayPropertyObject *array = [ArrayPropertyObject createInRealm:realm withValue:@[@"", @[], @[]]];
+    [realm commitWriteTransaction];
+
+    id expectation = [self expectationWithDescription:@""];
```
jpsim: `__block`?

Realm/Tests/AsyncTests.m
```diff
@@ -0,0 +1,828 @@
...
+    [IntObject createInDefaultRealmWithValue:@[@0]];
+    [realm commitWriteTransaction];
+
+    [self waitForExpectationsWithTimeout:2.0 handler:nil];
+
+    [token stop];
+    [token2 stop];
+}
+
+- (void)testRLMResultsInstanceIsRedused {
```
jpsim: `testRLMResultsInstanceIsReused`

Realm/Tests/AsyncTests.m
```diff
@@ -0,0 +1,828 @@
...
+   __weak __block RLMResults *prev;
```
jpsim: indentation

Realm/RLMResults.h
```diff
@@ -31,6 +31,22 @@ RLM_ASSUME_NONNULL_BEGIN
  RLMResults can be queried with the same predicates as RLMObject and RLMArray
  and you can chain queries to further filter query results.
 
+ RLMResults always reflect the current state of the Realm on the current thread,
```
jpsim: This doc text is great and should be mirrored in `Results.swift`.

jpsim:
I'm done reviewing. :+1: this is nothing short of fantastic work! Extremely thorough tests. :clap: :100: 

I think your modifications to the examples are sufficient for now. At some point it could be useful to add an example demonstrating how async queries are useful with components as was mentioned in this PR's comments.

We'll need to copy some of the Results docs you added here to the Queries section of the docs.

@bdash did you want to review this further? It'd be nice if you could give this another look.

bigfish24:
Awesome work @tgoyne :clap:

Realm/ObjectStore/impl/async_query.hpp
```diff
@@ -0,0 +1,122 @@
...
+
+    Realm& get_realm() { return m_target_results->get_realm(); }
+    // Get the version of the current handover object
+    SharedGroup::VersionID version() const noexcept { return m_sg_version; }
+
+    bool is_alive() const noexcept;
+
+private:
+    // Target Results to update and a mutex which guards it
+    mutable std::mutex m_target_mutex;
```
bdash: Which members use this mutex for synchronization? The comment mentions only `m_target_results`, but the implementation appears to also use it for `m_unregistered`.

Realm/ObjectStore/impl/async_query.cpp
```diff
@@ -0,0 +1,265 @@
...
+, m_sort(target.get_sort())
+, m_sg_version(Realm::Internal::get_shared_group(*m_realm).get_version_of_current_transaction())
+{
+    Query q = target.get_query();
+    m_query_handover = Realm::Internal::get_shared_group(*m_realm).export_for_handover(q, MutableSourcePayload::Move);
+}
+
+AsyncQuery::~AsyncQuery()
+{
+    std::lock_guard<std::mutex> lock(m_target_mutex);
```
bdash: Why is it necessary to take the mutex here? By the time the destructor is reached, no other clients should have a reference to this object and there isn't anyone else to synchronize with?

Realm/ObjectStore/impl/async_query.cpp
```diff
@@ -0,0 +1,265 @@
...
+    return !m_unregistered;
+}
+
+void AsyncQuery::run()
+{
+    REALM_ASSERT(m_sg);
+
+    // This function must not touch any members touched in deliver(), as they
```
bdash: I'm a little confused by this comment as there are a number of members accessed within this function that are touched in `deliver()` (e.g., `m_target_results`, `m_have_callbacks`, `m_initial_run_complete`). Can you clarify what this comment is trying to say?

Realm/ObjectStore/impl/async_query.cpp
```diff
@@ -0,0 +1,265 @@
...
+        }
+    }
+
+    REALM_ASSERT(!m_tv.is_attached());
+
+    // If we've run previously, check if we need to rerun
+    if (m_initial_run_complete) {
+        // Make an empty tableview from the query to get the table version, since
+        // Query doesn't expose it
+        m_tv = m_query->find_all(0, 0, 0);
```
bdash: Why does this assign to `m_tv` rather than a local temporary? From what I can see all paths that hit this line proceed to assign a new value to `m_tv`.

Realm/ObjectStore/impl/async_query.cpp
```diff
@@ -0,0 +1,265 @@
...
+        // after delivering an error
+        std::lock_guard<std::mutex> callback_lock(m_callback_mutex);
+        m_callbacks.clear();
+    }
+}
+
+std::function<void (std::exception_ptr)> AsyncQuery::next_callback()
+{
+    std::lock_guard<std::mutex> callback_lock(m_callback_mutex);
+    for (++m_callback_index; m_callback_index < m_callbacks.size(); ++m_callback_index) {
```
bdash: If I'm reading this correctly, this is using unsigned integer overflow to get `m_callback_index` from its default value of `npos` into the range of indices that are valid for `m_callbacks`. Is there a reason this approach is used rather than having `call_callbacks` set `m_callback_index` to zero prior to making the first call to `next_callback`?

Realm/ObjectStore/impl/async_query.hpp
```diff
@@ -0,0 +1,122 @@
...
+    Results* m_target_results;
+
+    std::shared_ptr<Realm> m_realm;
+    const SortOrder m_sort;
+    const std::thread::id m_thread_id = std::this_thread::get_id();
+
+    // The source Query, in handover from iff m_sg is null
```
bdash: s/from/form/?

Realm/ObjectStore/impl/realm_coordinator.cpp
bdash: I see now that this is likely a complete comment as there is an implied "hold a readlock".

Realm/ObjectStore/impl/realm_coordinator.cpp
```diff
@@ -0,0 +1,453 @@
...
+    LangBindHelper::advance_read(*m_query_sg, *m_query_history,
+                                 m_advancer_sg->get_version_of_current_transaction());
+
+    // Transfer all new queries over to the main SG
+    for (auto& query : m_new_queries) {
+        query->detatch();
+        query->attach_to(*m_query_sg);
+    }
+
+    if (!m_new_queries.empty()) {
```
bdash: Is it possible for `m_new_queries` to be empty here even though we already checked for that case in the early-return at the start of this function?

Realm/ObjectStore/results.cpp
```diff
@@ -325,8 +353,69 @@ Results Results::filter(Query&& q) const
...
 }
 
-Results::UnsupportedColumnTypeException::UnsupportedColumnTypeException(size_t column, const Table* table) {
+AsyncQueryCancelationToken Results::async(std::function<void (std::exception_ptr)> target)
+{
+    if (m_realm->config().read_only) {
+        throw InvalidTransactionException("Cannot create asynchronous query for read-only Realms");
+    }
+    if (m_realm->is_in_transaction()) {
+        throw InvalidTransactionException("Cannot create asynchronous query while in a write transaction");
```
bdash: Is the lack of async queries on read-only realms and while in a write transaction an inherent limitation, or just a limitation of the initial implementation?

Realm/ObjectStore/results.cpp
```diff
@@ -325,8 +353,69 @@ Results Results::filter(Query&& q) const
...
 }
+
+AsyncQueryCancelationToken::AsyncQueryCancelationToken(std::shared_ptr<_impl::AsyncQuery> query, size_t token)
+: m_query(std::move(query)), m_token(token)
+{
+}
+
+AsyncQueryCancelationToken::~AsyncQueryCancelationToken()
+{
+    if (auto query = std::atomic_load(&m_query)) {
```
bdash: What is it about the use of `AsyncQueryCancelationToken` that necessitates atomic operations on its `shared_ptr` member?

Realm/ObjectStore/shared_realm.hpp
```diff
@@ -101,15 +107,33 @@ namespace realm {
...
         // Realm after closing it will produce undefined behavior.
         void close();
 
         ~Realm();
 
-      private:
+        void init(std::shared_ptr<_impl::RealmCoordinator> coordinator);
         Realm(Config config);
 
+        // Give AsyncQuery direct access to the shared group as it needs it to
```
bdash: The comment mentions only `AsyncQuery` but the code also gives access to `RealmCoordinator`.

Realm/ObjectStore/results.hpp
```diff
@@ -144,6 +168,24 @@ class Results {
...
+    // Create an async query from this Results
+    // The query will be run on a background thread and delivered to the callback,
+    // and then rerun after each commit (if needed) and redelivered if it changed
+    AsyncQueryCancelationToken async(std::function<void (std::exception_ptr)> target);
+
+    bool wants_background_updates() const { return m_wants_background_updates; }
+
+    // Helper type to let AsyncQuery update the tableview without giving access
+    // to any other privates or letting anyone else do so
+    class AsyncFriend {
```
bdash: Should this be called `Internal` like `Realm`'s equivalent?

Realm/ObjectStore/impl/async_query.cpp
tgoyne: If unregister() is called while the query is running then the AsyncQuery will be destructed on the worker thread, so it needs a memory barrier to ensure the writes done in unregister() are visible. The actual lock is unnecessary, but thread sanitizer currently doesn't support explicit memory barriers.

Realm/ObjectStore/impl/async_query.cpp
tgoyne: It'd require locking m_callback_mutex one extra time, but I guess that probably doesn't matter.

Realm/ObjectStore/results.cpp
tgoyne: I think the write transaction limitation is inherent. The read-only limitation is because handover is implemented in SharedGroup (since Group has no concept of versions), and may go away with the Snapshot refactoring.

Realm/ObjectStore/results.cpp
tgoyne: It's required to make it safe to destroy a token from any thread (otherwise there is a data race if -stop is called on the correct thread but it's later deallocated on a different thread), and users have been very confused by things having to be deallocated on a specific thread in the past.

Realm/ObjectStore/results.cpp
jpsim: This is tracked in realm/realm-core#1317

Realm/ObjectStore/impl/async_query.hpp
bdash: This was addressed by the removal of `m_unregistered`.

Realm/Tests/ArrayPropertyTests.m
bdash: Unnecessary since `expectation` is not reassigned within the block.

bdash:
:+1:

jpsim:
Closes #3105

jpsim:
:+1:

PR #2848, opened 2015-11-12
[Results] Use unsafeBitcast instead force-cast when accessing elements

### LINKED ISSUE #2844: Swift dynamic objects access throws exception
Hey Realm folks!

I'm running into an exception when trying to access dynamic objects in a Result object:

```swift
class TestObject: Object {
    dynamic var id = 1
}

let realm = try! Realm()
realm.beginWrite()
let test = TestObject()
realm.add(test)
try! realm.commitWrite()
realm.dynamicObjects("TestObject").first  /// Throws an exception
```

The exception being 

```
Could not cast value of type 'Tests.TestObject' (0x7fad72535cc0) to 'RealmSwift.DynamicObject' (0x10bb57060).
```

It looks to be from these lines in `Result`:

```swift
    /**
    Returns the object at the given `index`.

    - parameter index: The index.

    - returns: The object at the given `index`.
    */
    public subscript(index: Int) -> T {
        get {
            throwForNegativeIndex(index)
            return rlmResults[UInt(index)] as! T
        }
    }

    /// Returns the first object in the results, or `nil` if empty.
    public var first: T? { return rlmResults.firstObject() as! T? }

    /// Returns the last object in the results, or `nil` if empty.
    public var last: T? { return rlmResults.lastObject() as! T? }
```

The `as! T?` here is causing the problem, because anything that subclasses Objects (all Realm objects basically) can't be cast with `as!` to DynamicObject, because DynamicObject (also a subclass) is unrelated to those subclasses.  the only as! that would probably work is to Object.

I looked into tests for dynamicObjects and could only find the one:

```swift
    func testDynamicObjects() {
        try! Realm().write {
            try! Realm().create(SwiftIntObject.self, value: [100])
            try! Realm().create(SwiftIntObject.self, value: [200])
            try! Realm().create(SwiftIntObject.self, value: [300])
        }

        XCTAssertEqual(0, try! Realm().dynamicObjects("SwiftStringObject").count)
        XCTAssertEqual(3, try! Realm().dynamicObjects("SwiftIntObject").count)
        assertThrows(try! Realm().dynamicObjects("Object"))
    }
```
Count here actually works.  So I couldn't find any test for actually accessing dynamic objects from a Results<DynamicObject>.

Is this out of scope for DynamicObjects?  Or is this a bug?

mrackwitz:
Fixes #2844. 
/c @bigfish24

### DIFF
--- a/RealmSwift-swift1.2/Tests/TestCase.swift
+++ b/RealmSwift-swift1.2/Tests/TestCase.swift
@@ -23,6 +23,18 @@ import Realm.Dynamic
 import RealmSwift
 import XCTest
 
+func XCTAssertEqual<T: Any where T: Equatable>(@autoclosure a: () -> T?, @autoclosure b: () -> T?, _ message: String? = nil, file: String = __FILE__, line: UInt = __LINE__) {
+    if let _a = a() {
+        if let _b = b() {
+            XCTAssertEqual(_a, _b, (message != nil ? message! : ""), file: file, line: line)
+        } else {
+            XCTFail((message != nil ? message! : "a != nil, b == nil"), file: file, line: line)
+        }
+    } else if let _ = b() {
+        XCTFail((message != nil ? message! : "a == nil, b != nil"), file: file, line: line)
+    }
+}
+
 func inMemoryRealm(inMememoryIdentifier: String) -> Realm {
     return Realm(configuration: Realm.Configuration(inMemoryIdentifier: inMememoryIdentifier))!
 }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ x.x.x Release notes (yyyy-MM-dd)
 
 ### Bugfixes
 
+* Fix crash when trying to retrieve object instances via `dynamicObjects`.
 * Fix crashes or incorrect results when calling `-[RLMRealm refresh]` during
   fast enumeration.
 * Add `Int8` support for `RealmOptional`, `MinMaxType` and `AddableType`.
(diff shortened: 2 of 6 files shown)


tgoyne:
What type is it bitcasting from in the dynamic case? RLMDynamicObject?

mrackwitz:
That depends on the schema: if there is an accessor class other than `RLMDynamicObject`, which was the case for the filed bug report, then this will be used instead.

jpsim:
I'd like to see more tests for different types, especially `RealmOptionalType`s.

mrackwitz:
Why `RealmOptionalType` specifically? `RealmOptional` isn't really involved in anyway when accessing values through subscripts.

jpsim:
Just testing a single required `Int` property is not sufficient to verify that this works as intended. We should be testing all our supported types, including optionals.

mrackwitz:
@jpsim: Addressed that.

RealmSwift-swift1.2/Tests/TestCase.swift
```diff
@@ -23,6 +23,18 @@ import Realm.Dynamic
...
 import XCTest
 
+func XCTAssertEqual<T: Any where T: Equatable>(@autoclosure a: () -> T?, @autoclosure b: () -> T?, _ message: String? = nil, file: String = __FILE__, line: UInt = __LINE__) {
+    if let _a = a() {
+        if let _b = b() {
+            XCTAssertEqual(_a, _b, (message != nil ? message! : ""), file: file, line: line)
+        } else {
+            XCTFail((message != nil ? message! : "a != nil, b == nil"), file: file, line: line)
+        }
+    } else if let _b = b() {
```
jpsim: don't you get a compiler warning here for not using `_b`? Shouldn't this be `_ = b()` instead?

jpsim:
:+1:

RealmSwift-swift1.2/Tests/TestCase.swift
mrackwitz: No, no compiler warning, but good point.

PR #2858, opened 2015-11-13
Annotate collections with generic types

tgoyne:
I got annoyed at the lack of autocomplete and added generic type annotations to (almost) all of the NSArrays and NSDictionaries in headers and property declarations. This is marked as a breaking change as Swift code using the obj-c API has to be updated in a few places to compile.


jpsim:
:+1: nice. We certainly should have done this sooner.

Issue #2862, opened 2015-11-16
Custom RLMConfiguration with objects shared in multiple realms slow?


vittoriom:
In our app we are using a `RLMRealm` for all the data we need to store. In the new version we're introducing a separate realm to store "favorites". We don't want the realm to have the same big schema the main realm has, so we're using `RLMConfiguration` and the `objectClasses` property to specify which part of the schema we want to bring with us. Keep in mind that the objects are a subset of those that we already have in the bigger realm.

What we noticed is that, even when the new realm is empty/really small (so the issue is not related to the size of the realm), accessing it takes huge amounts of time (in the order of 2.5 seconds to remove an item). Pausing the app with the debugger during one of these accesses showed that this is related to the creation of the `RLMConfiguration` (or the instantiation of the `RLMRealm` with this configuration?). We have to instantiate a realm every time we access it since we want to do that from different threads (queues, to be more specific). With the bigger realm this is not an issue because I think calling `RLMRealm.defaultRealm()` is more optimized/cached.

By using a static `RLMRealm` instance we managed to take the access time down to acceptable, but now we have to run everything on the main queue since this is the only queue that guarantees 1-to-1 mapping to a thread. 

Is there any known issue regarding custom schemas being particularly slow, maybe when the objects are a subset of bigger schemas already in use?

Thanks

vittoriom:
As an update, just having a static `RLMRealmConfiguration` works, meaning the access time is acceptable as well.
Maybe you want to investigate on the time it takes to create a `RLMRealmConfiguration` where the `objectClasses` property contains classes that are also shared with another realm?
Otherwise you can close the issue.

Thanks

tgoyne:
Setting the `objectClasses` property does a lot of work, and creating a single RLMRealmConfiguration that's reused is definitely the intended pattern. The documentation should probably do a much better job of calling this out.

vittoriom:
Hi @tgoyne, thanks for your reply. 
Good to know then, I was a bit worried that having a shared `RLMRealmConfiguration` object could have caused other not-easily-visible issues later on, but I think I'll close this issue now. Thanks again.

PR #2863, opened 2015-11-16
Document that RLMRealmConfiguration can be shared between threads

tgoyne:
See https://github.com/realm/realm-cocoa/issues/2862.


jpsim:
Something similar for the Swift side would also be useful, since it has the same performance implications with `objectTypes`.

jpsim:
@tgoyne I've added a similar comment on the Swift side. Can you please review? ObjC side looks good.

tgoyne:
:+1:

PR #2864, opened 2015-11-16
Fix race condition in multiprocess schema init

tgoyne:
If the schema was initialized by a different process between when the old schema was read and the write transaction was began, the schema init code would see the updated schema version but not re-read the schema, resulting in it thinking that a migration was required when the schema actually matched.

No changelog entry since we haven't shipped a release since when this was broken.

### DIFF
--- a/Realm/ObjectStore/shared_realm.cpp
+++ b/Realm/ObjectStore/shared_realm.cpp
@@ -180,6 +180,7 @@ bool Realm::update_schema(std::unique_ptr<Schema> schema, uint64_t version)
 {
     schema->validate();
 
+    // If the schema version matches, just verify that the schema itself also matches
     bool needs_update = !m_config.read_only && (m_config.schema_version != version || ObjectStore::needs_update(*m_config.schema, *schema));
     if (!needs_update) {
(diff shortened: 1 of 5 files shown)


Realm/ObjectStore/shared_realm.cpp
```diff
@@ -188,39 +189,60 @@ bool Realm::update_schema(std::unique_ptr<Schema> schema, uint64_t version)
...
-                                                             migration_function);
-        commit_transaction();
-        return changed;
-    }
-    catch (...) {
-        if (is_in_transaction()) {
-            cancel_transaction();
+        auto new_realm = shared_from_this();
+        m_config.schema = std::move(schema);
+        m_config.schema_version = version;
```
bdash: Can you clarify why updating `m_config` with the new schema and version was moved into the migration function?

tgoyne: It's a net reduction in the amount of copying-and-shuffing-stuff-around code for dealing with that we need the unmodified config for the old realm in the migration and want everything left unchanged if an error occurs during the migration, but now that I'm looking at the code it's only a net reduction because it's wrong...

bdash:
:+1:

PR #2870, opened 2015-11-18
Add Xcode 7 compatibility to Realm Plugin.

TimOliver:
This PR adds the required UUIDs for Xcode versions 7.0, 7.1 (Including 7.1.1) and 7.2-beta to enable the Realm Plugin to work with those versions.

Tested and confirmed working against Xcode 7.1.1 and Xcode 7.2-beta on OS X El Capitan.

/cc @jpsim @bdash

### DIFF
--- a/plugin/RealmPlugin/RealmPlugin-Info.plist
+++ b/plugin/RealmPlugin/RealmPlugin-Info.plist
@@ -36,6 +36,7 @@
 		<string>9F75337B-21B4-4ADC-B558-F9CADF7073A7</string>
 		<string>7FDF5C7A-131F-4ABB-9EDC-8C5F8F0B8A90</string>
 		<string>0420B86A-AA43-4792-9ED0-6FE0F2B16A13</string>
+		<string>7265231C-39B4-402C-89E1-16167C4CC990</string>
 	</array>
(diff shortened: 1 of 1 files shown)


jpsim:
Comparing this with [Alcatraz's plist](https://github.com/alcatraz/Alcatraz/blob/1.1.13/Alcatraz/Alcatraz-Info.plist), we appear to be missing `E969541F-E6F9-4D25-8158-72DC3545A6C6` and `8DC44374-2B35-4C57-A6FE-2AD66A36AAD9` from our list.

We also have a `XCGCReady` key that they don't, which I think is incorrect, since we've never tested the plugin with garbage collection.

TimOliver:
Fair enough! Those strings [appear to be](http://qiita.com/kotowo/items/39caf5d451f0a538824c) for Xcode 6.3.2 and Xcode 6.4-beta 3 respectively (I was looking at just Xcode 7 for this PR), but it might be best for completeness' sake to add all of them.

Alrighty. I'll remove that key as well!

TimOliver:
Let me know if this all looks good! :)

jpsim:
:+1: thanks Tim.

PR #2871, opened 2015-11-18
Fixes to Objective-C doc comments

samritchie:
I applied the same changes made to the Swift comments where appropriate; other than that it was just a few minor typos etc.

### DIFF
--- a/Realm/RLMRealm.h
+++ b/Realm/RLMRealm.h
@@ -28,7 +28,7 @@ RLM_ASSUME_NONNULL_BEGIN
  database.
 
  Realms can either be stored on disk (see +[RLMRealm realmWithPath:]) or in
- memory (see +[RLMRealm inMemoryRealmWithIdentifier:]).
+ memory (see RLMRealmConfiguration).
 
  RLMRealm instances are cached internally, and constructing equivalent RLMRealm
  objects (with the same path or identifier) multiple times on a single thread
@@ -113,7 +113,7 @@ RLM_ASSUME_NONNULL_BEGIN
 @property (nonatomic, readonly) BOOL inWriteTransaction;
 
 /**
- Returns an `RLMRealmConfiguration` that can be used to create this `RLMRealm` instance.
+ Returns the `RLMRealmConfiguration` that was used to create this `RLMRealm` instance.
  */
 @property (nonatomic, readonly) RLMRealmConfiguration *configuration;
 
@@ -170,7 +170,7 @@ typedef void(^RLMNotificationBlock)(NSString *notification, RLMRealm *realm);
 
  Only one write transaction can be open at a time. Write transactions cannot be
  nested, and trying to begin a write transaction on a `RLMRealm` which is
- already in a write transaction with throw an exception. Calls to
+ already in a write transaction will throw an exception. Calls to
  `beginWriteTransaction` from `RLMRealm` instances in other threads will block
  until the current write transaction completes.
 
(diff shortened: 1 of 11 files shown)


Realm/RLMRealm.h
```diff
@@ -211,13 +209,13 @@ typedef void(^RLMNotificationBlock)(NSString *notification, RLMRealm *realm);
 - (BOOL)commitWriteTransaction:(NSError **)error;
 
 /**
- Revert all writes made in the current write transaction and end the transaction.
+ Revert all writes made in the current write transaction and ends the transaction.
```
jpsim: Should also say "Reverts" to keep the tense consistent.

Realm/RLMRealm.h
```diff
@@ -266,10 +274,14 @@ typedef void(^RLMNotificationBlock)(NSString *notification, RLMRealm *realm);
 
  If set to YES (the default), changes made on other threads will be reflected
  in this Realm on the next cycle of the run loop after the changes are
- committed.  If set to NO, you must manually call -refresh on the Realm to
+ committed.  If set to NO, you must manually call `-refresh` on the Realm to
  update it to get the latest version.
 
- Even with this enabled, you can still call -refresh at any time to update the
+ Note that on background threads, the run loop is not run by default and you will
```
jpsim: Note that background threads do not have a run loop attached by default and you will...

Realm/RLMRealm.h
```diff
@@ -345,12 +357,12 @@ typedef void(^RLMNotificationBlock)(NSString *notification, RLMRealm *realm);
...
 /**
- Adds an object to be persisted it in this Realm.
+ Adds an object to be persisted in this Realm.
 
  Once added, this object can be retrieved using the `objectsWhere:` selectors
- on `RLMRealm` and on subclasses of `RLMObject`. When added, all linked (child)
- objects referenced by this object will also be added to the Realm if they are
- not already in it. If the object or any linked objects already belong to a
+ on `RLMRealm` and on subclasses of `RLMObject`. When added, all (child)
+ relationshipd referenced by this object will also be added to the Realm if they are
```
jpsim: relationships

Realm/RLMRealm.h
```diff
@@ -233,7 +231,7 @@ typedef void(^RLMNotificationBlock)(NSString *notification, RLMRealm *realm);
  but re-running the query which provided `oldObject` will once again return
  the valid object.
 
- Calling this when not in a write transaction will throw an exception.
+ @warning Calling this when not in a write transaction will throw an exception.
```
jpsim: I'd prefer to stay consistent in our write transaction warnings, and I prefer the "This method can only be called during a write transaction." wording.

Realm/RLMRealm.h
samritchie: Apple’s docs are inconsistent (as usual), but it looks like at some point they changed the language from “Every thread has a run loop associated” to “automatically created as needed”. I’ll update this in the Swift docs as well.

samritchie:
Also, if CFRunLoopGetCurrent() is being called on realm creation, there will a run loop attached on every thread with a realm. 

I was originally thinking it’s worth documenting how to run the run loop for autorefresh, however, I’m not sure whether the docs should recommend this vs just calling -refresh directly. I suspect the latter would be preferred unless you were already doing run loop programming in the background for some reason (which is fairly unlikely these days).

bdash:
I don't think it's useful to document how to run the runloop in this context since it's unlikely that one would want to do so solely for the purposes of Realm's auto-refreshing.

samritchie:
Okay, I’ve gone with “by default, background threads do not have an active run loop”, purely because ‘running run loop’ is a bit cumbersome. People know enough about run loops to have one running in a background thread will just skip over this, for everyone else it’s clear they need to call -refresh directly.

jpsim:
:+1: but you'll have to address the merge conflicts. We prefer rebasing against master over merging when these conflicts occur in PRs.

samritchie:
Thanks, I’ve squashed & rebased.

RealmSwift-swift1.2/Realm.swift
```diff
@@ -493,9 +493,9 @@ public final class Realm {
     committed.  If set to `false`, you must manually call `refresh()` on the Realm to
     update it to get the latest version.
 
-    Note that on background threads, the run loop is not run by default and you will 
-	will need to manually call `refresh()` in order to update to the latest version,
-	even if `autorefresh` is set to `true`.
+    Note that by default, background threads do not have an active run loop and you
```
jpsim: can you please remove the trailing space on this line?

RealmSwift-swift2.0/Realm.swift
```diff
@@ -467,10 +467,10 @@ public final class Realm {
     in this Realm on the next cycle of the run loop after the changes are
     committed.  If set to `false`, you must manually call `refresh()` on the Realm to
     update it to get the latest version.
-
-	Note that on background threads, the run loop is not run by default and you will
-	will need to manually call `refresh()` in order to update to the latest version,
-	even if `autorefresh` is set to `true`.
+	
+    Note that by default, background threads do not have an active run loop and you
```
jpsim: can you please remove the trailing space on this line?

jpsim:
closing in favor of #2888 (just trailing whitespace removed). I picked it up because it's now the weekend in @samritchie's Australia.

PR #2874, opened 2015-11-18
Update to core 0.95.0

### LINKED ISSUE #1885: It should be possible to debug an application using encrypted realms
We disabled debugging of applications using encrypted realms to address a frequent LLDB deadlock that would occur (issue #1625). We should address the underlying issue that causes the hang and reinstate the ability to debug applications that use encrypted realms. This will likely require changes in core.

### LINKED ISSUE #2827: To-Many relationships performance
I am not sure if I do something wrong or is it supposed to work like that, but just having a `List` declared in the `Object`, is hitting the performance quite visibly.

Code:
```swift
class Person: Object {
    dynamic var name: String = ""
    let locations = List<Location>()
}

class Location: Object {
    dynamic var latitude: Double = 0.0
    dynamic var longitude: Double = 0.0
}

let beginDate = NSDate()

var objects = [Person]()

for var i = 0; i < 70000; i++ {
    let person = Person()
    person.name = "Name Surname"
    objects.append(person)
}

do {
    let realm = try Realm()
    
    try realm.write({
        realm.add(objects)
    })
} catch let error as NSError {
    print("ERROR: \(error)")
}

let endDate = NSDate()
let timeDifference = endDate.timeIntervalSince1970 - beginDate.timeIntervalSince1970

print("DIFFERENCE: \(timeDifference)")
```
Result:
```
DIFFERENCE: 29.7714397907257
```

Code:
```swift
class Person: Object {
    dynamic var name: String = ""
//    let locations = List<Location>()
}

class Location: Object {
    dynamic var latitude: Double = 0.0
    dynamic var longitude: Double = 0.0
}

let beginDate = NSDate()

var objects = [Person]()

for var i = 0; i < 70000; i++ {
    let person = Person()
    person.name = "Name Surname"
    objects.append(person)
}

do {
    let realm = try Realm()
    
    try realm.write({
        realm.add(objects)
    })
} catch let error as NSError {
    print("ERROR: \(error)")
}

let endDate = NSDate()
let timeDifference = endDate.timeIntervalSince1970 - beginDate.timeIntervalSince1970

print("DIFFERENCE: \(timeDifference)")
```
Result:
```
DIFFERENCE: 0.884682893753052
```

To speed things up:
```
ProductName:	Mac OS X
ProductVersion:	10.11.1
BuildVersion:	15B42

Xcode 7.1
Build version 7B91b

/usr/local/bin/pod
0.39.0
Realm (0.96.2)
RealmSwift (0.96.2)

/bin/bash


(not in use here)

/usr/bin/git
git version 2.4.9 (Apple Git-60)
```

tgoyne:
Closes #1885. Closes #2827.

### DIFF
--- a/Realm/RLMRealm.mm
+++ b/Realm/RLMRealm.mm
@@ -89,9 +89,6 @@ NSData *RLMRealmValidatedEncryptionKey(NSData *key) {
         if (key.length != 64) {
             @throw RLMException(@"Encryption key must be exactly 64 bytes long");
         }
-        if (RLMIsDebuggerAttached()) {
-            @throw RLMException(@"Cannot open an encrypted Realm with a debugger attached to the process");
-        }
 #if TARGET_OS_WATCH
         @throw RLMException(@"Cannot open an encrypted Realm on watchOS.");
 #endif
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,12 +11,17 @@ x.x.x Release notes (yyyy-MM-dd)
 * Queries are no longer limited to 16 levels of grouping.
 * The block parameter of `-[RLMRealm transactionWithBlock:]`/`Realm.write(_:)` is 
   now marked as `__attribute__((noescape))`/`@noescape`.
+* Rework the implementation of encrypted Realms to no longer interfere with
+  debuggers.
 
 ### Bugfixes
 
 * Fix crashes or incorrect results when calling `-[RLMRealm refresh]` during
   fast enumeration.
 * Add `Int8` support for `RealmOptional`, `MinMaxType` and `AddableType`.
+* Set the default value for newly added non-optional NSData properties to a
+  zero-byte NSData rather than nil.
+* Fix a potential crash when deleting all objects of a class.
 
 0.96.2 Release notes (2015-10-26)
 =============================================================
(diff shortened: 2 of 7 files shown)


Realm/RLMRealm.mm
```diff
@@ -89,9 +89,6 @@ static bool shouldForciblyDisableEncryption() {
         if (key.length != 64) {
             @throw RLMException(@"Encryption key must be exactly 64 bytes long");
         }
-        if (RLMIsDebuggerAttached()) {
-            @throw RLMException(@"Cannot open an encrypted Realm with a debugger attached to the process");
-        }
 #if TARGET_OS_WATCH
         @throw RLMException(@"Cannot open an encrypted Realm on watchOS.");
 #endif
```
bdash: I think the switch away from Mach exceptions should allow encrypted Realms to work on watchOS and tvOS now?

bdash:
Also fixes #2827.

bdash:
:+1:

jpsim:
It'd be nice to add a changelog entry for #2827. Otherwise :+1: 

watchOS encryption support should be added in another PR IMO, since we'll need to test that before claiming it all works.

tgoyne:
> watchOS encryption support should be added in another PR IMO, since we'll need to test that before claiming it all works.

Agreed; it almost certainly works but we should still verify that first.

PR #2877, opened 2015-11-18
Validate schema changes made from other processes

tgoyne:
When multiple processes are sharing a single Realm file, they can potentially have different schema definitions that may or may not be compatible. This makes differences which do not require a schema version bump (i.e. one process having more or fewer tables than another process or disagreements about whether a property is indexed) explicitly supported, and throws an exception when incompatible schemas are encountered rather than crashing or silently corrupting data.

### DIFF
--- a/Realm/ObjectStore/impl/transact_log_handler.cpp
+++ b/Realm/ObjectStore/impl/transact_log_handler.cpp
@@ -43,6 +43,10 @@ class TransactLogHandler {
     // Change information for the currently selected LinkList, if any
     ColumnInfo* m_active_linklist = nullptr;
 
+    // Tables which were created during the transaction being processed, which
+    // can have columns inserted without a schema version bump
+    std::vector<size_t> m_new_tables;
+
     // Get the change info for the given column, creating it if needed
     static ColumnInfo& get_change(ObserverState& state, size_t i)
     {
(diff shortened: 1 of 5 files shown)


Realm/ObjectStore/impl/transact_log_handler.cpp
```diff
@@ -27,7 +27,110 @@
 using namespace realm;
 
 namespace {
-class TransactLogHandler {
+// A transaction log handler that just validates that all operations made are
+// onces supported by the object store
```
jpsim: s/onces/ones/

Realm/ObjectStore/impl/transact_log_handler.cpp
```diff
@@ -27,7 +27,110 @@
...
+    // allowed
+    bool add_search_index(size_t) { return true; }
+    bool remove_search_index(size_t) { return true; }
+
+    // Creating entirely new tables without a schema version bump is allowed, so
+    // we need to track if new columns are being added to a new table or an
+    // existing one
+    bool insert_group_level_table(size_t table_ndx, size_t, StringData)
+    {
+        m_new_tables.push_back(table_ndx);
```
bdash: Inserting a table at `table_ndx` suggests that the index of each table that had an index greater than that will have their index incremented. Depending on the order that calls are made, this could result in the indexes in `m_new_tables` becoming out of sync with the index the tables end up at.

Realm/ObjectStore/impl/transact_log_handler.cpp
```diff
@@ -82,56 +187,74 @@ class TransactLogHandler {
...
+
+    // Throw an exception if the currently modified table already existed before
+    // the current set of modifications
+    bool schema_error_unless_new_table()
+    {
+        if (std::find(begin(m_new_tables), end(m_new_tables), current_table()) == end(m_new_tables)) {
+            schema_error();
+        }
+        return true;
+    }
```
bdash: Why are `schema_error` and `schema_error_unless_new_table` duplicated in the subclass?

Realm/ObjectStore/impl/transact_log_handler.cpp
tgoyne: I wrote a test for this and it turns out that replication of non-end-insertions is broken in core.

tgoyne: https://github.com/realm/realm-core/pull/1344

bdash:
:+1:

PR #2881, opened 2015-11-19
Remove the exception TLS slot workaround in Swift tests

tgoyne:
https://github.com/realm/realm-core/pull/1339 fixed the leak of TLS slots in debug builds, so it's no longer needed.


bdash:
:+1:

PR #2883, opened 2015-11-19
Fix memory leak when using Object(value:) for subclasses with List/RealmOptional properties

### LINKED ISSUE #2882: Creating model object with init(value:_) causes memory leak with To-Many relationship
I have two models such as these:

```swift
class Person: Object {
    dynamic var name:String = ""
}

class Dog: Object {
    dynamic var name = ""
    dynamic var age = 0
    let owners = List<Person>()
}
```

Simply creating a dog object like this:

```swift
let myOtherDog = Dog(value: ["name" : "Pluto", "age": 3])
```
causes instruments to report a leak of List<Person> and RLMArray 
screen shot 2015-11-19 at 10 22 59 am

Instruments does not report a leak when just creating an object using init such as 

```swift 
let myDog = Dog()
```

Example Project: https://www.dropbox.com/s/kckg4xqv4uhstd3/RealmSwiftLeak.zip?dl=0

Info:
Realm (0.96.2)
RealmSwift (0.96.2)
OSX 10.10.5
XCode 7.1
CocoaPods 0.39.0

tgoyne:
Calling `[self init]` from `[RLMObjectBase initWithValue:schema:]` ends up
calling a second non-convenience Swift initializer on the object, which means
that all of the properties get initialized a second time and the old values are
leaked.

Closes #2882.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,8 @@ x.x.x Release notes (yyyy-MM-dd)
 * Fix a potential crash when deleting all objects of a class.
 * Fix performance problems when creating large numbers of objects with
   `RLMArray`/`List` properties.
+* Fix memory leak when using Object(value:) for subclasses with
+  `List` or `RealmOptional` properties.
 
 0.96.2 Release notes (2015-10-26)
 =============================================================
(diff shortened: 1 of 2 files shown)


bdash:
:+1:

PR #2886, opened 2015-11-20
Bug fix for missing parameter FormatUpgradeRequired exception constructor

TimOliver:
As a follow-up to https://github.com/realm/realm-cocoa/pull/2800, the CI was failing because a constructor was missing one of its required parameters.


bdash:
Why're there three commits on the branch? Only e78121a needs to be here, the two merge commits seem bogus.

bdash:
You should be able to drop the merge commits by doing:

    git rebase origin/master
    git push origin +to-file-upgrade-checker

TimOliver:
Thanks for that Mark! I was just looking into how to do that!

bdash:
:+1:

PR #2887, opened 2015-11-20
add ci-pr task to build.sh

jpsim:
This moves most of our Jenkins PR job configuration into this repo. /cc @bdash @tgoyne

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -801,6 +801,39 @@ case "$COMMAND" in
         ;;
 
     ######################################
+    # Continuous Integration
+    ######################################
+
+    "ci-pr")
+        mkdir -p build/reports
+
+        if [ "$target" = "docs" ]; then
+            ./build.sh set-swift-version
+            ./build.sh verify-docs
+        elif [ "$target" = "swiftlint" ]; then
+            ./build.sh verify-swiftlint
+        else
+            sha=$ghprbSourceBranch
+            REALM_SWIFT_VERSION=$swift_version
+            CONFIGURATION=$configuration
+            REALM_EXTRA_BUILD_ARGUMENTS='GCC_GENERATE_DEBUGGING_SYMBOLS=NO REALM_PREFIX_HEADER=Realm/RLMPrefix.h'
+            NSUnbufferedIO=YES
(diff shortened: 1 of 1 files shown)


jpsim:
I canceled the PR job because it wouldn't test this change. I'll run again with CI running `./build.sh ci-pr` once this gets reviewed.

bdash:
Could `NSUnbufferedIO=YES` be set unconditionally at the top of build.sh so that it applies to all invocations of `xcodebuild`, rather than just the ones for this CI command?

build.sh
```diff
@@ -801,6 +801,39 @@ case "$COMMAND" in
...
 
     ######################################
+    # Continuous Integration
+    ######################################
+
+    "ci-pr")
+        mkdir -p build/reports
+
+        if [ "$target" = "docs" ]; then
+            ./build.sh set-swift-version
```
tgoyne: We invoke calls to build.sh as `sh build.sh` everywhere else in this file.

build.sh
```diff
@@ -801,6 +801,39 @@ case "$COMMAND" in
...
+            ./build.sh verify-docs
+        elif [ "$target" = "swiftlint" ]; then
+            ./build.sh verify-swiftlint
+        else
+            sha=$ghprbSourceBranch
```
tgoyne: This doesn't appear to be used, and shouldn't be needed (since we need to have already downloaded the correct version to be running this...)

build.sh
```diff
@@ -801,6 +801,39 @@ case "$COMMAND" in
...
+            REALM_SWIFT_VERSION=$swift_version
```
tgoyne: These need to be exported to actually do anything as they're used from child processes.

build.sh
jpsim: The `sha` env variable is used by the CocoaPods installation examples to target the appropriate branch. #2808 will also use `sha` for Carthage examples.

jpsim:
Addressed all PR feedback.

tgoyne:
:+1:

jpsim:
I've updated the job config in Jenkins to use this and am re-running. Will merge if it works out.

PR #2892, opened 2015-11-23
Throw an exception when notifications blocks are added outside of a run loop

tgoyne:
The notification will never actually be called unless the user manually runs a run loop on the same thread afterwards, which is unlikely to be what the user actually expects.


jpsim:
awesome :+1:

PR #2895, opened 2015-11-24
Add support for comparison with key paths on both sides of the operator

bdash:
Adds support for comparisons of the form `class.unary OP property`, `class.unary OP class.unary`, and `class.many.@AGG.property OP property`, where `unary` is a unary relationship, and `many` is an `RLMArray` / `List` relationship.

The following are not part of this pull request:
* `ANY` / `SOME` / `ALL` / `NONE` with key paths on both sides of the operator.
* `class.many.@AGG.property OP class.many.@AGG.property`.
* `IN` with key paths on both sides.
* `BETWEEN` with key paths on both sides.

Depends on realm/realm-core#1360.

### DIFF
--- a/Realm/RLMQueryUtil.mm
+++ b/Realm/RLMQueryUtil.mm
@@ -204,7 +204,7 @@ public:
     NSUInteger column_index() const { return m_column->index(); }
     RLMPropertyType column_type() const { return m_column->type(); }
 
-    void validate_value(id value) const {
+    void validate_comparison(id value) const {
         switch (m_type) {
             case Count:
             case Average:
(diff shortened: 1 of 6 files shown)


anlaital:
:+1: :)

Realm/RLMQueryUtil.mm
```diff
@@ -139,37 +173,38 @@ void set_table() override {}
         Average,
     };
 
-    CollectionOperation(Type type, RLMProperty *linkProperty, RLMProperty *property)
+    CollectionOperation(Type type, ColumnReference link_column, util::Optional<ColumnReference> column)
         : m_type(type)
-        , m_link_property(linkProperty)
-        , m_property(property)
+        , m_link_column(link_column)
```
tgoyne: move these rather than copying?

Realm/RLMQueryUtil.mm
```diff
@@ -343,24 +393,24 @@ void validate_and_extract_between_range(id value, RLMProperty *prop, id *from, i
 void add_constraint_to_query(realm::Query &query, RLMPropertyType type,
                              NSPredicateOperatorType operatorType,
                              NSComparisonPredicateOptions predicateOptions,
-                             const std::vector<NSUInteger>& linkColumns, L lhs, R rhs);
+                             L lhs, R rhs);
 
-void add_between_constraint_to_query(realm::Query &query, std::vector<NSUInteger> const& indexes, RLMProperty *prop, id value) {
+void add_between_constraint_to_query(realm::Query &query, ColumnReference column, id value) {
```
tgoyne: This doesn't appear to need to take column by value.

tgoyne:
Looks good other than the handful of places that appear to be copying ColumnReferences unnecessarily. It's always nice when adding functionality ends up simplifying the code overall rather than complicating it.

Realm/RLMQueryUtil.mm
```diff
@@ -377,15 +427,18 @@ void add_between_constraint_to_query(realm::Query &query, std::vector<NSUInteger
...
     }
 }
 
 template<typename T>
 void add_binary_constraint_to_query(realm::Query & query,
                                     NSPredicateOperatorType operatorType,
-                                    NSUInteger index,
+                                    const ColumnReference& column,
                                     T value) {
+    RLMPrecondition(!column.has_links(), @"Unsupported operator", @"NSData properties cannot be queried over an object link.");
```
jpsim: Shouldn't this message read "Multi-level object equality link queries are not supported."?

jpsim: Oops, actually I think this is correct after all.

jpsim:
What an improvement! :+1: 

Needs a changelog entry.

PR #2899, opened 2015-11-24
Read-only Realm should be opened even in immutable directory

kishikawakatsumi:
Realm files in an app bundle should be opened if marked as read-only.
However, to open Realm in an app bundle will fail due to try to create a `*.note` file in a bundle.

This behavior introduced from the following commit:
https://github.com/realm/realm-cocoa/commit/9cfaa500a96482752c2536282644585ebef4c68f

This test (opening read-only Realm in app bundle) hits https://github.com/realm/realm-cocoa/blob/master/Realm/ObjectStore/impl/apple/external_commit_helper.cpp#L117 then throws exception.

cc @tgoyne @jpsim

### DIFF
--- a/Realm/Tests/RealmTests.mm
+++ b/Realm/Tests/RealmTests.mm
@@ -91,6 +91,32 @@ extern "C" {
     [NSFileManager.defaultManager setAttributes:@{NSFileImmutable: @NO} ofItemAtPath:RLMTestRealmPath() error:nil];
 }
 
+- (void)testReadOnlyFileInImmutableDirectory {
+    @autoreleasepool {
+        RLMRealm *realm = self.realmWithTestPath;
+        [realm beginWriteTransaction];
+        [StringObject createInRealm:realm withValue:@[@"a"]];
+        [realm commitWriteTransaction];
+    }
+
+    // Delete `*.lock` and `.note` files to simulate opening Realm in an app bundle
+    NSString *testRealmPath = RLMTestRealmPath();
+    [[NSFileManager defaultManager] removeItemAtPath:[testRealmPath stringByAppendingString:@".lock"] error:nil];
+    [[NSFileManager defaultManager] removeItemAtPath:[testRealmPath stringByAppendingString:@".note"] error:nil];
+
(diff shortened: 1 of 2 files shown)


kishikawakatsumi:
When read-only Realm, it doesn't use the notifier. Are these changes appropriate?

cc @tgoyne @jpsim

Realm/Tests/RealmTests.mm
```diff
@@ -91,6 +91,40 @@ - (void)testReadOnlyFile {
...
+    // Make parent directory immutable to simulate opening Realm in an app bundle
+    NSString *parentDirectoryOfTestRealmPath = [RLMTestRealmPath() stringByDeletingLastPathComponent];
+    [NSFileManager.defaultManager setAttributes:@{NSFileImmutable: @YES} ofItemAtPath:parentDirectoryOfTestRealmPath error:nil];
+
+    RLMRealm *realm;
+    // Read-only Realm should be opened even in immutable directory
+    XCTAssertNoThrow(realm = [self readOnlyRealmWithPath:RLMTestRealmPath() error:nil]);
+
+    XCTestExpectation *createdRealmInSubThread = [self expectationWithDescription:@"Instantiate Realm in sub thread"];
+    dispatch_async(dispatch_get_global_queue(DISPATCH_QUEUE_PRIORITY_DEFAULT, 0), ^{
```
tgoyne: This can just be `[self dispatchAsyncAndWait:^{ [self readOnlyRealmWithPath:RLMTestRealmPath() error:nil]; }]`, which also handles ensuring that the RLMRealm created on the background thread is deallocated before the test ends.

kishikawakatsumi: Cool. Thanks!

jpsim:
This needs a changelog entry, but otherwise this is good :+1: thanks @kishikawakatsumi!

kishikawakatsumi:
@jpsim I'm just wondering,  does changelog entry make users confuse? Because 9cfaa50 hasn't been released yet. It's been just merged to master.

jpsim:
Ah yes, you're right. I hadn't noticed we hadn't shipped this regression yet. No need for change log 👍

Issue #2754, opened 2015-10-27
[Swift] Copying object to another realm doesn't copy nested relationships


allbto:
I'm trying to copy an Object from one Realm to another using the method `Realm.create(type:, value:, update:)`

My database schema looks like this:

```
Device {
    ...
    List<Mood>
    ...
}

Mood {
    ...
    List<Schedule>
    ...
}

Schedule {
    ...
}
```

So I have a Device contained in my default realm, and here is how I try to copy it to another realm:

```swift
// let device: Device = ...

let editRealm = try! Realm(configuration: Realm.Configuration(inMemoryIdentifier: "AnIdentifier"))

editRealm.beginWrite()
let editDevice = editRealm.create(Device.self, value: device, update: true)
editRealm.commitWrite()
```

What happens is that I get the list of `Mood` in my `editDevice` but the `Mood` have an empty list of `Schedule`. They contained a few `Schedule` each on the main Realm.

Am I doing something wrong ? Should I implement a custom copy method for my `Device` ?

XCode 7.0.1, Swift 2.0, Realm 0.95.1

bigfish24:
Are all of the nested objects persisted?

allbto:
@bigfish24 Yes, I just checked, they belong to a realm and an objectSchema, they even appear on Realm Browser.

bigfish24:
@allbto could you provide a sample project? The reason being is that internally we should be handling nested relationships. The process should be working like this:

1. Swift `create()`
2. ObjC++ `RLMCreateObjectInRealmWithValue`
3. ObjC++ `RLMDynamicSet` when setting the property values
4. ObjC++ `RLMGetLinkedObjectForValue` to retrieve/create the linked object(s) for the property
5. ObjC++ `RLMCreateObjectInRealmWithValue` to create the linked object(s) for the property

If you can't provide any code, you can verify this path is occurring from the `create()` call.

TimOliver:
Hi @allbto!

Just touching base on this issue; are you able to provide us with a sample project to test this functionality?

Thanks! :)

smpanaro:
Hi, I think I might be experiencing a similar problem.

```swift
class Task {
    dynamic var name: String? = nil
    let steps = List<Step>()
}

class Step {
    dynamic var name: String? = nil
    let duration = RealmOptional<NSTimeInterval>()
}
```

[...]

PR #2901, opened 2015-11-25
Fixes issue where KVC fails while copying a Swift object with nested relationships from one Realm to another

bigfish24:
Without initializing the underlying ivar for a list property on an object contained within a list on its parent object, KVC will simply return a blank list.

Fixes: https://github.com/realm/realm-cocoa/issues/2754#issuecomment-159322825

### DIFF
--- a/Realm.xcodeproj/xcshareddata/xcschemes/Realm.xcscheme
+++ b/Realm.xcodeproj/xcshareddata/xcschemes/Realm.xcscheme
@@ -52,6 +52,16 @@
                ReferencedContainer = "container:Realm.xcodeproj">
             </BuildableReference>
          </TestableReference>
+         <TestableReference
+            skipped = "NO">
+            <BuildableReference
+               BuildableIdentifier = "primary"
+               BlueprintIdentifier = "5D660FD71BE98C7C0021E04F"
+               BuildableName = "RealmSwift Tests.xctest"
+               BlueprintName = "RealmSwift Tests"
+               ReferencedContainer = "container:Realm.xcodeproj">
+            </BuildableReference>
+         </TestableReference>
       </Testables>
       <MacroExpansion>
          <BuildableReference
--- a/Realm.xcodeproj/xcshareddata/xcschemes/Realm.xcscheme
+++ b/Realm.xcodeproj/xcshareddata/xcschemes/Realm.xcscheme
@@ -52,16 +52,6 @@
                ReferencedContainer = "container:Realm.xcodeproj">
             </BuildableReference>
          </TestableReference>
-         <TestableReference
-            skipped = "NO">
-            <BuildableReference
-               BuildableIdentifier = "primary"
-               BlueprintIdentifier = "5D660FD71BE98C7C0021E04F"
-               BuildableName = "RealmSwift Tests.xctest"
-               BlueprintName = "RealmSwift Tests"
-               ReferencedContainer = "container:Realm.xcodeproj">
-            </BuildableReference>
-         </TestableReference>
       </Testables>
       <MacroExpansion>
          <BuildableReference
--- a/Realm/RLMObjectStore.mm
+++ b/Realm/RLMObjectStore.mm
@@ -115,6 +115,35 @@ void RLMInitializeSwiftAccessorGenerics(__unsafe_unretained RLMObjectBase *const
     }
 }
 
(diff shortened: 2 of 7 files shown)


bigfish24:
@jpsim can you review?

Realm.xcodeproj/xcshareddata/xcschemes/Realm.xcscheme
```diff
@@ -52,6 +52,16 @@
...
+         <TestableReference
+            skipped = "NO">
+            <BuildableReference
+               BuildableIdentifier = "primary"
+               BlueprintIdentifier = "5D660FD71BE98C7C0021E04F"
+               BuildableName = "RealmSwift Tests.xctest"
+               BlueprintName = "RealmSwift Tests"
+               ReferencedContainer = "container:Realm.xcodeproj">
+            </BuildableReference>
+         </TestableReference>
```
bdash: I don't think this should be here. It's intended that the RealmSwift tests are run only when the RealmSwift scheme is selected.

Realm/RLMObjectStore.mm
```diff
@@ -108,6 +108,35 @@ void RLMInitializeSwiftAccessorGenerics(__unsafe_unretained RLMObjectBase *const
...
+        if (prop.type == RLMPropertyTypeArray) {
+            RLMArray *array = [RLMArrayLinkView arrayWithObjectClassName:prop.objectClassName
+                                                                    view:object->_row.get_linklist(prop.column)
+                                                                   realm:object->_realm
+                                                                     key:prop.name
+                                                            parentSchema:object->_objectSchema];
+            [RLMObjectUtilClass(YES) initializeListProperty:object property:prop array:array];
+        }
+        else if (auto ivar = prop.swiftIvar) {
+            auto optional = static_cast<RLMOptionalBase *>(object_getIvar(object, ivar));
```
bdash: It looks like you based this on an older version of `RLMInitializeSwiftAccessorGenerics` [than the one in master](https://github.com/realm/realm-cocoa/blob/c5b7519953fd2c76c34f13740b15239d9092d2a4/Realm/RLMObjectStore.mm#L106).

bigfish24: Oops I will fix!

Realm.xcodeproj/xcshareddata/xcschemes/Realm.xcscheme
bigfish24: I will remove, thanks!

jpsim:
There are several SwiftLint violations here which should be addressed:

```
ObjectCreationTests.swift:392:132: warning: Comma Spacing Violation: One space before and no after must be present next to any comma. (comma)
ObjectCreationTests.swift:400:42: warning: Comma Spacing Violation: One space before and no after must be present next to any comma. (comma)
ObjectCreationTests.swift:377: warning: Trailing Whitespace Violation: Lines should not have trailing whitespace. (trailing_whitespace)
ObjectCreationTests.swift:390: warning: Trailing Whitespace Violation: Lines should not have trailing whitespace. (trailing_whitespace)
ObjectCreationTests.swift:394: warning: Trailing Whitespace Violation: Lines should not have trailing whitespace. (trailing_whitespace)
ObjectCreationTests.swift:398: warning: Trailing Whitespace Violation: Lines should not have trailing whitespace. (trailing_whitespace)
ObjectCreationTests.swift:392: warning: Line Length Violation: Line should be 120 characters or less: currently 161 characters (line_length)
```

While we continue to support Swift 1.2, unit tests should be mirrored there as well, so please include the `testCreateWithDeeplyNestedObjectsFromAnotherRealm` test there.

This also warrants a CHANGELOG entry.

Otherwise it seems to be the right fix for this issue, thanks!

jpsim:
Since @bigfish24 is away, I'll update this.

tgoyne:
This does a lot of pointless extra work and creates a new RLMArray on each read.

tgoyne:
This is not the correct fix for the problem. The root issue is that `RLMInitializeSwiftAccessorGenerics()` is never called on objects which are returned from fast-enumerating RLMArray/RLMResults. `RLMGenerator` handles this for enumeration from Swift, but in this case the RLMArray is being enumerated from obj-c.

bdash:
Closing in favor of #3203.

PR #2906, opened 2015-11-25
Mirror RLMListBase for Swift Results with RLMResultsBase

bigfish24:
This means that the underlying `RLMResults` is available if you `import Realm.Private` module. Makes it easier for a component developer to create common code that operates on `RLMResults` and reuse for in a Realm Swift API version of the library.

### DIFF
--- a/Realm/RLMResultsBase.h
+++ b/Realm/RLMResultsBase.h
@@ -1,0 +1,21 @@
+//
+//  RLMResultsBase.h
+//  Realm
+//
+//  Created by Adam Fish on 11/25/15.
+//  Copyright © 2015 Realm. All rights reserved.
+//
+
+#import <Foundation/Foundation.h>
+
+@class RLMResults;
+
+// A base class for Swift generic Results to make it possible to interact with
+// them from obj-c
+@interface RLMResultsBase : NSObject <NSFastEnumeration>
+
+@property (nonatomic, strong) RLMResults *rlmResults;
+
+- (instancetype)initWithResults:(RLMResults *)rlmResults;
+
+@end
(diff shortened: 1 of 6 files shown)


bigfish24:
@jpsim and @tgoyne any objections to this?

tgoyne:
Components should not be using Realm.Private.

Realm/RLMResultsBase.h
```diff
@@ -0,0 +1,21 @@
...
+//  Copyright © 2015 Realm. All rights reserved.
+//
+
+#import <Foundation/Foundation.h>
+
+@class RLMResults;
+
+// A base class for Swift generic Results to make it possible to interact with
+// them from obj-c
+@interface RLMResultsBase : NSObject <NSFastEnumeration>
```
jpsim: This will cause `RLMResultsBase` to be part of the public API in both Objective-C and Swift, along with the two members declared here (`rlmResults` and `initWithResults:`). It's unfortunately necessary for `RLMListBase`, but avoidable here.

jpsim:
I understand the desire for a language-agnostic interface to Realm, but I don't think this is the solution.

Something like what we do for `Realm.RLMObjectUtil`/`RealmSwift.ObjectUtil` might be another option.

Or even a clearly namespaced, `@objc`-marked set of getters in Realm Swift:

```swift
@objc public class ObjectiveCCompability {
  public func rlmResults(results: Results) -> RLMResults { ... }
  public func rlmArray(list: List) -> RLMArray { ... }
  public func rlmCollection(collection: RealmCollectionType) -> RLMCollection { ... }
}
```

This would still be part of Realm Swift's public API, but would be clearly marked, like our existing dynamic* methods.

bigfish24:
@jpsim I like this idea because this would aggregate all of the functionality together, including accessing the `RLMRealm` for a `Realm`

tgoyne:
The `ObjectiveCCompability` idea sounds reasonable.

Another option would be to implement `_ObjectiveCBridgeable` for all of the Swift types which would let you just do `Realm() as RLMRealm`. Main downside is that it may change in the future.

jpsim:
Closing in favor of either `ObjectiveCCompatibility` or `_ObjectiveCBridgeable`, although I'd rather not rely on the private implementation details of a protocol.

PR #2924, opened 2015-12-02
Update existing TableViews in first() and last()

tgoyne:
The call to update_tableview() was in the wrong place, resulting in it creating the table view, but not updating an existing one. No changelog entry since it has not been shipped in a release.


bdash:
:+1:

Issue #2934, opened 2015-12-03
Move methods with `args:(va_list)args` to public interface


Rusik:
I want to make a default realm to my model classes to write just `[Model objectsWhere:...]` instead of `[Model objectsInRealm:[Model realmForModel] where:...]` every time in my code. So i have to override `+ (RLMResults *)objectsWhere:(NSString *)predicateFormat, ... ` in my model classes but i can't do it because can't just call super for method with variable arguments, i need `+ (RLMResults *)objectsInRealm:(RLMRealm *)realm where:(NSString *)predicateFormat args:(va_list)args;` for this.

Rusik:
Almost every class in Cocoa have this in public, for example:
```
// NSPredicate.h
+ (NSPredicate *)predicateWithFormat:(NSString *)predicateFormat argumentArray:(nullable NSArray *)arguments;
+ (NSPredicate *)predicateWithFormat:(NSString *)predicateFormat, ...;
+ (NSPredicate *)predicateWithFormat:(NSString *)predicateFormat arguments:(va_list)argList;
```
```
// NSString.h
- (instancetype)initWithFormat:(NSString *)format, ... NS_FORMAT_FUNCTION(1,2);
- (instancetype)initWithFormat:(NSString *)format arguments:(va_list)argList NS_FORMAT_FUNCTION(1,0);
```

jpsim:
Thanks for the request, @Rusik. I think I agree with you that this is useful to add to our public interface. I'll make a PR shortly to do this.

Rusik:
Great, thank you!

PR #2938, opened 2015-12-04
made va_list methods public

jpsim:
to allow other variadic methods to forward their arguments and to mirror what's done in Foundation APIs.

This addresses #2934. I've marked the methods as `:nodoc:` so they won't be included in our HTML documentation because I don't think they should be.

### DIFF
--- a/Realm/RLMQueryUtil.hpp
+++ b/Realm/RLMQueryUtil.hpp
@@ -17,6 +17,7 @@
 ////////////////////////////////////////////////////////////////////////////
 
 #import <Foundation/Foundation.h>
+#import "RLMConstants.h"
 #import <vector>
 
 namespace realm {
@@ -46,8 +47,7 @@ realm::SortOrder RLMSortOrderFromDescriptors(RLMObjectSchema *objectSchema, NSAr
 // This macro validates predicate format with optional arguments
 #define RLM_VARARG(PREDICATE_FORMAT, ARGS) \
 va_start(ARGS, PREDICATE_FORMAT);          \
-va_end(ARGS);                              \
-if (PREDICATE_FORMAT && ![PREDICATE_FORMAT isKindOfClass:[NSString class]]) {         \
-    NSString *reason = @"predicate must be an NSString with optional format va_list"; \
-    [NSException exceptionWithName:RLMExceptionName reason:reason userInfo:nil];       \
-}
+va_end(ARGS);
+
+// return predicate - throw for invalid format
+NSPredicate *RLMValidatedPredicate(id predicateFormat, va_list args);
--- a/Realm/RLMQueryUtil.hpp
+++ b/Realm/RLMQueryUtil.hpp
@@ -17,7 +17,6 @@
 ////////////////////////////////////////////////////////////////////////////
 
 #import <Foundation/Foundation.h>
-#import "RLMConstants.h"
 #import <vector>
 
 namespace realm {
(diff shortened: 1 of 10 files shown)


Rusik:
Hi @jpsim!
Maybe you should move `RLM_VARARG(predicateFormat, args)` with check
```
if (PREDICATE_FORMAT && ![PREDICATE_FORMAT isKindOfClass:[NSString class]]) {
    NSString *reason = @"predicate must be an NSString with optional format va_list";
    [NSException exceptionWithName:RLMExceptionName reason:reason userInfo:nil];
}
```
 from `+ (RLMResults *)objectsWhere:(NSString *)predicateFormat, ...` to `+ (RLMResults *)objectsWhere:(NSString *)predicateFormat args:(va_list)args`?

bdash:
I agree that pushing the checking performed by `RLM_VARARG` into the `va_list` version would be a good idea if these are made public.

jpsim:
@bdash @Rusik I've moved the predicate format validation into an `RLMValidatedPredicate` function so it's now also applied in the `va_list` methods. Anything else?

Realm/RLMQueryUtil.mm
```diff
@@ -71,6 +71,14 @@ static void RLMPrecondition(bool condition, NSString *name, NSString *format, ..
     return prop;
 }
 
+NSPredicate *RLMValidatedPredicate(id predicateFormat, va_list args) {
+    if (predicateFormat && ![predicateFormat isKindOfClass:[NSString class]]) {
```
tgoyne: What's the point of this check? The methods are all declared as taking a NSString and we don't check for violations of the static type anywhere else in the api.

jpsim: That's a *very* good point. Didn't we *used* to have overloads that could be either `NSString *` or `NSPredicate`? I think this is just a leftover from that. I'll revert that back.

Realm/RLMQueryUtil.hpp
```diff
@@ -46,8 +47,7 @@ realm::SortOrder RLMSortOrderFromDescriptors(RLMObjectSchema *objectSchema, NSAr
 // This macro validates predicate format with optional arguments
 #define RLM_VARARG(PREDICATE_FORMAT, ARGS) \
 va_start(ARGS, PREDICATE_FORMAT);          \
-va_end(ARGS);                              \
-if (PREDICATE_FORMAT && ![PREDICATE_FORMAT isKindOfClass:[NSString class]]) {         \
-    NSString *reason = @"predicate must be an NSString with optional format va_list"; \
-    [NSException exceptionWithName:RLMExceptionName reason:reason userInfo:nil];       \
-}
+va_end(ARGS);
```
bdash: It looks like the behavior of this macro is wrong (and was wrong prior to this change). The expected usage of `va_list` to forward variadic arguments is:

```c++
va_list args;
va_start(predicate, args);
target_function(predicate, args);
va_end(args);
```

`va_end` is documented as invalidating the `va_list`, so our practice of `va_start` / `va_end`/ `target_function` is relying on undefined behavior.

Realm/RLMQueryUtil.hpp
```diff
@@ -46,8 +47,7 @@ realm::SortOrder RLMSortOrderFromDescriptors(RLMObjectSchema *objectSchema, NSAr
...
+
+// return predicate - throw for invalid format
+NSPredicate *RLMValidatedPredicate(id predicateFormat, va_list args);
```
bdash: Why is `predicateFormat` declared as `id` rather than `NSString *`?

Realm/RLMQueryUtil.mm
bdash: Why does this intentionally avoid throwing when passed `nil`?

Realm/RLMQueryUtil.mm
```diff
@@ -71,6 +71,14 @@ static void RLMPrecondition(bool condition, NSString *name, NSString *format, ..
...
+        NSString *reason = @"predicate must be an NSString with optional format va_list";
+        @throw [NSException exceptionWithName:RLMExceptionName reason:reason userInfo:nil];
```
bdash: Using `@throw RLMException(…)` would tidy this up a little.

Realm/RLMQueryUtil.mm
jpsim: Again, if memory serves me right I think this was left over from very early code where you could pass in `nil` as a filter, which was the equivalent of `allObjects`. Not useful anymore obviously.

jpsim:
Feedback addressed.

bdash:
:+1:

PR #2944, opened 2015-12-04
Test v0.96.3

tgoyne:


jpsim:
One problem is that this doesn't include #2817 and CI machines no longer have appledoc installed. We can either c-p #2817 or install appledoc on the CI workers.

jpsim:
Any particular reason not to add #2759, #2822, #2811, #2870, #2796 & #2828? Those aren't breaking changes.

tgoyne:
Those all look fine. I didn't include them because they're all missing changelog entries.

jpsim:
0.96.3 is out!

PR #2946, opened 2015-12-04
Fix +[RLMObject createInRealm:withValue:] when given an NSArray

### LINKED ISSUE #2697: Version 0.96 and createInRealm:withValue:
Hi,

Since Realm 0.96, i got a strange bug.

```
[RProgram createInRealm: realm withValue:@[[RUserChannel objectForPrimaryKey:@(currentid)],
                                                             primary,
                                                             @([b[0] integerValue]), // startts
                                                             (...)
```
primary = NSString
startts = NSInteger

This code worked until 0.96, now i got this error:
`Invalid value '1584-1445009400' for property 'startts'`

Look like Realm put the value of `primary` in `startts`... If i uninstall / reinstall application all works fine, but i can't tell to users to uninstall / reinstall application ;)

Thanks

bdash:
It now matches values in the array with properties in the order that the properties are declared. This is the documented behavior.

We had been matching values in the order that the properties were stored in the Realm file. This order often matched the declared order, but isn't guaranteed to.

Fixes #2697.

I need to figure out how to write a test for this.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -41,6 +41,9 @@ x.x.x Release notes (yyyy-MM-dd)
 * Fix a race condition between commits and opening Realm files on new threads
   that could lead to a crash.
 * Fix several crashes when opening Realm files.
+* `-[RLMObject createInRealm:withValue:]`, `-[RLMObject createOrUpdateInRealm:withValue:]`, and
+  their variants for the default Realm now always match the contents of an `NSArray` against properties
+  in the same order as they are defined in the model,.
 
 0.96.2 Release notes (2015-10-26)
 =============================================================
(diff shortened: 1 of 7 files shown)


tgoyne:
`testRearrangeProperties` is the existing test which creates a Realm with a different column order then verifies that the accessor object works correctly.

bdash:
Thanks for the pointer!

bdash:
This is now ready for review.

CHANGELOG.md
```diff
@@ -41,6 +41,9 @@ x.x.x Release notes (yyyy-MM-dd)
 * Fix a race condition between commits and opening Realm files on new threads
   that could lead to a crash.
 * Fix several crashes when opening Realm files.
+* `-[RLMObject createInRealm:withValue:]`, `-[RLMObject createOrUpdateInRealm:withValue:]`, and
+  their variants for the default Realm now always match the contents of an `NSArray` against properties
+  in the same order as they are defined in the model,.
```
tgoyne: ,.

tgoyne:
It'd be good to also test `-[RLMMigration createObject:withValue:]` with prop order != col order, since funky things are done with the schema for migrations and while I suspect it works now, it's something that has a good chance of breaking.

jpsim:
:+1: great fix. Thanks for not ripping apart too many of the guts involved here.

tgoyne:
:+1:

PR #2947, opened 2015-12-04
Include #2817 in 0.96.3 release

jpsim:
Ran `git cherry-pick 23b3717...a823b2e`. /cc @tgoyne


tgoyne:
:+1:

jpsim:
I had to redo it with `git cherry-pick 23b3717^...a823b2e` because otherwise it was missing 23b3717. Should be good now.

PR #2948, opened 2015-12-05
Have github_release.rb fail if any of the carthage commands it runs fail

bdash:
Continuing on after failures can result in an incomplete Carthage.framework.zip being uploaded for the release.

/cc @jpsim @kishikawakatsumi


kishikawakatsumi:
:+1:

jpsim:
:+1:

PR #2954, opened 2015-12-07
Call TableView::clear() with RemoveMode::unordered

tgoyne:
By default it now does an ordered removal, which doesn't actually work with links.

The `TransactLogObserver` changes are to make it always assume unordered removal since it's currently being incorrectly reported as ordered, and we now never use ordered.

It'd be nice to have a test which actually hits the problematic conditions for ordered removal, but that'll probably take a while to write.

@jpsim @bdash

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,7 +18,9 @@ x.x.x Release notes (yyyy-MM-dd)
 
 ### Bugfixes
 
-* None.
+* Fix poor performance when calling `-[RLMRealm deleteObjects:]` on an
+  `RLMResults` which filtered the objects when there are other classes linking
+  to the type of the deleted objects.
 
 0.96.3 Release notes (2015-12-04)
 =============================================================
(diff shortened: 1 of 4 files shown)


tgoyne:
This needs a new core release first as unordered removal is broken without https://github.com/realm/realm-core/pull/1384.

Realm/ObjectStore/impl/transact_log_handler.cpp
```diff
@@ -256,7 +256,7 @@ class TransactLogObserver : public TransactLogValidator {
         return true;
     }
 
-    bool erase_rows(size_t row_ndx, size_t, size_t last_row_ndx, bool unordered)
+    bool erase_rows(size_t row_ndx, size_t, size_t last_row_ndx, bool)
```
jpsim: extra `bool` argument at the end?

bdash:
Given that the new implementation of `erase_rows` is only correct for unordered removals, should it assert or return false, etc., if called with unordered == false?

tgoyne:
I reverted the change to the transaction log handler, as the new core version fixes the bug it was working around.

bdash:
:+1:

PR #2961, opened 2015-12-08
Update doc tool & supported Xcodes in README

jpsim:
/cc @bdash


bdash:
I'd suggest dropping Xcode 6.3 from that list. I can't see any reason to support it at this point.

bdash:
:+1:

PR #2965, opened 2015-12-09
[build.sh] Fix fail to install via CocoaPods on case-sensitive filesystem

kishikawakatsumi:
Create `core/include/Realm` directory after copying ObjectStore headers.
Because cannot find ObjectStore related headers on case-sensitive filesystem.

On `0.96.3` and `master`, fail to install via CocoaPods due to the following error:

```
/Pods/Realm/Realm/ObjectStore/object_store_exceptions.cpp:19:10: 'object_store_exceptions.hpp' file not found
```

@jpsim @bdash


jpsim:
:+1: these issues are annoying

Issue #2976, opened 2015-12-10
Preloaded realm file (Exception error: 'In-memory identifier must not be empty')


rjjohnpatrickcruz:
Hi All,

Any help with my issue, Im using a preloaded realm file but when im trying to read the record inside the file it throws an exception. Please see my screenshots.

Thanks.

ViewController
<img width="520" alt="screen shot 2015-12-10 at 3 44 59 pm" src="https://cloud.githubusercontent.com/assets/5118009/11709780/4db9e87a-9f55-11e5-98d5-ee8068b0fb69.png">

Realm Class
<img width="323" alt="screen shot 2015-12-10 at 3 48 16 pm" src="https://cloud.githubusercontent.com/assets/5118009/11709799/7b53fcb2-9f55-11e5-8837-c03593e4b3ec.png">

Exception Error
<img width="434" alt="screen shot 2015-12-10 at 3 44 37 pm" src="https://cloud.githubusercontent.com/assets/5118009/11709790/5e1a1546-9f55-11e5-989d-65d0d731f5b1.png">

Realm File
[preloaded.realm.zip](https://github.com/realm/realm-cocoa/files/57748/preloaded.realm.zip)

jpsim:
The only possibility for that exception to be thrown based on the code in the screenshots you've shared is if `NSBundle.mainBundle().pathForResource("preloaded", ofType: "realm")` is returning `nil`.

However, since you're not setting a value for `inMemoryIdentifier`, we should instead be throwing a more useful exception along the lines of "A Realm Configuration must specify a path or an in-memory identifier.".

rjjohnpatrickcruz:
Hi @jpsim,

Please any recommendation on how can I debug this issue so that I can have idea on how to fix this.

Thanks.

jpsim:
> Please any recommendation on how can I debug this issue so that I can have idea on how to fix this.

What does `NSBundle.mainBundle().pathForResource("preloaded", ofType: "realm")` return? If it returns `nil`, then your main bundle doesn't include a `preloaded.realm` file. Is this file being copied into the bundle? Is the main bundle where you're calling this code the same bundle containing the `preloaded.realm` file? Are the file name and extension correct? Are you calling this from a command line app where there is no bundle?

rjjohnpatrickcruz:
Yup when I checked the return of the **NSBundle** it returns **nil**, yes I just copied the realm file to main bundle. And I'm calling it in the simulator.

[...]

PR #2977, opened 2015-12-10
improve config error when missing path+identifier

jpsim:
this would previously throw an Objective-C runtime exception: 'In-memory identifier must not be empty' which wasn't always clear, such as #2976. /cc @bdash

### DIFF
--- a/RealmSwift-swift1.2/RealmConfiguration.swift
+++ b/RealmSwift-swift1.2/RealmConfiguration.swift
@@ -147,10 +147,12 @@ extension Realm {
 
         internal var rlmConfiguration: RLMRealmConfiguration {
             let configuration = RLMRealmConfiguration()
-            if self.path != nil {
+            if path != nil {
                 configuration.path = self.path
-            } else {
+            } else if inMemoryIdentifier != nil {
                 configuration.inMemoryIdentifier = self.inMemoryIdentifier
+            } else {
+                fatalError("A Realm Configuration must specify a path or an in-memory identifier.")
             }
             configuration.encryptionKey = self.encryptionKey
             configuration.readOnly = self.readOnly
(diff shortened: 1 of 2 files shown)


RealmSwift-swift1.2/RealmConfiguration.swift
```diff
@@ -147,10 +147,12 @@ extension Realm {
 
         internal var rlmConfiguration: RLMRealmConfiguration {
             let configuration = RLMRealmConfiguration()
-            if self.path != nil {
+            if path != nil {
```
bdash: Why remove the `self.` when most other lines in the function use it, including the next line which assigns to `configuration.path`?

bdash:
While we're improving error messages in this area, would it be worth handling the case where both a path and in-memory identifier are specified?

:+1:

RealmSwift-swift1.2/RealmConfiguration.swift
jpsim: I just copy pasted it from the swift-2.0 case, where `self.` wasn't included in the `path` check. In general, I prefer to omit `self` unless required (some rationale [here](https://github.com/github/swift-style-guide#only-explicitly-refer-to-self-when-required)).

jpsim:
> While we're improving error messages in this area, would it be worth handling the case where both a path and in-memory identifier are specified?

In Objective-C, setting one nils out the other, so I'd rather stay consistent with that behavior.

bdash:
I didn't see that the setters for `path` / `inMemoryIdentifier` nil out the other, rendering my suggestion irrelevant.

RealmSwift-swift1.2/RealmConfiguration.swift
bdash: I don't care what we do, I just wish we would be consistent about it :-)

jpsim:
Not irrelevant, one could easily make the case that both being set is a programmer error, but that's not how it works now and I don't think it's critical (or even necessarily good) to change that now.

jpsim:
CI failure was due to #2978.

RealmSwift-swift1.2/RealmConfiguration.swift
jpsim: For better or worse, that's what SwiftLint is for (even though it doesn't enforce this particular rule yet).

PR #2982, opened 2015-12-11
Switch to using Swift 2.1.1

tgoyne:
Also need to update all the stuff in Jenkins.


jpsim:
There's a lot of duplication of the latest Swift version, we should keep that in a variable to share across the script.

jpsim:
`build.sh` still needs to be updated to work with `ios-swift` 2.1.1, the ios device tests & `ios7-static`.

jpsim:
:+1:

PR #2983, opened 2015-12-11
Don't measure the construction of the NSPredicate in testQueryConstruction

### LINKED ISSUE #2978: SwiftPerformanceTests.testQueryConstruction fails since core 0.95.4 on 32-bit
Usually at around ~80% slower. This only seems to happen on host 1 (32-bit iPhone 5 running iOS 8).

* https://ci.realm.io/job/objc_pr/configuration=Release,swift_version=2.1,target=ios-device-swift/2543/
* https://ci.realm.io/job/objc_pr/configuration=Release,swift_version=2.1,target=ios-device-swift/2538/
* https://ci.realm.io/job/objc_pr/configuration=Release,swift_version=2.1,target=ios-device-swift/2537/
* https://ci.realm.io/job/objc_pr/configuration=Release,swift_version=2.1,target=ios-device-swift/2536/
* https://ci.realm.io/job/objc_pr/configuration=Release,swift_version=2.1,target=ios-device-swift/2534/

See https://ci.realm.io/job/objc_pr/configuration=Release,swift_version=2.1,target=ios-device-swift/2543/testReport/junit/RealmSwift_Tests/SwiftPerformanceTests/testQueryConstruction/history/ for a history of this failing test.

I think this first started happening after we updated core from 0.95.1 to 0.95.4 (#2943). There were a few failures before then but only in the tvOS branch which was using a more recent build of core for a while.

Seems to me like the most likely culprit is https://github.com/realm/realm-core/pull/1363, which had the biggest impact on query code.

Thoughts @bdash?

bdash:
The parsing of the `NSPredicate` was moved into the benchmark loop by 2a2ae542d531a4fe6da132e8dbfdb22710989d7f, but it doesn't belong there.

This should fix #2978.

### DIFF
--- a/RealmSwift-swift2.0/Tests/PerformanceTests.swift
+++ b/RealmSwift-swift2.0/Tests/PerformanceTests.swift
@@ -247,8 +247,8 @@ class SwiftPerformanceTests: TestCase {
 
     func testQueryConstruction() {
         let realm = realmWithTestPath()
-        let predicate = "boolCol = false and (intCol = 5 or floatCol = 1.0) and objectCol = nil and doubleCol != 7.0 " +
-                        " and stringCol IN {'a', 'b', 'c'}"
+        let predicate = NSPredicate(format: "boolCol = false and (intCol = 5 or floatCol = 1.0) and " +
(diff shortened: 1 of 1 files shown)


jpsim:
:+1: wow, that's my bad. Thanks for investigating and fixing.

PR #2989, opened 2015-12-11
Implement KVC collection operators

### LINKED ISSUE #2937: NSPredicate keypath collection query fails on evaluateWithObject:
This was originally reported in an issue with RBQFetchedResultsController: https://github.com/Roobiq/RBQFetchedResultsController/issues/50

Here is the sample project (I made a few updates to the one the user originally provided):
[RBQFetchedResultsControllerExample.zip](https://github.com/realm/realm-cocoa/files/51558/RBQFetchedResultsControllerExample.zip)

If you run the project, you will notice that the predicate (`@"inTable = YES AND objects.@count > 0"`) for the `RBQFetchRequest` works correctly and only objects in the first section appear because those are the only objects with any linked objects in their `RLMArray` objects property.

However, if you click the delete button in the nav bar and then click insert, a new object will be created with a linked object and then it is added to Realm with a notification to trigger the FRC. The FRC will then evaluate the objects that are passed through the notification to verify they match the predicate for the fetch request.

In this case the predicate evaluation fails: `[self.predicate evaluateWithObject:object]`. From what I can tell it seems that internally, this triggers calling `valueForKey:` on the object, with the key being: `@count`.

Is this an incorrect parse of the predicate or do we need to enable support for this sort of `valueForKey:` request?

mrackwitz:
* [x] for `RLMResults`
* [x] for `RLMArray` (fixes #2937)

/c @bdash

### DIFF
--- a/Realm/RLMArray.mm
+++ b/Realm/RLMArray.mm
@@ -300,10 +300,22 @@ static void RLMValidateArrayBounds(__unsafe_unretained RLMArray *const ar,
 }
 
 - (id)valueForKeyPath:(NSString *)keyPath {
-    if (_backingArray) {
-        return [_backingArray valueForKeyPath:keyPath];
+    if (!_backingArray) {
+        return [super valueForKeyPath:keyPath];
     }
-    return [super valueForKeyPath:keyPath];
+    // Although delegating to valueForKeyPath: here would allow to support
+    // nested key paths as well, limiting functionality gives consistency
+    // between standalone and persisted arrays.
+    if ([keyPath characterAtIndex:0] == '@') {
+        NSRange operatorRange = [keyPath rangeOfString:@"." options:NSLiteralSearch];
+        if (operatorRange.location != NSNotFound) {
+            NSString *operatorKeyPath = [keyPath substringFromIndex:operatorRange.location + 1];
+            if ([operatorKeyPath rangeOfString:@"."].location != NSNotFound) {
+                @throw RLMException(@"Nested key paths are not supported yet for KVC collection operators.");
+            }
+        }
+    }
+    return [_backingArray valueForKeyPath:keyPath];
 }
 
 - (id)valueForKey:(NSString *)key {
--- a/Realm/RLMArray.mm
+++ b/Realm/RLMArray.mm
@@ -299,6 +299,13 @@ static void RLMValidateArrayBounds(__unsafe_unretained RLMArray *const ar,
     return [self objectsWithPredicate:[NSPredicate predicateWithFormat:predicateFormat arguments:args]];
 }
 
+- (id)valueForKeyPath:(NSString *)keyPath {
+    if (_backingArray) {
+        return [_backingArray valueForKeyPath:keyPath];
+    }
+    return [super valueForKeyPath:keyPath];
+}
+
 - (id)valueForKey:(NSString *)key {
     if ([key isEqualToString:RLMInvalidatedKey]) {
         return @NO; // Standalone arrays are never invalidated
--- a/Realm/RLMResults.mm
+++ b/Realm/RLMResults.mm
@@ -32,6 +32,7 @@
 #import "results.hpp"
 
 #import <objc/runtime.h>
+#import <objc/message.h>
 #import <realm/table_view.hpp>
 
 using namespace realm;
@@ -147,6 +148,12 @@ static const int RLMEnumerationBufferSize = 16;
 - (instancetype)initPrivate {
     self = [super init];
     return self;
+}
+
+static void assertKeyPathIsNotNested(NSString *keyPath) {
+    if ([keyPath rangeOfString:@"."].location != NSNotFound) {
+        @throw RLMException(@"Nested key paths are not supported yet for KVC collection operators.");
+    }
 }
 
 [[gnu::noinline]]
@@ -293,6 +300,13 @@ static inline void RLMResultsValidateInWriteTransaction(__unsafe_unretained RLMR
                 @throw RLMException(@"Unsupported KVC collection operator found in key path '%@'", keyPath);
             }
         }
+        NSString *operatorName = [keyPath substringWithRange:NSMakeRange(1, operatorRange.location - 1)];
+        NSString *operatorKeyPath = [keyPath substringFromIndex:operatorRange.location + 1];
+        NSAssert(operatorKeyPath.length > 0, @"Missing key path for KVC collection operator %@ in key path '%@'", operatorName, keyPath);
+        SEL opSelector = NSSelectorFromString([NSString stringWithFormat:@"_%@ForKeyPath:", operatorName]);
+        if ([self respondsToSelector:opSelector]) {
+            return ((id(*)(id, SEL, id))objc_msgSend)(self, opSelector, operatorKeyPath);
+        }
     }
     return [super valueForKeyPath:keyPath];
 }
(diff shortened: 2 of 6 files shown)


Realm/RLMResults.mm
```diff
@@ -293,6 +322,57 @@ - (void)setValue:(id)value forKey:(NSString *)key {
...
+}
+
+- (NSArray *)_unionOfObjectsForKeyPath:(NSString *)keyPath {
+    assertKeyPathIsNotNested(keyPath);
+    return translateErrors([&] {
+        return RLMCollectionValueForKey(self, keyPath);
+    });
+}
+
+- (NSArray *)_distinctUnionOfObjectsForKeyPath:(__unused NSString *)keyPath {
```
bdash: `__unused` shouldn't be here.

Realm/RLMResults.mm
```diff
@@ -282,6 +289,28 @@ - (NSUInteger)indexOfObject:(RLMObject *)object {
...
 
+- (id)valueForKeyPath:(NSString *)keyPath {
+    if ([keyPath characterAtIndex:0] == '@') {
+        NSRange operatorRange = [keyPath rangeOfString:@"." options:NSLiteralSearch];
+        if (operatorRange.location == NSNotFound) {
+            if ([keyPath isEqualToString:@"@count"]) {
+                return @(self.count);
+            }
+            else {
+                NSAssert(false, @"Unsupported KVC collection operator found in key path '%@'", keyPath);
```
bdash: `@throw RLMException` to be consistent with `assertKeyPathIsNotNested` above.

Realm/RLMResults.mm
```diff
@@ -282,6 +289,28 @@ - (NSUInteger)indexOfObject:(RLMObject *)object {
...
+            }
+        }
+        NSString *operatorName = [keyPath substringWithRange:NSMakeRange(1, operatorRange.location-1)];
+        NSString *operatorKeyPath = [keyPath substringFromIndex:operatorRange.location+1];
```
bdash: Whitespace around the `+`/`-`.

Realm/RLMResults.mm
mrackwitz: Would you do that as well for the remaining `NSAsserts` with actual conditions? Is there a point in being able to catch those?

bdash: I see only one other use of `NSAssert` in the project.

tgoyne: By default NSAssert is disabled in release builds, so it's not suitable for parameter validation.

mrackwitz: Good point. Replaced them by throwing `RLMExceptions`.

Realm/RLMArray.mm
```diff
@@ -299,6 +299,13 @@ - (RLMResults *)objectsWhere:(NSString *)predicateFormat args:(va_list)args
     return [self objectsWithPredicate:[NSPredicate predicateWithFormat:predicateFormat arguments:args]];
 }
 
+- (id)valueForKeyPath:(NSString *)keyPath {
+    if (_backingArray) {
+        return [_backingArray valueForKeyPath:keyPath];
```
bdash: One problem with delegating to `-[NSArray valueForKeyPath:]` like this is that it will result in standalone arrays supporting more of KVC than persisted arrays (since we reject nested key paths).

jpsim: I generally don't see that as a disadvantage IMHO. Yes, we risk some confusion here, but preventing this would prevent very valid use cases. Not to mention that it'd be a breaking change if users have depended on this before.

bdash: How could users be depending on this? Support for KVC operators is new in this change.

tgoyne: I don't really think it's a good idea, but we could actually delegate to the NSArray implementation of valueForKeyPath: for the things that we don't implement directly for the persisted version.

jpsim: `valueForKeyPath:` isn't overridden on `RLMArray` currently, but it does automatically forward `valueForKey:` to the backing array. Although if KVC operators only trigger `valueForKeyPath:` and not `valueForKey:` then you're right, you can't depend on this today.

mrackwitz: That's true. I don't think that highlighting that in the documentation will be sufficient. That would be easily missed in development.

Even if it seems to me that `RLMCollectionValueForKey` could easily support key paths as well instead just keys, which would cover the array operators. The numeric operators seem on the first look like they would be most efficiently implemented in Core. But KVC also allows nesting those operators (`@sum.employees.@unionOfObjects.pets.age`), which might not be feasible for such a solution, so that we would end-up using accessors for each numeric key path operator with a nested argument and aggregating the retrieved `NSNumber` values instead. This would mean that we would need to highlight performance considerations.

I'd tend to start with not supporting nested key paths at all consistenly, by checking here as well, whether the key path is nested.

mrackwitz: @jpsim: Only `@count` works with `valueForKey:`. So not supporting nested key paths for the sake of consistency for standalone arrays wouldn't be a breaking semantic change.

mrackwitz: I addressed that now by throwing an exception for standalone arrays as well.

Realm/RLMArray.mm
```diff
@@ -299,6 +299,25 @@ - (RLMResults *)objectsWhere:(NSString *)predicateFormat args:(va_list)args
     return [self objectsWithPredicate:[NSPredicate predicateWithFormat:predicateFormat arguments:args]];
 }
 
+- (id)valueForKeyPath:(NSString *)keyPath {
+    if (_backingArray) {
```
bdash: I'd typically write this using an early-return to avoid nesting the entirety of the method, and to make it clear that we never call through to `super` when we have a backing array. I'm not sure if we have a general policy on that though.

Realm/RLMResults.mm
```diff
@@ -282,6 +289,30 @@ - (NSUInteger)indexOfObject:(RLMObject *)object {
...
 
+- (id)valueForKeyPath:(NSString *)keyPath {
+    if ([keyPath characterAtIndex:0] == '@') {
+        NSRange operatorRange = [keyPath rangeOfString:@"." options:NSLiteralSearch];
+        if (operatorRange.location == NSNotFound) {
+            if ([keyPath isEqualToString:@"@count"]) {
+                return @(self.count);
+            }
+            else {
+                @throw RLMException(@"Unsupported KVC collection operator found in key path '%@'", keyPath);
```
bdash: I don't see a test case that covers this error path.

Realm/RLMResults.mm
```diff
@@ -282,6 +289,30 @@ - (NSUInteger)indexOfObject:(RLMObject *)object {
...
+            }
+        }
+        NSString *operatorName = [keyPath substringWithRange:NSMakeRange(1, operatorRange.location - 1)];
+        NSString *operatorKeyPath = [keyPath substringFromIndex:operatorRange.location + 1];
+        if (operatorKeyPath.length == 0) {
+            @throw RLMException(@"Missing key path for KVC collection operator %@ in key path '%@'", operatorName, keyPath);
```
bdash: I don't see a test case that covers this error path.

bdash:
A few of the error cases could do with increased test coverage, but it looks good beyond that. :+1:

Realm/RLMArray.mm
mrackwitz: So you'd early return if there is no backing array by calling through to `super`? That would at least save us one level in nesting and is more similar to the rest of the methods. :+1:

mrackwitz:
Increased test coverage, made an improvement to the error handling in a particular case (see the last commit) and reorganized the tests. Now they are for `RLMResults` all in one dedicated method (`testValueForCollectionOperationKeyPath`), but I omitted to test all the types to aggregate on, that should be well covered by other tests.

jpsim:
This looks good to me! :+1:

tgoyne:
:+1:

PR #2990, opened 2015-12-11
[RLMSchema] Fail instead dead-lock in +sharedSchema

### LINKED ISSUE #2868: Initializing a field with a realm call blocks on thread
Hey Realm people,

I was trying to initialize a field with a realm call, and got a blocked thread.

Here's an example of what I mean:

```swift
//
//  testRealm.swift
//
//  Created by Tom Wilkinson on 11/17/15.
//

import XCTest
import RealmSwift

class TopLevelObject: Object {
    dynamic var nested: NestedObject? = {
        if let object = try? Realm().objectForPrimaryKey(NestedObject.self, key: "default") {
           return object
        }
        return nil
    }()
}

class NestedObject: Object {
    dynamic var key: String = ""
    
        return "key"
    }
}

class testRealm: XCTestCase {

    func testExample() {
        let _ = TopLevelObject() // Blocks forever
    }
    
}
```

So first to explain why I'm trying for this behavior.  It's useful in my case to query another Realm object to find a default object that should be set for another Realm object's fields.  That default object may change dynamically throughout the lifetime of the Realm, so its not the type of thing that I want to hardcode in, and also would ideally be set on init, and wouldn't have to be set by the user every time.  Concretely, imagine you have a `Session` object, and a `Task` object that has a `session` field that you want to default populate with the current active `Session,` where the current active `Session` is marked with a field of its own, and stored in Realm.

And I think what's causing this is mostly the following lines in RLMSchema.mm

```objc
// schema based on runtime objects
+ (instancetype)sharedSchema {
    static std::once_flag onceFlag;
    std::call_once(onceFlag, [&](){
        RLMSchema *schema = [[RLMSchema alloc] init];

        unsigned int numClasses;
        std::unique_ptr<__unsafe_unretained Class[], decltype(&free)> classes(objc_copyClassList(&numClasses), &free);
        [self registerClasses:classes.get() count:numClasses];

        // set class array
        schema.objectSchema = s_partialSharedSchema.objectSchema;

        // set shared schema
        s_sharedSchema = schema;
    });
    return s_sharedSchema;
}
```

The first time we get to that `std::call_once` is from the `Realm().objectForPrimaryKey()` call to setup the Realm from the configuration, but then to get the properties to get the shared schema, there's this unfortunate swift quirk:

```objc
// For Swift classes we need an instance of the object when parsing properties
    id swiftObjectInstance = isSwiftClass ? [[objectClass alloc] init] : nil;
```

Which then goes and tries to call `Realm().objectForPrimaryKey()` again and hits the `std::call_once` wall and deadlocks.

The workaround for me isn't a big deal, I just have to force users to remember to initialize that field themselves, but I thought I would bring it to your attention in case it's a bug you feel like should be worked on or at least mentioned in docs to avoid (or throw an exception?  Hard to figure out at runtime whether a user is going to try and call Realm inside their init).  Or is this something that I'm doing wrong?

mrackwitz:
If a Swift property is initialized to a computed value, which attempts to open a Realm, e.g. to query for a given object, this can led to recursion while initializing the runtime schema, because Swift objects have to be initialized.
Fixes #2868.

Note: Because that affects only the `sharedSchema` and there is no API to reset those, I had to base the test on `RLMMultiProcessTestCase`, which makes this nearly impossible to debug. Because that behavior is only present for Swift object classes, this had to be done in a Swift test, which required to make a test assert to catch and check exceptions available to those as well.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,8 @@ x.x.x Release notes (yyyy-MM-dd)
 * Many forms of queries with key paths on both sides of the comparison operator
   are now supported.
 * Add support for KVC collection operators in `RLMResults` and `RLMArray`.
+* Fail instead of deadlocking in `+[RLMRealm sharedSchema]`, if a Swift property is initialized
+  to a computed value, which attempts to open a Realm on its own.
 
 ### Bugfixes
 
--- a/Realm/RLMSchema.mm
+++ b/Realm/RLMSchema.mm
@@ -180,8 +180,13 @@ static NSMutableDictionary *s_localNameToClass = [[NSMutableDictionary alloc] in
 
 // schema based on runtime objects
 + (instancetype)sharedSchema {
+    static mach_port_t threadID;
+    if (pthread_mach_thread_np(pthread_self()) == threadID) {
+        @throw RLMException(@"Illegal recursive call of +[%@ %@]. Note: Properties of Swift `Object` classes must not be prepopulated with queried results from a Realm.", self, NSStringFromSelector(_cmd));
+    }
     static std::once_flag onceFlag;
     std::call_once(onceFlag, [&](){
+        threadID = pthread_mach_thread_np(pthread_self());
         RLMSchema *schema = [[RLMSchema alloc] init];
 
         unsigned int numClasses;
@@ -193,6 +198,8 @@ static NSMutableDictionary *s_localNameToClass = [[NSMutableDictionary alloc] in
 
         // set shared schema
         s_sharedSchema = schema;
+
+        threadID = 0;
     });
     return s_sharedSchema;
 }
--- a/Realm.xcodeproj/project.pbxproj
+++ b/Realm.xcodeproj/project.pbxproj
@@ -49,6 +49,8 @@
 		02AFB4671A80343600E11938 /* ResultsTests.m in Sources */ = {isa = PBXBuildFile; fileRef = 02AFB4621A80343600E11938 /* ResultsTests.m */; };
 		02AFB4681A80343600E11938 /* ResultsTests.m in Sources */ = {isa = PBXBuildFile; fileRef = 02AFB4621A80343600E11938 /* ResultsTests.m */; };
 		02E334C31A5F41C7009F8810 /* DynamicTests.m in Sources */ = {isa = PBXBuildFile; fileRef = E81A1FBA1955FE0100FDED82 /* DynamicTests.m */; };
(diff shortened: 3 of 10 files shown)


Realm/RLMSchema.mm
```diff
@@ -180,8 +180,13 @@ + (void)registerClasses:(Class *)classes count:(NSUInteger)count {
 
 // schema based on runtime objects
 + (instancetype)sharedSchema {
+    static mach_port_t threadID;
+    if (pthread_mach_thread_np(pthread_self()) == threadID) {
+        @throw RLMException(@"Illegal recursive call of +[%@ %@]. Note: Properties of Swift `Object` classes must not be prepopulated with queried results from a Realm.", self, NSStringFromSelector(_cmd));
```
jpsim: Why evaluate `self` and `NSStringFromSelector(_cmd)` here when this isn't designed to be subclassed & overridden? Just hardcode `+[RLMSchema sharedSchema]`.

jpsim:
The changes made to Swift 2.0 should be mirrored for Swift 1.2. And I'm not convinced that checking the thread ID here is correct. We should also prevent concurrent access to a partially initialized schema, so we should replace `threadID` with a `schemaIsInitializing` boolean that is set during the initialization period, and replace `if (pthread_mach_thread_np(pthread_self()) == threadID)` with `if (schemaIsInitializing)`.

tgoyne:
> replace if (pthread_mach_thread_np(pthread_self()) == threadID) with if (schemaIsInitializing).

That would make it throw if `[RLMSchema sharedSchema]` is called simultaneously on two different threads, which currently works correctly.

jpsim:
> That would make it throw if [RLMSchema sharedSchema] is called simultaneously on two different threads, which currently works correctly.

ah, nevermind then :+1:.

Realm/RLMSchema.mm
mrackwitz: Names can changes. Stringly-encoded type information tends to be not updated accordingly.

CHANGELOG.md
```diff
@@ -16,6 +16,8 @@ x.x.x Release notes (yyyy-MM-dd)
 * Many forms of queries with key paths on both sides of the comparison operator
   are now supported.
 * Add support for KVC collection operators in `RLMResults` and `RLMArray`.
+* Fail instead dead-lock in `+[RLMRealm sharedSchema]`, if a Swift property is initialized
```
tgoyne: instead of deadlocking

tgoyne:
:+1:

PR #2998, opened 2015-12-15
Add support for subqueries

bdash:
Expressions that compare the count of a subquery against a constant value are supported. Comparing against non-constant values is not yet supported, nor is comparing anything but the count of the subquery.

This depends on realm/realm-core#1405 and cannot be merged until we're using a core release that contains that change.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -36,6 +36,10 @@ x.x.x Release notes (yyyy-MM-dd)
   properties.
 * Indexing `BOOL`/`Bool` and `NSDate` properties are now supported.
 * Swift: Add support for indexing optional properties.
+* `NSPredicate`'s `SUBQUERY` operator is now supported. It has the following limitations:
+  * `@count` is the only operator that may be applied to the `SUBQUERY` expression.
+  * The `SUBQUERY(…).@count` expression must be compared with a constant.
+  * Correlated subqueries are not yet supported.
 
 ### Bugfixes
 
--- a/Realm/RLMPredicateUtil.hpp
+++ b/Realm/RLMPredicateUtil.hpp
@@ -1,0 +1,36 @@
+////////////////////////////////////////////////////////////////////////////
+//
+// Copyright 2015 Realm Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+#import <Foundation/Foundation.h>
+
+class RLMPredicateExpressionTransformer {
+    using ExpressionVisitor = NSExpression *(*)(NSExpression *);
+
+public:
+    static NSPredicate *transform(NSPredicate *predicate, ExpressionVisitor visitor) {
+        RLMPredicateExpressionTransformer self(visitor);
+        return self.visit(predicate);
+    }
+
+private:
+    RLMPredicateExpressionTransformer(ExpressionVisitor visitor) : m_visitor(visitor) { }
+
+    NSExpression *visit(NSExpression *expression) const;
+    NSPredicate *visit(NSPredicate *predicate) const;
+
+    ExpressionVisitor m_visitor;
+};
--- a/Realm/RLMPredicateUtil.hpp
+++ b/Realm/RLMPredicateUtil.hpp
@@ -17,20 +17,5 @@
 
 #import <Foundation/Foundation.h>
 
-class RLMPredicateExpressionTransformer {
-    using ExpressionVisitor = NSExpression *(*)(NSExpression *);
-
-public:
-    static NSPredicate *transform(NSPredicate *predicate, ExpressionVisitor visitor) {
-        RLMPredicateExpressionTransformer self(visitor);
-        return self.visit(predicate);
-    }
-
-private:
-    RLMPredicateExpressionTransformer(ExpressionVisitor visitor) : m_visitor(visitor) { }
-
-    NSExpression *visit(NSExpression *expression) const;
-    NSPredicate *visit(NSPredicate *predicate) const;
-
-    ExpressionVisitor m_visitor;
-};
+using ExpressionVisitor = NSExpression *(*)(NSExpression *);
+NSPredicate *transformPredicate(NSPredicate *, ExpressionVisitor);
--- a/Realm/RLMQueryUtil.mm
+++ b/Realm/RLMQueryUtil.mm
@@ -1015,7 +1015,7 @@ void update_query_with_subquery_expression(RLMSchema *schema, RLMObjectSchema *o
 
     // Eliminate references to the iteration variable in the subquery.
     NSPredicate *subqueryPredicate = [subqueryExpression.predicate predicateWithSubstitutionVariables:@{ subqueryExpression.variable : [NSExpression expressionForEvaluatedObject] }];
-    subqueryPredicate = RLMPredicateExpressionTransformer::transform(subqueryPredicate, simplify_self_value_for_key_path_function_expression);
+    subqueryPredicate = transformPredicate(subqueryPredicate, simplify_self_value_for_key_path_function_expression);
 
     Query subquery = collectionMemberObjectSchema.table->where();
     RLMUpdateQueryWithPredicate(&subquery, subqueryPredicate, schema, collectionMemberObjectSchema);
--- a/Realm/RLMQueryUtil.mm
+++ b/Realm/RLMQueryUtil.mm
@@ -21,6 +21,7 @@
 #import "RLMArray_Private.hpp"
 #import "RLMObject_Private.hpp"
 #import "RLMObjectSchema_Private.hpp"
+#import "RLMPredicateUtil.hpp"
 #import "RLMProperty_Private.h"
 #import "RLMSchema_Private.h"
 #import "RLMUtil.hpp"
@@ -142,6 +143,12 @@ public:
     auto resolve(Query& query) const
     {
         return table_for_query(query)->template column<T>(index());
+    }
+
+    template <typename T>
+    auto resolveWithSubquery(Query& query, Query subquery) const
+    {
+        return table_for_query(query)->template column<T>(index(), std::move(subquery));
     }
 
     RLMProperty *property() const { return m_property; }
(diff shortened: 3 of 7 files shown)


bigfish24:
@bdash I am happy to test this a bit since I have a few projects that will benefit. What are the unsupported advanced use cases that are not supported with subqueries?

bdash:
1. Correlated subqueries (e.g., a subquery that contains a reference a property of the parent table) are not supported.
2. `SUBQUERY(…).@count` is the only form that's currently supported. I'm not sure if there's any use case for operations other than `@count` on the subquery.
3. The subquery expression must be compared with a constant. Supporting comparisons against more general expressions is clearly desirable for the future.

astigsen:
>  I'm not sure if there's any use case for operations other than `@count` on the subquery.

I could imagine use cases for the rest of the aggregate operators (`sum`, `avg`, `min`, `max`, `first`, `last`) and possibly future operators like `distinct` and `group_by`. But I don't think there is any pressing need for any of them right now.

bigfish24:
Rebased this off of [0.97.0](https://github.com/realm/realm-cocoa/commit/280b2b42af861c18a956ff98cb9e4675279ddf45) and updated to work with Core v0.95.7.

bigfish24:
@bdash nice work! I was able to build the framework and have done a few limited tests, but seems to be working. Will continue to do more thorough testing against a large Salesforce dataset.

maremmle:
Hi, I am super confused about subqueries. 
As TiM commented on [Stackoverflow](http://stackoverflow.com/a/32776267/1879171) and from this PR it is apparently not yet implemented nor released. But at the same time it already in the docs at the [NSPredicate Cheatsheet](https://realm.io/news/nspredicate-cheatsheet/). 

So …
  1. I am right that it's not ready yet? 
  2. Why is there something in the docs, that isn't implemented / available yet? 

thx. Martin

bdash:
The [`NSPredicate` cheatsheet](https://realm.io/news/nspredicate-cheatsheet/) covers all features of `NSPredicate`, including those which Realm does not yet support. The features which Realm supports are called out via pink dots.

maremmle:
@bdash thanks

bigfish24:
@bdash this seems to fail when the subquery is evaluating a keypath versus a property, i.e.:

``` objc
// Simplified description of the object model
// Object : RLMObject
// @property RLMArray<Field *><Field> *fields;

// Field: RLMObject
// @property PicklistEntry *picklistValue;

// PicklistEntry: RLMObject
// @property NSString *value

[Object objectsWhere:@"SUBQUERY(fields, $field, $field.picklistValue.value == 'Dormant' AND ...).@count > 0"];
```

>*** Terminating app due to uncaught exception 'Invalid property name', reason: 'Property 'Dormant' not found in object of type 'Field''

bigfish24:
@bdash false alarm, the key path subquery evaluation is working. My issue was caused by how `NSPredicate` parses `SUBQUERY` expressions. I was building the predicate string incorrectly causing the predicate to be incorrectly parsed (but which wasn't immediately apparent in the debugger).

Here's a quick summary of the quirks around predicate parsing with `SUBQUERY`:

``` objc
// Correct predicate parsing with string without substitutions on subquery keypath
NSPredicate *passPredicate = [NSPredicate predicateWithFormat:@"SUBQUERY(array, $object, $object.boolCol = YES AND $object.objectCol.stringCol = 'string0').@count > 0"];

// Correct predicate parsing with substitution through NSString
NSString *predicateString = [NSString stringWithFormat:@"SUBQUERY(array, $object, $object.boolCol = YES AND %@.%@ = 'string0').@count > 0",@"$object",@"objectCol.stringCol"];
NSPredicate *pass1Predicate = [NSPredicate predicateWithFormat:predicateString];

// Correct predicate parsing with substitution through NSPredicate, but not substituting in variable
// NOTE: '%K' is required to prevent quotations to surround substitution
NSPredicate *pass2Predicate = [NSPredicate predicateWithFormat:@"SUBQUERY(array, $object, $object.boolCol = YES AND $object.%K = 'string0').@count > 0",@"objectCol.stringCol"];

// Failed predicate parsing with substitution through NSPredicate including variable via %K
// NOTE: the variable '$object' is not correctly parsed due to the substitution
NSPredicate *failPredicate = [NSPredicate predicateWithFormat:@"SUBQUERY(array, $object, $object.boolCol = YES AND %K.%K = 'string0').@count > 0",@"$object",@"objectCol.stringCol"];

// Failed predicate parsing with substitution through NSPredicate including variable via %@
// NOTE: the substitution results in the key path parsed as a constant due to quotations
NSPredicate *fail2Predicate = [NSPredicate predicateWithFormat:@"SUBQUERY(array, $object, $object.boolCol = YES AND %@.%@ = 'string0').@count > 0",@"$object",@"objectCol.stringCol"];
```

bdash:
This is no longer blocked as we now have a core release with support for subqueries.

Realm/RLMQueryUtil.mm
```diff
@@ -85,12 +86,14 @@ size_t find_first(size_t start, size_t end) const override
     }
     void set_table(const Table*) override {}
     const Table* get_table() const override { return nullptr; }
+    std::unique_ptr<Expression> clone(QueryNodeHandoverPatches*) const override { return nullptr; }
```
tgoyne: AFAICT from looking at the code that calls this, it does need to return a new object or OrNote::find_first_local() will deref a null pointer.

Realm/RLMPredicateUtil.hpp
```diff
@@ -0,0 +1,36 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+#import <Foundation/Foundation.h>
+
+class RLMPredicateExpressionTransformer {
```
tgoyne: There doesn't appear to be any reason for this to be a class rather than just a function, since the public interface is just a function, and the use of the class in the implementation of that function just rearranges the order of the arguments to the visit functions.

Realm/RLMQueryUtil.mm
```diff
@@ -980,6 +989,34 @@ void update_query_with_column_expression(RLMSchema *schema, RLMObjectSchema *des
                             std::move(left), std::move(right));
 }
 
+NSExpression *simplify_self_value_for_key_path_function_expression(NSExpression *expression) {
+    // -[NSPredicate predicateWithSubtitutionVariables:] results in function expressions of the form [SELF valueForKeyPath:]
+    // that update_query_with_predicate cannot handle. Replace such expressions with equivalent NSKeyPathExpressionType expressions.
+    if (expression.expressionType == NSFunctionExpressionType
+        && expression.operand.expressionType == NSEvaluatedObjectExpressionType
+        && [expression.function isEqualToString:@"valueForKeyPath:"]) {
+        if (NSString *keyPath = expression.arguments.firstObject.keyPath) {
```
tgoyne: I find this indentation hard to read, as it's hard to scan where the conditional ends and the body of the block begins. I'd find it easier to read if the continuation of the condition was indented further, or perhaps even extracted to a separate function entirely (although in this case that'd render this function a bit pointless).

Realm/RLMQueryUtil.mm
```diff
@@ -1064,15 +1101,34 @@ void update_query_with_predicate(NSPredicate *predicate, RLMSchema *schema,
             update_query_with_value_expression(schema, objectSchema, query, compp.rightExpression.keyPath,
                                                compp.leftExpression.constantValue, compp);
         }
+        else if (exp1Type == NSFunctionExpressionType) {
+            NSExpression *functionExpression = compp.leftExpression;
+            if (functionExpression.operand.expressionType == NSSubqueryExpressionType && [functionExpression.function isEqualToString:@"valueForKeyPath:"] && functionExpression.arguments.count == 1) {
+                NSExpression *keyPathExpression = functionExpression.arguments.firstObject;
+                if ([keyPathExpression.keyPath isEqualToString:@"@count"]
+                    && compp.rightExpression.expressionType == NSConstantValueExpressionType
+                    && [compp.rightExpression.constantValue isKindOfClass:[NSNumber class]]) {
```
tgoyne: See note above about indentation.

Realm/RLMQueryUtil.mm
```diff
@@ -980,6 +989,34 @@ void update_query_with_column_expression(RLMSchema *schema, RLMObjectSchema *des
...
+        && [expression.function isEqualToString:@"valueForKeyPath:"]) {
+        if (NSString *keyPath = expression.arguments.firstObject.keyPath) {
+            return [NSExpression expressionForKeyPath:keyPath];
+        }
+    }
+    return expression;
+}
+
+void update_query_with_subquery_expression(RLMSchema *schema, RLMObjectSchema *objectSchema, realm::Query& query,
+                                           NSExpression *subqueryExpression, NSPredicateOperatorType operatorType, NSUInteger value) {
```
tgoyne: Why is value NSUInteger? The value passed to it is signed and it's cast back to int64_t when it's used.

Realm/RLMQueryUtil.mm
bdash: Yeah, you're right. Brainfart on my part.

Realm/RLMPredicateUtil.hpp
bdash: The class holds a reference to the visitor function to avoid having to pass it to every call to `visit` within the visiting code. The class can be moved to the .mm file though.

Realm/RLMQueryUtil.mm
bdash: I addressed both this and the other gnarly if statement in b227714 by introducing helper functions. It's more code, but a lot easier to follow.

CHANGELOG.md
```diff
@@ -36,6 +36,10 @@ x.x.x Release notes (yyyy-MM-dd)
   properties.
 * Indexing `BOOL`/`Bool` and `NSDate` properties are now supported.
 * Swift: Add support for indexing optional properties.
+* `NSPredicate`'s `SUBQUERY` operator is now supported. It has the following limitations:
+  * `@count` is the only operator that may be applied to the `SUBQUERY` expression.
```
jpsim: we should file GitHub issues tracking these limitations so we remember to address them eventually.

Realm/Tests/QueryTests.m
```diff
@@ -180,8 +180,9 @@ - (void)testPredicateNotSupported
     // LinkList equality is unsupport since the semantics are unclear
     XCTAssertThrows(([ArrayOfAllTypesObject objectsWhere:@"ANY array = array"]));
 
-    // subquery
-    XCTAssertThrows(([ArrayOfAllTypesObject objectsWhere:@"SUBQUERY(array, $obj, $obj.intCol = 5).@count > 1"]));
+    // Unsupported variants of subqueries.
+    RLMAssertThrowsWithReasonMatching(([ArrayOfAllTypesObject objectsWhere:@"SUBQUERY(array, $obj, $obj.intCol = 5).@count == array.@count"]), @"SUBQUERY.*compared with a constant number");
+    RLMAssertThrowsWithReasonMatching(([ArrayOfAllTypesObject objectsWhere:@"SUBQUERY(array, $obj, $obj.intCol = 5) == 0"]), @"SUBQUERY.*immediately followed by .@count");
```
jpsim: AFAICT a common subquery pattern would be `prop IN SUBQUERY(...)` so it'd be nice to have a test to confirm we're throwing an exception with a helpful message in that case.

CHANGELOG.md
bdash: Filed #3169.

Realm/Tests/QueryTests.m
```diff
@@ -1748,6 +1749,25 @@ - (void)testLinksToDeletedOrMovedObject
...
+                                                       @{@"name": @"Jill",  @"age": @40, @"hired": @YES},
+                                                       @{@"name": @"Joe",  @"age": @40, @"hired": @YES}]]];
+    [CompanyObject createInRealm:realm
+                       withValue:@[@"second company", @[@{@"name": @"Bill", @"age": @35, @"hired": @YES},
+                                                        @{@"name": @"Don",  @"age": @45, @"hired": @NO},
+                                                        @{@"name": @"Tim",  @"age": @60, @"hired": @NO}]]];
+    [realm commitWriteTransaction];
+
+    XCTAssertEqual(1U, [CompanyObject objectsWhere:@"SUBQUERY(employees, $employee, $employee.age > 30 AND $employee.hired = FALSE).@count > 0"].count);
+    XCTAssertEqual(2U, [CompanyObject objectsWhere:@"SUBQUERY(employees, $employee, $employee.age < 30 AND $employee.hired = TRUE).@count == 0"].count);
```
jpsim: Would it be valuable to test chaining this query with another and confirming that the final query is properly auto-updated?

bdash: I'm not sure. Auto-updating isn't specific to the type of query being performed, so why would it be useful to test explicitly in this case?

tgoyne: Would be good to use RLMAssertCount here to at least minimally test handover of subqueries.

tgoyne:
:+1:

jpsim:
:+1:

PR #3005, opened 2015-12-15
Handle queries on objects that get deleted or moved

vigonius:
cc @jpsim @bdash @tgoyne 

Depends on https://github.com/realm/realm-core/pull/1397

### DIFF
--- a/Realm/RLMQueryUtil.mm
+++ b/Realm/RLMQueryUtil.mm
@@ -503,7 +503,7 @@ void add_link_constraint_to_query(realm::Query & query,
     RLMPrecondition(query.get_table()->get_link_target(column.index()).get() == obj->_row.get_table(),
                     @"Invalid value origin", @"Object must be from the Realm being queried");
 
-    query.links_to(column.index(), obj->_row.get_index());
+    query.links_to(column, obj->_row);
 }
 
 void add_link_constraint_to_query(realm::Query & query,
--- a/Realm/Tests/TransactionTests.m
+++ b/Realm/Tests/TransactionTests.m
@@ -17,9 +17,40 @@
 ////////////////////////////////////////////////////////////////////////////
 
 #import "RLMTestCase.h"
+#import <XCTest/XCTest.h>
 
 @interface TransactionTests : RLMTestCase
 @end
+
+
+
+RLM_ARRAY_TYPE(Breeder)
+
+@interface Breeder : RLMObject
+@end
+
+@implementation Breeder
+@end
+
+@interface Dog : RLMObject
+
+@property RLMArray<Breeder> *breeders;
+
+@end
+
+@implementation Dog
+@end
+
+@interface Person : RLMObject
+
+@property Dog *dog;
+
+@end
+
+@implementation Person
+@end
+
+
 
 @implementation TransactionTests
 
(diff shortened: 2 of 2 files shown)


Realm/Tests/TransactionTests.m
```diff
@@ -52,4 +83,102 @@ - (void)testTransactionMisuse
...
+        
+        [realm deleteObject:breeder];
+        
+    }];
+    
+    XCTAssertNoThrow(results.count, @"Moved Items test Failed.");
+    XCTAssertEqual(results.count,(unsigned long)0);
+}
+
+- (void)testLinkQueryWhenLinkIsDeleted
```
bdash: My reading is that this single test exercises the same code paths as the two preceding tests. Do the other two tests exercise any behavior that this one doesn't? If not, it seems like we'd only need the one test case, and can avoid adding the new model types.

bdash: One thing that these tests don't cover is whether an object can still be found when removal of other rows in its table causes its row index to change. That seems worth having coverage of.

Realm/RLMQueryUtil.mm
```diff
@@ -503,7 +503,7 @@ void add_link_constraint_to_query(realm::Query & query,
     RLMPrecondition(query.get_table()->get_link_target(column.index()).get() == obj->_row.get_table(),
                     @"Invalid value origin", @"Object must be from the Realm being queried");
 
-    query.links_to(column.index(), obj->_row.get_index());
+    query.links_to(column, obj->_row);
```
bdash: Removing the call to `.index()` will prevent this from compiling.

Realm/Tests/TransactionTests.m
jpsim: @vigonius says that's good idea and will do that tonight.

Realm/Tests/TransactionTests.m
```diff
@@ -52,4 +83,102 @@ - (void)testTransactionMisuse
     XCTAssertThrows([realm deleteObject:obj], @"Outside writetransaction");
 }
 
+- (void)testDeleteBug
+{
+    RLMRealm *realm = [RLMRealm defaultRealm];
+    [realm transactionWithBlock:^{
```
jpsim: Realms are reset between all tests, so it shouldn't be necessary to delete all objects at the beginning of tests.

Realm/Tests/TransactionTests.m
```diff
@@ -17,10 +17,41 @@
...
+#import <XCTest/XCTest.h>
 
 @interface TransactionTests : RLMTestCase
 @end
 
+
+
+RLM_ARRAY_TYPE(Breeder)
+
+@interface Breeder : RLMObject
```
jpsim: instead of adding new types `Breeder` and `Dog`, I'd encourage you to reuse an existing model defined in `RLMTestObjects`.

Realm/Tests/TransactionTests.m
```diff
@@ -52,4 +83,102 @@ - (void)testTransactionMisuse
...
+        
+        Dog *second_dog = [[Dog alloc] init];
+        [second_dog.breeders addObject:second_breeder];
+        [realm addObject:second_dog];
+        
+        [realm deleteObject:breeder];
+        
+    }];
+    
+    XCTAssertNoThrow(results.count, @"Moved Items test Failed.");
```
jpsim: not necessary since we're accessing `results.count` on the next line.

bdash:
I'm going to redo this.

PR #3007, opened 2015-12-16
[RLMObjectBase] fix analyzer warning

### LINKED ISSUE #3006: Analysis issue in 0.96.3
Assertion: Access to instance variable '_objectSchema' results in a dereference of a null pointer (loaded from variable 'self')
Realm/Realm/RLMObjectBase.mm:102

jpsim:
Fixes #3006 /cc @bdash @tgoyne 

Should we be failing more loudly here if your model objects can't be initialized?


tgoyne:
:+1: 

`[super init]` can only return nil if someone has monkey-patched NSObject (or changed our superclass...), so no point in trying to do anything fancy.

jpsim:
CI failure unrelated (perf test deviation).

PR #3009, opened 2015-12-16
[CONTRIBUTING] Add guidelines about the format of commit messages

mrackwitz:
Based on https://swift.org/contributing/#commit-messages with minor adaptions.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -43,6 +43,21 @@ $(which git && git --version)
 
 We love contributions to Realm! If you'd like to contribute code, documentation, or any other improvements, please [file a Pull Request](https://github.com/realm/realm-cocoa/pulls) on our GitHub repository. Make sure to accept our [CLA](#CLA) and to follow our [style guide](https://github.com/realm/realm-cocoa/wiki/Objective-C-Style-Guide).
 
+### Commit Messages
+
+Although we don’t enforce a strict format for commit messages, we prefer that you follow the guidelines below, which are common among open source projects. Following these guidelines helps with the review process, searching commit logs and documentation of implementation details. At a high level, the contents of the commit message should convey the rationale of the change, without delving into much detail. For example, `setter names were not set right` leaves the reviewer wondering about which bits and why they weren’t “right”. In contrast, `[RLMProperty] Correctly capitalize setterName` conveys almost all there is to the change.
+
+Below are some guidelines about the format of the commit message itself:
+
+* Separate the commit message into a single-line title and a separate body that describes the change.
+* Make the title concise to be easily read within a commit log.
+* In changes that are restricted to a specific part of the code, include a [tag] at the start of the line in square brackets — for example, `[RLMArray] …`, `[scripts] …` or  `[Podspec] …`. This tag helps email filters and searches for post-commit reviews.
+* Make the body concise, while including the complete reasoning. Unless required to understand the change, additional code examples or other details should be left to the pull request.
+* If the commit fixes a bug, include the number of the issue in the message.
+* Use the first person present tense - for example "Fix …" instead of "Fixes …" or "Fixed …".
+* For text formatting and spelling, follow the same rules as documentation and in-code comments — for example, the use of capitalization and periods.
+* If the commit is a bug fix on top of another recently committed change, or a revert or reapply of a patch, include the Git revision number of the prior related commit, e.g. `Revert abcd3fg because it caused #1234`.
+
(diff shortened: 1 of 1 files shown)


jpsim:
:+1:

CONTRIBUTING.md
```diff
@@ -43,6 +43,21 @@ $(which git && git --version)
...
+
+Below are some guidelines about the format of the commit message itself:
+
+* Separate the commit message into a single-line title and a separate body that describes the change.
+* Make the title concise to be easily read within a commit log.
+* In changes that are restricted to a specific part of the code, include a [tag] at the start of the line in square brackets — for example, `[RLMArray] …`, `[scripts] …` or  `[Podspec] …`. This tag helps email filters and searches for post-commit reviews.
+* Make the body concise, while including the complete reasoning. Unless required to understand the change, additional code examples or other details should be left to the pull request.
+* If the commit fixes a bug include the number of the issue in the message.
```
bdash: s/include the/include, the/

I think this point should be more specific about using the phrasing that GitHub will use to flag the issue to be automatically closed when the commit is merged (e.g., Fixes #XYZ).

CONTRIBUTING.md
```diff
@@ -43,6 +43,21 @@ $(which git && git --version)
...
 
+### Commit Messages
+
+Although we don’t enforce a strict format for commit messages, we prefer that you follow the guidelines below, which are common among open source projects. Following these guidelines helps with the review process, searching commit logs and documentation of implementation details. At a high level, the contents of the commit message should convey the rationale of the change, without delving into much detail. For example, `setter names were not set right` leaves the reviewer wondering about which bits and why they weren’t “right”. In contrast, `[RLMProperty] Correctly capitalize setterName` conveys almost all there is to the change.
+
+Below are some guidelines about the format of the commit message itself:
+
+* Separate the commit message into a single-line title and a separate body that describes the change.
+* Make the title concise to be easily read within a commit log.
+* In changes that are restricted to a specific part of the code, include a [tag] at the start of the line in square brackets — for example, `[RLMArray] …`, `[scripts] …` or  `[Podspec] …`. This tag helps email filters and searches for post-commit reviews.
```
bdash: This is the only part that I disagree with. These tags clutter up the titles and don't provide much value. It's not like we have tens of pull requests a day that we have to filter through.

tgoyne: I also find them noisy and of little value.

CONTRIBUTING.md
```diff
@@ -43,6 +43,21 @@ $(which git && git --version)
...
+* Make the body concise, while including the complete reasoning. Unless required to understand the change, additional code examples or other details should be left to the pull request.
+* If the commit fixes a bug include the number of the issue in the message.
+* Use the first person present tense - for example "Fix …" instead of "Fixes …" or "Fixed …".
```
bmunkholm: is this important? I would strive for only really valuable rules that has a big value.

CONTRIBUTING.md
```diff
@@ -43,6 +43,21 @@ $(which git && git --version)
...
+* For text formatting and spelling, follow the same rules as documentation and in-code comments — for example, the use of capitalization and periods.
```
bmunkholm: Is this important?

bmunkholm:
@mrackwitz Once the review here is finished, I would suggest you to promote this in #Engineering channel to other teams.

CONTRIBUTING.md
mrackwitz: I think, we should strive for proper and clear use of language, which e.g. includes for me to make use of capitalization instead of writing all lowercase or UPPERCASE.

CONTRIBUTING.md
mrackwitz: I like them when going through the history on a branch to understand in retrospective what was going on as it makes it easy to skim through the commits and filter out only those which were really relevant. The examples given here were all for components, where you could argue that would be also easy to figure out just from the list of files which were modified. But beside that I've also seen and used some in addition, which describe more the type of change which was applied e.g. `[Style]`, `[Doc]`, `[Refactor]`, `[Test]`, if and only if a change is entirely dedicated to that.

jpsim: @mrackwitz can you please just remove this point from the document since it's the only part that's been disputed? That way we can get the rest of the document in, and take the debate about tags to Slack and move on.

mrackwitz:
Removed the debatable point about tags in commit messages as suggested by @jpsim.

PR #3011, opened 2015-12-17
Always allow adding notification blocks on the main thread

### LINKED ISSUE #2939: Exception adding a notification block on OS X in applicationDidFinishLaunching:
On OS X, the call `CFRunLoopCopyCurrentMode(CFRunLoopGetCurrent())` will return NULL if called on a thread where the runloop is stopped. This will happen on the main thread in an OS X Cocoa application during `applicationDidFinishLaunching:`.

The `-addNotificationBlock:` call checks for an active run loop (in `RLMIsInRunLoop`) using this code and throws an exception otherwise. Unfortunately that means that notifications can’t be added in `applicationDidFinishLaunching:` unless queued for later dispatch using `dispatch_after`.

tgoyne:
Closes #2939.

@jpsim Should get this in for 0.97.

### DIFF
--- a/Realm/RLMUtil.mm
+++ b/Realm/RLMUtil.mm
@@ -333,6 +333,14 @@ BOOL RLMIsDebuggerAttached()
 }
 
 BOOL RLMIsInRunLoop() {
+    // The main thread may not be in a run loop yet if we're called from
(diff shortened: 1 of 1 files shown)


jpsim:
:+1: we almost forgot this!

Issue #2968, opened 2015-12-09
A signed resource has been added, modified, or deleted"


dannycohn:

bdash:
@dannycohn, can you please provide any information at all about what problem you're seeing, what you were trying to do when you saw it, and with which version of Realm you saw it?

dannycohn:
Sorry, I thought I stopped and closed the tab, didn't mean to submit the issue. I figured it out.

dannycohn:
Dec  9 08:19:48 iPhone itunesstored[86] <Error>: 0x218b000 __MobileInstallationInstallForLaunchServices_block_invoke240: Returned error Error Domain=MIInstallerErrorDomain Code=13 "Failed to verify code signature of <MIExecutableBundle : path = /private/var/mobile/Library/Caches/com.apple.mobile.installd.staging/temp.s8jTQN/extracted/Payload/Connect-Order-652.app/Frameworks/Realm.framework identifier = io.Realm.Realm type = Framework> : 0xe8008017 (Signed resources have been added, removed, or modified)" UserInfo=0x16627af0 {LibMISErrorNumber=-402620393, LegacyErrorString=ApplicationVerificationFailed, FunctionName=-[MICodeSigningVerifier performValidationWithError:], NSLocalizedDescription=Failed to verify code signature of <MIExecutableBundle : path = /private/var/mobile/Library/Caches/com.apple.mobile.installd.staging/temp.s8jTQN/extracted/Payload/Connect-Order-652.app/Frameworks/Realm.framework identifier = io.Realm.Realm type = Framework> : 0xe8008017 (Signed resources have been added, removed, or modified), SourceFileLine=192}

jpsim:
An explanation of what's happening here would go a long way in helping us understand what's happening, @dannycohn.

For example, what version of Realm/iOS/Xcode is this error happening on?

The log you've shared seems to be from an iOS device, how did you install your app on that device (TestFlight, HockeyApp, Xcode, App Store, enterprise certificate, etc)?

How did you install Realm in your app (CocoaPods, Carthage, prebuilt binaries, from source, with/without bitcode)?

dannycohn:
Ya sorry, I meant to come back and put in more info.

[...]

PR #3012, opened 2015-12-17
[Script] Avoid code signing bcsymbolmap files

kishikawakatsumi:
Move copying bcsymbolmap files to before the code signing.
Because these files to be erased or being moved, problems are occurred if they are signed.

Fixes https://github.com/realm/realm-cocoa/issues/2968

cc @bdash @jpsim


jpsim:
:+1:

Issue #2984, opened 2015-12-11
sumOfProperty wrong


shaozi:
Hi, Please see the question on SO: http://stackoverflow.com/q/34216584/528629
I did a deleteAllObjects and then add them back quickly. But the sumOfProperty gives me 0 instead of the correct sum. Meanwhile, a for loop that add each works.

mrackwitz:
Hey, I saw your question and gave some first general hints in the comments, but I couldn't yet figure out where this difference is originating from just from the snippets you posted there. Could you share a simple sample project with us, which reproduces this behavior? You can also send us code privately to help@realm.io.

shaozi:
I created a repository: https://github.com/shaozi/RealmSumReproduce.git

bdash:
Thanks for the excellent test case, @shaozi! I was able to reproduce the problem by following your instructions, and could easily see where the bug is.

The underlying issue is that `TableView` appears to fail to update `m_num_detached_refs` in some cases. This can result in `TableViewBase::aggregate` thinking that all refs in the table view are detached, causing it to early-exit with a result of 0. It can also result in an incorrect result when the table view is empty, since the assumption is that for an empty table view `m_num_detached_refs` will be zero. This will need to be fixed in core. I'll file a core issue once I have a chance to boil the test case down to something that uses core directly.

shaozi:
Great! Once it is fixed, will this method be faster than loop and sum?

mrackwitz:
@shaozi: Yes, once this is fixed, it should be vastly faster than looping manually over the objects. Internally we're able to iterate over all relevant values of the column and to add up those without having to instantiate full accessor objects in the Objective-C runtime for every object in the array.

shaozi:
When will the fix be merged?

jpsim:
> When will the fix be merged?

It's fixed in core, but we need a new release of our core binaries, which should happen soon.

jpsim:
A fix for this was merged in #3015. Thanks all for reporting this and helping us identify what was wrong!

PR #3015, opened 2015-12-18
Upgrade to core 0.95.6

tgoyne:
Still need to update the performance baselines (to be faster, hopefully...).


bdash:
Fixes #2984.

jpsim:
:+1: we can merge this now as long as we create a new GH issue to track updating the performance baselines.

bdash:
AFAIK we no longer need to update the performance baselines for this. :+1:

Issue #2900, opened 2015-11-24
Use default configuration in +[RLMRealm realmWithPath:]


jpsim:
This will have to be treated as a breaking change.

jpsim:
Done in #3017.

PR #3017, opened 2015-12-18
use default configuration in +[RLMRealm realmWithPath:]

jpsim:
Addresses #2900. /cc @bdash @tgoyne


tgoyne:
The Swift version also needs to be updated, since it doesn't use the one-arg realmWithPath:.

jpsim:
> The Swift version also needs to be updated, since it doesn't use the one-arg realmWithPath:.

Great point. Updated.

tgoyne:
:+1:

PR #3018, opened 2015-12-18
update all Xcode projects to recommended settings

jpsim:
silences Xcode warnings /cc @bdash

### DIFF
--- a/Realm.xcodeproj/project.pbxproj
+++ b/Realm.xcodeproj/project.pbxproj
@@ -1996,6 +1996,7 @@
 				E83EAC7B1BED3D880085CCD2 /* Release */,
 			);
 			defaultConfigurationIsVisible = 0;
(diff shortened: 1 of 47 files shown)


Realm.xcodeproj/project.pbxproj
```diff
@@ -1906,6 +1906,7 @@
 			isa = XCBuildConfiguration;
 			baseConfigurationReference = 5D659E6E1BE0398E006515A0 /* Debug.xcconfig */;
 			buildSettings = {
+				ONLY_ACTIVE_ARCH = YES;
```
bdash: This appears to be unnecessary. We already set this in Debug.xcconfig.

bdash:
Seems fine other than the unnecessary configuration setting. :+1:

PR #3027, opened 2015-12-22
Specifying strong attribute to `underlyingValue` to avoid warning

poolqf:
Hi!

I am having troubles with RLMOptionalBase. It has no specific attributes for the `underlyingValue` property, so Xcode is showing a warning:

:warning: `error: no 'assign', 'retain', or 'copy' attribute is specified - 'assign' is assumed`

As I have `Treat warnings as errors` set to YES, it is not letting me compile my project.

Appart from that, I think `assign` is not the correct assumption in this case. As Travis corroborates:
```
default property attribute 'assign' not appropriate for non-GC object [-Werror,-Wobjc-property-no-attribute]
```

Which is also making Travis not to pass

### DIFF
--- a/Realm/RLMOptionalBase.h
+++ b/Realm/RLMOptionalBase.h
@@ -29,6 +29,6 @@
 
 @property (nonatomic, unsafe_unretained) RLMProperty *property;
 
-@property (nonatomic) id underlyingValue;
+@property (nonatomic, strong) id underlyingValue;
 
 @end


bdash:
What version of Xcode are you seeing the warning about `underlyingValue` with?

Under ARC, properties of object types are considered `strong` unless explicitly declared otherwise. This means your proposed code change does not change the behavior of the generated code.

poolqf:
I'm using Xcode 7.2. 

But I can also see the "assign is assumed" log on Travis.

bdash:
Is the log from Travis publicly viewable?

bdash:
From looking at the Clang source code, this error appears to be emitted only when compiling with ARC disabled. Realm must be built with ARC enabled. Building with ARC disabled is not supported.

poolqf:
Nope... Sorry
I am pasting here the chunk of output where you can see it

PS. Using Xcode 7.1 in travis, not 7.2, sorry

```
While building module 'RealmSwift' imported from XXX-Swift.h:89:
While building module 'Realm' imported from XXX/Carthage/Build/iOS/RealmSwift.framework/Headers/RealmSwift-Swift.h:89:
In file included from <module-includes>:10:
XXX/Carthage/Build/iOS/Realm.framework/PrivateHeaders/RLMOptionalBase.h:32:1: error: no 'assign', 'retain', or 'copy' attribute is specified - 'assign' is assumed [-Werror,-Wobjc-property-no-attribute]
@property (nonatomic) id underlyingValue;
^
XXX/Carthage/Build/iOS/Realm.framework/PrivateHeaders/RLMOptionalBase.h:32:1: error: default property attribute 'assign' not appropriate for non-GC object [-Werror,-Wobjc-property-no-attribute]
2 errors generated.
While building module 'RealmSwift' imported from XXX-Swift.h:89:
In file included from <module-includes>:1:
XXX/Carthage/Build/iOS/RealmSwift.framework/Headers/RealmSwift-Swift.h:89:9: fatal error: could not build module 'Realm'
@import Realm;
 ~~~~~~~^~~~~
3 errors generated.
****-Swift.h:89:9: fatal error: could not build module 'RealmSwift'
@import RealmSwift;
 ~~~~~~~^~~~~~~~~~

```

poolqf:
The App has ARC enabled

bdash:
Can you please share the compiler command-line that appears in the `xcodebuild` output immediately preceding the errors?

poolqf:
The xctool command is the following:

`xctool build -workspace XXX.xcworkspace -scheme XXX -sdk iphoneos9.1 -configuration Release OBJROOT=/Users/travis/build/XXX/XXX/build SYMROOT=/Users/travis/build/XXX/XXX/build ONLY_ACTIVE_ARCH=NO`

bdash:
What I'm after is the command-line that `xcodebuild` used to invoke `clang`, `swiftc`, etc. `xcodebuild` displays this in its output immediately prior to any errors and warnings that the command produced. The arguments that are passed to the compiler will help us to understand what's going on here.

poolqf:
Here you are:

` /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/clang -x objective-c-header -arch armv7 -fmessage-length=0 -fdiagnostics-show-note-include-stack -fmacro-backtrace-limit=0 -std=gnu99 -fobjc-arc -fmodules -gmodules -fmodules-cache-path=/Users/travis/Library/Developer/Xcode/DerivedData/ModuleCache -fno-autolink -fmodules-prune-interval=86400 -fmodules-prune-after=345600 -fbuild-session-file=/Users/travis/Library/Developer/Xcode/DerivedData/ModuleCache/Session.modulevalidation -fmodules-validate-once-per-build-session -Xclang -fmodule-implementation-of -Xclang Redbooth -Wno-trigraphs -fpascal-strings -Os -Werror -Wno-missing-field-initializers -Wno-missing-prototypes -Werror=return-type -Wunreachable-code -Wno-implicit-atomic-properties -Werror=deprecated-objc-isa-usage -Werror=objc-root-class -Wno-arc-repeated-use-of-weak -Wduplicate-method-match -Wno-missing-braces -Wparentheses -Wswitch -Wunused-function -Wno-unused-label -Wno-unused-parameter -Wunused-variable -Wunused-value -Wempty-body -Wconditional-uninitialized -Wno-unknown-pragmas -Wno-shadow -Wno-four-char-constants -Wno-conversion -Wconstant-conversion -Wint-conversion -Wbool-conversion -Wenum-conversion -Wshorten-64-to-32 -Wpointer-sign -Wno-newline-eof -Wno-selector -Wno-strict-selector-match -Wundeclared-selector -Wno-deprecated-implementations -DCOCOAPODS=1 -DCONFIGURATION_Release -DNS_BLOCK_ASSERTIONS=1 -DOBJC_OLD_DISPATCH_PROTOTYPES=0 -isysroot `

poolqf:
@bdash any ideas?

bdash:
I'm still of the belief that it shouldn't be possible to hit this warning when building with ARC enabled. The compiler invocation you shared shows `-fobjc-arc` being passed, resulting in ARC being enabled. It's possible that there's a bug in Clang or Xcode that results in that flag flag not being propagated to the compilation of the module, but it's hard to investigate that angle without being able to reproduce the issue myself. For now I think the right thing to do is to just land this change and to attempt to investigate further if we see a similar issue in the future.

poolqf:
Thanks @bdash , if I find out more about it, I'll tell you

PR #3031, opened 2015-12-24
fix invalid Swift docs

jpsim:
some Swift docs:

- were missing parameter definitions
- were missing throws documentation
- had superfluous returns documentation
- were using documentation incompatible with SwiftLint's valid docs rule

After these changes, SwiftLint runs successfully as of [0.5.3](https://github.com/realm/SwiftLint/releases/tag/0.5.3).

/cc @bdash @tgoyne

### DIFF
--- a/RealmSwift-swift2.0/Object.swift
+++ b/RealmSwift-swift2.0/Object.swift
@@ -221,9 +221,14 @@ public class Object: RLMObjectBase {
 
     // MARK: Equatable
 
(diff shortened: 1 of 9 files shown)


tgoyne:
:+1: 

I was going to say that the `throws` docs should mention which error codes the NSError is expected to have, but the error codes aren't actually exposed in the Swift API anywhere.

Issue #601, opened 2014-07-17
Fine-grained notifications


ipeisong:
Can I find the object type info? The parameter realm seems does not contain any specific info. That's to say I need update UIs all over the app correct? Thanks.

bigfish24:
This isn't really an ideal solution, but I built a replacement for NSFetchedResultsController, RBQFetchedResultsController, that works along side a singleton class: RBQRealmNotificationManager that broadcasts object changes. 

https://github.com/Roobiq/RBQFetchedResultsController

Currently the notification manager is limited to the developer manually logging the changes when they occur (which will then be rebroadcasted after Realm updates), but an obvious improvement would be to subclass RLMObject and RLMRealm to log the change or the add/delete automatically.

xtazz:
Any update on that? This can really help especially when displaying Realm objects in the table view to nicely display row animations for deleted, inserted, added or updated rows. Thank you!

Coltspb:
:+1: I need it too

mgleason:
> Support for more fine-grained notifications (per object) will come soon.

... Any update/timeframe for this (been almost 9-months:()?

Antol:
+1

skeeet:
All hope is gone by now :(

esummers:
@skeeet Just a guess, but they will probably use the syncing architecture they are developing to implement that.

bigfish24:
While I know it would be ideal if Realm supported it natively now, if you need something soon check out: https://github.com/Roobiq/RBQRealmNotificationManager. 

It's easy to track changes when editing a RLMObject by building a wrapper method for the object that uses blocks such as:

-(void)editObject:(void(^)(id anObject))block
{
     block(self);
     [[RBQRealmChangeLogger defaultLogger] didChangeObject:self];
}

this way every time you edit an object, just use this method to get automatic change tracking:

[anObject editObject:^(id anObject) {
    anObject.property = foo;
}

the RBQRealmNotificationManager will pass along the change to any listeners. (This is what powers RBQFetchedResultsController that I referenced earlier).

mgleason:
Yea, but RBQRealmNotificationManager doesn't seem to work with swift.  

Even after fixing the #import "Realm.h" issues, it still won't fetch results - although  fetchedResultsController.performFetch returns success, but there are zero results...

bigfish24:
RE: Swift, sorry about that but we are still using objc for our projects so I haven't tested any of the code with Swift. 

[...]

PR #3043, opened 2015-12-28
added "is_dirty" flag during Realms sync process

bronenos:
After changes are written in one Realm, and before they will be synchronized with others, the next runloop cycle might gets fired. So add "is_dirty" flag that we can check to ensure other Realms are synchronized and we can call our custom completion blocks.

### DIFF
--- a/Realm/ObjectStore/impl/apple/external_commit_helper.cpp
+++ b/Realm/ObjectStore/impl/apple/external_commit_helper.cpp
@@ -327,6 +327,11 @@ void ExternalCommitHelper::listen()
 
 void ExternalCommitHelper::notify_others()
 {
+    std::lock_guard<std::mutex> lock(m_realms_mutex);
+    for (auto const& realm : m_realms) {
+        realm.realm->set_is_dirty(true);
+    }
+    
     notify_fd(m_notify_fd);
 }
 #endif
--- a/Realm/ObjectStore/shared_realm.cpp
+++ b/Realm/ObjectStore/shared_realm.cpp
@@ -97,6 +97,8 @@ Realm::Realm(Config config)
         throw RealmFileException(RealmFileException::Kind::FormatUpgradeRequired, m_config.path, "The Realm file format must be allowed to be upgraded "
         "in order to proceed.");
     }
+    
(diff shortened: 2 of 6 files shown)


tgoyne:
Can you give a concrete example of how you would use this?

bronenos:
Well, the following code really prevented me from working with unsynced main Realm in completion block. Because the next runloop cycle sometimes arrives earlier then notification of file descriptor inside Realm.

```
static func performSaving(block block: ExecutionBlock, completion: CompletionBlock?) {
  let callerQueue = NSOperationQueue.currentQueue()
  
  self.writingThread.addOperationWithBlock {
    let realm = self.currentRealm()
    try! realm.write {
      block(realm)
    }
    
    if let completion = completion {
      self.waitUntilRealmRefreshedAndExecuteBlock(completion, onQueue: callerQueue!)
    }
  }
}

private static func waitUntilRealmRefreshedAndExecuteBlock(block: dispatch_block_t, onQueue queue: NSOperationQueue) {
  queue.addOperationWithBlock {
    if self.currentRealm().isDirty {
      self.waitUntilRealmRefreshedAndExecuteBlock(block, onQueue: queue)
    }
    else {
      block()
    }
  }
}
```

jpsim:
@bronenos can you please help me understand why you'd do this rather than simply refresh the Realm? I'm not convinced that this property allows for usage scenarios not already covered.

jpsim:
@bronenos I'm closing this because I lack an understanding of the problem space.

If you don't think your use cases are covered by refreshing the Realm like I've suggested above, please open a new GitHub issue describing what you're trying to achieve and we can determine if there's either an existing way to accomplish that, or if we should add new functionality (maybe even something like what you've proposed here).

Thanks!

bronenos:
Well, if you don't mind of me forcing the `refresh` after every transaction, by design, I will try.
Thanks for your reply.

jpsim:
> Well, if you don't mind of me forcing the refresh after every transaction, by design, I will try.

Only when necessary. Whenever you would have checked `isDirty` with this change.

Note that I think many of these scenarios will be addressed once we have #601.

PR #3046, opened 2015-12-30
fixed FileAccessError

fewspider:


jpsim:
This is incorrect. `RLMError.FileAccess` is the right value: https://github.com/realm/realm-cocoa/blob/7d97479d49ea896859eb3b32d5caf18b5ce1b9ba/Realm/RLMConstants.h#L68

PR #3056, opened 2016-01-04
Forward the source error location in several Swift test helpers

tgoyne:
assertThrows(), assertSucceeds() and assertFails() captured the calling source location, but did not actually do anything with it.


bdash:
:+1:

jpsim:
:+1:

PR #3057, opened 2016-01-04
Fix using -[RLMResults setValue:forKey:] on a key used in the query

### LINKED ISSUE #3028: Crash when Results.setValue(_:forKeyPath:) mutates objects to no longer match predicate
Hi, after upgrading realm to the latest version `v0.97.0` my code crash:

```Swift
let remove = db.objects(VODBrand.self)
    .filter("topBrand = %@", RealmOptional<Bool>(true))
if remove.count > 0 {
    remove.setValue(RealmOptional<Bool>(false), forKeyPath: "topBrand")
}
```

`remove.count` have a value of 1, but `RLMCollectionSetValueForKey()` methods is looping 3 objects:

schermata 2015-12-22 alle 19 30 55

Here the crash:

```
Terminating app due to uncaught exception 'RLMException', reason: 'Index 2 is out of bounds (must be less than 1)'
```

What's up?

tgoyne:
This was broken by aadd90501c70fc09e17d04e5f25fd7b732e6a9cc, which indirectly made it so that indexInSource: brings the Results back in to sync. Changing that back would be difficult, so just directly iterate over a frozen tableview instead.

Also adds a test to ensure this usage will continue to work in the future.

Fixes #3028.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,7 +16,8 @@ x.x.x Release notes (yyyy-MM-dd)
 
 ### Bugfixes
 
-* None.
+* Fix incorrect results or crashes when using `-[RLMResults setValue:forKey:]`
+  on an RLMResults which was filtered on the key being set.
 
 0.97.0 Release notes (2015-12-17)
 =============================================================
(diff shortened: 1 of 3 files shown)


jpsim:
:+1:

PR #3058, opened 2016-01-05
track Swift version used to compile Realm Swift

jpsim:
This will give us a better idea of when it's reasonable to drop older Swift versions. /cc @tgoyne @bdash

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -817,7 +817,7 @@ case "$COMMAND" in
     # CocoaPods
     ######################################
     "cocoapods-setup")
-        if [[ "$2" != "without-core" ]]; then
+        if [[ "$2" != "swift" ]]; then
             sh build.sh download-core
             mv core/librealm.a core/librealm-osx.a
             if [[ "$REALM_SWIFT_VERSION" = "1.2" ]]; then
@@ -838,7 +838,7 @@ case "$COMMAND" in
           fi
         done
 
-        if [[ "$2" != "without-core" ]]; then
+        if [[ "$2" != "swift" ]]; then
           # CocoaPods doesn't support multiple header_mappings_dir, so combine
           # both sets of headers into a single directory
           rm -rf include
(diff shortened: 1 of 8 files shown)


tgoyne:
cocoapods-setup doesn't call set-swift-version because we'd gone with the explicit separate branches for swift versions with CP, so the file never gets created.

jpsim:
> cocoapods-setup doesn't call set-swift-version because we'd gone with the explicit separate branches for swift versions with CP, so the file never gets created.

Right. I suppose we could write the file based on the user's value of `xcrun swift --version` at `cocoapods-setup` time, although that will only get updated at each `pod install`. If https://github.com/apple/swift-evolution/pull/69 lands in Swift 2.2, then this won't be too problematic.

Can you think of a better way to do this? I wouldn't go so far as installing a build phase to update the file in `cocoapods-setup`.

jpsim:
@tgoyne @bdash CocoaPods fix in place.

build.sh
```diff
@@ -852,6 +858,8 @@ case "$COMMAND" in
             cp -R include/Realm include/realm
           fi
           touch include/Realm/RLMPlatform.h
+        else
+          echo "let swiftLanguageVersion = \"$(get_swift_version)\"" > RealmSwift/SwiftVersion.swift
```
bdash: Why is this in the `else` block? I'm not sure how this logic relates to the `without-core` argument.

jpsim: This definitely needs a comment, but it's because `without-core` is only used when installing Realm Swift.

jpsim: Actually, it makes more sense at this point to rename the `without-core` argument to `swift`.

bdash:
:+1:

build.sh
```diff
@@ -412,6 +412,12 @@ case "$COMMAND" in
         ln -s "RealmSwift-swift$version" RealmSwift
         git update-index --assume-unchanged RealmSwift || true
 
+        SWIFT_VERSION_FILE="RealmSwift/SwiftVersion.swift"
+        CONTENTS="let swiftLanguageVersion = \"$version\""
+        if ! grep "$CONTENTS" "$SWIFT_VERSION_FILE"; then
```
tgoyne: `grep -q` to avoid noise in the build logs

jpsim: Good call, thanks.

PR #3059, opened 2016-01-05
update build.sh's build_docs() for jazzy 0.5.0

jpsim:
[jazzy 0.5.0](https://github.com/realm/jazzy/releases/tag/v0.5.0) now has the ability to inject custom content into the `<head></head>` of the generated docs, which means we don't have to maintain a fork of the HTML templates just to add a site icon & the analytics script.

The default Swift version was also bumped to 2.1.1 in this version.

Note that CI can only have a single version of gems installed at a time, so it's still running 0.4.1. Once this PR has been reviewed and approved, I'll update CI to use jazzy 0.5.0.

/cc @tgoyne @bdash


docs/custom_head.html
```diff
@@ -0,0 +1,17 @@
+<link rel="icon" href="https://realm.io/img/favicon.ico">
+    <link rel="apple-touch-icon-precomposed" sizes="57x57" href="https://realm.io/img/favicon-57x57.png" />
```
bdash: The indentation looks inconsistent here. Most of these `link` elements have an unnecessary trailing `/`.

jpsim: indentation is like this on purpose to look nice in the generated HTML. I can remove the trailing `/`s though.

jpsim:
I'm installing jazzy 0.5.0 on CI and will re-run this job once the installation is complete. If you have any further comments on this, now's the time to share them.

PR #3060, opened 2016-01-05
Minor fix to give better error message for misspelled keypath operators.

astigsen:
This gives a better error message for misspelled keypath operators (and in general accidental use of `@` outside of collection operators). It used to just crash.


jpsim:
:+1: thanks alex!

PR #3067, opened 2016-01-05
Not throw and allow registering notification blocks on a thread without a runloop in special case

bigfish24:
Not throw and allow registering notification blocks on a thread without a runloop as long as we are on the same thread as the transaction, since the calling of the block won't occur via the named pipe.

This is a strange use-case but came up with `RBQFetchedResultsController` via `RBQRealmNotificationManager`: https://github.com/Roobiq/RBQFetchedResultsController/issues/52

`RBQRealmNotificationManager` registers a notification block on the same thread as the transaction, which is probably not a common use-case beyond a component like this.

### DIFF
--- a/Realm/RLMRealm.mm
+++ b/Realm/RLMRealm.mm
@@ -395,7 +395,8 @@ static void CheckReadWrite(RLMRealm *realm, NSString *msg=@"Cannot write to a re
     if (!block) {
         @throw RLMException(@"The notification block should not be nil");
     }
-    if (!RLMIsInRunLoop()) {
+    if (!RLMIsInRunLoop() &&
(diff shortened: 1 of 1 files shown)


bigfish24:
cc @tgoyne and @jpsim

tgoyne:
This just disables the runloop check entirely, as we throw an exception if you try to add a notification block from a different thread.

bigfish24:
ah yeah that's right... hmm I guess there isn't a good way to know if the notification block is only there to listen to the transaction on the current thread. So just working around the check in this specific use-case is more appropriate.

bigfish24:
ok will just close this for now

PR #3070, opened 2016-01-06
enable PR CI builds on Travis CI

jpsim:
This will serve as a backup in case our main CI setup is experiencing
difficulties.

It will also allow non-Realm employees to see the results of some

I've only ported a subset of our standard CI PR matrix here because:

  1. Travis limits their matrix builds to 50 minutes *in aggregate*
  2. Some of our standard CI PR matrix jobs require dependencies not available
     to Travis, like codesigning certificates and provisioning profiles.

The downside to integrating this is that we'll have yet another file to update
when making changes to our build system or CI setup.

This commit also avoids grepping on RealmSwift/SwiftVersion.swift when the file
doesn't exist, preventing a warning from being printed.

/cc @tgoyne @bdash @mrackwitz

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,35 @@
+language: objective-c
+osx_image: xcode7.2
+branches:
+  only: master
+script: placeholder # workaround for https://github.com/travis-ci/travis-ci/issues/4681
+matrix:
+  include:
+    - script: ./build.sh verify-swiftlint
+      env: JOB=verify-swiftlint
+      before_install: brew install swiftlint
+    - script: ./build.sh verify-docs
+      env: JOB=verify-docs
+      before_install: gem install jazzy
+    - script: ./build.sh verify-osx-swift
+      env: JOB=verify-osx-swift
+    - script: ./build.sh test-tvos
+      env: JOB=test-tvos
+    - script: ./build.sh verify-osx
+      env: JOB=verify-osx
+    - script: ./build.sh test-ios-swift
+      env: JOB=test-ios-swift
+    - script: ./build.sh test-ios-static
+      env: JOB=test-ios-static
+
+    ############################################################################
+    # These jobs pass but are disabled because they occasionally make Travis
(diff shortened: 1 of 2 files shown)


bdash:
Seems fine to me. Would be worth renaming the "default" check to something more descriptive.

.travis.yml
```diff
@@ -0,0 +1,35 @@
...
+  only: master
+script: placeholder # workaround for https://github.com/travis-ci/travis-ci/issues/4681
+matrix:
+  include:
+    - script: ./build.sh verify-swiftlint
+      env: JOB=verify-swiftlint
+      before_install: brew install swiftlint
+    - script: ./build.sh verify-docs
+      env: JOB=verify-docs
+      before_install: gem install jazzy
```
mrackwitz: We might be able to save some additional time by making use of [Travis' optional caching logic for bundler](https://docs.travis-ci.com/user/caching/#Enabling-Bundler-caching).

jpsim: Installing jazzy takes less than 30 seconds in Travis, and is only done for one of the matrix builds, and I don't know if using the bundler cache would negatively affect other builds, so I'd rather not.

Also, the docs you linked to show to specify `language: ruby`, so I'm not sure how supported it would be to use it with `language: objective-c`.

mrackwitz: >I don't know if using the bundler cache would negatively affect other builds, so I'd rather not.

Beside CocoaPods, I don't see what else could be affected by that change.

>Also, the docs you linked to show to specify `language: ruby`, so I'm not sure how supported it would be to use it with `language: objective-c`.

Directly above that subsection, in the introduction of [the section](https://docs.travis-ci.com/user/caching/#Bundler), they explain that this is available for both type of images.

jpsim:
> Seems fine to me. Would be worth renaming the "default" check to something more descriptive.

That would certainly be nice, although according to @emanuelez, the [GitHub PR Builder Jenkins plugin](https://github.com/janinko/ghprb) we use doesn't support this.

PR #3081, opened 2016-01-08
Bool and NSDate properties can be indexed.

kishikawakatsumi:
Core supports indexing Bool and NSDate properties.
realm-java also supports indexing Bool and NSDate properties.
It is better to align the specification with realm-java.

Also, indexing of Bool and NSDate had been supported implicitly in v0.96. v0.97 breaks this behavior.

cc @tgoyne @jpsim

### DIFF
--- a/Realm/ObjectStore/property.hpp
+++ b/Realm/ObjectStore/property.hpp
@@ -51,6 +51,7 @@ namespace realm {
         std::string object_type;
         bool is_primary = false;
         bool is_indexed = false;
+        bool is_indexable() const { return type == PropertyTypeInt || type == PropertyTypeBool || type == PropertyTypeString || type == PropertyTypeDate; }
         bool is_nullable = false;
 
         size_t table_column = -1;
--- a/Realm/ObjectStore/schema.cpp
+++ b/Realm/ObjectStore/schema.cpp
@@ -89,7 +89,7 @@ void Schema::validate() const
 
             // check indexable
             if (prop.is_indexed) {
-                if (prop.type != PropertyTypeString && prop.type != PropertyTypeInt) {
+                if (!prop.is_indexable()) {
                     exceptions.emplace_back(PropertyTypeNotIndexableException(object.name, prop));
                 }
             }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ x.x.x Release notes (yyyy-MM-dd)
   of the specified path was not found when a copy should be written. 
 * Greatly improve performance when deleting objects with one or more indexed
   properties.
+* Indexing `BOOL`/`Bool` and `NSDate` properties are now supported.
 
 ### Bugfixes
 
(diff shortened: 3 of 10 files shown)


Realm/ObjectStore/schema.cpp
```diff
@@ -89,7 +89,7 @@ void Schema::validate() const
 
             // check indexable
             if (prop.is_indexed) {
-                if (prop.type != PropertyTypeString && prop.type != PropertyTypeInt) {
+                if (prop.type != PropertyTypeString && prop.type != PropertyTypeInt && prop.type != PropertyTypeBool && prop.type != PropertyTypeDate) {
```
jpsim: it'd be preferable to have an `is_indexable` bool on property type rather than inline here.

jpsim:
I wasn't aware that these types were now indexable in core. Thanks for the PR. I'm certainly in favor of doing this, at the very least for compatibility with files created in java.

This needs a changelog entry and to update the Objective-C tests. I would also prefer if we had more test coverage using indexed properties than we do now, but if it's sufficiently tested in core, that's even better.

pjebs:
When will this be ready for Realm-iOS?

jpsim:
@pjebs this needs to undergo more review from the rest of the cocoa team before we can merge. It won't be ready until that happens.

jpsim:
@kishikawakatsumi it looks like Java also supports indexing `byte[]` fields, so we should add `NSData` to this.

> This needs a changelog entry and to update the Objective-C tests. I would also prefer if we had more test coverage using indexed properties than we do now, but if it's sufficiently tested in core, that's even better.

I still think we need more test coverage than just checking to see if marking a property as indexed gets reflected in its `RLMProperty` representation.

bdash:
Core doesn't support indexing the binary columns that we use to store `NSData` properties. The Java documentation don't mention `byte[]` as supported either, only `byte`.

jpsim:
> Core doesn't support indexing the binary columns that we use to store NSData properties. The Java documentation don't mention byte[] as supported either, only byte.

Sorry, my mistake. I must have misread that.

jpsim:
Closes #3104

kishikawakatsumi:
Addressed feedback.
Add Objective-C tests.
Add Swift tests for other indexable properties.

Can you review this @jpsim

Realm/ObjectStore/property.hpp
```diff
@@ -51,6 +51,7 @@ namespace realm {
         std::string object_type;
         bool is_primary = false;
         bool is_indexed = false;
+        bool is_indexable() const { return type == PropertyTypeInt || type == PropertyTypeBool || type == PropertyTypeString || type == PropertyTypeDate; };
```
tgoyne: Extra trailing semicolon.

tgoyne:
:+1:

jpsim:
Looks good! Just needs a changelog entry then :+1:

kishikawakatsumi:
Added changelog.

jpsim:
:+1:

PR #3084, opened 2016-01-08
verify that binaries have bitcode if necessary

jpsim:
/cc @tgoyne

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -158,7 +158,11 @@ build_combined() {
     clean_retrieve $os_path $out_path $product_name
 
     # Combine ar archives
-    xcrun lipo -create "$simulator_path/$binary_path" "$os_path/$binary_path" -output "$out_path/$product_name/$module_name"
+    LIPO_OUTPUT="$out_path/$product_name/$module_name"
+    xcrun lipo -create "$simulator_path/$binary_path" "$os_path/$binary_path" -output "$LIPO_OUTPUT"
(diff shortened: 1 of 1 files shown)


tgoyne:
:+1:

build.sh
```diff
@@ -813,6 +817,36 @@ case "$COMMAND" in
...
+        if [[ ${#archs_array[@]} < 2 ]]; then
+            exit 1 # Early exit if not a fat binary
+        fi
+
+        TEMPDIR=$(mktemp -d $TMPDIR/realm-bitcode-check.XXXX)
+
+        for arch in $archs; do
+            lipo -thin "$arch" "$BINARY" -output "$TEMPDIR/$arch"
+            if otool -l "$TEMPDIR/$arch" | grep -q "segname __LLVM"; then
+                exit 0
```
bdash: If I'm reading this right, we're only checking a single slice. Are we assuming that if one slice has bitcode that they all do? If so, why not check only the first architecture and return a result based on that?

jpsim: simulator slices don't have bitcode. We could hardcode a list of architectures where we'd expect bitcode if you like.

bdash: Doesn't seem worth it. A comment could help clarify the current behavior and reason for it.

jakajancar:
@jpsim @bdash I believe this will detect `-fembed-bitcode-marker`, even if there is not bitcode (`-fembed-bitcode`).

jpsim:
@jakajancar can you recommend an improvement?

PR #3089, opened 2016-01-11
Delete RLMPredicateUtil

tgoyne:
It's a bunch of code that I find thoroughly baffling and it doesn't actually do much of anything.

@bdash


bdash:
:+1:

tgoyne:
CI failures are a mix of the iOS 7 simulator failing to mmap libraries and the download of the source zip from github getting interrupted. Former issue went away after restarting the VM and all of the jobs hit by the latter passed at least once, so errors seem spurious.

PR #3090, opened 2016-01-12
Refactor the query tests a bit

tgoyne:
Except for where the other variants are specifically being tested, switch everything to using `objectsWhere:` rather than a random mishmash of that, `objectsInRealm:where` and `objectsWithPredicate:`.

Remove the last few uses of the dynamic API in places where the dynamic API isn't specifically being tested.

Add a macro `RLMAssertCount()` which concisely constructs a query and checks the count. The main point of this is to add a convenient hook point for testing async queries, but it also makes all of the tests way less verbose and more consistent.


bdash:
:+1:

jpsim:
:+1:

PR #3110, opened 2016-01-15
[build.sh] Print stderr when curl fails

### LINKED ISSUE #3023: Better logging / instructions on issues when downloading core
Our `build.sh` script intentionally deactivates all logging, which might happen by curl. That's nice to have non-verbose output for the happy path, but if something goes wrong, you're out of luck.
We should improve the logging there, as there can be various reasons, why that might fail.

mrackwitz:
Fixes #3023.

The used CURL options are the following for reference:
-f: (HTTP) Fail silently (no output at all) on server errors.
-s: Silent or quiet mode.
-S: When used with -s it makes curl show an error message if it fails.
-L: Follow redirects.
-o: Write output to given path

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -282,8 +282,15 @@ download_core() {
     CORE_TMP_TAR="${TMP_DIR}/core-${REALM_CORE_VERSION}.tar.bz2.tmp"
     CORE_TAR="${TMP_DIR}/core-${REALM_CORE_VERSION}.tar.bz2"
     if [ ! -f "${CORE_TAR}" ]; then
-        curl -f -L -s "https://static.realm.io/downloads/core/realm-core-${REALM_CORE_VERSION}.tar.bz2" -o "${CORE_TMP_TAR}" ||
-          (echo "Downloading core failed. Please try again once you have an Internet connection." && exit 1)
+        local CORE_URL="https://static.realm.io/downloads/core/realm-core-${REALM_CORE_VERSION}.tar.bz2"
+        set +e # temporarily disable immediate exit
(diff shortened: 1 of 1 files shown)


build.sh
```diff
@@ -282,8 +282,15 @@ download_core() {
     CORE_TMP_TAR="${TMP_DIR}/core-${REALM_CORE_VERSION}.tar.bz2.tmp"
     CORE_TAR="${TMP_DIR}/core-${REALM_CORE_VERSION}.tar.bz2"
     if [ ! -f "${CORE_TAR}" ]; then
-        curl -f -L -s "https://static.realm.io/downloads/core/realm-core-${REALM_CORE_VERSION}.tar.bz2" -o "${CORE_TMP_TAR}" ||
-          (echo "Downloading core failed. Please try again once you have an Internet connection." && exit 1)
+        local CORE_URL="https://static.realm.io/downloads/core/realm-core-${REALM_CORE_VERSION}.tar.bz2"
+        set +e # temporarily disable immediate exit
+        local ERROR # sweeps the exit code unless declared separatedly
```
jpsim: "separatedly" -> "separately"

jpsim:
LGTM :+1:

tgoyne:
In general I'd prefer to use the long-form of flags whenever they aren't totally obvious. Here that'd be `curl --fail --silent --show-error --location "$CORE_URL" -o "${CORE_TMP_TAR}"` which I think is somewhat more understandable.

mrackwitz:
Fair point. I saw that as well, but decided instead to document them in the commit message. I don't think that e.g. the meaning of `--location` is really obvious.

Issue #3097, opened 2016-01-13
ignoredProperties are not ignored


danielsaidi:
Hi,

I am using Realm Swift 0.97.0 and am having the same problem as previously reported (e.g. https://github.com/realm/realm-cocoa/pull/2734).

In short, I am having a computed UIImage property, let's call it "image":

```
var image: UIImage {
   get { ... }
   set { ... }
}
```

I used to have the type be UIImage?, but refactored it to be UIImage since the link above said that the bug only affected optional properties in Swift.

In order to not have the UIImage property causing problems, I add it to the ignoredProperties list, so that that function looks like this:

```
override class func ignoredProperties() -> [String] {
   return ["image", "hasId"]
}
```

However, this does not work. I still get the following error:

Uncaught exception: 'UIImage' is not supported as an RLMObject property. All properties must be primitives, NSString, NSDate, NSData, RLMArray, or subclasses of RLMObject.

mrackwitz:
Could you please share your full model definition?

danielsaidi:
Sure, I have posted the complete model in terms of design and inheritance, but have removed redundant properties to make the post manageable.

My entity class (let's call it Item) inherits from a base class called Entity:

```
class Entity: RLMObject {
    
    dynamic var id = ""
    
    var hasId: Bool {
        get { return id.characters.count > 0 }
    }
    
    override class func primaryKey() -> String {
        return "id"
    }
    
    override class func ignoredProperties() -> [String] {
        return ["hasId"]
    }
}
```

The Item class then specifies some new properties, as such:

```
class Card: Entity {

    dynamic var name = ""
    dynamic var imageString = ""
    
    
    var image: UIImage {
        get { return stringToImage(imageString) }
        set { imageString = imageToString(newValue) ?? "" }
    }
    

    override class func ignoredProperties() -> [String] {
        return ["image", "hasId"]
    }


    private func imageToString(image: UIImage?) -> String {
        if (image == nil) {
            return ""
        }
        let data = UIImagePNGRepresentation(image!)
        return data != nil ? data!.base64EncodedStringWithOptions(.Encoding64CharacterLineLength) : ""
    }
    
    private func stringToImage(string: String) -> UIImage {
        if (string.characters.count == 0) {
            return UIImage()
        }
        
        let imageString = string.stringByReplacingOccurrencesOfString("\r\n", withString: "")
        let options = NSDataBase64DecodingOptions(rawValue: 0)
[...]

PR #3120, opened 2016-01-20
Use the right Swift ObjectUtil class depending on parent class

mrackwitz:
`RealmSwiftObjectUtil` doesn't support `RLMObject` classes (and shouldn't from my understanding). That can cause that just linking Realm Swift screws up `RLMObject` model definitions from Swift as seen in #3097.
This introduces the protocol `ObjectProtocol` implemented by RealmSwiftObject and SwiftFakeObject  and tests for the conformance to that as a workaround to allow invalid test classes.
/c @jpsim @tgoyne

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,8 @@ x.x.x Release notes (yyyy-MM-dd)
 
 ### Bugfixes
 
-* None.
+* Using Realm Objective-C from Swift while having Realm Swift linked no longer causes that the
+  declared `ignoredProperties` are not taken into account.
 
 0.98.2 Release notes (2016-02-18)
 =============================================================
--- a/Realm/RLMObjectBase.mm
+++ b/Realm/RLMObjectBase.mm
@@ -188,6 +188,10 @@ static id RLMValidatedObjectForProperty(id obj, RLMProperty *prop, RLMSchema *sc
     return [RLMSchema sharedSchemaForClass:self.class];
 }
 
++ (Class)objectUtilClass:(BOOL)isSwift {
+    return RLMObjectUtilClass(isSwift);
+}
+
 - (NSString *)description
 {
     if (self.isInvalidated) {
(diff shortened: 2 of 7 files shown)


tgoyne:
This makes it so that Swift-defined RLMObject subclasses are not treated as Swift classes in some of the places that they need to be.

mrackwitz:
@tgoyne: The only places which are concerned from this changeset are within `RLMObjectSchema` as the property `isSwiftClass` continues to reflect where the class was defined instead what superclass it is inheriting from. Can you please point out where exactly you think something goes wrong?

tgoyne:
Swift-defined RLMObject subclasses need to use initSwiftPropertyWithName for RLMArray properties, and the Any/String? error message applies to them.

mrackwitz:
Ah right, so I would pass both `isSwiftClass` and `isSwiftObject` into the method to change the handling there? Alternatively, we could have a further function on base of `RLMObjectUtil` which takes a class and goes over the runtime, but that could mean a slight performance degradation in comparison.

tgoyne:
Is there any obvious reason it wouldn't work to have a class method on RLMObject and RealmSwift.Object (and the fake base classes) which returns the appropriate ObjectUtil class? It seems like that'd both simpler and faster than what the code is currently doing. The only downside I can think of is that it might not be possible to fully prevent the Swift version from showing up in autocomplete (the obj-c version just needs to not be in the public header).

mrackwitz:
That works fine, but it pesters indeed Realm Swift's public API.

mrackwitz:
I addressed what @tgoyne proposed here and rebased it now. So I'm waiting now for CI feedback and would merge it then after adding a changelog item, if no one objects.

bdash:
Can a test case be written for this?

mrackwitz:
Our current test harness doesn't really allow testing failures at link time.
Theoretically we could let our tests for Realm Objective-C link also again Realm Swift, but that could allow that other issues could be introduced.

PR #3123, opened 2016-01-21
Implement +automaticallyNotifiesObserversForKey: on RLMObjectBase

### LINKED ISSUE #3111: Adding automaticallyNotifiesObserversOfFoo for each property hurts Realm initialization time
A user emailed help@realm.io about the performance of Realm initialization with Realm 0.97.0. Their application has a very large data model, comprising of over 90 subclasses of `RealmSwift.Object`. They provided an Instruments trace that shows the initialization time being dominated by calls to `class_addMethod` within `RLMAccessorClassForObjectClass`. These calls add implementations of `+automaticallyNotifiesObserversOf<Property>` for each property to prevent the KVO machinery from sending notifications for properties that we manage. Each of the calls to `class_addMethod` flushes the Objective-C runtime cache for the class, which is relatively expensive when done repeatedly. It looks like KVO supports disabling automatic notification using a single `+automaticallyNotifiesObserversForKey:` method. Depending on how performance-sensitive this is we could either implement this directly on `RLMObject`, or add custom implementations on each accessor class.

bdash:
Prior to this `RLMAccessorClassForObjectClass` added an implementation of `+automaticallyNotifiesObserversOf<Property>` to the class for each property in the schema. Given that the results of these methods is cached by Foundation after the first call it's not necessary to optimize them in this manner. In fact, doing so makes the creation of accessor classes during schema initialization much slower due to the increased number of calls to `class_addMethod`.

Fixes #3111.

/cc @tgoyne

### DIFF
--- a/Realm/RLMObjectBase.mm
+++ b/Realm/RLMObjectBase.mm
@@ -302,6 +302,19 @@ static id RLMValidatedObjectForProperty(id obj, RLMProperty *prop, RLMSchema *sc
     _observationInfo->kvoInfo = observationInfo;
 }
 
++ (BOOL)automaticallyNotifiesObserversForKey:(NSString *)key
+{
+    const char* className = class_getName(self);
+    const char accessorClassPrefix[] = "RLMAccessor_";
+    if (!strncmp(className, accessorClassPrefix, sizeof(accessorClassPrefix) - 1)) {
+        if (self.sharedSchema[key]) {
+            return NO;
+        }
+    }
+
+    return [super automaticallyNotifiesObserversForKey:key];
+}
+
 @end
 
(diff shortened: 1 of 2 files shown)


tgoyne:
:+1:

Realm/RLMObjectBase.mm
```diff
@@ -302,6 +302,19 @@ - (void)setObservationInfo:(void *)observationInfo {
     _observationInfo->kvoInfo = observationInfo;
 }
 
++ (BOOL)automaticallyNotifiesObserversForKey:(NSString *)key
+{
+    const char* className = class_getName(self);
```
mrackwitz:
```diff
-const char* className
+const char *className
```
 :point_up: :grimacing:

PR #3135, opened 2016-01-25
Pull a bunch of incidental changes from the ObjectStore repo

tgoyne:
There are a bunch of little non-functional differences between the two copies of the code that made diffing them pointlessly annoying. This brings them back into sync except for where the differences are related to actual functional changes.


jpsim:
:+1:

PR #3137, opened 2016-01-26
fix CocoaPods installation examples

jpsim:
by checking if there's a workspace after running pod install.

So it turns out that CocoaPods *doesn't* fail to install Realm on the first try, but rather that we were testing our installation examples incorrectly.


jpsim:
Fixes #2701.

PR #3143, opened 2016-01-26
Restore test location reporting in Swift tests

tgoyne:
Switch between String and StaticString based on the Xcode version.


jpsim:
Awesome!

Issue #3054, opened 2016-01-04
EXC_BAD_ACCESS and SIGABRT when using Notifications


angerman:
After distilling the code in question down to a workable sample, I'd like to report an issue with Views and Notifications.
Assume the following model:
```
class Person : Object {
    dynamic var name : String = ""
    var children = List<Person>()
}
```
The reclusive nature doesn't have any impact as far as I can tell. Using the alternative of three models to crete a hierarchy A -> B -> C exhibits the same.

Using the following code to create three objects, link them in a chain and finally deleting objects from that chain at different positions, yields different results for the observer.
```
                let realm = try! Realm()
                let p1 = Person(); p1.name = "Tim"
                let p2 = Person(); p2.name = "Tom"; p2.children.append(p1)
                let p3 = Person(); p3.name = "Max"; p3.children.append(p2)
                // Max -> Tom -> Tim
                try! realm.write {
                    realm.add([p1,p2,p3])
                }
                
                let max = realm.objects(Person).filter("name = 'Max'").first!
                let children = max.children.filter("name = name")
                let grandchildren = children.first!.children.filter("name = name")
                
                let token = realm.addNotificationBlock { notification, realm in
                    print("---")
                    print(max)
                    print("---")
                    print(children)
                    print("---")
                    print(grandchildren)
                }
                try! realm.write {
                    // realm.delete(p3) // this will cause EXC_BAD_ACCESS in Realm`realm::TableViewBase::is_in_sync() const
                    // realm.delete(p2) // this will cause SIGABRT in Realm`realm::LinkView::size(...) const with '[realm-core-0.95.5] Assertion failed: is_attached()'
                    realm.delete(p1) // this will succeed.
                }
```
The `filter` used here is clearly the identity filter, but allows us to go from Arrays to Results, for illustrative purposes. In our non sample code, we use filters that do not necessarily resemble the identity.

The underlying reason seems obvious (a view who's predicate chain is invalidated, does not make a lot of sense).

This sample code was compiled using Cocoapods with
```
Using Realm (0.97.0)
Using RealmSwift (0.97.0)
```

The question is, how to detect if an object in a notification block is still valid / attached, prior to accessing it?

mrackwitz:
Thanks for bringing this to our attention and the work you put in to distill it down to a workable sample.

[...]

PR #3147, opened 2016-01-27
add workaround for #3054

jpsim:
This test would crash with core 0.95.5. /cc @bdash 

I'm sure the test case can be further reduced and probably even be done without adding a new test object model, but I haven't been able to do so myself.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,6 +31,7 @@ x.x.x Release notes (yyyy-MM-dd)
 * Fix crashes or incorrect results for queries that use relationship equality
   in cases where the `RLMResults` is kept alive and instances of the target class
   of the relationship are deleted.
+* Fix crash when accessing Results dependent on a deleted relationship.
 
 0.97.0 Release notes (2015-12-17)
 =============================================================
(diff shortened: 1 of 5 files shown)


bdash:
IIRC this solution only fixes half of the problem. In particular, I think that if you reorder the calls to `firstObject` and `count` in your test case you'll still crash. I'd suggest waiting for the complete fix to come from core (realm/realm-core#1459) unless there's an urgent need for a partial workaround.

jpsim:
> IIRC this solution only fixes half of the problem. In particular, I think that if you reorder the calls to firstObject and count in your test case you'll still crash. I'd suggest waiting for the complete fix to come from core (realm/realm-core#1459) unless there's an urgent need for a partial workaround.

Yes, this is only a partial fix, but if our goal is still to cut a release with the current version of core, better to ship a partial fix than no fix at all.

Alternatively, we could delay our release until a new core release is cut with realm/realm-core#1459. I don't think I'll be able to completely fix the SF's CI environment today, so if this happens soon, then yes we should wait.

bdash:
If we don't have a good reason to land a partial fix, we should hold off until the complete fix is available. I don't see this necessitating that we delay our planned bug fix release.

jpsim:
Preventing user apps from crashing even with proper usage of our API is reason enough to land a partial fix if you ask me, but if a core release with realm/realm-core#1459 is imminent anyway, then I'm fine closing this and waiting instead.

jpsim:
After talking more about this offline with @bdash, I've decided to close this and wait for the right fix rather than merge this very brittle fix.

PR #3153, opened 2016-01-29
Don't invoke 'sh build.sh binary-has-bitcode' unless we're building with Xcode > 6

bdash:
Xcode 6 doesn't support bitcode so there's no reason to try to verify that the binary we produced contains it.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -160,9 +160,14 @@ build_combined() {
     # Combine ar archives
     LIPO_OUTPUT="$out_path/$product_name/$module_name"
     xcrun lipo -create "$simulator_path/$binary_path" "$os_path/$binary_path" -output "$LIPO_OUTPUT"
-    if [[ "$destination" != "" ]] && [[ "$config" == "Release" ]]; then
+
+    if [[ $(xcode_major_version) != "6" && "$destination" != "" && "$config" == "Release" ]]; then
         sh build.sh binary-has-bitcode "$LIPO_OUTPUT"
     fi
+}
+
+xcode_major_version() {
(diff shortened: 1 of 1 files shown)


build.sh
```diff
@@ -160,11 +160,16 @@ build_combined() {
     # Combine ar archives
     LIPO_OUTPUT="$out_path/$product_name/$module_name"
     xcrun lipo -create "$simulator_path/$binary_path" "$os_path/$binary_path" -output "$LIPO_OUTPUT"
-    if [[ "$destination" != "" ]] && [[ "$config" == "Release" ]]; then
+
+    if [[ $(xcode_major_version) != "6" && "$destination" != "" && "$config" == "Release" ]]; then
```
jpsim: Could just check that `REALM_SWIFT_VERSION != 1.2`, it should always be set.

bdash: I thought about that, but it didn't feel right to me since the relationship between bitcode and the Swift version isn't as clear as the relationship between bitcode and the Xcode version. If you think it's simpler, I can switch it though.

bdash: I guess since we use `REALM_SWIFT_VERSION` to select the bitcode or non-bitcode version of core, it's sensible to use it here too.

jpsim: We use the swift version as an Xcode version check in lots of places in `build.sh`. We should really use a variable more appropriately named, but just need to do the check once.

jpsim:
Wow, a passing build! Quick, let's merge before something else breaks!

(implementation looks good. :+1:)

PR #3154, opened 2016-01-29
fix use of grep in build.sh binary-has-bitcode

jpsim:
I could reproduce the `build.sh package-ios-static` failure locally, and this fixed it, strangely enough. /cc @bdash

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -827,7 +827,9 @@ case "$COMMAND" in
 
     "binary-has-bitcode")
         BINARY="$2"
-        if otool -l "$BINARY" | grep -q "segname __LLVM"; then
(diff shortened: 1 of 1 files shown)


jpsim:
This passed the `objc_ios` job of `releasability_pipeline`: https://ci.realm.io/job/releasability_pipeline/1317/console

bdash:
I really wish we understood why this was a problem. We use `grep -q` in a few other places in build.sh, and it'd be unfortunate if they also weren't behaving as we expect.

jpsim:
As far as I can tell, the `-q` flag does change the result in some cases: https://github.com/freebsd/freebsd/blob/master/usr.bin/grep/util.c#L211-L238

PR #3173, opened 2016-02-03
Remove support for Swift 1.2 from Realm Swift

### LINKED ISSUE #3106: Remove Swift 1.x support
This should be done prior to releasing 0.98.

bdash:
A small amount of support for Swift 1.2 still remains for sake of running Realm Objective-C's unit tests with Xcode 6.x. Numerous references to a `REALM_SWIFT_VERSION` of 1.2 remain in `build.sh` as that is how we currently distinguish between Xcode 6.x and Xcode 7.x builds, which is necessary for deciding to use a bitcode or non-bitcode version of core.

Fixes #3106.


build.sh
```diff
@@ -423,10 +423,13 @@ case "$COMMAND" in
         ln -s "RealmSwift-swift$version" RealmSwift
         git update-index --assume-unchanged RealmSwift || true
 
-        SWIFT_VERSION_FILE="RealmSwift/SwiftVersion.swift"
-        CONTENTS="let swiftLanguageVersion = \"$version\""
-        if [ ! -f "$SWIFT_VERSION_FILE" ] || ! grep -q "$CONTENTS" "$SWIFT_VERSION_FILE"; then
-            echo "$CONTENTS" > "$SWIFT_VERSION_FILE"
+        # Only write SwiftVersion.swift if RealmSwift supports the given version of Swift.
+        if [[ -e "RealmSwift-swift$version" ]]; then
```
jpsim: Why not move this check higher in `set-swift-version`, failing as early as possible when building with an unsupported version of Swift, avoiding the extra script that was added in your later commit?

jpsim:
`Realm/Tests/Swift1.2` should also be removed.

build.sh
bdash: `set-swift-version` is called when building the Objective-C framework since the Objective-C tests include Swift source files, and we need to set up the appropriate symlink to find the test files.

bdash:
Why should `Realm/Tests/Swift1.2` be removed? It still seems useful to verify that the Objective-C framework can be used from Swift 1.2, since AFAIK that's not something we can prevent.

jpsim:
Good point. Ok, let's keep that in there.

tgoyne:
The ReactKit example should probably be ported to Swift 2 rather than deleted entirely (at the time it was written ReactKit didn't support swift 2 yet). Otherwise everything looks correct.

jpsim:
:+1:

tgoyne:
:+1:

PR #3174, opened 2016-02-04
Use realm::List in RLMArrayLinkView

tgoyne:
The only functional changes should be in the error handling. Some of the newly added tests fail on master due to c++ exceptions not being properly wrapped in a few spots, and some edge cases threw exceptions with incorrect or unhelpful error messages.

The main point of this is that it's a prereq for future objectstore changes which actually add new functionality to List.


jpsim:
Looks good to me! :+1:

bdash:
:+1:

jpsim:
Is this blocked on anything @tgoyne?

PR #3182, opened 2016-02-05
Fix a crash when a deleted LinkView is implicitly handed over for bg query updates

tgoyne:
See https://secure.helpscout.net/conversation/168310017/2922.

https://gist.github.com/tgoyne/fd89f097d1b44c8e3f9c reproduces the crash that this fixes, but I couldn't figure out how to make it deterministic without the non-test change. The use of addNotificationBlock is just to make it insensitive to the foibles of the optimizer and is not required to hit the bug.

This is probably better fixed in core (and in fact some core changes are needed since the post-0.96.0 changes to deleted linkview handling break this workaround), but since it's a regression from 0.97.1 IMO we need to ship a fix promptly.

@bdash @jpsim

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,19 @@
+x.x.x Release notes (yyyy-MM-dd)
+=============================================================
+
+### API breaking changes
+
+* None.
+
+### Enhancements
+
+* None.
+
+### Bugfixes
+
+* Fix an occasional crash when a `RLMArray`/`List` which has previously been
+  filtered is deleted.
+
 0.98.0 Release notes (2016-02-04)
 =============================================================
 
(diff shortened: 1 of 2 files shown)


tgoyne:
Note that the "can't be tested deterministically" is something I'm working on solving as part of the fine-grained notifications stuff as that introduces a lot more things that really need to be tested which can't with the current approach.

bigfish24:
:+1: tested against assertion failure I observed in 0.98.0 and this fixes it

jpsim:
:+1:

If we're to cut a point release with this change, we should first cut a release of core containing realm/realm-core#1459 & realm/realm-core#1469 (among other fixes).

bigfish24:
:+1: this did fix the bug I was observing (even though I confused myself that it wasn't for a bit...)

tgoyne:
It turns out that this was actually fixed by https://github.com/realm/realm-core/pull/1469 and I apparently did something stupid when I previously tried to check if it was.

Issue #3052, opened 2016-01-03
CocoaPods 1.0.0.beta.X can't integrate Realm as a dynamic framework


icanzilb:
Just updated my project to cocoapods 1.0.0 RC and my realm to 0.97 but can't get the realm pod to compile anymore. Installed + reinstalled few times + cleaned derived data but the error is still here.

![image](https://cloud.githubusercontent.com/assets/633862/12078574/bf95e320-b217-11e5-87c5-56104faa7bd7.png)

Xcode can't find a bunch of files in Realm/include/realm and can't compile the Realm pod. Maybe an issue with the 1.0.0 version of Cocoapods?

icanzilb:
The problem exists also in a brand new project. Xcode 7.2 - create new single view project - add a basic podfile:

use_frameworks!

target 'Realm-test1' do
    pod 'RealmSwift'
end

Realm pod doesn't compile anymore with the same error:

![image](https://cloud.githubusercontent.com/assets/633862/12079647/17ddf01c-b241-11e5-89dc-28a4204a17c4.png)

weyert:
Yes, I have the same problem. If I try to build Realm myself for my iOS project I don't have this problem but then it's complaining about unsupported binaries when archiving it for TestFlight.

jpsim:
I've started looking into this. The issue seems limited to CocoaPods 1.0.0RC1 and only with `use_frameworks!`. Doesn't happen with static builds. I'll have to spend more time looking into this to figure out what's happening. In the meantime, please use a stable release of CocoaPods with Realm as dynamic frameworks.

cday:
I ran into the same issue. Tried to resolve the issue by getting rid of local changes and moving CocoaPods back to 0.39.0. However, once I was back down to 0.39.0 anytime I did `pod install` I am getting the same error. I am basically stuck in a situation where I cannot add or update pods to my project.

davejlong:
I find myself unable to use Cocoapods 0.39, because Alamofire only support the 1.0 betas now. Has anyone found a solution that works with both Alamofire and Realm?

jpsim:
@davejlong no this issue is still not fixed. I suggest you use the last version of Alamofire that works with CocoaPods 0.39... which is still the latest stable version of CocoaPods, so it's strange that Alamofire would deprecate its support for it.

mishagray:
[...]

PR #3188, opened 2016-02-08
Fix warnings when building via CocoaPods

### LINKED ISSUE #3178: Warning: Multiple build commands for output file
I've just updated to Realm 0.98 (via Pods) and now I have these two  warnings in Xcode 7.2.1:

> Check dependencies
> 
> Warning: Multiple build commands for output file /Users/Minikin/Library/Developer/Xcode/DerivedData/appName-dabrczkkehpfzsdbpumeqbbdscji/Build/Products/Debug-iphonesimulator/Realm.framework/PrivateHeaders/cached_realm.hpp
> Warning: Multiple build commands for output file /Users/Minikin/Library/Developer/Xcode/DerivedData/appName-dabrczkkehpfzsdbpumeqbbdscji/Build/Products/Debug-iphonesimulator/Realm.framework/PrivateHeaders/external_commit_helper.hpp

Thank you.

tgoyne:
Only include headers needed for the Realm.Private module in private_header_files and just preserve_path the assembled include directory. This makes it not copy a whole bunch of files into the framework which don't need to be copied, and in some cases were generating warnings due to headers with the same file name at different paths.

This requires fixing all of the objectstore include paths to not rely on Xcode's header mapping stuff, which needed to be done anyway.

Also put the core headers in a different directory from the obj-c headers to remove the need to do convoluted things to handle case-sensitive filesystems.

Fixes #3178.

### DIFF
--- a/Realm.podspec
+++ b/Realm.podspec
@@ -39,19 +39,11 @@ Pod::Spec.new do |s|
                               'include/Realm/RLMObjectBase_Dynamic.h'
 
                               # Realm.Private module
-  private_header_files      = 'include/Realm/RLMAccessor.h',
-                              'include/Realm/RLMArray_Private.h',
+  private_header_files      = 'include/Realm/*_Private.h',
+                              'include/Realm/RLMAccessor.h',
                               'include/Realm/RLMListBase.h',
-                              'include/Realm/RLMMigration_Private.h',
-                              'include/Realm/RLMObjectSchema_Private.h',
                               'include/Realm/RLMObjectStore.h',
-                              'include/Realm/RLMObject_Private.h',
-                              'include/Realm/RLMOptionalBase.h',
-                              'include/Realm/RLMProperty_Private.h',
-                              'include/Realm/RLMRealmConfiguration_Private.h',
-                              'include/Realm/RLMRealm_Private.h',
-                              'include/Realm/RLMResults_Private.h',
-                              'include/Realm/RLMSchema_Private.h'
+                              'include/Realm/RLMOptionalBase.h'
 
   source_files              = 'Realm/*.{m,mm}',
                               'Realm/ObjectStore/*.cpp',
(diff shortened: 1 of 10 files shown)


jpsim:
Does this help address #3052?

tgoyne:
No, that still fails because the project generated by cocoapods 1.0 tries to copy the headers to `Realm.framework/Realm` (which fails due to that not being a directory) rather than `Realm.framework/Headers/`.

Realm.podspec
```diff
@@ -33,22 +33,31 @@ Pod::Spec.new do |s|
...
                               'include/Realm/RLMObjectBase_Dynamic.h'
 
-  private_header_files      = 'include/Realm/*{Accessor,RealmUtil,ListBase,ObjectStore,Private,SwiftSupport}.h',
-                              'include/realm.hpp',
-                              'include/realm/realm_nmmintrin.h',
-                              'include/realm/util/*.h',
-                              'include/realm/**/*.hpp',
-                              'include/Realm/**/*.hpp'
+                              # Realm.Private module
+  private_header_files      = 'include/Realm/RLMAccessor.h',
```
jpsim: all the hardcoded paths are unfortunate here. Can't we glob `include/Realm/*_Private.h` and then explicitly include the 4 others?

build.sh
```diff
@@ -889,25 +889,19 @@ case "$COMMAND" in
...
-          cp Realm/ObjectStore/impl/apple/*.hpp include/Realm/impl/apple
-          # Create lowercase `realm` header directory for a case-sensitive filesystem.
-          if [ ! -e include/realm ]; then
-            cp -R include/Realm include/realm
-          fi
-          touch include/Realm/RLMPlatform.h
+          mkdir -p include
+          mv core/include include/core
+
+          mkdir -p include/impl/apple
```
jpsim: at least this simplifies our case-sensitive file system support (in theory)

jpsim:
Ideally we'd reduce how many paths are hardcoded in the podspec, but if we can't for whatever reason, that shouldn't block us from merging. :+1:

jpsim:
:+1:

PR #3189, opened 2016-02-09
Update to core 0.96.1

### LINKED ISSUE #3054: EXC_BAD_ACCESS and SIGABRT when using Notifications

tgoyne:
Closes #3054.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,11 @@
+x.x.x Release notes (yyyy-MM-dd)
+=============================================================
+
+### Bugfixes
+
+* Fix crashes when deleting an object containing an `RLMArray`/`List` which had
+  previously been queried.
+
 0.98.0 Release notes (2016-02-04)
 =============================================================
 
--- a/Realm/ObjectStore/impl/transact_log_handler.cpp
+++ b/Realm/ObjectStore/impl/transact_log_handler.cpp
@@ -133,8 +133,9 @@ public:
     bool insert_substring(size_t, size_t, size_t, StringData) { return true; }
     bool erase_substring(size_t, size_t, size_t, size_t) { return true; }
     bool optimize_table() { return true; }
-    bool set_int_unique(size_t, size_t, int_fast64_t) { return true; }
-    bool set_string_unique(size_t, size_t, StringData) { return true; }
+    bool set_int_unique(size_t, size_t, size_t, int_fast64_t) { return true; }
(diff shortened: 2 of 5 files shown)


jpsim:
This should close #3054. I had a unit test for this in #3147 which would be nice to include here.

tgoyne:
Travis is consistently failing with simulator issues, but this now passes on jenkins.

jpsim:
:+1:

PR #3202, opened 2016-02-11
Fix reporting of inWriteTransaction in notifications triggered by beginWrite

### LINKED ISSUE #3199: Realm is left in a state where calling beginWrite crashes.
Firstly I apologize for not being able to provide a reproducible case for this. This started happening directly after upgrading to v0.98.1, so I'll try and roll back a bit to see if it fixes the issue.

The exception is as follows:

```
*** Terminating app due to uncaught exception 'RLMException', reason: 'Wrong transactional state (no active transaction, wrong type of transaction, or transaction already in progress)'
*** First throw call stack:
(
	0   CoreFoundation                      0x0000000105d43e65 __exceptionPreprocess + 165
	1   libobjc.A.dylib                     0x0000000107894deb objc_exception_throw + 48
	2   Realm                               0x0000000103ad10e8 -[RLMRealm beginWriteTransaction] + 136
	3   RealmSwift                          0x0000000103f83abd _TFC10RealmSwift5Realm10beginWritefS0_FT_T_ + 45
```

The Realm's inWriteTransaction is ```false```. The source of the exception is ```_realm->begin_transaction();```, so I'm not able to debug this further as I don't have the source for that.

tgoyne:
Set m_in_transaction before calling `promote_to_write()`, and set it back to false if that throws rather than setting it afterwards, so that notifications called due to the implicit `advance_read()` see the correct value.

Also includes some uninteresting tests that hit a few of the previously uncovered bits of `realm::Realm`.

May fix #3199.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,8 @@ x.x.x Release notes (yyyy-MM-dd)
 ### Bugfixes
 
 * Use of KVC collection operators on Swift collection types no longer throws an exception.
+* Fix reporting of inWriteTransaction in notifications triggered by
+  `beginWriteTransaction`.
 
 0.98.1 Release notes (2016-02-10)
 =============================================================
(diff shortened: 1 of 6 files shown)


bdash:
:+1:

PR #3203, opened 2016-02-11
Ensure that RLMFastEnumerator initializes our Swift generics machinery for the accessor objects it creates

### LINKED ISSUE #2905: Optional property values missing when copying between realms.
Hi!

I have some objects:

```swift
class Task: Object {
    dynamic var name: String? = nil
    let steps = List<Step>()
}

class Step: Object {
    dynamic var name: String? = nil
    let duration = RealmOptional<NSTimeInterval>()
}
```

When I copy a `Task` from a read-only compressed realm into the default realm, everything is copied over except the `duration` property which shows up as 0 in the Realm browser. I verified using the Realm browser that the `duration` property is not 0 in the read-only Realm.

Here's a test project (it's a .zip file, so you'll have to change the extension when you download it):
test-proj

My built frameworks were too big to upload, but I can find a way to upload them if needed.

Here's the output I get in my console when running the project:
```
Objects in main realm before deletion: Results<Task> (

)
Deleting all objects from main Realm.
Creating read-only Realm.
Read-only Realm saved to /Users/......./readonly.realm
Objects in the read-only realm: Results<Task> (
	[0] Task {
		name = taskName;
		steps = RLMArray <0x7fcd3b585300> (
			[0] Step {
				name = stepName;
				duration = 3.5;
			}
		);
	}
)
Inserted read-only objects into main Realm.
Objects in main realm after: Results<Task> (
	[0] Task {
		name = taskName;
		steps = RLMArray <0x7fcd3b726d90> (
			[0] Step {
				name = stepName;
				duration = (null);          <<<<< I think this should be 3.5
			}
		);
	}
)
Objects in read-only realm after: Results<Task> (
	[0] Task {
		name = taskName;
		steps = RLMArray <0x7fcd3b738740> (
			[0] Step {
				name = stepName;
				duration = 3.5;
			}
		);
	}
)
```

This is an almost perfect replica of my issue in my other project. The only difference is that my other project persists the default Realm across launches but this one does not seem to do that. 

Let me know what I'm doing wrong or if there's anything else I can provide!

I'm on XCode 7.1 (7B91b) / OS X 10.11.1 / Realm 0.96.2 / Carthage 0.9.2

bdash:
Fixes #2905.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,8 @@ x.x.x Release notes (yyyy-MM-dd)
 * Use of KVC collection operators on Swift collection types no longer throws an exception.
 * Fix reporting of inWriteTransaction in notifications triggered by
   `beginWriteTransaction`.
+* The contents of `List` and `Optional` properties are now correctly preserved when copying
+  a Swift object from one Realm to another.
 
 0.98.1 Release notes (2016-02-10)
 =============================================================
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,7 +15,8 @@ x.x.x Release notes (yyyy-MM-dd)
 * Fix reporting of inWriteTransaction in notifications triggered by
   `beginWriteTransaction`.
 * The contents of `List` and `Optional` properties are now correctly preserved when copying
-  a Swift object from one Realm to another.
+  a Swift object from one Realm to another, and performing other operations that result in a
+  Swift object graph being recursively traversed from Objective-C.
 
 0.98.1 Release notes (2016-02-10)
 =============================================================
--- a/RealmSwift-swift2.0/Tests/ObjectCreationTests.swift
+++ b/RealmSwift-swift2.0/Tests/ObjectCreationTests.swift
@@ -388,17 +388,23 @@ class ObjectCreationTests: TestCase {
             "arrayCol": [SwiftBoolObject(value: [true]), SwiftBoolObject()] as AnyObject,
         ]
 
-        realmWithTestPath().beginWrite()
-        let otherRealmObject = realmWithTestPath().create(SwiftListOfSwiftObject.self, value: ["array": [SwiftObject(value: values), SwiftObject(value: values)]])
-        try! realmWithTestPath().commitWrite()
+        let realmA = realmWithTestPath()
+        let realmB = try! Realm()
 
-        try! Realm().beginWrite()
-        let object = try! Realm().create(SwiftListOfSwiftObject.self, value: otherRealmObject)
-        try! Realm().commitWrite()
+        var realmAObject: SwiftListOfSwiftObject!
+        try! realmA.write {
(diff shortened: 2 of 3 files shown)


RealmSwift-swift2.0/Tests/ObjectCreationTests.swift
```diff
@@ -375,6 +375,35 @@ class ObjectCreationTests: TestCase {
...
+            "dateCol": NSDate(timeIntervalSince1970: 2) as NSDate,
+            "objectCol": SwiftBoolObject(value: [true]) as AnyObject,
+            "arrayCol": [SwiftBoolObject(value: [true]), SwiftBoolObject()] as AnyObject,
+        ]
+
+        realmWithTestPath().beginWrite()
+        let otherRealmObject = realmWithTestPath().create(SwiftListOfSwiftObject.self, value: ["array": [SwiftObject(value: values), SwiftObject(value: values)]])
+        try! realmWithTestPath().commitWrite()
+
+        try! Realm().beginWrite()
```
jpsim: `beingWrite()` doesn't throw

RealmSwift-swift2.0/Tests/ObjectCreationTests.swift
```diff
@@ -375,6 +375,35 @@ class ObjectCreationTests: TestCase {
...
+        let object = try! Realm().create(SwiftListOfSwiftObject.self, value: otherRealmObject)
+        try! Realm().commitWrite()
+
+        XCTAssertNotEqual(otherRealmObject, object)
+        XCTAssertEqual(object.array.count,2)
```
jpsim: fyi the lack of space after the comma will trigger a SwiftLint violation

CHANGELOG.md
```diff
@@ -12,6 +12,8 @@ x.x.x Release notes (yyyy-MM-dd)
 ### Bugfixes
 
 * Use of KVC collection operators on Swift collection types no longer throws an exception.
+* The contents of `List` and `Optional` properties are now correctly preserved when copying
```
jpsim: That content also wasn't being preserved when initializing an object with the `Object(value:)` initializer IIRC.

bdash: They're not preserved when doing anything that results in the objects graph being recursively traversed from Objective-C. I'm not sure what the complete list of operations that do that is, or whether it's beneficial to list them all here.

jpsim: True, but the entry kind of implies that copying between Realms is the only time this will happen at the moment, which is misleading. You could just add "and other operations resulting in the object graph being recursively traversed from Objective-C" at the end of the sentence. Although most users won't understand the full ramifications of that, if they're affected, they know what to look for.

bdash: Done.

jpsim:
It would have been nice to test the optional side of that too, but not strictly necessary.

:+1:

tgoyne:
:+1:

PR #3208, opened 2016-02-12
Make partial schema initialization more robust

### LINKED ISSUE #3200: Crash when having two realms with different schemes.
Once again, I apologize for not being able to provide an isolated reproducible project for this.

I have two distinct realms inside the app:

```
private lazy var realm: Realm = {
    var configuration = Realm.Configuration.defaultConfiguration
    configuration.objectTypes = [A1.self, A2.self, A3.self, A4.self, A5.self, A6.self, A7.self, A8.self, A9.self, A10.self, A11.self, A12.self]
    configuration.migrationBlock = {
        migration, oldSchemaVersion in
        if oldSchemaVersion < 1 {
            // migration stuff
        }
        if oldSchemaVersion < 2 {
            // migration stuff
        }
    }
    return try! Realm(configuration: configuration)
}()

// note that the in-memory realm contains ALL Realm subclasses inheriting from Object
private lazy var inMemoryRealm: Realm = {
    let configuration = Realm.Configuration(path: nil, inMemoryIdentifier: "someIdentifier", encryptionKey: nil, readOnly: false, schemaVersion: 0, migrationBlock: nil, objectTypes: nil)
    return try! Realm(configuration: configuration)
}()
```

Now the crucial part is the instantiation order. 

If you instantiate the in-memory Realm first ```void Schema::validate() const``` is called having 22 ```ObjectSchema```s. This number includes both the 12 classes that are marked to be persisted inside the regular Realm as well as the 10 others that are only used by the in-memory Realm. Then ```validate``` is called for the persistent Realm having 12 ```ObjectSchema``` classes as expected. There is no problem here and everything is working fine.

However, if you instantiate the persistent Realm first, then its schema is validated having 12 ```ObjectSchema``` classes as expected. But when the in-memory Realm is instantiated, its schema is validated but it contains *only* 10 ```ObjectSchema```s. Then an exception is thrown by ```exceptions.emplace_back(MissingObjectTypeException(object.name, prop));``` stating that:

```
fatal error: 'try!' expression unexpectedly raised an error: Error Domain=io.realm Code=0 "Migration is required due to the following errors: 
- Target type 'A6' doesn't exist for property 'user'." UserInfo={
    NSLocalizedDescription = "Migration is required due to the following errors: \n- Target type 'A6' doesn't exist for property 'user'.";
}: file /Library/Caches/com.apple.xbs/Sources/swiftlang/swiftlang-700.1.101.15/src/swift/stdlib/public/core/ErrorType.swift, line 50
```

tgoyne:
Make it so that creating a standalone object only initializing the entire shared schema if it needs to in order to validate link/array properties, and otherwise only initialize that class's schema.

Fix a case where triggering initialization of the shared schema during partial schema init would result in the classes used in the partial schema not appearing in the shared schema.

Fix creating standalone instances of subclasses when the parent class has been used in a partial schema and the shared schema has not been initialized.

Fix some probably innocuous data races.

Reduce the number of calls to class_replaceMethod() during schema init.

Fixes #3200.

### DIFF
--- a/Realm/RLMSchema.mm
+++ b/Realm/RLMSchema.mm
@@ -313,8 +313,10 @@ static void RLMRegisterClassLocalNames(Class *classes, NSUInteger count) {
 
 - (NSString *)description {
     NSMutableString *objectSchemaString = [NSMutableString string];
-    for (RLMObjectSchema *objectSchema in self.objectSchema) {
-        [objectSchemaString appendFormat:@"\t%@\n", [objectSchema.description stringByReplacingOccurrencesOfString:@"\n" withString:@"\n\t"]];
+    NSArray *sort = @[[NSSortDescriptor sortDescriptorWithKey:@"className" ascending:YES]];
+    for (RLMObjectSchema *objectSchema in [self.objectSchema sortedArrayUsingDescriptors:sort]) {
+        [objectSchemaString appendFormat:@"\t%@\n",
+         [objectSchema.description stringByReplacingOccurrencesOfString:@"\n" withString:@"\n\t"]];
     }
     return [NSString stringWithFormat:@"Schema {\n%@}", objectSchemaString];
 }
--- a/Realm/RLMSchema.mm
+++ b/Realm/RLMSchema.mm
@@ -67,7 +67,7 @@ static RLMObjectSchema *RLMRegisterClass(Class cls) {
     // set standalone class on shared shema for standalone object creation
     schema.standaloneClass = RLMStandaloneAccessorClassForObjectClass(schema.objectClass, schema);
 
-    // override sharedSchema classs methods for performance
+    // override sharedSchema class methods for performance
     RLMReplaceSharedSchemaMethod(cls, schema);
 
     s_privateObjectSubclasses[schema.className] = schema;
@@ -189,7 +189,7 @@ static void RLMRegisterClassLocalNames(Class *classes, NSUInteger count) {
 
 + (RLMObjectSchema *)sharedSchemaForClass:(Class)cls {
     @synchronized(s_localNameToClass) {
-        // We create instancse of Swift objects during schema init, and they
+        // We create instances of Swift objects during schema init, and they
         // obviously need to not also try to initialize the schema
         if (s_sharedSchemaState == SharedSchemaState::Initializing) {
             return nil;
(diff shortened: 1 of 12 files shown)


Realm/RLMSchema.mm
```diff
@@ -43,172 +43,207 @@ @interface RLMSchema ()
...
+    schema.standaloneClass = RLMStandaloneAccessorClassForObjectClass(schema.objectClass, schema);
 
-    RLMSchema *schema = [[self alloc] init];
-    NSMutableArray *schemas = [NSMutableArray arrayWithCapacity:count];
-    for (Class cls in classes) {
-        if (!RLMIsObjectSubclass(cls)) {
-            @throw RLMException(@"Can't add non-Object type '%@' to a schema.", cls);
-        }
-        [schemas addObject:[cls sharedSchema]];
+    // override sharedSchema classs methods for performance
```
bdash: s/classs/class/

Realm/RLMSchema.mm
```diff
@@ -43,172 +43,207 @@ @interface RLMSchema ()
...
-            RLMReplaceSharedSchemaMethod(cls, schema);
+    return schema;
+}
 
-            if ([cls shouldIncludeInDefaultSchema]) {
-                [schemas addObject:schema];
-            }
++ (RLMObjectSchema *)sharedSchemaForClass:(Class)cls {
+    @synchronized(s_localNameToClass) {
+        // We create instancse of Swift objects during schema init, and they
```
bdash: s/instancse/instances/

Realm/RLMSchema.mm
```diff
@@ -43,172 +43,207 @@ @interface RLMSchema ()
...
-        @throw RLMException(@"Illegal recursive call of +[%@ %@]. Note: Properties of Swift `Object` classes must not be prepopulated with queried results from a Realm.", self, NSStringFromSelector(_cmd));
-    }
-    static std::once_flag onceFlag;
-    std::call_once(onceFlag, [&](){
-        threadID = pthread_mach_thread_np(pthread_self());
-        RLMSchema *schema = [[RLMSchema alloc] init];
+    @synchronized(s_localNameToClass) {
+        if (s_sharedSchemaState == SharedSchemaState::Initialized) {
+            // Init was done on a different thread while we were waiting for
+            // the @synchronized
```
bdash: It took me a moment to figure out that this code is indeed specific to when we're waiting on the lock. It isn't apparent at this point in the function that the implementation of `+sharedSchema` is overridden when initialization completes, and therefore this code won't be called if initialization was already complete when `+sharedSchema` was called.

bdash:
:+1:

PR #3215, opened 2016-02-15
Avoid holding locks while calling transaction::advance()

### LINKED ISSUE #3201: Deadlock accessing realm objects from different threads
I'm using realm notification block to read updates from database on background thread. Having many NSThreads, access to realm objects deadlock.

This issue appeared first in 0.98.0, and reproducible in 0.98.1. Right now i'm using 0.97.1 which is working fine in my case.

[deadlock.txt](https://github.com/realm/realm-cocoa/files/126803/deadlock.txt)

tgoyne:
It can call user code due to change notifications, which leads to deadlocks if that code then tries to add async queries (and advancing is a potentially expensive operation, so doing it while holding a lock inhibits parallelism anyway).

Fixes #3201.

### DIFF
--- a/Realm/ObjectStore/impl/realm_coordinator.cpp
+++ b/Realm/ObjectStore/impl/realm_coordinator.cpp
@@ -407,34 +407,55 @@ void RealmCoordinator::advance_to_ready(Realm& realm)
     auto& sg = Realm::Internal::get_shared_group(realm);
     auto& history = Realm::Internal::get_history(realm);
 
-    {
-        std::lock_guard<std::mutex> lock(m_query_mutex);
-
-        SharedGroup::VersionID version;
+    auto get_query_version = [&] {
         for (auto& query : m_queries) {
-            version = query->version();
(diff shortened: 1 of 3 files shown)


Realm/ObjectStore/impl/realm_coordinator.cpp
```diff
@@ -407,34 +407,53 @@ void RealmCoordinator::advance_to_ready(Realm& realm)
...
+    }
+
+    while (true) {
+        // Advance to the ready version without holding any locks because it
+        // may end up calling user code (in did_change() notifications)
         transaction::advance(sg, history, realm.m_binding_context.get(), version);
 
+        // Reacquire the lock and recheck the query version, as the queries may
+        // have advanced to a later version while we didn't hold the lock. If
+        // so, we need to released the lock and re-advance
```
jpsim: release

jpsim:
:+1:

bdash:
:+1:

PR #3216, opened 2016-02-15
Allow aggregate operations to be used on key paths that start with a unary relationship

### LINKED ISSUE #2955: Support ANY modifier with nested keypaths
The predicate string `ANY presentation.speakers.name == 'John Doe'` should be valid for the following models:

```swift
class Event: Object {
    dynamic var presentation: Presentation?
}

class Presentation: Object {
    let speakers: List<Speaker>()
}

class Speaker: Object {
    dynamic var name = ""
}
```

To work around this issue, simply omit the `ANY` keyword: `presentation.speakers.name == 'John Doe'`.

Originally reported in http://stackoverflow.com/questions/34125883/why-im-getting-aggregate-operations-can-only-be-used-on-rlmarray-properties.

bdash:
The constraint is now that the key path must traverse a to-many relationship, but it is no longer required to be the first key in the path.

This allows for the following queries that were formerly rejected:

    ANY company.employees.age > 30
    company.employees.@avg.age > 30
    SUBQUERY(company.employees, $e, $e.age > 30).@count > 0

Fixes #2955.

### DIFF
--- a/Realm/RLMQueryUtil.mm
+++ b/Realm/RLMQueryUtil.mm
@@ -703,50 +703,46 @@ void add_constraint_to_query(realm::Query &query, RLMPropertyType type,
     }
 }
 
-ColumnReference column_reference_from_key_path(RLMSchema *schema, RLMObjectSchema *desc,
+ColumnReference column_reference_from_key_path(RLMSchema *schema, RLMObjectSchema *objectSchema,
                                                NSString *keyPath, bool isAggregate)
 {
+    RLMProperty *property;
     std::vector<size_t> indexes;
-    RLMProperty *prop = nil;
 
-    NSString *prevPath = nil;
+    bool keyPathContainsToManyRelationship = false;
+
     NSUInteger start = 0, length = keyPath.length, end = NSNotFound;
     do {
         end = [keyPath rangeOfString:@"." options:0 range:{start, length - start}].location;
-        NSString *path = [keyPath substringWithRange:{start, end == NSNotFound ? length - start : end - start}];
-        if (prop) {
-            RLMPrecondition(prop.type == RLMPropertyTypeObject || prop.type == RLMPropertyTypeArray,
-                            @"Invalid value", @"Property '%@' is not a link in object of type '%@'", prevPath, desc.className);
-            indexes.push_back(prop.column);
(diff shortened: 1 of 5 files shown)


jpsim:
I think this resolves #2955.

bdash:
It does!

Realm/RLMQueryUtil.mm
```diff
@@ -703,50 +703,46 @@ void add_constraint_to_query(realm::Query &query, RLMPropertyType type,
...
+            REALM_ASSERT(property.objectClassName);
+            objectSchema = schema[property.objectClassName];
         }
-        prevPath = path;
+
         start = end + 1;
     } while (end != NSNotFound);
 
-    return ColumnReference(prop, indexes);
+    if (isAggregate && !keyPathContainsToManyRelationship) {
```
jpsim: these two conditions are essentially the same, so they could be merged into `isAggregate != keyPathContainsToManyRelationship`.

jpsim:
LGTM :+1:

Realm/RLMQueryUtil.mm
bdash: They need to generate different error messages so I don't think this helps?

Realm/RLMQueryUtil.mm
```diff
@@ -703,50 +703,46 @@ void add_constraint_to_query(realm::Query &query, RLMPropertyType type,
...
+        @throw RLMPredicateException(@"Invalid predicate",
+                                     @"Aggregate operations can only be used on key paths that include an array property");
+    } else if (!isAggregate && keyPathContainsToManyRelationship) {
+        @throw RLMPredicateException(@"Invalid predicate",
+                                     @"Aggregate operations must be used on key paths that include an array property");
```
tgoyne: I think it'd be clearer to say "Key paths that include an array property must use aggregate operations". The current wording could reasonably be interpreted to mean the same thing as the error above, when it's actually the opposite case.

bdash: Sounds good.

jpsim: Yeah, definitely clearer.

tgoyne:
:+1:

PR #3219, opened 2016-02-15
Add support for predicates of the form "%@ IN collection"

### LINKED ISSUE #1550: Support querying "%@ IN (RLMArray keypath)"
**Summary**
Querying existence of objects in an RLMArray needs a specific query syntax.

**Steps**
Query an RLMArray of objects:
 - First with: (ANY types = %@)
 - Then with (%@ IN types)

**Expected**
(%@ IN types) should work as it makes grammatical sense, even if array contains one value.

**Actual**
Only (ANY types = %@) works, and the second option does not.
 Error message: Predicate with IN operator must compare a KeyPath with an aggregate with two values

**Workaround**
Had to figure this out manually, would be good to put this in the docs for querying to-many relationships.

bdash:
Rewrite `%@ IN collection` to the equivalent `ANY collection = %@` since that form is already supported.

Fixes #1550.


jpsim:
:tada: :+1:

PR #3220, opened 2016-02-16
Work around Xcode 7.2.1 failing to run tests in the iOS simulator for unknown reasons

bdash:
Resetting the simulator between the two xcodebuild invocations in the workaround for <rdar://problem/23055637> works around the new issue.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -171,6 +171,9 @@ xc_work_around_rdar_23055637() {
     # build (<http://openradar.appspot.com/23055637>). Work around this by having the test phases intentionally
     # exit after they finish building the first time, then run the tests for real.
     ( REALM_EXIT_AFTER_BUILDING_TESTS=YES xc "$1" ) || true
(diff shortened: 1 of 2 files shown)


bdash:
It looks like this somehow causes problems for the tvOS build on Travis. After running reset-simulators.sh it can no longer find the tvOS simulator. Huh. Jenkins looks fine. Perhaps this is due to which Xcode version is the default in the different environments?

bdash:
I've not been able to reproduce the Travis tvOS issue locally :disappointed:

bdash:
Ok, so what appears to be happening is that reset-simulators.sh thinks that the tvOS 9.1 simulator is a duplicate of the tvOS 9.0 simulator, and so it deletes it. reset-simulators.sh needs to be smarter about this.

bdash:
This is ready for review now!

/cc @jpsim @tgoyne

tgoyne:
:+1:

jpsim:
:+1: :100:

PR #3221, opened 2016-02-16
Split SchemaValidationException into two of SchemaValidationException and SchemaMismatchException

kishikawakatsumi:
Because SchemaValidationException is thrown both cases that a schema definition is incorrect and that two schema definitions are mismatched.
In the former case, the migration does not solve the problem. But the exception message shows "Migration is required..."

*e.g.* An exception that says "migration is retuired" when define the following model class
```
@interface TestObject : RLMObject
@property NSData *data;
@end

@implementation TestObject
+ (NSArray<NSString *> *)indexedProperties {
    return @[@"data"];
}
@end

*** Terminating app due to uncaught exception 'RLMException', reason: 'Migration is required due to the following errors: 
- Can't index property TestObject.data: indexing a property of type 'data' is currently not supported'
```

Therefore the latter as SchemaMismatchException, to distinguish between the two cases. 
Also, the message of SchemaValidationException is changed to "Schema validation is failed...".

CC @tgoyne @bdash

### DIFF
--- a/Realm/ObjectStore/object_store.cpp
+++ b/Realm/ObjectStore/object_store.cpp
@@ -168,7 +168,7 @@ void ObjectStore::verify_schema(Schema const& actual_schema, Schema& target_sche
         errors.insert(errors.end(), more_errors.begin(), more_errors.end());
     }
     if (errors.size()) {
-        throw SchemaValidationException(errors);
+        throw SchemaMismatchException(errors);
     }
 }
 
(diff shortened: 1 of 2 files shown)


Realm/ObjectStore/object_store.cpp
```diff
@@ -513,10 +513,18 @@ DuplicatePrimaryKeyValueException::DuplicatePrimaryKeyValueException(std::string
     m_what = "Primary key property '" + property.name + "' has duplicate values after migration.";
 }
 
-
 SchemaValidationException::SchemaValidationException(std::vector<ObjectSchemaValidationException> const& errors) :
     m_validation_errors(errors)
 {
+    m_what ="Schema validation is failed due to the following errors: ";
```
tgoyne: Missing a space after the `=`, and drop the "is" in the message.

kishikawakatsumi:
Thanks! Feedback addressed @tgoyne

jpsim:
Looks fine to me :+1:

jpsim:
Is this blocked on anything @kishikawakatsumi?

kishikawakatsumi:
No, I thought it has still one `+1`. I'll merge it.

PR #3222, opened 2016-02-16

kishikawakatsumi:
Closes https://github.com/realm/realm-cocoa/issues/1692
This PR depends on https://github.com/realm/realm-cocoa/pull/3221

Add `deleteRealmIfMigrationNeeded` configuration option. If the value of the property is `YES`, the Realm file will be deleted then re-create with a new schema when a migration is required.

If you increase schema version and set this flag to `YES`, migration will be executed (the file will not be deleted).

@tgoyne @jpsim


Realm/ObjectStore/shared_realm.cpp
```diff
@@ -42,6 +42,7 @@ Realm::Config::Config(const Config& c)
 , encryption_key(c.encryption_key)
 , schema_version(c.schema_version)
 , migration_function(c.migration_function)
+, delete_realm_if_maigration_needed(c.delete_realm_if_maigration_needed)
```
jpsim: `maigration` should be `migration`

Realm/RLMRealmConfiguration.h
```diff
@@ -71,6 +71,9 @@ RLM_ASSUME_NONNULL_BEGIN
 /// The block which migrates the Realm to the current version.
 @property (nonatomic, copy, nullable) RLMMigrationBlock migrationBlock;
 
+/// Delete the Realm file if migration is needed and re-create with the new schema.
+@property (nonatomic, getter = shouldDeleteRealmIfMigrationNeeded) BOOL deleteRealmIfMigrationNeeded;
```
jpsim: the custom getter here seems superfluous. `deleteRealmIfMigrationNeeded` reads like a fine getter name to me.

Realm/Tests/MigrationTests.mm
```diff
@@ -200,6 +200,22 @@ - (void)assertNoMigrationRequiredForChangeFrom:(NSArray *)from to:(NSArray *)to
     RLMAssertRealmSchemaMatchesTable(self, [RLMRealm realmWithConfiguration:config error:nil]);
 }
 
+- (void)assertDeletingRealmFilesForChangeFrom:(NSArray *)from to:(NSArray *)to {
```
jpsim: since this is just used from one caller (`testDeleteRealmIfMigrationNeeded`) it would seem preferable to keep the logic there rather than have this helper method?

RealmSwift-swift2.0/RealmConfiguration.swift
```diff
@@ -58,6 +58,8 @@ extension Realm {
         - parameter readOnly:           Whether the Realm is read-only (must be true for read-only files).
         - parameter schemaVersion:      The current schema version.
         - parameter migrationBlock:     The block which migrates the Realm to the current version.
+        - parameter deleteRealmIfMigrationNeeded: If the value of this property is YES, the Realm will be cleard
```
jpsim:
```
If `true`, the Realm file will be deleted and recreated when a migration is required.
```

Realm/Tests/MigrationTests.mm
kishikawakatsumi: Oh yes, I was going to add two or three more tests. For now, I kept the logic there.

Realm/ObjectStore/shared_realm.cpp
```diff
@@ -127,33 +128,8 @@ void Realm::init(std::shared_ptr<RealmCoordinator> coordinator)
...
         // otherwise get the schema from the group
-        auto target_schema = std::move(m_config.schema);
-        auto target_schema_version = m_config.schema_version;
         m_config.schema_version = ObjectStore::get_schema_version(read_group());
         m_config.schema = std::make_unique<Schema>(ObjectStore::schema_from_group(read_group()));
-
-        // if a target schema is supplied, verify that it matches or migrate to
-        // it, as neeeded
-        if (target_schema) {
-            if (m_config.read_only) {
```
tgoyne: This logic does all need to exist, and it's somewhat concerning that apparently there are no tests for it.

tgoyne:
Deleting the file when a migration is needed should be done within the objectstore code rather than the binding-specific part.

Realm/RLMRealm.mm
```diff
@@ -373,9 +373,24 @@ + (instancetype)realmWithConfiguration:(RLMRealmConfiguration *)configuration er
...
                         realm->_realm->update_schema(schema.objectStoreCopy, config.schema_version);
                     }
+                } else {
+                    realm->_realm->update_schema(schema.objectStoreCopy, config.schema_version);
                 }
 
                 RLMRealmSetSchemaAndAlign(realm, schema);
+            } catch (SchemaMismatchException::SchemaMismatchException const& exception) {
+                if (configuration.deleteRealmIfMigrationNeeded) {
+                    NSError *deletiionError = nil;
```
uny: I would like this option (`deleteRealmIfMigrationNeeded`)!
I found it seemed like a typo to me.

kishikawakatsumi:
@tgoyne @jpsim Clear the data on migration needed instead deleting the file. It was done within objectstore.
How does that look?

Realm/ObjectStore/shared_realm.cpp
```diff
@@ -194,7 +195,28 @@ void Realm::update_schema(std::unique_ptr<Schema> schema, uint64_t version)
...
         }
 
-        ObjectStore::verify_schema(*m_config.schema, *schema, m_config.read_only);
+        auto errors = ObjectStore::verify_schema(*m_config.schema, *schema, m_config.read_only);
+        if (errors.size()) {
+            if (m_config.delete_realm_if_migration_needed) {
+                begin_transaction();
+                try {
+                    for (auto &object_schema : *m_config.schema) {
+                        ObjectStore::delete_data_for_object(m_group, object_schema.name);
```
jpsim: This feature should delete the file, not delete all rows for all tables within a write transaction.

kishikawakatsumi:
@tgoyne @jpsim I implemented delete/re-create file in `ObjectStore::verify_schema()`. It works but dirty implementation. Could you please review it?

kishikawakatsumi:
@jpsim @tgoyne 

There are two places where schema validation will fail.
One is here: https://github.com/realm/realm-cocoa/blob/master/Realm/RLMRealm.mm#L350

It is in the `Realm::init` method that is called from `openSharedRealm` method.

Another is here: https://github.com/realm/realm-cocoa/blob/master/Realm/RLMRealm.mm#L374

The former is called when a schema is specified with `objectClasses`,  the latter is called when that the schema isn't specified.

The problem is, the latter is called directly from the Objective-C layer, so  it is difficult to handle deleting and re-creating the file. It is also difficult to replace the Realm instances within the method.

Also it is difficult to determine whether the caller is former or latter. 

Do you have any thoughts?

kishikawakatsumi:
@tgoyne @bdash 

I need your advice. What do you think I should do?

It seems difficult for me that deleting Realm file when migration is needed is done within the ObjectStore code.

Schema validations are done in two places. 
The former is done during the initialization of the Realm instance. It is difficult to re-create Realm instance in this timing since the initialization method has an exclusive lock.

The latter is done much after the Realm was instantiated. So it needs to return the method to the Objective-C++ code to release and re-create Realm instance.

kishikawakatsumi:
Talked with @jpsim, we adopt https://github.com/realm/realm-cocoa/pull/3463 approach, for less affected existing code.

PR #3227, opened 2016-02-16
[Podspec] Fix issue with latest CocoaPods 1.0 betas

### LINKED ISSUE #3052: CocoaPods 1.0.0.beta.X can't integrate Realm as a dynamic framework

mrackwitz:
Attempts to fix #3052.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,8 @@ x.x.x Release notes (yyyy-MM-dd)
   that begin with an object relationship so long as there is a `RLMArray`/`List` property at some
   point in a key path.
 * Predicates of the form `%@ IN arrayProperty` are now supported.
+* Support CocoaPods 1.0 beginning from prerelease 1.0.0.beta.4 while retaining
+  backwards-compatibility to 0.39.
 
 ### Bugfixes
 
--- a/Realm.podspec
+++ b/Realm.podspec
@@ -16,34 +16,34 @@ Pod::Spec.new do |s|
   s.documentation_url       = "https://realm.io/docs/objc/#{s.version}"
   s.license                 = { :type => 'Apache 2.0', :file => 'LICENSE' }
 
-  public_header_files       = 'include/RLMArray.h',
-                              'include/RLMCollection.h',
-                              'include/RLMConstants.h',
-                              'include/RLMDefines.h',
-                              'include/RLMListBase.h',
-                              'include/RLMMigration.h',
-                              'include/RLMObject.h',
-                              'include/RLMObjectBase.h',
-                              'include/RLMObjectSchema.h',
-                              'include/RLMOptionalBase.h',
-                              'include/RLMPlatform.h',
-                              'include/RLMProperty.h',
-                              'include/RLMRealm.h',
-                              'include/RLMRealmConfiguration.h',
-                              'include/RLMResults.h',
-                              'include/RLMSchema.h',
-                              'include/Realm.h',
+  public_header_files       = 'include/**/RLMArray.h',
+                              'include/**/RLMCollection.h',
+                              'include/**/RLMConstants.h',
+                              'include/**/RLMDefines.h',
+                              'include/**/RLMListBase.h',
+                              'include/**/RLMMigration.h',
+                              'include/**/RLMObject.h',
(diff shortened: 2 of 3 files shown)


mrackwitz:
I could reproduce the CI failure locally. It works with CocoaPods 1.0, but it breaks backwards-compatibility to 0.39.0. The podspec itself still lints on 0.39.0, but the Clang module can't be included. It fails because of the inclusion of "non-modular headers". This supposedly happens because the headers in the sandbox, where CocoaPods links headers according to the `header_mapping_dir` overlap now with the framework headers. A not so recent Xcode version started to reject such configurations, from my understanding because framework module headers may only be included and found in the framework's header directory itself and not in any other search path. CocoaPods 1.0 ships a fix for exactly this problem, which broke our podspec in the first place.

One possible workaround to continue working with an older CocoaPods version is to delete the `Pods/Headers/Private/Realm` directory from a post installation Podfile hook.

mrackwitz:
@tgoyne: As you made some modifications recently to our podspec to address this as well, you may have some thoughts to this?

tgoyne:
I'm not a fan of requiring that people add extra stuff to their Podfile since there's no reasonable way to tell them that they need to do so. I'd sort of prefer to just outright drop support for 0.39 and require 1.0 over that.

With 1.0 I think we can skip coping the public headers to the include directory, since that was just a convoluted way of getting them out of the header search path for things using the dynamic framework.

mrackwitz:
>I'm not a fan of requiring that people add extra stuff to their Podfile since there's no reasonable way to tell them that they need to do so. I'd sort of prefer to just outright drop support for 0.39 and require 1.0 over that.

I'd prefer that as well and would be fine with just setting the minimum required version in the podspec to 1.0 once the final release is out. So the workaround was more thought as a reference, what users could do in the mean time, once we adopt these changes in our official podspec.

>With 1.0 I think we can skip coping the public headers to the include directory, since that was just a convoluted way of getting them out of the header search path for things using the dynamic framework.

Good to know, I wasn't aware that this was the original motivation. I thought that it was done so that the binding library headers work alongside the core headers with the way how CocoaPods places files. If there is now no strong reason to keep it like that, we should definitively get rid of it.

mrackwitz:
So, with https://github.com/CocoaPods/CocoaPods/pull/4933 shipped in the latest prerelease, I could implement the solution as discussed in our meeting, which I could successfully test with CocoaPods 0.39 and 1.0.0.beta.4.

mrackwitz:
The CI failure was because of job `ios-device-objc`, which seems unrelated.

jpsim:
What happens when `pod install` is first run with `0.39.0` and then again with `1.0.0.beta.4`? I imagine that'd be a fairly common scenario, so it'd be nice to support it.

If this change handles that scenario, I'm all for this. We should really have a changelog entry for this, since many users have been trying to use CocoaPods 1.0 with Realm already.

mrackwitz:
>What happens when `pod install` is first run with `0.39.0` and then again with `1.0.0.beta.4`?

CocoaPods would detect that a newer version runs and automatically nukes the cache, loads a fresh copy of Realm and integrates just like it would be installed the first time with 1.0.

mrackwitz:
Would appreciate a :+1: on the added CHANGELOG entry before merging.

CHANGELOG.md
```diff
@@ -11,6 +11,8 @@ x.x.x Release notes (yyyy-MM-dd)
   that begin with an object relationship so long as there is a `RLMArray`/`List` property at some
   point in a key path.
 * Predicates of the form `%@ IN arrayProperty` are now supported.
+* Support CocoaPods 1.0 beginning from prerelease 1.0.0.beta.4 while retaining
+  backwards-compatiblity to 0.39.
```
tgoyne: compatibility

tgoyne:
:+1:

PR #3228, opened 2016-02-16
fix inverted old/new versions

### LINKED ISSUE #3113: Incorrect message: Provided schema version 4 is less than last set version 3.
The message 

```cpp
m_what = "Provided schema version " + std::to_string(old_version) + " is less than last set version " + std::to_string(new_version) + ".";
```

is not always correct, because it can happen, e.g. when switching branches, that the Realm migration version is already higher than the last in code, where the message displays as follows:

```
Provided schema version 4 is less than last set version 3.
```

My CPP skill are absent so I didn't try to fix it myself, but I guess a simple comparison should be enough:

```
joiner = old_version > new_version ? "greater" : "less"
```

https://github.com/realm/realm-cocoa/blob/facde33440f2d46ece3e32ca60ff58f2a9028080/Realm/ObjectStore/object_store.cpp#L504

jpsim:
in InvalidSchemaVersionException. Fixes #3113. /cc @tgoyne

### DIFF
--- a/Realm/ObjectStore/object_store.cpp
+++ b/Realm/ObjectStore/object_store.cpp
@@ -504,7 +504,7 @@ bool ObjectStore::is_empty(const Group *group) {
 InvalidSchemaVersionException::InvalidSchemaVersionException(uint64_t old_version, uint64_t new_version) :
     m_old_version(old_version), m_new_version(new_version)
 {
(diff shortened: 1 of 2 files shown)


bdash:
Can we test this?

jpsim:
test adapted to match expected exception message

bdash:
:+1:

mrackwitz:
Shouldn't this get a changelog entry?

jpsim:
A changelog entry doesn't seem necessary here.

PR #3234, opened 2016-02-17
[ISSUE_TEMPLATE] Initial proposal

mrackwitz:
As seen in the [recent blog post](https://github.com/blog/2111-issue-and-pull-request-templates) on the GitHub blog, it's possible to create templates for issues and pull requests from now on. This PR should get the ball rolling for us and give a starting point for discussion.

Keep in mind that the template should be optimized in terms of readability as non-interpreted markdown as well.


bdash:
If we're going to do this I'd suggest putting the file in a `.github` subdirectory to avoid cluttering the top level of the repository.

ISSUE_TEMPLATE.md
```diff
@@ -0,0 +1,36 @@
...
+> ℹ Please replace this with what you tried to achieve.
+
+### Expected Results
+
+> ℹ Please replace this with what you expected to happen.
+
+### Actual Results
+
+> ℹ Please replace this with what actually happened.  
+> e.g. the stack trace of a crash
```
AndyDentFree: In the interests of readability for us, could we please encourage people to put things like _stack traces_ at the **end** of the issue? Maybe have a final section called **Supporting Details** and have the advice here to be something like _Please provide a short description here and append lengthy information such as stack traces at the end of the issue_

mrackwitz: I see your point, but I think it would be more important that stack traces are formatted as source code. As long as that is given, I think it's still relatively easy to keep an overview, especially given that the headings visually stand out.

mrackwitz:
>If we're going to do this I'd suggest putting the file in a .github subdirectory to avoid cluttering the top level of the repository.

:+1: While I would leave `CONTRIBUTING.md` located in the root of the repo, where it is currently.

ISSUE_TEMPLATE.md
```diff
@@ -0,0 +1,36 @@
+> ℹ Please fill out this template when filing an issue.
```
jpsim: what's with all these ℹs?

mrackwitz: Just an idea to make clear, that this is an informational message. As our repository has also a lot of traffic from beginners, who didn't used GitHub before and might not know about Markdown and how to handle that, I thought it might help to convey that message. Furthermore it could become a helpful marker for automatically processing issues.

mrackwitz:
Addressed @bdash's feedback.

.github/ISSUE_TEMPLATE.md
```diff
@@ -0,0 +1,36 @@
+> ℹ Please fill out this template when filing an issue.
+> All lines beginning with a quotation mark (`>`),
```
jpsim: Not a quotation mark

```md
> All lines marked as quotes (`>`)
```

no comma here either

.github/ISSUE_TEMPLATE.md
```diff
@@ -0,0 +1,36 @@
...
+> ℹ Please replace this with what you expected to happen.
+
+### Actual Results
+
+> ℹ Please replace this with what actually happened.  
+> e.g. the stack trace of a crash
+
+### Steps to Reproduce
+
+> ℹ Please provide steps necessary to take to reproduce this issue.
```
jpsim:
```md
> Please provide steps we can follow to reproduce this issue.
```

.github/ISSUE_TEMPLATE.md
```diff
@@ -0,0 +1,36 @@
...
+
+### Code Sample
+
+> ℹ Please provide a code sample that highlights the issue.
+> If necessary, please don't forget to provide at least an
```
jpsim:
```md
> If relevant, please include your model definitions.
> For larger code samples, links to external gists or repositories are preferred.
> Full Xcode projects that we can compile ourselves are ideal.
```

.github/ISSUE_TEMPLATE.md
```diff
@@ -0,0 +1,36 @@
...
+> excerpt from your model classes.
+> Full Xcode projects that we can compile ourselves are ideal.
+
+### Version of Realm and Tooling
+
+> ℹ Please share the versions of the used Realm frameworks, the Xcode
```
jpsim:
```md
> Please share which versions of Realm are affected, the Xcode
```

.github/ISSUE_TEMPLATE.md
```diff
@@ -0,0 +1,36 @@
...
+> and OS X version on which you compile, as well as the version of the
+> involved dependency manager. (CocoaPods / Cartage)
```
jpsim:
```md
> dependency manager involved (CocoaPods / Carthage).
```

jpsim:
Thanks for doing this, @mrackwitz! The sections involved here and information being requested all seem appropriate to me.

I'd prefer moving `CONTRIBUTING.md` into `.github/` to further remove some clutter in the top-level of the repo, but several links will have to be updated in the process.

I'd also prefer removing the ℹs and all of the "Please"s since it starts to be so redundant that it hurts readability. Saying please once at the start should be sufficient IMO.

mrackwitz:
>I'd prefer moving CONTRIBUTING.md into .github/ to further remove some clutter in the top-level of the repo, but several links will have to be updated in the process.

I think that would also drastically reduce the chances, that this document is read at all. Imho in comparison to all other files which are placed there, it's one of the most important because it's intended for humans. But just my 2cts. In addition, we should probably find some overall consensus across all our repos regarding that point, so that these files are always located at a predictable location.

>I'd also prefer removing the ℹs and all of the "Please"s since it starts to be so redundant that it hurts readability. Saying please once at the start should be sufficient IMO.

Good point. :+1:

jpsim:
> > I'd prefer moving `CONTRIBUTING.md` into `.github/`
>
> I think that would also drastically reduce the chances, that this document is read at all.

I really don't think many people read it when browsing the repo source, but rather by following the link when filing a new issue/PR, but I'm only speculating since we don't have stats on that.

However, this is the last topic we should be debating at this point, so I don't strongly oppose keeping it at the top-level.

> In addition, we should probably find some overall consensus across all our repos regarding that point, so that these files are always located at a predictable location.

Agreed. I propose we get this in cocoa first, and then bring it up for consideration for other bindings, pointing to the review history in this thread for context.

PR #3235, opened 2016-02-17
Improve shared schema initialization performance

### LINKED ISSUE #3230: Shared schema initialization performance is subpar with large schemas on some versions of iOS
On iOS 9.2, and likely other versions of iOS, the time spent initializing the shared schema in an application with many model classes (~100) is dominated by the Objective-C runtime repeatedly flushing caches when we add methods to the generated accessor classes. In the current iOS 9.3 beta this cache flushing appears to be significantly faster, resulting in shared schema initialization being an order of magnitude faster. Analysis suggests that some of the methods we add to the generated accessor classes are unnecessary. Eliminating this work should dramatically speed up shared schema initialization on iOS 9.2, and will likely be beneficial even after iOS 9.3 is released to users.

bdash:
On iOS 9.2, the time spent below `+[RLMSchema sharedSchema]` on the first call is dominated by the time that the Objective-C runtime spends flushing method caches as a result of calls to `class_addMethod` and `class_replaceMethod`. By eliminating several of these calls per accessor class we can take a significant chunk out of the cost of schema initialization. We do this by:

* No longer overriding `+shouldIncludeInDefaultSchema:` on generated accessor classes. It's no longer necessary as marking the class as generated has the same effect of excluding the class from the default schema.
* No longer overriding `+className` on accessor classes. `+[RLMObjectBase className]` now uses a map from `Class` to class name as an optimization rather than per-class overrides of the method. This is around 50% slower per invocation of `+className` than the previous approach, but this overhead should be dominated by any work performed by the caller.

On a user-supplied application this sequence of changes reduces the time spent below `+[RLMSchema sharedSchema]` during launch by 2/3 (on an iPad Air running iOS 9.2.1).

iOS 9.3 appears to have optimized the method cache flushing performed by `class_addMethod` to be dramatically faster. Once that version is the predominant version of iOS for users of Realm we may want to reconsider the approach to `+className` as the tradeoffs involved will be different.

Fixes #3230.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@ x.x.x Release notes (yyyy-MM-dd)
 
 ### Enhancements
 
-* None.
+* Initializing the shared schema is 3x faster.
 
 ### Bugfixes
 
--- a/Realm/RLMAccessor.mm
+++ b/Realm/RLMAccessor.mm
@@ -760,8 +760,6 @@ static Class RLMCreateAccessorClass(Class objectClass,
         }
     }
 
-    // implement className for accessor to return base className
-    RLMReplaceClassNameMethod(accClass, schema.className);
     RLMMarkClassAsGenerated(accClass);
 
     return accClass;
--- a/Realm/RLMAccessor.mm
+++ b/Realm/RLMAccessor.mm
@@ -689,14 +689,14 @@ static RLMAccessorCode accessorCodeForType(char objcTypeCode, RLMPropertyType rl
 // implement the class method className on accessors to return the className of the
 // base object
 void RLMReplaceClassNameMethod(Class accessorClass, NSString *className) {
-    Class metaClass = objc_getMetaClass(class_getName(accessorClass));
+    Class metaClass = object_getClass(accessorClass);
     IMP imp = imp_implementationWithBlock(^(Class){ return className; });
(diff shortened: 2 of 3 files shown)


tgoyne:
[Spinlocks are forbidden on iOS](https://lists.swift.org/pipermail/swift-dev/Week-of-Mon-20151214/000372.html)  because the QoS stuff can make them livelock.

How bad is the init performance if we only override className on the base class rather than on the base class and both generated accessor classes? The latter ones seem like they should be redundant, and I would expect cutting out two-thirds of them to give two thirds of the benefit...

bdash:
I'll gather some measurements when overriding `+className` on the base class only to confirm your intuition.

bdash:
Removing the overrides of `+className` from only the generated accessor classes appears to give a very similar performance improvement as removing all of the overrides (55% reduction in time spent below `+[RLMSchema sharedSchema]` rather than the 66% reduction with the other approach).

tgoyne:
:+1: 

Just needs a changelog entry.

jpsim:
Nice! :+1:

Changsung:
Thanks for the fix! I already propose the idea to improve the launch: https://groups.google.com/forum/#!topic/realm-cocoa/Am71xzMGh-Q.

PR #3250, opened 2016-02-22
Update to core 0.96.2

tgoyne:

Issue #3122, opened 2016-01-20
realm::LogicError crash


chrism1974:
I'm receiving the following crash "sometimes".  Using 0.97.0

Fatal Exception: realm::LogicError: Wrong transactional state (no active transaction, wrong type of transaction, or transaction already in progress)
```
Thread : Crashed: com.twitter.crashlytics.ios.exception
0  PipeTracker                    0x70d1d7 CLSProcessRecordAllThreads + 170146
1  PipeTracker                    0x70d1d7 CLSProcessRecordAllThreads + 170146
2  PipeTracker                    0x70d3e5 CLSProcessRecordAllThreads + 170672
3  PipeTracker                    0x701f5b CLSHandler + 124454
4  PipeTracker                    0x70c059 __CLSExceptionRecord_block_invoke + 165668
5  libdispatch.dylib              0x236bbdc3 _dispatch_client_callout + 22
6  libdispatch.dylib              0x236c5055 _dispatch_barrier_sync_f_invoke + 56
7  PipeTracker                    0x70bb5f CLSExceptionRecord + 164394
8  PipeTracker                    0x70b5a3 CLSTerminateHandler() + 162926
9  libc++abi.dylib                0x232dce17 std::__terminate(void (*)()) + 78
10 libc++abi.dylib                0x232dcf3b std::terminate() + 190
11 PipeTracker                    0x56391d realm::ArrayIntNull::~ArrayIntNull() (array_integer.hpp:297)
12 PipeTracker                    0x6628a3 realm::SharedGroup::end_read() + 7098
13 PipeTracker                    0x5a6eb5 realm::Realm::invalidate() (shared_realm.cpp:333)
14 PipeTracker                    0x59468f -[RLMRealm invalidate] (RLMRealm.mm:503)
15 PipeTracker                    0x424cc5 -[PendingDataOperation checkForPendingTransfersInContext:] (PendingDataOperation.m:405)
16 PipeTracker                    0x423fd7 -[PendingDataOperation start] (PendingDataOperation.m:147)
17 Foundation                     0x243247ed __NSOQSchedule_f + 192
18 libdispatch.dylib              0x236c5f97 _dispatch_queue_drain + 1762
19 libdispatch.dylib              0x236bef2f _dispatch_queue_invoke + 282
20 libdispatch.dylib              0x236c7325 _dispatch_root_queue_drain + 400
21 libdispatch.dylib              0x236c7193 _dispatch_worker_thread3 + 94
[...]

PR #3251, opened 2016-02-22
Remove Realm::m_in_transaction and use the value from the SharedGroup

tgoyne:
Depends on #3250.

No changelog entry because while it should make the specific symptoms seen in #3122 impossible, it's not obvious that it actually fixes any bugs.


bdash:
:+1:

jpsim:
:+1:

PR #3259, opened 2016-02-24
Fix core library name in the podspec

tgoyne:
0.96.2 changed the name of the core library for OS X, and while there's a symlink from the old to the new name, CocoaPods deletes the actual library and leaves just the symlink, which isn't very useful.


jpsim:
:+1:

PR #3260, opened 2016-02-24
Update the example projects to use a workspace for doing in-tree builds

bdash:
Using a workspace ensures that Realm and RealmSwift will be rebuilt when their source changes, while also building only the active architecture of the frameworks. It also makes for easier testing of changes against the example projects as the Realm framework source is directly accessible.


bdash:
I tested the packaging changes locally with `build.sh test-package-release` with positive results.

jpsim:
that's... considerably nicer! Looks like it speeds up the CI jobs too since fewer architectures are built :+1: :100:

bdash:
I'm not sure if it'll speed things up or not. With the existing approach, the CI jobs should already have the frameworks built so they shouldn't need to build anything as part of the examples. After this change the examples will have to build the Realm frameworks for a single architecture since the workspace uses a different build directory than `build.sh`. I couldn't find a good way to avoid that so I punted on it for now.

jpsim:
Ah, right that's a good point. Still a very clean win elsewhere.

Issue #3258, opened 2016-02-23
CocoaPod build.sh fails when TMPDIR isn't setup


semireg:
### Goals

Tried to run "pod install" on Xcode bot. Cocoapod install fails when running Realm's build.sh because TMPDIR env variable is not set.

### Expected Results

Should run without error.

### Actual Results

Installing Realm (0.98.2)
[!] /bin/bash -c 
set -e
sh build.sh cocoapods-setup

core is not a symlink. Deleting...
Downloading dependency: core 0.96.1
mkdir: /core_bin: Permission denied

### Steps to Reproduce

Unset TMPDIR and run build.sh.

### Code Sample

https://github.com/realm/realm-cocoa/blob/master/build.sh#L284

Running "export TMPDIR=/tmp" on a shell allows "pod install" to complete. However, a solution inside of build.sh would be preferred. Alternatively, it would be nice to know how to get launchd (or launchctl, or whatever is the ghost in the machine) to setup _xcsbuildd's environment properly.

We need a check here so it's not trying to mkdir at the root directory.
### Version of Realm and Tooling

El Cap 10.11.3 w/ Xcode Server bot running as user _xcsbuildd.

mrackwitz:
I think the right solution would be to add `TMPDIR` as an environment variable to the `/Library/Server/Xcode/Config/xcsbuildd.plist`, which is responsible for launching the job via `launchd`.
```diff
<?xml version="1.0" encoding="UTF-8"?>
<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
<plist version="1.0">
<dict>
+    <key>EnvironmentVariables</key>
+    <dict>
+           <key>TMPDIR</key>
+           <string>/tmp</string>
+    </dict>
```

Could you give that a try?

(Note: `launchd` is the daemon, who actually launches the jobs, while `launchctl` is a control utility allowing to interface launchd.)

bdash:
We should probably just have build.sh handle the case of missing `TMPDIR` environment variable and use `$(getconf DARWIN_USER_TEMP_DIR)` in that case.

PR #3268, opened 2016-02-25
use `getconf DARWIN_USER_TEMP_DIR` rather than `$TMPDIR`

jpsim:
$TMPDIR isn't always available. Addresses #3258.

I don't think this should cause any issues with previous uses of `$TMPDIR` since it's the same value as `DARWIN_USER_TEMP_DIR` when set (in my limited testing).

```
$ echo $TMPDIR
/var/folders/np/5y4sgt8x3r13sk7x0n31z_3m0000gn/T/
$ getconf DARWIN_USER_TEMP_DIR
/var/folders/np/5y4sgt8x3r13sk7x0n31z_3m0000gn/T/
```

/cc @bdash @mrackwitz

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -281,7 +281,7 @@ fi
 
 download_core() {
     echo "Downloading dependency: core ${REALM_CORE_VERSION}"
-    TMP_DIR="$TMPDIR/core_bin"
+    TMP_DIR="$(getconf DARWIN_USER_TEMP_DIR)/core_bin"
     mkdir -p "${TMP_DIR}"
     CORE_TMP_TAR="${TMP_DIR}/core-${REALM_CORE_VERSION}.tar.bz2.tmp"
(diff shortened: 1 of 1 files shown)


tgoyne:
:+1:

mrackwitz:
:+1: But shouldn't we just initialize `TMPDIR` with `getconf DARWIN_USER_TEMP_DIR`, if it is not set?

jpsim:
Not sure what that would net us

bdash:
It'd respect a custom value for `TMPDIR` if the user has set one for some reason, as well as ensuring that any subprocesses that care about `TMPDIR` have a valid value. I don't know if any of that is super important though.

mrackwitz:
What @bdash wrote, DRY and better readability / self-documentation.

PR #3288, opened 2016-03-03
Simplify AnyRealmCollection

jpsim:
Based on @gwengrid's talk at the *try! Swift* conference about type erasure, I realized that her implementation might have a few advantages over the approach we've previously taken.

The last commit is a breaking change and I'm not convinced it's exactly what we should do yet, so that should be up for discussion.

/cc @tgoyne @bdash


tgoyne:
I don't see how this is an improvement. It's the same amount of boilerplate (the reduction in line count comes entirely from deleting the pointless extra copy of the docs on the private subclass) and it adds a bunch of runtime overhead.

jpsim:
Ok, agreed. The runtime hit is definitely an issue. Thinking about it more, even the notification block changes aren't dependent on this method. Thanks for taking a look @tgoyne!

PR #3294, opened 2016-03-04
Update to issue template.

bmunkholm:
Added intro.
Removed some (obvious?) details.

@mrackwitz

Merge at will :-)

### DIFF
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -1,7 +1,17 @@
-> Please fill out this template when filing an issue.
-> All lines marked as quotes (`>`) instruct you with
-> what info we expect.
-> Remove this opening paragraph.
+> We LOVE to help with any issues or bugs you have!
+>
+> **Questions**: If you have questions about how to use Realm, ask on 
+> [StackOverflow](http://stackoverflow.com/questions/ask?tags=realm).
+> We monitor the `realm` tag.
+>
+> **Feature Request**: Just fill in the first two sections below.
+>
+> **Bugs**: To help you as fast as possible with an issue please describe your issue
+> and the steps you have taken to reproduce it in as many details as possible.
+>
+> Thanks for helping us help you! :-)
+>
+> Please remove this line and all above before submitting.
 
 ## Goals
 
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -39,8 +39,13 @@
 
 ## Version of Realm and Tooling
 
-> Share which versions of Realm are affected, the Xcode and OS X version
-> on which you compile, as well as the version of the
-> dependency manager involved (CocoaPods / Carthage).
-> [In the CONTRIBUTING guidelines](https://git.io/vgxJO), you will find
-> a script, which will help with that.
+[In the CONTRIBUTING guidelines](https://git.io/vgxJO), you will find a script,
+which will help determining these versions.
+
+Realm version: ?
+
+Xcode version: ?
+
+iOS/OSX version: ?
+
+Dependency manager + version: ?
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -35,6 +35,7 @@ What are steps we can follow to reproduce this issue?
 Provide a code sample or test case that highlights the issue.
 If relevant, include your model definitions.
 For larger code samples, links to external gists/repositories are preferred.
+Alternatively share confidentially via mail to help@realm.io.
 Full Xcode projects that we can compile ourselves are ideal!
 
 ## Version of Realm and Tooling
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -15,27 +15,27 @@
 
 ## Goals
 
-> Replace this with what you tried to achieve.
+What do you want to achieve?
 
 ## Expected Results
 
-> Replace this with what you expected to happen.
+What did you expected to happen?
 
 ## Actual Results
 
-> Replace this with what actually happened.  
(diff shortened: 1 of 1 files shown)


.github/ISSUE_TEMPLATE.md
```diff
@@ -1,36 +1,46 @@
...
+
+> **Questions**: If you have questions about how to use Realm, please ask on [SO](http://stackoverflow.com/questions/ask?tags=realm) - we monitor the Realm tag.
+
+> **Feature Request**: Just fill in the first two sections below.
+
+> **Bugs**: To help you as fast as possible with an issue or bug please describe your issue and the steps you have taken to reproduce it in as many details as possible.
+>
+> Thanks for helping us help you :-)
+>
+> Please remove this and above before submitting.
```
mrackwitz: Not sure about this. This is quite a lot of intro text. I think it's preferable to keep it short. This duplicates a lot of the info of CONTRIBUTING.md, but this is sadly rarely read. In advance we should keep breaking the text at a fixed width, so that it is easier to read in markdown syntax, which is displayed by default.

.github/ISSUE_TEMPLATE.md
```diff
@@ -1,36 +1,46 @@
...
 
 ## Goals
 
-> Replace this with what you tried to achieve.
+?
```
mrackwitz: I think a description of what we expect here in other words and that this should be replaced is actually helpful.

tgoyne:
I think the pointer to SO for questions and the body text changes are nice, but I agree with Marius that the placeholder descriptions of what goes in each section seems useful.

.github/ISSUE_TEMPLATE.md
bmunkholm: Some of them might be helpful. I just saw that half of them did actually not add any value at all. E.g. the details under "Steps to reproduce". So my intent was to get rid of the "noise" of text that didn't add useful value. Then I can see some might and we can keep those or we can add it if we see it's a problem? This way I think we can have just the intro and instructions mostly at the top, and after people have read that once, they will just quickly skip it in the future, but the template itself will be easier/lean.
But if there are specific headings were details are needed to explain it we can do that. In Java I made a slightly different version that had the details in the headings instead. Not sure it looks so good now though as the headings are big. But that could be an alternative as well.

bmunkholm: @mrackwitz @tgoyne : Can you point out any particular headings that has a valuable detailed description that we must keep?

.github/ISSUE_TEMPLATE.md
```diff
@@ -1,36 +1,50 @@
...
-> a script, which will help with that.
+Realm version: ?
+
+Xcode version: ?
+
+iOS/OSX version: ?
+
+Dependency manager + version: ?
+
+> [In the CONTRIBUTING guidelines](https://git.io/vgxJO), you will find a script,
```
mrackwitz: If I would go through from top to bottom and fill out the things and then see that there is something, which could help me to automate it, I'd think: "too late, let's go for it now". 
So putting this at the end reduces the likelihood that this is actually used while it does a very good job in ensuring that all version numbers which might needed are provided and save us some communication cycles.

.github/ISSUE_TEMPLATE.md
mrackwitz: >I just saw that half of them did actually not add any value at all. E.g. the details under "Steps to reproduce". So my intent was to get rid of the "noise" of text that didn't add useful value. Then I can see some might and we can keep those or we can add it if we see it's a problem? This way I think we can have just the intro and instructions mostly at the top, and after people have read that once, they will just quickly skip it in the future, but the template itself will be easier/lean.

I would prefer ideas how to improve these section hints, which you think are not helpful at the moment. I don't think that noise is an actual problem.

.github/ISSUE_TEMPLATE.md
```diff
@@ -1,36 +1,50 @@
...
 
 ## Actual Results
 
-> Replace this with what actually happened.  
-> e.g. the stack trace of a crash
+?
 
 ## Steps to Reproduce
 
-> Provide steps we can follow to reproduce this issue.
```
mrackwitz: I hoped this puts some emphasis on that **we** must be able to follow these steps and it isn't helpful at all, if some write transaction fails for a database we don't have access to. That at least was my intent behind it.

.github/ISSUE_TEMPLATE.md
```diff
@@ -1,36 +1,50 @@
...
 
 ## Expected Results
 
-> Replace this with what you expected to happen.
+?
 
 ## Actual Results
 
-> Replace this with what actually happened.  
-> e.g. the stack trace of a crash
```
mrackwitz: I think especially the last line gives some orientation where the crash log should be put.

.github/ISSUE_TEMPLATE.md
```diff
@@ -1,36 +1,50 @@
...
 
 ## Steps to Reproduce
 
-> Provide steps we can follow to reproduce this issue.
+?
 
 ## Code Sample
 
-> Provide a code sample that highlights the issue.
+> Provide a code sample or test case that highlights the issue.
```
mrackwitz: If you can provide a test case, a PR is even better. But I don't think we can expect that. I'm confident that those who can deliver a test will be aware that filing a PR is more helpful.

bmunkholm:
If you really feel strongly that some of those tips adds value, Feel free to keep them. But I would at least remove those tips that is mainly a repetition of the heading.
Feel free to edit this PR with those changes. I'm mostly very much pro for the intro (which should be common among products)

mrackwitz:
Ok, I can take over here and address that. But, I agree on the consistent intro, which I hope will help to (re)direct issues in the future and brings in a welcoming tone, we might have skipped so far in favor of brevity. :+1:

PR #3295, opened 2016-03-04
Report changes made via addOrUpdate/createOrUpdate to KVO

### LINKED ISSUE #3292: KVO notification after Realm is updated via realm.add(_, update: true) - possible?
Hi, I am trying to understand if it is already possible to use KVO to get updates from Realm when updating object's properties via `realm.add(_, update: true)` method.
Was this supposed to work already (RealmSwift 0.98.3)?

```Swift
func NewChildObject() -> ChildObject
{
	let res = ChildObject()
	res.userId = NSUUID().UUIDString
	res.pwd_guid = NSUUID().UUIDString
	res.name = "Peppa"
	return res
}
// ...
context("object in Realm") {
	var child: ChildObject!
	var observedName: String?
	var observationName: DisposableType?
	beforeEach {
		child = NewChildObject()
		// add to Realm
		try! realm.write { realm.add(child) }
		// then it's possible to observe changes via KVO
		observationName = child.rValueForKeyPath("name", sendInitial: true, retainStrongly: true).observe { name in
			observedName = name
			}.scopedDisposable
		expect(observedName) == child.name
	}
	afterEach {
		observationName?.dispose()
		observationName = nil
	}

	it("propagates object properties' changes to other instances with the same primary key once saved to Realm") {
		let child2 = NewChildObject()
		child2.userId = child.userId // primary key
		child2.name = "George"

		// update object
		try! realm.write { realm.add(child2, update: true) }

		// changes are propagated to other instances with the same primary key
		expect(child.name) == child2.name // (1)
		expect(observedName) == child2.name // <<<--- failed - expected to equal <George>, got <Peppa> (2)

		// another try
		try! realm.write { child2.name = "Alice" }
		expect(child.name) == child2.name // (3)
		expect(observedName) == child2.name // but this one works! (4)
	}
}
```
Since the original `child.name` property was updated as expected (1), I supposed that KVO notification would also be sent (2). Since case (4) already works when updating object via property setter, do you think making (2) work will take much effort?

Thank you!

tgoyne:
And fix a similar issue where modifying a RealmOptional property did not notify other accessors for the same row of the change.

Fixes #3292.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,9 @@ x.x.x Release notes (yyyy-MM-dd)
 
 ### Bugfixes
 
-* None.
+* Properly report changes made by adding an object to a Realm with
+  addOrUpdate:/createOrUpdate: to KVO observers for existing objects with that
+  primary key.
 
 0.98.3 Release notes (2016-02-26)
 =============================================================
--- a/Realm/RLMAccessor.mm
+++ b/Realm/RLMAccessor.mm
@@ -787,93 +787,93 @@ void RLMDynamicValidatedSet(RLMObjectBase *obj, NSString *propName, id val) {
         @throw RLMException(@"Invalid property value `%@` for property `%@` of class `%@`", val, propName, obj->_objectSchema.className);
     }
 
-    RLMWrapSetter(obj, prop.name, [&] {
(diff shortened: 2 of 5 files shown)


mrackwitz:
:+1:

Issue #3185, opened 2016-02-08
Realm issue: Use of unimplemented initialiser 'init(realm:schema:)' for class Storyboard


rameez-leftshift:
Following is the code snippet of my class:

```swift
public class StoryBoard: Object, Deserializable {
    public dynamic var status: Status?

    required public init() {
        status = Status()
        super.init()
    }

    required public init(data: [String: AnyObject]) {
        let new_id: Int = data["id"]! as! Int
        id =  "\(new_id)"
        super.init()
    }

    public class func fromJSON(data: [String : AnyObject]) -> StoryBoard? {
        return StoryBoard(data: data)
    }
}
```

I'm using swift version 2.1.1 and realm version 0.97.1. I have tried using the convenience initilaizer but couldn't resolve the issue. Thank you in advance. Also updated to 0.98.0, but the app still crashes.

kishikawakatsumi:
@rameez-leftshift Can you try to define convenience initializer instead overriding initializer? Like the following.

```swift
public class StoryBoard: Object {
    public dynamic var status: Status? = Status()

    convenience public init(data: [String: AnyObject]) { // Define convenience initializer
        let new_id: Int = data["id"]! as! Int
        id =  "\(new_id)"
        self.init() // Call self.init(), not super.init()
    }

    public class func fromJSON(data: [String : AnyObject]) -> StoryBoard? {
        return StoryBoard(data: data)
    }
}
```

Overriding initializer can work if you also implement `init(realm:schema:)`. But we don't recommend override `required init()` due to Swift's limitation.

See more details: https://github.com/realm/realm-cocoa/issues/1849#issuecomment-99547742

rameez-leftshift:
He i was able to get it work by removing the required public init and making the one with data params a convenience initialiser. Sorry forgot to update here have been busy off late.

mrackwitz:
Nice, thanks for taking the time to update us here. So I guess, we can close this now. :)

rameez-leftshift:
Yup sure. Thanks for the help appreciate it :)

adamstener:
I'm not able to get this working using just a convenience initializer.  If I only have a conv init, when I call self.init(), it expects the parameters to be filled in for the convenience init.

`//MARK: - Initialization
    convenience public init(myString: String) {
        let components: NSDateComponents = NSDateComponents()
        components.setValue(-UserConstants.Values.DefaultAge, forComponent: NSCalendarUnit.Year);
[...]

PR #3305, opened 2016-03-09
Mark all required initializers as NS_DESIGNATED_INITIALIZER

mrackwitz:
This is a follow up to investigations I did for #3185.

Even though something in that direction was brought up before by @kishikawakatsumi with #2514, which was finally rejected, I think what we currently have is problematic.

As seen in the second commit, it is currently possible to define an `Object` subclass, overwriting all initializers, which are declared as required, which perfectly compiles, whose object schema validates, but which can't be used universally at runtime as it e.g. fails when trying to access `valueForKey` on a `Results` collection of that object. The error in this case is:
>fatal error: use of unimplemented initializer 'init(realm:schema:)'

/c @jpsim @tgoyne

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,8 @@ x.x.x Release notes (yyyy-MM-dd)
 
 ### Bugfixes
 
-* None.
+* Mark further initializers in Objective-C as NS_DESIGNATED_INITIALIZER to prevent that these aren't
+  correctly defined in Swift Object subclasses, which don't qualify for auto-inheriting the required initializers.
 
 0.98.5 Release notes (2016-03-14)
 =============================================================
--- a/Realm/RLMObject.mm
+++ b/Realm/RLMObject.mm
@@ -41,6 +41,14 @@
     return [super init];
 }
 
+- (instancetype)initWithValue:(id)value schema:(RLMSchema *)schema {
+    return [super initWithValue:value schema:schema];
+}
+
+- (instancetype)initWithRealm:(__unsafe_unretained RLMRealm *const)realm
+                       schema:(__unsafe_unretained RLMObjectSchema *const)schema {
+    return [super initWithRealm:realm schema:schema];
+}
 
 #pragma mark - Convenience Initializers
 
--- a/Realm/RLMObject.mm
+++ b/Realm/RLMObject.mm
@@ -35,9 +35,14 @@
 // synthesized in RLMObjectBase
 @dynamic invalidated, realm, objectSchema;
 
+#pragma mark - Designated Initializers
+
 - (instancetype)init {
     return [super init];
 }
+
+
+#pragma mark - Convenience Initializers
 
 - (instancetype)initWithValue:(id)value {
     [self.class sharedSchema]; // ensure this class' objectSchema is loaded in the partialSharedSchema
@@ -48,6 +53,8 @@
 - (instancetype)initWithObject:(id)object {
     return [self initWithValue:object];
 }
+
+#pragma mark - Class-based Object Creation
 
 + (instancetype)createInDefaultRealmWithValue:(id)value {
     return (RLMObject *)RLMCreateObjectInRealmWithValue([RLMRealm defaultRealm], [self className], value, false);
@@ -87,6 +94,8 @@
     return [self createOrUpdateInRealm:realm withValue:object];
 }
 
+#pragma mark - Subscripting
+
 - (id)objectForKeyedSubscript:(NSString *)key {
     return RLMObjectBaseObjectForKeyedSubscript(self, key);
 }
@@ -94,6 +103,8 @@
 - (void)setObject:(id)obj forKeyedSubscript:(NSString *)key {
     RLMObjectBaseSetObjectForKeyedSubscript(self, key, obj);
 }
+
+#pragma mark - Getting & Querying
 
 + (RLMResults *)allObjects {
     return RLMGetObjects(RLMRealm.defaultRealm, self.className, nil);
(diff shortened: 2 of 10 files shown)


jpsim:
Requiring that APIs from the Realm Objective-C be implemented when using Realm Swift is an unacceptable solution.

mrackwitz:
Yeah, it is definitely not an UX / API improvement. But so far we allow to define broken object models, which are only partly usable / crash unexpectedly at runtime. If we can't avoid these constructors in the first place, I don't see any other solution than that. It is another strong motivation to define convenience constructors instead.

jcj52436999:
This message was created automatically by mail delivery software.

A message that you sent could not be delivered to one or more of its
recipients. This is a temporary error. The following address(es) deferred:

  jcjacksonengineering@gmail.com
    Domain joe-jackson.net has exceeded the max emails per hour (121/100 (121%)) allowed.  Message will be reattempted later

------- This is a copy of the message, including all the headers. ------
Received: from github-smtp2-ext6.iad.github.net ([192.30.252.197]:54736 helo=github-smtp2b-ext-cp1-prd.iad.github.net)
	by rs5.websitehostserver.net with esmtps (TLSv1.2:ECDHE-RSA-AES256-GCM-SHA384:256)
	(Exim 4.86_1)
	(envelope-from <noreply@github.com>)
	id 1afmNl-001y1C-Gz
	for jcjackson@joe-jackson.net; Tue, 15 Mar 2016 06:39:21 -0400
Date: Mon, 14 Mar 2016 17:48:23 -0700
DKIM-Signature: v=1; a=rsa-sha256; c=relaxed/relaxed; d=github.com;
	s=pf2014; t=1458002903;
	bh=2ECeyHhxE3o4K1AseiSx0z4VRDlHCBT3mcfKKySrWpw=;
	h=From:Reply-To:To:In-Reply-To:References:Subject:List-ID:
	 List-Archive:List-Post:List-Unsubscribe:From;
	b=rgemGaUqqJnQ8Bq9lxq93+MJm73+lPNdHPgMWQNykuNNjrWcTM000NqkpVW7JSKYy
	 1cH6BkbqAIixvSyZ3mZreXExMQqUdSr7BdWKDcCSccAMM1eWfQhgp/+K+5q5mmKHIB
	 kOjWQh0AAJhCEVE/y/KtswDTiCWOnij3qjDXC+0s=

yanndupuy:
This message was created automatically by mail delivery software.

A message that you sent could not be delivered to one or more of its
recipients. This is a temporary error. The following address(es) deferred:

  dupuyyann@gmail.com
    Domain zyapps.com has exceeded the max emails per hour (446/400 (111%)) allowed.  Message will be reattempted later

------- This is a copy of the message, including all the headers. ------
------ The body of the message is 10093 characters long; only the first
------ 5000 or so are included here.
Received: from github-smtp2-ext6.iad.github.net ([192.30.252.197]:60752 helo=github-smtp2b-ext-cp1-prd.iad.github.net)
	by serveur.ex2-p10.com with esmtps (TLSv1.2:ECDHE-RSA-AES256-GCM-SHA384:256)
	(Exim 4.86_1)
	(envelope-from <noreply@github.com>)
	id 1afmOs-000bG5-LQ
	for yann.dupuy@zyapps.com; Tue, 15 Mar 2016 11:40:40 +0100
Date: Tue, 15 Mar 2016 03:39:27 -0700
DKIM-Signature: v=1; a=rsa-sha256; c=relaxed/relaxed; d=github.com;
	s=pf2014; t=1458038367;
	bh=4uKi9B77OdSJo4Q4tOqhJL/8LJ8R3rqD5XrqSlfut7E=;
	h=From:Reply-To:To:In-Reply-To:References:Subject:List-ID:
	 List-Archive:List-Post:List-Unsubscribe:From;
	b=YVBXkTSB9YunWr4pVoDfXo5NRRphI8t3VwgAiKdlvRcAs3BePx5ofbkPtWufVjFkx
	 CrQf7hyhb6g3lHlrWDL+JBtLn7/M4Ze9vkCJkv8bPSsfDTajMFdBqjQAvUd3UOdNE7
	 OOUe2B2KnR5qbC/eoK0G0A3lJWaI38VdtTUl3AzA=

mrackwitz:
So how should we move on here?

jpsim:
Ah, what wasn't clear to me in this PR was that existing convenience initializers would continue to work:

```swift
class SwiftCustomInitializerObject: Object {
    dynamic var stringCol = ""

    convenience init(stringVal: String) {
        self.init()
        stringCol = stringVal
    }
}
```

I thought this was a breaking change, but it's not. The non-convenience initializer version never worked, but would after this change. I'm all for this.

Could you please also add a unit test for an object with a convenience initializer like the one I've shared above to ensure we don't regress on this. Especially since that's how [we recommend users add custom initializers](https://realm.io/docs/swift/latest/#adding-custom-initializers-to-object-subclasses) (which we should continue to do IMHO). This will need a rebase too.

mrackwitz:
>Could you please also add a unit test for an object with a convenience initializer like the one I've shared above to ensure we don't regress on this.

Well, we could test an object with a convenience initializer at runtime, but this wouldn't be factually different from any other object. The convenience initializer will be never called by the framework. Testing that a convenience initializer can be called in general is not our responsibility. We could define an object class with a convenience initializer and make sure that this compiles, while that only ensures that there is an inherited constructor without any arguments which can be called. That again is well-tested in the entire test suite. So I don't think that such a test would really make sense.

jpsim:
> Well, we could test an object with a convenience initializer at runtime, but this wouldn't be factually different from any other object. The convenience initializer will be never called by the framework.

`RealmSwift.Object` is fairly unique in that it inherits from an Objective-C subclass, along with a combination of inherited Objective-C, Swift, and user-defined (designated/convenience) initializers. I wouldn't be surprised if at some point an update to either Swift or Realm breaks convenience initializers to fail to compile or dispatch properly, which is why we should really have a test for this, considering that we encourage our users to use convenience initializers in our documentation

mrackwitz:
I added a unit test, which checks the correct initialization of the object schema when a convenience initializer was used.

The tvOS failure on Jenkins is very likely unrelated.

jpsim:
LGTM! :+1:

tgoyne:
:+1:

PR #3309, opened 2016-03-11
remove unnecessary swizzling

Changsung:
#3235
`RLMReplaceClassNameMethod` does unnecessary swizzling. In `RLMObjectBase.mm` the method which is `+ (NSString *)className ` is the same as `className` in the code. 

On iOS 9.2, Swizzling slows down the performance, so I hope it would be changed.

### DIFF
--- a/Realm/RLMSchema.mm
+++ b/Realm/RLMSchema.mm
@@ -107,7 +107,6 @@ static void RLMRegisterClassLocalNames(Class *classes, NSUInteger count) {
         }
 
         s_localNameToClass[className] = cls;
-        RLMReplaceClassNameMethod(cls, className);
     }
 }
 


bdash:
The swizzling is a performance optimization. The `+className` method is frequently called, and the the base class implementation is relatively expensive. Giving each class its own implementation significantly reduces the per-call cost of the method. In other words, this pull request is trading reduced work during initialization for increased overhead on every call to `+className`. Given that iOS 9.3 addresses the performance issue with swizzling and is likely to be released this month, I'm not sure that it's worth changing the performance tradeoff at this point.

jpsim:
I echo @bdash's sentiments on this performance tradeoff, and that the one we're making today is the best way forward considering future changes coming in iOS 9.3. You can see more work that was done in this area recently in #3235, which also explains a bit of the rationale that also applies here.

So thanks for the contribution, but we won't be accepting this PR. I hope you'll contribute to Realm again in the future!

PR #3311, opened 2016-03-11
Set TMPDIR value correctly if it's nil.

nakiwo:
`build.sh` set `TMPDIR`  to string `"getconf DARWIN_USER_TEMP_DIR"` if TMPDIR is empty.
This causes instlation failure. (when TMPDIR is empty)

`getconf DARWIN_USER_TEMP_DIR` should be enclosed by `$()`.


jpsim:
Thanks @nakiwo!

PR #3318, opened 2016-03-11
lint podspecs during build.sh verify-cocoapods

jpsim:
This will lint all podspecs (`Realm.podspec` & `RealmSwift.podspec`) /cc @bdash


bdash:
It looks like the CI job for verify-coocapods failed due to the worker having a cached copy of core tarball with the headers that generate the warning. Might have to remove that on all of the VMs prior to merging this?

:+1:

PR #3357, opened 2016-03-22
Extract primaryKey class method into PrimaryKeyObject protocol

mrackwitz:
:warning: This would be a breaking change.

By moving the class method `Object.primaryKey` into a protocol `PrimaryKeyObject`, we can restrict the methods which allow upserts only to those `Object` subclasses which define a primary key.


jpsim:
@tgoyne @bdash and I just discussed this offline and although the concept of making things more type-safe seemed good at first glance, in practice this ends up increasing the API footprint without much practical gain. Ultimately, I don't think this is a worthwhile addition, sorry.

PR #3359, opened 2016-03-22
Add change information to collection notifications

### LINKED ISSUE #687: Have RLMResults report query changesets when realm is updated / NSFetchedResultsController alternative
This would be great for animations similar to `NSFetchedResultsController`.

### LINKED ISSUE #3435: Document notification coalescing
## Goals

This is what the API docs say (thanks for fixing this) about `Results.addNotificationBlock(_)`:

> The block will be asynchronously called with the initial results, and then called again after each write transaction which causes the results to change.

I have code that creates a `Results` object and then writes one object that changes the results. In this case I'd expect my notification block to be called two times. The block however is called just one time.

## Expected Results

The block to be called for the initial result set

## Actual Results

The notification block is not called for the initial result set.

## Steps to Reproduce

```swift
Here's the complete source code to reproduce:

import UIKit
import RealmSwift

class Message: Object {
    dynamic var id = 0
    dynamic var text = ""
}

class ViewController: UIViewController {

    var token: NotificationToken?
    
    override func viewDidLoad() {
        super.viewDidLoad()
        
        let realm = try! Realm()
        try! realm.write {
            realm.deleteAll()
        }
        
        let hello1 = Message()
        hello1.text = "Hello!"

        try! realm.write {
            realm.add(hello1)
        }
        
        let feed = realm.objects(Message)
        
        print("initial count: \(feed.count)")
        token = feed.addNotificationBlock {newResults, _ in
            print("notification count: \(newResults!.count)")
        }

        let hello = Message()
        hello.text = "Hello!"
        
        try! realm.write {
            realm.add(hello)
        }
        print("changed count: \(feed.count)")
    }
    
}
```

Here's the output from the print statements:

image

Realm version: 0.98.7

Xcode version: 7,3

iOS/OSX version: latest

tgoyne:
This adds a third parameter to the notification block on Results/List/RLMResults/RLMArray/etc. that reports the indices in the collection which were deleted, inserted and modified. In addition, it skips sending the notification when the collection did not actually change, rather than notifying on every change to the relevant table(s). It does not include information about which property on the object at the given index changed or supply the old value at that index; for those KVO on each object must be used.

Most of the testing for this is done at the objectstore level, so see the [corresponding PR there](https://github.com/realm/realm-object-store/pull/61).

Closes #687. Closes #3435.

Things left to do:

 - [X] Functionality
   - [X] Change information for List
   - [X] Change information for unsorted Results
   - [X] Change information for sorted Results
   - [X] Change information for Results derived from List
   - [X] Finalize API
 - [x] Testing
   - [X] Write tests covering everything that can reasonably be tested in object-store
   - [X] Obj-c integration tests
   - [X] Swift integration tests
   - [X] TSan on object store tests
   - [x] TSan on realm-cocoa tests
   - [X] Fuzz-test unsorted query changesets
   - [X] Fuzz-test sorted query changesets
   - [X] Fuzz-test unsorted linkview changesets
   - [X] Fuzz-test sorted linkview changesets
 - [ ] Documentation
   - [X] Write API docs
   - [ ] Update the guide
   - [ ] Release post
   - [X] Update existing UITableView examples
   - [X] <s>Add an NSOutlineView example</s>
 - [ ] Performance
   - [x] Small changesets, Unsorted
   - [x] Small changesets, Sorted
   - [x] Small changesets, List
   - [X] Many changesets, Unsorted
   - [X] Many changesets, Sorted
   - [x] Many changesets, List
   - [x] Large changesets, Unsorted
   - [x] Large changesets, Sorted
   - [ ] Large changesets, List

My current goal performance-wise is to have everything be fast enough that no one will need a way to opt-out of the changeset calculations, as that'd make the API unpleasant. Since most of the work is done on background threads and doesn't block the UI this should be easily attainable.


RealmSwift-swift2.0/RealmCollectionType.swift
```diff
@@ -41,6 +41,120 @@ public final class RLMGenerator<T: Object>: GeneratorType {
...
+            case .Update(_, let deletions, let insertions, let modifications):
+                // Query results have changed, so apply them to the TableView
+                self.tableView.beginUpdates()
+                self.tableView.insertRowsAtIndexPaths(insertions, withRowAnimation: .Automatic)
+                self.tableView.deleteRowsAtIndexPaths(deletions, withRowAnimation: .Automatic)
+                self.tableView.reloadRowsAtIndexPaths(modifications, withRowAnimation: .Automatic)
+                self.tableView.endUpdates()
+                break
+            case .Error(let err):
+                // An error occured while opening the Realm file on the background worker thread
```
TimOliver: I noticed this when copying this code for the documentation: 'occurred' needs another 'r' there.

examples/ios/swift-2.2/TableView/TableViewController.swift
```diff
@@ -46,8 +46,25 @@ class TableViewController: UITableViewController {
...
+            case .Update(_, let deletions, let insertions, let modifications):
+                // Query results have changed, so apply them to the TableView
+                self.tableView.beginUpdates()
+                self.tableView.insertRowsAtIndexPaths(insertions, withRowAnimation: .Automatic)
+                self.tableView.deleteRowsAtIndexPaths(deletions, withRowAnimation: .Automatic)
+                self.tableView.reloadRowsAtIndexPaths(modifications, withRowAnimation: .Automatic)
+                self.tableView.endUpdates()
+                break
+            case .Error(let err):
+                // An error occured while opening the Realm file on the background worker thread
```
TimOliver: Same as the other one: 'occurred' needs another 'r'. :)

examples/ios/objc/TableView/TableViewController.m
```diff
@@ -51,8 +51,25 @@ - (void)viewDidLoad
 
     // Set realm notification block
     __weak typeof(self) weakSelf = self;
-    self.notification = [self.array addNotificationBlock:^(RLMResults *data, NSError *error) {
-        [weakSelf.tableView reloadData];
+    self.notification = [self.array addNotificationBlock:^(RLMResults *data, RLMCollectionChange *changes, NSError *error) {
```
TimOliver: While the headers and documentation all say `RLMCollectionChange *change`, this one defines `RLMCollectionChange *changes` with `changes` being pluralized.

For consistency, and for more natural sounding English, I think the headers should be changed to say `changes` as well. But that's definitely something we can discuss.

jpsim:
Instead of having two separate `addNotificationBlock` overloads for `NSIndexPath`/`Int` block parameters, could we use platform-specific typealiases? e.g. `Int` for OS X, `NSIndexPath` for others?

Otherwise, our users will end up fighting Swift's type inference system:

image

vs

image

tgoyne:
After thinking about it some more and looking at some code using the notifications, I'm now thinking that the obj-c NSIndexPath version should be `[change insertionsInSection:0]` (withSection?), and the Swift one should just be removed entirely.

Only generating paths for section 0 makes it not useful for grouped table views, and generalizing it doesn't really make it worse for the simple case. In Swift, you can just do `insertions.map { NSIndexPath(section: 0, index: $0) }`, so the whole thing is kinda pointless.

tgoyne:
I tried using fine-grained notifications for the GroupedTableView example, and it turns out that the current design doesn't work for that when a single transaction modifies multiple sections.  The problem is that all of the sections need to be updated within a single UITableView batch as it rechecks the sizes of all of the groups when you call `endUpdates`, but from within the notification block there's no good way to know if there's still more blocks which will be called.  Some ideas for how to handle this:

1. Don't, and just document it as an unsupported use-case
2. Do something like `let tokens = [results1, results2, results].addNotificationBlock { [changes] in ... }` to make it possible to get all of the change notifications at once. I'm not really sure what the obj-c API would look like for this.
3. Delay sending the Realm DidChange notification until after the async queries have been updated so that you can end the batch there
4. Something less awful?

Realm/ObjectStore/collection_notifications.hpp
```diff
@@ -0,0 +1,71 @@
...
+    ~NotificationToken();
+
+    NotificationToken(NotificationToken&&);
+    NotificationToken& operator=(NotificationToken&&);
+
+    NotificationToken(NotificationToken const&) = delete;
+    NotificationToken& operator=(NotificationToken const&) = delete;
+
+private:
+    util::AtomicSharedPtr<_impl::BackgroundCollection> m_query;
```
bdash: `m_query` doesn't seem like an appropriate name for this member given that the thing it holds is described as a collection.

Realm/ObjectStore/impl/background_collection.hpp
```diff
@@ -0,0 +1,205 @@
...
+    void add_changes(CollectionChangeBuilder change) { m_accumulated_changes.merge(std::move(change)); }
+    void set_table(Table const& table);
+    std::unique_lock<std::mutex> lock_target();
+
+private:
+    virtual void do_attach_to(SharedGroup&) = 0;
+    virtual void do_detach_from(SharedGroup&) = 0;
+    virtual void do_prepare_handover(SharedGroup&) = 0;
+    virtual bool do_deliver(SharedGroup&) { return true; }
+    virtual bool do_add_required_change_info(TransactionChangeInfo&) { return true; }
```
bdash: What does the bool return value of `do_deliver` and `do_add_required_change_info` signify?

Realm/ObjectStore/impl/background_collection.cpp
```diff
@@ -0,0 +1,832 @@
...
+}
+
+void BackgroundCollection::detach()
+{
+    REALM_ASSERT(m_sg);
+    do_detach_from(*m_sg);
+    m_sg = nullptr;
+}
+
+CollectionChangeBuilder::CollectionChangeBuilder(IndexSet deletions,
```
bdash: It might be worth moving CollectionChangeBuilder into its own file.

Realm/ObjectStore/impl/background_collection.cpp
```diff
@@ -0,0 +1,832 @@
...
+                                                 IndexSet insertions,
+                                                 IndexSet modifications,
+                                                 std::vector<Move> moves)
+{
+    this->deletions = std::move(deletions);
+    this->insertions = std::move(insertions);
+    this->modifications = std::move(modifications);
+    this->moves = std::move(moves);
```
bdash: Any reason to initialize these in the constructor body rather than as part of the initialization list?

Realm/ObjectStore/impl/background_collection.cpp
```diff
@@ -0,0 +1,832 @@
...
+    };
+    struct Row {
+        size_t row_index;
+        size_t tv_index;
+    };
+
+    IndexSet const& m_modified;
+    std::vector<Match> m_longest_matches;
+
+    std::vector<Row> a, b;
```
bdash: Can `a` and `b` be given more descriptive names?

Realm/ObjectStore/impl/background_collection.cpp
```diff
@@ -0,0 +1,832 @@
...
+            for (auto const& pair : prev) {
+                if (pair.j + 1 == j)
+                    return pair.len + 1;
+            }
+            return 1;
+        };
+
+        Match best = {begin1, begin2, 0, 0};
+
+        for (size_t i = begin1; i < end1; ++i) {
```
bdash: It's hard to see at a high level what this loop is doing. The generic variable names don't help. Can you add a comment explaining briefly what this loop is doing?

Realm/ObjectStore/impl/transact_log_handler.cpp
```diff
@@ -408,24 +414,154 @@ class TransactLogObserver : public TransactLogValidator {
...
+    bool select_link_list(size_t col, size_t row, size_t)
+    {
+        mark_dirty(row, col);
+
+        m_active = nullptr;
+        for (auto& o : m_info.lists) {
+            if (o.table_ndx == current_table() && o.row_ndx == row && o.col_ndx == col) {
+                m_active = o.changes;
+                // need to use last match for multiple source version logic
+//                break;
```
bdash: Should this `break` be commented out?

Realm/ObjectStore/util/atomic_shared_ptr.hpp
```diff
@@ -0,0 +1,137 @@
...
+            m_ptr = ptr.m_ptr;
+            m_mutex.unlock();
+            ptr.m_mutex.unlock();
+        }
+        return *this;
+    }
+
+    AtomicSharedPtr& operator=(AtomicSharedPtr&& ptr)
+    {
+        std::lock(m_mutex, ptr.m_mutex);
```
bdash: Should we check for self assignment here? It looks as though we'd deadlock if that were attempted.

bdash:
There's some very dense code in the change computation and index set implementation. I've read over it looking for obvious errors, but have not studied it in depth. It looks to have very good test coverage so I'm not particularly concerned about going over it with a fine-tooth comb, but I have called out one or two places where a higher-level description of what a function is doing would help.

bdash:
@tgoyne, what work remains for the "Large changesets, List" item on your checklist?

bdash:
NotificationTests.m has a number of errors with Xcode 6.4 in CI:

    Realm/Tests/NotificationTests.m:234:61: error: no known instance method for selector 'section'
        XCTAssert(deletionPaths.count == 0 || [deletionPaths[0] section] == section);

Realm/ObjectStore/util/atomic_shared_ptr.hpp
tgoyne: Self move-assignment on standard library types is UB, so it's reasonably safe to just ignore.

bdash:
I love the newly-added comments explaining the move detection code!

👍

imton:
Thank you @tgoyne and the team!

Realm/ObjectStore/impl/realm_coordinator.cpp
```diff
@@ -201,10 +197,28 @@ void RealmCoordinator::clear_cache()
     }
 }
 
+void RealmCoordinator::clear_all_caches()
```
jpsim: This doesn't appear to be used anywhere. Is it still worth keeping around?

Realm/Tests/NotificationTests.m
```diff
@@ -0,0 +1,633 @@
...
+    ExpectNoChange(self, ^(RLMRealm *realm) {
+        RLMArray *array = [[[ArrayPropertyObject allObjectsInRealm:realm] firstObject] intArray];
+        IntObject *io = [IntObject createInRealm:realm withValue:@[@5]];
+        [array insertObject:io atIndex:2];
+    });
+}
+
+- (void)testExchangeObjects {
+    // adjacent swap: one move, since second is redundant
+//    ExpectChange(self, @[@1, @0], @[], @[], ^(RLMRealm *realm) {
```
jpsim: is there a reason these are still commented out?

tgoyne: I never implemented the non-naive version of swap handling that produces nice diffs.

loudenvier:
Is this functionality available for Realm Java?

PR #3360, opened 2016-03-22
Allow renaming properties in migrations

### LINKED ISSUE #2574: Make renaming properties in migrations easier
Migrating renames of object and array properties is difficult when the linked object has a primary key. Assigning the new property to the old property results in a new object always being inserted even if an object with that primary key already exists, and there's no way to query for the existing object with that primary key during a migration. We should have something like `[migration copyValuesFromProperty:@"oldName" toProperty:@"newName"]` which would handle all of this.

See #2570.

jpsim:
Fixes #2574. @tgoyne @bdash

### DIFF
--- a/Realm/ObjectStore/object_store.cpp
+++ b/Realm/ObjectStore/object_store.cpp
@@ -252,7 +252,10 @@ static void copy_property_values(const Property& source, const Property& destina
 // set references to tables on targetSchema and create/update any missing or out-of-date tables
 // if update existing is true, updates existing tables, otherwise validates existing tables
 // NOTE: must be called from within write transaction
-void ObjectStore::create_tables(Group *group, Schema &target_schema, bool update_existing) {
+std::vector<Property> ObjectStore::create_tables(Group *group, Schema &target_schema, bool update_existing) {
+    // properties to delete
+    std::vector<Property> to_delete;
+
     // first pass to create missing tables
     std::vector<ObjectSchema *> to_update;
     for (auto& object_schema : target_schema) {
@@ -295,8 +298,12 @@ void ObjectStore::create_tables(Group *group, Schema &target_schema, bool update
             current_prop.table_column -= deleted;
 
             auto target_prop = target_object_schema->property_for_name(current_prop.name);
-            if (!target_prop || (property_has_changed(current_prop, *target_prop)
-                                 && !property_can_be_migrated_to_nullable(current_prop, *target_prop))) {
+            // mark property for deletion
+            if (!target_prop) {
+                to_delete.push_back(current_prop);
+            }
+            else if ((property_has_changed(current_prop, *target_prop) &&
+                      !property_can_be_migrated_to_nullable(current_prop, *target_prop))) {
                 if (deleted == current_schema.persisted_properties.size() - 1) {
                     // We're about to remove the last column from the table. Insert a placeholder column to preserve
                     // the number of rows in the table for the addition of new columns below.
@@ -358,6 +365,26 @@ void ObjectStore::create_tables(Group *group, Schema &target_schema, bool update
             set_primary_key_for_object(group, target_object_schema->name, "");
         }
     }
+    return to_delete;
+}
+
+void ObjectStore::remove_properties(Group *group, Schema &target_schema, std::vector<Property> to_delete) {
+    for (auto& target_object_schema : target_schema) {
+        TableRef table = table_for_object_type(group, target_object_schema.name);
+        ObjectSchema current_schema(group, target_object_schema.name);
+        size_t deleted = 0;
+        for (auto& current_prop : current_schema.properties) {
+            current_prop.table_column -= deleted;
+            for (auto& single_to_delete : to_delete) {
+                if (current_prop.name == single_to_delete.name &&
+                    current_prop.object_type == single_to_delete.object_type) {
+                    table->remove_column(current_prop.table_column);
+                    ++deleted;
+                    current_prop.table_column = npos;
+                }
+            }
+        }
+    }
 }
 
 bool ObjectStore::is_schema_at_version(const Group *group, uint64_t version) {
@@ -402,9 +429,10 @@ void ObjectStore::update_realm_with_schema(Group *group, Schema const& old_schem
 
     // create tables
     create_metadata_tables(group);
-    create_tables(group, schema, migrating);
+    auto to_delete = create_tables(group, schema, migrating);
 
     if (!migrating) {
+        remove_properties(group, schema, to_delete);
         // If we aren't migrating, then verify that all of the tables which
         // were already present are valid (newly created ones always are)
         verify_schema(old_schema, schema, true);
@@ -419,6 +447,7 @@ void ObjectStore::update_realm_with_schema(Group *group, Schema const& old_schem
     // apply the migration block if provided and there's any old data
     if (get_schema_version(group) != ObjectStore::NotVersioned) {
         migration(group, schema);
+        remove_properties(group, schema, to_delete);
 
         validate_primary_column_uniqueness(group, schema);
     }
--- a/Realm/ObjectStore/object_store.cpp
+++ b/Realm/ObjectStore/object_store.cpp
@@ -373,7 +373,7 @@ void ObjectStore::remove_properties(Group *group, Schema &target_schema, std::ve
         TableRef table = table_for_object_type(group, target_object_schema.name);
         ObjectSchema current_schema(group, target_object_schema.name);
         size_t deleted = 0;
-        for (auto& current_prop : current_schema.properties) {
+        for (auto& current_prop : current_schema.persisted_properties) {
             current_prop.table_column -= deleted;
             for (auto& single_to_delete : to_delete) {
                 if (target_object_schema.name == single_to_delete.first &&
@@ -554,7 +554,7 @@ void ObjectStore::rename_property(Group *group, Schema& passed_schema, StringDat
     table->rename_column(old_property->table_column, new_name);
     table->remove_column(column_to_remove);
     // update table_column for each property since it may have shifted
-    for (auto& current_prop : passed_object_schema->properties) {
+    for (auto& current_prop : passed_object_schema->persisted_properties) {
         auto target_prop = matching_schema.property_for_name(current_prop.name);
         current_prop.table_column = target_prop->table_column;
     }
--- a/Realm/ObjectStore/object_store.cpp
+++ b/Realm/ObjectStore/object_store.cpp
@@ -432,7 +432,6 @@ void ObjectStore::update_realm_with_schema(Group *group, Schema const& old_schem
     auto to_delete = create_tables(group, schema, migrating);
 
     if (!migrating) {
-        remove_properties(group, schema, to_delete);
         // If we aren't migrating, then verify that all of the tables which
         // were already present are valid (newly created ones always are)
         verify_schema(old_schema, schema, true);
--- a/Realm/ObjectStore/object_store.cpp
+++ b/Realm/ObjectStore/object_store.cpp
@@ -252,9 +252,9 @@ static void copy_property_values(const Property& source, const Property& destina
 // set references to tables on targetSchema and create/update any missing or out-of-date tables
 // if update existing is true, updates existing tables, otherwise validates existing tables
 // NOTE: must be called from within write transaction
-std::vector<Property> ObjectStore::create_tables(Group *group, Schema &target_schema, bool update_existing) {
+std::vector<std::pair<std::string, Property>> ObjectStore::create_tables(Group *group, Schema &target_schema, bool update_existing) {
     // properties to delete
-    std::vector<Property> to_delete;
+    std::vector<std::pair<std::string,Property>> to_delete;
 
     // first pass to create missing tables
     std::vector<ObjectSchema *> to_update;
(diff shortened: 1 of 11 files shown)


Realm/ObjectStore/object_store.cpp
```diff
@@ -400,9 +427,10 @@ void ObjectStore::update_realm_with_schema(Group *group, Schema const& old_schem
 
     // create tables
     create_metadata_tables(group);
-    create_tables(group, schema, migrating);
+    auto to_delete = create_tables(group, schema, migrating);
 
     if (!migrating) {
+        remove_properties(group, schema, to_delete);
```
tgoyne: It doesn't make any sense to be calling remove_properties() in the not migrating case.

Realm/ObjectStore/object_store.cpp
```diff
@@ -487,6 +516,57 @@ void ObjectStore::delete_data_for_object(Group *group, StringData object_type) {
...
+    if (old_property->type != new_property->type ||
+        old_property->object_type != new_property->object_type) {
+        throw PropertyRenameTypeMismatchException(*old_property, *new_property);
+    }
+    if (passed_object_schema->property_for_name(old_name) != nullptr) {
+        throw PropertyRenameOldStillExistsException(old_name, new_name);
+    }
+    size_t column_to_remove = new_property->table_column;
+    table->rename_column(old_property->table_column, new_name);
+    table->remove_column(column_to_remove);
```
tgoyne: This changes the column numbers for any other newly-added columns after the one being removed, which doesn't appear to be handled.

jpsim: that'll be handled in `verify_schema` at the end of the migration.

tgoyne: At the end of the migration is too late as the column number can be used during the migration.

jpsim: Ah yes, ok will adjust.

jpsim: Done.

bdash:
We talked about this briefly today and I mentioned a couple of issues with the current approach:

1. The "deleted" properties are accessible on objects from the new version within the migration block.
2. Renaming a property then later introducing a new property with the old name doesn't look like it'll work.
3. Renaming the same property multiple times over the lifetime of a schema doesn't look like it'll work.

jpsim:
> 1: The "deleted" properties are accessible on objects from the new version within the migration block.

This can be addressed by renaming the columns of deleted properties to a unique ID at the start of the migration, and maintain a map of names/uuids to use in the renaming operation.

> 2: Renaming a property then later introducing a new property with the old name doesn't look like it'll work.

I think it's necessary to remove all new schema validation altogether because of this, and other multi-version schema change scenarios.

> 3: Renaming the same property multiple times over the lifetime of a schema doesn't look like it'll work.

How about falling back to looking up the property by name in the new Realm? This would allow the following scenarios:

```swift
// v0
class A: Object {
  dynamic var p1 = 0
}
// v1
class A: Object {
  dynamic var p2 = 0
}
```

---

I'd like to get feedback on those changes before implementing them /cc @bdash @tgoyne.

Overall, a "simple" operation like renaming a property is complicated by not tracking intermediate schema changes. Other migration operations are also blocked on this design decision.

jpsim:
Latest feedback and issues from @bdash have now been addressed, so this is once again ready for review @tgoyne @bdash

Realm/ObjectStore/object_store.cpp
```diff
@@ -355,6 +365,23 @@ void ObjectStore::create_tables(Group *group, Schema &target_schema, bool update
...
+}
+
+void ObjectStore::remove_properties(Group *group, Schema &target_schema, std::unordered_map<std::string, Property> to_delete) {
+    for (auto& target_object_schema : target_schema) {
+        TableRef table = table_for_object_type(group, target_object_schema.name);
+        ObjectSchema current_schema(group, target_object_schema.name);
+        size_t deleted = 0;
+        for (auto& current_prop : current_schema.properties) {
+            current_prop.table_column -= deleted;
+            if (to_delete.find(current_prop.name) != to_delete.end()) {
```
tgoyne: This doesn't appear to work if multiple classes have properties with the same name.

Realm/ObjectStore/object_store.cpp
```diff
@@ -416,7 +443,8 @@ void ObjectStore::update_realm_with_schema(Group *group, Schema const& old_schem
     // apply the migration block if provided and there's any old data
     if (get_schema_version(group) != ObjectStore::NotVersioned) {
         migration(group, schema);
-
+        remove_properties(group, schema, to_delete);
```
tgoyne: `std::move(to_delete)`

Realm/RLMProperty.mm
```diff
@@ -429,6 +429,23 @@ - (id)copyWithZone:(NSZone *)zone {
     return prop;
 }
 
+- (RLMProperty *)copyWithNewName:(NSString *)name {
+    RLMProperty *prop = [self copy];
+    prop->_name = name;
+    prop->_type = _type;
```
tgoyne: All of these appear to be redundant since it's calling `copy` above.

jpsim:
I've beefed up the tests and fixed some update issues in the process, so this is ready for yet another round...

Realm/ObjectStore/object_store.cpp
```diff
@@ -355,6 +362,26 @@ void ObjectStore::create_tables(Group *group, Schema &target_schema, bool update
...
+void ObjectStore::remove_properties(Group *group, Schema &target_schema, std::vector<Property> to_delete) {
+    for (auto& target_object_schema : target_schema) {
+        TableRef table = table_for_object_type(group, target_object_schema.name);
+        ObjectSchema current_schema(group, target_object_schema.name);
+        size_t deleted = 0;
+        for (auto& current_prop : current_schema.properties) {
+            current_prop.table_column -= deleted;
+            for (auto& single_to_delete : to_delete) {
+                if (current_prop.name == single_to_delete.name &&
+                    current_prop.object_type == single_to_delete.object_type) {
```
tgoyne: `object_type` is the type of the linked object for object/array properties, not the type of the containing object.

jpsim: Fixed.

Realm/ObjectStore/object_store.cpp
```diff
@@ -486,6 +515,65 @@ void ObjectStore::delete_data_for_object(Group *group, StringData object_type) {
...
+    if (matching_schema == schema.end()) {
+        throw PropertyRenameMissingOldObjectTypeException(object_type);
+    }
+    Property *old_property = matching_schema->property_for_name(old_name);
+    if (old_property == nullptr) {
+        throw PropertyRenameMissingOldPropertyException(old_name, new_name);
+    }
+    TableRef table = table_for_object_type(group, object_type);
+    if (!table) {
+        throw PropertyRenameMissingNewObjectTypeException(object_type);
```
tgoyne: This line isn't reachable (it's effectively the same check as the `matching_schema == schema.end()` one above, but it throws a different exception...).

Realm/ObjectStore/object_store.cpp
```diff
@@ -486,6 +515,65 @@ void ObjectStore::delete_data_for_object(Group *group, StringData object_type) {
     }
 }
 
+void ObjectStore::rename_property(Group *group, Schema& passed_schema, StringData object_type, StringData old_name, StringData new_name) {
+    Schema schema = schema_from_group(group);
```
tgoyne: This could instead be:
```
    TableRef table = table_for_object_type(group, object_type);
    if (!table) {
        throw PropertyRenameMissingNewObjectTypeException(object_type);
    }
    ObjectSchema matching_schema(group, object_type);
```

which avoids the very heavyweight operation of reading the entire schema only to look at a single table.

Realm/ObjectStore/object_store.cpp
```diff
@@ -486,6 +515,65 @@ void ObjectStore::delete_data_for_object(Group *group, StringData object_type) {
...
+    TableRef table = table_for_object_type(group, object_type);
+    if (!table) {
+        throw PropertyRenameMissingNewObjectTypeException(object_type);
+    }
+    auto passed_object_schema = passed_schema.find(object_type);
+    if (passed_object_schema == passed_schema.end()) {
+        throw PropertyRenameMissingNewObjectTypeException(object_type);
+    }
+    Property *new_property = matching_schema->property_for_name(new_name);
+    if (new_property == nullptr) {
```
tgoyne: This block could use a comment about why this is a sensible thing to do.

Realm/ObjectStore/object_store.cpp
```diff
@@ -486,6 +515,65 @@ void ObjectStore::delete_data_for_object(Group *group, StringData object_type) {
...
+    size_t column_to_remove = new_property->table_column;
+    table->rename_column(old_property->table_column, new_name);
+    table->remove_column(column_to_remove);
+    // update table_column for each property after the one we just removed
+    for (auto& current_prop : passed_object_schema->properties) {
+        auto target_prop = matching_schema->property_for_name(current_prop.name);
+        current_prop.table_column = target_prop->table_column;
+    }
+    if (new_property->is_primary && !old_property->is_primary) {
+        set_primary_key_for_object(group, object_type, new_name);
```
tgoyne: PK is stored as the name, so I don't think this does anything.

Realm/ObjectStore/object_store.cpp
```diff
@@ -486,6 +515,65 @@ void ObjectStore::delete_data_for_object(Group *group, StringData object_type) {
...
+    }
+    if (new_property->requires_index() && !old_property->requires_index()) {
+        table->add_search_index(old_property->table_column);
```
tgoyne: Also do the inverse and remove indexes?

jpsim:
It'd be great to get this in the next release. Let me know if you have any concerns over this @tgoyne @bdash

tgoyne:
The error when a property is left with a name that does not appear in the final schema is misleading and confusing (it says that a migration is required, when a migration has just run).

Realm/RLMMigration.mm
```diff
@@ -146,4 +147,19 @@ - (BOOL)deleteDataForClassName:(NSString *)name {
...
+    RLMObjectSchema *objectSchema = [RLMObjectSchema objectSchemaForObjectStoreSchema:objectStoreSchema];
+    NSMutableArray *mutableObjectSchemas = [NSMutableArray arrayWithArray:_realm.schema.objectSchema];
+    [mutableObjectSchemas replaceObjectAtIndex:[mutableObjectSchemas indexOfObject:_realm.schema[className]]
+                                    withObject:objectSchema];
+    objectSchema.realm = _realm;
+    _realm.schema.objectSchema = [mutableObjectSchemas copy];
+    for (RLMProperty *property in objectSchema.properties) {
+        property.column = realm::ObjectStore::schema_from_group(_realm.group).find(className.UTF8String)->property_for_name(property.name.UTF8String)->table_column;
+    }
+    RLMClearAccessorCache();
```
tgoyne: It should not ever be necessary to call this function outside of code specifically testing the accessor caching logic.

jpsim: How else do you suggest doing this?

tgoyne: Well I don't know what problem it's trying to solve, so...

Realm/RLMMigration.mm
```diff
@@ -146,4 +147,19 @@ - (BOOL)deleteDataForClassName:(NSString *)name {
     return true;
 }
 
+- (void)renamePropertyForClass:(NSString *)className oldName:(NSString *)oldName newName:(NSString *)newName {
+    realm::ObjectStore::rename_property(_realm.group, *_realm->_realm->config().schema, className.UTF8String, oldName.UTF8String, newName.UTF8String);
+    auto objectStoreSchema = *realm::ObjectStore::schema_from_group(_realm.group).find(className.UTF8String);
```
tgoyne: `ObjectSchema objectStoreSchema(_realm.group, className.UTF8String);`

Realm/RLMMigration.mm
```diff
@@ -146,4 +147,19 @@ - (BOOL)deleteDataForClassName:(NSString *)name {
...
+    RLMObjectSchema *objectSchema = [RLMObjectSchema objectSchemaForObjectStoreSchema:objectStoreSchema];
+    NSMutableArray *mutableObjectSchemas = [NSMutableArray arrayWithArray:_realm.schema.objectSchema];
+    [mutableObjectSchemas replaceObjectAtIndex:[mutableObjectSchemas indexOfObject:_realm.schema[className]]
+                                    withObject:objectSchema];
+    objectSchema.realm = _realm;
+    _realm.schema.objectSchema = [mutableObjectSchemas copy];
+    for (RLMProperty *property in objectSchema.properties) {
+        property.column = realm::ObjectStore::schema_from_group(_realm.group).find(className.UTF8String)->property_for_name(property.name.UTF8String)->table_column;
```
tgoyne: `objectStoreSchema.property_for_name(property.name.UTF8String)->table_column`

Realm/RLMMigration.mm
jpsim: You're right, this isn't needed anymore, I had just added this temporarily while debugging and forgot to remove it.

jpsim:
> The error when a property is left with a name that does not appear in the final schema is misleading and confusing (it says that a migration is required, when a migration has just run).

Fixed. Now says "Renamed property '...' is not in the latest model.".

tgoyne:
👍

PR #3361, opened 2016-03-22
make Swift 2.2 new default Swift version

jpsim:
This just moves our project infrastructure to use Swift 2.2 as the default Swift version. This doesn't modify any of the Swift code itself to address deprecation warnings. That may be done at a later time. This PR is mostly to test against our updated CI infrastructure.


jpsim:
This is blocked on getting device tests working with Xcode 7.3.

When running iOS simulator tests from Xcode 7.3 with the `master` branch, Xcode fails before the tests can run, prompting with an `Invalid Argument` message (no further information is provided by Xcode)
Removing the `TestApp` host app target allows simulator tests to run, which is what I've done temporarily in this PR. But this host app is necessary to run unit tests on-device.
So we need to find a way to be able to run both device tests and simulator tests in Xcode 7.3, and we haven’t found one yet. Any help appreciated!

bdash:
Device and simulator tests are now working with Xcode 7.3.

bdash:
@jpsim, can you please take a look at the changes I added. Your earlier changes all look good to me.

tgoyne:
:+1:

jpsim:
:+1: thanks for all of this @bdash!

PR #3369, opened 2016-03-24
Adds support for async writes

### LINKED ISSUE #3136: Support for asynchronous writes
What are the current thoughts on adding a shortcut syntax for doing async transactions (where operations happen on a background thread)?

Here is the current recommended way of doing it manually (from the docs):

``` swift
// Query from any thread
dispatch_async(dispatch_queue_create("background", nil)) {
  let realm = try! Realm()
  let theDog = realm.objects(Dog).filter("age == 1").first
  try! realm.write {
    theDog!.age = 3
  }
}
```

It is very flexible, but it does open up the opportunity to accidentally forgetting to get a new Realm, and when working with non-default realms, you have to be careful to pass it the right configuration (also, you probably want it wrapped in an autoreleasepool, like in some of the later examples).

Would it make sense to offer a simpler version that ensured that you get the right realm (and that it get released afterwards)? Maybe something like this:

``` swift
// Query and update on background thread
realm.async { realm in
  let theDog = realm.objects(Dog).filter("age == 1").first
  try! realm.write {
    theDog!.age = 3
  }
}
```

Or even simpler if you know that you need to write:

``` swift
// Write via background thread
realm.async_write { realm in
  let theDog = realm.objects(Dog).filter("age == 1").first
  theDog!.age = 3
}
```

This kind of shortcut syntax would obviously need a bit of thought in regard to how you would then handle errors (can't open realm, commit fails), but offering the option to supply a closure called on error might satisfy that.

bigfish24:
1. Block with a thread local RLMRealm instance
2. Block with a thread local RLMObject instance (handed over from calling thread)
3. Block with a thread local RLMArray/RLMResults (handed over from the calling thread)

Tests for each. I am more dubious of the implementation for RLMResults, since the difference between all the various backing objects.

Fixes #3136

### DIFF
--- a/Realm/ObjectStore/shared_realm.hpp
+++ b/Realm/ObjectStore/shared_realm.hpp
@@ -158,6 +158,8 @@ namespace realm {
 
         // FIXME private
         Group *read_group();
+        
+        SharedGroup& shared_group();
     };
 
     class RealmFileException : public std::runtime_error {
--- a/Realm/RLMObject.mm
+++ b/Realm/RLMObject.mm
@@ -24,6 +24,11 @@
 #import "RLMRealm_Private.hpp"
 #import "RLMQueryUtil.hpp"
 
+#include <future>
+#include <realm/group_shared.hpp>
+#import <realm/link_view.hpp>
+#import <realm/row.hpp>
+
 // We declare things in RLMObject which are actually implemented in RLMObjectBase
 // for documentation's sake, which leads to -Wunimplemented-method warnings.
 // Other alternatives to this would be to disable -Wunimplemented-method for this
--- a/Realm/RLMObject.mm
+++ b/Realm/RLMObject.mm
@@ -24,11 +24,6 @@
 #import "RLMRealm_Private.hpp"
 #import "RLMQueryUtil.hpp"
 
-#include <future>
-#include <realm/group_shared.hpp>
-#import <realm/link_view.hpp>
-#import <realm/row.hpp>
-
 // We declare things in RLMObject which are actually implemented in RLMObjectBase
 // for documentation's sake, which leads to -Wunimplemented-method warnings.
 // Other alternatives to this would be to disable -Wunimplemented-method for this
--- a/Realm/RLMRealm.h
+++ b/Realm/RLMRealm.h
@@ -18,6 +18,8 @@
 
 #import <Foundation/Foundation.h>
 #import <Realm/RLMDefines.h>
+#import <Realm/RLMCollection.h>
+
 
 @class RLMRealmConfiguration, RLMObject, RLMSchema, RLMMigration, RLMNotificationToken;
 
@@ -500,6 +502,22 @@ typedef void (^RLMMigrationBlock)(RLMMigration *migration, uint64_t oldSchemaVer
  @see                 RLMMigration
  */
 + (NSError *)migrateRealm:(RLMRealmConfiguration *)configuration;
+
+#pragma mark - Async Writes
+
+typedef void (^RLMAsyncWriteBlock)(RLMRealm *realm);
+
+typedef void (^RLMAsyncWriteObjectBlock)(RLMRealm *realm, id object);
+
+typedef void (^RLMAsyncWriteCollectionBlock)(RLMRealm *realm, id<RLMCollection> collection);
+
+typedef void (^RLMCompletionBlock)(NSError *error);
+
+- (void)writeAsyncWithBlock:(RLMAsyncWriteBlock)block completion:(RLMCompletionBlock)completion;
+
+- (void)writeObjectAsync:(RLMObject *)object withBlock:(RLMAsyncWriteObjectBlock)block completion:(RLMCompletionBlock)completion;
+
+- (void)writeCollectionAsync:(id<RLMCollection>)collection withBlock:(RLMAsyncWriteCollectionBlock)block completion:(RLMCompletionBlock)completion;
 
 @end
 
--- a/Realm/RLMRealm.h
+++ b/Realm/RLMRealm.h
@@ -18,10 +18,10 @@
 
 #import <Foundation/Foundation.h>
 #import <Realm/RLMDefines.h>
-#import <Realm/RLMCollection.h>
-
 
 @class RLMRealmConfiguration, RLMObject, RLMSchema, RLMMigration, RLMNotificationToken;
+
+@protocol RLMCollection;
 
 RLM_ASSUME_NONNULL_BEGIN
 
--- a/Realm/RLMRealm.mm
+++ b/Realm/RLMRealm.mm
@@ -45,8 +45,7 @@
 #include <realm/disable_sync_to_disk.hpp>
 #include <realm/version.hpp>
 
-#include <future>
-#include <realm/group_shared.hpp>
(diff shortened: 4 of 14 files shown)


bigfish24:
@jpsim, @tgoyne and @bdash I wanted to get the ball rolling on this and leave it up to the smarter minds to shape direction.

Questions I pose to you:
1. What do you think about the API design?
2. Am I missing something for `RLMResults`?
3. I assume this can be refactored further into the object store?

Haven't touched Swift since I assume more changes will come.

Realm/ObjectStore/shared_realm.hpp
```diff
@@ -158,6 +158,8 @@ namespace realm {
 
         // FIXME private
         Group *read_group();
+        
+        SharedGroup& shared_group();
```
tgoyne: The Realm's shared group should not be publicly exposed.

Realm/RLMObject.mm
```diff
@@ -24,6 +24,11 @@
 #import "RLMRealm_Private.hpp"
 #import "RLMQueryUtil.hpp"
 
+#include <future>
+#include <realm/group_shared.hpp>
```
tgoyne: Use #import for all headers in obj-c++ files.

Realm/RLMRealm.h
```diff
@@ -18,6 +18,8 @@
 
 #import <Foundation/Foundation.h>
 #import <Realm/RLMDefines.h>
+#import <Realm/RLMCollection.h>
```
tgoyne: Forward-declare what you need from that header rather than including it.

Realm/RLMRealm.mm
```diff
@@ -780,4 +787,158 @@ - (void)detachAllEnumerators {
...
+                completion(error);
+            });
+            return;
+        }
+        
+        [bgRealm transactionWithBlock:^{
+            block(bgRealm);
+        } error:&error];
+        
+        dispatch_async(dispatch_get_main_queue(), ^{
```
tgoyne: The completion block dispatching needs to be done via the Realm notification system and not dispatch_async to work with async queries and fine-grained notifications.

Realm/RLMRealm.mm
```diff
@@ -780,4 +787,158 @@ - (void)detachAllEnumerators {
...
+        if (error) {
+            dispatch_async(dispatch_get_main_queue(), ^{
+                completion(error);
+            });
+            return;
+        }
+        
+        newRealm->_realm->read_group();
+        auto& new_sg = newRealm->_realm->shared_group();
+        std::unique_ptr<realm::Row> new_row( new_sg.import_from_handover(move(handover_row)) );
```
tgoyne: This will fail for any nontrivial cases of writes being made on multiple threads (e.g. if asyncWrite is called twice in a row...) because objects can only be handed over to a sharedgroup at the same version as the source one.

tgoyne:
API-wise I'd be inclined to have just a single method that takes an `id` which can be an RLMObject, RLMArray, RLMResults, or fast-enumerable of any of those. Being able to pass an NSArray of specific objects to use on the background thread is probably an important use-case, and having ten different methods would get unwieldy.

Realm/RLMRealm.mm
bigfish24: adjusted to fire the completion block off the notification, let me know if it looks right

Realm/RLMRealm.mm
bigfish24: Yeah @bdash mentioned this. I am not certain how to enforce the versions.

Realm/ObjectStore/shared_realm.hpp
bigfish24: I guess the object store `realm` could have methods to generate the intermediate handover object and another to accept this intermediate object and return the actual handed over object.

Realm/RLMRealm.mm
```diff
@@ -780,4 +786,165 @@ - (void)detachAllEnumerators {
...
+
+- (void)writeAsyncWithBlock:(RLMAsyncWriteBlock)block completion:(RLMCompletionBlock)completion {
+    RLMRealmConfiguration *config = self.configuration;
+    
+    __block NSError *error;
+    __block BOOL didComplete;
+    
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wunused-parameter"
+    __block RLMNotificationToken *token = [self addNotificationBlock:^(NSString *notification, RLMRealm *realm) {
```
JadenGeller: This shouldn't need to be a `__block` parameter since the value of the pointer isn't being mutated inside the block, right?

tgoyne: Without __block it captures the value of the variable at the time the block is created (i.e. nil, since the block is created before addNotificationBlock is called).

JadenGeller:
Are there any fundamental problems with the implementation? It looks pretty decent to me, so I'm trying to understand if this was never merged because deciding on an API was difficult or because of some other issue.

Realm/RLMRealm.mm
JadenGeller: Oh! That's subtle, thanks for explaining!

bigfish24:
@JadenGeller this implementation doesn't actually work in a multi-threaded environment because at the time it wasn't straightforward to "pin" a version of the shared group such that when the async block runs that version is still around and hasn't advanced due to another thread. This is now possible via https://github.com/realm/realm-core/pull/1626 so you would need to incorporate the pinning functionality.

bigfish24:
Firing the completion block also isn't too elegant, wondering if there is a better way to do this cc @tgoyne?

Realm/ObjectStore/shared_realm.hpp
bigfish24: @JadenGeller note @tgoyne comment, so the handover logic really should be moved to the object store so that we don't make `SharedGroup` public from binding perspective.

bigfish24:
Overall though this PR does demonstrate the basic semantics of using core's handover functionality, but it was never merged primarily since it doesn't work in multithread scenario, but secondarily because we never really discussed what the API should look like.

tgoyne:
A rough idea of what the object-store level API could look like:

```
class Realm {
    std::shared_ptr<HandoverThing> package_for_handover(std::vector<std::variant<List&, Results&, Row&>> objects_to_hand_over);
    std::vector<std::variant<List, Results, Row>> import_from_handover(HandoverThing& handover);
};
```

`HandoverThing` is an opaque object defined within `shared_realm.cpp` which stores a Realm::Config object along with the actual Handover objects, pins the version those objects are from when it is created, and unpins that version either when it is destroyed or consumed (by passing it to `realm_from_handover()`).

`import_from_handover()` ends the Realm's read transaction, begins a new read transaction at the HandoverThing's version, and then consumes the handover object and returns an array of the objects imported into the current Realm.

The code in RLMRealm would loop over the passed-in objects to be handed over adding the ObjectStore members to a vector, passes that to `package_for_handover()`, does a `dispatch_aync()` to a background queue, gets an RLMRealm for the current thread, calls `import_from_handover()`, constructs obj-c/swift wrappers for those objects, begins a write transaction, and calls the block with the array of objects.

bdash:
Has this been superseded by #3828? If so, should this be closed?

PR #3370, opened 2016-03-24
iOS Playground demoing RealmSwift

icanzilb:
This PR adds a Playground to the iOS swift example app - it shows everything one needs to know to get started with Realm in an interactive fashion.


jpsim:
Thanks so much for this @icanzilb! I just made a few changes and pushed #3385 and will continue there.

PR #3379, opened 2016-03-25
Fix -[RLMResults indexOfObjectWithPredicate:] for RLMResults instances that were created by filtering an RLMArray

### LINKED ISSUE #3319: Results.indexOf crashing app
I'm trying to get the index of the first occurrence of `Predicate` in a `Results` object using `indexOf(predicate: NSPredicate)` but it's crashing every time. 

Model Object:
```Swift
class Service: Object {
	dynamic var timestamp: NSDate?
	dynamic var server: Server?
	dynamic var backgroundMonitoring = true
	
	dynamic var name: String = ""
	dynamic var displayName: String = ""
	dynamic var installed: Bool = false
	dynamic var enabled: Bool = false
	dynamic var monitored: Bool = false
	dynamic var running: Bool = false
	
	override static func indexedProperties() -> [String] {
		return [
			"timestamp",
			"backgroundMonitoring",
			"name",
			"displayName",
			"installed",
			"enabled",
			"monitored",
			"running"
		]
	}
}
```
Code that triggers error:
```Swift
// This loop is just to print index/values of the Results to make sure the data is good
for (key, item) in data.enumerate() {
	print(key, (item as! Service).displayName)
}

let predicate = NSPredicate(format: "displayName BEGINSWITH[c] 'P'")
if let index = data.indexOf(predicate) { // This is the line that errors out
	return index
}
```
Console output of the For/In Loop .. indexOf should return 18
```
0 Apache SpamAssassin™
1 Apache Web Server
2 cPanel Daemon
3 cPanel DAV Daemon
4 cPanel DNS (Domain Name System) Admin Cache
5 cPanel Greylisting Daemon
6 cPanel Log and Bandwidth Processor
7 cPHulk Daemon
8 Cron Daemon
9 DNS (Domain Name System) Server
10 Exim Mail Server
11 Exim Mail Server (on another port)
12 FTP (File Transfer Protocol) Server
13 IMAP (Internet Mail Access Protocol) Server
14 IP (Internet Protocol) Aliases
15 Mailman
16 MySQL Server
17 Name Service Cache Daemon
18 Passive OS (Operating System) Fingerprinting Daemon
19 PHP-FPM service for cPanel Daemons
20 POP3 (Post Office Protocol 3) Server
21 rsyslog System Logger Daemon
22 SSH (Secure Shell) Daemon
23 TailWatch Daemon
24 TaskQueue Processor
```

indexOf goes to RealmSwift/Results.swift
```Swift
public func indexOf(predicate: NSPredicate) -> Int? {
        return notFoundToNil(rlmResults.indexOfObjectWithPredicate(predicate)) // Triggers here
 }
```
And finally in RealmSwift/Util.swift
```Swift
internal func notFoundToNil(index: UInt) -> Int? {
    if index == UInt(NSNotFound) {
        return nil
    }
    return Int(index) // Here
}
```
Stack trace
```
#0	0x000000010152c224 in notFoundToNil(UInt) -> Int? at /Users/REMOVED/xcode/iOS/REMOVED/Pods/RealmSwift/RealmSwift/Util.swift:28
#1	0x0000000101528634 in Results.indexOf<A where ...>(NSPredicate) -> Int? at /Users/REMOVED/xcode/iOS/REMOVED/Pods/RealmSwift/RealmSwift/Results.swift:143
#2	0x00000001001d3688 in TableViewDataSource.tableView<A where ...>(UITableView, sectionForSectionIndexTitle : String, atIndex : Int) -> Int at /Users/REMOVED/xcode/iOS/REMOVED/REMOVED/TableViewDataSource.swift:120
#3	0x00000001001d3850 in @objc TableViewDataSource.tableView<A where ...>(UITableView, sectionForSectionIndexTitle : String, atIndex : Int) -> Int ()
#4	0x0000000185fe8308 in -[UITableView _sectionIndexChangedToIndex:title:] ()
#5	0x0000000185eec20c in -[UITableView _sectionIndexChanged:] ()
#6	0x0000000185cdbe50 in -[UIApplication sendAction:to:from:forEvent:] ()
#7	0x0000000185cdbdcc in -[UIControl sendAction:to:forEvent:] ()
#8	0x0000000185cc3a88 in -[UIControl _sendActionsForEvents:withEvent:] ()
#9	0x0000000185eec138 in -[UITableViewIndex _selectSectionForTouch:withEvent:] ()
#10	0x0000000185eebf18 in -[UITableViewIndex beginTrackingWithTouch:withEvent:] ()
#11	0x0000000185ce4544 in -[UIControl touchesBegan:withEvent:] ()
#12	0x0000000185cdb168 in -[UIWindow _sendTouchesForEvent:] ()
#13	0x0000000185cd3e30 in -[UIWindow sendEvent:] ()
#14	0x0000000185ca44cc in -[UIApplication sendEvent:] ()
#15	0x0000000185ca2794 in _UIApplicationHandleEventQueue ()
#16	0x0000000180f68efc in __CFRUNLOOP_IS_CALLING_OUT_TO_A_SOURCE0_PERFORM_FUNCTION__ ()
#17	0x0000000180f68990 in __CFRunLoopDoSources0 ()
#18	0x0000000180f66690 in __CFRunLoopRun ()
#19	0x0000000180e95680 in CFRunLoopRunSpecific ()
#20	0x00000001823a4088 in GSEventRunModal ()
#21	0x0000000185d0cd90 in UIApplicationMain ()
#22	0x0000000100051b3c in main at /Users/REMOVED/xcode/iOS/REMOVED/REMOVED/AppDelegate.swift:19
#23	0x0000000180a368b8 in start ()

```

```
ProductName:	Mac OS X
ProductVersion:	10.11.3
BuildVersion:	15D21

Xcode 7.2.1
Build version 7C1002

pod is /usr/local/bin/pod
pod is /usr/local/bin/pod
0.39.0
Realm (0.98.4)
RealmSwift (0.98.4)

bash is /bin/bash

carthage is /usr/local/bin/carthage
carthage is /usr/local/bin/carthage
0.9.2
(not in use here)

git is /usr/local/git/bin/git
git is /usr/bin/git
git is /usr/local/git/bin/git
git version 2.5.0
```

bdash:
Fixes #3319.

/cc @tgoyne 

This is working around a core issue (realm/realm-core#1565). It gives correct results for calls to `indexOfObjectWithPredicate:` at the expense of making them do more work (potentially a lot more), even in cases that aren't affected by the bug. I'm not sure if that's a good tradeoff for us to ship, but it may prove useful to anyone that is affected by #3319.

### DIFF
--- a/Realm/RLMResults.mm
+++ b/Realm/RLMResults.mm
@@ -290,7 +290,7 @@ static inline void RLMResultsValidateInWriteTransaction(__unsafe_unretained RLMR
     // FIXME: We're only looking for a single object so we'd like to be able to use `Query::find`
     // for this, but as of core v0.97.1 it gives incorrect results if the query is restricted
     // to a link view (<https://github.com/realm/realm-core/issues/1565>).
-    auto table_view = query.find_all();
+    auto table_view = query.find_all(0, -1, 1);
     if (!table_view.size()) {
         return NSNotFound;
     }
--- a/Realm/RLMResults.mm
+++ b/Realm/RLMResults.mm
@@ -286,11 +286,15 @@ static inline void RLMResultsValidateInWriteTransaction(__unsafe_unretained RLMR
 
     Query query = translateErrors([&] { return _results.get_query(); });
     RLMUpdateQueryWithPredicate(&query, predicate, _realm.schema, _objectSchema);
-    size_t index = query.find();
-    if (index == realm::not_found) {
+
+    // FIXME: We're only looking for a single object so we'd like to be able to use `Query::find`
+    // for this, but as of core v0.97.1 it gives incorrect results if the query is restricted
+    // to a link view (<https://github.com/realm/realm-core/issues/1565>).
+    auto table_view = query.find_all();
+    if (!table_view.size()) {
         return NSNotFound;
     }
-    return _results.index_of(index);
+    return _results.index_of(table_view.get_source_ndx(0));
 }
 
 - (id)objectAtIndex:(NSUInteger)index {
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,8 @@ x.x.x Release notes (yyyy-MM-dd)
 
 * Mark further initializers in Objective-C as NS_DESIGNATED_INITIALIZER to prevent that these aren't
   correctly defined in Swift Object subclasses, which don't qualify for auto-inheriting the required initializers.
+* `-[RLMResults indexOfObjectWithPredicate:]` now returns correct results
+  for `RLMResults` instances that were created by filtering an `RLMArray`.
 
 0.98.6 Release notes (2016-03-25)
 =============================================================
--- a/Realm/Tests/ArrayPropertyTests.m
+++ b/Realm/Tests/ArrayPropertyTests.m
@@ -475,11 +475,12 @@
     EmployeeObject *po1 = [EmployeeObject createInRealm:realm withValue:@{@"name": @"Joe",  @"age": @40, @"hired": @YES}];
     [EmployeeObject createInRealm:realm withValue:@{@"name": @"John", @"age": @30, @"hired": @NO}];
     EmployeeObject *po3 = [EmployeeObject createInRealm:realm withValue:@{@"name": @"Jill", @"age": @25, @"hired": @YES}];
+    EmployeeObject *po4 = [EmployeeObject createInRealm:realm withValue:@{@"name": @"Bill", @"age": @55, @"hired": @YES}];
 
     // create company
     CompanyObject *company = [[CompanyObject alloc] init];
     company.name = @"name";
-    [company.employees addObjects:@[po3, po1]];
+    [company.employees addObjects:@[po3, po1, po4]];
 
     // test standalone
     XCTAssertEqual(0U, [company.employees indexOfObjectWhere:@"name = 'Jill'"]);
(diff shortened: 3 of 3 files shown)


bdash:
We could potentially limit the workaround to only cases that would be affected by the bug, but that would require knowing if a `Query` is restricted by a view. AFAICT that information is not exposed.

Realm/RLMResults.mm
```diff
@@ -286,11 +286,15 @@ - (NSUInteger)indexOfObjectWithPredicate:(NSPredicate *)predicate {
 
     Query query = translateErrors([&] { return _results.get_query(); });
     RLMUpdateQueryWithPredicate(&query, predicate, _realm.schema, _objectSchema);
-    size_t index = query.find();
-    if (index == realm::not_found) {
+
+    // FIXME: We're only looking for a single object so we'd like to be able to use `Query::find`
+    // for this, but as of core v0.97.1 it gives incorrect results if the query is restricted
+    // to a link view (<https://github.com/realm/realm-core/issues/1565>).
+    auto table_view = query.find_all();
```
tgoyne: `find_all(0, -1, 1)` would make the overhead over `find()` fairly small.

tgoyne:
#3359 makes Results track the LinkView it's derived from so I could extract that change if needed, but a find_all() limited to one match is probably fine.

Realm/RLMResults.mm
bdash: Oh, good point! That'd just give us the overhead of working with the `TableView`, which at least will be fixed rather than dependent on the number of rows we're searching.

tgoyne:
Actually, if I understand things correctly I think we could just do the following with the current behavior of `find()`:

    auto tv = results.get_tableview();
    Query query = tv.get_table()->where(tv);
    RLMUpdateQueryWithPredicate(&query, predicate, _realm.schema, _objectSchema);
    size_t index = query.find();
    return index == realm::not_found ? NSNotFound : index;

This would actually be faster than the current code for everything but Table-backed Results because index_of() is O(N) (could special-case Table-backed results to make it just as fast)

bdash:
I'm not sure how I feel about optimizing based on `find`s current bogus behavior. That would mean regressing performance once the core issue is fixed.

tgoyne:
:+1:

PR #3395, opened 2016-03-29
Remove the manual storage of observationInfo on RLMObject

### LINKED ISSUE #3346: Storing the kvoInfo pointer in RLMObservationInfo breaks KVO with associated objects
KVO supports the ability to unregister observers at the point where associated objects are cleared on the observed object (as part of dealloc). This trick is used by [postmates/PMKVObserver](https://github.com/postmates/PMKVObserver) to implement safe automatic unregistering of observers.

However, doing this requires that the opaque pointer returned by `-observationInfo` is still valid at the point where associated objects are released. Under normal circumstances this works perfectly fine. But `RLMObjectBase` breaks this functionality. `RLMObjectBase` overrides `-observationInfo` and `-setObservationInfo:` to store the opaque pointer as a member variable of an ivar it has of type `std::unique_ptr<RLMObservationInfo>`. Unfortunately, when using Obj-C++, the destructors of any C++ member variables are invoked prior to the associated objects being cleared, which means the `std::unique_ptr` is destructed, which resets it back to the null pointer. This means that by the time associated objects are cleared, `-[RLMObjectBase observationInfo]` is guaranteed to return `NULL`. And this causes any calls to `-removeObserver:forKeyPath:context:` called by the associated objects being released to trigger an exception as KVO believes that the observer is not registered. And if the associated object somehow knew this was going to happen and skips the call to `-removeObserver:...`, the KVO machinery will instead throw an exception because the `RLMObjectBase` was deallocated without unregistering all observers (it grabs the observation info prior to the object being deallocated and then detects that the observation info still has registered observers after the object was deallocated).

The simple way to fix all of this is to remove the `void *kvoInfo` member variable from `RLMObservationInfo` and instead to just add it as a separate ivar to `RLMObjectBase`. There doesn't seem to be any benefit to keeping it on the `RLMObservationInfo` and this will fix the behavior of `-observationInfo` while the associated objects are being cleared.

This issue was originally filed as postmates/PMKVObserver#9, at which point I started reading the Realm source to figure out what's going on.

tgoyne:
This was originally done for the sake of awful things that thankfully turned out to be unneeded and was left only as a free minor performance optimization, but in practice it doesn't help and it breaks removing observers via associated objects.

Fixes #3346,

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,8 @@ x.x.x Release notes (yyyy-MM-dd)
   correctly defined in Swift Object subclasses, which don't qualify for auto-inheriting the required initializers.
 * `-[RLMResults indexOfObjectWithPredicate:]` now returns correct results
   for `RLMResults` instances that were created by filtering an `RLMArray`.
+* Adjust how RLMObjects are destroyed in order to support using an associated
+  object on an RLMObject to remove KVO observers from that RLMObject.
 
 0.98.6 Release notes (2016-03-25)
 =============================================================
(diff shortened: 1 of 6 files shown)


bdash:
:+1:

PR #3399, opened 2016-03-29
deprecate duplicate realm configuration properties

### LINKED ISSUE #3396: Remove redundant properties on RLMRealm already present on RLMConfiguration
`path`, `readonly` & `schema` are on both `RLMRealm` and `RLMConfiguration`. Those properties should be removed from `RLMRealm`.

jpsim:
Closes #3396. /cc @tgoyne @bdash 

I also think it would make sense to move `RLMRealm.schema` to `RLMRealmConfiguration.schema` (renaming `customSchema` to `schema` in the process), but that's a bit more involved so I want to make sure we agree it's a good idea before I go down that path.

To do:

- [x] Determine if we should move `RLMRealm.schema` to `RLMRealmConfiguration.schema`
- [x] Reflect these deprecations in Realm Swift

### DIFF
--- a/Realm/RLMRealmConfiguration_Private.hpp
+++ b/Realm/RLMRealmConfiguration_Private.hpp
@@ -1,0 +1,24 @@
+////////////////////////////////////////////////////////////////////////////
+//
+// Copyright 2014 Realm Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+////////////////////////////////////////////////////////////////////////////
+
+#import "RLMRealmConfiguration_Private.h"
+#include "shared_realm.hpp"
+
+@interface RLMRealmConfiguration ()
+- (realm::Realm::Config&)config;
+@end
--- a/Realm/RLMRealmConfiguration_Private.hpp
+++ b/Realm/RLMRealmConfiguration_Private.hpp
@@ -17,7 +17,7 @@
 ////////////////////////////////////////////////////////////////////////////
 
 #import "RLMRealmConfiguration_Private.h"
-#include "shared_realm.hpp"
+#import "shared_realm.hpp"
 
 @interface RLMRealmConfiguration ()
 - (realm::Realm::Config&)config;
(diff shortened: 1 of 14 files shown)


tgoyne:
The internal uses of `readOnly` inside RLMRealm need to use `_realm.config().read_only`, as `-[RLMRealm configuration]` creates a new `RLMRealmConfiguration` object, which is not at all inexpensive.

Creating Realms with custom schemas is not functionality I think we should expose (it's potentially a big can of worms that no one has even asked for), and having a read-only property on a class intended primarily for passing in user-supplied parameters would be very weird.

jpsim:
Please review @tgoyne @bdash

jpsim:
@bdash comments explaining use of deprecated property added

bdash:
It could be better to use `realm.config.config.path` for the cases where it's necessary to check the on-disk path used by in-memory realms, rather than continuing to use the deprecated method. I don't feel strongly about that though.

bdash:
👍

jpsim:
Now using object store config to validate the underlying location for in-memory Realms in unit tests rather than a deprecated property.

Realm/RLMRealmConfiguration_Private.hpp
```diff
@@ -0,0 +1,24 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+////////////////////////////////////////////////////////////////////////////
+
+#import "RLMRealmConfiguration_Private.h"
+#include "shared_realm.hpp"
```
tgoyne: #import

Realm/Tests/RealmConfigurationTests.mm
```diff
@@ -118,26 +117,27 @@ - (void)testDefaultConfiugrationUsesValueSemantics {
...
+        XCTAssertTrue([realmPath hasSuffix:@"/default"]);
+        XCTAssertTrue([realmPath hasPrefix:NSTemporaryDirectory()]);
     }
 
     config.schemaVersion = 1;
     RLMRealmConfiguration.defaultConfiguration = config;
     @autoreleasepool {
-        RLMRealm *realm = RLMRealm.defaultRealm;
-        XCTAssertEqual(1U, [RLMRealm schemaVersionAtPath:realm.path error:nil]);
+        __unused RLMRealm *realm = RLMRealm.defaultRealm;
```
tgoyne: No point in leaving the variable if it's unused.

RealmSwift-swift2.0/Tests/PerformanceTests.swift
```diff
@@ -361,7 +361,7 @@ class SwiftPerformanceTests: TestCase {
                 }
             }
         }
-        _ = realm.path
+        _ = realm
```
tgoyne: This needs to actually access something on the object which the compiler can't prove doesn't have side effects.

Realm/Tests/RealmConfigurationTests.mm
jpsim: actually it was necessary to access the Realm here to cause the migration to happen, but I've refactored this to no longer have an unused variable.

jpsim:
Addressed your feedback @tgoyne

jpsim:
Any final feedback since my last modifications @tgoyne?

tgoyne:
👍

PR #3401, opened 2016-03-29
improve property type exceptions to include the property name

### LINKED ISSUE #3324: More informative exception message for unsupported property types
## Goals

If I'm creating a Realm object and it has a property of unsupported type (like NSURL) the init throws. Since I generally don't expect this to happen in production I don't catch this exception and I get the error description in the console.

The description however doesn't tell me which property is of an unsupported type, which I'd expect so I can quickly fix the problem.

## Expected Results

I'd expect the property name in the error message - something like "The property imageURL is of type NSURL, which is not supported..." in the error description.

## Actual Results

The error message does not include the name of the offending property:

*** Terminating app due to uncaught exception 'RLMException', reason: ''NSURL' is not supported as an RLMObject property. All properties must be primitives, NSString, NSDate, NSData, RLMArray, or subclasses of RLMObject. See https://realm.io/docs/objc/latest/api/Classes/RLMObject.html for more information.'

## Steps to Reproduce

Define a property of unsupported type in a realm object, e.g.:

dynamic var imageUrl: NSURL?

Realm 0.98.3 / Xcode 7.2.1

jpsim:
Closes #3324. /cc @tgoyne @bdash @icanzilb

### DIFF
--- a/Realm/RLMProperty.mm
+++ b/Realm/RLMProperty.mm
@@ -109,7 +109,7 @@ BOOL RLMPropertyTypeIsNumeric(RLMPropertyType propertyType) {
 }
 
 // determine RLMPropertyType from objc code - returns true if valid type was found/set
--(BOOL)setTypeFromRawType {
+- (BOOL)setTypeFromRawType {
     const char *code = _objcRawType.UTF8String;
     _objcType = *code;    // first char of type attr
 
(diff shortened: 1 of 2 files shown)


Realm/RLMProperty.mm
```diff
@@ -203,7 +205,9 @@ -(BOOL)setTypeFromRawType {
                 // verify type
                 Class cls = [RLMSchema classForString:className];
                 if (!cls) {
-                    @throw RLMException(@"'%@' is not supported as an RLMObject property. All properties must be primitives, NSString, NSDate, NSData, RLMArray, or subclasses of RLMObject. See https://realm.io/docs/objc/latest/api/Classes/RLMObject.html for more information.", className);
+                    @throw RLMException(@"Property '%@' is declared as '%@', which is not a supported RLMObject property type. "
+                                        @"All properties must be primitives, NSString, NSDate, NSData, RLMArray, or subclasses of RLMObject. "
```
tgoyne: Not a new issue, but NSNumber is missing from this list.

jpsim: Good point. Updated.

tgoyne:
:+1:

PR #3402, opened 2016-03-30
[build.sh] download core files using local cache first

6david9:
When the network is bad, it hard to build project success. The building process is pending on download core files. I have modify the build.sh file to using local cache first when downloading "realm-core-${version}.tar.bz2".

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -289,17 +289,25 @@ download_core() {
     mkdir -p "${TMP_DIR}"
     CORE_TMP_TAR="${TMP_DIR}/core-${REALM_CORE_VERSION}.tar.bz2.tmp"
     CORE_TAR="${TMP_DIR}/core-${REALM_CORE_VERSION}.tar.bz2"
+    CORE_CACHE_NAME="realm-core-${REALM_CORE_VERSION}.tar.bz2"
+
     if [ ! -f "${CORE_TAR}" ]; then
-        local CORE_URL="https://static.realm.io/downloads/core/realm-core-${REALM_CORE_VERSION}.tar.bz2"
-        set +e # temporarily disable immediate exit
-        local ERROR # sweeps the exit code unless declared separately
-        ERROR=$(curl --fail --silent --show-error --location "$CORE_URL" --output "${CORE_TMP_TAR}" 2>&1 >/dev/null)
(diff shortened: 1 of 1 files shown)


jpsim:
Thanks for the contribution @6david9 and for accepting our CLA!

Are you aware that if you have a `core/` directory (not a symlink), `build.sh` will use what's in there rather than attempt to download a core build from the internet? Would that approach fulfill your needs rather than this?

I'm wondering what this allows you to do that isn't already possible with the `core/` directory approach.

6david9:
@jpsim from build.sh line 349:

    if [ -d core -a -d ../realm-core -a ! -L core ]; then
        using local cache

   what does `../realm-core` do? I don't have `realm-core` in parent folder, so it alway not using local cache.

jpsim:
That's not what I meant. I mean a `core` directory in the root of the Realm repo with the unarchived contents of the tarball download.

bdash:
The logic in build.sh that @6david9 points out means that we only uses a `core` directory if `../realm-core` also exists and is a directory.

6david9:
@bdash Yes. I have try @jpsim 's way, but I found the build.sh maybe have some issue. 
I think this pr should be closed.

bdash:
Note that you can achieve what you're after by ensuring the core tarball exists in `$TMPDIR/core_bin/`. The only downside is that as this path is within `$TMPDIR`, it will need to be replaced after a reboot.

6david9:
@bdash In my mac, echo $TMPDIR is 

    echo $TMPDIR
    /var/folders/07/hz7q5y6j3hgd709yk1scvhw80000gn/T/

is not `/tmp` and it may be different each time. so is hard for me to push tarball to `$TMPDIR/core_bin/`.

bdash:
I don't understand why that's a problem. A script can fetch the value of `$TMPDIR` to determine where to copy the core tarball.

6david9:
Why not make it more friendly to people who have not read the whole of build.sh? I think it's easier to tell people who face the same situation just download the tarball to the root of repo or move unarchive content to the root of repo than tell him to write another script to avhieve this.

6david9:
if you think i should write another script to solve this problem, close it.

PR #3411, opened 2016-04-06
Add Launch Screen Storyboards to Each Example App

TimOliver:
I noticed this while I was making an example app to demonstrate the fine-grained notification API.

This PR adds a `LaunchScreen.storyboard` file and the appropriate configurations to every Objective-C and Swift example app that we have. This allows the apps to properly accomodate the various iOS device screen sizes, and does away with any blurry upscaling we may have previously had.

/cc @jpsim @mrackwitz


mrackwitz:
Looks good! :+1:

jpsim:
Instead of having 16 different storyboards checked into source control, it'd be nice to just reuse one. Fine if they all say `Realm Examples` instead of their individual names IMO.

TimOliver:
Fair enough! I've consolidated them down to one storyboard per project now.

jpsim:
:+1: thanks Tim!

PR #3419, opened 2016-04-07
Add support for linking objects properties

### LINKED ISSUE #1324: Support queries on inverse relationships (backlinks)
It looks like the only way I can query inverse relationships is to create explicit links instead of using the recommendation for creating inverse relationships found in the docs (http://realm.io/docs/cocoa/#relationships).

Note: creating explicit links is explained here: https://github.com/realm/realm-cocoa/issues/677#issuecomment-50182237

bdash:
Linking objects properties provide a live view on which objects have a given relationship with an object:
```swift
class Person: Object {
    dynamic var name = ""
    dynamic var age = 0
    let children = List<Person>()
    let parents = LinkingObjects(fromType: Person.self, property: "children")
}
```

They're an improvement over the existing `Object.linkingObjects(_:forProperty:)` in three key ways:

1. They're a live, automatically updating view, rather than a point in time snapshot.
2. They can be used in queries:

    ```swift
    // People that have a child that have a parent named Diane.
    realm.objects(Person).filter("ANY children.parents.name == 'Diane'")
    ```
3. `LinkingObjects` is a subclass of `Results` and so supports all of methods you'd expect:

    ```swift
    // Which of my parents are over the age of 56?
    bdash.parents.filter("age > 56")
    
    // Calculate the age of my parents.
    bdash.parents.average("age")
    ```

Fixes #1324.

This depends on an unreleased version of Realm core.

Remaining work:
- [x] Schema validation to ensure that the specified class and property links to this object.
- [x] Notifications? I think these should work out of the box, but need tests to verify.
- [x] More testing of the non-query aspects.

Outstanding issues:
- [x] Objective-C `RLMLinkingObjects` properties are `nil` for standalone objects, but Swift `LinkingObjects` cannot be. How should we reconcile this?
- [ ] Do we need to support KVO on linking objects properties in the initial implementation?
- [x] Is there a nicer way to declare the properties in Objective-C?
- [x] Is it ok to make `Results` non-final? Can we avoid it?
- [x] What should we do with `Object.linkingObjects(_:forProperty:)`?


Realm/RLMObjectSchema.mm
```diff
@@ -56,23 +56,38 @@ - (instancetype)initWithClassName:(NSString *)objectClassName objectClass:(Class
...
 
 // create property map when setting property array
 -(void)setProperties:(NSArray *)properties {
     _properties = properties;
+    _propertiesInDeclaredOrder = nil;
+    [self _propertiesDidChange];
+}
+
+- (void)setComputedProperties:(NSArray *)computedProperties
+{
```
mrackwitz: Brace should go on the same line like method name for consistency.

Realm/RLMObjectSchema.mm
```diff
@@ -56,23 +56,38 @@ - (instancetype)initWithClassName:(NSString *)objectClassName objectClass:(Class
...
+    _computedProperties = computedProperties;
+    [self _propertiesDidChange];
+}
+
+- (void)_propertiesDidChange
+{
```
mrackwitz: Brace should be on the same line like method name for consistency.

RealmSwift-swift2.0/Results.swift
```diff
@@ -394,3 +394,28 @@ extension Results: RealmCollectionType {
...
+internal protocol LinkingObjectsInfo {
+    var objectClassName: String { get }
+    var propertyName: String { get }
+}
+
+/**
+ LinkingObjects is an auto-updating container type that represents a group of objects that
+ link to a given object.
+ */
+public class LinkingObjects<T: Object> : Results<T>, LinkingObjectsInfo {
```
mrackwitz: I think we could make this `final`?

cmelchior:
Two comments:
- Any reason you cannot just overload Results to also support this use case? Introducing a completely new public type + opening Results seems like something we would want to avoid if possible.
- Have you done any thinking about restricting the number of backlinks to only 1? At least on the Java side that would make some use cases easier, but we also have annotations to help with with the specification:

```
// Only one parent, but ideally that should also be enforced at insert time. Just crashing when querying seems a bit meh.
@BackLink 
private Person parent;

// Multiple parents exposed as a standard RealmResults.
@BackLink
private RealmResults<Person> parents;
```

Realm/Tests/QueryTests.m
```diff
@@ -33,12 +33,33 @@ @implementation NonRealmEmployeeObject
...
 @property NSInteger age;
+@property RLMArray<PersonObject> *children;
+@property (readonly) RLMLinkingObjects *parents;
 @end
 
 @implementation PersonObject
+
++ (NSDictionary *)linkingObjectsProperties
+{
+    return @{@"parents": @{@"class": @"PersonObject", @"property": @"children"} };
```
mrackwitz: Missing spaces. (`s/@{@"/@{ @"/`)

Realm/RLMObject.h
```diff
@@ -278,6 +278,21 @@ RLM_ASSUME_NONNULL_BEGIN
...
+ Implement to return a dictionary providing information related to linking objects properties.
+
+ Properties of type RLMLinkingObjects must have a corresponding entry in the dictionary to provide
+ information about the origin of the link that they represent. Their corresponding value in the
+ dictionary must be of the form:
+
+     @{ @"class": @"SomeClass", @"property": @"someProperty" }
+
+ Where `SomeClass` is a class that has a property `someProperty` that forms a relationship with this class.
+
```
mrackwitz: I think we should add an example for a full return value for further clarification. e.g.
```objc
@{ @"linkingProperty": @{ @"class": @"SomeClass", @"property": @"someProperty" } }
```

Realm/RLMProperty.mm
```diff
@@ -343,8 +381,11 @@ - (instancetype)initWithName:(NSString *)name
 
     _name = name;
     _indexed = indexed;
+    _objectClassName = linkingObjectsInfo[@"class"];
```
mrackwitz: These should be validated to be present within the dictionary and of the expected type.

mrackwitz:
Awesome work! Really excited about this. :tada:

---

>What should we do with Object.linkingObjects(_:forProperty:)?

I'd vote for deprecate it with the first minor release which ships these changes and remove it with the following.

---

>Any reason you cannot just overload Results to also support this use case? Introducing a completely new public type + opening Results seems like something we would want to avoid if possible.

While I think, that would be a good idea implementation-wise, given that we would not need to open up `Results` in that way, it may lead to confusion that the initializer which need to be exposed for that purpose could be used out of scope of object schema definition. (e.g. to retrieve all linked objects)
But maybe it would be helpful for documentation purposes to have a dedicated type. That would also give the concept a proper name, which could be referred to in case of questions. I can see how that could become a real advantage.

---

>Have you done any thinking about restricting the number of backlinks to only 1?

I think that would be something, we should do for Objective-C - just for convenience. In Swift, we would need to introduce something like a singular `LinkingObject` which would be another box not offering real advantage over a collection of one element beside documenting the multiplicity of the relationship.

bdash:
> Any reason you cannot just overload Results to also support this use case? Introducing a completely new public type + opening Results seems like something we would want to avoid if possible.

Reusing `RLMResults` for this purpose in Objective-C seemed like it would be somewhat error prone. Users can have properties of type `RLMResults` for purposes other than backlinks making it impossible to determine whether a property being absent from the dictionary returned by `+[RLMObject linkingObjectsProperties]` is intentional or due to a typo. A new type allows us to enforce that a `RLMLinkingObjects` property has a corresponding entry in the dictionary returned by `+linkingObjectsProperties`. It also makes it clearer in a class declaration what the property is since the only other knowledge that it contains linking objects is hidden in the class implementation.

> Have you done any thinking about restricting the number of backlinks to only 1?

I'm skeptical about supporting this at present given that we can't enforce there being only a single backlink to a given object.

mrackwitz:
For Swift at least we could have the following without introducing any semantic ambiguities.
```swift
class Person : Object {
    dynamic var name = ""
    dynamic var age = 0
    let children = List<Person>()
    let parents = Results(linkingObjectsFromType: Person.self, property: "children")
}
```

bdash:
Yes, the reasoning I mentioned for Objective-C doesn't apply for Swift. For Swift the choice of a separate type is primarily about consistency with Objective-C. IMO it's also more readable than using `Results` directly.

Realm/RLMProperty.mm
bdash: We can't usefully validate anything at this point. Some validation is performed within `-setTypeFromRawType`, and the majority will be performed at the object store layer within `Schema::validate`.

jpsim:
Another downside to using `Results` for Swift properties is that the constructor would also be available but invalid from other contexts. Putting that in its own type offers stronger context and less chance of accidental auto-complete or confusion IMO.

> > Have you done any thinking about restricting the number of backlinks to only 1?
>
> I'm skeptical about supporting this at present given that we can't enforce there being only a single backlink to a given object.

We could offer a shortcut to an optional first backlink, equivalent to `LinkingObjects(fromType: Person.self, property: "child").first`, but I don't see the value there for the same reasons @bdash shared.

> Objective-C `RLMLinkingObjects` properties are `nil` for standalone objects, but Swift `LinkingObjects` cannot be. How should we reconcile this?

Return a `LinkingObjects` backed by an empty RLMResults? I'd actually say that'd be preferable to `nil` in Objective-C too.

> Do we need to support KVO on linking objects properties in the initial implementation?

Without reviewing too deeply, I'd suspect this already works with KVO when tracking top-level results properties like `count`. If it's not too much work to assess what's missing for more complete KVO support, it'd be nice to at least do that before we determine whether or not it's worth the investment. At the very least we should track completing KVO support as a GH issue if we decide to punt on it.

> Is there a nicer way to declare the properties in Objective-C?

We could introduce some sort of `RLMLinkingObjectsDefinition` class in Objective-C which would likely avoid some user issues with remembering the NSDictionary format:

```objc
@interface RLMLinkingObjectsDefinition : NSObject
@property (nonatomic, copy) NSString *propertyName;
@property (nonatomic, weak) Class linkingClass;
@property (nonatomic, copy) NSString *linkingPropertyName;
@end
```

Naming alternative could be `RLMBacklinkDefinition` depending on if we want to use that term at all.

> Is it ok to make `Results` non-final? Can we avoid it?

We could avoid it by typaliasing `LinkingObjects` to `Results`, which would be nicer in some ways, worse in others.

> What should we do with `Object.linkingObjects(_:forProperty:)`?

I'd vote to deprecate it too.

bdash:
> Return a `LinkingObjects` backed by an empty RLMResults? I'd actually say that'd be preferable to `nil` in Objective-C too.

What behavior do we a `LinkingObjects` returned by a standalone object to have after the object is added to a Realm? Should it become live, or remain detached and empty?

jpsim:
> What behavior do we expect a LinkingObjects returned by a standalone object to have after the object is added to a Realm? Should it become live, or remain detached and empty?

Again, I'd compare the Results of linking objects as a query. If you perform a query referencing a standalone object, it returns an empty Results. In a perfect world adding that object to a realm would "upgrade" the Results to reflect that, but in practice that's not what happens today. In the grand scheme of things, probably fine to do the same for now with linking objects.

bdash:
This is now ready for review feedback.

/cc @jpsim, @tgoyne, @mrackwitz, @kishikawakatsumi

Realm/RLMProperty.mm
mrackwitz: Makes sense, my fault. :+1:

Realm/RLMObject.h
tgoyne: Dictionaries with a fixed set of keys are somewhat of an API antipattern. `[RLMLinkDescriptor descriptorWithClass:MyClass.class property:@"someProperty"]` (or `[RLMLinkDescriptor descriptorWithClassName:@"MyClass" property:@"someProperty"]`) would be a bit less error-prone.

An array of strings using keypath syntax would be a lot less verbose, but probably not a gain overall.

Realm/RLMObject.h
```diff
@@ -278,6 +278,21 @@ RLM_ASSUME_NONNULL_BEGIN
...
+ @return     NSDictionary mapping property names to information about linking objects properties.
+ */
++ (NSDictionary *)linkingObjectsProperties;
```
tgoyne: Return type should have a generic annotation.

Realm/RLMObjectSchema.mm
```diff
@@ -56,23 +56,36 @@ - (instancetype)initWithClassName:(NSString *)objectClassName objectClass:(Class
...
+    [self _propertiesDidChange];
+}
+
+- (void)setComputedProperties:(NSArray *)computedProperties {
+    _computedProperties = computedProperties;
+    [self _propertiesDidChange];
+}
+
+- (void)_propertiesDidChange {
     NSMutableDictionary *map = [NSMutableDictionary dictionaryWithCapacity:_properties.count];
```
tgoyne: _properties.count + _computedProperties.count

Realm/RLMObjectSchema.mm
```diff
@@ -268,6 +312,7 @@ - (id)copyWithZone:(NSZone *)zone {
 
     // call property setter to reset map and primary key
     schema.properties = [[NSArray allocWithZone:zone] initWithArray:_properties copyItems:YES];
+    schema.computedProperties = [[NSArray allocWithZone:zone] initWithArray:_computedProperties copyItems:YES];
```
tgoyne: Does this need to deep copy? IIRC for persisted properties that's purely for the `column` property, and that doesn't appear to be used for computed ones.

Realm/RLMQueryUtil.mm
```diff
@@ -141,41 +141,88 @@ void set_base_table(const Table*) override {}
...
-    explicit ColumnReference(RLMProperty* property) : m_property(property) { }
 
-    template <typename T>
+    template <typename T, typename std::enable_if<!std::is_same<T, Link>::value>::type* = nullptr>
     auto resolve(Query& query) const
     {
         return table_for_query(query)->template column<T>(index());
     }
 
+    template <typename T, typename std::enable_if<std::is_same<T, Link>::value>::type* = nullptr>
```
tgoyne: Since it doesn't need to be partially specialized, I think this could just be an explicit specialization for `<Link>`.

Realm/RLMResults.mm
```diff
@@ -237,6 +237,13 @@ + (instancetype)resultsWithObjectSchema:(RLMObjectSchema *)objectSchema
     return ar;
 }
 
++ (instancetype)emptyDetachedResults
+{
+    RLMResults *results = [[self alloc] initPrivate];
+    results->_results = Results();
```
tgoyne: `_results` should be already default-initialized by the runtime.

Realm/RLMObjectSchema.mm
bdash: `RLMProperty` has some mutable members so it seemed safest to avoid any chance of sharing them.

bdash:
Rebasing on top of Thomas's fine grained notifications work revealed a new core bug. I've submitted a PR for review, but we'll need a new core release before we can merge this.

jpsim:
This looks excellent @bdash! My only concern is cosmetic: I dislike the `klass` name. Could we call it `objectClass` for consistency with other parts of the API that refer to `objectClassName` for example?

To address your remaining checklist items, yes I think it's fine to make Results non-final, there's not much users can do to hurt themselves by subclassing Results. As for KVO support, I do think that needs to be documented and an issue be created to track addressing that, but it's certainly not worth delaying this next release to get that in.

Once core 0.98.1 is out, a big 👍 from me!

mrackwitz:
I agree with what @jpsim commented here, especially the big :+1:!

tgoyne:
CI failure is an actual Swift error. The obvious resolutions I can see are:

1. Just remove addNotificationBlock from the protocol (it wasn't there in previous versions so this would be just not adding a feature as opposed to a breaking change)
2. Make LinkingObjects a function which returns an appropriately-constructed Results (and thus push all of the functionality currently there into Results, which is kinda eh).
3. Make LinkingObjects inherit from ResultsBase, and either duplicate a bunch of wrappers or push the functionality there from Results.

tgoyne:
There doesn't need to be extensive tests, but a test or two that minimally verifies that fine-grained notifications for backlink results aren't totally non-functional would be nice.

bdash:
None seem like great options. 1 seems like the least bad to me. Anyone else have strong feelings on this?

jpsim:
Agreed, 1 is least bad. It's easier to make an addition afterwards.

bdash:
While adding the tests for fine-grained notifications that @tgoyne requested I've discovered a few issues that need to be addressed. At least one appears to require a fix in core 😦

bdash:
Latest changes pick up core v0.98.2, which includes the fixes for the core issues I mentioned above.

bdash:
The [CI failure](https://ci.realm.io/job/objc_pr/3148/) shows a real issue in the core fix. We'll need another core update with an updated fix.

bdash:
Updating the Swift backlinks example to use `LinkingObjects` has caused the Swift compiler to segfault:

```
0  swift                    0x000000010b0944eb llvm::sys::PrintStackTrace(llvm::raw_ostream&) + 43
1  swift                    0x000000010b0937d6 llvm::sys::RunSignalHandlers() + 70
2  swift                    0x000000010b094b4f SignalHandler(int) + 287
3  libsystem_platform.dylib 0x00007fff9867852a _sigtramp + 26
4  libsystem_platform.dylib 0x00007fa6d1495840 _sigtramp + 954323760
5  swift                    0x0000000108ea85cb emitNominalMetadataRef(swift::irgen::IRGenFunction&, swift::NominalTypeDecl*, swift::CanType) + 2715
6  swift                    0x0000000108eeb31f emitWitnessTableAccessorCall(swift::irgen::IRGenFunction&, swift::NormalProtocolConformance const*, swift::CanType, llvm::Value**) + 79
7  swift                    0x0000000108ee962d swift::irgen::emitWitnessMethodValue(swift::irgen::IRGenFunction&, swift::CanType, llvm::Value**, swift::SILDeclRef, swift::ProtocolConformanceRef, swift::irgen::Explosion&) + 189
8  swift                    0x0000000108f244e1 swift::SILVisitor<(anonymous namespace)::IRGenSILFunction, void>::visit(swift::ValueBase*) + 8225
9  swift                    0x0000000108f1ec4b swift::irgen::IRGenModule::emitSILFunction(swift::SILFunction*) + 9787
10 swift                    0x0000000108e79fd8 swift::irgen::IRGenModuleDispatcher::emitGlobalTopLevel() + 600
11 swift                    0x0000000108f05a5e performIRGeneration(swift::IRGenOptions&, swift::ModuleDecl*, swift::SILModule*, llvm::StringRef, llvm::LLVMContext&, swift::SourceFile*, unsigned int) + 1278
12 swift                    0x0000000108f05f06 swift::performIRGeneration(swift::IRGenOptions&, swift::SourceFile&, swift::SILModule*, llvm::StringRef, llvm::LLVMContext&, unsigned int) + 70
13 swift                    0x0000000108de9a1c performCompile(swift::CompilerInstance&, swift::CompilerInvocation&, llvm::ArrayRef<char const*>, int&) + 15004
14 swift                    0x0000000108de541d frontend_main(llvm::ArrayRef<char const*>, char const*, void*) + 2781
15 swift                    0x0000000108de0e3c main + 1932
16 libdyld.dylib            0x00007fff982aa5ad start + 1
```

😭

bdash:
Using a Swift 2.2.1 snapshot build the crash is different, but more descriptive:

```
Assertion failed: ((getType()->isEqual(foundInherited->getType()) || foundInherited->getType()->isSuperclassOf(getType(), nullptr)) && "inherited conformance does not match type"), function getInheritedConformance, file /Users/buildnode/jenkins/workspace/oss-swift-2.2-package-osx/swift/lib/AST/ProtocolConformance.cpp, line 526.
0  swift                    0x0000000104ec247b llvm::sys::PrintStackTrace(llvm::raw_ostream&) + 43
1  swift                    0x0000000104ec1706 llvm::sys::RunSignalHandlers() + 70
2  swift                    0x0000000104ec2b4f SignalHandler(int) + 383
3  libsystem_platform.dylib 0x00007fff8845652a _sigtramp + 26
4  libsystem_platform.dylib 0x00007fff601e9ac5 _sigtramp + 3621336501
5  swift                    0x0000000104ec2926 abort + 22
6  swift                    0x0000000104ec2901 __assert_rtn + 81
7  swift                    0x0000000102ec94b9 swift::ProtocolConformance::getInheritedConformance(swift::ProtocolDecl*) const + 1513
8  swift                    0x0000000102ece7df swift::Substitution::subst(swift::ModuleDecl*, llvm::ArrayRef<swift::Substitution>, llvm::DenseMap<swift::TypeBase*, swift::Type, llvm::DenseMapInfo<swift::TypeBase*>, llvm::detail::DenseMapPair<swift::TypeBase*, swift::Type> >&, llvm::DenseMap<swift::ArchetypeType*, llvm::ArrayRef<swift::ProtocolConformanceRef>, llvm::DenseMapInfo<swift::ArchetypeType*>, llvm::detail::DenseMapPair<swift::ArchetypeType*, llvm::ArrayRef<swift::ProtocolConformanceRef> > >&) const + 1327
9  swift                    0x0000000102ece1e5 swift::Substitution::subst(swift::ModuleDecl*, swift::GenericParamList*, llvm::ArrayRef<swift::Substitution>) const + 1509
10 swift                    0x0000000102938c77 swift::TypeSubstCloner<swift::GenericCloner>::visitWitnessMethodInst(swift::WitnessMethodInst*) + 103
11 swift                    0x00000001029348cb swift::SILCloner<swift::GenericCloner>::visitSILBasicBlock(swift::SILBasicBlock*) + 75
12 swift                    0x0000000102934bd9 swift::SILCloner<swift::GenericCloner>::visitSILBasicBlock(swift::SILBasicBlock*) + 857
13 swift                    0x0000000102933f8f swift::GenericCloner::populateCloned() + 399
14 swift                    0x0000000102945032 swift::GenericCloner::cloneFunction(swift::SILFunction*, llvm::DenseMap<swift::TypeBase*, swift::Type, llvm::DenseMapInfo<swift::TypeBase*>, llvm::detail::DenseMapPair<swift::TypeBase*, swift::Type> >&, llvm::DenseMap<swift::TypeBase*, swift::Type, llvm::DenseMapInfo<swift::TypeBase*>, llvm::detail::DenseMapPair<swift::TypeBase*, swift::Type> >&, llvm::StringRef, swift::ApplySite, std::__1::function<void (swift::SILInstruction*, swift::SILInstruction*)>) + 274
15 swift                    0x0000000102944a44 swift::trySpecializeApplyOfGeneric(swift::ApplySite, swift::SILFunction*&, swift::CloneCollector&) + 2772
16 swift                    0x00000001029e40f2 (anonymous namespace)::GenericSpecializer::run() + 658
17 swift                    0x000000010294e158 swift::SILPassManager::runPassesOnFunction(llvm::ArrayRef<swift::SILFunctionTransform*>, swift::SILFunction*) + 1224
18 swift                    0x000000010294ee6f swift::SILPassManager::runFunctionPasses(llvm::ArrayRef<swift::SILFunctionTransform*>) + 847
19 swift                    0x000000010294fe04 swift::SILPassManager::runOneIteration() + 1012
20 swift                    0x0000000102959a15 swift::runSILOptimizationPasses(swift::SILModule&) + 213
21 swift                    0x000000010255fb99 performCompile(swift::CompilerInstance&, swift::CompilerInvocation&, llvm::ArrayRef<char const*>, int&) + 14473
22 swift                    0x000000010255b4ce frontend_main(llvm::ArrayRef<char const*>, char const*, void*) + 2814
23 swift                    0x00000001025563cc main + 1532
24 libdyld.dylib            0x00007fff8c7195ad start + 1
25 libdyld.dylib            0x000000000000003c start + 1938713232
```

bdash:
Wrote up https://bugs.swift.org/browse/SR-1323 about the assertion failure.

Realm/RLMProperty.h
```diff
@@ -92,4 +97,17 @@ RLM_ASSUME_NONNULL_BEGIN
 
 @end
 
+
+/**
+ This class describes a specific property on a given class.
+ */
+@interface RLMPropertyDescriptor : NSObject
+
++ (instancetype)descriptorWithClass:(Class)objectClass propertyName:(NSString *)propertyName;
```
mrackwitz: Have we ever considered using selectors for the property name? That would give users a bit more compile-time safety when using Realm Objective-C from Swift / Realm Swift.

bdash: Swift generates warnings since it doesn't consider properties to have selectors:
```
examples/ios/swift-2.2/Backlink/AppDelegate.swift:27:75: warning: no method declared with Objective-C selector 'dogs'
    let owners = LinkingObjects(fromType: Person.self, property: Selector("dogs"))
```

tgoyne: There was a recently-accepted proposal to fix that, but Swift 3 will also have `#keypath(Person.dogs)` for a checked keypath string.

bdash: Looks like that is [SE-0062](https://github.com/apple/swift-evolution/blob/master/proposals/0062-objc-keypaths.md). It'd be nicer if it was more than a compiler-checked string literal since using that will mean restating the type in both arguments to the `LinkingObjects` initializer.

mrackwitz: Great. I wasn't fully aware of `#keypath`.

bdash:
I'm working on extending test coverage of `LinkingObjects` now that it doesn't share code with `Results`.

RealmSwift-swift2.0/LinkingObjects.swift
```diff
@@ -0,0 +1,438 @@
...
+ LinkingObjects always reflect the current state of the Realm on the current thread,
+ including during write transactions on the current thread. The one exception to
+ this is when using `for...in` enumeration, which will always enumerate over the
+ linking objects when the enumeration is begun, even if some of them are deleted or
+ modified to no longer link to the target object during the enumeration.
+
+ LinkingObjects can only be used as a property on `Object` models. The property must
+ be declared as `let` and cannot be `dynamic`.
+ */
+public class LinkingObjects<T: Object>: NSObject, UntypedLinkingObjects {
```
jpsim: `final`?

bdash:
Adding more extensive testing of `LinkingObjects` revealed two further issues with `Query`s that are restricted by `TableView`s that need to be addressed in core. I'll submit pull requests to address those shortly.

RealmSwift-swift2.0/Tests/RealmCollectionTypeTests.swift
```diff
@@ -19,39 +19,65 @@
 import XCTest
 import RealmSwift
 
-class SwiftAggregateObjectList: Object {
-    let list = List<SwiftAggregateObject>()
+class CTTAggregateObject: Object {
```
tgoyne: CTT?

examples/ios/objc/Backlink/AppDelegate.m
```diff
@@ -23,16 +23,8 @@
 @interface Dog : RLMObject
 @property NSString *name;
 @property NSInteger age;
-@property (readonly) NSArray *owners; // Realm doesn't persist this property because it is readonly
+@property (readonly) RLMLinkingObjects *owners; // Realm doesn't persist this property because it is readonly
```
tgoyne: Comment no longer really makes sense.

RealmSwift-swift2.0/Tests/RealmCollectionTypeTests.swift
bdash: Initialism of CollectionTypeTests, to distinguish these types from similarly-named ones used elsewhere in the RealmSwift tests.

bdash:
The performance regression caught by CI appears to have primarily been introduced by #3458. The extra test classes in this PR increased the regression by just enough to push it over the failure threshold. 3fd6f91 should fix the regression sufficiently to make CI happy again.

tgoyne:
👍

jkmathew:
Hi @bdash 
I have related issue.
I have `Friend` class and a `FriendGroup` class. `FriendGroup` has a to may relation to `Friend` also, `Friend` has an inverse relation to `FriendGroup`. Now I want to fetch all friends to a  `Results<Friend>` who are not added to any groups.

My class declaration is like,
```
class Friend: Object {
    @objc dynamic var name = ""
    let group = LinkingObjects(fromType: FriendGroup.self, property: "members")    
}

class FriendGroup: Object {
    @objc dynamic var name = ""
  
    let members = List<Friend>()
}

```

And I tried fetching like
```
let predicae = NSPredicate(format: "group.count < %ld", 1)
friends = realm?.objects(Friend.self).filter(predicae).sorted(byKeyPath: "name")
```

But its crashing with stack trace
```
*** Terminating app due to uncaught exception 'Invalid property name', reason: 'Property 'count' not found in object of type 'FriendGroup''
*** First throw call stack:
(
	0   CoreFoundation                      0x0000000106ca51cb __exceptionPreprocess + 171
	1   libobjc.A.dylib                     0x0000000105a2bf41 objc_exception_throw + 48
	2   Realm                               0x000000010465e4c1 _ZL15RLMPreconditionbP8NSStringS0_z + 657
	3   Realm                               0x0000000104666b97 _ZN12_GLOBAL__N_120key_path_from_stringEP9RLMSchemaP15RLMObjectSchemaP8NSString + 1015
	4   Realm                               0x0000000104663f8c _ZN12_GLOBAL__N_112QueryBuilder30column_reference_from_key_pathEP15RLMObjectSchemaP8NSStringb + 140
	5   Realm                               0x0000000104662fa2 _ZN12_GLOBAL__N_112QueryBuilder22apply_value_expressionEP15RLMObjectSchemaP8NSStringP11objc_objectP21NSComparisonPredicate + 338
	6   Realm                               0x0000000104660033 _ZN12_GLOBAL__N_112QueryBuilder15apply_predicateEP11NSPredicateP15RLMObjectSchema + 5075
	7   Realm                               0x000000010465e67b _Z19RLMPredicateToQueryP11NSPredicateP15RLMObjectSchemaP9RLMSchemaRN5realm5GroupE + 347
	8   Realm                               0x000000010479b57b _ZZ35-[RLMResults objectsWithPredicate:]ENK4$_11clEv + 283
	9   Realm                               0x0000000104795206 _ZL15translateErrorsIZ35-[RLMResults objectsWithPredicate:]E4$_11EDaOT_P8NSString + 38
	10  Realm                               0x0000000104795188 -[RLMResults objectsWithPredicate:] + 72
	11  RealmSwift                          0x00000001050e0459 _T010RealmSwift7ResultsC6filterACyxGSo11NSPredicateCF + 153
	12  Invite                              0x00000001041a4212 _T06Invite20ImportViewControllerC11viewDidLoadyyF + 514
	13  Invite                              0x00000001041a4a94 _T06Invite20ImportViewControllerC11viewDidLoadyyFTo + 36
	14  UIKit                               0x0000000107844d51 -[UIViewController loadViewIfRequired] + 1235
	15  UIKit                               0x000000010784519e -[UIViewController view] + 27
	16  UIKit                               0x000000010787713b -[UINavigationController _startCustomTransition:] + 954
	17  UIKit                               0x000000010788d894 -[UINavigationController _startDeferredTransitionIfNeeded:] + 686
	18  UIKit                               0x000000010788eb90 -[UINavigationController __viewWillLayoutSubviews] + 115
	19  UIKit                               0x0000000107ae52ae -[UILayoutContainerView layoutSubviews] + 231
	20  UIKit                               0x0000000107775551 -[UIView(CALayerDelegate) layoutSublayersOfLayer:] + 1331
	21  QuartzCore                          0x000000010f85f4ba -[CALayer layoutSublayers] + 153
	22  QuartzCore                          0x000000010f8635a9 _ZN2CA5Layer16layout_if_neededEPNS_11TransactionE + 401
	23  QuartzCore                          0x000000010f7ec1cd _ZN2CA7Context18commit_transactionEPNS_11TransactionE + 365
	24  QuartzCore                          0x000000010f817ae4 _ZN2CA11Transaction6commitEv + 500
	25  QuartzCore                          0x000000010f818830 _ZN2CA11Transaction17observer_callbackEP19__CFRunLoopObservermPv + 76
	26  CoreFoundation                      0x0000000106c47db7 __CFRUNLOOP_IS_CALLING_OUT_TO_AN_OBSERVER_CALLBACK_FUNCTION__ + 23
	27  CoreFoundation                      0x0000000106c47d0e __CFRunLoopDoObservers + 430
	28  CoreFoundation                      0x0000000106c2c324 __CFRunLoopRun + 1572
	29  CoreFoundation                      0x0000000106c2ba89 CFRunLoopRunSpecific + 409
	30  GraphicsServices                    0x000000010e37e9c6 GSEventRunModal + 62
	31  UIKit                               0x00000001076a6d30 UIApplicationMain + 159
	32  Invite                              0x000000010419b2e7 main + 55
	33  libdyld.dylib                       0x000000010743bd81 start + 1
)
libc++abi.dylib: terminating with uncaught exception of type NSException
```

group is a `LinkingObjects` object, then why its crashing

Issue #3413, opened 2016-04-06
NSString and NSDate property values are removed after migration, when property goes in requiredProperties


rsahara:
## Goals

Migrate a realm db, keeping old property values if the type of that property is not changed.

## Expected Results

Before migration: NSString nullable property: value "hello world"
After migration: NSString non nullable property: value "hello world"

Before migration: NSDate nullable property: value 2016/04/06 11:45:51
After migration: NSDate non nullable property: value 2016/04/06 11:45:51

## Actual Results

Before migration: NSString nullable property: value "hello world"
After migration: NSString non nullable property: value ""

Before migration: NSDate nullable property: value 2016/04/06 11:45:51
After migration: NSDate non nullable property: value 1970/01/01 9:00:00

Check models ModelSearchParam and ModelTab in the attached realm files persistent_before.realm and persistent_after.realm.
[persistent.zip](https://github.com/realm/realm-cocoa/files/206107/persistent.zip)

## Steps to Reproduce

Perform a migration by moving a property from nullable to non nullable without changing its type/name.

## Code Sample

```
@interface ModelTab : RLMObject
@property NSDate*	modelUpdate;
@property BOOL		modelDelete;
@property NSString* tabId;
@property NSString* tabTitle;
@property int		tabIndex;
@property int		tabType;
@property int		siteId;
@property ModelSearchParam*	searchParam;
@property int		tabParam;
@end

@implementation ModelTab
+ (NSString*)primaryKey {
	return @"tabId";
}
+ (NSDictionary *)defaultPropertyValues {
	return @{
			 @"modelUpdate":[NSDate date],
			 @"modelDelete":@NO,
			 @"tabId":[[NSUUID UUID] UUIDString],
			 @"tabTitle":@"",
			 @"siteId":[NSNumber numberWithInt:ModelSiteIdWomen],
			 @"tabParam":[NSNumber numberWithInt:ModelTabParamNone],
			 };
}
// Just added this method in v. 96 (previous v. is 87)
+ (NSArray *)requiredProperties {
	return @[
			 @"modelUpdate",
			 @"tabTitle",
			 ];
}
@end
```

## Version of Realm and Tooling

Realm version: 0.96.3 (same issue with 0.98.6 linked to the app that perform the migration)

Xcode version: 7.0.1

iOS/OSX version: 9.0

Dependency manager + version: CocoaPods 0.39.0

kishikawakatsumi:
@rsahara 

[...]

PR #3420, opened 2016-04-08
[Migration] Keep old value when nullable to required auto migration

kishikawakatsumi:
If change the property attribute from nullable to required, automatic migration works. But the values will be cleared. This behavior is difficult to understand what happened.
So keep the value if just changing nullability.

If the old value is null, the new value will be an empty string, a zero, a zero-length data or a UNIX epoch date.

This user confused the current behavior https://github.com/realm/realm-cocoa/issues/3413.

CC @tgoyne @jpsim @bdash

### DIFF
--- a/Realm/ObjectStore/object_store.cpp
+++ b/Realm/ObjectStore/object_store.cpp
@@ -152,6 +152,13 @@ static inline bool property_can_be_migrated_to_nullable(const Property& old_prop
         && new_property.name == old_property.name;
 }
 
+static inline bool property_can_be_migrated_to_required(const Property& old_property, const Property& new_property) {
+    return old_property.type == new_property.type
+    && old_property.is_nullable
+    && !new_property.is_nullable
+    && new_property.name == old_property.name;
+}
+
 void ObjectStore::verify_schema(Schema const& actual_schema, Schema& target_schema, bool allow_missing_tables) {
     std::vector<ObjectSchemaValidationException> errors;
     for (auto &object_schema : target_schema) {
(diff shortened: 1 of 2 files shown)


jpsim:
I think something else is happening in #3413: a "nullable to required" property transition doesn't quality for an automatic migration, so just doing so will cause a migration to be required. When bumping the schema and writing a migration block, it'll then be necessary to copy data from the old property to the new, adjusting for any potential nulls along the way.

I'd prefer to blow up in a big way when a user fails to copy data to the new required property as we do now, rather than have it work "most of the time" when there's a value but not for nulls, which could easily go unnoticed by the developer.

rsahara:
@kishikawakatsumi @jpsim 

Just uploaded a sample project in case it can help: https://github.com/rsahara/realmtest

Logs with REALMVERSION=1

```
test db path: /Users/rsahara/Library/Developer/CoreSimulator/Devices/E62451F2-95DA-4D99-AB35-C197539BD48F/data/Containers/Data/Application/F249D5E7-4E72-4EA7-A519-877D34C0B8D2/tmp/test.realm
Created ModelTest with values [2016-05-12 08:41:19 +0000] [none] [123]
```

Logs with REALMVERSION=2
```
test db path: /Users/rsahara/Library/Developer/CoreSimulator/Devices/E62451F2-95DA-4D99-AB35-C197539BD48F/data/Containers/Data/Application/B3D8E06E-CCF6-4785-B21F-5B794D332B5C/tmp/test.realm
Migration without touching any values
Found ModelTest with values [1970-01-01 00:00:00 +0000] [] [123]
```

This project shows how the migration ends silently with the "unknown values" above. I didn't notice those values changed at first. Neither the old values, nor the default values I provided are used.
A bad migration can kill a service, so I think something to avoid those values, unexpected to the developer, would be helpful.
Thanks!

jpsim:
Discussing this more, @kishikawakatsumi and I determined that it should be a programmer error to not set new values in the new Realm's required property. So we're closing this.

PR #3421, opened 2016-04-08
Have -[RLMResults indexOfObjectWithPredicate:] respect the sort order

### LINKED ISSUE #3410: [RLMResults indexOfObjectWhere:...] returning not first matching index
## Code Sample
```objc
//Object
@interface RLMItem : RLMObject
@property NSString *code;
@property NSString *details;
@end
@implementation RLMItem
+ (NSString *)primaryKey { return @"code"; }
+ (NSArray *)indexedProperties { return @[@"details"]; }
@end

//Query
RLMResults *results = [[RLMItem allObjects] sortedResultsUsingProperty:@"details" ascending:YES];
NSString *letter = @"D"; //any letter
NSUInteger index = [results indexOfObjectWhere:@"details BEGINSWITH %@", letter]; // <- not expected/wrong index
```

## Expected Results

Expecting index of first object with letter D, because I've made sorting.

## Actual Results

Getting index of middle object (random?) that matches query.

## Goal
Make alphabet scroll using indexOfObjectWhere:

Realm version: 0.98.5

Xcode version: 7.3

iOS/OSX version: 10.10.4

Dependency manager + version: Cocoapods 0.39.0

bdash:
If the `RLMResults` has been sorted then the returned index should be the first object that is matched by the predicate, not an arbitrary object.

Fixes #3410.

/cc @jpsim @tgoyne 

The performance here could be improved if `TableView` supported partial sorting or selecting the nth element based on the sort criteria. Those doesn't really fit with `TableView`'s current design for sorting though.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,8 @@ x.x.x Release notes (yyyy-MM-dd)
   for `RLMResults` instances that were created by filtering an `RLMArray`.
 * Adjust how RLMObjects are destroyed in order to support using an associated
   object on an RLMObject to remove KVO observers from that RLMObject.
+* `-[RLMResults indexOfObjectWithPredicate:]` now returns the index of the first matching object for a
+  sorted `RLMResults`, matching its documented behavior.
 
 0.98.6 Release notes (2016-03-25)
 =============================================================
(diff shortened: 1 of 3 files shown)


jpsim:
:+1:

tgoyne:
:+1:

Issue #3372, opened 2016-03-25
Crash when delete object in 0.98.5


nixzhu:
Crash at:

```
    // only here to follow the general scheme, base class method could be used instead
    void apply_patch(HandoverPatch& patch, Group& group)
    {
        TableViewBase::apply_patch(patch, group);
    }
```

```
(lldb) bt all
* thread #1: tid = 0xd7b07, 0x22380e8a libsystem_pthread.dylib`pthread_mutex_lock + 10, queue = 'com.apple.main-thread', stop reason = EXC_BAD_ACCESS (code=1, address=0x230)
    frame #0: 0x22380e8a libsystem_pthread.dylib`pthread_mutex_lock + 10
    frame #1: 0x01a2714a Realm`realm::Table::register_row_accessor(realm::RowBase*) const + 90
    frame #2: 0x01a20300 Realm`realm::RowBase::apply_patch(realm::RowBaseHandoverPatch&, realm::Group&) + 80
    frame #3: 0x019e09d2 Realm`realm::LinksToNode::apply_handover_patch(std::__1::vector<std::__1::unique_ptr<realm::QueryNodeHandoverPatch, std::__1::default_delete<realm::QueryNodeHandoverPatch> >, std::__1::allocator<std::__1::unique_ptr<realm::QueryNodeHandoverPatch, std::__1::default_delete<realm::QueryNodeHandoverPatch> > > >&, realm::Group&) + 206
    frame #4: 0x019d4662 Realm`realm::Query::apply_patch(realm::QueryHandoverPatch&, realm::Group&) + 350
    frame #5: 0x01a4ab16 Realm`realm::TableViewBase::apply_patch(realm::TableViewHandoverPatch&, realm::Group&) + 138
  * frame #6: 0x018c9506 Realm`realm::TableView::apply_patch(this=0x17c7d550, patch=0x16789210, group=0x169c9a00) + 22 at table_view.hpp:602
    frame #7: 0x018594ec Realm`realm::TableView::apply_and_consume_patch(this=0x17c7d550, patch=0x167694f0, group=0x169c9a00) + 36 at table_view.hpp:583
    frame #8: 0x017c13fc Realm`std::__1::unique_ptr<realm::TableView, std::__1::default_delete<realm::TableView> > realm::SharedGroup::import_from_handover<realm::TableView>(this=0x169c9a00, handover=<unavailable>) + 396 at group_shared.hpp:912
[...]

PR #3429, opened 2016-04-13
update core version to 0.97.3

jpsim:
Fix I'm referring to is realm/realm-core#1585 /cc @bdash @tgoyne


bdash:
Core v0.97.3 also includes a fix for #3372.

jpsim:
You're right. Added.

bdash:
👍

tgoyne:
:+1:

PR #3437, opened 2016-04-14
Set error if mmap address space is exhausted

### LINKED ISSUE #2269: Set NSError on Realm initialization failure due to inability to mmap
Otherwise, users have no way to gracefully handle the failure in their apps, forcing them to either catch all exceptions upon Realm initialization or crash.

We're tracking this in core as https://github.com/realm/realm-core/issues/885.

mrackwitz:
This is based on the assumption that the address space will be exhausted most commonly either on trying to map a Realm database which doesn't fit in the left address space or on committing a writing transaction when the memory is resized, while this could also happen on (auto-)refreshes.
Fixes #2269.

/c @jpsim @bdash @tgoyne

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,7 +32,8 @@ x.x.x Release notes (yyyy-MM-dd)
 
 ### Enhancements
 
-* None.
+* Return a recoverable `RLMErrorAddressSpaceExhausted` error rather than crashing when failing to
+  mmap in Realm initialization or write commit.
 
 ### Bugfixes
 
--- a/Realm/RLMConstants.h
+++ b/Realm/RLMConstants.h
@@ -81,6 +81,8 @@ typedef NS_ENUM(NSInteger, RLMError) {
         process which cannot share with the current process due to an
         architecture mismatch. */
     RLMErrorIncompatibleLockFile  = 8,
+    /** Returned by RLMRealm if there is insufficient available address space. */
+    RLMErrorAddressSpaceExhausted = 9,
 };
 
 #pragma mark - Constants
--- a/Realm/Tests/RealmTests.mm
+++ b/Realm/Tests/RealmTests.mm
@@ -177,6 +177,7 @@ extern "C" {
     RLMAssertThrowsWithCodeMatching([RLMRealm realmWithConfiguration:config error:nil], RLMErrorFileFormatUpgradeRequired);
 }
 
+#if TARGET_OS_IPHONE && (!TARGET_IPHONE_SIMULATOR || !TARGET_RT_64_BIT)
 - (void)testExceedingVirtualAddressSpace {
     RLMRealmConfiguration *config = [RLMRealmConfiguration defaultConfiguration];
 
@@ -227,6 +228,7 @@ extern "C" {
         XCTAssertNoThrow([RLMRealm realmWithConfiguration:config error:nil]);
     }
 }
+#endif
 
 #pragma mark - Adding and Removing Objects
 
--- a/Realm/Tests/RealmTests.mm
+++ b/Realm/Tests/RealmTests.mm
@@ -27,6 +27,8 @@ extern "C" {
 }
 
 #import <thread>
+#import <mach/mach_init.h>
+#import <mach/vm_map.h>
 
 @interface RLMRealm ()
 + (BOOL)isCoreDebug;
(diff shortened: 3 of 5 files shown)


CHANGELOG.md
```diff
@@ -8,7 +8,7 @@ x.x.x Release notes (yyyy-MM-dd)
 
 ### Enhancements
 
-* None.
+* Fail with `RLMErrorAddressSpaceExhausted` if unable to acquire more mmap memory.
```
jpsim: Return a recoverable `RLMErrorAddressSpaceExhausted` error rather than crashing when failing to mmap in Realm initialization or write commit.

Realm/RLMConstants.h
```diff
@@ -81,6 +81,9 @@ typedef NS_ENUM(NSInteger, RLMError) {
         process which cannot share with the current process due to an
         architecture mismatch. */
     RLMErrorIncompatibleLockFile  = 8,
+    /** Returned by RLMRealm if unable to acquire more mmap memory because it
```
jpsim: Returned by RLMRealm when there is insufficient available mmap address space.

RealmSwift-swift2.0/Error.swift
```diff
@@ -87,9 +87,13 @@ public enum Error: ErrorType {
     /// cannot share with the current process due to an architecture mismatch.
     case IncompatibleLockFile
 
-    /// Returned by RLMRealm if a file format upgrade is required to open the file,
+    /// Error thrown by Realm if a file format upgrade is required to open the file,
     /// but upgrades were explicilty disabled.
     case FileFormatUpgradeRequired
+
+    /// Error thrown by Realm if unable to acquire more mmap memory because it would exhaust the
```
jpsim: Error thrown by Realm when there is insufficient available mmap address space.

Realm/RLMConstants.h
mrackwitz: That's better. I'll stick with *if* instead of *when* though for consistency.

jpsim:
Some context from a previous conversation: although this exception can be thrown in `advance_read`, it's not easily recoverable there because all accessors will become invalid. Also, advance_read can happen implicitly in so many situations that it would be impractical to allow error handling there.

One solution to this might be to allow users to register custom error handlers that we would invoke for them, but the recovery options would still be very limited, so probably not worthwhile implementing unless there's a strong demand for it.

I'll wait for the compile issues to be sorted out before doing a full review.

mrackwitz:
This passes now on CI. (Jenkins failure is due to the device build being aborted.)

tgoyne:
It'd be nice to have a test that verifies that we actually get the correct error and that it's recoverable. An idea for how to do this is:

    create realm file with a 1MB string, close realm
    size = 1024 * 1024 * 1024
    while (size >= 1024 * 1024) {
        kern_return_t ret = vm_allocate(size)
        if (ret == KERN_NO_SPACE)
            size /= 2
        else
            add allocation to list to free
    }
    try to open realm, assert that expected error is given
    deallocate all allocations above
    try again to open realm and verify that it works

This would obviously need to be 32-bit only. It may turn out to be unreasonably slow to run it as part of the normal test suite, but it'd still be good to at least do a one-time verification.

mrackwitz:
@tgoyne: Thanks for the very detailed proposal how to test that. I was thinking about that and @ironage was also giving me some good feedback. I'll report back, once I was able to reproduce it, whether it is actually recoverable and if it's a good idea to test that continuously.

jpsim:
:+1: if testing confirms this works

mrackwitz:
I think, I could reproduce it, assuming that my test case makes sense.
It didn't seemed like it was a huge performance regression for the test suite at least locally for me, but it might look different for our CI VMs.

mrackwitz:
I think it would be good if someone could review my test case and give a :+1: if everything is alright.

Realm/Tests/RealmTests.mm
```diff
@@ -175,6 +177,58 @@ - (void)testFileFormatUpgradeRequiredButDisabled {
...
+        stringObj.stringCol = largeString;
+        [realm addObject:stringObj];
+        [realm commitWriteTransaction];
+    }
+
+    struct VirtualMemoryChunk {
+        vm_address_t address;
+        vm_size_t size;
+    };
+    typedef struct VirtualMemoryChunk VirtualMemoryChunk;
```
bdash: This `typedef` shouldn't be necessary since this code is in an Objective-C++ file.

Realm/Tests/RealmTests.mm
```diff
@@ -175,6 +177,58 @@ - (void)testFileFormatUpgradeRequiredButDisabled {
     RLMAssertThrowsWithCodeMatching([RLMRealm realmWithConfiguration:config error:nil], RLMErrorFileFormatUpgradeRequired);
 }
 
+- (void)testExceedingVirtualAddressSpace {
```
bdash: How long does this test run for on an OS X machine with a lot of memory? I guess it's relatively quick since it maps in large chunks (1GB), and doesn't touch any of the mapped pages?

Realm/RLMConstants.h
```diff
@@ -81,6 +81,8 @@ typedef NS_ENUM(NSInteger, RLMError) {
         process which cannot share with the current process due to an
         architecture mismatch. */
     RLMErrorIncompatibleLockFile  = 8,
+    /** Returned by RLMRealm if there is insufficient available mmap address space. */
```
bdash: I don't think it's useful to mention mmap here.

RealmSwift-swift2.0/Error.swift
```diff
@@ -87,9 +89,12 @@ public enum Error: ErrorType {
     /// cannot share with the current process due to an architecture mismatch.
     case IncompatibleLockFile
 
-    /// Returned by RLMRealm if a file format upgrade is required to open the file,
+    /// Error thrown by Realm if a file format upgrade is required to open the file,
     /// but upgrades were explicilty disabled.
     case FileFormatUpgradeRequired
+
+    /// Error thrown by Realm if there is insufficient available mmap address space.
```
bdash: I don't think it's useful to mention mmap here.

bdash:
👍

Realm/Tests/RealmTests.mm
tgoyne: For me, takes 2 seconds for OS X/64-bit sim, 200 ms for 32-bit sim, and 500 ms on a 64-bit device. IMO that's long enough to justify making it device/32-bit sim only.

Realm/RLMConstants.h
jpsim: Could just omit the word from the sentence. Same below.

Realm/Tests/RealmTests.mm
jpsim: Agreed

jpsim:
:+1:

PR #3439, opened 2016-04-14
Add write transactions which do not produce local notifications

tgoyne:
This is very WIP, but it needs some bikeshedding over the API and what to call it.

This address a problem we discovered with fine-grained notifications when the UI is updated before the data model, rather than the reactive approach of the model being the canonical source of truth that the UI merely displays. The simplest motivating use-case is UITableView reordering; when using the built-in functionality for dragging rows around the delegate is merely notified of changes already displayed in the UI, and re-applying those changes in the collection notification block after the Realm is committed will corrupt the order to the UITV. Because notifications are coalesced, users can't reasonably filter out these redundant updates themselves. and even if they could it would be very easy to get wrong.

This makes it so that when beginning a write transaction, a user can specify that the transaction should not produce notifications on the current thread, as the things they want to update with the notifications are already up-to-date. In addition, this variant of `beginWriteTransaction` blocks until all async notifications are ready and delivers them, which is needed to avoid having things getting really weird.

An alternative approach would be to instead let the user pass in the tokens for notification blocks which they want to suppress. This is more flexible and makes it work with commits on a different thread from the notified thread, but maybe would make the API even more complex.


astigsen:
How does Apple's FetchResultsController deal with this problem (if it does at all)?

jpsim:
An API idea: passing in the notification tokens for which you want the transaction to skip.

```swift
func beginWrite(ignoringNotificationTokens: [NotificationToken] = [])
```

```objc
- (void)beginWriteTransactionIgnoringNotificationTokens:(NSArray<RLMNotificationToken *>)tokens;
```

tgoyne:
> How does Apple's FetchResultsController deal with this problem (if it does at all)?

http://stackoverflow.com/questions/1077568/how-to-implement-re-ordering-of-coredata-records

tl;dr is that it updates synchronously when you call `save`, so you can manually track whether or not a set of calls on the delegate are from user-driven changes. This doesn't really work with our model of async updates.

astigsen:
> tl;dr is that it updates synchronously when you call save,so you can manually track whether or not a set of calls on the delegate are from user-driven changes.

That seem unsafe. What happens if changes happen on other threads while you save (and disable updates)? That could make the core data state update with changes from the other save while you have manually disabled updates.

rudedogg:
I hate to ask, but I was wondering if you have an estimate for when this will be merged? It would solve some issues I've run into with a collection view.

tgoyne:
The current implementation doesn't totally work and I've been focusing on bug fixes and ensuring other people are unblocked. I hope to get back to this in the next few days.

TiagoVeloso:
Any updates on this?

JadenGeller:
Closing this since it's being tracked in https://github.com/realm/realm-cocoa/issues/3665 😁

PR #3440, opened 2016-04-15
Align documentation of addNotificationBlock and explain coalescing

### LINKED ISSUE #3435: Document notification coalescing

mrackwitz:
Fixes #3435.

* [x] Swift docs
* [ ] Objective-C docs 

/c @jpsim @icanzilb @TimOliver

### DIFF
--- a/RealmSwift-swift2.0/List.swift
+++ b/RealmSwift-swift2.0/List.swift
@@ -407,9 +407,10 @@ public final class List<T: Object>: ListBase {
 
     The block will be asynchronously called with the initial list, and then
     called again after each write transaction which changes the list or any of
-    the items in the list. You must retain the returned token for as long as
-    you want the results to continue to be sent to the block. To stop receiving
-    updates, call stop() on the token.
+    the items in the list.
+     
+    You must retain the returned token for as long as you want updates to continue
+    to be sent to the block. To stop receiving updates, call stop() on the token.
 
     - parameter block: The block to be called each time the list changes.
     - returns: A token which must be held for as long as you want notifications to be delivered.
--- a/RealmSwift-swift2.0/List.swift
+++ b/RealmSwift-swift2.0/List.swift
@@ -419,6 +419,22 @@ public final class List<T: Object>: ListBase {
     notifications may be coalesced. That can include the notification about the
     initial list.
      
+    This will never be actually received by the passed block, if a write transcation is
+    executed directly after setting up the observation like seen in the example below:
+
+        let person = realm.objects(Person).first!
+        print("dogs.count: \(person.dogs.count)") // => 0
+        person.dogs.addNotificationBlock { dogs in
+            // Only fired once for the example
+            print("dogs.count: \(dogs.count)") // will only print "dogs.count: 1"
+        }
+        try! realm.write {
+            let dog = Dog()
+            dog.name = "Rex"
+            person.dogs.append(dog)
+        }
+        // end of runloop execution context
+     
     You must retain the returned token for as long as you want updates to continue
     to be sent to the block. To stop receiving updates, call stop() on the token.
 
--- a/RealmSwift-swift2.0/List.swift
+++ b/RealmSwift-swift2.0/List.swift
@@ -409,6 +409,11 @@ public final class List<T: Object>: ListBase {
     called again after each write transaction which changes the list or any of
     the items in the list.
      
+    The block is called on the same thread as it was added on, and can only
+    be added on threads which are currently within a run loop. Unless you are
+    specifically creating and running a run loop on a background thread, this
+    normally will only be the main thread.
+     
     You must retain the returned token for as long as you want updates to continue
     to be sent to the block. To stop receiving updates, call stop() on the token.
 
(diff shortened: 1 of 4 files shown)


RealmSwift-swift2.0/List.swift
```diff
@@ -407,9 +407,36 @@ public final class List<T: Object>: ListBase {
...
-    the items in the list. You must retain the returned token for as long as
-    you want the results to continue to be sent to the block. To stop receiving
-    updates, call stop() on the token.
+    the items in the list.
+     
+    The block is called on the same thread as it was added on, and can only
+    be added on threads which are currently within a run loop. Unless you are
+    specifically creating and running a run loop on a background thread, this
+    normally will only be the main thread.
+
```
TimOliver: The block is always called on the same thread on which it was registered, and can only be done so on threads that have an active run loop. Since background threads don't usually have a run loop unless you explicitly create them, this normally only applies to the main thread.

RealmSwift-swift2.0/List.swift
```diff
@@ -407,9 +407,36 @@ public final class List<T: Object>: ListBase {
...
+     
+    The block is called on the same thread as it was added on, and can only
+    be added on threads which are currently within a run loop. Unless you are
+    specifically creating and running a run loop on a background thread, this
+    normally will only be the main thread.
+     
+    Notifications can't be delivered as long as the runloop is blocked by
+    other activity. When notifications can't be delivered instantly, multiple
+    notifications may be coalesced. That can include the notification about the
+    initial list.
```
TimOliver: Notifications cannot be delivered while the runloop is being blocked by other activity. When notifications aren't able to be delivered instantly, they may be coalesced. That can even apply to the initial notification that is sent after the block is registered.

RealmSwift-swift2.0/List.swift
```diff
@@ -407,9 +407,36 @@ public final class List<T: Object>: ListBase {
...
+     
+    This will never be actually received by the passed block, if a write transcation is
+    executed directly after setting up the observation like seen in the example below:
+
```
TimOliver: Notifications are only posted at the start of each run loop iteration. If a notification block is registered, and a write transaction occurs in the same run loop iteration, no notifications from that block will occur. As such, the Realm write transaction in the following code will not trigger a notification.

RealmSwift-swift2.0/Realm.swift
```diff
@@ -443,20 +443,27 @@ public final class Realm {
...
-    are currently within a run loop. Unless you are specifically creating and
-    running a run loop on a background thread, this normally will only be the
-    main thread.
+    independent from the thread or process.
+
+    The block is called on the same thread as it was added on, and can only
+    be added on threads which are currently within a run loop. Unless you are
+    specifically creating and running a run loop on a background thread, this
+    normally will only be the main thread.
+
```
TimOliver: Same as above. :)

RealmSwift-swift2.0/List.swift
mrackwitz:
```diff
-and can only be added on threads which are currently within a run loop.
+and can only be done so on threads that have an active run loop.
```
I don't think that the change from "be added" to "be done so" really makes sense in the sentence. Beside that I think there is a little semantic difference between calling this from within a run loop to calling this from a thread, which has an active run loop.

RealmSwift-swift2.0/List.swift
mrackwitz: Well, that's not quite what is happening and not what I meant. This means as well that this has to improved, I guess. :grin: The first notification about the initial state is never received, because the runloop is still blocked by activity, which is in that case the write transaction directly happening after registering the block. So the initial notification will be coalesced with the notification about the state after the commit.

RealmSwift-swift2.0/List.swift
mrackwitz: Is this really an improvement?
```diff
-When notifications can't be delivered instantly…
+When notifications aren't able to be delivered instantly…
```

mrackwitz:
This was picked up by @tgoyne on #3359, which will require larger updates to the API docs of those methods.

PR #3442, opened 2016-04-15
Deprecate mixed properties

### LINKED ISSUE #3347: Deprecate mixed
Realm's Mixed/Any type is rarely used by end users, undocumented and not generally useful. The Objective-C binding is the only one to support it (it's not available to Swift, React Native or Java).

We should deprecate its support as soon as possible and remove it entirely prior to a 1.0 release.

mrackwitz:
Fixes #3347.

This warns only once on initialization of the shared runtime schema and points out which property cause the warning.
There is no warning yet for dynamic schemas, but as we consider them a more advanced feature, a deprecation warning there might not be necessary at all cost.

#### ToDo

* [x] Add changelog entry
* [x] Decide whether a warning is needed for dynamic schemas

/c @jpsim

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@ x.x.x Release notes (yyyy-MM-dd)
   `-[RLMNotificationToken stop]`.
 * `RLMRealm.path` and `RLMRealm.readOnly` have been deprecated in favor of their
   counterparts on `RLMRealmConfiguration`.
+* Deprecate properties of type `id`/`AnyObject`. This type was rarely used,
+  rarely useful and unsupported in every other Realm binding.
 
 ### Enhancements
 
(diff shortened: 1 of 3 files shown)


jpsim:
@mrackwitz I pushed up an alternative approach in https://github.com/realm/realm-cocoa/compare/mr-deprecate-mixed...jp-mr-deprecate-mixed, which I'd like you to consider. Feel free to either cherry pick those commits onto your branch, or borrow from them more granularly if you'd like. If for whatever reason you don't think that approach is preferable, let me know.

mrackwitz:
@jpsim: Something like that was the first alternative I considered as well, but I think it'd preferable to log not only the property name, but also the class name. What do you think outweights?

jpsim:
I think the property name is enough for users to help identify their mixed properties, but we could add the class name to the property initializers if necessary, while still retaining the other advantages of the alternate approach.

PR #3444, opened 2016-04-15
update core to 0.97.4

### LINKED ISSUE #3194: writeCopyToPath:encryptionKey:error: creates a file that cannot be opened
I'm trying to use writeCopyToPath:encryptionKey:error: to compress my realm file, but I'm noticing that the copied version cannot be opened. I've triple checked to make sure the encryption key is correct, and I have disabled all functions that use Realm to isolate just the copying process. I have also put the copy function in an autoreleasepool as per some suggestions, but to no avail. The original file can be opened with the key, but the copy cannot.

I have tried this on 0.97.0 and 0.98.0

Thanks

jpsim:
Fixes #3194 /cc @tgoyne @bdash

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,7 +12,8 @@ x.x.x Release notes (yyyy-MM-dd)
 
 ### Bugfixes
 
-* None.
+* Fixed a bug that caused encrypted files created using
+  `-[RLMRealm writeCopyToPath:encryptionKey:error:]` to fail to open.
 
 0.98.7 Release notes (2016-04-13)
 =============================================================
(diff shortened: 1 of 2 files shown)


tgoyne:
:+1:

CHANGELOG.md
```diff
@@ -12,7 +12,8 @@ x.x.x Release notes (yyyy-MM-dd)
 
 ### Bugfixes
 
-* None.
+* Fixed a bug that caused encrypted files created using
```
bdash: I'd soften this somewhat since the bug only caused problems for a subset of files in specific conditions.

bdash:
👍

PR #3455, opened 2016-04-19
Rename all instances of primary key to object ID

### LINKED ISSUE #2849: Rename "Primary Key" concept to "Object ID"
We should transition from using the concept of Primary Key, to introducing the following two concepts: ObjectId, and Unique, resulting in the following attribute hierarchy: ObjectId > Unique > Indexed. The fourth attribute, Required, will remain as it is currently.

The reasons to move away from Primary Key as as follows:
1. The concept of a Primary Key will be unclear to anyone starting on Realm that does not have preliminary database experience. In general, we prefer to use traditional Object concepts rather than DB concepts
2. Our current (and preferred) implementation of the Primary Key / ObjectId behavior, is not consistent with that of typical Primary Keys, as having 1 object with a null PK is allowed.

(“ObjectId” is the preferred term over “Identity” because identity can mean other things like use identity / authentication. “ObjectId is the preferred term over “Unique Identifier” because of the overlap with the attribute “Unique”)

Expected subtasks:
- [ ] Rename “Primary Key” to “ObjectId”. Keep both syntaxes for a certain amount of time if possible. This is a breaking API change, so will need to be done before 1.0.
- [ ] Introduce a “Unique” attribute for properties that should be unique but not the Primary Key / ObjectId — or split to separate issue)
- [ ] Add ability to have composite ObjectIds (objects have one ObjectId that is the composite of several properties) —or split to separate issue.
- [ ] Add the ability to have composite unicity constraints (one unicity constraint as a combination of several properties — or split to seperate issue
- [ ] Call out in the docs that “ObjectId” is similar to the concept of a “Primary Key” but highlight the fact one ObjectId per model can be null
- [ ] Call out in the docs the hierarchy: ObjectId > Unique > Indexed

jpsim:
Closes #2849.

This is to be discussed and considered for 0.99 based on the conversations in #2849 and Slack#bindings. /cc @tgoyne @bdash @timanglade


tgoyne:
Would `objectForID:` and `objectInRealm:forID:` be too inconsistent? It'd be a bit less ugh.

jpsim:
That was my first choice too, but opted against it based on the rationale for the full Object ID name in #2849

Realm/RLMObject.h
```diff
@@ -322,7 +332,21 @@ RLM_ASSUME_NONNULL_BEGIN
...
+ Get the single object with the given object ID from the default Realm.
+
+ Returns the object from the default Realm which has the given object ID, or
+ `nil` if the object does not exist. This is slightly faster than the otherwise
+ equivalent `[[SubclassName objectsWhere:@"objectIDPropertyName = %@", key] firstObject]`.
+
+ This method requires that `objectID` be overridden on the receiving subclass.
+
+ @return    An object of the subclass type or nil if an object with the given object ID does not exist.
+ @see       -objectID
```
bdash: Shouldn't this be `+objectID`?

Realm/RLMObject.h
```diff
@@ -322,7 +332,21 @@ RLM_ASSUME_NONNULL_BEGIN
...
+ */
++ (nullable instancetype)objectForObjectID:(nullable id)objectID;
```
bdash: I think including `ObjectID` in the name is redundant given object is already mentioned at the start of the name.

Is *for* the best word to use? `objectForID` sounds awkward to me. `objectWithID` sounds more natural.

jpsim: `objectWithID:` is fine by me.

jpsim:
After much internal debate, we've opted not to do this at this time. Closing.

PR #3457, opened 2016-04-19
Fix a use-after-free when removing KVO observers via an associated object

tgoyne:
Associated objects are actually removed *after* dealloc is called, so supporting this means supporting calling removeObserver:forKeyPath: on an RLMObject after dealloc.


jpsim:
Seems ok. How'd you find out about this? :+1:

tgoyne:
Asan complained about it.

PR #3458, opened 2016-04-20
Deprecate string-based path APIs in favor of NSURL-based ones

### LINKED ISSUE #2593: NSURL-compatible Realm.Configuration APIs
Hello,

After updating to Swift 2.0, all my string path manipulation code didn't compile anymore, and requires the use of `NSURL` instead. (E.g. `stringByAppendingPathComponent`)

With RealmSwift mostly supporting string paths (which I originally thought was a great approach), it doesn't work so well with the new `NSURL` oriented approach Apple is pushing for.

Now, the realm Swift API is absolutely great to work with, and for me to raise an issue like this just means how a minor issue quickly stands out amidst the clean API.

Thanks,
Maz

jpsim:
Closes #2593.

~~Depends on #3455. Diff with that PR: https://github.com/realm/realm-cocoa/compare/jp/api/rename-primary-key-to-object-id...jp/api/url-paths~~ (UPDATE: no longer the case)

I'm torn as to whether or not this is something we should be doing. On the one hand, it's aligned with recent Foundation API changes. On the other, we gain (almost) nothing internally by representing paths as URLs, since it's all string based under the hood.

I still wanted to do this work to fully assess if it's the right move, as we're going through other deprecations.

Thoughts @tgoyne @bdash @mrackwitz?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,31 +9,31 @@ x.x.x Release notes (yyyy-MM-dd)
   `-[RLMResults addNotificationBlock:]` now takes another parameter.
 * The following Objective-C APIs have been deprecated in favor of newer or preferred versions:
 
-| Deprecated API                                         | New API                                                   |
-|:-------------------------------------------------------|:----------------------------------------------------------|
-| `-[RLMRealm removeNotification:]`                      | `-[RLMNotificationToken stop]`                            |
-| `RLMRealmConfiguration.path`                           | `RLMRealmConfiguration.fileURL`                           |
-| `RLMRealm.path`                                        | `RLMRealmConfiguration.fileURL`                           |
-| `RLMRealm.readOnly`                                    | `RLMRealmConfiguration.readOnly`                          |
-| `+[RLMRealm realmWithPath:]`                           | `+[RLMRealm realmWithFileURL:]`                           |
-| `+[RLMRealm writeCopyToPath:error:]`                   | `+[RLMRealm writeCopyToFileURL:encryptionKey:error:]`     |
-| `+[RLMRealm writeCopyToPath:encryptionKey:error:]`     | `+[RLMRealm writeCopyToFileURL:encryptionKey:error:]`     |
-| `+[RLMRealm schemaVersionAtPath:error:]`               | `+[RLMRealm schemaVersionAtFileURL:encryptionKey:error:]` |
-| `+[RLMRealm writeCopyToPath:error:]`                   | `+[RLMRealm writeCopyToFileURL:error:]`                   |
-| `+[RLMRealm writeCopyToPath:encryptionKey:error:]`     | `+[RLMRealm writeCopyToFileURL:encryptionKey:error:]`     |
-| `+[RLMRealm schemaVersionAtPath:error:]`               | `+[RLMRealm schemaVersionAtFileURL:error:]`               |
-| `+[RLMRealm schemaVersionAtPath:encryptionKey:error:]` | `+[RLMRealm schemaVersionAtFileURL:encryptionKey:error:]` |
+| Deprecated API                                         | New API                                               |
+|:-------------------------------------------------------|:------------------------------------------------------|
+| `-[RLMRealm removeNotification:]`                      | `-[RLMNotificationToken stop]`                        |
+| `RLMRealmConfiguration.path`                           | `RLMRealmConfiguration.fileURL`                       |
+| `RLMRealm.path`                                        | `RLMRealmConfiguration.fileURL`                       |
+| `RLMRealm.readOnly`                                    | `RLMRealmConfiguration.readOnly`                      |
+| `+[RLMRealm realmWithPath:]`                           | `+[RLMRealm realmWithURL:]`                       |
+| `+[RLMRealm writeCopyToPath:error:]`                   | `+[RLMRealm writeCopyToURL:encryptionKey:error:]`     |
+| `+[RLMRealm writeCopyToPath:encryptionKey:error:]`     | `+[RLMRealm writeCopyToURL:encryptionKey:error:]`     |
+| `+[RLMRealm schemaVersionAtPath:error:]`               | `+[RLMRealm schemaVersionAtURL:encryptionKey:error:]` |
+| `+[RLMRealm writeCopyToPath:error:]`                   | `+[RLMRealm writeCopyToURL:error:]`                   |
+| `+[RLMRealm writeCopyToPath:encryptionKey:error:]`     | `+[RLMRealm writeCopyToURL:encryptionKey:error:]`     |
+| `+[RLMRealm schemaVersionAtPath:error:]`               | `+[RLMRealm schemaVersionAtURL:error:]`               |
+| `+[RLMRealm schemaVersionAtPath:encryptionKey:error:]` | `+[RLMRealm schemaVersionAtURL:encryptionKey:error:]` |
 
 * The following Swift APIs have been deprecated in favor of newer or preferred versions:
 
-| Deprecated API                                | New API                                      |
-|:----------------------------------------------|:---------------------------------------------|
-| `Realm.removeNotification(_:)`                | `NotificationToken.stop()`                   |
-| `Realm.Configuration.path`                    | `Realm.Configuration.fileURL`                |
-| `Realm.path`                                  | `Realm.Configuration.fileURL`                |
-| `Realm.readOnly`                              | `Realm.Configuration.readOnly`               |
-| `Realm.writeCopyToPath(_:encryptionKey:)`     | `Realm.writeCopyToFileURL(_:encryptionKey:)` |
-| `schemaVersionAtPath(_:encryptionKey:error:)` | `schemaVersionAtFileURL(_:encryptionKey:)`   |
+| Deprecated API                                | New API                                  |
+|:----------------------------------------------|:-----------------------------------------|
+| `Realm.removeNotification(_:)`                | `NotificationToken.stop()`               |
+| `Realm.Configuration.path`                    | `Realm.Configuration.fileURL`            |
+| `Realm.path`                                  | `Realm.Configuration.fileURL`            |
+| `Realm.readOnly`                              | `Realm.Configuration.readOnly`           |
+| `Realm.writeCopyToPath(_:encryptionKey:)`     | `Realm.writeCopyToURL(_:encryptionKey:)` |
+| `schemaVersionAtPath(_:encryptionKey:error:)` | `schemaVersionAtURL(_:encryptionKey:)`   |
 
 ### Enhancements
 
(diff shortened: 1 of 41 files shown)


mrackwitz:
We might also want to take into account that `+[NSURL fileURLWithPath:]` / `-[NSURL initWithFileURLWithPath:]` does a little more work internally than just instantiating a file descriptor. It standardizes the path unless it is absolute and checks whether it represents a directory, where it accesses the file system for paths not ending in a `/` to check that. Differentiating whether the given path is a file or directory is something Core does already internally and brings custom error handling around it. The additional cost for this work that is not necessary by `NSURL` might be neglectable.

But if we'd consider keeping around a string-based API as alternative, we would need to store the string representation of paths, because otherwise it would break our API's getter/setter consistency, when the value returned by `path` differs from what is originally set, because it was standardized by `NSURL`.

That being said, I see how it would be nice to be able to get and set `NSURL` instances from Swift because of the way the Foundation methods related to path manipulation are currently (not) exposed in Swift, but it doesn't convince me that much that I'd be all for phasing out the string-based API.

Realm/RLMRealm.h
```diff
@@ -332,7 +341,22 @@ typedef void (^RLMNotificationBlock)(NSString *notification, RLMRealm *realm);
...
+/**
+ Write an encrypted and compacted copy of the RLMRealm to the given local URL.
+
+ The destination file cannot already exist.
+
+ Note that if this is called from within a write transaction it writes the
+ *current* data, and not data when the last write transaction was committed.
+
+ @param fileURL Local URL to save the Realm to.
+ @param key     64-byte encryption key to encrypt the new file with
```
mrackwitz: Please explain that this entirely optional and doesn't need to be provided.

Realm/RLMRealm.h
```diff
@@ -483,9 +507,22 @@ typedef void (^RLMMigrationBlock)(RLMMigration *migration, uint64_t oldSchemaVer
...
- @return            The version of the Realm at `realmPath` or RLMNotVersioned if the version cannot be read.
+ @return            The version of the Realm at `fileURL` or RLMNotVersioned if the version cannot be read.
+ */
++ (uint64_t)schemaVersionAtPath:(NSString *)realmPath encryptionKey:(nullable NSData *)key error:(NSError **)error DEPRECATED_MSG_ATTRIBUTE("use +[RLMRealm schemaVersionAtFileURL:encryptionKey:error:]");
+
+/**
+ Get the schema version for an encrypted Realm at a given local URL.
+
+ @param fileURL Local URL to a Realm file
+ @param key     64-byte encryption key.
```
mrackwitz: maybe?
>If the file is encrypted, the 64-byte encryption key. Otherwise pass in `nil`.

jpsim:
> But if we'd consider keeping around a string-based API as alternative

Exposing two different ways to get/set the file location is out of the question. We either stick with string-based paths or use NSURL full stop.

On the topic of NSURL doing additional validation work, that's very true. I'm assuming the overhead is negligible compared to the other work of opening a Realm, but I'd have to benchmark that to know for sure.

TimOliver:
I'm really torn on this one since I usually do all of my path manipulation as strings.

But if our goal is to ensure that we're consistent with the direction of the Apple's APIs, then it would definitely be appropriate to move all of our path handling to `NSURL`. While strings are still appropriate in Objective-C, I think the trend in Swift is definitely to push towards `NSURL` as much as possible.

Additionally, I think changing the configuration property from `path` to `fileURL` is definitely a lot more concise and easier to intuit. :)

tgoyne:
Foundation appears to always just use `...WithURL:` even when only a file URL is supported, so the `WithFileURL:` is inconsistent with that and seems more verbose for little gain.

jpsim:
> Foundation appears to always just use `...WithURL:` even when only a file URL is supported, so the `WithFileURL:` is inconsistent with that and seems more verbose for little gain.

Then for consistency, `RLMRealmConfiguration.fileURL` would become `RLMRealmConfiguration.URL`, which is a bit misleading and an all-caps variable isn't exactly pretty.

And as mentioned offline, Foundation does use "file URL" in `+[NSURL fileURLWithPath:]`.

bdash:
I'm of the opinion that `RLMRealmConfiguration.fileURL` is useful to make it clear that it specifies the URL of the file. This is similar to `NSDocument`'s `fileURL` property.

I don't feel strongly about whether to use `FileURL` or just `URL` (or even just `File`) in the method names. I think there's an argument for using just `URL` since it's clear from context that only a file URL makes sense.

jpsim:
Since both @tgoyne and @bdash said they'd prefer just `URL` rather than `FileURL` in method names, I've changed it to just that. I've kept parameter names as `fileURL` though, which will continue to provide strong context to users in autocomplete and documentation.

Any further feedback?

Realm/RLMRealm.h
```diff
@@ -332,7 +341,26 @@ typedef void (^RLMNotificationBlock)(NSString *notification, RLMRealm *realm);
  @param error On input, a pointer to an error object. If an error occurs, this pointer is set to an actual error object containing the error information. You may specify nil for this parameter if you do not want the error information.
  @return YES if the realm was copied successfully. Returns NO if an error occurred.
 */
-- (BOOL)writeCopyToPath:(NSString *)path encryptionKey:(NSData *)key error:(NSError **)error;
+- (BOOL)writeCopyToPath:(NSString *)path encryptionKey:(NSData *)key error:(NSError **)error DEPRECATED_MSG_ATTRIBUTE("use +[RLMRealm writeCopyToURL:encryptionKey:error:]");
+
+/**
+ Write an encrypted and compacted copy of the RLMRealm to the given local URL.
```
tgoyne: Write a compacted and optionally encrypted

Realm/RLMRealm.h
```diff
@@ -483,9 +511,22 @@ typedef void (^RLMMigrationBlock)(RLMMigration *migration, uint64_t oldSchemaVer
                     that describes the problem. If you are not interested in
                     possible errors, pass in `NULL`.
 
- @return            The version of the Realm at `realmPath` or RLMNotVersioned if the version cannot be read.
+ @return            The version of the Realm at `fileURL` or RLMNotVersioned if the version cannot be read.
+ */
++ (uint64_t)schemaVersionAtPath:(NSString *)realmPath encryptionKey:(nullable NSData *)key error:(NSError **)error DEPRECATED_MSG_ATTRIBUTE("use +[RLMRealm schemaVersionAtURL:encryptionKey:error:]");
+
+/**
+ Get the schema version for an encrypted Realm at a given local URL.
```
tgoyne: for a Realm

Realm/RLMRealm.h
```diff
@@ -483,9 +511,22 @@ typedef void (^RLMMigrationBlock)(RLMMigration *migration, uint64_t oldSchemaVer
...
+
+ @param fileURL Local URL to a Realm file
+ @param key     Optional 64-byte encryption key.
```
tgoyne: 64-byte key the file was encrypted with, or `nil` if it is unencrypted.

Realm/RLMRealm.mm
```diff
@@ -162,12 +162,16 @@ + (instancetype)defaultRealm {
...
++ (instancetype)realmWithURL:(NSURL *)fileURL {
     RLMRealmConfiguration *configuration = [RLMRealmConfiguration defaultConfiguration];
-    configuration.path = path;
+    configuration.fileURL = fileURL;
     return [RLMRealm realmWithConfiguration:configuration error:nil];
 }
 
-+ (instancetype)realmWithPath:(NSString *)path
++ (instancetype)realmWithURL:(NSURL *)fileURL
                           key:(NSData *)key
```
tgoyne: realign

Realm/RLMRealm_Private.h
```diff
@@ -61,11 +61,11 @@ void RLMRealmTranslateException(NSError **error);
...
- @see RLMRealm realmWithPath:
- @see RLMRealm realmWithPath:readOnly:error:
- @see RLMRealm realmWithPath:encryptionKey:readOnly:error:
+ @see RLMRealm realmWithURL:
+ @see RLMRealm realmWithURL:readOnly:error:
+ @see RLMRealm realmWithURL:encryptionKey:readOnly:error:
  */
-+ (instancetype)realmWithPath:(NSString *)path
++ (instancetype)realmWithURL:(NSURL *)fileURL
                           key:(NSData *)key
```
tgoyne: realign

Realm/Tests/DynamicTests.m
```diff
@@ -33,7 +33,7 @@ @implementation DynamicTests
 - (void)testDynamicRealmExists {
     @autoreleasepool {
         // open realm in autoreleasepool to create tables and then dispose
-        RLMRealm *realm = [RLMRealm realmWithPath:RLMTestRealmPath()];
+        RLMRealm *realm = [RLMRealm realmWithURL:[NSURL fileURLWithPath:RLMTestRealmPath()]];
```
tgoyne: RLMTestRealmPath should just be changed to RLMTestRealmURL

jpsim:
Ok I just cleaned up tests and other issues you addressed in review @tgoyne.

tgoyne:
👍

CHANGELOG.md
```diff
@@ -3,14 +3,37 @@ x.x.x Release notes (yyyy-MM-dd)
...
+| `RLMRealm.path`                                        | `RLMRealmConfiguration.fileURL`                       |
+| `RLMRealm.readOnly`                                    | `RLMRealmConfiguration.readOnly`                      |
+| `+[RLMRealm realmWithPath:]`                           | `+[RLMRealm realmWithURL:]`                       |
+| `+[RLMRealm writeCopyToPath:error:]`                   | `+[RLMRealm writeCopyToURL:encryptionKey:error:]`     |
+| `+[RLMRealm writeCopyToPath:encryptionKey:error:]`     | `+[RLMRealm writeCopyToURL:encryptionKey:error:]`     |
+| `+[RLMRealm schemaVersionAtPath:error:]`               | `+[RLMRealm schemaVersionAtURL:encryptionKey:error:]` |
+| `+[RLMRealm writeCopyToPath:error:]`                   | `+[RLMRealm writeCopyToURL:error:]`                   |
+| `+[RLMRealm writeCopyToPath:encryptionKey:error:]`     | `+[RLMRealm writeCopyToURL:encryptionKey:error:]`     |
+| `+[RLMRealm schemaVersionAtPath:error:]`               | `+[RLMRealm schemaVersionAtURL:error:]`               |
+| `+[RLMRealm schemaVersionAtPath:encryptionKey:error:]` | `+[RLMRealm schemaVersionAtURL:encryptionKey:error:]` |
```
bdash: You have some duplicate entries on the left, and incorrect entries on the right.

bdash:
👍

CHANGELOG.md
jpsim: thanks, should be fixed now

PR #3459, opened 2016-04-20
Fix the initial ref count for WeakRealmNotifier

### LINKED ISSUE #3270: RealmCoordinator::get_realm() leaks memory
**I don't know how to fix the memory leak.**
screen shot 2016-02-25 at 5 35 27 pm

tgoyne:
Adding the run loop source to the run loop retains it, so the initial refcount should be 0, not 1.

Fixes #3270.


bdash:
👍

PR #3461, opened 2016-04-21
Improve error when appending to array property before the schema is ready

### LINKED ISSUE #3454: Property validation exceptions should include class name
## Goals

Define a RLMObject or Object subclass in the model with an invalid property declaration and try to access a Realm.

## Expected Results

An error message, which points me directly to the name of the class which contains the invalid property.

>Property 'Person.friendNames' is of type 'RLMArray<NSString>' which is not a supported RLMArray object type.
>Property 'Person.id' requires a protocol defining the contained type - example: NSNumber<RLMInt>.
>Property 'Person.friends' requires a protocol defining the contained type - example: RLMArray<Person>.

## Actual Results

An error message, which only includes the property name, but no info where that was defined.

>Property 'friendNames' is of type 'RLMArray<NSString>' which is not a supported RLMArray object type.
>Property 'id' requires a protocol defining the contained type - example: NSNumber<RLMInt>.
>Property 'friends' requires a protocol defining the contained type - example: RLMArray<Person>.

mrackwitz:
Fixes #3454.

* [x] Changelog entry

/c @jpsim @tgoyne

### DIFF
--- a/Realm/RLMArray.mm
+++ b/Realm/RLMArray.mm
@@ -136,6 +136,10 @@ static void RLMValidateMatchingObjectType(RLMArray *array, RLMObject *object) {
     if (!object) {
         @throw RLMException(@"Object must not be nil");
     }
+    if (!object->_objectSchema) {
+        @throw RLMException(@"Object cannot be inserted unless the schema is initialized. "
+                            "This can happen if you try to insert objects into a RLMArray / List from a default value or from an overriden standalone initializer (`init()`).");
+    }
     if (![array->_objectClassName isEqualToString:object->_objectSchema.className]) {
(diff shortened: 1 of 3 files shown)


Realm/RLMArray.mm
```diff
@@ -136,6 +136,10 @@ static void RLMValidateMatchingObjectType(RLMArray *array, RLMObject *object) {
     if (!object) {
         @throw RLMException(@"Object must not be nil");
     }
+    if (!object->_objectSchema) {
+        @throw RLMException(@"Object can not be inserted as long as the schema is not initialized. "
```
jpsim: "Object cannot be inserted until the schema is initialized."

jpsim:
👍

PR #3462, opened 2016-04-21

### LINKED ISSUE #1692: Allow configuring a Realm file to be deleted when a migration is required

kishikawakatsumi:
Closes #1692

This PR is for discussion.

Move calling `[RLMSchema sharedSchema]` earlier than initialization of the Realm, to combine verifying schema to one place.
https://github.com/realm/realm-cocoa/compare/kk-delete-if-migration3?expand=1#diff-b20e44b1f2acd3a4d649580c4a87b52aR349

This PR modifies the binding code but almost done within ObjectStore code.

Will you let me know what you think about it?
@tgoyne @jpsim


kishikawakatsumi:
Talked with @jpsim, we adopt https://github.com/realm/realm-cocoa/pull/3463 approach for now, for less affected existing code.

PR #3463, opened 2016-04-21

### LINKED ISSUE #1692: Allow configuring a Realm file to be deleted when a migration is required

kishikawakatsumi:
Closes #1692

This PR is for discussion.

Deleting Realm when migration is done within the binding code. I know it is not the best solution. But it should work. If it won't make it in time for the schedule, I think it will be worth choosing this implementation.

What do you think about it?
@tgoyne @jpsim


Realm/RLMRealm.mm
```diff
@@ -346,7 +349,21 @@ + (instancetype)realmWithConfiguration:(RLMRealmConfiguration *)configuration er
...
-        realm->_realm = [self openSharedRealm:config error:error];
+        try {
+            realm->_realm = [self openSharedRealm:config error:error];
+        }
+        catch (SchemaMismatchException const& exception) {
+            if (configuration.deleteRealmIfMigrationNeeded) {
+                BOOL success = [[NSFileManager defaultManager] removeItemAtPath:configuration.path error:nil];
+                if (success) {
+                    realm = nil;
+                    return [self realmWithConfiguration:configuration error:error];
```
jpsim: shouldn't this be `realm->_realm = [self openSharedRealm:config error:error];`?

Realm/RLMRealmConfiguration.h
```diff
@@ -71,6 +71,9 @@ RLM_ASSUME_NONNULL_BEGIN
 /// The block which migrates the Realm to the current version.
 @property (nonatomic, copy, nullable) RLMMigrationBlock migrationBlock;
 
+/// Clear all tables if migration is needed and re-create with the new schema.
```
jpsim: I think this is clearer and more accurate:

> Recreate the Realm file with the new schema if a migration is required.

RealmSwift-swift2.0/RealmConfiguration.swift
```diff
@@ -58,6 +58,8 @@ extension Realm {
         - parameter readOnly:           Whether the Realm is read-only (must be true for read-only files).
         - parameter schemaVersion:      The current schema version.
         - parameter migrationBlock:     The block which migrates the Realm to the current version.
+        - parameter deleteRealmIfMigrationNeeded: If `true`, all tables will be cleared and recreated
```
jpsim: > If `true`, recreate the Realm file with the new schema if a migration is required.

RealmSwift-swift2.0/RealmConfiguration.swift
```diff
@@ -120,6 +123,9 @@ extension Realm {
         /// The block which migrates the Realm to the current version.
         public var migrationBlock: MigrationBlock? = nil
 
+        /// Delete the Realm file if migration is needed and re-create with the new schema.
```
jpsim: This comment is inconsistent with the Objective-C version.

Realm/RLMRealm.mm
```diff
@@ -277,6 +277,9 @@ + (SharedRealm)openSharedRealm:(Realm::Config const&)config error:(NSError **)ou
     try {
         return Realm::get_shared_realm(config);
     }
+    catch (SchemaMismatchException const& ex) {
+        throw;
```
jpsim: Rather than unconditionally throw here, I think you need to do the following:

```cpp
if (config.delete_realm_if_migration_needed) {
    throw;
} else {
    RLMRealmTranslateException(outError);
}
```

jpsim:
A few questions:

1. Would it be a good idea to log something when Realm files are automatically deleted?
2. If a migration is required and two threads or processes are concurrently accessing the file, could this lead to a crash while the file is deleted from under the accessing thread/process?
3. Would it be wise to add a warning to the documentation, recommending this for development use only?

tgoyne:
Multiple threads initializing at once will work fine since we hold a global lock during init. Multiple processes could run into problems, though. I think it's fine to just document that since it's unlikely to be an actual issue for people.

Realm/RLMRealm.mm
```diff
@@ -375,6 +392,17 @@ + (instancetype)realmWithConfiguration:(RLMRealmConfiguration *)configuration er
                 }
 
                 RLMRealmSetSchemaAndAlign(realm, schema);
+            } catch (SchemaMismatchException const& exception) {
+                if (configuration.deleteRealmIfMigrationNeeded) {
+                    BOOL success = [[NSFileManager defaultManager] removeItemAtPath:configuration.path error:nil];
```
tgoyne: removeItemAtURL:configuration.fileURL

kishikawakatsumi:
Feedback addressed.

kishikawakatsumi:
> Would it be wise to add a warning to the documentation, recommending this for development use only?

I think like this feature would be worth to use for purgeable or re-creatable data (e.g. API response cache). I don't think it would be reasonable recommending this for development use only.

jpsim:
> I think like this feature would be worth to use for purgeable or re-creatable data

Fair enough, I just wanted to make sure it was considered. How about logging when the file's deleted though?

jpsim:
By the way, SwiftLint caught missing documentation: the `deleteRealmIfMigrationNeeded` parameter documentation was added to the deprecated `Realm.Configuration` initializer, but not the new one...

Since we won't be adding this parameter to the deprecated initializer, you can remove the documentation from that init and move it to the non-deprecated init.

tgoyne:
If we're endorsing it for use in non-dev scenarios then the docs need something like "WARNING: this is not safe to use if the Realm file is being shared between multiple processes"

kishikawakatsumi:
@jpsim Done!

Realm/RLMRealm.mm
```diff
@@ -382,6 +402,17 @@ + (instancetype)realmWithConfiguration:(RLMRealmConfiguration *)configuration er
                 }
 
                 RLMRealmSetSchemaAndAlign(realm, schema);
+            } catch (SchemaMismatchException const& exception) {
+                if (configuration.deleteRealmIfMigrationNeeded) {
+                    BOOL success = [[NSFileManager defaultManager] removeItemAtURL:configuration.fileURL error:nil];
+                    if (success) {
+                        realm = nil;
+                        return [self realmWithConfiguration:configuration error:error];
```
jpsim: This code path isn't tested and worries me a bit. I suspect it should be invalidating the Realm if `beganReadTransaction` was set.

Could you add a unit test and determine if the realm should be invalidated before set to `nil`?

jpsim:
A basic Swift test would be nice, so would a test of the second codepath I mentioned inline.

Realm/RLMRealm.mm
kishikawakatsumi: I see. I added a test case for the code path.
I'll investigate for invalidating.

Realm/Tests/MigrationTests.mm
```diff
@@ -407,6 +409,57 @@ - (void)testChangingIntSizeDoesNotRequireMigration {
     [self assertNoMigrationRequiredForChangeFrom:@[from] to:@[to]];
 }
 
+- (void)testDeleteRealmWithSetCustomSchamaIfMigrationNeeded {
```
jpsim: "Schama" -> "Schema"

Realm/RLMRealm.mm
kishikawakatsumi: I think it doesn't need to invalidate. Because schema verification error occurred before updating the schema. A transaction starts when updating the schema after schema verification was passed.

Any thought? @jpsim @tgoyne

tgoyne: `invalidate` wouldn't do anything relevant. I think it shouldn't be needed with what we currently do before hitting this point, but calling `realm->_realm->close()` before setting `realm` to `nil` wouldn't hurt and would make things a bit more robust for future changes.

jpsim:
Add a Swift test and I'm happy with this. @tgoyne @bdash would be great to get this in the next release, so I'd appreciate if you could take time to review

kishikawakatsumi:
I added `realm->_realm->close()` before setting `realm` to `nil`.
I added two test cases for Swift.

tgoyne:
👍

jpsim:
👍

marchy:
Is there any way to be notified of when this happens? ie: callback to run custom code, or try/catch clause. We need to clear out some additional data (ie: user defaults) when this happens but currently the error is fully handled in the Realm code and I don't see any API surface to know when the Realm deletion happens.

PR #3466, opened 2016-04-22
Remove `addNotificationBlock(_:)` from `RealmCollectionType`.

jpsim:
Cherry-picked from #3419.

> It no longer compiles now that `Results` is not final. It was added as part of the fine-grained notifications work and has not yet shipped in a release, so it's ok to remove. We can look into bringing it back in the future if we can find a way to do so.

/cc @bdash


bdash:
👍

PR #3467, opened 2016-04-22
Deprecate old notification block APIs

jpsim:
As best I can tell, this was just missed in #3359 /cc @tgoyne 

I think this is compatible with #3419 but not 100% sure just yet. /cc @bdash

### DIFF
--- a/Realm/RLMCollection.h
+++ b/Realm/RLMCollection.h
@@ -20,7 +20,7 @@
 
 RLM_ASSUME_NONNULL_BEGIN
 
-@class RLMRealm, RLMResults, RLMObject, RLMSortDescriptor, RLMNotificationToken;
+@class RLMRealm, RLMResults, RLMObject, RLMSortDescriptor, RLMNotificationToken, RLMCollectionChange;
 
 /**
  A homogenous collection of `RLMObject`s like `RLMArray` or `RLMResults`.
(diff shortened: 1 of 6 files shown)


mrackwitz:
I think, we should deprecate these APIs you're removing here first to completely remove them with 1.0 to give our users better assistance migrating through the versions and adopting the changes.

jpsim:
The problem with deprecating these APIs rather than removing them is that the Swift compiler is a real pain to work with when methods have ambiguous overloads, which is how I discovered this in the first place.

For example, the following won't compile if we keep both overloads (even if one is deprecated):

```swift
let token = results.addNotificationBlock { changes in }
```

For it to work, you need to be explicit with the type:

```swift
let token = results.addNotificationBlock { (changes: RealmCollectionChange) in }
```

tgoyne:
The old method should certainly be marked as deprecated, but having to explicitly give the type of an argument to a block is not exactly an unusual experience in Swift.

jpsim:
> The old method should certainly be marked as deprecated

If we do this, we need to do it consistently for the other notification block APIs that have been changed.

> but having to explicitly give the type of an argument to a block is not exactly an unusual experience in Swift.

I'd like to think we've set ourselves to a higher bar of quality than the usual experiences, for example supporting Swift versions long after even Apple doesn't.

mrackwitz:
As the method has the same name just another type signature, the compiler will do likely do at least an acceptable job at hinting how that has to change. I'm all for completely removing it already now, because it's a breaking change in one way or another, if you haven't specified the types of the expected arguments before. So a :+1: from my side.

tgoyne:
It's not a breaking change either way. The one argument version is ambiguous due to that the single arg could be a tuple (which is something that they're killing in Swift 3), but the two-arg version is unambiguously the old one.

> If we do this, we need to do it consistently for the other notification block APIs that have been changed.

If there's a path to doing so that doesn't lead to long-term unpleasantness then I'd be all for it, but I don't have any ideas for how to do so other than using a different name, and that just gives us a breaking change later when we decide to rename it back to `addNotificationBlock`.

> I'd like to think we've set ourselves to a higher bar of quality than the usual experiences, for example supporting Swift versions long after even Apple doesn't.

Yes, and I consider doing our best to avoid breaking existing code without a warning period to be an important part of making a library pleasant to use. Trying to update your deps and having compilation fail is unpleasant.

jpsim:
Ok, there's definitely a nice-ish way to keep the old Swift one deprecated. I'm doing that now.

jpsim:
Done.

bdash:
👍

tgoyne:
👍

mrackwitz:
:+1:

PR #3468, opened 2016-04-22
Fixes warning on Xcode 7.3

icanzilb:
changed typealias to associatedtype


jpsim:
Sorry @icanzilb, but we need to continue supporting previous versions of Swift 2.x, so we can't do this just yet.

icanzilb:
fair enough :)

PR #3478, opened 2016-04-25
Fix a few async notifier issues

tgoyne:
This may or may not fix any actual issues hit by users. The newly added test hits the "m_advancer_sg->get_version_of_current_transaction() == new_notifiers.front()->version()" assertion failure, but does so in a way that is very unlikely to be the same as how users are hitting it (at the minimum, to hit it via the public API would require active runloops on multiple threads).

The problem fixed in the second commit is almost certainly not relevant to any actual users.

@bdash @jpsim

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,9 @@ x.x.x Release notes (yyyy-MM-dd)
 
 ### Bugfixes
 
-* None.
+* Fix a scenario that could lead to the assertion failure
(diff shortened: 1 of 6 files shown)


bdash:
👍

tgoyne:
It turns out it actually is possible to hit the problematic scenario here without doing anything especially weird by doing Query -> Write -> Query on the main thread and having the background worker run immediately in between the two Results objects being destroyed when the autorelease pool is drained.

PR #3479, opened 2016-04-25
Update -[AsyncTests testErrorHandling] to work with core 0.100

tgoyne:
Also works with older versions. Deleting the file no longer works due to that the realm file is no longer actually opened multiple times, but setting the max open files to zero will work as long as each SharedGroup opens any files at all.

@bdash


jpsim:
Cool! 👍

PR #3480, opened 2016-04-26
Added New Linking Objects Syntax to Example Apps

TimOliver:
A very minor PR that brings the 'Backlinks' sample app in line with the new linking objects properties feature syntax.

/cc @bdash @jpsim


bdash:
Thanks, Tim! :+1:

PR #3483, opened 2016-04-26
Fix an assertion failure when a ListNotifier's LV was already deleted

tgoyne:
Fixes https://github.com/realm/realm-core/issues/1728.


bdash:
:+1:

PR #3484, opened 2016-04-26
Fix a debug assertion failure when deleting objects in row order

tgoyne:
Dispatching to erase() from CollectionChangeBuilder::move_over() doesn't actually work as erase() doesn't update m_move_mapping (since m_move_mapping only makes sense for unordered deletions). I'm not sure if this would ever actually cause any functional problems in non-debug builds, but I hit it while trying to reproduce other issues.


jpsim:
Seems reasonable to me 👍

PR #3488, opened 2016-04-27
[Examples] Fix linker error for RACTableView target

### LINKED ISSUE #3482: the RACTableView example of ios-xcode7-objc has a build error
## Goals

What do you want to achieve?
> Try to run RACTableView example

## Expected Results

What did you expected to happen?
> no errors

## Actual Results

What did happened instead?  
>A building error happened, said that Xcode can't find the ReactiveCocoa library.

## Steps to Reproduce

>Just 'pod install', then open the 'RealmExamples.xcworkspace', run.

## Code Sample
NO.

# Version of Realm and Tooling
Realm version: 0.98.8
Xcode version: 7.3

iOS/OSX version: 10.11

## Solution:
The error is the 'Per-configuration Build Products Path' in 'Build Settings', the path is '$(BUILD_DIR)/$(CONFIGURATION)$(EFFECTIVE_PLATFORM_NAME)-static'. If remove the suffix '-static' , it will be OK.

mrackwitz:
The scheme for the ReactKitTableView was not shared and must be autocreated, before it can be built.

CocoaPods (0.39) requires that implicit dependencies are built.
Furthermore the `LIBRARY_SEARCH_PATHS` must be set to the default build directory, because we use a custom one (with the suffix `-static`) to link against the Realm static iOS framework while CocoaPods builds the targets into the default build directory.

Fixes #3482.


jpsim:
👍 would also be nice to actually test it as part of `build.sh examples-ios` though

mrackwitz:
Good that you bring that up. So far we ignore the `Podfile.lock` and the `Pods` directory. I'd be in favor of committing them to our repo to avoid slowing down CI by having to run `pod install` to build the examples, but that would obviously mean that we would vendor third-party code in our repository. Are we okay with that?

jpsim:
I doubt that running `pod install` on CI will appreciably slow it down and would prefer that to vendoring 3rd party code used for an example into our repo.

mrackwitz:
CI failures seem to be unrelated, while there are quite a few additional Swift deprecation warnings through the 3rd party libs being actually build now.

I made sure that the examples will built with CocoaPods 0.39 and 1.0.0.beta.8.

jpsim:
👍

PR #3503, opened 2016-04-29
Update to core 0.100.0

### LINKED ISSUE #875: NSDate being truncated to the second
NSDates are being truncated to a whole second when being stored and then retrieved from the realm. Looking at the date setter/getter, the NSTimeInterval is being cast to a std::time_t. NSTimeInterval is a double with sub second precision. My C++ is rusty but I believe time_t is a integer value of whole seconds. Could the underlying storage of NSDate be changed to a double?

### LINKED ISSUE #2159: Crashing during a write transaction makes other processes "hang" indefinitely

tgoyne:
Brings nanosecond-precision dates, robust mutex emulation for the write lock, and reduced memory usage due to sharing memory mappings between threads.

## FILES WRITTEN BY THIS VERSION OF REALM CANNOT BE OPENED BY PREVIOUS VERSIONS OF REALM

Fixes #2159. Fixes #875.


Realm/ObjectStore/property.hpp
```diff
@@ -23,25 +23,15 @@
...
-        Any    = 6,
-        /** Date type: NSDate */
-        Date   = 7,
-        /** Object type. See [Realm Models](https://realm.io/docs/objc/latest/#models) */
+        Any    = 6, // Deprecated and will be removed in the future
+        Date   = 8,
         Object = 12,
-        /** Array type. See [Realm Models](https://realm.io/docs/objc/latest/#models) */
         Array  = 13,
         /** Linking objects type. See [Realm Models](https://realm.io/docs/objc/latest/#models) */
```
bdash: Probably worth nuking this comment as well.

Realm/Tests/ObjectTests.m
```diff
@@ -1526,7 +1575,7 @@ - (void)testIndex
     XCTAssertTrue(boolProperty.indexed, @"indexed property should have an index");
 
     RLMProperty *dateProperty = schema[IndexedObject.className][@"dateCol"];
-    XCTAssertTrue(dateProperty.indexed, @"indexed property should have an index");
+    XCTAssertFalse(dateProperty.indexed, @"non-indexed property should have an index");
```
bdash: s/should have an index/should not have an index/ ?

Realm/Tests/RealmTests.mm
```diff
@@ -166,14 +168,57 @@ - (void)testFilePermissionDenied {
...
+
+- (void)testFileFormatUpgradeRequiredButReadOnly {
+    RLMRealmConfiguration *config = [RLMRealmConfiguration defaultConfiguration];
+    config.readOnly = true;
+
+    NSURL *bundledRealmURL = [[NSBundle bundleForClass:[RealmTests class]] URLForResource:@"fileformat-pre-null" withExtension:@"realm"];
+    [NSFileManager.defaultManager copyItemAtURL:bundledRealmURL toURL:config.fileURL error:nil];
 
-    RLMAssertThrowsWithCodeMatching([RLMRealm realmWithConfiguration:config error:nil], RLMErrorFileFormatUpgradeRequired);
+    RLMAssertThrowsWithCodeMatching([RLMRealm realmWithConfiguration:config error:nil], RLMErrorFileAccess);
```
jpsim: unfortunate that core no longer throws a "format upgrade required" exception here. How should we distinguish between a read error and a file format upgrade error in the browser?

Realm/ObjectStore/index_set.cpp
```diff
@@ -414,7 +414,9 @@ void IndexSet::add_shifted_by(IndexSet const& shifted_by, IndexSet const& values
     copy(old_it, old_end, std::back_inserter(builder));
     m_data = builder.finalize();
 
-    REALM_ASSERT_DEBUG((size_t)std::distance(as_indexes().begin(), as_indexes().end()) == expected);
+#ifdef REALM_DEBUG
+    REALM_ASSERT((size_t)std::distance(as_indexes().begin(), as_indexes().end()) == expected);
```
jpsim: why this change?

Realm/RLMUtil.hpp
```diff
@@ -157,14 +157,28 @@ static inline realm::BinaryData RLMBinaryDataForNSData(__unsafe_unretained NSDat
...
+    timeInterval -= seconds;
+    seconds += static_cast<int64_t>(NSTimeIntervalSince1970);
+
+    // Seconds and nanoseconds have to have the same sign
+    if (timeInterval < 0 && seconds > 0) {
+        ++timeInterval;
+        --seconds;
+    }
+    auto nanoseconds = static_cast<int32_t>(timeInterval * 1'000'000'000.0);
+    return {seconds, nanoseconds};
```
rrrlasse: A) You can use lrint(timeInterval * 1'000'000'000.0) to round to nearest, instead of static_cast.

B) I'm not convinced that nanosecond part is always guaranteed to be within our valid range due to the weird IEEE pitfalls. If it's outside range, Timestamp will assert. So maybe guard against that before construction

Else :+1:

rrrlasse: Oh, you also need to check if the input NSDate is within valid range for a Timestamp. Timestamp has a smaller range. But it's perhaps more of a theoretical problem...

Realm/Tests/ObjectTests.m
```diff
@@ -1526,7 +1575,7 @@ - (void)testIndex
     XCTAssertTrue(boolProperty.indexed, @"indexed property should have an index");
 
     RLMProperty *dateProperty = schema[IndexedObject.className][@"dateCol"];
-    XCTAssertTrue(dateProperty.indexed, @"indexed property should have an index");
+    XCTAssertFalse(dateProperty.indexed, @"non-indexed property should no have an index");
```
bmunkholm: no -> not

RealmSwift-swift2.0/Tests/SwiftTestObjects.swift
```diff
@@ -251,11 +251,11 @@ class SwiftIndexedPropertiesObject: Object {
     dynamic var dataCol = NSData()
 
     override class func indexedProperties() -> [String] {
-        return ["stringCol", "intCol", "int8Col", "int16Col", "int32Col", "int64Col", "boolCol", "dateCol"]
+        return ["stringCol", "intCol", "int8Col", "int16Col", "int32Col", "int64Col", "boolCol"]
```
bmunkholm: date can still be indexed?

RealmSwift-swift2.0/Object.swift
```diff
@@ -160,7 +160,7 @@ public class Object: RLMObjectBase {
 
     /**
     Return an array of property names for properties which should be indexed. Only supported
-    for strings, integers, booleans and NSDate properties.
+    for string, integer and boolean properties.
```
bmunkholm: date can still be indexed?

Realm/Tests/ObjectTests.m
```diff
@@ -87,7 +88,7 @@ @interface IndexedObject : RLMObject
 @implementation IndexedObject
 + (NSArray *)indexedProperties
 {
-    return @[@"stringCol", @"integerCol", @"intCol", @"longCol", @"longlongCol", @"boolCol", @"dateCol", @"optionalIntCol", @"optionalBoolCol"];
+    return @[@"stringCol", @"integerCol", @"intCol", @"longCol", @"longlongCol", @"boolCol", @"optionalIntCol", @"optionalBoolCol"];
```
bmunkholm: date can still be indexed?

Realm/RLMObject.h
```diff
@@ -233,8 +233,8 @@ RLM_ASSUME_NONNULL_BEGIN
 #pragma mark - Customizing your Objects
 
 /**
- Return an array of property names for properties which should be indexed. Only supported
- for strings, integers, booleans and NSDate properties.
+ Return an array of property names for properties which should be indexed.
+ Only supported for string, integer and  boolean properties.
```
bmunkholm: date can still be indexed?

Realm/Tests/RealmTests.mm
tgoyne: Core never threw "format upgrade required" for this (it's an entirely new test). The browser doesn't open files in read-only mode.

Realm/ObjectStore/index_set.cpp
tgoyne: https://github.com/realm/realm-core/pull/1630 made it so that the body of an assert always needs to be compilable.

RealmSwift-swift2.0/Tests/SwiftTestObjects.swift
tgoyne: I disabled it as it was throwing a LogicError saying that it wasn't supported, but now it works fine so I guess I just had a bug elsewhere.

Realm/RLMUtil.hpp
danielpovlsen: Be were that the double should be truncated, *not* rounded.

Realm/RLMUtil.hpp
```diff
@@ -157,14 +157,37 @@ static inline realm::BinaryData RLMBinaryDataForNSData(__unsafe_unretained NSDat
...
+        return {std::numeric_limits<int64_t>::max(), 1'000'000'000 - 1};
+    if (timeInterval - NSTimeIntervalSince1970 < std::numeric_limits<int64_t>::min())
+        return {std::numeric_limits<int64_t>::min(), -1'000'000'000 + 1};
+
+    auto seconds = static_cast<int64_t>(timeInterval);
+    timeInterval -= seconds;
+    seconds += static_cast<int64_t>(NSTimeIntervalSince1970);
+
+    // Seconds and nanoseconds have to have the same sign
+    if (timeInterval < 0 && seconds > 0) {
```
bdash: The variable name `timeInterval` threw me a little when reading this part of the code as the variable now only contains the sub-second portion of the original time interval. I was a little confused as to how the two adjustments below would result in the values having the same sign until I realized this.

Realm/Tests/ObjectTests.m
```diff
@@ -651,26 +652,117 @@ - (void)testObjectSubclass {
...
+}
+
+- (void)testExactRepresentationOfDatesAroundEpoch {
+    NSDate *date = [NSDate dateWithTimeIntervalSince1970:0];
+    NSTimeInterval time = date.timeIntervalSinceReferenceDate;
+    testDatesInRange(time - .001, time + .001, ^(NSDate *d1, NSDate *d2) {
+        XCTAssertEqualObjects(d1, d2);
     });
-    dateObject.dateCol = date;
+}
```
bdash: Would it be worth testing the exact representation of dates around `NSDate`'s epoch as well?

bdash:
👍

Realm/Tests/ObjectTests.m
```diff
@@ -651,26 +652,145 @@ - (void)testObjectSubclass {
...
+
+- (void)testExactRepresentationOfDatesAroundReferenceDate {
+    RLMRealm *realm = [RLMRealm defaultRealm];
+    [realm beginWriteTransaction];
+
+    NSDate *zero = [NSDate dateWithTimeIntervalSinceReferenceDate:0];
+    DateObject *dateObject = [DateObject createInRealm:realm withValue:@[zero]];
+    XCTAssertEqualObjects(dateObject.dateCol, date);
```
jpsim: CI failed because `date` isn't declared in this test.

jpsim:
👍 other than missing variable in test.

PR #3505, opened 2016-04-30
re-add addNotificationBlock(_:) vending 'Self' to RealmCollectionType

jpsim:
this was previously removed because in preparation for `Results` to no longer be `final`, but we ended up going with a different approach. /cc @bdash @tgoyne

### DIFF
--- a/RealmSwift-swift2.0/RealmCollectionType.swift
+++ b/RealmSwift-swift2.0/RealmCollectionType.swift
@@ -297,6 +297,61 @@ public protocol RealmCollectionType: CollectionType, CustomStringConvertible {
 
     // MARK: Notifications
 
+    /**
+     Register a block to be called each time the collection changes.
+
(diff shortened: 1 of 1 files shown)


bdash:
#3466 also removed:
```
override func addNotificationBlock(wrapper: Wrapper, block: (RealmCollectionChange<Wrapper>) -> ())
```

from `_AnyRealmCollection`. Is that something we should also be bringing back?

jpsim:
`addNotificationBlock(wrapper:block:)` is no longer necessary in `_AnyRealmCollection` anymore because all `RealmCollectionType`s declare and implement their own specialization of `addNotificationBlock(block: (RealmCollectionChange<Self>) -> Void) -> NotificationToken`, so they don't need to "proxy" through this intermediate method.

bdash:
:+1:

PR #3511, opened 2016-05-02
Fix parameters in doc comments

woxtu:


jpsim:
Thanks @woxtu, but this is still incorrect since it uses the Swift 1 style documentation comment format. I've pushed a fix here: #3512.

PR #3512, opened 2016-05-02
fix doc comments in RealmCollectionChange.Update

jpsim:
continued from #3511.


jpsim:
Unfortunately there's no official way to document Swift enum associated values, and no thus way that jazzy will semantically parse, so even this is missing the first argument... if anyone has better ideas on how this could be documented, please share.

bdash:
Seems fine. 👍

PR #3515, opened 2016-05-02
Fix RLMArray -addNotificationBlock: missing error

plokijuhhb:


jpsim:
Thanks for the PR, @plokijuhhb, this change looks right to me. Could you please sign our CLA? https://github.com/realm/realm-cocoa/blob/master/CONTRIBUTING.md#cla

We'll need you to do this before we can accept any patches.

plokijuhhb:
Thanks for your reply, I already sign you CLA ,Please check it.

jpsim:
I see it, thanks for your contribution!

PR #3516, opened 2016-05-02
Fix an assertion failure in IndexSet::do_add()

tgoyne:
ChunkedVector::erase() was returning the wrong iterator when a chunk was removed entirely, resulting in the next operation using that iterator to fail.

The [objectstore version of this commit](https://github.com/realm/realm-object-store/commit/f277639cb8386d60e79284b551a5c32df60f06ba) has a basic test case that hits this, but no obj-c test because hitting it via the obj-c api would be tremendously awkward as that codepath is used only in the changeset merging logic.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,8 @@ x.x.x Release notes (yyyy-MM-dd)
 
 ### Bugfixes
 
-* None.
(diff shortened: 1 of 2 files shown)


jpsim:
👍 changelog?

Issue #3494, opened 2016-04-28
Realm triggers modification change notifications where there should be none.


kunalsood:
## Goals

As an example, an RSS Reader App, which has a `Feed` object, and a `FeedItem` object.
`Feed` has a to-many relationship `items` to `FeedItem`.
`FeedItem` has a to-one relationship `feed` to `Feed`.

Every time I do a feed refresh, new feed items are downloaded, and imported into the database. Both relationships are created explicitly.

My goal is to display FeedItem objects to the user, using RLMResults to populate my UI, and make animated changes to the UI when new items are downloaded using RLMResults' change notifications API.

## Expected Results

RLMResults change notifications will only include change-sets for items that have been inserted, deleted, or modified.

## Actual Results

RLMResults change notifications includes modification change-sets for every `FeedItem` object already in the database before the write transaction that added new objects was committed.

## Steps to Reproduce

Create new "Single View Application" project in Xcode, follow the steps to add Realm framework to the project, copy & paste the code sample below into the root view controller, run the app & watch the logs.
First it will produce logs like this..

```
2016-04-28 13:55:09.682 RLMIssueSample[10785:4734055] 0 results at first load
2016-04-28 13:55:09.688 RLMIssueSample[10785:4734055] Staring Import
2016-04-28 13:55:09.738 RLMIssueSample[10785:4734097] Committed Write Transaction, Saved 100 total items in Realm
2016-04-28 13:55:09.739 RLMIssueSample[10785:4734055] 0 deletions, 100 insertions, 0 modifications, 100 total objects
```
Which is fine as there was nothing in the DB yet, so, no modification change-sets. But wait for the next set of change-set notifications & consequential logs. Which will show something like...

```
2016-04-28 13:55:09.804 RLMIssueSample[10785:4734097] Committed Write Transaction, Saved 200 total items in Realm
2016-04-28 13:55:09.811 RLMIssueSample[10785:4734055] 0 deletions, 100 insertions, 100 modifications, 200 total objects
2016-04-28 13:55:09.834 RLMIssueSample[10785:4734097] Committed Write Transaction, Saved 300 total items in Realm
2016-04-28 13:55:09.867 RLMIssueSample[10785:4734055] 0 deletions, 100 insertions, 200 modifications, 300 total objects
2016-04-28 13:55:09.899 RLMIssueSample[10785:4734097] Committed Write Transaction, Saved 400 total items in Realm
[...]

PR #3522, opened 2016-05-03
Improve change calculation performance for nontrivial object graphs

tgoyne:
Skip doing any checking at all if none of the tables reachable from the root table have been modified (which can happen if the table version was bumped due to insertions, unrelated backlinks, or unlinked-to rows being deleted in linked tables).

Add cycle checking rather than relying on the max depth to handle it, as the worst case was O(N^16) if the cycle involved a LinkList of size N.

Track which rows have been confirmed to have not been modified.

Cache the information about the links for each of the relevant tables as checking the table schema can get somewhat expensive.

Fixes the performance part of #3494, with the specific scenario that I tested going from not finishing after 30 minutes to taking ~400ms.

### DIFF
--- a/Realm/ObjectStore/impl/collection_notifier.cpp
+++ b/Realm/ObjectStore/impl/collection_notifier.cpp
@@ -26,36 +26,125 @@
 using namespace realm;
 using namespace realm::_impl;
 
-bool TransactionChangeInfo::row_did_change(Table const& table, size_t idx, int depth) const
+std::function<bool (size_t)>
+CollectionNotifier::get_modification_checker(TransactionChangeInfo const& info,
+                                             Table const& root_table)
 {
-    if (depth > 16)  // arbitrary limit
+    // First check if any of the tables accessible from the root table were
+    // actually modified. This can be false if there were only insertions, or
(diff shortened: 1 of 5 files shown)


Realm/ObjectStore/impl/collection_notifier.cpp
```diff
@@ -26,38 +26,127 @@
...
+}
+
+void DeepChangeChecker::find_related_tables(std::vector<RelatedTable>& out, Table const& table)
+{
+    auto table_ndx = table.get_index_in_group();
+    if (any_of(begin(out), end(out), [=](auto& tbl) { return tbl.table_ndx == table_ndx; }))
+        return;
 
-    for (size_t i = 0, count = table.get_column_count(); i < count; ++i) {
+    size_t info = out.size();
```
bdash: `info` is not a good name for this.

bdash:
Is this an area it's worth adding a performance test for?

👍

tgoyne:
I wrote a perf test for it, but the setup time to get enough objects for the change calculations to actually take a measurable amount of time is pretty unreasonable now (it used to only need about 10...).

PR #3523, opened 2016-05-03
Fix BETWEEN queries that traverse List properties to ensure that a single related object satisfies the BETWEEN criteria

bdash:
We previously allowed different objects in the array to satisfy the lower and upper bounds, which is not consistent with how Foundation evaluates predicates.

/cc @jpsim @tgoyne

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,9 @@ x.x.x Release notes (yyyy-MM-dd)
 
 ### Bugfixes
 
-* None.
+* Fix `BETWEEN` queries that traverse `RLMArray`/`List` properties to ensure that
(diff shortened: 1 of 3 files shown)


jpsim:
Are there performance implications to no longer be using `Query::between`? As far as I can tell that's not really optimized over two less/greater than constraints, but I'm not sure.

If that's the case then 👍.

bdash:
`Query::between` has the same performance as us generating `>=` and `<=` comparisons in a group. I did realize that my change would slow down queries that don't involve arrays or backlinks though. I pushed a fix for that in afd459f74282d3a9504d7a2383428a1aad73279f.

tgoyne:
👍

PR #3524, opened 2016-05-04
fix assertion failure in clean_up_dead_notifiers

jpsim:
we shouldn't be reading `m_advancer_sg->get_transact_stage()` without first checking if `m_advancer_sg` is set /cc @tgoyne 

I think this should fix https://secure.helpscout.net/conversation/199189388/4080 (internal link)

### DIFF
--- a/Realm/ObjectStore/impl/realm_coordinator.cpp
+++ b/Realm/ObjectStore/impl/realm_coordinator.cpp
@@ -299,8 +299,8 @@ void RealmCoordinator::clean_up_dead_notifiers()
         }
     }
     if (swap_remove(m_new_notifiers)) {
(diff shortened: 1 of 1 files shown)


tgoyne:
The advancer SG is created when the first notifier is added to m_new_notifiers and is never destroyed, so I think the assert is in the correct place and the `&& m_advancer_sg` check is just pointless (it's required for the `m_notifier_sg` case a few lines due to that shared group being opened later). The linked issue is unlikely to be caused by `m_advancer_sg` being null as `get_transact_stage()` would try to read from page 0 if called on a null pointer and thus produce EXC_BAD_ACCESS.

jpsim:
You're right, now this is a matter of tracing down why the state could ever be different than `Reading`...

PR #3528, opened 2016-05-04
Fixed reading the schema version from read-only realms

conradev:
I added a test case demonstrating the issue, and a proposed fix for the issue.

Let me know if I should apply a different fix or make any changes

### DIFF
--- a/Realm/Tests/MigrationTests.mm
+++ b/Realm/Tests/MigrationTests.mm
@@ -243,6 +243,35 @@ RLM_ARRAY_TYPE(MigrationObject);
     XCTAssertEqual(1U, [RLMRealm schemaVersionAtURL:config.fileURL encryptionKey:nil error:nil]);
 }
 
+- (void)testGetSchemaVersionOnReadOnlyRealm {
+    NSFileManager *fileManager = [NSFileManager defaultManager];
+    NSURL *directoryURL = [RLMDefaultRealmURL() URLByDeletingLastPathComponent];
+    
+    NSError *error = nil;
+    NSURL *readOnlyDirectoryURL = [directoryURL URLByAppendingPathComponent:@"ReadOnly"];
+    XCTAssertTrue([fileManager createDirectoryAtURL:readOnlyDirectoryURL withIntermediateDirectories:YES attributes:nil error:&error]);
+    XCTAssertNil(error);
+    
+    RLMRealmConfiguration *config = [RLMRealmConfiguration defaultConfiguration];
+    config.fileURL = [readOnlyDirectoryURL URLByAppendingPathComponent:@"default.realm"];
+    @autoreleasepool { [RLMRealm realmWithURL:config.fileURL]; }
+    XCTAssertEqual(0U, [RLMRealm schemaVersionAtURL:config.fileURL encryptionKey:nil error:nil]);
+    
+    error = nil;
+    XCTAssertTrue([fileManager setAttributes:@{NSFilePosixPermissions: @0x140} ofItemAtPath:config.fileURL.path error:&error]);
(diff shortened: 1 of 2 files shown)


Realm/Tests/MigrationTests.mm
```diff
@@ -243,6 +243,34 @@ - (void)testGetSchemaVersion {
...
+    XCTAssertNil(error);
+    
+    RLMRealmConfiguration *config = [RLMRealmConfiguration defaultConfiguration];
+    config.fileURL = [readOnlyDirectoryURL URLByAppendingPathComponent:@"default.realm"];
+    @autoreleasepool { [RLMRealm realmWithURL:config.fileURL]; }
+    XCTAssertEqual(0U, [RLMRealm schemaVersionAtURL:config.fileURL encryptionKey:nil error:nil]);
+    
+    error = nil;
+    XCTAssertTrue([fileManager setAttributes:@{NSFilePosixPermissions: @0x100} ofItemAtPath:config.fileURL.path error:&error]);
+    XCTAssertTrue([fileManager setAttributes:@{NSFilePosixPermissions: @0x100} ofItemAtPath:readOnlyDirectoryURL.path error:&error]);
```
bdash: This leaves a directory named ReadOnly that takes special work to remove (i.e., `rm -rf` chokes on it). It'd be worth having the test case clean up after itself as best it can.

conradev: Is the test case succeeding? I attempt to do this in line 270, so I'm confused as to why the directory would still be there.

conradev: Apparently not, from the CI:

```
"Error Domain=NSCocoaErrorDomain Code=513 "“ReadOnly” couldn’t be removed because you don’t have permission to access it." UserInfo=
```

bdash: Ah, I missed those lines when skimming the test. The attempt to remove the files appears to be failing due to the restrictive permissions the tests set on the directories. See https://travis-ci.org/realm/realm-cocoa/jobs/127903409, for instance, which says:

> /Users/travis/build/realm/realm-cocoa/Realm/Tests/MigrationTests.mm:271: error: -[MigrationTests testGetSchemaVersionOnReadOnlyRealm] : ((error) == nil) failed: "Error Domain=NSCocoaErrorDomain Code=513 "“ReadOnly” couldn’t be removed because you don’t have permission to access it." UserInfo={NSFilePath=/Users/travis/Library/Application Support/io.realm.TestHost/ReadOnly, NSUserStringVariant=(

conradev: Pushed a fix – it should now properly clean up after itself.

tgoyne:
Unconditionally opening it in readOnly mode could result in incorrect results and/or crashes if the file is being modified by another process at the same time (or another thread, but that at least is solvable and would require doing very weird things anyway), so I think it needs to be a parameter. Maybe it should just take an RLMRealmConfiguration? Presumably if you're calling this method you're going to be creating a config object anyway, but half of the config properties aren't applicable.

conradev:
That makes sense. I was worried about the consequences of opening a realm in readonly mode without the user specifying it. I'm not sure which would option would be best, but I like the idea of the schema version check taking an `RLMConfiguration`.

bdash:
@conradev, is there an issue that this fixes? If we're going to take an entirely different approach to fixing this it'd be good to close out this pull request in favor of an issue tracking the problem we're trying to fix.

conradev:
We ship a Demo realm inside of the Workflow bundle (and because it is in the bundle it can only be opened as read-only). I want to check the schema version of this realm before deciding to open it, in case it is too new.

bdash:
Sorry, I meant a GitHub issue filed against the realm-cocoa repository.

conradev:
Ah, yes, I see, there isn't to my knowledge

jpsim:
Closing due to inactivity.

PR #3530, opened 2016-05-04
Remove List::get_version_counter()

tgoyne:
This was a temporary hack for list notifications which was made unnecessary by find-grained notifications.


jpsim:
👍

PR #3548, opened 2016-05-10
Fix RLMLinkingObjects properties on RLMObject subclasses declared in Swift

### LINKED ISSUE #3546: Realm crashes when using linkingObjectsOfClass
## Goals

Use ```linkingObjectsProperties``` approach  since the ```linkingObjectsOfClass``` method been deprecated

## What do you want to achieve?

ability to link object together 

## What did happened instead?  

```
2016-05-10 16:12:36.446 FuseiOS-Core[43078:2278675] *** Terminating app due to uncaught exception 'RLMException', reason: 'Property 'conversations' is of type RLMLinkingObjects but +linkingObjectsProperties did not specify the class or property that is the origin of the link.'
*** First throw call stack:
```

## Code Sample

```
public class FSCMessageModel: RLMObject {
    
    /// Message Id
    public dynamic var id = NSNumber(integer: 0)
    
    public dynamic var conversations: RLMLinkingObjects?

    public override class func linkingObjectsProperties() -> [String : RLMPropertyDescriptor] {    
        return ["conversations" : RLMPropertyDescriptor(withClass: FSCConversationModel.self, propertyName: "messages")]
     }
}
```

```
public class FSCConversationModel: RLMObject {
    
    /// conversation Id
    public dynamic var id = NSNumber(integer: 0)
    
    /// conversation messages
    public dynamic var messages = RLMArray(objectClassName: String(FSCMessageModel))
}
```

Realm version: v0.102.0 (Objective-C version) project is mix and match.

Xcode version: 7.3

iOS/OSX version: 9.3

## Code where it fails:
RLMProperty.mm: line 241
```
                if (!_objectClassName || !_linkOriginPropertyName) {
                    @throw RLMException(@"Property '%@' is of type RLMLinkingObjects but +linkingObjectsProperties did not specify the class "
                                        "or property that is the origin of the link.", _name);
                }
```

seems like ```RLMObjectSchema``` create ```RLMProperty``` object without objectClassName and linkOriginPropertyName

bdash:
`RLMProperty` is initialized in a slightly different codepath for classes declared in Swift than Objective-C. The Swift codepath was missing the necessary logic to handle the linking objects property descriptor.

Fixes #3546.

/cc @jpsim @tgoyne

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,8 @@ x.x.x Release notes (yyyy-MM-dd)
 
 ### Bugfixes
 
-* None.
+* `RLMLinkingObjects` properties declared in Swift subclasses of `RLMObject`
+  now work correctly.
 
 0.102.0 Release notes (2016-05-09)
 =============================================================
--- a/Realm/RLMObjectSchema.mm
+++ b/Realm/RLMObjectSchema.mm
@@ -196,6 +196,7 @@ using namespace realm;
         if (isSwiftClass) {
             prop = [[RLMProperty alloc] initSwiftPropertyWithName:propertyName
                                                           indexed:[indexed containsObject:propertyName]
+                                           linkPropertyDescriptor:linkingObjectsProperties[propertyName]
                                                          property:props[i]
                                                          instance:swiftObjectInstance];
         }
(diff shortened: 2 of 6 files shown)


bdash:
CI says this regresses one of the Swift performance tests. Will investigate.

tgoyne:
The perf regression it's complaining about appears to have already landed on master. Running the perf tests locally that test takes .187 seconds on 0.100.0 and .300 seconds on master.

I suspect one or more of the CI devices are missing perf test baselines.

tgoyne:
The failing perf test is an expected regression due to the robust mutex changes (opening a SG is now slower due to that it needs to open more files), so it isn't anything related to this PR.

👍

PR #3550, opened 2016-05-10
Lazily initialize Swift LinkingObjects instances

bdash:
`LinkingObjects` instances were being initialized eagerly by `RLMInitializeSwiftAccessorGenerics`, which was resulting in a backlinks `TableView` being created for every Swift instance of a type that contained a `LinkingObjects` property. Lazily initializing these properties means that the work of building the backlinks view is deferred until the property is first accessed. This significantly reduces the cost of allocating a Swift object that has a `LinkingObjects` property.

/cc @tgoyne @jpsim @austinzheng 

This was visible in an Instruments trace a user provided via email.

### DIFF
--- a/Realm.xcodeproj/project.pbxproj
+++ b/Realm.xcodeproj/project.pbxproj
@@ -564,7 +564,7 @@
 		3FE556431B9A43E5002A1129 /* schema.hpp */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.cpp.h; name = schema.hpp; path = ObjectStore/schema.hpp; sourceTree = "<group>"; };
 		3FE79FF719BA6A5900780C9A /* RLMSwiftSupport.h */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.c.h; path = RLMSwiftSupport.h; sourceTree = "<group>"; };
 		3FEC4A3D1BBB188B00F009C3 /* SwiftSchemaTests.swift */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.swift; path = SwiftSchemaTests.swift; sourceTree = "<group>"; };
-		5D1534B71CCFF545008976D7 /* LinkingObjects.swift */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.swift; name = LinkingObjects.swift; path = "RealmSwift-swift2.0/LinkingObjects.swift"; sourceTree = "<group>"; };
+		5D1534B71CCFF545008976D7 /* LinkingObjects.swift */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.swift; name = LinkingObjects.swift; path = "RealmSwift-swift2.0/LinkingObjects.swift"; sourceTree = SOURCE_ROOT; };
 		5D2E8F651C98DC0D00187B09 /* RLMProperty_Private.hpp */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.cpp.h; path = RLMProperty_Private.hpp; sourceTree = "<group>"; };
 		5D30B74A1B42039F00D90C5A /* RLMDefines.h */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.c.h; path = RLMDefines.h; sourceTree = "<group>"; };
 		5D3E1A2C1C1FC6D5002913BA /* RLMPredicateUtil.hpp */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.cpp.h; path = RLMPredicateUtil.hpp; sourceTree = "<group>"; };
@@ -851,6 +851,7 @@
 			children = (
 				5D660FE31BE98D670021E04F /* Aliases.swift */,
 				29B7FDF51C0DA6560023224E /* Error.swift */,
+				5D1534B71CCFF545008976D7 /* LinkingObjects.swift */,
 				5D660FE41BE98D670021E04F /* List.swift */,
 				5D660FE51BE98D670021E04F /* Migration.swift */,
 				5D660FE61BE98D670021E04F /* Object.swift */,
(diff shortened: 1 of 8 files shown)


RealmSwift-swift2.0/LinkingObjects.swift
```diff
@@ -22,12 +22,33 @@ import Realm
...
+
+    internal func attachTo(object object: RLMObjectBase, property: RLMProperty) {
+        self.object = RLMWeakObjectHandle(object: object)
+        self.property = property
+        self.cachedRLMResults = nil
+    }
+
+    internal var rlmResults: RLMResults {
+        if cachedRLMResults == nil {
+            if let object = self.object, let property = self.property {
```
bdash: SwiftLint complains about this construct:
> warning: Conditional Binding Cascade Violation: Repeated `let` statements in conditional binding cascade should be avoided.

Any suggestions as to how cleanly avoid multiple `let` statements here?

austinzheng: You should be able to just do:

```
if let a = optionalA, b = optionalB, ... {

}
```

tgoyne:
Why `RLMWeakObjectHandle` over just a weak pointer to the object?

bdash:
There's no guarantee that anything will keep the object alive long enough. Consider:

```swift
let linkingObjects = realm.objects(SomeClass).first.linkingObjects
print(linkingObjects.count)
```

austinzheng:
Should that class be called `RLMLazyObjectHandle` instead? The point is that the object is instantiated on demand, right?

tgoyne:
That makes sense. I'm mostly just concerned that the extra row accessor could be expensive due to that it needs to grab a lock, but I suppose the TableView was doing that anyway.

bdash:
Now that I think about it, `RLMWeakObjectHandle` could also hold a weak reference to the original object to sometimes allow skipping the allocation of a new instance within `-object`. It wouldn't avoid the need for it to hold a row accessor itself though, so it may not be worth it.

@austinzheng, I used "weak" in the name since the important part of its behavior is that it doesn't keep the object accessor alive. "lazy" doesn't feel like a good fit for the semantics it provides.

bdash:
The CI failure is only the pre-existing `testRealmCreationCached` regression in Swift device tests. Moving to Review.

Realm.xcodeproj/project.pbxproj
```diff
@@ -564,7 +564,7 @@
 		3FE556431B9A43E5002A1129 /* schema.hpp */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.cpp.h; name = schema.hpp; path = ObjectStore/schema.hpp; sourceTree = "<group>"; };
 		3FE79FF719BA6A5900780C9A /* RLMSwiftSupport.h */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.c.h; path = RLMSwiftSupport.h; sourceTree = "<group>"; };
 		3FEC4A3D1BBB188B00F009C3 /* SwiftSchemaTests.swift */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.swift; path = SwiftSchemaTests.swift; sourceTree = "<group>"; };
-		5D1534B71CCFF545008976D7 /* LinkingObjects.swift */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.swift; name = LinkingObjects.swift; path = "RealmSwift-swift2.0/LinkingObjects.swift"; sourceTree = "<group>"; };
+		5D1534B71CCFF545008976D7 /* LinkingObjects.swift */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.swift; name = LinkingObjects.swift; path = "RealmSwift-swift2.0/LinkingObjects.swift"; sourceTree = SOURCE_ROOT; };
```
mrackwitz: This hard-codes the Swift version by setting the file-reference relative to the source root. It should be instead relative to the group: `path = "LinkingObjects.swift"; sourceTree = "<group>";`

Note: This can't be setup from Xcode's UI and requires manual project file modification.

bdash: Weird. Not sure why Xcode changed this. All I did was drag the file to the correct place in the UI. Will fix.

Realm/RLMObject.mm
```diff
@@ -219,3 +219,31 @@ - (void)setValue:(id)value forUndefinedKey:(NSString *)key {
...
+
+- (instancetype)initWithObject:(RLMObjectBase *)object {
+    if (!(self = [super init])) {
+        return nil;
+    }
+
+    _row = object->_row;
+    _realm = object->_realm;
+    _objectSchema = object->_objectSchema;
+    _objectClass = object.class;
```
tgoyne: This doesn't actually need the object class; it can just create an RLMObjectBase. It also doesn't need the object schema, but that's a bit iffier since that relies on RLMDynamicGet() not needing it.

Realm/RLMObject.mm
```diff
@@ -219,3 +219,31 @@ - (void)setValue:(id)value forUndefinedKey:(NSString *)key {
...
+
+    return self;
+}
+
+- (RLMObjectBase *)object {
+    RLMObjectBase *object = [[_objectClass alloc] initWithRealm:_realm schema:_objectSchema];
+    object->_row = _row;
```
tgoyne: Moving from _row would save a lock acquisition when the RLMWeakObjectHandle is deallocated.

bdash: That's true. The class or method name will need to be updated to reflect the fact that it can only be converted into an object once.

Realm/RLMObject.mm
bdash: For `LinkingObjects` purposes you're right that `RLMObjectBase` is sufficient. I went with returning the object's class as it seemed more correct, and avoids any potential problems if in the future the objects returned by this class make their way back to user code. It doesn't seem like it's expensive enough relative to the other work that happens initializing Swift generics to worry about for performance reasons.

tgoyne:
👍 

CI failure unrelated.

PR #3552, opened 2016-05-10
Speed up creating RealmSwift.Realm instances

tgoyne:
Nothing particularly exciting, but while checking why the relevant perf test was failing I noticed it was spending a lot of time on unnecessary things. This makes both `Realm()` and `self.testRealmPath()` (in our tests) take about a third less time, and switches the testRealmCreationCached test over to testing what is hopefully a more realistic scenario.


bdash:
👍

jpsim:
👍

Issue #3542, opened 2016-05-10
Feature Request: Create observer/callback/block when deleteRealmIfMigrationNeeded is set to true


daffodilistic:
## Goals

It would be nice if Realm has a "hook" via KVO, notification, callback or block which is called when a deletion is needed BEFORE the deletion occurs, when `deleteRealmIfMigrationNeeded` is set to true for a `Realm.Configuration`

For example:

```
let config = Realm.Configuration(
                schemaVersion: 1,
                deleteOnMigrationBlock: { 
                    let defaults: NSUserDefaults = NSUserDefaults.standardUserDefaults()
                    defaults.removeObjectForKey("OfflineFriendListUpdateTime")
                    defaults.synchronize()
                }
            })

config.deleteRealmIfMigrationNeeded = true
Realm.Configuration.defaultConfiguration = config
```

## Expected Results

To be able to execute code before deletion of the Realm file occurs, like unsetting user preferences, showing a loading bar, etc.

mrackwitz:
You can workaround that once #3559 is released by implementing this strategy on your own rather than using `deleteRealmIfMigrationNeeded`. You can achieve that by catching the error thrown on a schema mismatch.

```swift
let config = Realm.Configuration.defaultConfiguration
let realm: Realm
do {
  realm = try Realm(configuration: config)
  // …
} catch Error.SchemaMismatch {
  // Do some work before the deletion occurs
  let defaults = NSUserDefaults.standardUserDefaults()
  defaults.removeObjectForKey("OfflineFriendListUpdateTime")
  defaults.synchronize()

  // Delete the realm file
  try? NSFileManager.defaultManager().removeItemAtURL(config.fileURL)
  
  // Try again
  realm = try Realm(configuration: config)
}
```

kubajakowski:
@mrackwitz I have similar code in my DBService but Realm doesn't create new file after deletion...

```
    private var configuration: Realm.Configuration
    {
        var configuration = Realm.Configuration.defaultConfiguration
        configuration.deleteRealmIfMigrationNeeded = true
        return configuration
    }
    
    lazy var realm: Realm? = {
        do {
            //trying to create realm
            let realm = try Realm(configuration: self.configuration)
            return realm
        }catch (let e) {
            debugPrint(e)
            //here got Error Domain=io.realm Code=5 Directory at path '.../deafult.realm' does not exist
            //so all good, file shoud be deleted when migration is needed
            //trying to recreate
[...]

PR #3559, opened 2016-05-11
Explicit error code for schema mismatch

### LINKED ISSUE #3534: Readonly database migration error/warning
I was trying to track down an error as to why my database needed migration but it wasn't running the migration block and I noticed its because I was opening it read-only.  

It would be great if you could throw an error message that said something like:  `Database Needs Migration but is opened in read-only mode`

Thanks

mrackwitz:
This makes it easier to catch when a migration is required and provide a custom error handling, instead of relying on the built-in delete strategy.

This allows to workaround #3542. This fixes #3534.

/c @jpsim @austinzheng @kishikawakatsumi

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,8 @@ x.x.x Release notes (yyyy-MM-dd)
 
 ### Enhancements
 
-* None.
+* Return `RLMErrorSchemaMismatch` error rather than the more generic `RLMErrorFail`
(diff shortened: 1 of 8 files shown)


jpsim:
👍 nice, changelog entry?

tgoyne:
👍

kishikawakatsumi:
👍 Thanks!

Issue #3438, opened 2016-04-14
Some Realm errors don't include sufficient detail to understand their cause


bdash:
In #3426 a user is seeing an error message of `Unable to open a realm at path '…/foo.realm'`. This error message appears to be generated in the following cases, and possibly others:

1. Opening a file that does not contain a valid Realm file header.
2. Attempting to open an encrypted Realm file with an incorrect encryption key.
3. Attempting to open a directory as a Realm file, or other POSIX-level errors when opening the Realm file.

In the first case we should make it clear that what is being opened is not a valid Realm file.

In the second case we should make it clear that either the encryption key is incorrect or the file is not a valid Realm file (the two cases are indistinguishable).

For other errors that occur when opening the Realm file we should preserve as much information as is possible (e.g., the underlying `std::system_error`, or `errno` value) and use it to populate `NSUnderlyingErrorKey` in the `NSError` we return.

fisch0920:
Currently running into this and would really love this improvement.

Thanks for all your hard work!!

austinzheng:
I pushed a change which should make future errors more descriptive. Please feel free to file another ticket if the next release of Realm Cocoa still has insufficiently descriptive errors.

fisch0920:
Thanks @austinzheng!

Issue #3069, opened 2016-01-05
Errors returned by Realm should use different domains for error codes from different sources


bdash:
The variants of `RLMMakeError` always use `RLMErrorDomain` for the domain of the error, but the error code itself can come from different sources (the value of the `RLMError` argument passed to the function, or `std::error_code::value`). The different sources use the same codes to represent different errors (e.g., 2 could mean either `RLMErrorFileAccess` or `ENOENT`, 4 could be either `RLMErrorFileExists` or `EINTR`).

I think we should be using `NSPOSIXErrorDomain` for `std::system_error`s with a category of `std::generic_category`. I'm not sure what the right approach is for errors from other categories. We could create an `RLMUnknownSystemErrorDomain`, or we could map them all to an `RLMUnknownError` code (is this what `RLMErrorFail` is for?) and include the underlying exception's code in the user info dictionary. Either approach would also benefit from having the category name in the user info dictionary.

austinzheng:
This is in. Unfortunately, it doesn't appear that Realm Cocoa actually throws non-Realm exceptions that can be propagated into `NSError` objects. If this changes in the future additional unit tests should be written.

PR #3560, opened 2016-05-11
Adding better error domain support for system errors

austinzheng:
A first shot at addressing #3069 and #3438. `NSError` objects created from system exceptions should be slightly more descriptive now. @bdash @tgoyne

Changes:
- Realm NSError now populated with the NSPOSIXErrorDomain when appropriate
- NSError user info now contains system error category
- Added RLMUnknownSystemErrorDomain for non-POSIX errors

### DIFF
--- a/Realm/ObjectStore/shared_realm.hpp
+++ b/Realm/ObjectStore/shared_realm.hpp
@@ -199,14 +199,16 @@ namespace realm {
             /** Thrown if the file needs to be upgraded to a new format, but upgrades have been explicitly disabled. */
             FormatUpgradeRequired,
         };
-        RealmFileException(Kind kind, std::string path, std::string message) :
-            std::runtime_error(std::move(message)), m_kind(kind), m_path(std::move(path)) {}
+        RealmFileException(Kind kind, std::string path, std::string message, std::string underlying) :
+            std::runtime_error(std::move(message)), m_kind(kind), m_path(std::move(path)), m_underlying(std::move(underlying)) {}
         Kind kind() const { return m_kind; }
         const std::string& path() const { return m_path; }
+        const std::string& underlying() const { return m_underlying; }
 
     private:
         Kind m_kind;
         std::string m_path;
+        std::string m_underlying;
     };
 
     class MismatchedConfigException : public std::runtime_error {
--- a/Realm/RLMUtil.mm
+++ b/Realm/RLMUtil.mm
@@ -244,10 +244,15 @@ NSError *RLMMakeError(RLMError code, const realm::RealmFileException& exception)
 }
 
 NSError *RLMMakeError(std::system_error const& exception) {
-    return [NSError errorWithDomain:RLMErrorDomain
+    BOOL isGenericCategoryError = (exception.code().category() == std::generic_category());
+    NSString *category = [NSString stringWithUTF8String:exception.code().category().name()];
+    NSString *errorDomain = isGenericCategoryError ? NSPOSIXErrorDomain : RLMUnknownSystemErrorDomain;
+
+    return [NSError errorWithDomain:errorDomain
                                code:exception.code().value()
                            userInfo:@{NSLocalizedDescriptionKey: @(exception.what()),
-                                      @"Error Code": @(exception.code().value())}];
+                                      @"Error Code": @(exception.code().value()),
+                                      @"Category": category}];
 }
 
 NSError *RLMMakeError(NSException *exception) {
--- a/Realm/RLMUtil.mm
+++ b/Realm/RLMUtil.mm
@@ -236,16 +236,18 @@ NSError *RLMMakeError(RLMError code, const realm::util::File::AccessError& excep
 }
 
 NSError *RLMMakeError(RLMError code, const realm::RealmFileException& exception) {
+    NSString *underlying = @(exception.underlying().c_str());
     return [NSError errorWithDomain:RLMErrorDomain
                                code:code
                            userInfo:@{NSLocalizedDescriptionKey: @(exception.what()),
                                       NSFilePathErrorKey: @(exception.path().c_str()),
-                                      @"Error Code": @(code)}];
+                                      @"Error Code": @(code),
+                                      @"Underlying": underlying.length == 0 ? @"n/a" : underlying}];
 }
 
 NSError *RLMMakeError(std::system_error const& exception) {
     BOOL isGenericCategoryError = (exception.code().category() == std::generic_category());
-    NSString *category = [NSString stringWithUTF8String:exception.code().category().name()];
+    NSString *category = @(exception.code().category().name());
     NSString *errorDomain = isGenericCategoryError ? NSPOSIXErrorDomain : RLMUnknownSystemErrorDomain;
 
     return [NSError errorWithDomain:errorDomain
--- a/Realm/Tests/AsyncTests.mm
+++ b/Realm/Tests/AsyncTests.mm
@@ -414,11 +414,7 @@
     XCTestExpectation *exp = [self expectationWithDescription:@""];
     auto token = [IntObject.allObjects addNotificationBlock:^(RLMResults *results, RLMCollectionChange *change, NSError *error) {
         XCTAssertNil(results);
-        XCTAssertNotNil(error);
-        // Validate error
-        XCTAssertEqualObjects(error.domain, RLMErrorDomain);
-        XCTAssertEqual(error.code, RLMErrorFail);
-        XCTAssert([error.localizedDescription rangeOfString:@"Too many open files"].location != NSNotFound);
+        RLMValidateRealmError(error, RLMErrorFail, @"Too many open files", nil);
         called = true;
         [exp fulfill];
     }];
@@ -437,10 +433,7 @@
     XCTestExpectation *exp2 = [self expectationWithDescription:@""];
     auto token2 = [IntObject.allObjects addNotificationBlock:^(RLMResults *results, RLMCollectionChange *change, NSError *error) {
         XCTAssertNil(results);
-        XCTAssertNotNil(error);
-        // Validate error
-        XCTAssertEqualObjects(error.domain, RLMErrorDomain);
-        XCTAssertEqual(error.code, RLMErrorFail);
+        RLMValidateRealmError(error, RLMErrorFail, @"Too many open files", nil);
         [exp2 fulfill];
     }];
     [realm beginWriteTransaction];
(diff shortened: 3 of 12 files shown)


austinzheng:
I added unit tests. However, per conversation with @bdash, there's more work to be done so this PR probably isn't ready for review yet.

Realm/ObjectStore/shared_realm.hpp
```diff
@@ -199,14 +199,16 @@ namespace realm {
             /** Thrown if the file needs to be upgraded to a new format, but upgrades have been explicitly disabled. */
             FormatUpgradeRequired,
         };
-        RealmFileException(Kind kind, std::string path, std::string message) :
-            std::runtime_error(std::move(message)), m_kind(kind), m_path(std::move(path)) {}
+        RealmFileException(Kind kind, std::string path, std::string message, std::string underlying) :
+        std::runtime_error(std::move(message)), m_kind(kind), m_path(std::move(path)), m_underlying(std::move(underlying)) {}
```
jpsim: the indentation here was intentional

jpsim:
Looks good to me! 👍

Realm/ObjectStore/shared_realm.hpp
austinzheng: Thanks. Looks like Xcode got confused and removed the indent.

austinzheng:
@jpsim I fixed the indentation and added another unit test.

bdash:
There don't appear to be any functional tests that look for `RLMUnknownSystemErrorDomain`. Is that because we don't expect those errors to ever be exposed directly, and the domain is just to ensure that they are kept distinct from our own error codes?

bdash:
There also don't appear to be any tests that look for `NSPOSIXErrorDomain`.

Realm/Tests/RealmTests.mm
```diff
@@ -1423,21 +1423,111 @@ - (void)testIsEmpty {
...
     configuration.fileURL = fileURL;
     XCTAssertNil([RLMRealm realmWithConfiguration:configuration error:&error], @"Invalid database");
     XCTAssertNotNil(error, @"Should populate error object");
+    // Validate error contents
+    XCTAssertEqualObjects(error.domain, RLMErrorDomain);
+    XCTAssertEqual(error.code, RLMErrorFileAccess);
+    XCTAssert([error.userInfo[@"Underlying"] rangeOfString:@"Realm file has bad size"].location != NSNotFound);
+}
+
+- (void)testRealmFileAccessBadEncryptionKey
```
tgoyne: There's a family of tests in EncryptionTests which cover this (testReopenWithNoKeyThrows, testReopenWithWrongKeyThrows, testOpenUnencryptedWithKeyThrows), so rather than duplicating the testing it'd be better to just make those tests actually check what exception is thrown.

Realm/Tests/RealmTests.mm
```diff
@@ -1423,21 +1423,111 @@ - (void)testIsEmpty {
...
+    XCTAssertNotNil(error);
+    // Validate error contents
+    XCTAssertEqualObjects(error.domain, RLMErrorDomain);
+    XCTAssertEqual(error.code, RLMErrorFileAccess);
+    XCTAssert([error.userInfo[@"Underlying"] rangeOfString:@"Realm file decryption failed"].location != NSNotFound);
+}
+
+- (void)testRealmFileAccessFileIsDirectory
+{
+    NSURL *fileURL = [NSURL fileURLWithPath:RLMRealmPathForFile(@"filename.realm")];
```
tgoyne: This test could be simplified by creating the directory at `RLMTestRealmURL()`, as RLMTestCase takes care of deleting anything that already exists at that path and also cleans it up after the test.

Realm/Tests/UtilTests.mm
```diff
@@ -56,6 +56,52 @@ - (void)testRLMExceptionWithCPlusPlusException {
...
+                                       NSLocalizedDescriptionKey : description,
+                                       @"Error Code" : @(code),
+                                       @"Category": [NSString stringWithUTF8String:std::system_category().name()]
+                                       };
+    XCTAssertEqualObjects(RLMMakeError(exception),
+                          [NSError errorWithDomain:RLMUnknownSystemErrorDomain code:code userInfo:expectedUserInfo]);
+}
+
+using namespace realm;
+#import "shared_realm.hpp"
```
tgoyne: Move to the top of the file after the other imports, and only use `using namespace` after the last header inclusion, as otherwise it'll apply to those headers and possibly break things.

austinzheng:
@bdash The error domain system in this PR segments errors into Realm-specific errors, POSIX system errors, and all other errors. It's not apparent to me that errors in the third category are ever actually thrown.

The `RLMRealmTranslateException` function itself is only invoked in two places: when opening a shared Realm and when adding a notification block. In the former case the only errors that can be emitted are Realm-specific ones. I'm not sure yet about the latter case.

Getting Realm itself to throw a POSIX error is kind of difficult. They only seem to be thrown in `external_commit_helper.cpp`. I'm not yet sure how to induce a situation where `ExternalCommitHelper` would fail and throw an exception. I'll continue looking into it.

Realm/Tests/AsyncTests.mm
```diff
@@ -415,6 +415,8 @@ - (void)testErrorHandling {
     auto token = [IntObject.allObjects addNotificationBlock:^(RLMResults *results, RLMCollectionChange *change, NSError *error) {
         XCTAssertNil(results);
         XCTAssertNotNil(error);
+        // Validate error
+        XCTAssertEqualObjects(error.domain, RLMErrorDomain);
```
tgoyne: In general any place where we're checking that something is equal to an exported constant in a test we want to use `XCTAssertEqual()` rather than `XCTAssertEqualObjects()` so that hardcoding the value somewhere rather than using the constant will produce a test failure (at least in debug builds; in release builds string interning will foil that).

Realm/Tests/RealmTests.mm
```diff
@@ -1423,21 +1423,111 @@ - (void)testIsEmpty {
...
+    assert(![[NSFileManager defaultManager] fileExistsAtPath:fileURL.path]);
     [[NSFileManager defaultManager] createFileAtPath:fileURL.path contents:fileContents attributes:nil];
 
     NSError *error;
     RLMRealmConfiguration *configuration = [RLMRealmConfiguration defaultConfiguration];
     configuration.fileURL = fileURL;
     XCTAssertNil([RLMRealm realmWithConfiguration:configuration error:&error], @"Invalid database");
     XCTAssertNotNil(error, @"Should populate error object");
+    // Validate error contents
+    XCTAssertEqualObjects(error.domain, RLMErrorDomain);
```
tgoyne: Checking the domain, code and message of an exception occurs in enough places that it seems worth extracting that to a single macro (and that will hopefully result in it being checked in *more* places, too...).

Realm/RLMUtil.mm
```diff
@@ -236,18 +236,25 @@ BOOL RLMIsObjectValidForProperty(__unsafe_unretained id const obj,
...
-                                      @"Error Code": @(code)}];
+                                      @"Error Code": @(code),
+                                      @"Underlying": underlying.length == 0 ? @"n/a" : underlying}];
 }
 
 NSError *RLMMakeError(std::system_error const& exception) {
-    return [NSError errorWithDomain:RLMErrorDomain
+    BOOL isGenericCategoryError = (exception.code().category() == std::generic_category());
+    NSString *category = @(exception.code().category().name());
+    NSString *errorDomain = isGenericCategoryError ? NSPOSIXErrorDomain : RLMUnknownSystemErrorDomain;
```
austinzheng: I actually think `system_domain` corresponds to POSIX errors, not `generic_domain`. I  might be wrong though. @bdash?

bdash: [`std::generic_category()`](http://en.cppreference.com/w/cpp/error/generic_category) corresponds to POSX errno codes:

> Obtains a reference to the static error category object for generic errors. The object is required to override the virtual function error_category::name() to return a pointer to the string "generic". It is used to identify error conditions that correspond to the POSIX errno codes.

austinzheng: Got it. Thanks.

austinzheng:
I updated the review based on Thomas' feedback but still need to write POSIX and unknown domain error code tests.

Realm/Tests/AsyncTests.mm
```diff
@@ -414,11 +414,7 @@ - (void)testErrorHandling {
     XCTestExpectation *exp = [self expectationWithDescription:@""];
     auto token = [IntObject.allObjects addNotificationBlock:^(RLMResults *results, RLMCollectionChange *change, NSError *error) {
         XCTAssertNil(results);
-        XCTAssertNotNil(error);
-        // Validate error
-        XCTAssertEqualObjects(error.domain, RLMErrorDomain);
-        XCTAssertEqual(error.code, RLMErrorFail);
-        XCTAssert([error.localizedDescription rangeOfString:@"Too many open files"].location != NSNotFound);
+        RLMValidateRealmError(error, RLMErrorFail, @"Too nany open files");
```
bdash: s/nany/many/

Hopefully this test is failing!

austinzheng: This test was not failing unfortunately, but I fixed the macro and the typo and made sure it worked with both good and bad test cases.

austinzheng:
I still haven't found a situation where a non-Realm exception is directly propagated to the user, but I did manage to write another test case.

Realm/Tests/RLMAssertions.h
```diff
@@ -54,10 +54,39 @@
...
 
 #define RLMAssertThrowsWithCodeMatching(expression, expectedCode, ...) \
 ({ \
     NSException *exception = RLMAssertThrows(expression, __VA_ARGS__); \
     XCTAssertEqual([exception.userInfo[NSUnderlyingErrorKey] code], expectedCode, __VA_ARGS__); \
 })
+
+#define RLMValidateRealmError(__error, __errnum, __description, __underlying)                        \
+({                                                                                                   \
+    NSString *__dsc = __description;                                                                 \
```
bdash: Identifiers that begin with two underscores are reserved.

Realm/Tests/RealmTests.mm
```diff
@@ -1448,23 +1448,76 @@ - (void)testIsEmpty {
     XCTAssertFalse(realm.isEmpty, @"Realm should not be empty after committing a write transaction that added an object.");
 }
 
+- (void)testRealmFileAccessNilPath {
+    XCTAssertThrows([RLMRealm realmWithURL:self.nonLiteralNil], @"nil path");
```
bdash: Should this be testing which error is thrown?

bdash:
👍

PR #3561, opened 2016-05-11
Update performance baselines

tgoyne:
Set `maxRegression` to 1ms for all tests, as it turns out that both the absolute and relative thresholds must be hit for a perf test to fail.

Update all baselines to the performance of current master. In most cases this is faster than the previous baseline for Swift tests (often significantly slow). A bunch of baselines were also previously missing entirely, and we had some baselines for a target which doesn't even exist in the project any more.

The following tests had their baselines made slower:

Realm.testDeleteAll, Realm.testQueryDeletion, RealmSwift.testQueryDeletion: These were made slower by adding more classes which link to the objects being deleted, and basically all of the time is spent updating backlinks.

Realm.testRealmCreationCached, RealmSwift.testRealmCreationCached: Made slower by the robust interprocess mutex change.

RealmSwift.testCountWhereTableView: Not sure why this is slower. Profiler says ~90% of the runtime is in the single call to `Query::find_all()`, so it's not due to binding changes.

Some of the KVO tests were modified because it turned out that they were actually mostly just benchmarking how long it takes to create an RLMArray, which got slower due to the change to make LinkList destruction thread-safe.

`testCrossThreadSyncLatency()` had regressed a bit due to core changes, but the changes in the first commit bring it back to its old speed.


RealmSwift-swift2.0/Realm.swift
```diff
@@ -493,7 +493,13 @@ public final class Realm {
     */
     @warn_unused_result(message="You must hold on to the NotificationToken returned from addNotificationBlock")
     public func addNotificationBlock(block: NotificationBlock) -> NotificationToken {
-        return rlmRealm.addNotificationBlock(rlmNotificationBlockFromNotificationBlock(block))
+        return rlmRealm.addNotificationBlock { rlmNotification, _ in
+            if rlmNotification == RLMRealmDidChangeNotification {
```
jpsim: could combine the if/else paths into just `block(notification: Notification(rawValue: rlmNotification)!, realm: self)`

Realm/RLMRealm.mm
```diff
@@ -487,10 +487,21 @@ - (void)removeNotification:(RLMNotificationToken *)token {
 - (void)sendNotifications:(NSString *)notification {
     NSAssert(!_realm->config().read_only, @"Read-only realms do not have notifications");
 
+    NSUInteger count = _notificationHandlers.count;
```
jpsim: I don't understand how these changes improve performance. Seems identical to me, unless the compiler is doing really dumb stuff.

1. the loop won't enter if `count == 0`
2. the loop will run the block exactly once if `count == 1`
3. the loop is identical for `count > 1`

What am I missing?

tgoyne: `_notificationHandlers.allObjects` is much slower than `anyObject`, so this avoids calling it unless it's actually needed.

RealmSwift-swift2.0/Realm.swift
tgoyne: Not using `Notification(rawValue: rlmNotification)` is the optimization here.

jpsim: wild

Realm/RLMRealm.mm
jpsim: alright then

jpsim:
So if the baselines were updated to the current state of master, why are the performance tests failing in CI?

RealmSwift-swift2.0/Realm.swift
tgoyne: The reason it's slower is that converts the NSString to a swift String (allocating several new objects for some reason and copying the string data) and then does a string comparison against the raw values, while this is just a message send to `isEqual:` which returns true from the pointer comparison fast-path.

tgoyne:
That's a good question!

tgoyne:
So it looks like the reason why a bunch of Swift tests got faster was due to OS-level changes, and not compiler improvements or changes in our code. The CI iPad is running 8.4, but I generated the baselines with one running 9.1, and the OS version isn't recorded in the baseline file.

tgoyne:
Now actually passes on CI. Turns out that iOS 8 has completely different performance from iOS 9.

On iOS 8 all of the tests that are mostly benchmarking accessor creation performance had a >10% regression, so I cut down on the work that RLMInitializeSwiftAccessorGenerics does when it's a Swift class but no properties actually need the extra init to get perf back to the old baseline.

A bunch of the swift performance tests had a relative standard deviation of > 25%, so I bumped up how much work they did.

bdash:
:+1:

jpsim:
👍

PR #3563, opened 2016-05-12
[Project] Move LinkingObjects.swift into RealmSwift group

mrackwitz:
I found the file reference in the root group. Furthermore it was based on a relative path which hard-coded the Swift version.


jpsim:
This is also done by #3550

mrackwitz:
Not exactly. It need to be fixed in #3550.

Issue #3470, opened 2016-04-22
Deprecate Xcode 6.4


jpsim:
Should be done relatively soon.

PR #3567, opened 2016-05-12
Remove support for Xcode 6.x & Swift prior to 2.2

### LINKED ISSUE #3469: Deprecate all Swift versions other than 2.2.x
We should soon deprecate Swift versions older than 2.2.x, which will allow us to avoid an `associatedtype` warning, make use of Swift language version compile-time checks and possibly simplify our build system depending on whether or not we feel comfortable removing the symlink infrastructure temporarily.

jpsim:
Closes #3469 & #3470 /cc @tgoyne @bdash @austinzheng 

It's debatable that we might want to keep some of the machinery for supporting multiple Xcode versions side by side, but until that proves to be necessary, I'd rather keep things as simple as possible.


tgoyne:
Swift 3 has a whole bunch of breaking changes, so there's zero chance that we won't need the version switching logic in just a month when the first Xcode DP with it comes out.

All of the preprocessor macros for supporting Xcode 6 need to be cleaned up.

jpsim:
> Swift 3 has a whole bunch of breaking changes, so there's zero chance that we won't need the version switching logic in just a month when the first Xcode DP with it comes out.

Yes, but Swift 2.2 introduced compile-time Swift version checking, so this should all be doable in the same source files.

bdash:
My suggestion would be to leave the version switching logic in place until we've had a chance to evaluate the scope of the changes that Swift 3 requires. 

Some of the Swift 3 changes look like they'll be quite invasive so it's not obvious to me that using Swift's version checking is the way to go. On top of that, we'll likely want to update our API to match the new Swift guidelines for things like argument labels. This will increase the divergence between the Swift 2.2 and 3 code paths, and could make documentation a challenge in a unified codebase.

tgoyne:
Yeah, my guess is that supporting both 2.2 and 3 would make the code preprocessor soup, and that can get quite a bit worse to deal with than the duplication was. If it turns out to not be so bad then we should definitely just use the version checks, though.

The release branch for the first Swift 3 DP is getting created today, so we are getting to the point where it'd be worth trying to port Realm to Swift 3, and thus finding out how much actually needs to change. Only real deadline for getting that ready is by WWDC, so polishy things can certainly wait a few weeks.

jpsim:
Alright, you've all convinced me, I've made #3575 which is a bit more conservative.

PR #3575, opened 2016-05-13
Remove support for Xcode 6.x & Swift prior to 2.2

### LINKED ISSUE #3469: Deprecate all Swift versions other than 2.2.x

jpsim:
Closes #3469 & #3470 /cc @tgoyne @bdash @austinzheng

This is a more conservative version of #3567 since it preserves much of our existing support for multiple Swift/Xcode versions.


jpsim:
Please review @tgoyne @bdash @austinzheng

build.sh
```diff
@@ -649,11 +644,26 @@ case "$COMMAND" in
 
     "verify-cocoapods")
         pod setup
```
tgoyne: Should do something like `git diff-index --quiet HEAD || ( echo "Cannot test the podspec with uncommitted changes"; exit 1)` since it does a hard reset.

build.sh
```diff
@@ -887,15 +891,14 @@ case "$COMMAND" in
     # CocoaPods
     ######################################
     "cocoapods-setup")
+        if [ ! -d core ]; then
+          # core directory won't exist if running `pod spec lint` locally
```
tgoyne: This comment doesn't really make any sense. This command is the thing that downloads core when installing the pod normally, so of course the core directory won't exist.

build.sh
```diff
@@ -887,15 +891,14 @@ case "$COMMAND" in
...
+          ./build.sh download-core
```
tgoyne: This script uses `sh build.sh` everywhere else.

austinzheng:
Your most recent source code changes look fine to me. I can't say much about the build scripts.

bdash:
Need to update the README too.

👍

PR #3576, opened 2016-05-13
Update to core 0.100.4

### LINKED ISSUE #3540: Realm 0.101: fails querying objects from read-only db in background threads
I ran into this pretty severe issue which appeared in Realm 0.101 (or maybe 0.100, didn't test with previous version): using a database in read-only mode only yields query results from the main thread. Querying the database from a background threads yields empty results sets.

## Expected Results

In Realm version prior to 0.100 this used to work just fine (I have a read-only DB in the application with static contents, alongside a read-write DB with dynamic contents).

## Actual Results

With Realm 0.101, querying the read-only DB in a background thread returns empty result sets

## Steps to Reproduce

Attached is a sample project that demonstrates the issue. It creates a simple DB, then closes it, then opens it read-only from two different queues (main and background) and only the main queue read returns query results for the objects written to the database.

With attached project:
```shell
$ pod install
$ open RealmStaticDBCrash.xcworkspace
```

Then build & run for simulator (iPhone 6S Plus, iOS 9.3). Test results are displayed in Xcode console.

## Code Sample

See attached project.
[RealmStaticDBCrash.zip](https://github.com/realm/realm-cocoa/files/255518/RealmStaticDBCrash.zip)

Realm version: 0.101.0

Xcode version: 7.3.1 OS X 10.11.4

iOS/OSX version: iOS simulator 9.3

### LINKED ISSUE #3547: Queries involving multiple levels of inverse relationships hit assertion failure in core
A user provided the following test case via email that reproducibly asserts:
```swift
class A: Object {
   dynamic var id = 0
    dynamic var name = ""
    let list = List<B>()
}

class B: Object {
    dynamic var name = ""
    let list = List<C>()
    let source = LinkingObjects(fromType: A.self, property: "list")
}

class C: Object {
    dynamic var name = ""
    let source = LinkingObjects(fromType: B.self, property: "list")
}

let realm = try! Realm()
return realm.objects(C).filter("ANY source.source.id == 1")
```

tgoyne:
Fixes #3540. Fixes #3547.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,10 @@ x.x.x Release notes (yyyy-MM-dd)
   objects, and then deleting some of the newly inserted objects within a single
   write transaction when there is an active notification block for a different
   object type which links to the objects being deleted.
+* Fix crashes and/or incorrect results when querying over multiple levels of
+  `LinkingObjects` properties.
+* Fix opening read-only Realms on multiple threads at once.
+* Fix a `BadTransactLog` exception when storing dates before the unix epoch (1970-01-01).
 
 0.102.0 Release notes (2016-05-09)
 =============================================================
--- a/Realm/Tests/QueryTests.m
+++ b/Realm/Tests/QueryTests.m
@@ -21,6 +21,37 @@
 
 #pragma mark - Test Objects
 
+@class LinkChain2, LinkChain3;
+
+@interface LinkChain1 : RLMObject
+@property int value;
+@property LinkChain2 *next;
+@end
+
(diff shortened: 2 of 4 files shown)


tgoyne:
CI failures are unrelated to the changes in this branch, and I'm poking at them in https://github.com/realm/realm-cocoa/pull/3578.

tgoyne:
@bdash @jpsim

bdash:
Thanks for taking care of writing the test cases for the query fix.

👍

jpsim:
👍

PR #3582, opened 2016-05-14
[Podspec] Fix import issues with CocoaPods 1.0 on OS X

mrackwitz:
Workaround for CocoaPods/CocoaPods#5313.

@TimOliver: This has potential to break MAS-compatiblity. Could you validate whether apps using frameworks like that pass AppStore's automated file validations?

/c @jpsim

### DIFF
--- a/Realm.podspec
+++ b/Realm.podspec
@@ -61,6 +61,11 @@ Pod::Spec.new do |s|
                                 'APPLICATION_EXTENSION_API_ONLY' => 'YES',
                                 'HEADER_SEARCH_PATHS' => '"${PODS_ROOT}/Realm/include/core"',
                                 'USER_HEADER_SEARCH_PATHS' => '"${PODS_ROOT}/Realm/include" "${PODS_ROOT}/Realm/include/Realm"' }
+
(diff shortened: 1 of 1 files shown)


jpsim:
@mrackwitz doesn't this break CocoaPods before 1.0? [See CI output](https://ci.realm.io/job/objc_pr/3315/configuration=Release,swift_version=2.2,target=cocoapods/console) using 0.39.0:

```
Setup completed

 -> Realm
 -> Realm (0.102.1)
    - ERROR | [OSX] xcodebuild: Returned an unsuccessful exit code. You can use `--verbose` for more information.
    - NOTE  | xcodebuild:  error: File exists and is not a symbolic link: /var/folders/bj/n03qf0s968d07kcp9bc4v_vw0000gl/T/CocoaPods/Lint/build/Release/Realm.framework/Headers

Analyzed 1 podspec.

[!] The spec did not pass validation, due to 1 error.
```

mrackwitz:
@jpsim: Good point. I didn't thought about that at all. But I don't see a way around that until CocoaPods is fixed. If we decide that backwards-compatibility to 0.39 is still a requirement at this point, we would likely need to document a post install Podfile-hook as a workaround in the meantime.

PR #3592, opened 2016-05-16
remove all deprecated APIs

jpsim:
/cc @tgoyne @bdash

### DIFF
--- a/Realm/RLMRealm.mm
+++ b/Realm/RLMRealm.mm
@@ -48,6 +48,7 @@ using namespace realm;
 using util::File;
 
 @interface RLMRealm ()
+@property (nonatomic, strong) NSHashTable *notificationHandlers;
 - (void)sendNotifications:(NSString *)notification;
 @end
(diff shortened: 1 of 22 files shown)


Realm/RLMRealm.mm
```diff
@@ -63,10 +64,12 @@ @interface RLMRealmNotificationToken : RLMNotificationToken
 
 @implementation RLMRealmNotificationToken
 - (void)stop {
-#pragma clang diagnostic push
-#pragma clang diagnostic ignored "-Wdeprecated-declarations"
-    [_realm removeNotification:self];
-#pragma clang diagnostic pop
+    [_realm verifyThread];
+    if (auto realmToken = RLMDynamicCast<RLMRealmNotificationToken>(self)) {
```
tgoyne: Don't need this cast.

tgoyne:
Need changelog entry.

👍 other than minor things.

PR #3593, opened 2016-05-16
Mostly remove Mixed property support

### LINKED ISSUE #3580: Remove support for mixed properties
Mixed properties were deprecated in v0.99 (#3442). We should remove them prior to 1.0.

tgoyne:
This leaves in just enough to be able to read from existing Mixed columns in migrations.

Also removes some old pre-Swift 2 cruft related to String/String? being registered oddly with the obj-c runtime.

Closes #3580.


jpsim:
So this will prevent accessing mixed properties but will keep it stored in the Realm. Should we instead be removing those columns during the migration?

tgoyne:
They'll be removed by the normal migration functionality since it's now impossible to declare a model with a mixed property.

jpsim:
> They'll be removed by the normal migration functionality since it's now impossible to declare a model with a mixed property.

Ha, yes that's obvious now... Looks good to me! 👍

Realm/RLMProperty.mm
```diff
@@ -181,9 +173,9 @@ - (BOOL)setTypeFromRawType {
             static const int linkingObjectsPrefixLen = sizeof(linkingObjectsPrefix) - 1;
 
             if (code[1] == '\0') {
-                // string is "@"
-                _type = RLMPropertyTypeAny;
-                _optional = false;
+                @throw RLMException(@"Property '%@' is declared as 'id', which is not a supported RLMObject property type. "
+                                    @"All properties must be primitives, NSString, NSDate, NSData, NSNumber, RLMArray, or subclasses of RLMObject. "
```
bdash: Should this include `RLMLinkingObjects`?

bdash:
👍

PR #3595, opened 2016-05-17
Add missing documentation

jpsim:
Although we detect missing documentation on CI jobs, we failed to update this check when jazzy changed the undocumented file location, so we let a few undocumented APIs slip through. /cc @tgoyne @bdash @austinzheng


bdash:
:+1:

PR #3599, opened 2016-05-17
Empty string shouldn't match anything with `CONTAINS`, `BEGINSWITH` and `ENDSWITH`

kishikawakatsumi:
Now, empty string matches everything in `CONTAINS`, `BEGINSWITH` and `ENDSWITH` query. But `-[NSArray filteredArrayUsingPredicate:]` and CoreData doesn't.


kishikawakatsumi:
@jpsim Should that be fixed? I would like to hear your view.

tgoyne:
IMO Foundation's behavior is wrong, but we should probably match it anyway.

bdash:
Agreed. We'll probably need to add the quirky Foundation behavior at Realm Cocoa's level too since I don't think it makes sense for other bindings.

bdash:
Note that we'd want to fix this for both comparisons against string literals and properties containing the empty string.

bdash:
Will be fixed by #3613.

jpsim:
Closing for #3613.

PR #3603, opened 2016-05-17
Fix Results to not become invalidated when it depends on a deleted object (LinkView or Row)

### LINKED ISSUE #3579: `RLMResults` should have an `invalidated` property
## Goals

Need a way to tell if a `Results` object is invalidated.
http://stackoverflow.com/questions/37201922/how-to-check-for-rlmresults-invalidated

## Expected Results

`Results` should have an `invalidated` property.

## Actual Results

It doesn't.

Realm version: 1.0.2

Xcode version: 7.2.1

iOS/OSX version: 10.11.4

Dependency manager + version: Carthage

bdash:
`Results` is now only invalidated when `invalidate` is called on the `Realm` that it is associated with.

Also exposes an `invalidated` property to `Results` and `LinkingObjects` to allow determining whether a given object is invalidated.

Fixes #3579.

/cc @jpsim @tgoyne @austinzheng

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,10 +5,14 @@ x.x.x Release notes (yyyy-MM-dd)
 
 * All functionality deprecated in previous releases has been removed entirely.
 * Support for Xcode 6.x & Swift prior to 2.2 has been completely removed.
+* `RLMResults`/`Results` now become empty when a `RLMArray`/`List` or row they depend on is deleted,
+  rather than throwing an exception when accessed.
 
 ### Enhancements
 
-* None.
+* Added `invalidated` properties to `RLMResults`/`Results`, `RLMLinkingObjects`/`LinkingObjects`,
+  `RealmCollectionType` and `AnyRealmCollection`. These properties report whether the Realm
+  the object is associated with has been invalidated.
 
 ### Bugfixes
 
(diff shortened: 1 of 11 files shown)


tgoyne:
👍 

Needs changelog entry.

austinzheng:
Looks good.

Is there any use case where a user would want a Results' valid status to depend on a deleted object? The idea is that (e.g.) a list might be deleted, but the Results containing a subset of that list's items is still usable by the application?

bdash:
@austinzheng, are you asking why a user would want to use a `Results` that depends on a deleted object? The short answer is that it can be difficult in an application to know that an object a `Results` depends on has been removed. It's more useful in most contexts for a `Results` to be empty than to throw an exception when accessed.

jpsim:
👍 though it is quite odd that we don't have `invalidated` on RLMRealm at this point. I could have sworn we had that already...

tgoyne:
RLMRealm automatically begins a new read transaction when you do something that requires it, so you can just keep using the Realm after calling `invalidate` on it. OTOH, RLMObject etc. can't be reattached after beginning the new read transaction because we don't have any sort of immutable unique id.

bdash:
Calling `-[RLMRealm invalidate]` doesn't leave the realm itself in an invalid state, only the accessors, so I don't think an `invalidated` property on `RLMRealm` would make sense.

bdash:
CI is failing due to changes to `verify-cocoapods` made as part of #3575.

bdash:
Failing CI tests are unrelated.

PR #3604, opened 2016-05-17
Delete realm rather than run migration if deleteRealmIfMigrationNeeded is set

### LINKED ISSUE #3583: Make deleteRealmIfMigrationNeeded delete the file regardless of schema version changes
It might not be obvious that even if `deleteRealmIfMigrationNeeded` is set to true, migrations are still executed if a higher schema version is specified. We should point out in the API docs, that the Realm file is just deleted if there are outstanding migrations beyond those which are specified with the configuration.

/c @kishikawakatsumi @icanzilb

jpsim:
Closes #3583 /cc @kishikawakatsumi @tgoyne @bdash

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@ x.x.x Release notes (yyyy-MM-dd)
 * Support for Xcode 6.x & Swift prior to 2.2 has been completely removed.
 * `RLMResults`/`Results` now become empty when a `RLMArray`/`List` or row they depend on is deleted,
   rather than throwing an exception when accessed.
+* Migrations are no longer run when `deleteRealmIfMigrationNeeded` is set,
+  recreating the file instead.
 
 ### Enhancements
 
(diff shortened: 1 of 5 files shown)


kishikawakatsumi:
Implementation is 👍 

For tests, I worried about testing only one code path that custom schema was set. It would be nice if there is a test for another code path not to set custom schema. @jpsim

mrackwitz:
Any objections on documenting that even more in detail?

>Recreate the Realm file with the current schema if a migration would be required. This is the case, when the stored schema differs from the provided schema or the stored schema version differs from the version on this configuration. This prevents that schema migration take place.
>Note: This doesn't disable file format migrations.

jpsim:
> For tests, I worried about testing only one code path that custom schema was set

Since we have other tests that cover custom vs shared schema code paths, it's unnecessary to duplicate here IMO.

jpsim:
Thanks for the suggestions to improve documentation @mrackwitz, I've made some modifications.

tgoyne:
👍 

Swiftlint is complaining about trailing whitespace.

PR #3606, opened 2016-05-17
Fix build.sh verify-cocoapods

jpsim:
rather than use pod spec lint, use our own installation examples /cc @bdash @tgoyne

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -644,11 +644,11 @@ case "$COMMAND" in
 
     "verify-cocoapods")
         cd examples/installation
-        test-ios-objc-cocoapods
-        test-ios-swift-cocoapods
-        test-osx-objc-cocoapods
-        test-watchos-objc-cocoapods
-        test-watchos-swift-cocoapods
+        sh build.sh test-ios-objc-cocoapods
+        sh build.sh test-ios-swift-cocoapods
+        sh build.sh test-osx-objc-cocoapods
+        sh build.sh test-watchos-objc-cocoapods
+        sh build.sh test-watchos-swift-cocoapods
         ;;
 
     "verify-osx-encryption")
--- a/build.sh
+++ b/build.sh
@@ -643,36 +643,12 @@ case "$COMMAND" in
         ;;
 
     "verify-cocoapods")
-        pod setup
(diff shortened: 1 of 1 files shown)


build.sh
```diff
@@ -643,36 +643,12 @@ case "$COMMAND" in
...
-          git tag -d $tag
-          git reset --hard HEAD^ || true
-          exit 1
-        )
-        git tag -d $tag
-        git reset --hard HEAD^ || true
         cd examples/installation
-        sh build.sh test-ios-objc-cocoapods || exit 1
-        sh build.sh test-ios-swift-cocoapods || exit 1
+        test-ios-objc-cocoapods
```
bdash: Should this be `sh build.sh test-ios-objc-cocoapods || exit `?

jpsim: I don't think so, because all of `build.sh` has `set -o pipefail; set -e`, so I don't think this was ever needed.

build.sh
```diff
@@ -643,36 +643,12 @@ case "$COMMAND" in
...
-          git tag -d $tag
-          git reset --hard HEAD^ || true
-          exit 1
-        )
-        git tag -d $tag
-        git reset --hard HEAD^ || true
         cd examples/installation
-        sh build.sh test-ios-objc-cocoapods || exit 1
-        sh build.sh test-ios-swift-cocoapods || exit 1
+        sh build.sh test-ios-objc-cocoapods
```
bdash: The installation examples default to testing master if an environment variable named `sha` is not set. I'm not sure that's what we want.

The installation examples are also only capable of testing changes that have been pushed to GitHub. We should consider warning if there are unpushed changes on the branch we're testing so the user isn't surprised that changes aren't being tested when they should be.

jpsim: sure, how do you propose we do that?

bdash: Not sure which part you're asking about.

1. `git rev-parse --abbrev-ref HEAD`
2. `if [[ $(git log -1 @{push}..) == "" ]]; then echo "Good to go."; else echo "Uh-oh, local changes."; fi`

bdash:
CI failures are unrelated.

Issue #3607, opened 2016-05-18
Fix unit test macro


austinzheng:
Right now, `- [EncryptionTests testOpenUnencryptedWithKeyThrows]` is failing on CI. The error description is not very good, and that's because the macro `RLMValidateRealmError` doesn't report failure messages very well. Change `RLMValidateRealmError` to use `RLMAssertMatches`, and root cause the failing test.

PR #3611, opened 2016-05-18
Improving error validation test macro

austinzheng:
The error validation macro should now print out descriptive error messages when things don't pass. The string comparison is a little more robust now, as well. #3607

### DIFF
--- a/Realm/Tests/RLMAssertions.h
+++ b/Realm/Tests/RLMAssertions.h
@@ -63,24 +63,26 @@
     XCTAssertEqual([exception.userInfo[NSUnderlyingErrorKey] code], expectedCode, __VA_ARGS__); \
 })
 
-#define RLMValidateRealmError(macro_error, macro_errnum, macro_description, macro_underlying)        \
-({                                                                                                   \
-    NSString *macro_dsc = macro_description;                                                         \
-    NSString *macro_usl = macro_underlying;                                                          \
-    NSError *macro_castErr = (NSError *)macro_error;                                                 \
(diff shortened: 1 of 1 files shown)


Realm/Tests/RLMAssertions.h
```diff
@@ -63,24 +63,26 @@
...
-    }                                                                                                \
+#define RLMValidateRealmError(macro_error, macro_errnum, macro_description, macro_underlying)            \
+({                                                                                                       \
+    NSString *macro_dsc = macro_description;                                                             \
+    NSString *macro_usl = macro_underlying;                                                              \
+    macro_dsc = [macro_dsc lowercaseString];                                                             \
+    macro_usl = [macro_usl lowercaseString];                                                             \
+    NSError *macro_castErr = (NSError *)macro_error;                                                     \
+    XCTAssertNotNil(macro_castErr);                                                                      \
+    XCTAssertEqual(macro_castErr.domain, RLMErrorDomain, @"Was expecting the error domain '%@', but got '%@' instead", RLMErrorDomain, macro_castErr.domain); \
```
bdash: I don't think the description is necessary here as `XCTAssertEqual` should generate a sensible failure message by default.

austinzheng: I fixed this for the error number test. I decided to keep it for the domain test because this test checks for whether or not the interned copy of the error domain is being used, as well as whether it says the right thing.

jpsim:
seems fine

PR #3612, opened 2016-05-18
Date Parsing during JSON load

tim12s:
Support loading of JSON using Realm.IO in the Javascript Date Time Format as per Javascript JSON convention aligning to ISO8601. e.g. Only "2014-01-01T23:28:56.782Z". At a minimum, this format should be supported as the acceptable standard.

### DIFF
--- a/Realm/RLMObjectBase.mm
+++ b/Realm/RLMObjectBase.mm
@@ -92,7 +92,12 @@ static id RLMValidatedObjectForProperty(id obj, RLMProperty *prop, RLMSchema *sc
             [dateFormatter setDateFormat:@"yyyy-MM-dd'T'HH:mm:ss.SSS'Z'"]; 
         }
         NSDate *date = [dateFormatter dateFromString:obj]; 
-        return date; 
+        if (date) {
+            return date; 
+        } else {
+            // if not convertible to prop throw
+            @throw RLMException(@"Invalid date value '%@' for property '%@' which should be either date or string in format [yyyy-MM-dd'T'HH:mm:ss.SSS'Z']", obj, prop.name);
+        }
     }
 
     // check for object or array of properties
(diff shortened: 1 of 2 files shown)


bdash:
Hi @tim12s, have you seen #694 and the links it contains to frameworks that support mapping from JSON to Realm objects? Those frameworks are likely to be a better place to handle converting dates formatted as strings into `NSDate`s for consumption by Realm.

tim12s:
Hi,

I use ObjectMapper & JSONCodable in some relatively complex mappings.   At this point there are a number of edge cases where I do not want to use ObejctMapper for loading JSON into Realm.  Having said this, I agree that putting custom mappers inside realm is the wrong thing to do and should leave any complex mapping to a library built for it.

This pull request extends the existing basic JSON Loading capability within realm to utilize dates in the most common/acceptable format, since there is no agreed specification.  Any test cases utilizing dates, etc, at the moment, requires some form of object mapper is adds unnecessary cruft.  The different libraries each have a number of "qwirks" when creating Realm objects.

This pull request is a simple enhancement of the basic json loading capability in realm and utilizes the most defensible date format json format .

tim12s:
Hi,

I'm uncertain why verify-docs is failing but it should be I've also added a specific test case and i've corrected an issue identified by a negative regression test.

I dont believe there is anything else that should be done other than agree whether it is acceptable to merge or not.

Thanks,
-Timothy

bdash:
Hi @tim12s, thanks for taking the time to submit this pull request, and my apologies for taking so long to get back to you about it. As I mentioned in an earlier comment Realm doesn't do any parsing of data itself, and we don't intend to add parsing at this time. We feel this behavior is a much better fit for a JSON mapping layer rather than Realm itself.

rromanchuk:
@tim12s hi from the future,  i'm not realm folks always make this suggested, when the suggesting is literally just the same problem we are trying to solve 

> much better fit for a JSON mapping layer rather than Realm itself.

> At this point there are a number of edge cases where I do not want to use ObejctMapper for loading JSON into Realm.

It's not even an edge, it's required if you want to utilize partials, you have to unmap the mappers by hand

PR #3613, opened 2016-05-18
Fix empty strings and null on the right side of BEGINSWITH, ENDSWITH and CONTAINS

bdash:
Foundation never matches an empty or null string with `BEGINSWITH`, `ENDSWITH` and `CONTAINS`. Core's query engine treats them as always matching since they're logically a match at any character position within the string. To emulate Foundation's semantics we generate additional criteria on the core query.

/cc @kishikawakatsumi @tgoyne @jpsim @austinzheng

### DIFF
--- a/Realm/RLMQueryUtil.mm
+++ b/Realm/RLMQueryUtil.mm
@@ -402,10 +402,8 @@ void add_bool_constraint_to_query(realm::Query &query, NSPredicateOperatorType o
 }
 
 
-struct SkipCheckForEmptyAndNullStrings {};
-
 template <typename T>
-void add_substring_constraint_to_query(SkipCheckForEmptyAndNullStrings, Query& query, NSPredicateOperatorType operatorType, bool caseSensitive, Columns<String> column, T value)
+void add_substring_constraint_without_empty_string_checks_to_query(Query& query, NSPredicateOperatorType operatorType, bool caseSensitive, Columns<String> column, T value)
 {
     switch (operatorType) {
         case NSBeginsWithPredicateOperatorType:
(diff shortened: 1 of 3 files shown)


Realm/RLMQueryUtil.mm
```diff
@@ -401,17 +401,12 @@ void add_bool_constraint_to_query(realm::Query &query, NSPredicateOperatorType o
...
-                                    T value) {
-    bool caseSensitive = !(predicateOptions & NSCaseInsensitivePredicateOption);
-    bool diacriticInsensitive = (predicateOptions & NSDiacriticInsensitivePredicateOption);
-    RLMPrecondition(!diacriticInsensitive, @"Invalid predicate option",
-                    @"NSDiacriticInsensitivePredicateOption not supported for string type");
 
+struct SkipCheckForEmptyAndNullStrings {};
+
+template <typename T>
+void add_substring_constraint_to_query(SkipCheckForEmptyAndNullStrings, Query& query, NSPredicateOperatorType operatorType, bool caseSensitive, Columns<String> column, T value)
```
tgoyne: Is this just a weird way of giving the function a different name, or is there some reason it needs to be done this way?

bdash: Leftovers from an earlier approach that I didn't simplify sufficiently. Will fix.

tgoyne:
👍

kishikawakatsumi:
Is matching zero-length data any data type properties OK? Just wondering. Because Foundation doesn't support `CONTAINS`, `BEGINSWITH` and `ENDSWITH` query for NSData properties.

Other than 👍

bdash:
I'd forgotten that we support those operators on binary properties too. We should handle them in the same way we do with strings. I'll update this PR.

MilkBiscuit:
Is there any progress on this bug? 
It does not make sense to match an empty string with KEYWORD search.

We had same problem on Android platform as well.

jpsim:
Sorry, nothing to report yet.

austinzheng:
This PR was revived as #5200 and merged into `master-3.0`.

PR #3614, opened 2016-05-19
remove podspec target checks

jpsim:
since CocoaPods 0.39.0 is the minimum CocoaPods version required, these are guaranteed to be present.

watchos was added in 0.38.0 and tvos was added in 0.39.0

/cc @mrackwitz


mrackwitz:
:+1: That's been overdue. I wanted to tackle that since a while. 😁

PR #3620, opened 2016-05-19
Update to core 1.0.0

### LINKED ISSUE #3601: Creating data values on object after deletion of another object
Hello,
I am having the problem, that when I am deleting a realm object, it is deleting the object but at the same time it is modifying another object.
This image is from the Realm Browser before i delete an object.

realm-vorher

This image is after i deleted it.
realm-nachher

You can see that the data properties, which were nil before, are now having a data value.
I am using Swift, realm 102.1,  Xcode 7.3.1 and the simulator is running iOS 9.3

This is my realm object class and the delete Method
```
import Foundation
import RealmSwift

class MessageRealm: Object {
    dynamic var absenderId: String? = nil
    dynamic var absenderName: String? = nil
    dynamic var datum: NSDate? = nil
    dynamic var gruppenId: String? = nil
    dynamic var image: NSData? = nil
    dynamic var latitude = 0.0
    dynamic var longitude = 0.0
    dynamic var messageId: String? = nil
    dynamic var uebergreifendId : String? = nil
    dynamic var question: String? = nil
    dynamic var sound: NSData? = nil
    dynamic var video: NSData? = nil
    dynamic var livePhoto: NSData? = nil
    dynamic var text: String? = nil
    
    dynamic var wurdeVerschickt: Bool = false
    let runtergeladenAngeschaut = List<RuntergeladenRealm>()
    let answers = List<WertRealm>()
    let results = List<WertRealm>()
    let userIDs = List<WertRealm>()

    override class func primaryKey() -> String {
        return "messageId"
    }

}
```

```

override func collectionView(collectionView: JSQMessagesCollectionView!, didDeleteMessageAtIndexPath indexPath: NSIndexPath!) {
        let message = messages[indexPath.row]

        let nachricht = realm.objectForPrimaryKey(MessageRealm.self, key: message.messageId)
        if nachricht != nil {
            try! self.realm.write({ () -> Void in
                realm.delete(nachricht!)
            })
            
            messages.removeAtIndex(indexPath.row)
            
            
        }
        else {
            messages.removeAtIndex(indexPath.row)
            
        }
        
    }
```

tgoyne:
Removes the bitcode/non-bitcode core logic and adjusts some of the exception translation to deal with that core is now throwing different exceptions for many errors.

Fixes #3601.

### DIFF
--- a/Realm/RLMUtil.mm
+++ b/Realm/RLMUtil.mm
@@ -228,10 +228,24 @@ NSError *RLMMakeError(RLMError code, const realm::util::File::AccessError& excep
 }
 
 NSError *RLMMakeError(RLMError code, const realm::RealmFileException& exception) {
-    NSString *underlying = @(exception.underlying().c_str());
+    // Errors for `open()` now include the path in the exception message, but
+    // RealmFileException's message already has the path, so remove the second
+    // copy as it makes errors much less readable.
+    NSString *underlying;
+    auto pathPos = exception.underlying().find(exception.path());
+    if (pathPos != std::string::npos) {
+        auto trimmed = exception.underlying();
+        trimmed.replace(pathPos, exception.path().size(), "");
+        underlying = @(trimmed.c_str());
+    }
+    else {
+        underlying = @(exception.underlying().c_str());
+    }
+
+    NSString *msg = underlying.length ? [NSString stringWithFormat:@"%s: %@", exception.what(), underlying] : @(exception.what());
     return [NSError errorWithDomain:RLMErrorDomain
                                code:code
-                           userInfo:@{NSLocalizedDescriptionKey: @(exception.what()),
+                           userInfo:@{NSLocalizedDescriptionKey: msg,
                                       NSFilePathErrorKey: @(exception.path().c_str()),
                                       @"Error Code": @(code),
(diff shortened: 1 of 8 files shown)


jpsim:
👍

Realm/RLMUtil.mm
```diff
@@ -228,10 +228,24 @@ BOOL RLMIsObjectValidForProperty(__unsafe_unretained id const obj,
...
+    if (pathPos != std::string::npos) {
+        auto trimmed = exception.underlying();
+        trimmed.replace(pathPos, exception.path().size(), "");
+        underlying = @(trimmed.c_str());
+    }
+    else {
+        underlying = @(exception.underlying().c_str());
+    }
+
+    NSString *msg = underlying.length ? [NSString stringWithFormat:@"%s: %@", exception.what(), underlying] : @(exception.what());
```
bdash: Why is it now desirable to include the underlying message in the localized description?

tgoyne: There's a whole bunch of errors that now result in this exception rather than runtime_error, which resulted in the actual error message moving from what() to underlying().

austinzheng:
Looks good to me. Thanks for improving the error reporting functionality.

PR #3621, opened 2016-05-19
remove redundant `exit 1`s from build.sh

jpsim:
since the whole script is run with `set -o pipefail; set -e` /cc @bdash


mrackwitz:
:+1:

tgoyne:
👍

PR #3628, opened 2016-05-20
Remove +[RLMRealm realmWithURL:key:readOnly:inMemory:dynamic:schema:error]

tgoyne:
The browser has switched to using RLMRealmConfiguration, so the only things left using it were a handful of tests which were easily updated to not use it.


jpsim:
👍

bdash:
:+1:

PR #3629, opened 2016-05-20
Copyediting RLMArray documentation

austinzheng:
@jpsim @bdash @mrackwitz @tgoyne 

I figure many smaller commits will be easier to review than one big commit, and will also lessen the chances of mishandled merge conflicts. If any of you prefer one big commit let me know.

Changes:
- Usage of 'managed' and 'unmanaged' terminology
- Backticks around all type names
- Copyediting for consistency with Apple documentation writing style

### DIFF
--- a/Realm/RLMArray.h
+++ b/Realm/RLMArray.h
@@ -26,36 +26,36 @@ NS_ASSUME_NONNULL_BEGIN
 
 /**
 
- RLMArray is the container type in Realm used to define to-many relationships.
+ `RLMArray` is the container type in Realm used to define to-many relationships.
 
- Unlike an NSArray, RLMArrays hold a single type, specified by the `objectClassName` property.
+ Unlike an `NSArray`, `RLMArray`s hold a single type, specified by the `objectClassName` property.
  This is referred to in these docs as the “type” of the array.
 
- When declaring an RLMArray property, the type must be marked as conforming to a
+ When declaring an `RLMArray` property, the type must be marked as conforming to a
  protocol by the same name as the objects it should contain (see the
- `RLM_ARRAY_TYPE` macro). RLMArray properties can also use Objective-C generics
+ `RLM_ARRAY_TYPE` macro). `RLMArray` properties can also use Objective-C generics
  if available. For example:
 
      RLM_ARRAY_TYPE(ObjectType)
      ...
      @property RLMArray<ObjectType *><ObjectType> *arrayOfObjectTypes;
 
- RLMArrays can be queried with the same predicates as RLMObject and RLMResults.
+ `RLMArray`s can be queried with the same predicates as `RLMObject` and `RLMResult`s.
 
- RLMArrays cannot be created directly. RLMArray properties on RLMObjects are
+ `RLMArray`s cannot be created directly. `RLMArray` properties on `RLMObject`s are
  lazily created when accessed, or can be obtained by querying a Realm.
 
  ### Key-Value Observing
 
- RLMArray supports array key-value observing on RLMArray properties on RLMObject
- subclasses, and the `invalidated` property on RLMArray instances themselves is
- key-value observing compliant when the RLMArray is attached to a persisted
- RLMObject (RLMArrays on standalone RLMObjects will never become invalidated).
+ `RLMArray` supports array key-value observing on `RLMArray` properties on `RLMObject`
+ subclasses, and the `invalidated` property on `RLMArray` instances themselves is
+ key-value observing compliant when the `RLMArray` is attached to a managed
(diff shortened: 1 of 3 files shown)


Realm/RLMArrayLinkView.mm
```diff
@@ -90,7 +90,7 @@ static void throwError() {
         throw;
     }
     catch (realm::InvalidTransactionException const&) {
-        @throw RLMException(@"Cannot modify persisted RLMArray outside of a write transaction");
+        @throw RLMException(@"Cannot modify managed RLMArray outside of a write transaction");
```
austinzheng: Nothing in our code depends on this string, but maybe our users might?

tgoyne: Like most of Foundation, we don't officially support doing anything meaningful after catching the exceptions we throw at all, much less doing things based on the message.

Realm/RLMArray.h
```diff
@@ -349,14 +348,14 @@ NS_ASSUME_NONNULL_BEGIN
 #pragma mark - Unavailable Methods
 
 /**
- -[RLMArray init] is not available because RLMArrays cannot be created directly.
- RLMArray properties on RLMObjects are lazily created when accessed, or can be obtained by querying a Realm.
+ `- (RLMArray)init` is not available because `RLMArray`s cannot be created directly.
```
bdash: Why change from `-[RLMArray init]` to `- (RLMArray)init`? The former is the conventional way of referring to an instance method.

Realm/RLMArray.h
```diff
@@ -26,59 +26,59 @@ NS_ASSUME_NONNULL_BEGIN
...
- Unlike an NSArray, RLMArrays hold a single type, specified by the `objectClassName` property.
+ Unlike an `NSArray`, `RLMArray`s hold a single type, specified by the `objectClassName` property.
  This is referred to in these docs as the “type” of the array.
 
- When declaring an RLMArray property, the type must be marked as conforming to a
+ When declaring an `RLMArray` property, the type must be marked as conforming to a
  protocol by the same name as the objects it should contain (see the
- `RLM_ARRAY_TYPE` macro). RLMArray properties can also use Objective-C generics
+ `RLM_ARRAY_TYPE` macro). `RLMArray` properties can also use Objective-C generics
  if available. For example:
```
tgoyne: "if available" is no longer applicable since we dropped support for Xcode 6. Probably a bunch of other similar little references to generics not being available.

Realm/RLMArray.h
```diff
@@ -349,14 +348,14 @@ NS_ASSUME_NONNULL_BEGIN
...
- RLMArray properties on RLMObjects are lazily created when accessed, or can be obtained by querying a Realm.
+ `- (RLMArray)init` is not available because `RLMArray`s cannot be created directly.
+ `RLMArray` properties on `RLMObject`s are lazily created when accessed, or can be obtained by querying a Realm.
  */
 - (instancetype)init __attribute__((unavailable("RLMArrays cannot be created directly")));
 
 /**
- +[RLMArray new] is not available because RLMArrays cannot be created directly.
- RLMArray properties on RLMObjects are lazily created when accessed, or can be obtained by querying a Realm.
+ `+ (RLMArray)new` is not available because `RLMArray`s cannot be created directly.
```
bdash: Ditto about the change from `+[RLMArray new]` to `+ (RLMArray)new` here.

Realm/RLMArray.h
austinzheng: No reason, I didn't know that was a convention. I'll change them back.

tgoyne: The previous version was correct here.

Realm/RLMArray.h
```diff
@@ -26,59 +26,59 @@ NS_ASSUME_NONNULL_BEGIN
...
- Unlike an NSArray, RLMArrays hold a single type, specified by the `objectClassName` property.
+ Unlike an `NSArray`, `RLMArray`s hold a single type, specified by the `objectClassName` property.
  This is referred to in these docs as the “type” of the array.
 
- When declaring an RLMArray property, the type must be marked as conforming to a
+ When declaring an `RLMArray` property, the type must be marked as conforming to a
  protocol by the same name as the objects it should contain (see the
- `RLM_ARRAY_TYPE` macro). RLMArray properties can also use Objective-C generics
- if available. For example:
+ `RLM_ARRAY_TYPE` macro). In addition, the property can be made generic for better
```
bdash: "made generic" doesn't feel like correct phrasing to me. The _type_ is the thing that can be made generic. When declaring the property you're only choosing whether to opt into to using generics. Something partway between the original and your version seems like a better fit:
> the property can be declared using Objective-C generics for better compile-time type safety

bdash:
👍

PR #3633, opened 2016-05-20
Factor out some duplicated error translation and improve the resulting messages

tgoyne:
`-writeCopyToURL` and `commitWriteTransaction` had their own incomplete copies of the file error translation that produced generally worse errors, so refactor things to reuse the same error translation as when opening Realms, add better tests for the error reporting, and improve some of the relevant error messages a bit (primarily in that the underlying error is now only tacked on in the generic AccessError case, and is done in a more appropriate place).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@ x.x.x Release notes (yyyy-MM-dd)
 
 ### Enhancements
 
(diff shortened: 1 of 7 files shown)


bdash:
:+1:

jpsim:
👍

PR #3634, opened 2016-05-20
Copyediting RLMMigration, RLMResults, and RLMSchema

austinzheng:
@bdash @tgoyne @jpsim @mrackwitz 

Sorry for all the PR spam!

Note that I use the term "results collection"; let me know if you prefer just plain "results".

### DIFF
--- a/Realm/RLMMigration.h
+++ b/Realm/RLMMigration.h
@@ -25,33 +25,35 @@ NS_ASSUME_NONNULL_BEGIN
 @class RLMObject;
 
 /**
-Provides both the old and new versions of an object in this Realm. Object properties can only be
-accessed using keyed subscripting.
-
-@param oldObject Object in original RLMRealm (read-only).
-@param newObject Object in migrated RLMRealm (read-write).
+ A block type which provides both the old and new versions of an object in the Realm. Object 
+ properties can only be accessed using keyed subscripting.
+ 
+ @see `-[RLMMigration enumerateObjects:block:]`
+ 
+ @param oldObject The object from the original Realm (read-only).
+ @param newObject The object from the migrated Realm (read-write).
 */
 typedef void (^RLMObjectMigrationBlock)(RLMObject * __nullable oldObject, RLMObject * __nullable newObject);
 
 /**
- RLMMigration is the object passed into a user defined RLMMigrationBlock when updating the version
- of an RLMRealm instance.
-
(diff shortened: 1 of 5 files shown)


Realm/RLMSchema.h
```diff
@@ -35,38 +35,40 @@ NS_ASSUME_NONNULL_BEGIN
...
 
 #pragma mark - Methods
 
 /**
- Returns an RLMObjectSchema for the given class name in this RLMSchema.
+ Returns an `RLMObjectSchema` for the given class name in the schema.
 
  @param className   The object class name.
- @return            RLMObjectSchema for the given class in this RLMSchema.
+ @return            A `RLMObjectSchema` for the given class in the schema.
```
bdash: An

Realm/RLMMigration.h
```diff
@@ -25,91 +25,97 @@ NS_ASSUME_NONNULL_BEGIN
...
 
 /**
- RLMMigration is the object passed into a user defined RLMMigrationBlock when updating the version
- of an RLMRealm instance.
-
- This object provides access to the RLMSchema current to this migration.
+ `RLMMigration` instances encapsulate information intended to facilitate a schema migration.
+ 
+ A `RLMMigration` instance is passed into a user-defined `RLMMigrationBlock` block when updating
+ the version of a Realm. This instance provides access to the old and new database schemas.
```
tgoyne: It does a lot more than that. Something like "This instance provides access to the old and new database schemas, the objects in the Realm, and provides functionality for modifying the Realm during the migration".

Realm/RLMResults.h
```diff
@@ -24,53 +24,54 @@ NS_ASSUME_NONNULL_BEGIN
...
- RLMResults always reflect the current state of the Realm on the current thread,
+ `RLMResults` always reflect the current state of the Realm on the current thread,
  including during write transactions on the current thread. The one exception to
  this is when using `for...in` fast enumeration, which will always enumerate
  over the objects which matched the query when the enumeration is begun, even if
  some of them are deleted or modified to be excluded by the filter during the
  enumeration.
 
- RLMResults are initially lazily evaluated, and only run queries when the result
+ `RLMResults` are lazily evaluated, and only run queries when the result
```
tgoyne: "initially" is correct here; once you access it the results are eagerly kept up-to-date for as long as you continue to access the results after every write transaction.

PR #3636, opened 2016-05-23
Remove property with type 'id' from the RubyMotion sample

MarkVillacampa:
I have investigated a little and have not seen an alternative to an 'id' property. If someone let's me know an alternative I will be happy to modify the PR with the fix :)


jpsim:
ah, thanks for fixing this @MarkVillacampa! We removed support for mixed properties a few versions ago but forgot about this example!

PR #3643, opened 2016-05-24
Fix reading of optional scalar types via the dynamic API

### LINKED ISSUE #3641: Optional Int property contains nil instead of zero during migration
## Goals

I want to change property type from _optional_ Int to _optional_ String. See RealmBuddy class in my [schema](https://github.com/romanzes/dont-forget-to-say-ios/blob/migration_issue/Don't%20Forget%20To%20Say/RealmModels.swift#L25). To achieve it I use the following [code](https://github.com/romanzes/dont-forget-to-say-ios/blob/migration_issue/Don't%20Forget%20To%20Say/RealmMigration.swift#L33):
``` Swift
migration.enumerate(RealmBuddy.className()) { oldObject, newObject in
    if let contactId = oldObject!["contactId"] as? Int {
        newObject!["contactId"] = "\(contactId)"
    }
}
```

## Expected Results

I expect **nil** Ints to be converted to **nil** Strings.

## Actual Results

Ints with **nil** value are being transformed to **"0"** Strings. It appears that oldObject!["contactId"] equals zero in that case. Strangely, Realm Browser also shows nil values for contactId field as 0 (in file with the old schema). It may seem like they are indeed zeros, but if you load them via code (old version of code, before migration), they will be nils as they should be.

## Steps to Reproduce

1. `git clone git@github.com:romanzes/dont-forget-to-say-ios.git`
2. `cd dont-forget-to-say-ios`
3. `git checkout migration_issue`
4. `pod install`
5. Run Xcode and open [MigrationTests.swift](https://github.com/romanzes/dont-forget-to-say-ios/blob/migration_issue/Dont%20Forget%20To%20Say%20Tests/MigrationTests.swift) file
6. Run test and make sure it fails

## Code Sample

My project is self-educational and open-source so I provide link to it directly: https://github.com/romanzes/dont-forget-to-say-ios/tree/migration_issue. Notice that I put failing test in a separate branch named migration_issue. I also created a branch named [test_v0_realm](https://github.com/romanzes/dont-forget-to-say-ios/tree/test_v0_realm) to demonstrate that provided [.realm](https://github.com/romanzes/dont-forget-to-say-ios/blob/migration_issue/Dont%20Forget%20To%20Say%20Tests/default-v0.realm) file indeed has an entity with nil contactId (it is named "Roman Petrenko").

Realm version: 0.103.1

Xcode version: 7.3

iOS/OSX version: iOS 9.3 / OS X 10.11.4

bdash:
Optional properties of scalar types were getting the wrong Objective-C type code, leading to the dynamic accessors taking the non-optional code path. This resulted in `nil` values being returned as 0.

Fixes #3641.

/cc @tgoyne @jpsim @austinzheng

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,8 @@ x.x.x Release notes (yyyy-MM-dd)
 * Fix an assertion failure which could occur when opening a Realm after opening
   that Realm failed previously in some specific ways in the same run of the
   application.
+* Reading optional integers, floats, and doubles during a migration block now
+  correctly returns `nil` rather than 0 when the stored value is `nil`.
 
 0.103.1 Release notes (2016-05-19)
 =============================================================
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,7 +14,7 @@ x.x.x Release notes (yyyy-MM-dd)
 * Fix an assertion failure which could occur when opening a Realm after opening
   that Realm failed previously in some specific ways in the same run of the
   application.
-* Reading optional integers, floats, and doubles during a migration block now
+* Reading optional integers, floats, and doubles from within a migration block now
   correctly returns `nil` rather than 0 when the stored value is `nil`.
 
 0.103.1 Release notes (2016-05-19)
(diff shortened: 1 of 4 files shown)


tgoyne:
There appears to not be any Swift tests for this either (unless I just can't find them), so it'd be good to add those as well as this is an area they could easily differ.

austinzheng:
Looks good.

jpsim:
👍

tgoyne:
👍

PR #3651, opened 2016-05-25
Optimize accessor class creation

tgoyne:
In total this cuts the time to create the accessor classes in half. In absolute terms it's not an exciting speedup (it was only ~10 ms for our test schema before on an iPad Air running 9.1), but supporting schema changes while the app is running involves creating a lot more accessor classes, which lead to it starting to be an issue.

### DIFF
--- a/Realm/RLMAccessor.mm
+++ b/Realm/RLMAccessor.mm
@@ -486,46 +486,6 @@ static id RLMAccessorUnmanagedSetter(RLMProperty *prop, RLMAccessorCode accessor
     return nil;
 }
 
-// macros/helpers to generate objc type strings for registering methods
-#define GETTER_TYPES(C) C "@:"
-#define SETTER_TYPES(C) "v@:" C
-
-// getter type strings
-// NOTE: this typecode is really the the first character of the objc/runtime.h type
-//       the @ type maps to multiple core types (string, date, array, mixed, any which are id in objc)
-static const char *getterTypeStringForObjcCode(char code) {
-    switch (code) {
-        case 's': return GETTER_TYPES("s");
-        case 'i': return GETTER_TYPES("i");
-        case 'l': return GETTER_TYPES("l");
-        case 'q': return GETTER_TYPES("q");
-        case 'f': return GETTER_TYPES("f");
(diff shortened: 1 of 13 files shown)


Realm/RLMAccessor.mm
```diff
@@ -549,40 +549,40 @@ static void RLMSuperSet(RLMObjectBase *obj, NSString *propName, id val) {
     superSetter(obj, prop.setterSel, val);
 }
 
-// getter/setter for unmanaged
-static IMP RLMAccessorStandaloneGetter(RLMProperty *prop, RLMAccessorCode accessorCode) {
+// getter/setter for standalone
```
jpsim: @austinzheng's been working on moving to consistently refer to standalone objects as "unmanaged"

tgoyne: This file still uses Standalone everywhere except this one comment. The switch in terminology should be done across the entire implementation at once or it just makes things more confusing.

austinzheng: I agree. I didn't want to touch the actual code so close to release. We can go back later and rename everything properly.

jpsim:
This appears to break Realm Swift's reflection, at least in `ObjectUtil.getOptionalProperties(_:)`.

jpsim:
This is looking really good. I think the only things are getting the changelog entry to the right spot and realigning with the "managed/unmanaged" naming that was just merged.

tgoyne:
I've rebased this and brought it back up to date as I found myself wanting the refactored accessor creation code from this for the array of primitives stuff.

Realm/RLMObjectStore.mm
```diff
@@ -41,14 +41,21 @@
 using namespace realm;
 
 void RLMRealmCreateAccessors(RLMSchema *schema) {
+    const size_t bufferSize = sizeof("RLM:Managed  ") // includes null terminator
```
bdash: Is having two trailing spaces here intentional?

tgoyne: There's a space before and after the counter in the generated class name.

bdash: Ah, that's a little subtle.

PR #3657, opened 2016-05-26
Using 'managed'/'unmanaged' everywhere

austinzheng:
This should be the last copyediting/terminology PR. Thanks for your time.

@bdash @tgoyne @jpsim

### DIFF
--- a/Realm/RLMMigration.h
+++ b/Realm/RLMMigration.h
@@ -75,15 +75,17 @@ typedef void (^RLMObjectMigrationBlock)(RLMObject * __nullable oldObject, RLMObj
 
 /**
  Creates and returns an `RLMObject` instance of type `className` in the Realm being migrated.
+ 
+ The `value` argument is used to populate the object. It can be a key-value coding compliant object, an array or 
+ dictionary returned from the methods in `NSJSONSerialization`, or an array containing one element for each managed
+ property. An exception will be thrown if any required properties are not present and those properties were not defined
+ with default values.
+
+ When passing in an `NSArray` as the `value` argument, all properties must be present, valid and in the same order as
+ the properties defined in the model.
 
  @param className   The name of the `RLMObject` class to create.
- @param value       The value used to populate the object. This can be any key-value coding compliant
-                    object, or an array or dictionary returned from the methods in `NSJSONSerialization`, or
-                    an `NSArray` containing one element for each persisted property. An exception will be
-                    thrown if any required properties are not present and those properties were not defined with
-                    default values.
-                    When passing in an `NSArray`, all properties must be present,
-                    valid and in the same order as the properties defined in the model.
+ @param value       The value used to populate the object.
  */
 - (RLMObject *)createObject:(NSString *)className withValue:(id)value;
 
--- a/Realm/ObjectStore/shared_realm.cpp
+++ b/Realm/ObjectStore/shared_realm.cpp
@@ -309,7 +309,7 @@ void Realm::verify_thread() const
 void Realm::verify_in_write() const
 {
     if (!is_in_transaction()) {
-        throw InvalidTransactionException("Cannot modify persisted objects outside of a write transaction.");
+        throw InvalidTransactionException("Cannot modify managed objects outside of a write transaction.");
     }
 }
 
--- a/Realm/RLMAccessor.h
+++ b/Realm/RLMAccessor.h
@@ -35,7 +35,7 @@ NS_ASSUME_NONNULL_BEGIN
 
 // get accessor classes for an object class - generates classes if not cached
 Class RLMAccessorClassForObjectClass(Class objectClass, RLMObjectSchema *schema, NSString *prefix);
-Class RLMStandaloneAccessorClassForObjectClass(Class objectClass, RLMObjectSchema *schema);
+Class RLMUnmanagedAccessorClassForObjectClass(Class objectClass, RLMObjectSchema *schema);
 
 // Check if a given class is a generated accessor class
 bool RLMIsGeneratedClass(Class cls);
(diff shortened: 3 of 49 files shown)


CHANGELOG.md
```diff
@@ -629,7 +629,7 @@ Prebuilt frameworks are now built with Xcode 7.1.
 
 ### Bugfixes
 
-* Setting the primary key property on persisted `RLMObject`s / `Object`s
+* Setting the primary key property on managed `RLMObject`s / `Object`s
```
austinzheng: Is this also true for computed properties?

CHANGELOG.md
```diff
@@ -689,7 +689,7 @@ Prebuilt frameworks are now built with Xcode 7.1.
 
 ### Enhancements
 
-* Persisted properties of `RLMObject`/`Object` subclasses are now Key-Value
+* Managed properties of `RLMObject`/`Object` subclasses are now Key-Value
```
austinzheng: Is this also true for computed properties?

CHANGELOG.md
bdash: Computed properties cannot be set and also cannot be used as the primary key.

jpsim:
I'd rather not modify older parts of the changelog @austinzheng. Otherwise these changes look fine to me. Don't forget to update the cocoa docs in the website repo too!

tgoyne:
I agree on not changing the changelog.

The new wording is incorrect for `createInRealm:withValue:` &c. RLMLinkingObjects/LinkingObjects properties are managed properties, but won't be used from input dictionaries/other KVC objects and cannot appear in input arrays. I don't have any great ideas for how to update the wording to fix this.

austinzheng:
What about this for a docstring?

```
/**
 Creates an instance of a Realm object with a given value, and adds it to the specified Realm.
 
 If nested objects are included in the argument, `createInRealm:withValue:` will be recursively called
 on them.
 
 The `value` argument can be any key-value coding compliant object, an array or dictionary returned from the methods in
 `NSJSONSerialization`, or an `NSArray` containing one element for each managed property. An exception will be thrown if
 any required properties are not present and those properties were not defined with default values. Do not pass in a
 `RLMLinkingObjects` instance, either by itself or as a member of a collection.
 
 When passing in an `NSArray` as the `value` argument, all properties must be present, valid and in the same order as
 the properties defined in the model.

 @param realm    The Realm which should manage the newly-created object.
 @param value    The value used to populate the object.

 @see   `defaultPropertyValues`
 */
```

As far as I can tell it's impossible to give a `param` a multi-paragraph description.

jpsim:
> As far as I can tell it's impossible to give a `param` a multi-paragraph description.

Correct, in both Swift and ObjC, since Jazzy renders doc comments as Markdown, converting `@param` to unordered list items, and AFAICT in Markdown you can't have multiple paragraphs within a single bullet list item.

jpsim:
There are trailing spaces in the Swift files FYI, causing SwiftLint violations.

austinzheng:
Are you okay with the rewritten description of `value`? I want to make sure it's not misleading.

jpsim:
Could you please fix the merge issues here, I'd be happy to take another look afterwards.

austinzheng:
I've rebased onto master. Would appreciate a review whenever you have time to spare. @bdash @tgoyne @jpsim

mrackwitz:
This makes still retrospective modifications to the changelog. Shouldn't those be reverted?

Realm/RLMMigration.h
```diff
@@ -75,15 +75,17 @@ typedef void (^RLMObjectMigrationBlock)(RLMObject * __nullable oldObject, RLMObj
 
 /**
  Creates and returns an `RLMObject` instance of type `className` in the Realm being migrated.
+ 
+ The `value` argument is used to populate the object. It can be a key-value coding compliant object, an array or 
+ dictionary returned from the methods in `NSJSONSerialization`, or an `NSArray` containing one element for each managed
+ property. An exception will be thrown if any required properties are not present and those properties were not defined
+ with default values. Do not pass in a `RLMLinkingObjects` instance, either by itself or as a member of a collection.
```
jpsim: I don't understand how this last sentence is useful.

Realm/RLMMigration.h
```diff
@@ -75,15 +75,17 @@ typedef void (^RLMObjectMigrationBlock)(RLMObject * __nullable oldObject, RLMObj
...
+
+ When passing in an `NSArray` as the `value` argument, all properties must be present, valid and in the same order as
```
jpsim: we interchangeably use "`NSArray`" or "array" in this comment. Let's stick with "array".

jpsim:
Just need to revert changes to changelog and address my two small comments, then 👍 from me.

austinzheng:
I'm going to squash, because not doing so makes reverting stuff too difficult.

austinzheng:
Does anyone have any objection to this API documentation for the following method (and others like it)?

```objective-c
 Creates an `RLMObject` instance of type `className` in the Realm, and populates it using a given object.
 
 The `value` argument is used to populate the object. It can be a key-value coding compliant object, an array or
 dictionary returned from the methods in `NSJSONSerialization`, or an array containing one element for each managed
 property. An exception will be thrown if any required properties are not present and those properties were not defined
 with default values.

 When passing in an array as the `value` argument, all properties must be present, valid and in the same order as the
 properties defined in the model.

 @warning This method is useful only in specialized circumstances, for example, when building components
          that integrate with Realm. If you are simply building an app on Realm, it is recommended to
          use `[RLMObject createInDefaultRealmWithValue:]`.

 @param value    The value used to populate the object.

 @return    An `RLMObject` instance of type `className`.
 */
-(RLMObject *)createObject:(NSString *)className withValue:(id)value;
```

jpsim:
Your last comment seems fine @austinzheng

austinzheng:
Please file issues against me if you want any more documentation changes.

PR #3662, opened 2016-05-27
Fix Swift playground packaging

### LINKED ISSUE #3507: Running Playground Example with RealmSwift Scheme
## Goals

Run the GettingStarted.playground from RealmExamples

## Expected Results

The instructions say: "In the scheme selector choose RealmSwift > iPhone 6s"

## Actual Results

There is no RealmSwift in my scheme selector. Screenshot:

screen shot 2016-04-30 at 12 19 47 pm

## Steps to Reproduce

1. Open RealmExamples.xcworkspace. 
2. In the Project Navigator, open GettingStarted.playground. 
3. Try to follow the instructions there.

## Code Sample

N/A

Mac OS X 10.11.4 15E65

Xcode 7.3 7D175

Realm 0.100.0

jpsim:
Fixes #3507. /cc @austinzheng @tgoyne @bdash 

Xcode playgrounds only support 3rd party frameworks when they're built as part of the Xcode workspace, so we need a bit of a hack to nest the prebuilt frameworks into an empty Swift framework that's imported into the playground. An alternative could have been to package the Realm Swift source files as part of the release zip.

### DIFF
--- a/examples/ios/swift-2.2/RealmExamples.xcodeproj/project.pbxproj
+++ b/examples/ios/swift-2.2/RealmExamples.xcodeproj/project.pbxproj
@@ -7,10 +7,6 @@
 	objects = {
 
 /* Begin PBXBuildFile section */
-		0295B91619D103A70036D6C3 /* libc++.dylib in Frameworks */ = {isa = PBXBuildFile; fileRef = 0295B91519D103A70036D6C3 /* libc++.dylib */; };
-		0295B91719D103BE0036D6C3 /* libc++.dylib in Frameworks */ = {isa = PBXBuildFile; fileRef = 0295B91519D103A70036D6C3 /* libc++.dylib */; };
-		0295B91819D103C60036D6C3 /* libc++.dylib in Frameworks */ = {isa = PBXBuildFile; fileRef = 0295B91519D103A70036D6C3 /* libc++.dylib */; };
-		0295B91919D103CB0036D6C3 /* libc++.dylib in Frameworks */ = {isa = PBXBuildFile; fileRef = 0295B91519D103A70036D6C3 /* libc++.dylib */; };
 		227D20F21CB609A1008F641B /* LaunchScreen.xib in Resources */ = {isa = PBXBuildFile; fileRef = 227D20F11CB609A1008F641B /* LaunchScreen.xib */; };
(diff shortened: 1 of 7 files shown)


jpsim:
rebased

austinzheng:
For what it's worth, it looks fine to me.

bdash:
> Xcode playgrounds only support 3rd party frameworks when they're built as part of the Xcode workspace

Can you elaborate on this requirement? Is this due to the framework search path used by the playground, or something more?

jpsim:
> Can you elaborate on this requirement? Is this due to the framework search path used by the playground, or something more?

The only official documentation on this I could find was [this](https://developer.apple.com/library/ios/recipes/Playground_Help/Chapters/ImportFramework.html), which fails to mention any use of prebuilt frameworks. I tried copying the framework to the search paths, but that wasn't enough.

examples/ios/swift-2.2/RealmExamples.xcodeproj/project.pbxproj
```diff
@@ -1089,6 +1141,56 @@
 			};
 			name = Release;
 		};
+		E8F14D0F1CF8FD0C00564AF5 /* Debug */ = {
+			isa = XCBuildConfiguration;
+			buildSettings = {
+				CLANG_ANALYZER_NONNULL = YES;
```
bdash: A lot of these settings are redundant and repeat those set at the project level. Only the handful that have different values than the project level need to be set on the target.

bdash:
👍

examples/ios/swift-2.2/RealmExamples.xcodeproj/project.pbxproj
jpsim: The only ones I could find being also defined at the project level were `CLANG_ENABLE_MODULES` and `IPHONEOS_DEPLOYMENT_TARGET`. Are there others?

I'm not even sure if I'm doing this right, I'm looking at the build settings for `E80575AF19BA55E700376620` and `E80575B019BA55E700376620`.

bdash: I'd assumed there were more that were duplicates. Out of the others that aren't duplicates, a lot are just unnecessary. I wouldn't bother spending any more time tidying this up though given how obscure this new target is.

PR #3664, opened 2016-05-28
Fixing sequence iterators

austinzheng:
@jpsim 

Changes:
- Updated sequence, collection, and iterator protocols and types for Swift 3
- Restored use of NSFastEnumerationIterator
- Reenabled a test

### DIFF
--- a/Realm/Swift/RLMSupport.swift
+++ b/Realm/Swift/RLMSupport.swift
@@ -29,7 +29,7 @@ extension RLMObject {
     }
 }
 
-public final class RLMGenerator: GeneratorType {
+public final class RLMIterator: IteratorProtocol {
     private let generatorBase: NSFastGenerator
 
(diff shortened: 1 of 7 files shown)


jpsim:
👍

bdash:
Looks like the Realm/Swift directory may need separate versions for Swift 2.2 vs 3.0.

jpsim:
@bdash you're absolutely right, though we haven't started porting `Realm/Tests/Swift` at all yet.

bdash:
I'm referring to RLMSupport.swift, not the tests for using Realm Objective-C from Swift. I think it's a mistake to modify RLMSupport.swift it in place. It should be copied for Swift 3 like RealmSwift was (or #if'd, whichever).

jpsim:
Ah yes, good point!

PR #3672, opened 2016-05-30
Allow to create or update objects with RealmOptional as primary key

### LINKED ISSUE #3671: RealmSwift.Object with numeric optional primary keys can't be added with update=true
## Goals

Add an object of a class defined from Swift with a numeric optional primary key to a Realm.

## Expected Results

Successfully creating the object.

## Actual Results

>** Terminating app due to uncaught exception 'NSUnknownKeyException', reason: '[ valueForUndefinedKey:]: this class is not key value coding-compliant for the key (null).'

## Code Sample

```swift
class Post: Object {
    let postId = RealmOptional<Int>()
    dynamic var postImage : PostImage?
    dynamic var postUser: User?

        return "postId"
    }
 }

class PostTest : XCTestCase {
   func testExample() {
        let realm = try! Realm()
        realm.beginWrite()
        let post = Post()
        post.postId.value = json["image"]["image_id"].intValue
        realm.add(post, update: true)
        try! realm.commitWrite()
   }
}

```

mrackwitz:
Fixes #3671.

/c @kishikawakatsumi @jpsim

### DIFF
--- a/Realm/RLMObjectStore.mm
+++ b/Realm/RLMObjectStore.mm
@@ -182,7 +182,7 @@ void RLMAddObjectToRealm(__unsafe_unretained RLMObjectBase *const object,
 
     // get or create row
     bool created;
-    auto primaryGetter = [=](__unsafe_unretained RLMProperty *const p) { return [object valueForKey:p.getterName]; };
+    auto primaryGetter = [=](__unsafe_unretained RLMProperty *const p) { return [object valueForKey:p.name]; };
     object->_row = (*schema.table)[RLMCreateOrGetRowForObject(schema, primaryGetter, createOrUpdate, created)];
 
     RLMCreationOptions creationOptions = RLMCreationOptionsPromoteStandalone;
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,8 @@ x.x.x Release notes (yyyy-MM-dd)
 
 ### Bugfixes
 
-* None.
+* Swift: Fix an error thrown when trying to create or update `Object` instances via
+  `add(:_update:)` with a primary key property of type `RealmOptional`.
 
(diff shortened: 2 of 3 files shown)


kishikawakatsumi:
👍

Realm/RLMObjectStore.mm
```diff
@@ -182,7 +182,12 @@ void RLMAddObjectToRealm(__unsafe_unretained RLMObjectBase *const object,
 
     // get or create row
     bool created;
-    auto primaryGetter = [=](__unsafe_unretained RLMProperty *const p) { return [object valueForKey:p.getterName]; };
+    auto primaryGetter = [=](__unsafe_unretained RLMProperty *const prop) {
+        if (prop.swiftIvar) {
+            return static_cast<RLMOptionalBase *>(object_getIvar(object, prop.swiftIvar)).underlyingValue;
+        }
+        return [object valueForKey:prop.getterName];
```
kishikawakatsumi: It seems that `[object valueForKey:prop.getterName]` should be `[object valueForKey:prop.name]`?

mrackwitz: Yep, that's better. :+1:

Realm/RLMObjectStore.mm
```diff
@@ -182,7 +182,7 @@ void RLMAddObjectToRealm(__unsafe_unretained RLMObjectBase *const object,
 
     // get or create row
     bool created;
-    auto primaryGetter = [=](__unsafe_unretained RLMProperty *const p) { return [object valueForKey:p.getterName]; };
+    auto primaryGetter = [=](__unsafe_unretained RLMProperty *const p) { return  [object valueForKey:p.name]; };
```
kishikawakatsumi: Extra whitespace?

jpsim:
👍

PR #3682, opened 2016-05-31
Making RLMSupport.swift versioned by Swift version

austinzheng:
@jpsim @tgoyne @bdash 

Would appreciate a review and advice as to whether this is the best way to split up RLMSupport.swift, whenever you have time.

Changes:
- Added separate versions of RLMSupport.swift per Swift version
- Changed build.sh to generate symlink for support file whenever a build is triggered
- build.sh now updates the support file symlink whenever a build is triggered from the command line
- Added a new Xcode build target to generate the Realm support file. This target is now a dependency of the Realm, Realm iOS static, and various test build targets


jpsim:
Given how small `RLMSupport.swift` is, and how we just want users to drag and drop it in their Xcode projects, I think we're better off keeping it to a single file and using Swift version compiler checks for whatever divergences are necessary.

bdash:
If we can't do this with #if's in a single file, I'd suggest we handle this in the same way as for other Swift versioning:

1. Move Realm/Swift -> Realm/Swift-swift2.2
2. Copy Realm/Swift-swift2.2 to Realm/Swift-swift3.0dev, and update as necessary for Swift 3.
3. Update build.sh's `set-swift-version` command to update the Realm/Swift symlink to point at the appropriate directory. This should already be invoked at the necessary places during our build process.

austinzheng:
Is there actually a way to check Swift version using `#if`? As far as I know it only works with build configurations and OS versions.

tgoyne:
Given the fairly small size of RLMSupport.swift it seems reasonable to me to just have a single top-level `#if` and duplicate the entire file for each version, as opposed to finer-grained `#if`s around just the things that differ.

Actually, it may be less annoying to do that for everything than to have two separate files with the symlink switching. Doubling the size of all of the Swift files is mildly unpleasant, but maybe less unpleasant than two sets of files?

PR #3684, opened 2016-05-31
Moving failing test away from identity equality comparison

austinzheng:
@jpsim @tgoyne @bdash 

My primary objective is to run the addition of "isEqual:" past you all. Is identity equality essential to these tests, or do we only care that (in this case) the object is properly added to the end of the array?

If this is approved I'll fix all the tests that broke because of the lack of "isEqual:".

Changes:
- Removed some uses of IUOs (since there are still Swift bugs related to them)
- Added an "isEqual:" overload for SwiftStringObject
- Reenabled a test

### DIFF
--- a/RealmSwift-swift3.0-dev/Object.swift
+++ b/RealmSwift-swift3.0-dev/Object.swift
@@ -223,7 +223,7 @@ public class Object: RLMObjectBase {
 
     - parameter object: Object to compare for equality.
     */
-    public override func isEqual(to object: AnyObject?) -> Bool {
+    public override func isEqual(_ object: AnyObject?) -> Bool {
         return RLMObjectBaseAreEqual(self as RLMObjectBase?, object as? RLMObjectBase)
     }
 
--- a/RealmSwift-swift3.0-dev/Tests/ListTests.swift
+++ b/RealmSwift-swift3.0-dev/Tests/ListTests.swift
@@ -20,8 +20,8 @@ import XCTest
 import RealmSwift
 
 class ListTests: TestCase {
-    var str1: SwiftStringObject!
-    var str2: SwiftStringObject!
+    var str1: SwiftStringObject?
+    var str2: SwiftStringObject?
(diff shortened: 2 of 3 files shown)


tgoyne:
I'm actually not sure why any of those tests other than the Standalone version are passing on Swift 2, since for the other ones `List`'s subscript should be returning a newly created object.

tgoyne:
So, apparently we define `isEqual:` with different semantics on `RLMObject` and `RealmSwift.Object`. The obj-c version only overrides the behavior if a primary key is set, and otherwise just uses pointer identity. The reason for this is that two objects which compare equal to each other need to have the same hash to work as the key of Foundation containers, and without a primary key we don't have anything we can hash to make that true. The Swift version however always overrides equality, and as a result I assume `Object` doesn't work as a key for a Dictionary without a PK?

If Swift 3 is changing things related to how equality works then I guess it'd be a good time to align Object with RLMObject. Rather that comparing the value of the property, I would be inclined to expose `-[RLMObject isEqualToObject:]` (i.e. `RLMObjectBaseAreEqual()`) and either use that in the `isEqual()` override or use it directly in the tests.

tgoyne:
The reason for using `RLMObjectBaseAreEqual()` rather than comparing the property values in those tests is that adding an unmanaged object to a managed List should turn the object into a managed object and not just add a copy of it to the Realm, so we want to verify that that's happened for all of the mutation methods.

austinzheng:
@tgoyne That makes sense, thanks for the detailed explanation. I'll update the diff accordingly.

astigsen:
> t two objects which compare equal to each other need to have the same hash to work as the key of Foundation containers, and without a primary key we don't have anything we can hash to make that true.

How about hashing on a combination of table and row id? It would not be stable (the row id could change), but as long as you dont cache it, it should be valid for a comparison at any point in time.

tgoyne:
Foundation requires that the hash be stable.

PR #3685, opened 2016-05-31
Fixing many more tests

austinzheng:
This accounts for most, but not all the tests.

@tgoyne @jpsim @bdash 

Changes:
- Realm collections are now RandomAccessCollections
- Gave list's "descriptionWithMaxDepth" a custom Objective-C selector, since it wasn't being found
- Reenabled many tests which were disabled because of the `isEqual` bug from the previous commit
- Some tests rewritten to use fewer IUOs

### DIFF
--- a/RealmSwift-swift3.0-dev/List.swift
+++ b/RealmSwift-swift3.0-dev/List.swift
@@ -31,7 +31,7 @@ public class ListBase: RLMListBase {
         return descriptionWithMaxDepth(depth: RLMDescriptionMaxDepth)
     }
 
-    @objc private func descriptionWithMaxDepth(depth: UInt) -> String {
+    @objc(descriptionWithMaxDepth:) private func descriptionWithMaxDepth(depth: UInt) -> String {
         let type = "List<\(_rlmArray.objectClassName)>"
         return gsub(pattern: "RLMArray <0x[a-z0-9]+>", template: type, string: _rlmArray.description(withMaxDepth: depth)) ?? type
     }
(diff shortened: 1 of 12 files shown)


RealmSwift-swift3.0-dev/List.swift
```diff
@@ -31,7 +31,7 @@ public class ListBase: RLMListBase {
         return descriptionWithMaxDepth(depth: RLMDescriptionMaxDepth)
     }
 
-    @objc private func descriptionWithMaxDepth(depth: UInt) -> String {
+    @objc(descriptionWithMaxDepth:) private func descriptionWithMaxDepth(depth: UInt) -> String {
```
tgoyne: Does this bridge correctly without the overridden selector if you exactly match the `description(withMaxDepth:)` that the RLMArray version appears to be brdiged to Swift as?

jpsim: do Swift `@objc` methods get de-clang-imported? In other words, would `description(maxDepth: UInt)` infer the right selector?

austinzheng: The code that is calling this code is Objective-C++; neither `@selector(description:)` nor `@selector(description(maxDepth:))` are accepted.

austinzheng: Changing the Swift function signature to `@objc private func descriptionWithMaxDepth(_ depth: UInt) -> String` seems to work. (I had to explicitly omit the first parameter.)

PR #3692, opened 2016-06-01
Fixes for Swift 3 DP 1

austinzheng:
@bdash @tgoyne @jpsim 

Changes:
- Fixed imported enum cases
- Annotated some APIs with @discardableResult
- Added "_ =" to tests that intentionally ignore return values to quiet warnings

NOTE: JP pointed out a few tests that are still commented out. A subsequent commit will restore them.

### DIFF
--- a/Realm/RLMCollection.mm
+++ b/Realm/RLMCollection.mm
@@ -311,7 +311,7 @@ RLMNotificationToken *RLMAddNotificationBlock(id objcCollection,
                 rethrow_exception(err);
             }
             catch (...) {
-                NSError *error;
+                NSError *error = nil;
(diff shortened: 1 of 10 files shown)


tgoyne:
I sorta think `create()` should be discardable since calling it purely for the side effect is perfectly reasonable.

`migrateRealm()` shouldn't be (actually, it should be throwing an exception...).

austinzheng:
Agreed. It makes sense to just want to persist an object to a Realm directly without further manipulating it.

austinzheng:
I force pushed over the previous commit since there's not much point in comparing the differences. I made a few functional changes.

tgoyne:
👍

PR #3693, opened 2016-06-01
Re-enabling and fixing more tests

austinzheng:
@tgoyne @jpsim @bdash 

Thank you for all your reviews! These are the last failing tests to be enabled for Swift 3; I went through and double-checked.


RealmSwift-swift3.0-dev/LinkingObjects.swift
```diff
@@ -209,7 +209,7 @@ public final class LinkingObjects<T: Object>: LinkingObjectsBase {
      - parameter key:   The name of the property.
      */
     public override func setValue(_ value: AnyObject?, forKey key: String) {
-        return setValue(value, forKeyPath: key)
+        return rlmResults.setValue(value, forKeyPath: key)
```
austinzheng: This was causing infinite recursion.

jpsim:
👍

PR #3697, opened 2016-06-02
Pull in assorted object store changes

tgoyne:
Only functional changes are two small bug fixes; everything else is fixes for things we don't actually do or minor shuffling around for Android support.


bdash:
👍

PR #3712, opened 2016-06-06
explicitly use 'Type.self' notation when passing types to Realm Swift APIs

### LINKED ISSUE #3711: Always use "Type.self" in documentation
According to http://thread.gmane.org/gmane.comp.lang.swift.devel/2148, it turns out that `realm.objects(MyModel)` works due to a bug in Swift, so we should always use an explicit `.self` when referencing types in Realm APIs in our documentation.

Additionally, APIs that take more than one argument _require_ the use of `.self` (e.g. `LinkingObjects(fromType: Person.self, property: "dogs")`) so it's inconsistent to sometimes refer to types without an explicit `.self`.

jpsim:
/cc @austinzheng 

fixes #3711

### DIFF
--- a/RealmSwift-swift2.2/LinkingObjects.swift
+++ b/RealmSwift-swift2.2/LinkingObjects.swift
@@ -343,7 +343,7 @@ public final class LinkingObjects<T: Object>: LinkingObjectsBase {
      the write transaction.
 
      ```swift
-     let dog = realm.objects(Dog).first!
+     let dog = realm.objects(Dog.self).first!
      let owners = dog.owners
      print("owners.count: \(owners.count)") // => 0
      let token = owners.addNotificationBlock { changes in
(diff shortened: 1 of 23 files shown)


austinzheng:
Looks good. Thanks for making these changes!

tgoyne:
I'm still not a fan of using `.self` everywhere, and I don't see a reason to change it now.

tgoyne:
The fact that it working is just a bug isn't news; we knew that when we were first developing the Swift API and it came up in our discussion then about whether or not to omit `.self` when possible.

jpsim:
> The fact that it working is just a bug isn't news; we knew that when we were first developing the Swift API and it came up in our discussion then about whether or not to omit .self when possible.

Actually, we decided to unify our documentation around this a few months ago (though I don't recall where the discussion took place) and consistency was the motivating factor.

jpsim:
I'm merging this since it's just a matter of time until apple/swift@4a60b6c makes its way into a release.

Issue #3337, opened 2016-03-16
Realm Swift detects lazy properties as optional, crashes when attempting to interpret them as RLMOptionalBase


bdash:
A [user on Stack Overflow has reported seeing a crash](http://stackoverflow.com/questions/36024833/realm-adding-object-with-compound-primarykey-error) when attempting to save an object using code that looks like so:

```swift
class MyObject: Object {
    dynamic var someProperty: String = ""
    dynamic lazy var compoundPrimaryKey: String = self.compoundPrimaryKeyValue()
    private func compoundPrimaryKeyValue() -> String { return "SomeValue" }
    override static func primaryKey() -> String? { return "compoundPrimaryKey" }
}

try! Realm().write {
    Realm().add(MyObject())
}
```

This results in either a crash or an exception being thrown:

> 2016-03-15 18:07:57.988 TestHost[12404:378454] -[__NSTaggedDate underlyingValue]: unrecognized selector sent to instance 0x103aa3add

The crash and exception originate from `RLMAddObjectToRealm`:

```
* thread #1: tid = 0x5dec6, 0x00000001046658d6 Realm`RLMAddObjectToRealm(object=0x00006080000e6300, realm=0x000060000067c940, createOrUpdate=false) + 1830 at RLMObjectStore.mm:203, queue = 'com.apple.main-thread', stop reason = breakpoint 1.1
  * frame #0: 0x00000001046658d6 Realm`RLMAddObjectToRealm(object=0x00006080000e6300, realm=0x000060000067c940, createOrUpdate=false) + 1830 at RLMObjectStore.mm:203
    frame #1: 0x0000000104526eaa RealmSwift`Realm.add(object=0x00006080000e6300, update=false, self=0x0000608000028e60) -> () + 714 at Realm.swift:217
    frame #2: 0x00000001041f1c06 RealmSwift Tests`ObjectTests.(storeRealm=0x0000608000028e60) -> ()).(closure #1) + 54 at ObjectTests.swift:346
    frame #3: 0x0000000104529337 RealmSwift`thunk + 39 at Realm.swift:0
    frame #4: 0x0000000104777946 Realm`-[RLMRealm transactionWithBlock:error:](self=0x000060000067c940, _cmd="transactionWithBlock:error:", block=0x000060800004e880, outError=domain: nil - code: 140735238246560) + 86 at RLMRealm.mm:506
[...]

PR #3717, opened 2016-06-06
Prohibiting non-ignored lazy properties in models

austinzheng:
@jpsim @bdash @tgoyne 

Fixing #3337 by prohibiting managed lazy properties in Realm model objects.

Changes:
- Non-ignored lazy properties in models will now cause an exception to be thrown when the schema is built
- Unit tests

NOTE: I'd like to write one more test - ensuring that a Realm Objective-C `RLMObject` subclass defined in Swift with a `lazy var` throws the exception properly. However, it doesn't seem to be possible to use `RLMSetTreatFakeObjectAsRLMObject()` properly from Swift. I'll look into this some more.

### DIFF
--- a/Realm/RLMObjectSchema.mm
+++ b/Realm/RLMObjectSchema.mm
@@ -171,6 +171,16 @@ using namespace realm;
     return schema;
 }
 
++ (nullable NSString *)baseNameForLazySwiftProperty:(NSString *)propertyName {
+    // A Swift lazy var shows up as two separate children on the reflection tree: one named 'x', and another that is
+    // optional and is named 'x.storage'. Note that '.' is illegal in either a Swift or Objective-C property name.
+    NSString *const storageSuffix = @".storage";
+    if ([propertyName hasSuffix:storageSuffix]) {
+        return [propertyName substringToIndex:propertyName.length - storageSuffix.length];
+    }
+    return nil;
+}
+
 + (NSArray *)propertiesForClass:(Class)objectClass isSwift:(bool)isSwiftClass {
     Class objectUtil = [objectClass objectUtilClass:isSwiftClass];
     NSArray *ignoredProperties = [objectUtil ignoredPropertiesForClass:objectClass];
@@ -267,6 +277,18 @@ using namespace realm;
             }
             if (auto type = RLMCoerceToNil(propertyType)) {
                 if (existing == NSNotFound) {
+                    // Check to see if this optional property is an underlying storage property for a Swift lazy var.
+                    // Managed lazy vars are't allowed.
+                    // NOTE: Revisit this once property behaviors are implemented in Swift.
+                    if (NSString *lazyPropertyBaseName = [self baseNameForLazySwiftProperty:propertyName]) {
+                        if ([ignoredProperties containsObject:lazyPropertyBaseName]) {
+                            // This property is the storage property for a ignored lazy Swift property. Just continue.
+                            return;
+                        } else {
+                            @throw RLMException(@"Lazy managed property '%@' is not allowed on a Realm Swift object class. Either add the property to the ignored properties list or make it non-lazy.", lazyPropertyBaseName);
+                        }
+                    }
+                    // The current property isn't a storage property for a lazy Swift property.
                     property = [[RLMProperty alloc] initSwiftOptionalPropertyWithName:propertyName
                                                                               indexed:[indexed containsObject:propertyName]
                                                                                  ivar:class_getInstanceVariable(objectClass, propertyName.UTF8String)
--- a/Realm/Tests/Swift2.2/SwiftPropertyTypeTest.swift
+++ b/Realm/Tests/Swift2.2/SwiftPropertyTypeTest.swift
@@ -103,4 +103,20 @@ class SwiftPropertyTypeTest: RLMTestCase {
         XCTAssertEqual(obj.int32, v32)
         XCTAssertEqual(obj.int64, v64)
     }
+
+    func testLazyVarProperties() {
+        let realm = realmWithTestPath()
+        let succeeded : Void? = try? realm.transactionWithBlock() {
+            realm.addObject(SwiftLazyVarObject())
+        }
+        XCTAssertNotNil(succeeded, "Writing an NSObject-based object with an lazy property should work.")
+    }
+
+    func testIgnoredLazyVarProperties() {
+        let realm = realmWithTestPath()
+        let succeeded : Void? = try? realm.transactionWithBlock() {
+            realm.addObject(SwiftIgnoredLazyVarObject())
+        }
+        XCTAssertNotNil(succeeded, "Writing an object with an ignored lazy property should work.")
+    }
 }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,8 @@ x.x.x Release notes (yyyy-MM-dd)
   transactions have been committed.
 * Fix a "Destruction of mutex in use" assertion failure after an error while
   opening a file.
+* Realm now throws an exception if an `Object` subclass is defined with a managed Swift `lazy` property.
+  Objects with ignored `lazy` properties should now work correctly.
 * Update the LLDB script to work with recent changes to the implementation of `RLMResults`.
 
 1.0.0 Release notes (2016-05-25)
--- a/Realm/Tests/Swift2.2/SwiftTestObjects.swift
+++ b/Realm/Tests/Swift2.2/SwiftTestObjects.swift
@@ -141,3 +141,12 @@ class SwiftLinkTargetObject: RLMObject {
         return ["backlinks": RLMPropertyDescriptor(withClass: SwiftLinkSourceObject.self, propertyName: "link")]
     }
 }
+
+class SwiftLazyVarObject : RLMObject {
+    dynamic lazy var lazyProperty : String = "hello world"
+}
+
+class SwiftIgnoredLazyVarObject : RLMObject {
+    dynamic lazy var ignoredVar : String = "hello world"
+    override class func ignoredProperties() -> [String] { return ["ignoredVar"] }
+}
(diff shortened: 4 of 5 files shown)


Realm/Tests/Swift2.2/SwiftPropertyTypeTest.swift
```diff
@@ -103,4 +103,12 @@ class SwiftPropertyTypeTest: RLMTestCase {
         XCTAssertEqual(obj.int32, v32)
         XCTAssertEqual(obj.int64, v64)
     }
+
+    func testIgnoredLazyVarProperties() {
+        let realm = realmWithTestPath()
+        let succeeded : Void? = try? realm.transactionWithBlock() {
+            realm.addObject(SwiftIgnoredLazyVarObject())
+        }
+        XCTAssertNotNil(succeeded, "Writing an object with an ignored lazy property should work.")
```
jpsim: this should check `succeeded` for `true`, not just non-nil.

jpsim: sorry, no I guess this is correct, my mistake

jpsim:
LGTM 👍

austinzheng:
I added a test for an `NSObject`-derived model object with lazy properties. In this case there's no issue - the 'shadow' optional property isn't created, since the object isn't a Swift object, and everything works as expected.

Realm/RLMObjectSchema.mm
```diff
@@ -171,6 +171,21 @@ + (instancetype)schemaForObjectClass:(Class)objectClass {
     return schema;
 }
 
++ (bool)_isInvalidLazyProperty:(NSString *)propertyName ignoredProps:(NSArray *)ignoredProperties {
+    NSString *const storageSuffix = @".storage";
+    const NSUInteger suffixLength = [storageSuffix length];
```
bdash: The various accesses of the `length` property can use use dot notation.

Realm/RLMObjectSchema.mm
```diff
@@ -171,6 +171,21 @@ + (instancetype)schemaForObjectClass:(Class)objectClass {
...
+    if (![ignoredProperties containsObject:propertyName]
+        && [propertyName length] > suffixLength
+        && [propertyName hasSuffix:storageSuffix]) {
+        // Snip the name
```
bdash: Not sure this adds any value.

Realm/RLMObjectSchema.mm
```diff
@@ -171,6 +171,21 @@ + (instancetype)schemaForObjectClass:(Class)objectClass {
     return schema;
 }
 
++ (bool)_isInvalidLazyProperty:(NSString *)propertyName ignoredProps:(NSArray *)ignoredProperties {
+    NSString *const storageSuffix = @".storage";
+    const NSUInteger suffixLength = [storageSuffix length];
+    if (![ignoredProperties containsObject:propertyName]
+        && [propertyName length] > suffixLength
+        && [propertyName hasSuffix:storageSuffix]) {
```
bdash: Checking `[propertyName length] > suffixLength` seems redundant since `-hasSuffix:` would have to perform a similar check internally in order to return true.

austinzheng:
Thanks @bdash! I tightened up the logic and made some other fixes.

Realm/RLMObjectSchema.mm
```diff
@@ -267,6 +278,12 @@ + (NSArray *)propertiesForClass:(Class)objectClass isSwift:(bool)isSwiftClass {
             }
             if (auto type = RLMCoerceToNil(propertyType)) {
                 if (existing == NSNotFound) {
+                    // Check to see if this optional property is an underlying storage property for a Swift lazy var.
+                    // Managed lazy vars are't allowed.
+                    // NOTE: Revisit this once property behaviors are implemented in Swift.
+                    if ([self _isDisallowedLazyProperty:propertyName ignoredProps:ignoredProperties]) {
+                        @throw RLMException(@"Lazy managed properties are not allowed on Realm Swift object classes.");
```
mrackwitz: The exception message could include a reference to the property name and a hint that this needs to be either ignored or not defined as lazy.

mrackwitz:
LGTM :+1: beside my comment.

Realm/RLMObjectSchema.mm
```diff
@@ -171,6 +171,24 @@ + (instancetype)schemaForObjectClass:(Class)objectClass {
     return schema;
 }
 
++ (bool)_isDisallowedLazyProperty:(NSString *)propertyName
+                     ignoredProps:(NSArray *)ignoredProperties
+                      snippedName:(NSString **)snippedName {
+    // A Swift lazy var shows up as two separate children on the reflection tree: one named 'x', and another that is
```
bdash: It might be clearer for this method to become `+baseNameForLazySwiftProperty:` that returns the base property name or nil when it's not the name of the storage of the lazy Swift property, and to have `+propertiesForClass:isSwift:` handle checking if it's ignored and throwing the exception.

If that's not an improvement, I'd suggest `s/ignoredProps/ignoredProperties/` and `s/snippedName/basePropertyName/`.

austinzheng: Yeah, I think that's better than trying to return both a boolean and a String.

Realm/RLMObjectSchema.mm
```diff
@@ -171,22 +171,14 @@ + (instancetype)schemaForObjectClass:(Class)objectClass {
...
-        if (![ignoredProperties containsObject:theSnippedName]) {
-            if (snippedName) {
-                *snippedName = theSnippedName;
-            }
-            return true;
-        }
-    }
-    return false;
++ (nullable NSString *)baseNameForLazySwiftProperty:(NSString *)propertyName {
+  // A Swift lazy var shows up as two separate children on the reflection tree: one named 'x', and another that is
```
bdash: Four spaces for indentation, please!

austinzheng: Sorry! I wish Xcode had per-project indentation settings...

Realm/RLMObjectSchema.mm
```diff
@@ -281,14 +281,21 @@ + (NSArray *)propertiesForClass:(Class)objectClass isSwift:(bool)isSwiftClass {
...
-                    if (lazyPropertyBaseName && ![ignoredProperties containsObject:lazyPropertyBaseName]) {
-                        @throw RLMException(@"Lazy managed property '%@' is not allowed on a Realm Swift object class. Either add the property to the ignored properties list or make it non-lazy.", lazyPropertyBaseName);
+                    if (lazyPropertyBaseName) {
+                        if ([ignoredProperties containsObject:lazyPropertyBaseName]) {
+                            // This property is the storage property for a ignored lazy Swift property. Just continue.
+                            return;
+                        } else {
+                            @throw RLMException(@"Lazy managed property '%@' is not allowed on a Realm Swift object class. Either add the property to the ignored properties list or make it non-lazy.", lazyPropertyBaseName);
+                        }
+                    } else {
```
bdash: Strictly speaking there's no need for the `else` block given that all paths out of the `if` block leave the function.

austinzheng: Yeah. I wanted to make it a little more explicit. Would you prefer a coding style that didn't have the `else`?

Realm/RLMObjectSchema.mm
```diff
@@ -281,14 +281,21 @@ + (NSArray *)propertiesForClass:(Class)objectClass isSwift:(bool)isSwiftClass {
                     // Managed lazy vars are't allowed.
                     // NOTE: Revisit this once property behaviors are implemented in Swift.
                     NSString *lazyPropertyBaseName = [self baseNameForLazySwiftProperty:propertyName];
-                    if (lazyPropertyBaseName && ![ignoredProperties containsObject:lazyPropertyBaseName]) {
-                        @throw RLMException(@"Lazy managed property '%@' is not allowed on a Realm Swift object class. Either add the property to the ignored properties list or make it non-lazy.", lazyPropertyBaseName);
+                    if (lazyPropertyBaseName) {
```
bdash: Since this is Objective-C++ you can pull the declaration / initialization of `lazyPropertyBaseName` into the conditional:

```objc
if (NSString *lazyPropertyBaseName = [self baseNameForLazySwiftProperty:propertyName]) {
    …
}
```

This also has the advantage of limiting the scope of the variable to the `if`'s body.

Realm/RLMObjectSchema.mm
bdash: I personally tend to avoid `else`s after `return`s, but I don't think that we as a team have a guideline one way or the other.

bdash:
:+1:

Realm/RLMObjectSchema.mm
```diff
@@ -267,6 +277,18 @@ + (NSArray *)propertiesForClass:(Class)objectClass isSwift:(bool)isSwiftClass {
             }
             if (auto type = RLMCoerceToNil(propertyType)) {
                 if (existing == NSNotFound) {
+                    // Check to see if this optional property is an underlying storage property for a Swift lazy var.
+                    // Managed lazy vars are't allowed.
+                    // NOTE: Revisit this once property behaviors are implemented in Swift.
+                    if (NSString *lazyPropertyBaseName = [self baseNameForLazySwiftProperty:propertyName]) {
+                        if ([ignoredProperties containsObject:lazyPropertyBaseName]) {
+                            // This property is the storage property for a ignored lazy Swift property. Just continue.
+                            return;
```
bdash: Shouldn't this be `continue`, not `return`?

austinzheng: I think `return` is correct. This code is running in an `enumerateKeysAndObjectsUsingBlock` block.

bdash: Right you are. The `for` loop I was remembering is further out than that.

Issue #3335, opened 2016-03-15
FIFO for notifications should move into new .management directory alongside lock files


bdash:
Realm core is moving the lock and log files into a `.management` directory alongside the realm (e.g., `foo.realm.management` for `foo.realm`). We should move the `.note` file that we create for interprocess notifications into the directory as well.

austinzheng:
We should figure out whether this counts as a "breaking change", if we want to do it after all, and if so how we want to mitigate any effects on our users.

bdash:
Can we phase this move in over multiple releases to reduce the impact on users?

For instance, in v1.1 we could hard link the fifo into the management directory so it can be opened from either location. In a future version we could switch to only creating the fifo inside the management directory. I've not spent much time thinking about whether this can be done safely in the presence of multiple processes though.

austinzheng:
I don't see any implementation issues with doing so. Would the idea be to switch to only creating the FIFO inside the directory at the same time we make other breaking changes?

bdash:
It wouldn't necessarily have to coincide with other breaking changes. It's unlikely that a user will want an app using v1.0 to use the same file as another of their apps using v1.5, so we might be able to get away with saying that files can only be shared between processes if they're all using "similar enough" versions of Realm.

austinzheng:
That's a good point. Is there a way we can enforce this so users don't get weird behavior where notifications aren't delivered?

What is the advantage of moving the file into the `management` directory, by the way? Just for consistency?

bdash:
It's weird to have a subdirectory for the extra files we create alongside the Realm but be creating extra files outside of it as well.

PR #3727, opened 2016-06-07
Putting FIFO file in new "realm.management" directory

austinzheng:
Per #3335.

@tgoyne @jpsim @bdash

### DIFF
--- a/Realm/ObjectStore/impl/apple/external_commit_helper.cpp
+++ b/Realm/ObjectStore/impl/apple/external_commit_helper.cpp
@@ -96,7 +96,7 @@ ExternalCommitHelper::ExternalCommitHelper(RealmCoordinator& parent)
 
 #if !TARGET_OS_TV
     auto dir_path = parent.get_path() + ".management/";
-    auto path = dir_path + ".note";
+    auto path = dir_path + "notification_pipe.note";
 
     // Create and open the named pipe
     int ret = mkdir(dir_path.c_str(), 0766);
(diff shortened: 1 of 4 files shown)


tgoyne:
A file named just ".note" is weird. Should be something like "notification_pipe".

Moving this file breaks sharing the realm file with previous versions of Realm. We haven't really decided if we consider that a breaking change.

austinzheng:
I think I took the ticket description too literally. I'll fix the file name.

austinzheng:
Is there anything I can do to move this ticket forward? If I create a symlink in the original location will that be enough to get it approved?

Realm/ObjectStore/impl/apple/external_commit_helper.cpp
```diff
@@ -95,10 +95,20 @@ ExternalCommitHelper::ExternalCommitHelper(RealmCoordinator& parent)
     }
 
 #if !TARGET_OS_TV
-    auto path = parent.get_path() + ".note";
+    auto dir_path = parent.get_path() + ".management/";
+    auto path = dir_path + "notification_pipe.note";
```
jpsim: does this need an extension? Weird to call it "note{...}.note". Why not just `notification_pipe`?

jpsim:
A symlink in the current location seems sensible as a transition step. Seems like we ought to find a way to make this relevant to users in the changelog too.

jpsim:
Putting this "on hold" as it hasn't been actively maintained.

austinzheng:
Closing this PR since it would be far less work to just redo it from scratch than to try rebasing it onto HEAD.

Issue #3725, opened 2016-06-07
APIs for helping people unit test with Realm


austinzheng:
As a follow-up to #3345:

How should we expose APIs or other affordances to help people write correct unit tests that involve Realms?

I talked with Thomas briefly about just exposing `[RLMRealm resetRealmState]` to public users. It turns out that this method is difficult to use correctly (i.e. in a way that won't cause tests to crash nondeterministically). 

We have some [shared test code](https://github.com/realm/realm-cocoa/blob/master/Realm/Tests/RLMTestCase.m#L127) that we use for our own tests. Perhaps we could expose an XCTest-agnostic helper class that abstracts out whatever we need to do to make tests run correctly, and have users use that instead? Strawman example:

```swift
class RealmUnitTestHelper {
    init(setupBlock: RealmTestBlock = nil, teardownBlock: RealmTestBlock = nil)

    // In-memory Realm is named automatically
    func runTest(testBlock: RealmTestBlock)

    func runTestWithRealmNamed(name: String, testBlock: RealmTestBlock)
}

// Usage:
class MyTest : XCTestCase {
    let helper = RealmUnitTestHelper(setupBlock: { ... }, teardownBlock: { ... })

    func testFoo() {
        helper.runTest() { ... }
    }

    func testBar() {
        helper.runTest() { ... }
    }
}
```

Any other ideas are welcome.

@bdash @tgoyne @jpsim @mrackwitz @kishikawakatsumi @stel

tgoyne:
Having to wrap the body of every test in an extra block is somewhat unpleasant: it results in an extra level of indentation for everything, and the extra boilerplate encourages doing more in each test body rather than splitting things into as small of tests as possible. Something along those lines would be to provide a wrapper for `invokeTest()`:

    class MyTest : XCTestCase {
        let helper = RealmUnitTestHelper(setupBlock: { ... }, teardownBlock: { ... })

        override func invokeTest() {
            helper.invokeTest { super.invokeTest() }
        }

        func testDefaultRealm() {
            let realm = Realm() // default realm is set to a per-test file
            // ... do stuff ...
        }

        func testNamedRealm() {
            let realm = helper.realmNamed("named for in memory realm")
            // ... do stuff ...
        }

        func testCustomConfig() {
            let realm = helper.configuationWithName("name")
        }
    }

austinzheng:
[...]

PR #3730, opened 2016-06-08
Adding unit test helper class

austinzheng:
@tgoyne @bdash @jpsim

Here is a helper class that I'm hoping can help our users write more robust unit tests involving Realms. It's basically a helper class based off our `RLMTestCase` base test case, but with no XCTest dependencies.

The idea is that a user would create an instance of this class belonging to every test case class they make which uses Realms. They'd have to override `invokeTest` on their test class to call the helper's `invokeTest` method. They'd use this class's methods to perform async and sync work, but write their tests as usual otherwise.

This isn't so much of a review as it is asking what you think of the basic concept. How is the API? Could it use more helper methods? Is it too cumbersome? Let me know your thoughts.

(If this is something we want to pursue further I'll add documentation and a Swift wrapper.)

### DIFF
--- a/Realm/RLMUnitTestHelper.h
+++ b/Realm/RLMUnitTestHelper.h
@@ -1,0 +1,73 @@
+////////////////////////////////////////////////////////////////////////////
+//
+// Copyright 2016 Realm Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+////////////////////////////////////////////////////////////////////////////
+
+#import <Foundation/Foundation.h>
+
+@class RLMRealm, RLMSchema;
+
+/**
+ A helper class intended for use when writing unit tests involving Realms, especially those involving the use of GCD.
+
+ To use this class, create a property on your `XCTestCase` class of type `RLMUnitTestHelper`, and either instantiate the
+ test helper upon initialization or lazily when requested.
+
+ Your `XCTestCase` test case subclass must override the `-invokeTest` method. In that method, call the helper's
+ `invokeTestWithBlock:` method. Inside the block passed to `-invokeTestWithBlock:`, call `[super invokeTest]`. You can
+ also do additional work within the block if necessary.
+
+ The test helper overrides the default Realm configuration to use either a special file or special in-memory identifier.
+ Each test's Realms are completely isolated from the Realms of other tests. All internal state will automatically be
+ reset after each test.
+
+ Obtain this special Realm for test use simply by calling `[RLMRealm defaultRealm]`. You may also obtain Realms with
+ other configurations explicitly, as usual.
+
+ Instead of using GDC functions directly, use the `-dispatch:` and `dispatchAndWait:` methods instead.
+ */
+@interface RLMUnitTestHelper : NSObject
+
+/**
+ An on-disk test Realm for your unit test to use. The Realm is cleaned up and properly destroyed after each test suite.
+ */
+@property (nonatomic, readonly) RLMRealm *onDiskTestRealm;
+
+/**
+ An in-memory test Realm for your unit test to use. The Realm is properly destroyed after each test suite.
+ */
+@property (nonatomic, readonly) RLMRealm *inMemoryTestRealm;
+
+/**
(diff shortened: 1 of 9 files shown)


tgoyne:
I think this is definitely something that'd be useful to have. A good next step would be to make RLMTestCase use it, and try to make it just a thing wrapper around the helper. There's probably some functionality which doesn't make sense to push into the helper, but the bulk of it would probably be good to expose.

austinzheng:
I tried pushing as much of `RLMTestCase`'s functionality into the helper as possible. There are a few methods that are either really specific to testing Realm itself, or too heavily dependent upon XCTestCase's APIs to abstract out properly. Let me know what you think.

Realm/RLMUnitTestHelper.h
```diff
@@ -0,0 +1,68 @@
...
+
+ To use this class, create a property on your `XCTestCase` class of type `RLMUnitTestHelper`, and either instantiate the
+ test helper upon initialization or lazily when requested.
+
+ Your `XCTestCase` test case subclass must override the `-invokeTest` method. In that method, call the helper's
+ `invokeTestWithBlock:` method. Inside the block passed to `-invokeTestWithBlock:`, call `[super invokeTest]`. You can
+ also do additional work within the block if necessary.
+
+ The test helper overrides the default Realm configuration to use either a special file or special in-memory identifier.
+ Each test's Realms are completely isolated from the Realms of other tests. The state of the underlying bindings will
```
jpsim: rephrase to "All internal state will automatically be reset after each test"?

Realm/RLMUnitTestHelper.h
```diff
@@ -0,0 +1,68 @@
...
+ automatically be reset after each test. Obtain this special Realm for test use simply by calling `[RLMRealm realm]`.
```
jpsim: what's `[RLMRealm realm]`?

Realm/RLMUnitTestHelper.m
```diff
@@ -0,0 +1,117 @@
...
+}
+
+- (instancetype)initUsingInMemoryRealms:(BOOL)useInMemoryRealms {
+    if (self = [super init]) {
+        NSURL *url = nil;
+        if (!useInMemoryRealms) {
+            NSFileManager *manager = [NSFileManager defaultManager];
+            NSError *error = nil;
+            url = [manager URLForDirectory:NSCachesDirectory
+                                         inDomain:NSUserDomainMask appropriateForURL:nil create:YES error:&error];
```
jpsim: we usually align `:`s in Objective-C

Realm/RLMUnitTestHelper.h
austinzheng: Oops! That should be `[RLMRealm defaultRealm]`. Sorry!

jpsim:
Seems kludgy to override the default configuration for this. Why not a `RLMUnitTestHelper.testRealm` property that `XCTFail`s on access if creating the Realm fails?

jpsim:
That's all the feedback I have for now. I may have more when this is rebased.

austinzheng:
@jpsim I would think that overriding the default configuration would be more elegant, as the test writer could just use the default Realm. However, I defer to your judgment. Having a specific `testRealm` would also make the implementation cleaner, so I'm +1 on that overall.

austinzheng:
I made the requested changes. The test helper now has two properties for user convenience, one for an in-memory Realm and one for an on-disk Realm. If a user _really_ wants to make the test Realm the default Realm, they can just set the default configuration to that of one of the test Realms.

austinzheng:
#3725 (linking issue)

Realm/RLMUnitTestHelper.h
```diff
@@ -0,0 +1,73 @@
...
+ also do additional work within the block if necessary.
+
+ The test helper overrides the default Realm configuration to use either a special file or special in-memory identifier.
+ Each test's Realms are completely isolated from the Realms of other tests. All internal state will automatically be
+ reset after each test.
+
+ Obtain this special Realm for test use simply by calling `[RLMRealm defaultRealm]`. You may also obtain Realms with
+ other configurations explicitly, as usual.
+
+ Instead of using GDC functions directly, use the `-dispatch:` and `dispatchAndWait:` methods instead.
```
jpsim: GCD

jpsim:
Can you also update the Swift tests to use this helper?

austinzheng:
An update. I had to port the helper over to Swift, rather than simply using a typealias, because `RLMRealm` and `Realm` aren't interchangeable. I have it integrated into the Swift unit tests, but some of them are still failing and I'm trying to figure out why. Given that our unit tests are very different from those we expect our customers to write (we test features on Realms themselves; they write tests that happen to involve Realms), I may just back out on the deep integration into `TestCase.swift`.

jpsim:
Marking `S:On Hold` until @austinzheng can get back to this.

jpsim:
Closing due to severe age. Tracked in #5029.

PR #3733, opened 2016-06-09
Add some documentation for readOnly

### LINKED ISSUE #3731: readonly Realm does not track changes
I noticed that a (Swift) Realm with the readOnly flag specified in its Configuration will not track changes made in another process, which it does when opening the database read-write.
This makes sense (kind of), since the readOnly feature seems to be intended for use with databases within the application bundle that will never change. Still, this behaviour should be documented.

## Goals
I have an app with a bundled Today extension using a Realm database in a shared App Group location. The extension does not modify the database, hence I would like to open it read-only.
Why? The extension and the main app share a lot of code and I would like to minimise the chance of inadvertent changes to the database.

## Expected Results
When the widget updates, I expect previously fetched objects from the database to be up-to-date.

## Actual Results
Stale data is returned from fetched objects.
Attempts to use `realm.refresh()` produce a crash.

## Mitigation
When setting `configuration.readOnly = false`, it works as expected.

```
ProductName:	Mac OS X
ProductVersion:	10.11.5
BuildVersion:	15F34

Xcode 7.3.1
Build version 7D1014

/usr/local/bin/pod
1.0.0
(not in use here)

/bin/bash

/usr/local/bin/carthage
0.16.2
github "realm/realm-cocoa" "v1.0.0"

/usr/bin/git
git version 2.7.4 (Apple Git-66)
```

tgoyne:
AFAICT we didn't actually say anywhere that you couldn't open a file in both read-only and read-write mode at the same time.

Closes #3731.

### DIFF
--- a/Realm/RLMRealmConfiguration.h
+++ b/Realm/RLMRealmConfiguration.h
@@ -64,7 +64,15 @@ NS_ASSUME_NONNULL_BEGIN
 /// A 64-byte key to use to encrypt the data, or `nil` if encryption is not enabled.
 @property (nonatomic, copy, nullable) NSData *encryptionKey;
 
-/// Whether the Realm is read-only (must be `YES` for read-only files).
+/// Whether to open the Realm in read-only mode.
+///
+/// This is required to be able to open Realm files which are not writeable or
+/// are in a directory which is not writeable. This should only be used on files
+/// which will not be modified by anyone while they are open, and not just to
+/// get a read-only view of a file which may be written to by another thread or
+/// process. Opening in read-only mode requires disabling Realm's reader/writer
+/// coordination, so committing a write transaction from another process will
+/// result in crashes.
 @property (nonatomic) BOOL readOnly;
 
 /// The current schema version.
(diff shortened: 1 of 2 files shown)


Realm/RLMRealmConfiguration.h
```diff
@@ -64,7 +64,15 @@ NS_ASSUME_NONNULL_BEGIN
 /// A 64-byte key to use to encrypt the data, or `nil` if encryption is not enabled.
 @property (nonatomic, copy, nullable) NSData *encryptionKey;
 
-/// Whether the Realm is read-only (must be `YES` for read-only files).
+/// Whether to open the Realm in read-only mode.
+///
+/// This is required to be able to open Realm files which are not writeable or
+/// are in a directory which is not writeable. This should only be used on files
+/// which will be modified by anyone while they are open, and not just to get a
```
austinzheng: "which will not be modified"?

austinzheng:
Line comment aside, looks good.

jpsim:
👍

PR #3737, opened 2016-06-10
Ensure that the initial notification for Results is always .Initial

### LINKED ISSUE #3501: RealmCollectionChange.Initial notification not always received
Newly introduced RealmCollection.addNotificationBlock method not always works as expected. I need to track down only when items are inserted or deleted from the collection. But sometimes the first passed value is `.Update` while `.Initial` is not passed at all. So I forced to use my own `_initialTags` flag. Here's the code:

```
var _initialTags = false
tokens.append(hashtags.addNotificationBlock { (changes: RealmCollectionChange<Results<Hashtag>>) in
    switch changes {
    case .Update(_, let deletions, let insertions, _):
        if deletions.isEmpty && insertions.isEmpty && _initialTags { break }
            fallthrough
    case .Initial:
        self.storedHashtags.removeAll()
        self.storedHashtags.appendContentsOf(self.hashtags)
        if !self.searchMode { self.updateOutline() }
        _initialTags = true
        break
    case .Error(let error):
        fatalError("\(error)")
    }
})
```

```
ProductName:	Mac OS X
ProductVersion:	10.11.4
BuildVersion:	15E65

Xcode 7.3
Build version 7D175

~/.gem/ruby/2.0.0/bin/pod
0.39.0
Realm (0.99.1)
RealmSwift (0.99.1)

/bin/bash


(not in use here)

/usr/bin/git
git version 2.6.4 (Apple Git-63)
```

tgoyne:
There was logic which attempted to ensure this was the case, but it wasn't
tested, wasn't actually used, and didn't actually work.

Fixes #3501.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,9 @@ x.x.x Release notes (yyyy-MM-dd)
 * Fix a crash when reading the shared schema from an observed Swift object.
 * Fix crashes or incorrect results when passing an array of values to
   `createOrUpdate` after reordering the class's properties.
+* Ensure that the initial call of a Results notification block is always passed
+  .Initial even if there is a write transaction between when the notification
+  is added and when the first notification is delivered.
 
 1.0.1 Release notes (2016-06-12)
 =============================================================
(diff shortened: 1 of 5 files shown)


tgoyne:
@bdash

bdash:
👍

PR #3742, opened 2016-06-13
Adding Swift 3 preview support

austinzheng:
This PR adds Swift 3 preview support to our master branch, per JP's request. The code compiles/tests pass as of the June 6 developer snapshot.

Let me know if you have any questions, or if anything should be changed etc.

@tgoyne @jpsim @bdash


jpsim:
I think this is good for now, since it shouldn't affect anyone not using the Swift 3 preview. 👍 

The only problem with this now I think is the fact that Realm Objective-C tests will fail to build with the Swift 3 preview because `Realm/Tests/Swift3.0-dev` doesn't exist. Might not be worth doing at all though until Swift 3 stabilizes a bit more.

tgoyne:
A stub version of the swift 3 obj-c tests with just empty files sounds reasonable to me for now.

austinzheng:
I updated the branch with another commit to get things working for both `Realm` and `RealmSwift` targets.

austinzheng:
I'm still getting sporadic compiler segfaults when I try running the tests for the `Realm` target:

```
While emitting IR SIL function @_TFE5TestsCSo8RLMArray12makeIteratorfT_CS_11RLMIterator for 'makeIterator' at /Users/austinzheng/Developer/realm-swift3-cocoa/Realm/Swift/RLMSupport.swift:49:16
```

austinzheng:
@tgoyne @jpsim @bdash Final ping; I've commented out the `Sequence` extensions on the Objective-C types per JP, and am ready to merge once I get an approval.

jpsim:
👍 this is good for now.

Maybe worth filing issues for completing work on `RLMSupport.swift` and porting Realm Objective-C Swift tests to Swift 3. Also filing a Swift bug for the Swift compiler segfault.

PR #3745, opened 2016-06-14
[Examples ios swift] Add CollectionView example app target

mrackwitz:
Adds a sample using a `UICollectionView`, which needs little different glue code to propagate changes from our fine-grained notifications to the collection view. This is mainly intended as a reference for our users.

/c @icanzilb @TimOliver

### DIFF
--- a/examples/ios/swift-2.2/CollectionView/CollectionViewController.swift
+++ b/examples/ios/swift-2.2/CollectionView/CollectionViewController.swift
@@ -1,0 +1,135 @@
+//
+//  CollectionViewController.swift
+//  CollectionView
+//
+//  Created by Marius Rackwitz on 14.6.16.
+//  Copyright © 2016 Realm. All rights reserved.
+//
+
+import UIKit
+import RealmSwift
+
+class DemoObject: Object {
+    dynamic var title = ""
+    dynamic var date = NSDate()
+}
+
+class Cell: UICollectionViewCell {
+    @IBOutlet var label: UILabel!
+    @IBOutlet var dateLabel: UILabel!
+
+    static var dateFormatter: NSDateFormatter = {
+        let dateFormatter = NSDateFormatter()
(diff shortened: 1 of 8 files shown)


mrackwitz:
CI failures are unrelated.

mrackwitz:
ping and /c @jpsim

Completely reviewing this PR with all the boilerplate stuff in here is difficult. I'd appreciate a brief look at [CollectionViewController.swift](https://github.com/realm/realm-cocoa/blob/mr/example/collectionview/examples/ios/swift-2.2/CollectionView/CollectionViewController.swift).

examples/ios/swift-2.2/CollectionView/CollectionViewController.swift
```diff
@@ -0,0 +1,135 @@
...
+        label.text = object.title
+        dateLabel.text = Cell.dateFormatter.stringFromDate(object.date)
+    }
+}
+
+class CollectionViewController: UICollectionViewController {
+    var notificationToken: NotificationToken? = nil
+
+    lazy var realm = try! Realm()
+    lazy var results: Results<DemoObject> = {
```
jpsim: why is the closure useful here?

jpsim: wouldn't you just write this?

```swift
lazy var results = self.realm.objects(DemoObject)
```

jpsim:
I don't get the motivation for this example. It seems to be exactly like the TableViewExample, except for the very few places where UICollectionView's API differs.

mrackwitz:
These very few differences are exactly the motivation. Giving those of our users, who are not familiar with the API to propagate batched updates to a `UICollectionView`, a copy-paste ready boilerplate on how to achieve that with Realm.

jpsim:
In my opinion, I don't think this provides enough differentiated value compared to our existing examples to include here. Closing.

PR #3750, opened 2016-06-15
Fix reading the shared schema from object instances

### LINKED ISSUE #3719: [Realm 1.0.0] Realm object becomes "NSKVONotifying_RLMStandalone_User" and is not treated as swift class --> crash
Hi team,

I have a class "User" inheriting from Realm object. Its name is normally "ProjectName.User", and it will be treated as Swift class, according to the following code:

```
@implementation RLMSwiftSupport
+ (BOOL)isSwiftClassName:(NSString *)className {
    return [className rangeOfString:@"."].location != NSNotFound;
}
```

However, when "self.dynamicType.sharedSchema()" function is called, it becomes "NSKVONotifying_RLMStandalone_User" and not be treated as Swift class anymore. When I debug, the "cls" in following code is NSKVONotifying_RLMStandalone_User. I do add some KVO observations to unmanaged Realm object and remove them when adding object to realm.

```
void RLMReplaceSharedSchemaMethod(Class accessorClass, RLMObjectSchema *schema) {
    Class metaClass = object_getClass(accessorClass);
    IMP imp = imp_implementationWithBlock(^(Class cls) {
        // This can be called on a subclass of the class that we overrode it on
        // if that class hasn't been initialized yet
        if (cls == accessorClass) {
            return schema;
        }
        return [RLMSchema sharedSchemaForClass:cls];
    });
    class_addMethod(metaClass, @selector(sharedSchema), imp, "@@:");
}
```

Since it is not Swift class (does not contain character "." in name), Realm cannot "Mirror" its instance to get the properties list --> crash.

I do not have this issue in Realm 0.97 and before. Since upgrading to 1.0.0, I had that problem.

Thanks.

tgoyne:
The overridden sharedSchema class method only worked properly when called
directly on the class it's overridden on, but actual object instances are
always a generated subclass of that class. This roundaboutly ended up working
by coincidence in most cases due to it finding the correct shared schema via
the lookup by name, but for KVO-generated subclasses with a different name it'd
instead end up trying to generate a new RLMObjectSchema for the runtime-created
subclass. For obj-c classes this worked and was just pointless extra work, but
for Swift classes it crashed.

Fixes #3719.


bdash:
👍

austinzheng:
Looks good.

Is there any danger of the `while` loop not terminating if the method is improperly invoked?

tgoyne:
It'll crash if it hits a root class without encountering the desired class, which I'd consider acceptable behavior for someone grabbing the implementation of a method and invoking it on an unrelated class.

PR #3751, opened 2016-06-15
[WIP] no longer run reset-simulators.sh

jpsim:
Opening this PR to investigate how CI breaks (if at all) if we don't reset the simulators between jobs now that we have Xcode 7 & 8 on CI machines.


jpsim:
I'm done with this testing for now, I'll reopen another WIP PR if I need to test this again in the future.

PR #3760, opened 2016-06-16
Allow throws within write block

JadenGeller:
Currently, the user cannot call throwing functions within a `write` block unless they handle recovery _within_ the block. If the user wishes to propagate the error outside the block, this leads to very convoluted code.

```swift
var error: ErrorType?
try Realm().write {
  do {
    try createInterdimensionalPortal()
  } catch let ohNoes {
    error = ohNoes
  }
}
if let error = error { throw ohNoes }
```

Even worse, this will not cancel the write transaction, so any changes made before the `throw` will still be committed to the database. This pull request annotates the block parameter of `write` as `throws` so that it can optionally throw an exception, propagating that exception outside of the write block and canceling the transaction. Now, the above code can be properly written:
```swift
try Reaml().write {
  try createInterdimensionalPortal()
}
```

### DIFF
--- a/RealmSwift-swift2.2/Realm.swift
+++ b/RealmSwift-swift2.2/Realm.swift
@@ -98,6 +98,9 @@ public final class Realm {
     /**
      Performs actions contained within the given block inside a write transaction.
 
+     If the block throws an error, the transaction will be canceled, reverting any writes made in the block before
+     the error was thrown.
+
      Write transactions cannot be nested, and trying to execute a write transaction on a Realm which is already
      participating in a write transaction will throw an error. Calls to `write` from `Realm` instances in other threads
      will block until the current write transaction completes.
@@ -109,9 +112,17 @@ public final class Realm {
      - parameter block: The block containing actions to perform.
 
      - throws: An `NSError` if the transaction could not be completed successfully.
+               If `block` throws, the propagated `ErrorType`.
      */
-    public func write(@noescape block: (() -> Void)) throws {
-        try rlmRealm.transactionWithBlock(block)
+    public func write(@noescape block: (() throws -> Void)) throws {
+        beginWrite()
+        do {
+            try block()
+        } catch let error {
+            if inWriteTransaction { cancelWrite() }
+            throw error
+        }
+        if inWriteTransaction { try commitWrite() }
     }
 
     /**
--- a/RealmSwift-swift2.2/Tests/RealmTests.swift
+++ b/RealmSwift-swift2.2/Tests/RealmTests.swift
@@ -25,6 +25,10 @@ import XCTest
 import Foundation
 
 class RealmTests: TestCase {
+    enum TestError: ErrorType {
+        case intentional
+    }
+
     func testFileURL() {
         XCTAssertEqual(try! Realm(fileURL: testRealmURL()).configuration.fileURL,
                        testRealmURL())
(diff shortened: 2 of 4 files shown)


RealmSwift-swift2.2/Tests/RealmTests.swift
```diff
@@ -255,6 +255,29 @@ class RealmTests: TestCase {
...
+    func testThrowsWrite() {
+        enum TestError: ErrorType { case Exception }
+        
+        do {
+            try Realm().write {
+                throw TestError.Exception
+            }
+            XCTFail("Expected exception thrown in write block to be rethrown.")
+        }
+        catch TestError.Exception { }
```
JadenGeller: I realize we have an `assertThrows` function, but it wasn't working for me. Does using `try!` turn a Swift exception into an Objective-C exception? Should I expect this to work on Swift error types, or only `NSError`?

austinzheng: `assertThrows` only checks for Objective-C exceptions. I don't think there's any correspondence between `try` and Obj-C exceptions, so you may need to write your own Swift error-specific test harness.

tgoyne: `assertFails()` checks for a Swift exception.

RealmSwift-swift2.2/Realm.swift
```diff
@@ -96,7 +96,8 @@ public final class Realm {
     // MARK: Transactions
 
     /**
-     Performs actions contained within the given block inside a write transaction.
+     Performs actions contained within the given block inside a write transaction. If the block throws an exception, the
```
tgoyne: First line of the docs should always be a (brief) single sentence description of the function, and all addition details go in their own paragraphs.

RealmSwift-swift2.2/Tests/RealmTests.swift
```diff
@@ -255,6 +255,29 @@ class RealmTests: TestCase {
         }
         XCTAssertEqual(try! Realm().objects(SwiftStringObject.self).count, 0)
     }
+    
+    func testThrowsWrite() {
+        enum TestError: ErrorType { case Exception }
```
JadenGeller: Is it reasonable to create a specific test error type, or would `NSError` with a test domain be preferred? Right now, this causes a listing error due to nested types.

JadenGeller: Apparently `assertFails` requires that I throw a `Realm.Error` rather than any `ErrorType`. I'm not sure the best way to test this. If I were to throw an existing `Error` case, I couldn't be sure that some unrelated failure didn't occur.

RealmSwift-swift2.2/Realm.swift
JadenGeller: Okay, I'll add a paragraph break before the second sentence.

RealmSwift-swift2.2/Tests/RealmTests.swift
JadenGeller: I threw `Realm.Error.Fail` in the test, but I'm not too happy with that as the test could erroneously succeed. Any suggestions @austinzheng, @tgoyne?

JadenGeller: Sorry, this is the up-to-date line: https://github.com/realm/realm-cocoa/blob/78c8690be82048fd21858cc3c8128efceec12360/RealmSwift-swift2.2/Tests/RealmTests.swift#L262

RealmSwift-swift2.2/Realm.swift
```diff
@@ -98,6 +98,9 @@ public final class Realm {
     /**
      Performs actions contained within the given block inside a write transaction.
 
+     If the block throws an exception, the transaction will be canceled, reverting any writes made in the block before
```
jpsim: "If the block throws an error". Apple never refers to Swift errors as "exceptions", but rather "Swift errors"

PR #3762, opened 2016-06-16
Use a separate index for column -> property lookups rather than sorting the properties

tgoyne:
This makes things a bit less weird when constructing a schema manually for the
dynamic API (since the properties array visible to you will actually match what
you passed in), is required to support having the array of properties be a
subset of the table columns, and simplifies handling columns being rearranged a bit.

Also a little faster since it's O(N) rather than O(N log N), but if you have
enough properties for it to be even remotely noticeable something has gone very
wrong.

@bdash

### DIFF
--- a/Realm/RLMObjectSchema.mm
+++ b/Realm/RLMObjectSchema.mm
@@ -41,8 +41,8 @@ using namespace realm;
 @implementation RLMObjectSchema {
     // table accessor optimization
     realm::TableRef _table;
-    NSArray *_propertiesInDeclaredOrder;
     NSArray *_swiftGenericProperties;
+    std::vector<RLMProperty *> _propertiesInTableColOrder;
 }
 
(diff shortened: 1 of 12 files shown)


tgoyne:
ping

Realm/RLMObjectSchema.mm
```diff
@@ -41,8 +41,8 @@ @interface RLMObjectSchema ()
 @implementation RLMObjectSchema {
     // table accessor optimization
     realm::TableRef _table;
-    NSArray *_propertiesInDeclaredOrder;
     NSArray *_swiftGenericProperties;
+    std::vector<RLMProperty *> _propertiesInTableColOrder;
```
bdash: I'd suggest either `_propertiesInTableColumnOrder` or `_propertiesInTableOrder`. Seems a bit pointless to abbreviate to save 3 characters when the variable name is already > 25 characters long.

bdash:
👍

PR #3766, opened 2016-06-17
Fix warnings when building with Xcode 8

tgoyne:
Apply the build settings changes Xcode 8 wants and fix a few trivial nullability issues.


austinzheng:
👍

jpsim:
👍

jpsim:
Actually nvm, there's a problem with the `-Owholemodule` flag not being recognized.

I _think_ it's `SWIFT_WHOLE_MODULE_OPTIMIZATION = YES;` for Xcode 7 and `SWIFT_OPTIMIZATION_LEVEL = "-Owholemodule";` for Xcode 8.

bdash:
👍

tgoyne:
This now actually works in both Xcode 7 and 8, but I don't understand what Xcode 8 is doing. If I set `SWIFT_OPTIMIZATION_LEVEL = "-Owholemodule";` in the xcconfig it over-escapes the value and passes `"-Owholemodule"` to the compiler, `SWIFT_OPTIMIZATION_LEVEL = -Owholemodule;` does nothing (the command line gets just -O and not `-O -whole-module`), and with the weird hack I went with the command line is correct.

bdash:
That sounds worthy of a Radar!

tgoyne:
I tried to put together a repro case for a radar and it turns out it's just a bug in Xcode's handling of xcconfig files changing while the project is open. After clearing DerivedData and closing/reopening the project it works correctly, so presumably the presence of a user-defined variable in the value was just making it so that Xcode didn't cache the value when it shouldn't.

PR #3767, opened 2016-06-17
Improve error message when using NSNumber in Swift

### LINKED ISSUE #3705: Improve error message when using NSNumber properties in Swift
For iOS, using Realm1.0.0, Swift and Xcode 7.3.1. Realm frameworks manually added from the `realm-swift-1.0.0` zip file.

```
        do {
            realm = try Realm()
        } catch let error as NSError {
            print("\(error)")
        }
```

lldb output attached.
[realm-exc_bad_access.txt](https://github.com/realm/realm-cocoa/files/298656/realm-exc_bad_access.txt)

JadenGeller:
Fixes #3705

### DIFF
--- a/Realm/RLMProperty.mm
+++ b/Realm/RLMProperty.mm
@@ -352,6 +352,9 @@ static bool rawTypeIsComputedProperty(NSString *rawType) {
     }
     else if ([_objcRawType isEqualToString:@"@\"NSNumber\""]) {
         const char *numberType = [propertyValue objCType];
+        if (!numberType) {
+            @throw RLMException(@"Can't persist NSNumber without default value: use a Swift-native number type or provide a default value.");
+        }
         switch (*numberType) {
             case 'i':
             case 'l':
--- a/RealmSwift/Tests/TestUtils.mm
+++ b/RealmSwift/Tests/TestUtils.mm
@@ -30,7 +30,7 @@ static void initializeSharedSchema() {
     [RLMSchema sharedSchema];
 }
 
-void RLMAssertThrows(XCTestCase *self, dispatch_block_t block, NSString *name, NSString *message, NSString *fileName, NSUInteger lineNumber) {
+void RLMAssertThrowsWithName(XCTestCase *self, dispatch_block_t block, NSString *name, NSString *message, NSString *fileName, NSUInteger lineNumber) {
     BOOL didThrow = NO;
     @try {
         block();
@@ -40,6 +40,28 @@ void RLMAssertThrows(XCTestCase *self, dispatch_block_t block, NSString *name, N
         if (![name isEqualToString:e.name]) {
             NSString *msg = [NSString stringWithFormat:@"The given expression threw an exception named '%@', but expected '%@'",
                              e.name, name];
+            [self recordFailureWithDescription:msg inFile:fileName atLine:lineNumber expected:NO];
+        }
+    }
+    if (!didThrow) {
+        NSString *prefix = @"The given expression failed to throw an exception";
+        message = message ? [NSString stringWithFormat:@"%@ (%@)",  prefix, message] : prefix;
+        [self recordFailureWithDescription:message inFile:fileName atLine:lineNumber expected:NO];
+    }
+}
+
+void RLMAssertThrowsWithReasonMatching(XCTestCase *self, dispatch_block_t block, NSString *regexString, NSString *message, NSString *fileName, NSUInteger lineNumber) {
+    BOOL didThrow = NO;
+    @try {
+        block();
+    }
+    @catch (NSException *e) {
+        didThrow = YES;
+        NSString *reason = e.reason;
+        NSRegularExpression *regex = [NSRegularExpression regularExpressionWithPattern:regexString options:(NSRegularExpressionOptions)0 error:nil];
+        if ([regex numberOfMatchesInString:reason options:(NSMatchingOptions)0 range:NSMakeRange(0, reason.length)] == 0) {
+            NSString *msg = [NSString stringWithFormat:@"The given expression threw an exception with reason '%@', but expected to match '%@'",
+                             reason, regexString];
             [self recordFailureWithDescription:msg inFile:fileName atLine:lineNumber expected:NO];
         }
     }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@ x.x.x Release notes (yyyy-MM-dd)
 
 ### Enhancements
 
-* None.
+* Improve error message when using NSNumber incorrectly in Swift models.
 
 ### Bugfixes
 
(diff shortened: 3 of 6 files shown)


JadenGeller:
//cc @bdash

jpsim:
@JadenGeller could you please add a test for this, also verifying the message string, in `ObjectSchemaInitializationTests.testInvalidObjects`?

Realm/RLMProperty.mm
```diff
@@ -343,6 +343,9 @@ - (instancetype)initSwiftPropertyWithName:(NSString *)name
     }
     else if ([_objcRawType isEqualToString:@"@\"NSNumber\""]) {
         const char *numberType = [propertyValue objCType];
+        if (!numberType) {
+            @throw RLMException(@"Can't persist NSNumber directly: use a Swift-native number type.");
```
mrackwitz: > or provide a default value, which allows to infer the type.

JadenGeller: Are you suggesting we allow `NSNumber` to be used in Swift? That seems like it would provide confusing semantics since it user code doesn't annotate what type of number it stores.

mrackwitz: That's what we already do and some users rely on it, even though it's not directly documented. But you can declare your Realm models in Swift while using Realm Objective-C.

JadenGeller: Oh, I see. What type do you suggest it default to? I'd presume a floating-point representation would be best so no unexpected truncations will occur.

Is there any way we can provide proper error messages to Realm Swift users while maintaining backwards compatibility for Realm Objective-C users with Swift models?

tgoyne: It shouldn't default to anything. If the user provides a default value we can get what type of number was used to construct the NSNumber (that's what the line above is doing).

JadenGeller: The Realm Objective-C docs don't mention support for storing non-nullable `NSNumber`s—the primitive type should be used—so we could at least throw an exception for the non-`Optional` case. This would not have helped #3705 though.

mrackwitz: Exactly. :+1:

>Is there any way we can provide proper error messages to Realm Swift users while maintaining backwards compatibility for Realm Objective-C users with Swift models?

No, I don't see how there would be a way as `RLMProperty` is agnostic about the class it belongs to. The only way would be making the initializer failable or catching the exception from the object schema creation and then rethrowing it, which I'd rather avoid just for that purpose.

JadenGeller: Ooohhh, I misunderstood the @mrackwitz comment. I had thought it was a suggested alternative implementation, but I now understand that it was an addition to the exception message. Sorry for the confusion!

jpsim:
when making updates to the Swift side of things, please remember to mirror those changes in all Swift versions we're targeting. In this case, Swift 3

JadenGeller:
@bdash @jpsim @mrackwitz I've made the suggested changes. I feel that the exception message may be a bit confusing, so please provide feedback if you agree.

> Can't persist NSNumber directly: use a Swift-native number type or provide a default value, which allows to infer the type

JadenGeller:
@jpsim How should I go about "verifying the message string"? The `named:` parameter of `assertThrows` checks an exception name, not message, and the function `assertThrowsWithReasonMatching` is only found on `RLMTestCase`, not `TestCase`. Is this a mistake?

jpsim:
Exposing an equivalent to `assertThrowsWithReasonMatching` for Swift tests seems sensible to me.

RealmSwift-swift2.2/Tests/TestUtils.mm
```diff
@@ -53,3 +53,25 @@ void RLMAssertThrows(XCTestCase *self, dispatch_block_t block, NSString *name, N
 bool RLMHasCachedRealmForPath(NSString *path) {
     return RLMGetAnyCachedRealmForPath(path.UTF8String);
 }
+
+void RLMAssertThrowsWithReasonMatchingSwift(XCTestCase *self, dispatch_block_t block, NSString *regexString, NSString *message, NSString *fileName, NSUInteger lineNumber) {
```
JadenGeller: This code is duplicated from from the Realm Objective-C `TestUtils.mm` file. Would it be preferable to include that file in both test targets? We already have precedent of duplicating `RLMAssertThrows` across multiple files, but it definitely is gross having to maintain both separately.

tgoyne: Pulling the helper functions for both frameworks into a file shared between them sounds like a good idea.

JadenGeller: Will do. Where should this file live? Should I create a shared folder?

Realm/RLMProperty.mm
```diff
@@ -343,6 +343,9 @@ - (instancetype)initSwiftPropertyWithName:(NSString *)name
     }
     else if ([_objcRawType isEqualToString:@"@\"NSNumber\""]) {
         const char *numberType = [propertyValue objCType];
+        if (!numberType) {
+            @throw RLMException(@"Can't persist NSNumber directly: use a Swift-native number type or provide a default value, which allows to infer the type");
```
bdash: "allows to infer the type" reads oddly to me.

JadenGeller: I agree. This phrasing was suggested to me by https://github.com/realm/realm-cocoa/pull/3767#discussion_r67742399. I'd be okay to drop off "which allows to infer the type" entirely. Thoughts?

JadenGeller: (I honestly still find it a bit surprising that users can use `NSNumber` in Swift if a default value is provided. Is this intentional, or just a side effect of how our model parsing works? As a user, I wouldn't expect a default value of `1` would limit me to integers and `1.0` to floats.)

tgoyne: It's intentional. There aren't really any good ways to let the user specify the type (can't get the generic type at runtime, can't use the protocol hack we use in obj-c, and using a subclass means that you have to have casts at runtime to assign NSNumbers to the property (and it disables the use of tagged pointers)).

JadenGeller: As a user though, I would not expect changing my default value from this
```swift
class Foo: RLMObject {
    var bar: NSNumber = 0.5
}
```
to this
```swift
class Foo: RLMObject {
    var bar: NSNumber = 1
}
```
to require a schema migration. That would be extremely surprising.

JadenGeller: @tgoyne I think that if a user is to use `NSNumber` in Swift, they should have to override some method to indicate the type of that number.

JadenGeller:
@bdash @jpsim I made the changes you requested. I'll merge this later today if there are no objections.

RealmSwift/Tests/TestUtils.mm
```diff
@@ -50,6 +50,28 @@ void RLMAssertThrows(XCTestCase *self, dispatch_block_t block, NSString *name, N
     }
 }
 
+void RLMAssertThrowsWithReasonMatching(XCTestCase *self, dispatch_block_t block, NSString *regexString, NSString *message, NSString *fileName, NSUInteger lineNumber) {
```
JadenGeller: Note that this is copied from Realm Objective-C's Swift helpers. I decided to duplicate the code (rather than merging the files) since the other functions in this file are not applicable to Realm Objective-C's tests.

tgoyne:
👍 

Needs a changelog entry.

Issue #3768, opened 2016-06-18
Implicitly unwrapped relationships using Swift 3


Anviking:
## Goals
Use implicitly unwrapped optional relationship properties in a Swift Realm model using Xcode 8 / Swift 3.

## Expected Results
Should not crash on launch. Should behave just like an optional property, but implicitly unwrapped. This worked in Swift 2.

## Actual Results
Throws in RLMObjectSchema.mm:317 on application launch
```objc
for (RLMProperty *property in propArray) {
        if (!property.optional && property.type == RLMPropertyTypeObject) { // remove if/when core supports required link columns
            @throw RLMException(@"The `%@.%@` property must be marked as being optional.", [objectClass className], property.name);
        }
    }
```

## Steps to Reproduce / Code Sample

```swift
class User: Object {
    var name: String = ""
}

class Dog: Object {
    var owner: User!
}

// Somewhere
let realm = try! Realm()
```

## Version of Realm and Tooling

Realm version: ebf32fe57d8d02aaea5de1ed141909b8362cb673

Xcode version: 8 beta 1

iOS/OSX version: 10

Dependency manager + version: N/A

PR #3769, opened 2016-06-18
Check for implicitly unwrapped relationships in `getOptionalProperties`

Anviking:
Quick, potential fix for #3768

Playing around in a playground I found that `childMirror.displayStyle` for a `T!` (where `T: Object`) is `.enum` while `childMirror.displayStyle` for `T?` is `.optional` using Swift 3.0. Haven't explicitly tested on any previous Swift version, but I assume this is what's causing the problem.

Also haven't verified that the specific implicitly unwrapped properties in my app works, but it does at least launch now.

### DIFF
--- a/RealmSwift/Tests/ObjectSchemaInitializationTests.swift
+++ b/RealmSwift/Tests/ObjectSchemaInitializationTests.swift
@@ -182,6 +182,15 @@ class ObjectSchemaInitializationTests: TestCase {
         let types = Set(schema.properties.map { $0.type })
         XCTAssertEqual(types, Set([.string, .string, .data, .date, .object, .int, .float, .double, .bool]))
     }
+    
+    func testImplicitlyUnwrappedOptionalsAreParsedAsOptionals() {
+        let schema = SwiftImplicitlyUnwrappedOptionalObject().objectSchema
+        XCTAssertTrue(schema["optObjectCol"]!.isOptional)
+        XCTAssertTrue(schema["optNSStringCol"]!.isOptional)
+        XCTAssertTrue(schema["optStringCol"]!.isOptional)
+        XCTAssertTrue(schema["optBinaryCol"]!.isOptional)
+        XCTAssertTrue(schema["optDateCol"]!.isOptional)
+    }
 
     func testNonRealmOptionalTypesDeclaredAsRealmOptional() {
         assertThrows(RLMObjectSchema(forObjectClass: SwiftObjectWithNonRealmOptionalType.self))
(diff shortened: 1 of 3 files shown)


jpsim:
Hi @Anviking, thanks for doing this (and accepting our CLA, I just verified that). This would bring us back to the behavior in Swift 2.2, which I'm in favor of. Realm can never guarantee that an object property will _always_ be fulfilled, but if the user (e.g. you) explicitly ignores that, we should continue to support that behavior IMHO.

We also had unit tests that verified this behavior in Swift 2.2 ([source](https://github.com/realm/realm-cocoa/blob/9c978d96fc3e6cbdf446a0e2b28591434e52daa3/RealmSwift-swift2.2/Tests/ObjectSchemaInitializationTests.swift#L190-L197)). Could you also re-introduce these in the Swift 3 tests please?

Anviking:
@jpsim done!

Anviking:
(And rebased against master yesterday)

RealmSwift/Tests/ObjectSchemaInitializationTests.swift
```diff
@@ -182,6 +182,17 @@ class ObjectSchemaInitializationTests: TestCase {
         let types = Set(schema.properties.map { $0.type })
         XCTAssertEqual(types, Set([.string, .string, .data, .date, .object, .int, .float, .double, .bool]))
     }
+    
+    func testImplicitlyUnwrappedOptionalProperties() {
```
jpsim: Is there a reason the Swift 2.2 test wasn't sufficient? It's important to keep differences between Swift version implementations to a minimum to facilitate the transition.

```swift
func testImplicitlyUnwrappedOptionalsAreParsedAsOptionals() {
    let schema = SwiftImplicitlyUnwrappedOptionalObject().objectSchema
    XCTAssertTrue(schema["optObjectCol"]!.optional)
    XCTAssertTrue(schema["optNSStringCol"]!.optional)
    XCTAssertTrue(schema["optStringCol"]!.optional)
    XCTAssertTrue(schema["optBinaryCol"]!.optional)
    XCTAssertTrue(schema["optDateCol"]!.optional)
}
```

jpsim:
Thanks for updating this @Anviking! Just had one last comment about the unit test.

RealmSwift/Tests/ObjectSchemaInitializationTests.swift
Anviking: Ok! Not at all, just falsely believed style in `testOptionalProperties` had been introduced after this PR when I rebased.

Anviking: Fixed and squashed now!

jpsim:
Looks good. Thanks for your contribution @Anviking!

PR #3771, opened 2016-06-19
Combine files for all Swift versions

### LINKED ISSUE #3759: Merge Realm Swift source files
We currently support multiple Swift versions by having a copy of the root `RealmSwift` directory for each supported Swift version (e.g. `RealmSwift-swift3.0`) and at build time, run an Xcode script build phase to detect the Swift version in use for that project and establish a symbolic link from `RealmSwift` to the version-appropriate directory. This is _very_ hacky and now that Swift supports [compile-time version checks](https://github.com/apple/swift-evolution/blob/master/proposals/0020-if-swift-version.md), we have a "better way".

Issues with the current approach:

* We flatten the directories when installing via CocoaPods based on the `xcode-select`'d Xcode version, so changing Xcode versions after installing doesn't update the Swift version used.
* Xcode doesn't handle symlinks well: compiler diagnostics (including SwiftLint violations) don't render in the source files because the real path is different than what Xcode points to.
* It adds significant complexity to the project & build system:
  * Xcode build phases in a few places
  * CocoaPods link flattening step
  * `build.sh set-swift-version`
  * `scripts/swift-version.sh`
  * modifying git indexes to avoid committing symlink changes
* Minor point: not scalable to non-Xcode build systems in the future (e.g. SwiftPM)

---

Solution: concatenate the Swift 2.2 & Swift 3.0 files with a single giant `#if swift(<3.0) ... #else ... #endif` in each file. For some files it may be appropriate to share more code.

---

Potential reasons not to do this:

* We should investigate if combining the files causes Xcode to be (more) sluggish
* ???

jpsim:
Closes #3759.

There's room for improvement to minimize the amount of duplicated code, but at least this addresses the issues described in #3759.

Also resolves CocoaPods support for Swift 3 (see [tweet](https://twitter.com/leoderbois/status/744161106059493377)), since we no longer run `set-swift-version` in `build.sh cocoapods-setup`.

/cc @austinzheng @tgoyne


tgoyne:
I suspect the cocoapods failure is because we were relying on the symlinnk functionality for the os x core library, which should be a trivial fix.

Otherwise looks good. w.r.t. sharing code between the versions I'd prefer to at least start with only having a single `if swift(>=3.0)` check per file and only share things that can easily fit with that constraint (i.e. what's in Util.swift is fine, but having a bunch of `#if`s gets very hard to work with).

austinzheng:
+1 to what Thomas said. Let's keep the Swift version checks as simple as possible.

austinzheng:
I don't want to rush anyone, but I think it would be great if we could expedite this particular PR. Any nontrivial changes to the Swift files depend on this going in first.

jpsim:
CI failures unrelated. Merging. I addressed @tgoyne's comment about the CocoaPods core symlink failure in 005ad61.

PR #3776, opened 2016-06-20
Fix handling of Swift 3 properties which are renamed when bridged to obj-c

### LINKED ISSUE #3773: Properties on the form isFoo are inaccessible using Swift 3
Properties on the form `isFoo` are renamed to `foo` when imported to ObjC from model in Swift 3 / Xcode 8. This causes a crash upon access.

```swift
class Dog: Object {
    var isFoo = false
}

// Somewhere
let realm = try! Realm()
try! realm.write {
        let a = Dog()
        a.isFoo = false
        realm.add(a)
}
```
## Expected Results
Properties declared in Swift  should work with realm regardless of their name.

## Actual Results
```
'RLMException', reason: 'No value or default value specified for property 'foo' in 'Dog''
	*** First throw call stack:
	(
		0   CoreFoundation                      0x0000000100d94deb __exceptionPreprocess + 171
		1   libobjc.A.dylib                     0x000000010354331e objc_exception_throw + 48
		2   Realm                               0x000000010074d6ac RLMAddObjectToRealm + 2220
		3   RealmSwift                          0x0000000100be0e57 
```

In `RLMAddObjectToRealm`:
prop.getterSel is "foo"
`respondsToSelector` fails on line 205 fails since `Dog.isFoo` is now (assume this is new) imported to ObjC as
```objc
@property (nonatomic, getter=isFoo) BOOL foo;
``` 

Realm version: ebf32fe57d8d02aaea5de1ed141909b8362cb673

Xcode version: 8 beta 1

iOS/OSX version: iOS 10

tgoyne:
Fixes #3773.

### DIFF
--- a/Realm/RLMProperty.mm
+++ b/Realm/RLMProperty.mm
@@ -279,7 +279,7 @@ static bool rawTypeIsComputedProperty(NSString *rawType) {
     }
 }
 
-- (bool)parseObjcProperty:(objc_property_t)property {
+- (bool)parseObjcProperty:(objc_property_t)property isSwift:(bool)isSwift {
     unsigned int count;
     objc_property_attribute_t *attrs = property_copyAttributeList(property, &count);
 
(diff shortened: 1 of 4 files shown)


jpsim:
👍 on implementation. Could use a changelog entry and we really need to come up with a strategy for keeping multiple Swift versions up to date (tests were only added for Swift 2.2).

tgoyne:
There are both Swift 2.2 and Swift 3 tests here.

I don't think changelog entries for bugfixes in our swift 3 support make sense until we actually have a release with swift 3.

PR #3782, opened 2016-06-21
[ISSUE_TEMPLATE] Make use of details html element

mrackwitz:
I [noticed](https://gist.github.com/orta/057eab0832773aa299a6ebd39187ab06) that the `<details>` HTML tag is supported by GitHub-flavoured Markdown and thought that might be a neat way to get the intro of our issue template out of the way without bothering users creating issues to remove that part. In addition to that we can use it to collapse longer sections. 😳

/c @jpsim @bmunkholm

### DIFF
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -1,17 +1,17 @@
-> We LOVE to help with any issues or bugs you have!
->
-> **Questions**: If you have questions about how to use Realm, ask on 
-> [StackOverflow](http://stackoverflow.com/questions/ask?tags=realm).
-> We monitor the `realm` tag.
->
-> **Feature Request**: Just fill in the first two sections below.
->
-> **Bugs**: To help you as fast as possible with an issue please describe your issue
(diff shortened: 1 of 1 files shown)


jpsim:
👍

mrackwitz:
It turns out that there seems to be an issue with current Safari betas not predictable showing and hiding the details as seen [here](https://github.com/CocoaPods/CocoaPods/pull/5575). As that could potentially affect a fair chunk of users visiting this repo, I'd leaning towards not merging that until this is resolved.

bdash:
The issue template should only be things that we want the user to read. Having content in it that's hidden by default suggests the content isn't important enough to belong in the issue template in the first place. I'm also not clear why it's useful to have collapsable sections in the issue template that default to expanded. Can you elaborate on why you feel this is beneficial?

jpsim:
@mrackwitz can we just close this if it causes issues for Safari and these changes don't have a substantial effect on our bug reports?

mrackwitz:
Yes, let's close this.

PR #3797, opened 2016-06-24
Remove Xcode 8 code signing requirement

austinzheng:
@jpsim @tgoyne @bdash @mrackwitz 

This removes the requirement for Xcode 8 users to configure code signing when building Realm for device targets.


bdash:
What impact does this have on building for iOS devices from within Xcode 7?

austinzheng:
There shouldn't be any. Building to both my phone and the generic iOS device target seem to work fine. Do you know if someone else might have more experience with build configurations?

bdash:
I ask because presumably `CODE_SIGN_IDENTITY` was being set for a reason. Maybe it is just something that Xcode set unnecessarily, or was only useful in older versions of Xcode?

Configuration/Base.xcconfig
```diff
@@ -38,7 +38,8 @@ WARNING_CFLAGS = -Wmismatched-tags -Wunused-private-field;
 
 OTHER_CPLUSPLUSFLAGS = $(OTHER_CPLUSPLUSFLAGS) -isystem core/include -IRealm/ObjectStore;
 
-CODE_SIGN_IDENTITY[sdk=iphone*] = iPhone Developer;
+CODE_SIGNING_REQUIRED = NO;
+CODE_SIGN_IDENTITY[sdk=*] =;
```
bdash: Would removing `CODE_SIGN_IDENTITY` have the same effect?

austinzheng: if I remove `CODE_SIGN_IDENTITY` completely, iPhone and OS X build fine but watch and TV builds fail. I assume this is an artifact of them hacking in support for the latter two platforms?

austinzheng: Per @mrackwitz suggestion, `CODE_SIGN_IDENTITY =;` also seems to work.

bdash:
I looked at this with Austin and we found that this prevented running tests on iOS devices. After some experimentation we came up with a different approach, which I've taken in #3801.

PR #3801, opened 2016-06-24
Fix code signing with Xcode 8

bdash:
Automatic provisioning is enabled for the targets that require signing (frameworks, tests, and test host). Signing is explicitly disabled on OS X as it is not necessary for our purposes.
    
Both of Xcode 8's approaches to provisioning appear to require hard-coding a development team identifier in the project file. This will likely prevent people outside of Realm from building from source using Xcode 8 without first changing the team used for code signing for each target via Xcode's target editor. Hopefully this situation is improved before Xcode 8 GMs.
    
These changes have no effect on Xcode 7's behavior.


austinzheng:
👍 , per our conversation.

jpsim:
Do we know how CocoaPods/Carthage are dealing with this?

bdash:
CocoaPods is dealing with this by disabling signing. That works for their purposes as they're only concerned with building frameworks, and they can defer the signing of those frameworks to when they're copied into the user's app bundle.

PR #3804, opened 2016-06-27
Delete a bunch of dead code in RLMObjectStore

tgoyne:
`validateNested` was always false, so everything related to that code path was unreachable.


jpsim:
👍

bdash:
👍

PR #3809, opened 2016-06-29
Upgrade to core 1.3.0

### LINKED ISSUE #3793: Terminating app due to uncaught exception 'RLMException', reason: 'flock() failed: Interrupted system call
## Goals

What do you want to achieve?

A normal user flow

## Expected Results

No crash

## Actual Results

App crash for a reason that a can't find any info about

## Steps to Reproduce

I don't know what triggered it, it seams it happend during a write transaction, but it doesn't happen consistently.

## Code Sample

    JVTRoom *room = [self getRoomFromId:roomId];

    if (room) {

        room.isPresentedRoom = presented;

    }

Realm version: ?

Realm ( 1.0.1 )

Xcode version: ?

7.3.1

iOS/OSX version: ?

9.3.2

Cocoapods 1.0.0

tgoyne:
Fixes #3793.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,13 @@ x.x.x Release notes (yyyy-MM-dd)
 
 ### Enhancements
 
-* Swift: A `write` block may now `throw`, reverting any changes already made in the transaction.
+* Swift: A `write` block may now `throw`, reverting any changes already made in
+  the transaction.
+* Reduce address space used when committing write transactions.
+* Significantly reduce the download size of prebuilt binaries and slightly
+  reduce the final size contribution of Realm to applications.
+* Improve performance of accessing RLMArray properties and creating objects
+  with List properties.
 
 ### Bugfixes
 
@@ -18,6 +24,10 @@ x.x.x Release notes (yyyy-MM-dd)
 * Ensure that the initial call of a Results notification block is always passed
   .Initial even if there is a write transaction between when the notification
   is added and when the first notification is delivered.
+* Fix a crash when deleting all objects in a Realm while fast-enumerating query
+  results from that Realm.
+* Handle EINTR from flock() rather than crashing.
+* Fix incorrect behavior following a call to `[RLMRealm compact]`.
 
 1.0.1 Release notes (2016-06-12)
 =============================================================
--- a/build.sh
+++ b/build.sh
@@ -14,7 +14,7 @@ set -o pipefail
 set -e
 
 # You can override the version of the core library
-: ${REALM_CORE_VERSION:=1.1.1} # set to "current" to always use the current build
+: ${REALM_CORE_VERSION:=1.3.0} # set to "current" to always use the current build
 
 # You can override the xcmode used
 : ${XCMODE:=xcodebuild} # must be one of: xcodebuild (default), xcpretty, xctool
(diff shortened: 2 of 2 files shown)


austinzheng:
👍

CHANGELOG.md
```diff
@@ -7,11 +7,19 @@ x.x.x Release notes (yyyy-MM-dd)
 
 ### Enhancements
 
-* Swift: A `write` block may now `throw`, reverting any changes already made in the transaction.
+* Swift: A `write` block may now `throw`, reverting any changes already made in
+  the transaction.
+* Reduce address space used when committing write transactions.
+* Significantly reduce the download size of prebuilt binaries and slightly
+  reduce the final size contribution of Realm to applications.
```
bdash: What accounted for the reduction in size contributed to applications?

bdash:
👍

CHANGELOG.md
tgoyne: The changes to util::terminate slightly reduced the linked size (but not by a very exciting amount).

CHANGELOG.md
```diff
@@ -7,11 +7,19 @@ x.x.x Release notes (yyyy-MM-dd)
...
 
 ### Bugfixes
 
 * Fix a crash when reading the shared schema from an observed Swift object.
+* Fix a crash when deleting all objects in a Realm while fast-enumerating query
+  results from that Realm.
+* Handle EINTR from flock() rather than crashing.
+* Fix incorrect behavior following a call to `[RLMRealm compact]`.
```
bmunkholm: Could we describe how users could have experienced this?

tgoyne: It's not actually public (although a few users have used it anyway) and we didn't get any reports of problems with it, so a wall of text about a problem that probably never effected any of the people reading this doesn't seem useful.

jpsim:
👍

CHANGELOG.md
bmunkholm: +1

cmelchior:
We are investigating what looks like a Core sorting bug on Android right now, so you might want to hold merging this until we have verified it either way.

tgoyne:
Added a changelog entry for the sort order change. IMO from our perspective it's a bug fix as it brings us closer to the platform's sort order, and as we've never specified anything about our sort order it's only a breaking change in the sense that every functional change ever is.

cmelchior:
I am a bit puzzled how you can say that. You are replacing one broken implementation with another? I assume you had the same functionality as described here: https://realm.io/docs/java/latest/#sorting-and-querying-on-string. At least users had time to work around that. 

Changing that to something else that doesn't work might make sense, but right now the ramifications are not clear. Danish is a small language but what if e.g French or German is suddenly broken as well. As a minimum this should at least be a major version bump.

CHANGELOG.md
```diff
@@ -7,11 +7,21 @@ x.x.x Release notes (yyyy-MM-dd)
...
+* Align the sort order for punctuation with the order used by obj-c/swift sort
```
cmelchior: Only having this is misleading IMO, because other things have most likely changed as well.

tgoyne:
That section of the docs is either nonsense or incorrect, depending on how you interpret it. We only support sorting Latin Extended 2 in the sense that it produces a consistent result, but that's true of the entire unicode range. We certainly don't produce "correct" results when sorting them and never have, as that's impossible with a locale-insensitive sort.

JadenGeller:
Any news on this merge? It seems like the conversation died down.

tgoyne:
Core is going to switch back to the old sort order so we might as well just wait for that.

bdash:
👍

PR #3815, opened 2016-06-30
Fix `migrateRealm:` to follow Cocoa error conventions

JadenGeller:
Cocoa methods ought to use an error parameter rather than returning an error. This also allows Swift to import the function as throwing.

Fixes https://github.com/realm/realm-cocoa/issues/3816

### DIFF
--- a/Realm/RLMRealm.mm
+++ b/Realm/RLMRealm.mm
@@ -693,16 +693,21 @@ REALM_NOINLINE void RLMRealmTranslateException(NSError **error) {
     }
 }
 
-+ (NSError *)migrateRealm:(RLMRealmConfiguration *)configuration {
++ (BOOL)migrateRealm:(RLMRealmConfiguration *)configuration error:(NSError **)error {
+
     if (RLMGetAnyCachedRealmForPath(configuration.config.path)) {
         @throw RLMException(@"Cannot migrate Realms that are already open.");
     }
 
+    NSError *localError; // Prevents autorelease
+    BOOL success;
     @autoreleasepool {
-        NSError *error = nil;
-        [RLMRealm realmWithConfiguration:configuration error:&error];
-        return error;
+        success = [RLMRealm realmWithConfiguration:configuration error:&localError] != nil;
     }
+    if (!success && error) {
+        *error = localError; // Must set outside pool otherwise will free anyway
+    }
+    return success;
 }
 
 - (RLMObject *)createObject:(NSString *)className withValue:(id)value {
--- a/Realm/RLMRealm.mm
+++ b/Realm/RLMRealm.mm
@@ -693,8 +693,14 @@ REALM_NOINLINE void RLMRealmTranslateException(NSError **error) {
     }
 }
 
-+ (BOOL)migrateRealm:(RLMRealmConfiguration *)configuration error:(NSError **)error {
++ (nullable NSError *)migrateRealm:(RLMRealmConfiguration *)configuration {
+    // Preserves backwards compatibility
+    NSError *error;
+    [self performMigrationForConfiguration:configuration error:&error];
+    return error;
+}
 
++ (BOOL)performMigrationForConfiguration:(RLMRealmConfiguration *)configuration error:(NSError **)error {
     if (RLMGetAnyCachedRealmForPath(configuration.config.path)) {
         @throw RLMException(@"Cannot migrate Realms that are already open.");
     }
--- a/Realm/Tests/EncryptionTests.mm
+++ b/Realm/Tests/EncryptionTests.mm
@@ -207,11 +207,11 @@
         migrationRan = YES;
     };
 
-    XCTAssertFalse([RLMRealm migrateRealm:configuration error:nil]);
+    XCTAssertFalse([RLMRealm performMigrationForConfiguration:configuration error:nil]);
     XCTAssertFalse(migrationRan);
 
     configuration.encryptionKey = key;
-    XCTAssertTrue([RLMRealm migrateRealm:configuration error:nil]);
+    XCTAssertTrue([RLMRealm performMigrationForConfiguration:configuration error:nil]);
     XCTAssertTrue(migrationRan);
 }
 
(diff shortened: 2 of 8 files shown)


austinzheng:
Looks good to me! 🎉

tgoyne:
This is an API breaking change so it needs to wait for 2.0.

JadenGeller:
@tgoyne Is there a 2.0 branch that such changes are merged into, or do they just wait?

Realm/RLMRealm.mm
```diff
@@ -693,16 +693,17 @@ + (uint64_t)schemaVersionAtURL:(NSURL *)fileURL encryptionKey:(NSData *)key erro
...
     if (RLMGetAnyCachedRealmForPath(configuration.config.path)) {
         @throw RLMException(@"Cannot migrate Realms that are already open.");
     }
 
     @autoreleasepool {
-        NSError *error = nil;
-        [RLMRealm realmWithConfiguration:configuration error:&error];
-        return error;
+        [RLMRealm realmWithConfiguration:configuration error:error];
+        if (error) return NO;
```
JadenGeller: iirc, the value of an error parameter should not be checked unless the functions returns `nil` or `NO`

Realm/Tests/EncryptionTests.mm
```diff
@@ -207,11 +207,11 @@ - (void)testExplicitMigration {
         migrationRan = YES;
     };
 
-    XCTAssertNotNil([RLMRealm migrateRealm:configuration]);
+    XCTAssertTrue([RLMRealm migrateRealm:configuration error:nil]);
     XCTAssertFalse(migrationRan);
 
     configuration.encryptionKey = key;
-    XCTAssertNil([RLMRealm migrateRealm:configuration]);
+    XCTAssertTrue([RLMRealm migrateRealm:configuration error:nil]);
```
JadenGeller: This should be `XCTAssertFalse`

Realm/Tests/EncryptionTests.mm
```diff
@@ -207,11 +207,11 @@ - (void)testExplicitMigration {
         migrationRan = YES;
     };
 
-    XCTAssertNotNil([RLMRealm migrateRealm:configuration]);
+    XCTAssertTrue([RLMRealm migrateRealm:configuration error:nil]);
```
JadenGeller: This should be `XCTAssertFalse`

JadenGeller:
This fails tests right now with quite a few EXC_BAD_ACCESS exceptions. Do we guarantee to return `nil`/`NO` on error conditions? Simply setting the error pointer is not sufficient.

> When dealing with errors passed by reference, it’s important to test the return value of the method to see whether an error occurred, as shown above. Don’t just test to see whether the error pointer was set to point to an error. ([Apple](https://developer.apple.com/library/ios/documentation/Cocoa/Conceptual/ProgrammingWithObjectiveC/ErrorHandling/ErrorHandling.html))

Realm/RLMRealm.mm
```diff
@@ -693,15 +693,14 @@ + (uint64_t)schemaVersionAtURL:(NSURL *)fileURL encryptionKey:(NSData *)key erro
...
+
     if (RLMGetAnyCachedRealmForPath(configuration.config.path)) {
         @throw RLMException(@"Cannot migrate Realms that are already open.");
     }
 
     @autoreleasepool {
-        NSError *error = nil;
-        [RLMRealm realmWithConfiguration:configuration error:&error];
-        return error;
+        return [RLMRealm realmWithConfiguration:configuration error:error] == nil;
```
JadenGeller: This is the opposite of what you want, use `!=` :P

bdash:
Passing null for the error parameter of `+[RLMRealm realmWithConfiguration:error]` will result it in throwing an exception if it encounters an error (look for calls to `RLMSetErrorOrThrow`). That may be what's tripping you up in this change?

JadenGeller:
@bdash I actually think the `@autorelease` pool is causing an issue.

JadenGeller:
I could declare an additional reference to the error outside the autorelease pool, but I feel like this should be unnecessary. Is there a more elegant solution?

JadenGeller:
Seems like that's the way to do it ¯\_(ツ)_/¯ http://lists.apple.com/archives/objc-language/2011/Aug/msg00006.html

Realm/RLMRealm.mm
```diff
@@ -693,16 +693,21 @@ + (uint64_t)schemaVersionAtURL:(NSURL *)fileURL encryptionKey:(NSData *)key erro
...
+    NSError *localError; // Prevents autorelease
+    BOOL success;
     @autoreleasepool {
-        NSError *error = nil;
-        [RLMRealm realmWithConfiguration:configuration error:&error];
-        return error;
+        success = [RLMRealm realmWithConfiguration:configuration error:&localError] != nil;
+    }
+    if (!success && error) {
+        *error = localError; // Must set outside pool otherwise will free anyway
```
JadenGeller: Tricky autorelease pools 👹 should work now though

JadenGeller:
@tgoyne What if we additively introduce this method while deprecating the old one?

JadenGeller:
Never mind, keeping both in Swift makes them ambiguous under normal use. We could make the new one no longer a top level function, but I'd honestly rather axe the function if we have no real use case for it. (Although lots of our tests use it, apparently, so it's semi-useful…)

JadenGeller:
Would it be more fitting for to introduce `Realm.performMigration(with configuration:)` or `Migration.perform(for configuration)`? I personally don't love either.

JadenGeller:
I think it's best to stay consistent with the Objective-C API for now, moving it to static method on `Realm` with the same name it currently has. I think we might want to reconsider either removing or renaming this function with Realm 2.0 or Swift 3.0 eventually.

JadenGeller:
Turns out there's a bug with the way Objective-C throwing functions are imported into Swift when there's ambiguity, so I went ahead an renamed both [new] functions as to not break compatibility for users of Realm Objective-C in Swift.

JadenGeller:
Deprecated the old functions without changing their signatures and introduced `performMigration(for:) throws` in Swift and `performMigrationForConfiguration:error:` in Objective-C. Also removed the deprecated function in Swift 3.0.

tgoyne:
👍 

Needs a changelog entry.

Issue #3816, opened 2016-07-01
Correctly follow NSError Cocoa conventions


JadenGeller:
The Realm Objective-C API does not always follow proper NSError conventions and, as such, some functions are not properly imported into Swift as `throws`. Specifically, `NSError` should be an out parameter, not a return value. Additionally, a function that contains an `NSError` parameter should follow one of the supported return [conventions](https://github.com/apple/swift-clang/blob/stable/include/clang/Basic/AttrDocs.td#L1808).

The following functions have been identified to improperly follow the Cocoa conventions:
- `migrateRealm:`
  - This function returns an `NSError`, but should return `BOOL` and have an `NSError` out parameter.
- `schemaVersionAtURL:encryptionKey:error:`
  - This function is sensible. It returns a specific constant when an error occurs, and will otherwise return a different value. Unfortunately, the Swift importer doesn't support such behavior. We should file a radar. As a temporarily workaround, we can mark this method as `NS_REFINED_FOR_SWIFT` and reimplement it as throwing in Swift.

These changes will improve the experience of working with Realm Objective-C in the Swift programming language. Additionally, this will move us a step closer to replacing our separate Swift module with an Swift-annotated Objective-C module (if we decide to go down that path) (https://github.com/realm/realm-cocoa/issues/3790).

JadenGeller:
https://github.com/realm/realm-cocoa/pull/3815 fixes the conventions of the `migrateRealm:` function

JadenGeller:
Swift issue tracking [Default value annotations for Objective-C parameters](https://bugs.swift.org/browse/SR-1970) and tracking [Importing error convention for specific non-zero constant](https://bugs.swift.org/browse/SR-1971). If we feel these are important features, it might be reasonable to draft a Swift evolution proposal or even prototype an implementation, but I think the workarounds are easy enough that this isn't the best use of time.

PR #3817, opened 2016-07-01
Fix `schmaVersionAtURL:encryptionKey:error` to throw when imported into Swift

JadenGeller:
Fixes an issue where `schemaVersionAtURL:encryptionKey:error`, when used in Swift from the Objective-C API, would not import as throwing. Also, this now check return value to determine if there's an error in accordance to Cocoa conventions.

Fixes https://github.com/realm/realm-cocoa/issues/3816

### DIFF
--- a/Realm/RLMRealm.h
+++ b/Realm/RLMRealm.h
@@ -462,7 +462,8 @@ typedef void (^RLMMigrationBlock)(RLMMigration *migration, uint64_t oldSchemaVer
 
  @return The version of the Realm at `fileURL`, or `RLMNotVersioned` if the version cannot be read.
  */
-+ (uint64_t)schemaVersionAtURL:(NSURL *)fileURL encryptionKey:(nullable NSData *)key error:(NSError **)error;
++ (uint64_t)schemaVersionAtURL:(NSURL *)fileURL encryptionKey:(nullable NSData *)key error:(NSError **)error
+NS_REFINED_FOR_SWIFT;
 
 /**
  Performs the given Realm configuration's migration block on a Realm at the given path.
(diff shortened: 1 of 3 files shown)


JadenGeller:
Note that this doesn't break the API for anyone _except_ Swift users of the Realm Objective-C API. If this is unacceptable expect in a major version bump, I'll add the original API back as well and deprecate it.

JadenGeller:
Note that `NS_REFINED_FOR_SWIFT` was required since there is currently no way to indicate to the importer that a specific, non-zero return value indicates failure ([existing conventions](https://github.com/apple/swift-clang/blob/stable/include/clang/Basic/AttrDocs.td#L1808)).

tgoyne:
I'd be inclined to just leave the swift 2.2 RLMRealm API as-is (and not even bother deprecating it) since we get the free break with Swift 3.

JadenGeller:
@tgoyne I don't think I can apply `NS_REFINED_FOR_SWIFT` to only a single Swift version though, right?

tgoyne:
The Swift 2.2 implementation in RLMSupport.swift can implement the old API while the Swift 3.0 one implements the proper version.

JadenGeller:
@tgoyne The Swift function cannot exactly reimplement the Objective-C function since they'd share the same selector.

> Method 'schemaVersionAtURL(_:encryptionKey:error:)' with Objective-C selector 'schemaVersionAtURL:encryptionKey:error:' conflicts with method '__schemaVersionAtURL(_:encryptionKey:error:)' with the same Objective-C selector

With that in mind, should we keep with the deprecation approach?

tgoyne:
Marking the Swift method as `@nonobjc` should work.

JadenGeller:
@tgoyne Perfect! I had no idea that existed! Wish Xcode gave me a fixit there.

Issue #3790, opened 2016-06-22
Build proof of concept Realm with NS_REFINED_FOR_SWIFT and NS_SWIFT_NAME


jpsim:
This could potentially let us avoid have two frameworks like we do now: Realm.framework & RealmSwift.framework.

But we may not be able to get most of our current Swift API like this. But it's worth looking into how far we can get.

JadenGeller:
Even more useful, we can rename functions with the `swift_name` clang attribute.
```objc
#define SWIFT_NAME(x) __attribute__((swift_name(#x)))
+ (instancetype)defaultRealm SWIFT_NAME(init());
```

bdash:
AKA `NS_SWIFT_NAME`.

JadenGeller:
Swift 3.0 made `swift_name` a whole lot more powerful: https://github.com/apple/swift-evolution/blob/master/proposals/0044-import-as-member.md

JadenGeller:
Note that our `enums` would not be fully compatible between Objective-C and Swift. For example, the Objective-C code right now uses `NSString` values to indicate the type of [notification](https://realm.io/docs/swift/latest/api/Enums/Notification.html#/s:FO10RealmSwift12Notification9DidChangeFMS0_S0_) while Swift uses enums with `String` raw values. We could either (a) use string constants in both, (b) use integer-backed enums in both, or (c) reimplement this function for Swift in an extension and mark the Objective-C version unavailable.

In the case of enums with associated values (such as swift fine-grain notifications), we either have to ditch them in favor of the Objective-C implementation or reimplement this function in a Swift extension.

JadenGeller:
Another issue is that we cannot bridge Objective-C classes to Swift struct overlays. Though a protocol `_ObjectiveCBridgeable` exists, it isn't intended to be used by 3rd parties. Foundation uses this to provide value-semantic versions of some of its classes, like [`Date`](https://github.com/apple/swift/blob/d68c6dcd6b4be19cff9f6e82e2811884de56b6fa/stdlib/public/SDK/Foundation/Date.swift), but we couldn't (yet) use it for providing a value-semantic version of `Configuration`. We'd either have to (a) [hopefully temporarily] have totally separate APIs surrounding value-semantic classes or (b) settle for a less-Swifty, reference-semantic API in certain places.

austinzheng:
That's a good point. Do we have any 'structy' data types besides the configuration object?

JadenGeller:
[...]

PR #3818, opened 2016-07-01
Introduce `RLMNotification` typedef with Swift annotation

JadenGeller:
The importer will automatically convert our notification string constants into a Swift enum with `String` raw values. This is an improvement to the Objective-C API when used in Swift. Since this requires a feature only available in Xcode 8, this will _not_ break the API for existing users until they upgrade to Swift 3 (which is a breaking change anyway).

This is an incremental change that's necessary if we decide to go through with https://github.com/realm/realm-cocoa/issues/3790.

### DIFF
--- a/Realm/RLMConstants.h
+++ b/Realm/RLMConstants.h
@@ -18,6 +18,14 @@
 
 #import <Foundation/Foundation.h>
 
+// For compatibility with Xcode 7, before extensible string enums were introduced,
+#ifdef NS_EXTENSIBLE_STRING_ENUM
+#define RLM_EXTENSIBLE_STRING_ENUM_CASE_SWIFT_NAME(_, extensible_string_enum) NS_SWIFT_NAME(extensible_string_enum)
+#else
+#define NS_EXTENSIBLE_STRING_ENUM
+#define RLM_EXTENSIBLE_STRING_ENUM_CASE_SWIFT_NAME(fully_qualified, _) NS_SWIFT_NAME(fully_qualified)
+#endif
+
 #pragma mark - Enums
 
 /**
@@ -125,6 +133,11 @@ typedef NS_ENUM(NSInteger, RLMError) {
 #pragma mark - Notification Constants
 
 /**
+ A notification indicating that changes were made to a Realm.
+*/
+typedef NSString * RLMNotification NS_EXTENSIBLE_STRING_ENUM;
+
+/**
  This notification is posted by a Realm when the data in that Realm has changed.
 
  More specifically, this notification is posted after a Realm has been refreshed to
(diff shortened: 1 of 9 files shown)


Realm/RLMConstants.h
```diff
@@ -18,6 +18,14 @@
 
 #import <Foundation/Foundation.h>
 
+// For compatibility with Xcode 7, before extensible string enums were introduced,
+#ifdef NS_EXTENSIBLE_STRING_ENUM
```
JadenGeller: I'm not confident this is sufficient. If the `NS_EXTENSIBLE_STRING_ENUM` works in Swift 2.x when compiled with Xcode 8, then this will break the API on upgrade from Xcode 7 to Xcode 8 even if the developer uses a legacy Swift version. I'm not sure if this is problematic though, as it's a simple rename and would only break on upgrade.

I would definitely appreciate hearing others' thoughts on this though because I'm not confident I considered all possibilities.

JadenGeller:
Also, if I'm understanding Jenkins right, this change seems to break Jazzy? https://ci.realm.io/job/objc_pr/3619/configuration=Release,swift_version=2.2,target=docs/console //cc @jpsim

jpsim:
The jazzy failure is because the `RLMNotification` typedef is a public symbol but doesn't have a documentation-style comment (`/** ... */`)

JadenGeller:
@tgoyne @austinzheng Could you review and +1? 🙂

RealmSwift/Realm.swift
```diff
@@ -475,10 +475,13 @@ public final class Realm {
...
-                block(notification: Notification.DidChange, realm: self)
-            } else if rlmNotification == NSNotification.Name.RLMRealmRefreshRequired.rawValue {
-                block(notification: Notification.RefreshRequired, realm: self)
+            switch rlmNotification {
+            case RLMNotification.DidChange:
+                block(notification: .DidChange, realm: self)
+            case RLMNotification.RefreshRequired:
+                block(notification: .RefreshRequired, realm: self)
+            default:
+                fatalError("Unhandled notification type: \(rlmNotification)")
```
JadenGeller: The only _functional_ change is we now crash when propagating an unknown notification type in Swift. We used to silently ignore such a condition. Either way, this should never happen in user code. Maybe I should change `fatalErrror` to `preconditionFailure` to make this more clear?

austinzheng:
This looks fine to me, but at some point the Swift enum case names are going to change from starting with an uppercase letter to starting with a lowercase letter. Does that affect your work at all? If you want, maybe you can fold in that change.

JadenGeller:
@austinzheng That should be possible to change afterwards. I figured it made sense for all enums to be changed in a single commit later one.

austinzheng:
👍

Realm/RLMConstants.h
```diff
@@ -18,6 +18,14 @@
...
+#define RLM_EXTENSIBLE_STRING_ENUM_CASE_SWIFT_NAME(_, extensible_string_enum) NS_SWIFT_NAME(extensible_string_enum)
+#else
+#define NS_EXTENSIBLE_STRING_ENUM
```
bdash: I don't think we want to define this macro in our public headers. It'll break other code doing a similar #ifdef check on it.

JadenGeller: @bdash The macro needs to be used in our public headers though (since `NS_EXTENSIBLE_STRING_ENUM` and `NS_SWIFT_NAME` annotations must be in the header). Do you think that the `RLM` prefix isn't sufficient?

bdash: I think you've misunderstood what I'm referring to. By doing `#define NS_EXTENSIBLE_STRING_ENUM` any code that imports our header and does a check of the form `#ifdef NS_EXTENSIBLE_STRING_ENUM` will now see that as defined, but it's not defined in the way that they're likely to expect. What we should do instead is use `RLM_EXTENSIBLE_STRING_ENUM` in our headers in place of `NS_EXTENSIBLE_STRING_ENUM`, and conditionally define it to either `NS_EXTENSIBLE_STRING_ENUM` or nothing.

JadenGeller: Ohhhh, I understand now. Yeah, good catch!

PR #3827, opened 2016-07-05
Fix createOrUpdate when column order does not match property order

tgoyne:
It was incorrectly using the table column as an index in the input array for the primary key value.

@bdash @jpsim


jpsim:
thanks 👍

bdash:
👍

PR #3828, opened 2016-07-05
Introduce async handover

JadenGeller:
Allows users to easily perform asynchronous writes (https://github.com/realm/realm-cocoa/issues/3136)
```swift
realm.async(handingOver: object) { realm, object in
    try! realm.write {
        object.boolCol = true
    }
}
```


JadenGeller:
What happens if the user accesses a not-yet-created object on the dispatch queue that they previously saved a reference to (before beginning a read transaction at an earlier version)? We should make sure this produces a proper error message.

Realm/RLMRealm.mm
```diff
@@ -705,6 +705,37 @@ + (NSError *)migrateRealm:(RLMRealmConfiguration *)configuration {
...
+    dispatch_async(queue, ^{
+        @autoreleasepool {
+            RLMRealm *realm = [RLMRealm realmWithConfiguration:config error:nil];
+            std::vector<std::unique_ptr<realm::Row>> acceptedRows = realm->_realm->accept_handover(*package);
+
+            NSMutableArray<RLMObject *> *acceptedObjects = [NSMutableArray arrayWithCapacity:classNames.count];
+            for (NSUInteger i = 0; i < classNames.count; i++) {
+                RLMObjectSchema *schema = [realm.schema schemaForClassName:classNames[i]];
+                [acceptedObjects addObject: (RLMObject *)RLMCreateObjectAccessor(realm, schema,
+                                                                                 acceptedRows[i]->get_index())];
```
JadenGeller: @bdash @tgoyne Is there a better way to create a `RowExpr` from a `BasicRow`?

Realm/RLMRealm.mm
```diff
@@ -705,6 +705,37 @@ + (NSError *)migrateRealm:(RLMRealmConfiguration *)configuration {
     }
 }
 
+- (void)dispatchAsync:(dispatch_queue_t)queue handingOver:(NSArray<RLMObject *> *)objectsToHandOver
+        withBlock:(void(^)(RLMRealm *, NSArray<RLMObject *> *))block {
+    std::vector<realm::Row> rowsToHandOver;
+    rowsToHandOver.reserve(objectsToHandOver.count);
+    NSMutableArray<NSString *> *classNames = [NSMutableArray arrayWithCapacity:objectsToHandOver.count];
```
JadenGeller: I couldn't find a better way to store schema information than by name. We cannot store the `RLMObjectSchema` object since it is associated with a particular `RLMRealm`. Is there a way to store an index instead? I don't think that would be safe with runtime schema changes though. Is the current implementation compatible with runtime schema changes/ @tgoyne

tgoyne: You could store indexes by doing `[realm.schema.objectSchema indexOfObject:object->_objectSchema]` but it probably wouldn't actually end up being faster.

The source and destination SharedGroups have to be at the same transaction version so there shouldn't be any way for there to be runtime schema change-related issues.

RealmSwift/Realm.swift
```diff
@@ -1240,6 +1240,18 @@ public final class Realm {
         }
     }
 
+    public func async<C: RangeReplaceableCollectionType where C.Generator.Element: Object>(
```
JadenGeller: @jpsim Is this the sort of polymorphic support you imagined?

JadenGeller:
I don't think it's worth having a variadic overload since the closure argument would still be an array. If we offered such an overload, it would be very confusing that they arguments did not match.
```swift
realm.async(handingOver: objectA, objectB) { realm, allObjects in … }
```

RealmSwift/Realm.swift
JadenGeller: @jpsim I'm actually not sure how much sense it makes to offer a variant supporting polymorphic containers. In what situations would a user want to pass a non-array collection of Realm objects across thread boundaries?

- `Set`: It's not sensible to keep a set of objects since they're mutable so their hash value is subject to change. I guess a user might treat their objects as if they're immutable, or they must base the hash value solely on the object identifier, but these seem like less common situations.
- `Dictionary`: Though I do think it's reasonable to pass dictionaries across thread boundaries, this would _not_ be supported by the above overload since `Dictionary` isn't a `RangeReplaceableCollectionType`. I'm not sure how worthwhile such an overload is though. It would have to be clear that _only_ Realm objects are handed over in a thread safe way. And I think there are better alternatives anyway.
- `Sequence`: We can't pass an arbitrary sequence across thread boundaries without exposing it as either an `Array` or an `AnySequence` on the other end. It seems like the user should just build an array before passing, otherwise we may mislead the user into thinking that the sequenced is not traversed on handover (or that it is traversed lazily on access).

What are your thoughts?

JadenGeller: Note that I'm arguing against a `<C: RangeReplaceableCollectionType where C.Generator.Element: Object>` overload, not a `<O: Object>` overload.

JadenGeller:
Currently this Objective-C API is implemented.
```objc
- (void)dispatchAsync:(dispatch_queue_t)queue withBlock:(void(^)(RLMRealm *))block;
- (void)dispatchAsync:(dispatch_queue_t)queue handingOverObject:(RLMObject *)objectToHandOver
            withBlock:(void(^)(RLMRealm *, RLMObject *))block;
- (void)dispatchAsync:(dispatch_queue_t)queue handingOverObjects:(NSArray<RLMObject *> *)objectsToHandOver
            withBlock:(void(^)(RLMRealm *, NSArray<RLMObject *> *))block;
```

Do we think it's worthwhile to double the number of overloads to support a "default" queue argument?
```objc
- (void)dispatchAsyncWithBlock:(void(^)(RLMRealm *))block;
- (void)dispatchAsyncHandingOverObject:(RLMObject *)objectToHandOver withBlock:(void(^)(RLMRealm *, RLMObject *))block;
- (void)dispatchAsyncHandingOverObjects:(NSArray<RLMObject *> *)objectsToHandOver
            withBlock:(void(^)(RLMRealm *, NSArray<RLMObject *> *))block;
```

JadenGeller:
@bdash @tgoyne @jpsim @austinzheng

bdash:
I'm not super keen on the API.

* A queue-based API feels a little misleading since the objects are only valid for the duration of the block, and an attempt to use them on the same queue in the future is not valid.
* Methods prefixed `dispatchAsync` feel out of place on our `RLMRealm` class.

Realm/RLMRealm.mm
bdash: This is fine.

astigsen:
> A queue-based API feels a little misleading since the objects are only valid for the duration of the block, and an attempt to use them on the same queue in the future is not valid.

Are there any good use cases for needing to specify a specific queue? Otherwise it would be nice to avoid the need for the extra argument.

JadenGeller:
As @jpsim mentioned in https://github.com/realm/realm-cocoa/issues/3136#issuecomment-175328699, the user might want to specify the queue priority. I think it's a reasonable thing to expose (for more reasons than just that).

@bdash Any suggestions of a better naming scheme? We could write them as functions on the dispatch queue / top level functions I guess. The Realm could even be inferred based on the arguments (throwing if not the same for each). This kinda makes the queue a more prevalent part of the API though, so it doesn't fit well with your first concern.
```swift
queue.async(handingOver: [obj1, obj2]) { 
  // do stuff
}
```

JadenGeller:
@astigsen The argument is defaulted in Swift, and overloads can be provided in Objective-C to make it "defaulted" if necessary.

JadenGeller:
@bdash @astigsen Just talked with @jpsim offline, and he felt that we should definitely include the queue parameter. We can document that the objects are only valid for the duration of the block. I don't think this should be a huge surprise to users given that Realm objects are thread confined, and there's no guarantee what thread a queue runs on.

We also discussed the Objective-C overloads. We decided that a reasonable solution would be to only include the queue parameter in the most complicated overload rather than including every combination of conveniences.

RealmSwift/Realm.swift
```diff
@@ -1231,6 +1231,29 @@ public final class Realm {
...
+    }
+
+    public func async<O: Object>(onQueue queue: dispatch_queue_t = dispatch_get_global_queue(DISPATCH_QUEUE_PRIORITY_BACKGROUND, 0),
+                               handingOver object: O, execute block: (Realm, O) -> ()) {
+        rlmRealm.dispatchAsync(queue, handingOverObject: unsafeBitCast(object, RLMObject.self)) { realm, objects in
+            block(Realm(realm), unsafeBitCast(objects, O.self))
+        }
+    }
+
+    public func async<O: Object>(onQueue queue: dispatch_queue_t = dispatch_get_global_queue(DISPATCH_QUEUE_PRIORITY_BACKGROUND, 0),
```
JadenGeller: What's everyones thoughts on this explicitly taking an `Array` vs. it being `SequenceType` constrained?

I have a slight preference toward only accepting `Array` parameters since we vend an `Array` in the block. If we accepted any `SequenceType`, users could pass `RLMResults` in we'd strictly load and handover each object, rather than handing over the entire results (though maybe this is something that should be addressed by specialized overloads? https://github.com/realm/realm-cocoa/issues/3863)

Offline, @jpsim brought up the concern that users often don't realize that `Array.init` has a `SequenceType` overload, and thus won't understand how to pass their non-array collection.

//cc @austinzheng

bdash:
> @bdash Any suggestions of a better naming scheme? 

CoreData uses `performBlock:` for a similar type of behavior in its API. That's awkward given the extra arguments we need to include though.

> We could write them as functions on the dispatch queue / top level functions I guess.

It's a nice idea, but discoverability is likely to be an issue. There's also no obvious equivalent for Objective-C.

JadenGeller:
> CoreData uses performBlock: for a similar type of behavior in its API. That's awkward given the extra arguments we need to include though.
Agreed. We could call it `handOver` or something, but that name is less self-explanatory and doesn't work in the case where you want to perform async without handing over any objects.

> It's a nice idea, but discoverability is likely to be an issue. There's also no obvious equivalent for Objective-C.
Agreed. The equivalent would be top-level C functions, but this is not a great Objective-C API.

I'm not sure how to improve on `dispatchAsync`.

bdash:
> Agreed. We could call it handOver or something, but that name is less self-explanatory and doesn't work in the case where you want to perform async without handing over any objects.

Why is performing async work without handing over any objects an API we need to provide? It already exists as `dispatch_async`.

JadenGeller:
> Why is performing async work without handing over any objects an API we need to provide? It already exists as dispatch_async.

I'm not sure how necessary it is, but it essentially "passes" the Realm asynchronously. It's nice to have, I think. I don't feel strongly about it though, I just followed what @jpsim proposed in https://github.com/realm/realm-cocoa/issues/3136#issuecomment-175328699.

Realm/RLMRealm.h
```diff
@@ -495,11 +495,21 @@ __deprecated_msg("Use `performMigrationForConfiguration:error:`") NS_REFINED_FOR
  */
 + (BOOL)performMigrationForConfiguration:(RLMRealmConfiguration *)configuration error:(NSError **)error;
 
-- (void)dispatchAsync:(dispatch_queue_t)queue withBlock:(void(^)(RLMRealm *))block;
-- (void)dispatchAsync:(dispatch_queue_t)queue handingOverObject:(RLMObject *)objectToHandOver
-            withBlock:(void(^)(RLMRealm *, RLMObject *))block;
-- (void)dispatchAsync:(dispatch_queue_t)queue handingOverObjects:(NSArray<RLMObject *> *)objectsToHandOver
-            withBlock:(void(^)(RLMRealm *, NSArray<RLMObject *> *))block;
+- (void)dispatchAsyncWithBlock:(void(^)(RLMRealm *))block
```
JadenGeller: I updated the Objective-C methods to follow the naming conventions @jpsim suggested.

RealmSwift/Realm.swift
```diff
@@ -22,6 +22,16 @@ import Realm.Private
 
 #if swift(>=3.0)
 
+private let defaultQueue: DispatchQueue = {
+    let attributes: DispatchQueue.GlobalAttributes
+    if #available(OSXApplicationExtension 10.10, *) {
+        attributes = DispatchQueue.GlobalAttributes.qosBackground
+    } else {
+        attributes = DispatchQueue.GlobalAttributes.priorityBackground
```
JadenGeller: Xcode warns that we're using a deprecated symbol here, and I'm not sure how to silence it. I could define the queue from Objective-C (where this was not deprecated?) and use it from Swift, but that seems gross.

JadenGeller:
It was decided in https://github.com/realm/realm-cocoa/issues/3863 that we'll introduce a common protocol for types that support handover. I'm working on updating the pull request to include that feature.

Realm/RLMRealm.mm
```diff
@@ -713,6 +713,58 @@ + (BOOL)performMigrationForConfiguration:(RLMRealmConfiguration *)configuration
...
+                       block:(void(^)(RLMRealm *, NSArray<RLMObject *> *))block {
+    std::vector<realm::Row> rowsToHandOver;
+    rowsToHandOver.reserve(objectsToHandOver.count);
+    NSMutableArray<NSString *> *classNames = [NSMutableArray arrayWithCapacity:objectsToHandOver.count];
+    for (RLMObject *object in objectsToHandOver) {
+        if (self != object.realm) {
+            @throw RLMException(@"Can only hand over objects from the Realm they belong to.");
+        }
+        rowsToHandOver.push_back(object->_row);
+        [classNames addObject: object.objectSchema.className];
```
tgoyne: No space after colon in obj-c.

Realm/RLMRealm.mm
```diff
@@ -713,6 +713,58 @@ + (BOOL)performMigrationForConfiguration:(RLMRealmConfiguration *)configuration
...
+    }
+    std::shared_ptr<Realm::HandoverPackage> package = _realm->package_for_handover(rowsToHandOver);
+
+    RLMRealmConfiguration *config = self.configuration;
+    dispatch_async(queue, ^{
+        @autoreleasepool {
+            RLMRealm *realm = [RLMRealm realmWithConfiguration:config error:nil];
```
tgoyne: Although it's *very unlikely* for this to fail, it is possible (e.g. due to running out of memory or hitting the maximum number of open files), so we do need to forward the error on to the user's block.

JadenGeller: How do you suggest we expose this error? I think we need either (1) a second closure for failure or (2) a nullable error argument to the callback in addition to making all other arguments nullable as well.

It seems like (1) is the better option, but it breaks trailing closure syntax in Swift unless we make it the first argument. We could define it like this
```swift
public func async<O: Object>(onQueue queue: DispatchQueue = defaultQueue,
                             handingOver objects: [O],
                             onFailure recover: (ErrorType) -> () = { fatalError(String($0)) },
                             execute block: (Realm, [O]) -> ()) { … }
```
and call it like this
```swift
realm.async(handingOver: [carl, daisy], onFailure: { error in performCustomRecovery(error) }) { realm, objects in
    // do async stuff
}
```

Let me know if you have a more clever solution, @tgoyne or @austinzheng.

tgoyne:
I don't think the name `async()` really works. With a dispatch queue all you can do is submit blocks to it, so the individual function names don't need to communicate that fact, but that isn't the case for a Realm. A CoreData-esq `performBlock:` also seems like a poor fit, since unlike CoreData the Realm doesn't have it's own internal queue. In Swift 3 the current API feels like it'd be more natural as a protocol extension on the DispatchQueue, but that doesn't help with obj-c or swift 2.

Realm/RLMRealm.mm
tgoyne: Passing multiple callbacks is pretty unpleasant. The notification callbacks take the route of a nullable NSError in obj-c and an enum for success/error in swift, so barring a good reason to be different I'd be inclined to do the same thing for consistency.

JadenGeller: One last idea: We could return a sort of promise that can be used to recover from error. The API would look like this
```swift
@discardableResult
public func async<O: Object>(onQueue queue: DispatchQueue = defaultQueue,
                             handingOver objects: [O],
                              execute block: (Realm, [O]) -> ()) -> AsyncError { … }
```
and would be called like this
```swift
let asyncError = async(handingOver: [foo, bar]) { realm, objects in

}
// ignore async error, and it'll crash on error
// or handle async error
asyncError.handle { error in performCustomRecovery(error) }
```

I think this API is interesting and nicely hides the error handling from most users (who don't need it), but it's very different than Cocoa conventions and our current API, so I don't think it's necessarily a great idea. Just throwing it out there though.

JadenGeller: @tgoyne Though I think using an enum is a clean solution, it doesn't give us the ability to detect when the user ignores our error. In the case of an additional parameter, we can provide a default that crashes the application. With an enum though, I expect to see a lot of code written like
```swift
realm.async(handingOver: [foo, bar]) { result in
  if let .success(realm, objects) = result {
    // do stuff async
  }
}
```
This approach brings error handling into a much more prevalent position for the user. Since most of these errors are not going to be things the user can recover from, it introduces boilerplate code that a user will likely mess up by ignoring the error.

JadenGeller: I guess we could provide a `value() throws` function on the result enum, and the user could use it
```swift
realm.async(handingOver: [foo, bar]) { result in
  let (realm, objects) = try! result.value()
  try! realm.write { 
    // write stuff async
  }
}
```
but we're still introducing so many hoops that the user has to jump through to do stuff asynchronously.

JadenGeller:
I agree that it'd be neat to expose these functions as an extension on `DispatchQueue` in Swift 3. Wouldn't that imply that we could expose these functions as top level functions taking a `dispatch_queue_t` in Objective-C and Swift 2.x? I don't think that'd be that odd.

Realm/RLMRealm.mm
JadenGeller: After talking with @tgoyne and @austinzheng offline, we decided that an `unwrap() throws` method is probably the best approach. This means that RealmSwift will include a new type
```swift
enum Result<Element> {
	case success(Element)
	case failure(ErrorType)
	
	func unwrap() throws -> Element {
		switch self {
		case .success(let value):
			return value
		case .failure(let error):
			throw error
		}
	}
}
```

bdash: `Result` seems confusingly close to our existing `Results` type.

JadenGeller: Good point. Unfortunately this is a [pretty](https://doc.rust-lang.org/std/result/enum.Result.html) [common](https://github.com/antitypical/Result) [name](https://github.com/kittinunf/Result). I'm having troubling thinking of any alternatives, but we definitely can't use such a similar name. Any suggestions?

bdash: What's the rationale for taking a different approach than our notification callbacks use?

JadenGeller: We can't simply introduce a single, non-generic type since we have overloads for passing 0 objects, 1 object, or many objects. Further, these overloads are generic, leading to many different function signatures in practice. We either have to introduce a generic result type.

The motivation for the `unwrap` function is that we want to make it difficult to accidentally ignore the error. Additionally, we want to make the API as concise as possible. While I do not consider the current design particularly concise, it's better than this boilerplate:
```swift
realm.async(handingOver: [foo, bar]) { result in
    switch result {
    case let .success((realm, object)):
        try! realm.write {
            // do async writes
        }
    case let .failure(error):
        fatalError(error)
    }
}
```

JadenGeller:
This is what the API could look like if we added extensions to `DispatchQueue`:
```swift
DispatchQueue.global(attributes: .qosBackground).async(handingOver: [foo, bar]) { result in
    let (realm, objects) = try! result.unwrap()
    try! realm.write {

    }
}
```

The biggest thing I notice: This makes the `DispatchQueue` a very prevalent part of the API. Before, we hid it away in a default parameter for advanced users. Now it brings it front and center.

I still think that it really does make sense, especially in Swift, to define the async method as an extension on the queue. Despite that though, it's definitely not the easiest API to show users, and doesn't make the prettiest sample code.

astigsen:
> it's definitely not the easiest API to show users, and doesn't make the prettiest sample code.

Both of which are really high priority requirements :-)

JadenGeller:
I don't disagree, but it feels very elegant to follow `DispatchQueue`'s API. I don't see why a dispatch async call that hands over a Realm object should look *simpler* than a standard dispatch async all.

JadenGeller:
The API has been updated to expose an `exportForThreadHandover` function that returns an `HandoverPackage<T>` with a throwing `importOnCurrentThread` function. This API is the most versatile (can be used w/o GCD) and avoids the issue of async error handling.

```swift
let package = realm.exportForThreadHandover([foo, bar])
DispatchQueue.global().async {
    let (realm, objects) = try! package.importOnCurrentThread()
    try! realm.write {
        // do stuff
    }
}
```

This function maintains the exact same type safety that the previously discussed closure variant would. We would also like to include a convenience API for users who don't need to handle errors. There are a few options (names not final):
1. `realm.async(handingOver: [foo, bar]) { realm, objects in try! realm.write { … } }`
2. `realm.asyncWrite(handingOver: [foo, bar]) { realm, objects in … }`
3. `queue.async(handingOver: [foo, bar]) { realm, objects in try! realm.write { … } }`
Though `realm.asyncWrite` is the most concise option, users that would like to perform an async read would need to use the significantly lengthier API. If we thing async reads will be common, we should probably opt for `realm.async`. Note that all 3 of these APIs hide some error handling (crashing on failure), `asyncWrite` more than the others.

JadenGeller:
//cc @jpsim @austinzheng @bdash

astigsen:
How about:

`realm.writeAsync(handingOver: [foo, bar]) { realm, objects in … }`

It would have the nice effect that it show up in auto-completion every time someone does a write transaction, so it would help educating users about the option to do it async.

JadenGeller:
@astigsen Sure. Do you think an async read convenience API is unnecessary then?

JadenGeller:
I'll name the equivalent Objective-C function `[RLMRealm transactionAsyncWithObjects:block:]` to match the Swift naming.

astigsen:
> Do you think an async read convenience API is unnecessary then?

I think we can wait with that until people get a bit of experience with the write API, and we see if that makes them request an async read API as well.

bdash:
`transactionAsyncWithObjects:` sounds incredibly awkward.

JadenGeller:
@bdash I agree. It would be preferable to use `writeAsyncWithObjects:block:`, but we've already established a precedent of using `transaction` instead of `write` in Objective-C.

We could use `asyncTransactionWithObjects:block:`, but this loses the auto-complete that @astigsen was looking for.

I'm open to better suggestions.

tgoyne:
`transactionWithBlock`'s name is wrong by Cocoa's naming scheme (it's a noun phrase, so it should return a transaction object). Since it's just a single method I don't think we should continue this mistake, especially when it has such awkward results.

JadenGeller:
@tgoyne Would you you be for deprecating the `transaction` family of APIs and replacing them with a `write` family of APIs? I think it's a good idea to do eventually, but I'm not sure how you'd feel about such a major API change in a non-major release. (Remember, there's `transactionWithBlock:`, `beginTransaction`, `commitTransaction`, `cancelTransaction`)

It would be weird to only deprecate `transactionWithBlock:` and it would also be weird to introduce `writeAsyncWithBlock:` without also renaming the former.

tgoyne:
The other methods are `beginWriteTransaction` etc., not `beginTransaction`, and there's nothing wrong with them (they're all void-returning methods with verbs in the name).

JadenGeller:
@tgoyne Good catch. It seems like `performTransactionWithBlock:` would be the Cocoa convention-following name most similar to our current API.

I think we should either:
1. Rename `transactionWithBlock:` to `performTransactionWithBlock:` and introduce `performTransactionAsyncWithBlock:` (follow Cocoa conventions with smallest API diff)
2. Rename `transactionWithBlock:` to `writeWithBlock:` and introduce `writeAsyncWithBlock:`; consider renaming `___WriteTransaction` family to `____Write` (aligning with Swift)
3. Keep the existing name and introduce `asyncTransactionWithBlock:` (no API changes, though harder to discover via autocomplete)

JadenGeller:
It would also be reasonable to name the Objective-C method `__handingOverObject:withBlock` instead of `__withObject:block`, but I think the latter more closely follows Objective-C conventions.

JadenGeller:
@bdash @tgoyne You commented earlier, but didn't give a 👍. Are you still reviewing the PR, or are there other changes you'd like to see? Also, y'all didn't love the naming. What are your thoughts on my suggested alternatives?

JadenGeller:
I'd also appreciate if the documentation could be reviewed since this is a pretty major addition. @jpsim

bdash:
I've been thinking about the `asyncWrite` convenience API a little and I'm having trouble understanding how we expect people to use it. There are two broad categories of writes that come to mind:
1. User-triggered writes
2. Non-user-triggered writes

The `asyncWrite` API seems like a poor fit for writes that result from a user interaction as it inherently leaves the state on the UI thread unchanged until the async write has committed and the UI thread's Realm has advanced to the new version. If the write takes a noticeable amount of time to commit (due to write contention, slow disk, etc.) the user will wonder if their change has been ignored as it's not reflected in the UI. An app that wishes to have user changes be immediately reflected in the UI would have to forgo binding their UI directly to our live accessor objects, or build some other mechanism on top of our accessors to integrate in-flight changes into the UI.

`asyncWrite` doesn't seem to bring much to the table for non-user-triggered writes either. There's typically work that needs to be performed before the write transaction proper to prepare the data that will be written (at the very least the data needs to be converted into instances of Realm model classes). This work should also be performed off the main thread, but it's not necessary (undesirable, even) to do it within the write transaction.

What am I missing?

Realm/RLMRealm.mm
```diff
@@ -712,6 +713,40 @@ + (BOOL)performMigrationForConfiguration:(RLMRealmConfiguration *)configuration
...
+                                                        NSArray<id<RLMThreadConfined>> * _Nonnull objects) {
+        REALM_ASSERT_DEBUG(objects.count == 1);
+        block(realm, objects[0]);
+    }];
+}
+
+- (void)transactionAsyncWithObjects:(NSArray<id<RLMThreadConfined>> *)objects
+                              block:(void(^)(RLMRealm *, NSArray<id<RLMThreadConfined>> *))block {
+    RLMThreadHandover *package = [self exportThreadHandoverWithObjects:objects];
+    dispatch_async(dispatch_get_global_queue(DISPATCH_QUEUE_PRIORITY_BACKGROUND, 0), ^{
```
bdash: I'm not sure about using `DISPATCH_QUEUE_PRIORITY_BACKGROUND` here. Nothing about an async write inherently indicates that it's of a lower priority. For instance, a priority of user initiated may make more sense if the write is triggered by a user action and the app is expecting the write to trigger an update of the UI. The background quality of service class throttles I/O, which may result in the transaction commit taking significantly longer than if it were running at a higher QoS class. The background priority queues are also serviced after higher-priority queues, meaning they can be starved if the app is performing a lot of work on any of the higher-priority queues.

JadenGeller: Okay, we'll use `qosDefault`.

JadenGeller:
@astigsen What're your thoughts on @bdash's concerns? Maybe we should merge the non-convenience API and later add a convenience API once users have had some time to use the feature?

Realm/ObjectStore/handover.cpp
```diff
@@ -0,0 +1,190 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+////////////////////////////////////////////////////////////////////////////
+
+// This is lets our internals access AnyHandover's private constructors.
+#define private public
```
tgoyne: This is not legal C++ (can't redefine keywords). Use `friend class` to give access to specific things.

Realm/ObjectStore/handover.cpp
```diff
@@ -0,0 +1,190 @@
...
+#include "handover.hpp"
+#undef private
+
+using namespace realm;
+
+#pragma mark AnyThreadConfined
```
tgoyne: No `#pragma mark` in object store files (it's a non-standard extension and the OS code isn't clang-only like the obj-c code is).

Realm/ObjectStore/handover.cpp
```diff
@@ -0,0 +1,190 @@
...
+        case Type::LinkViewRef:
+            m_link_view_ref.~LinkViewRef();
+            break;
+    }
+}
+
+AnyHandover AnyThreadConfined::export_for_handover(SharedGroup &shared_group) const {
+    switch (m_type) {
+        case AnyThreadConfined::Type::Row:
+            return AnyHandover (shared_group.export_for_handover(m_row));
```
tgoyne: Only language control-flow things have a space between the name and opening parenthesis.

Realm/ObjectStore/handover.cpp
```diff
@@ -0,0 +1,190 @@
...
+// This is lets our internals access AnyHandover's private constructors.
+#define private public
+#include "handover.hpp"
+#undef private
+
+using namespace realm;
+
+#pragma mark AnyThreadConfined
+
+AnyThreadConfined::AnyThreadConfined(const AnyThreadConfined& thread_confined) {
```
tgoyne: C++ coding style is a newline before `{` for function definitions only (obj-c code doesn't follow this).

Realm/ObjectStore/handover.cpp
JadenGeller: What do you mean by language control-flow? Declarations?

Realm/ObjectStore/handover.hpp
```diff
@@ -0,0 +1,120 @@
+////////////////////////////////////////////////////////////////////////////
+//
+// Copyright 2015 Realm Inc.
```
tgoyne: 2016

Realm/ObjectStore/handover.hpp
```diff
@@ -0,0 +1,120 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+////////////////////////////////////////////////////////////////////////////
+
+#ifndef handover_hpp
```
tgoyne: realm_handover_hpp

Realm/ObjectStore/handover.hpp
```diff
@@ -0,0 +1,120 @@
...
+#define handover_hpp
+
+#include <realm/row.hpp>
+#include <realm/query.hpp>
+#include <realm/group_shared.hpp>
+#include <realm/link_view.hpp>
+#include <realm/table_view.hpp>
+
+namespace realm {
+    class HandoverPackage;
```
tgoyne: Don't indent the body of namespaces.

Realm/ObjectStore/handover.hpp
```diff
@@ -0,0 +1,120 @@
...
+    class SharedGroup;
+
+    class AnyHandover;
+
+    // Type-erased wrapper for any type which must be exported to be handed between threads
+    class AnyThreadConfined {
+    public:
```
tgoyne: Our usual class member ordering is all public, then all private, with things within the sections being data members, constructors/etc., then other member functions.

Realm/ObjectStore/handover.hpp
```diff
@@ -0,0 +1,120 @@
...
+        AnyThreadConfined(Query query)               : m_type(Type::Query),       m_query(query)                 { }
+        AnyThreadConfined(TableRef table_ref)        : m_type(Type::TableRef),    m_table_ref(table_ref)         { }
+        AnyThreadConfined(TableView table_view)      : m_type(Type::TableView),   m_table_view(table_view)       { }
+        AnyThreadConfined(LinkViewRef link_view_ref) : m_type(Type::LinkViewRef), m_link_view_ref(link_view_ref) { }
+
+        AnyThreadConfined(const AnyThreadConfined& thread_confined);
+        AnyThreadConfined(AnyThreadConfined&& thread_confined);
+        ~AnyThreadConfined();
+
+        inline Type type() const { return m_type; }
```
tgoyne: `inline` here is redundant (member functions defined inline are implicitly `inline`, as otherwise including the header in multiple cpp files would always be a linker error).

Realm/ObjectStore/shared_realm.cpp
```diff
@@ -518,5 +518,50 @@ util::Optional<int> Realm::file_format_upgraded_from_version() const
...
+
+    ~HandoverPackage() {
+        if (is_awaiting_import()) {
+            REALM_TERMINATE("Handover package not imported, leaking pinned version");
+        }
+    }
+};
+
+std::shared_ptr<Realm::HandoverPackage> Realm::package_for_handover(std::vector<AnyThreadConfined> objects_to_hand_over) {
+    std::shared_ptr<Realm::HandoverPackage> handover = std::make_shared<HandoverPackage>();
```
tgoyne: The type of the value is explicitly spelled-out on the RHS, so this is one of the clearcut places for using `auto`.

Realm/ObjectStore/list.hpp
```diff
@@ -106,6 +106,8 @@ class List {
         size_t valid_count;
     };
 
+    LinkViewRef get_linkview() const { return m_link_view; }
```
tgoyne: It'd be nice to continue to avoid exposing the LinkView directly. Could maybe just make List/Result be able to produce an AnyThreadConfined?

tgoyne: Or maybe even the AnyHandover object directly.

JadenGeller: Yeah, I didn't realize this was something we didn't want to expose.

Realm/ObjectStore/handover.cpp
tgoyne: `if`/`while`/`for`/etc.

JadenGeller: Oh yes. But initialization also has a space in the case of declarations, no?
```c++
Foo x (arg1, arg2);
```

tgoyne: No.

JadenGeller: Is this spacing correct?
```swift
new (&m_row) RowHandover(std::move(handover.m_row));
```

Realm/ObjectStore/list.hpp
JadenGeller: We already expose it in `results.hpp`—why should we be inconsistent?

bdash: If you're referring to `Results::get_linkview`, it's not used after realm/realm-object-store@ef1c6ddc63d71a21c87434c056c6097ba8644abb, and should likely have been removed.

JadenGeller: Ok 😭

JadenGeller: Exposing `get_anythreadconfined` seems a lot more gross to me, especially considering there's a `linkview` member function on `AnyThreadConfined`.

astigsen:
> @astigsen What're your thoughts on @bdash's concerns? Maybe we should merge the non-convenience API and later add a convenience API once users have had some time to use the feature?

It is obviously not going to help when you want synchronous results, like when the UI has to be updated right away. But sometimes you do want to things async and that is where it will be useful.

There could be a good case for both having `writeAsync` and `readAsync`, since I agree that are cases where you want to do something before the write. But when I have been talking with users they have mostly complained about not being able to do async writes to objects without cumbersome re-fetching of all the objects.

I don't have any problem with merging it in two steps, but I do think that the convenience API is important.

JadenGeller:
I definitely feel we shouldn't provide both an `asyncRead` and `asyncWrite` API. The concept of reads does not exist in our users' mental model.

Since an async read is easily composable with the existing `write` method, and since an async read can thought of as passing a Realm with a bundle of objects between threads, my preference would be that we expose an async read convenience API. This API also doesn't have to expose a new way to handle write errors; users can use the same function they've always used, and can handle errors the exact same way.

If we do think that writes are such an overwhelmingly common case that an `asyncWrite` API ought to exist, then I think we shouldn't provide an async read convenience API since it'll lead to confusion. I'm definitely not convinced that most users need `asyncWrite`, or even that users who think they need it actually would benefit from it.

PR #3830, opened 2016-07-06
Replace RLMUpdateQueryWithPredicate with RLMPredicateToQuery

tgoyne:
Muting the query in-place used to be required due to issues with Query's node
ownership logic, but that's all been fixed and returning a new Query makes some
of the call sites nicer and none of them worse.

@bdash


bdash:
Nice!

👍

PR #3832, opened 2016-07-06
Improve `RLMRealm` Swift interface for initialization and transactions

JadenGeller:
Annotates `RLMRealm` members for better import into Swift as part of https://github.com/realm/realm-cocoa/issues/3790. Note that this is a source-breaking change for Swift users of Realm Objective-C. If we care to not break source compatibility for that particular user base, then we should delay merging this until a large version bump.

### DIFF
--- a/Realm/RLMRealm.h
+++ b/Realm/RLMRealm.h
@@ -53,6 +53,9 @@ NS_ASSUME_NONNULL_BEGIN
 
 #pragma mark - Creating & Initializing a Realm
 
+- (instancetype)init
+NS_UNAVAILABLE;
+
 /**
  Obtains an instance of the default Realm.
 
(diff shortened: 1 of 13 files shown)


JadenGeller:
@austinzheng

austinzheng:
Looks good to me! I don't know what our source-breaking story is for Swift users of Realm Objective-C, so please ask Thomas, Mark, or JP.

JadenGeller:
^^ @tgoyne @bdash @jpsim

tgoyne:
A breaking change like this would mean bumping the version to 2.0, and a month after we released 1.0 is far too early for that. The Swift 2.2 -> 3 transition is somewhat of a special case in that it's effectively a new language that we're targeting so there is no older versions to maintain compatibility with, but I'm assuming that anything based on annotating the obj-c API can't be swift 3-only.

JadenGeller:
To the best of my understanding, yes. Unless…
Is there a way the preprocessor can detect what version of Swift the user is compiling with?

JadenGeller:
Looks like such a macro might be coming soon! https://twitter.com/UINT_MIN/status/751164808553046016

JadenGeller:
Assigning to myself to remind me to update this once such a feature becomes available.

JadenGeller:
Actually, I'll close this PR for now, and make a new PR when such a feature is available. Until then, these sorts of changes can live on this branch: https://github.com/realm/realm-cocoa/tree/jg-improved-swift-import

PR #3834, opened 2016-07-06
Updates for compatibility with Xcode 8 beta 2

austinzheng:
Pretty straightforward. I don't think Swift recognizes `warn_unused_result` anymore, since the default behavior was switched around.

@tgoyne @jpsim @JadenGeller @bdash

### DIFF
--- a/RealmSwift/LinkingObjects.swift
+++ b/RealmSwift/LinkingObjects.swift
@@ -384,7 +384,6 @@ public final class LinkingObjects<T: Object>: LinkingObjectsBase {
      - parameter block: The block to be called with the evaluated linking objects and change information.
      - returns: A token which must be held for as long as you want updates to be delivered.
      */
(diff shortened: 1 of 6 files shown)


austinzheng:
I'm going to merge this in, since NFC, unless anyone objects.

JadenGeller:
I didn't realize that you could provide a message with `warn_unused_result`. Given that you could, that's a bit of a regression, but there's not much we can do about it.

+1

austinzheng:
Yeah, it's kind of unfortunate, but at least users will have to work harder to misuse the API now :).

JadenGeller:
Yeah, I'm definitely happy about the change of default, I just wish we could still provide a message. A bit of an oversight, it seems.

JadenGeller:
I'll write up a Swift Evolution proposal for this since I think the warning message is pretty important in our usage.

@austinzheng and I were thinking something along the lines of `@discardableResultWarning(message: "You must hold on to the NotificationToken returned from addNotificationBlock")`

PR #3835, opened 2016-07-06
Changing Realm Swift's Error type to just be a typealias for RLMError

austinzheng:
There's no real good reason to have `Error` in Realm Swift be its own separate enum. Having an `==` for non-specific `ErrorProtocol` instances is also kind of problematic.

@tgoyne @jpsim @bdash @JadenGeller

### DIFF
--- a/RealmSwift/Error.swift
+++ b/RealmSwift/Error.swift
@@ -21,106 +21,29 @@ import Realm
 #if swift(>=3.0)
 
 /**
-Enumeration that describes the error codes within the Realm error domain.
-The values can be used to catch a variety of _recoverable_ errors, especially those
-happening when initializing a Realm instance.
+ `Error` is an enum representing all recoverable errors. It is associated with the Realm error domain specified in
+ `RLMErrorDomain`.
 
-    let realm: Realm?
-    do {
-        realm = Realm()
-    } catch RealmSwift.Error.IncompatibleLockFile() {
-        print("Realm Browser app may be attached to Realm on device?")
-    }
+ `Error` is a Swift `ErrorType`:
 
-*/
-public enum Error: ErrorProtocol {
-    // swiftlint:disable variable_name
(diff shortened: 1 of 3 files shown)


JadenGeller:
+1

Will Jazzy still generate Swift documentation for `Error` cases?

austinzheng:
Ugh. Good question, let me try regenerating the docs and see what happens.

austinzheng:
I wasn't able to get Jazzy working with Xcode 8, but I do know what happens: we won't get the per-case documentation for `Error` anymore. We may want to move the summary of the error cases into the main documentation, and then add a link.

JadenGeller:
Aw, that definitely sucks. It's still probably a good move to consolidate these types though, so still +1 from me.

JadenGeller:
How difficult would it be to extend Jazzy so that "exported" classes also generate documentation? @jpsim

austinzheng:
I think what I'm going to do is as follows:

* Pull the descriptions for each error case into the online documentation.
* In both the objc and swift bindings, add a link in the type's docstrings to the documentation. (I'll leave the current ObjC per-case documentation in place.)

This is sort of like how the `PropertyType` alias is documented.

Would anyone have any objections to this?

JadenGeller:
How does this change interact with the [NSError bridging changes](https://github.com/apple/swift-evolution/blob/master/proposals/0112-nserror-bridging.md) in Swift 3? Does this change limit what sort of information we can include with errors in the future (since `RLMError ` is a primitive type).

jpsim:
Jazzy doesn't document inherited members, which also includes typealiases. See #190.

JadenGeller:
@jpsim Wrong issue number?

jpsim:
Right issue number, wrong repo: realm/jazzy#190

austinzheng:
Superseded by #3934, which also debated this issue.

PR #3836, opened 2016-07-06
Improve consistency add/delete methods on RLMReam

JadenGeller:
Improves the naming and typing of methods to add or delete multiple objects from a Realm.

### DIFF
--- a/Realm/RLMRealm.h
+++ b/Realm/RLMRealm.h
@@ -375,7 +375,7 @@ typedef void (^RLMNotificationBlock)(RLMNotification notification, RLMRealm *rea
 
  @see   `addObject:`
  */
-- (void)addObjects:(id<NSFastEnumeration>)array;
+- (void)addObjects:(NSObject<NSFastEnumeration> *)array;
 
 /**
  Adds or updates an existing object into the Realm.
@@ -401,11 +401,11 @@ typedef void (^RLMNotificationBlock)(RLMNotification notification, RLMRealm *rea
 
  @warning This method may only be called during a write transaction.
 
- @param array  An `NSArray`, `RLMArray`, or `RLMResults` of `RLMObject`s (or subclasses) to be added to the Realm.
+ @param array  An enumerable object such as `NSArray` or `RLMResults` which contains objects to be added to the Realm.
 
  @see   `addOrUpdateObject:`
  */
-- (void)addOrUpdateObjectsFromArray:(id)array;
+- (void)addOrUpdateObjects:(NSObject<NSFastEnumeration> *)array;
 
 /**
  Deletes an object from the Realm. Once the object is deleted it is considered invalidated.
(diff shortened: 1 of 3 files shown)


Realm/RLMRealm.h
```diff
@@ -401,11 +401,11 @@ typedef void (^RLMNotificationBlock)(RLMNotification notification, RLMRealm *rea
 
  @warning This method may only be called during a write transaction.
 
- @param array  An `NSArray`, `RLMArray`, or `RLMResults` of `RLMObject`s (or subclasses) to be added to the Realm.
+ @param array  An enumerable object such as `NSArray` or `RLMResults` which contains objects to be added to the Realm.
 
  @see   `addOrUpdateObject:`
  */
-- (void)addOrUpdateObjectsFromArray:(id)array;
+- (void)addOrUpdateObjects:(NSObject<NSFastEnumeration> *)array;
```
JadenGeller: Renaming `addOrUpdateObjectsFromArray` to `addOrUpdateObjects` is a source-breaking change. Should we also retain (and deprecate) the old name?

(Note that the method was renamed for consistency with `addObjects`.)

Realm/RLMRealm.h
```diff
@@ -423,11 +423,11 @@ typedef void (^RLMNotificationBlock)(RLMNotification notification, RLMRealm *rea
 
  @warning This method may only be called during a write transaction.
 
- @param array  An `RLMArray`, `NSArray`, or `RLMResults` of `RLMObject`s (or subclasses) to be deleted.
- 
+ @param array  An enumerable object such as `NSArray` or `RLMResults` which contains objects to be deleted.
```
JadenGeller: The doc-comments were changed to be consistent in phrasing between methods. Let me know if you notice a regression.

Realm/RLMRealm.mm
```diff
@@ -628,23 +628,20 @@ - (void)deleteObject:(RLMObject *)object {
     RLMDeleteObjectFromRealm(object, self);
 }
 
-- (void)deleteObjects:(id)array {
+- (void)deleteObjects:(NSObject<NSFastEnumeration> *)array {
```
JadenGeller: Previously, this method took an argument of type `id`, which is not very type safe. As such, we threw an exception when it was neither a type that conformed to `NSFastEnumeration` or a Realm collection. We'd like to mark this argument as `id<NSFastEnumeration>`, as the arguments of the other methods were, to ensure that only enumerable collections are passed in. Since Realm collections are themselves enumerable, this is a reasonable API.

In order to check whether or not an object is a Realm collection, we have to call `respondsToSelector:`, which is defined on the `NSObject` protocol. I think it's fair to assume that our API will only be used with NSObject subclasses (Swift collections should instead use a `SequenceType` variant of this method), so I marked its argument `NSObject<NSFastEnumeration>` and I did the same to the other methods for consistency.

Let me know if you think this is reasonable.

JadenGeller:
@austinzheng @jpsim @tgoyne @bdash

tgoyne:
Using `NSObject` rather than `id` is unidiomatic in obj-c and doesn't have any benefits.

While `addOrUpdateObjects:` is a better name, we are not at a point where we should be renaming methods just for the sake of a slightly better name.

JadenGeller:
@tgoyne I don't care much that it's a better or worse name, I just think it ought to be consistent with our other names. Having both `addOrUpdateObjectsFromArray:` and `deleteObject:`, which both taking identical argument types, is confusing.

The benefit gained from using `NSObject` instead of `id` is that the former defines `respondsToSelector:`, which we need in `deleteObjects:`. We could instead use `id` and cast to `NSObject` inside the method (under the assumption that all objects are `NSObject`s), but this seems worse and would break from Swift (though it should be `NS_REFINED_FOR_SWIFT` anyway).

PR #3840, opened 2016-07-07
Fix tests with Xcode 8 beta 2

bdash:
* The macOS 10.12 SDK includes a new class that declares a `column` property with different return type so it's now necessary to be explicit about the type of the object that we're accessing the `column` property on.
* Attempting to cast `CFRunLoopMode.defaultMode` to `CFTypeRef` aborts at runtime. Using the raw value of the enum works correctly.


austinzheng:
👍

PR #3841, opened 2016-07-07
Rework query building to not use column numbers from RLMProperty

tgoyne:
Explicitly pass in the Group to the query builder code and look up tables in it and column numbers in the Tables rather than getting those indirectly from RLMObjectSchema and RLMProperty. This eliminates any dependency on the private members of the schema classes and makes the query code ready for runtime schema changes. The slowdown from the extra column number lookups is in the range of a few percent.

Refactor the code by putting most of the free functions into a class that stores the Schema, Query and Group to avoid having to pass around quite so many values to every single function.

@bdash

### DIFF
--- a/Realm/RLMQueryUtil.mm
+++ b/Realm/RLMQueryUtil.mm
@@ -18,17 +18,18 @@
 
 #import "RLMQueryUtil.hpp"
 
-#import "RLMArray_Private.hpp"
+#import "RLMArray.h"
+#import "RLMObjectSchema.h"
 #import "RLMObject_Private.hpp"
-#import "RLMObjectSchema_Private.hpp"
 #import "RLMPredicateUtil.hpp"
-#import "RLMProperty_Private.h"
-#import "RLMSchema_Private.h"
+#import "RLMProperty.h"
+#import "RLMSchema.h"
 #import "RLMUtil.hpp"
 
+#import "object_store.hpp"
 #import "results.hpp"
 
-#include <realm.hpp>
+#include <realm/query_engine.hpp>
 
 using namespace realm;
 
@@ -73,6 +74,16 @@ RLMProperty *RLMValidatedProperty(RLMObjectSchema *desc, NSString *columnName) {
 }
 
 namespace {
+BOOL RLMPropertyTypeIsNumeric(RLMPropertyType propertyType) {
+    switch (propertyType) {
+        case RLMPropertyTypeInt:
+        case RLMPropertyTypeFloat:
+        case RLMPropertyTypeDouble:
+            return YES;
+        default:
+            return NO;
+    }
+}
 
 // FIXME: TrueExpression and FalseExpression should be supported by core in some way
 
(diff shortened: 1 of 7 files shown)


bdash:
Can you provide some context as to why looking up column indexes in the table helps with runtime schema changes?

> The slowdown from the extra column number lookups is in the range of a few percent.

Was this when measuring only the time it takes to build the query, without executing it?

tgoyne:
> Can you provide some context as to why looking up column indexes in the table helps with runtime schema changes?

Currently RLMProperty instances are shared between all RLMRealm instances for a given file, which is safe to do because they don't change. With runtime schema changes the `column` property would need to change at different times on different threads, so they either need to no longer be shared or the column number needs to no longer come from the RLMProperty. The former dramatically increases the time it takes to open realms on background threads when the realm is already open on another thread (more than doubles IIRC), so I'm currently pursuing the latter.

> Was this when measuring only the time it takes to build the query, without executing it?

With the execution time with zero rows in the table being queried, so the time in find_all() is negligible.

Realm/RLMQueryUtil.mm
```diff
@@ -18,17 +18,18 @@
...
-#import "RLMSchema_Private.h"
+#import "RLMProperty.h"
+#import "RLMSchema.h"
 #import "RLMUtil.hpp"
 
+#import "object_store.hpp"
 #import "results.hpp"
 
-#include <realm.hpp>
+#include <realm/query_engine.hpp>
```
bdash: Why query_engine.hpp rather than query.hpp?

tgoyne: We need query_expression.hpp but that doesn't actually compile without query_engine.hpp.

Realm/RLMQueryUtil.mm
```diff
@@ -138,36 +149,36 @@ void set_base_table(const Table*) override {}
...
-    template <typename T>
-    auto resolve_with_subquery(Query& query, Query subquery) const
+    template <typename T, typename... SubQuery>
+    auto resolve(SubQuery&&... subquery) const
     {
+        static_assert(sizeof...(SubQuery) < 2, "resolve() takes at most one subquery");
+        set_link_chain_on_table();
         if (type() != RLMPropertyTypeLinkingObjects) {
-            return table_for_query(query)->template column<T>(index(), std::move(subquery));
+            return m_table->template column<T>(index(), std::move(subquery)...);
```
bdash: Should be `std::forward<SubQuery>` rather than `std::move`?

Realm/RLMQueryUtil.mm
```diff
@@ -138,36 +149,36 @@ void set_base_table(const Table*) override {}
...
         }
         else {
-            RLMObjectSchema *link_origin_schema = m_schema[property().objectClassName];
-            Table& link_origin_table = *link_origin_schema.table;
-            size_t link_origin_column = link_origin_schema[property().linkOriginPropertyName].column;
-            return table_for_query(query)->template column<T>(link_origin_table, link_origin_column, std::move(subquery));
+            return resolve_backlink<T>(std::is_same<T, Link>(), std::move(subquery)...);
```
bdash: Ditto about `std::forward`.

Realm/RLMQueryUtil.mm
```diff
@@ -190,64 +201,80 @@ bool has_any_to_many_links() const {
...
-        if (!m_links.size()) {
-            return util::none;
-        }
-        auto links = m_links;
-        RLMProperty *property = links.back();
-        links.resize(links.size() - 1);
-        return ColumnReference(m_schema, property, std::move(links));
+    ColumnReference last_link_column() const {
+        REALM_ASSERT(!m_links.empty());
+        return {*m_query, *m_group, m_schema, m_links.back(), {m_links.begin(), m_links.end() - 1}};
```
bdash: Nice.

Realm/RLMQueryUtil.mm
```diff
@@ -190,64 +201,80 @@ bool has_any_to_many_links() const {
...
-    ColumnReference column_ignoring_links() const {
-        return {m_schema, m_property};
+    ColumnReference column_ignoring_links(Query& query) const {
+        return {query, *m_group, m_schema, m_property};
     }
 
 private:
-    Table* table_for_query(Query& query) const
+    template <typename T, typename... SubQuery>
+    auto resolve_backlink(std::true_type, SubQuery&&... subquery) const
```
bdash: It's not obvious what `true_type` signifies here. It's necessary to refer to the implementation of `resolve` to find out. We could clarify this with a comment, or by having a `resolve_backlink` overload that encapsulates the type dispatch.

bdash:
👍, with those minor points of feedback.

Realm/RLMQueryUtil.mm
tgoyne: What do you think of the following:

    struct ConstexprIf {
        template<typename Then, typename Else>
        static auto call(std::true_type, Then&& fn, Else&&) {
            return fn();
        }

        template<typename Then, typename Else>
        static auto call(std::false_type, Then&&, Else&& fn) {
            return fn();
        }
    };

    class ColumnReference {
        template <typename T, typename... SubQuery>
        auto resolve(SubQuery&&... subquery) const
        {
            static_assert(sizeof...(SubQuery) < 2, "resolve() takes at most one subquery");
            set_link_chain_on_table();
            if (type() != RLMPropertyTypeLinkingObjects) {
                return m_table->template column<T>(index(), std::move(subquery)...);
            }
            else {
                return ConstexprIf::call(std::is_same<T, Link>(), [&](auto...) {
                    return with_link_origin(m_property, [&](Table& table, size_t col) {
                        return m_table->template column<T>(table, col, std::move(subquery)...);
                    });
                }, [](auto...) -> Columns<T> { __builtin_unreachable(); });
            }
        }
    };

This fixes the locality issue and makes it much more clear what the true_type/false_type is doing (and that it's faking `if constexpr` via tag dispatch), but there's also some dreadfully subtle bits such as that the `auto...` is there to prevent the body of the lambda from being instantiated when the other case is used.

bdash: I think I'd prefer something closer to your original approach until we're able to use `if constexpr` directly.

bdash:
👍

Issue #3842, opened 2016-07-07
Update Swift test cases for Realm Objective-C to Swift 3.0


JadenGeller:
Currently, we have empty `#if swift(>=3.0)` branches in the Realm Objective-C Swift tests. It's probably worthwhile to add these back and make sure everything works as expected in Swift 3.0.

Issue #3843, opened 2016-07-07
Objective-C generics as properties of Swift classes show up as type `@` in the runtime


JadenGeller:
We should file a radar and/or a Swift bug report.

bdash:
Is this with all Xcode and Swift versions, or is it new in the Xcode 8 / Swift 3 betas?

JadenGeller:
Objective-C generics did not import into Swift (expect for classes with special bridging already like NSArray) until Swift 3.0 / Xcode 8. I should check that this still is an issue on the latest beta. I think I'm behind a version.

JadenGeller:
I just verified that this is an existing issue in beta 2. Hopefully we can get it fixed by the time Swift 3.0 is released. Regardless, there's a workaround, so we can still run our tests in the meantime. https://github.com/realm/realm-cocoa/commit/f70fb036f3e389f8be9fb74c1d5a54e2dd1f15cf

JadenGeller:
Filed [SR-2031](https://bugs.swift.org/browse/SR-2031).

PR #3846, opened 2016-07-08
Update Realm Objective-C's Swift Tests for Swift 3`

JadenGeller:
We currently don't have *any* tests for Realm Objective-C in Swift 3.0, so I updated the Swift 2.x ones. https://github.com/realm/realm-cocoa/issues/3842

### DIFF
--- a/Realm/RLMProperty.mm
+++ b/Realm/RLMProperty.mm
@@ -333,7 +333,12 @@ static bool rawTypeIsComputedProperty(NSString *rawType) {
 
     // temporarily workaround since Objective-C generics used in Swift show up as `@`
     if ([_objcRawType isEqualToString:@"@"]) {
-        _objcRawType = [NSString stringWithFormat:@"@\"%@\"", [propertyValue class]];
+        if (propertyValue) {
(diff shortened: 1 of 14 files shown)


JadenGeller:
@austinzheng @bdash

Realm/RLMProperty.mm
```diff
@@ -342,6 +342,16 @@ - (instancetype)initSwiftPropertyWithName:(NSString *)name
 
     id propertyValue = [obj valueForKey:_name];
 
+    // temporarily workaround since Objective-C generics used in Swift show up as `@`
```
bdash: Please make this a `FIXME: `, add information about which versions of Xcode and Swift this problem occurs in, and add a link to the Radar and/or swift.org bug report. It's important to make it easy for a future reader to determine whether this block of code can go away.

jpsim:
I agree with @bdash's comment on adding more info to your objc comment. Otherwise 👍. I didn't do a thorough review of the Swift 3 tests, but we'll need to do that again when Xcode 8 goes GM anyway.

JadenGeller:
Addresses https://github.com/realm/realm-cocoa/issues/3843

PR #3847, opened 2016-07-08
remove RLMSupport's test() function

jpsim:
was probably just debugging code introduced in #3817 /cc @JadenGeller


JadenGeller:
Oops, sorry about that! Definitely +1 😝

PR #3848, opened 2016-07-08
Improve type-safety of RLMResults subscript by making return type generic rather than `id`

JadenGeller:


JadenGeller:
@austinzheng

austinzheng:
👍

tgoyne:
Also need to update the swift 2.2 version of the tests as they now warn about pointless casts.

JadenGeller:
I thought Swift 2.2 didn't import Objective-C generics? I'll check it out and find out.

JadenGeller:
@tgoyne You're right, there *are* warning! But, not because the generics were imported into Swift 2.2. Changing the return type to the generic type imported it as `RLMObject`, the class constraint on the generic. 👍

bdash:
This should really have had an entry in the changelog

JadenGeller:
@bdash Sorry, should I create a new PR to add an entry? I'm not entirely clear on what sort of changes require an entry. Is it anything that breaks the API (warnings or errors), introduces features, or fixes major bugs?

bdash:
Any changes in API, new features, enhancements to existing features, or bug fixes should have changelog entries. The only things that typically don't get changelog entries are internal refactoring with no user-visible consequences. I'd also exclude Swift 3 specific changes for now since we don't officially support it, and we'll likely be announcing all of the changes in a single location when our support for it becomes official.

It'd be worth adding an entry still as we've not yet released this change.

PR #3849, opened 2016-07-08
Removed check for `RLMDynamicRealm`, a nonexistent class

JadenGeller:
The checks are actually only checking that it was a `RLMRealm`, which is statically known to be true and thus unnecessary. I assume they historically checked for a `RLMDynamicRealm`, but this class doesn't seem to exist anymore.

@tgoyne @bdash

### DIFF
--- a/Realm/Tests/Swift/SwiftDynamicTests.swift
+++ b/Realm/Tests/Swift/SwiftDynamicTests.swift
@@ -40,7 +40,6 @@ class SwiftDynamicTests: RLMTestCase {
         }
         let dyrealm = realm(withTestPathAndSchema: nil)
         XCTAssertNotNil(dyrealm, "realm should not be nil")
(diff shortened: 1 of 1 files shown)


Realm/Tests/Swift/SwiftDynamicTests.swift
```diff
@@ -40,7 +40,6 @@ class SwiftDynamicTests: RLMTestCase {
         }
         let dyrealm = realm(withTestPathAndSchema: nil)
         XCTAssertNotNil(dyrealm, "realm should not be nil")
-        XCTAssertTrue(dyrealm is RLMRealm, "realm should be of class RLMDynamicRealm")
```
JadenGeller: Source of warnings in Swift code since this check is trivial.

tgoyne:
👍 

I don't think RLMDynamicRealm ever actually existed.

PR #3855, opened 2016-07-11
Annotate nullability for RLMRealm_Dynamic.h

JadenGeller:
@austinzheng

### DIFF
--- a/Realm/RLMRealm_Dynamic.h
+++ b/Realm/RLMRealm_Dynamic.h
@@ -23,6 +23,8 @@
 
 @class RLMResults;
 
+NS_ASSUME_NONNULL_BEGIN
+
 @interface RLMRealm (Dynamic)
 
 #pragma mark - Getting Objects from a Realm
@@ -112,3 +114,5 @@
 -(RLMObject *)createObject:(NSString *)className withValue:(id)value;
 
 @end
+
+NS_ASSUME_NONNULL_END
--- a/Realm/Tests/QueryTests.m
+++ b/Realm/Tests/QueryTests.m
@@ -137,8 +137,8 @@
     XCTAssertThrows([[realm objects:@"" where:@"age > 25"] sortedResultsUsingProperty:@"age" ascending:YES], @"missing class name");
 
     // nil class name
-    XCTAssertThrows([realm objects:nil where:@"age > 25"], @"nil class name");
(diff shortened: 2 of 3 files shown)


austinzheng:
:shipit:

tgoyne:
This is a breaking change.

JadenGeller:
@tgoyne How so? It does introduce warnings in cases where the API is used incorrectly (and will throw an exception as a result), but this does not seem _breaking_ to me.

tgoyne:
Changing from an optional to a non-optional makes Swift code written for optionals not compile.

JadenGeller:
Got it. Just to be clear, we fully support Swift users of the Objective-C API, and we do not want to introduce changes that could break this API except large Realm or Swift version bumps. Is that correct?

tgoyne:
The dynamic API and using the obj-c framework from Swift are both discouraged (and in the case of the dynamic API deliberately not mentioned in the main guide), but are still both part of our supported public API.

We are trying to [follow SemVer](http://semver.org/), with the relevant part here being that if you can compile and run a thing using version X.Y of a library, you should also be able to compile and run it with X.(Y+1). Swift 3 is a somewhat unusual case in that there *isn't* a version X.Y that you can use with it, so X.(Y+1) doesn't have anything that it needs to maintain compatibility with.

We could of course just bump the major version every month and make breaking changes willy-nilly, but that tends to make people trying to use a library very unhappy.

JadenGeller:
Awesome, thanks for the explanation @tgoyne. I guess this'll have to hold off.

Do we have any way to track which PRs should go into the next breaking release? It seems like they could easily be lost in the abyss.

austinzheng:
This is something easy we could pull into 3.0, if deemed appropriate.

larkost:
The current failure in CI seems to be a glitch in Jenkins, specifically the workspace for the trigger job being malformed/empty. I don't have enough insight to be sure, but sending a new version (e.g. a comment change) through might be enough to get past it.

austinzheng:
Yeah, this is a hoary, old PR that's way out of date, so I'm not concerned about CI failures. I think once I rebase and push it'll start working fine.

austinzheng:
These changes were actually made (inadvertently, I think) in a different commit that made it into master, and so I will close this PR as being redundant.

Issue #3837, opened 2016-07-06
Should `RealmCollectionType` conform to `LazyCollectionType`?


JadenGeller:
Swift defines a type `LazyCollectionType: CollectionType` "on which normally-eager operations such as map and filter are implemented lazily." Since`RLMResults`, `RLMArray`, and `RLMLinkingObjects` lazily load objects into memory and perform operation such as filters lazily, would it make sense to conform `RealmCollectionType` to `LazyCollectionType`? Conformance ensures that operations like `map` don't load the entire collection into memory, instead only loading elements as they are needed.

Caveats
- This change will silently change the performance of some user code if they use certain default implementations on `CollectionType`. The functions will still produce the same results—usually more efficiently too—but might sometimes produce worse performance (for example, if they iterate over a mapped collection many successive times). Maybe we should rename the protocol so users realize that this change was made?

JadenGeller:
Here's what the implementation would look like: https://github.com/realm/realm-cocoa/tree/jg-lazy

austinzheng:
Cool! Is this going to affect the ability of a user to randomly index into the collection? Does that even matter?

JadenGeller:
No effect whatsoever.

JadenGeller:
Here's an example of where behavior will change:
```swift
let foo = results.map { x in
  print(x)
  return x
}
let bar = foo[n]
let baz = foo[n]
```
In the strict implementation, this would print every element once. In the lazy, implementation, this would print the `n`th element twice and no other element.

JadenGeller:
If the user would like strict collection semantics, they can aways use `Array.init(_:)` and operate on that. Currently, a user can opt into lazy semantics with realm collections using `myResults.lazy` to obtain a lazy view of that collection. I'm proposing we change the default semantics to lazy since that better reflects the current behavior of the collection.

JadenGeller:
@tgoyne What're you thoughts on this given that it both breaks source compatibility (in easily fixed, uncommon cases) and, more importantly, it mostly-silently changes the semantics of the collection? It might be reasonable to rename the protocol `RealmLazyCollection` as part of this change if we decide to go through with it.

tgoyne:
Since it's a breaking change it'd need to be swift 3-only.

I believe there was a reason we didn't use `LazyCollectionType` originally, but if so I don't know why. @jpsim may remember. If there's no hidden gotchas it sounds like the right thing to do to me.

jpsim:
[...]

PR #3866, opened 2016-07-13
Give Realm collections lazy semantics in Swift 3

JadenGeller:
Please review and look for any "hidden gotchas" of this change, @jpsim. Discussed in https://github.com/realm/realm-cocoa/issues/3837.


JadenGeller:
This PR only makes the change for Swift 3, so it is not breaking.

Should be put this in the release notes? We don't typically report changes that only apply to Swift 3, but it seems we should eventually acknowledge this change in release notes when we officially support Swift 3. @jpsim

jpsim:
I think we'll need to do a full review of the Swift 3 stuff when Xcode 8 goes GM anyway, and at that point we can add the appropriate items to the changelog.

jpsim:
👍

PR #3867, opened 2016-07-13
Update to core v1.3.1

### LINKED ISSUE #3853: Querying objects having linked objects and live updates
## Goals

In our project we have two kind of calendar event types (for example A and B) and we display them in one calendar view filtering them by `employeeId` and `date` (from the example they might seem quite similar but they are really different so treating them as similar ones are not an option). Up until now we were querying both A and B with separate queries and were having two references to `RLMResult`. Now we want to start using fine-grained notifications and redraw calendar view when events (both A and B) are updated, new ones are added or deleted. The problem is that A and B are updated during the same API call so there would be two separate notifications which would lead to redrawing calendar twice. The first idea was to wrap them into some kind of `CalendarEvent`:
```ObjC
@interface CalendarEvent : RLMObject
@property (nonatomic) A* a;
@property (nonatomic) B* b;
@end
```
but it's pretty impossible or at least will require a lot of refactoring around the app wrapping all API responses having `A` or `B` into `CalendarEvent`. Then we've found inverse relationships method and made `CalendarEventToken`:
```ObjC
@interface CalendarEventToken : RLMObject
@property (nonatomic) NSInteger identifier;
@property (readonly) RLMLinkingObjects *aEvents;
@property (readonly) RLMLinkingObjects *bEvents;
@end

@implementation CalendarEventToken
+ (NSString *)primaryKey {
    return @"identifier";
}
+ (NSDictionary *)linkingObjectsProperties {
    return @{
        @"aEvents": [RLMPropertyDescriptor descriptorWithClass:A.class propertyName:@"token"],
        @"bEvents": [RLMPropertyDescriptor descriptorWithClass:B.class propertyName:@"token"]
    };
}
@end
```
and of course added one additional `token` property for each A and B.

## Expected Results

Query
```ObjC
RLMResults *twoTypesOfEventsWrappedIntoTokenResults = [CalendarEventToken objectsWhere:@"(ANY aEvents.employeeId = %@) OR (ANY bEvents.employeeId = %@)", @(self.employeeId), @(self.employee)]
```
worked fine but when events have been updated RLMResults were not updated nor notification block has been called.

I understand that it's a odd case and not sure that it's a good practice doing that but we have not found any other solution. It would be totally understandable if CalendarEventTokens are not updated if their inner objects are updated but would be good to know that.

PS. If you have any ideas how we can solve our problem that would be awesome!

## Actual Results

Actual results are shown in demo project.
https://github.com/viktorasl/Realm-RLMLinkingObjects/blob/master/Realm-RLMLinkingObjects/ViewController.m

Realm version: 1.0.1 (and some other lower versions)

Xcode version: 7.3.1

iOS/OSX version: 9.3

Dependency manager + version: Cocoapods 1.0.1

bdash:
Fixes #3853.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,8 @@ x.x.x Release notes (yyyy-MM-dd)
 
 * Deprecated `migrateRealm:` in favor of new `performMigrationForConfiguration:error:` method
   that follows Cocoa's' NSError conventions.
-* Attempting to add an object with no properties to a Realm now throws.
+* Attempting to add an object with no properties to a Realm now throws rather than silently
+  doing nothing.
 
 ### Enhancements
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@ x.x.x Release notes (yyyy-MM-dd)
 
 * Deprecated `migrateRealm:` in favor of new `performMigrationForConfiguration:error:` method
   that follows Cocoa's' NSError conventions.
+* Attempting to add an object with no properties to a Realm now throws.
 
 ### Enhancements
 
@@ -28,6 +29,9 @@ x.x.x Release notes (yyyy-MM-dd)
   results from that Realm.
 * Handle EINTR from flock() rather than crashing.
 * Fix incorrect behavior following a call to `[RLMRealm compact]`.
+* Fix live updating and notifications for Results created from a predicate involving
+  an inverse relationship to be triggered when an object at the other end of the relationship
+  is modified.
 
 1.0.1 Release notes (2016-06-12)
 =============================================================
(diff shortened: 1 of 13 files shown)


CHANGELOG.md
```diff
@@ -5,6 +5,7 @@ x.x.x Release notes (yyyy-MM-dd)
 
 * Deprecated `migrateRealm:` in favor of new `performMigrationForConfiguration:error:` method
   that follows Cocoa's' NSError conventions.
+* Attempting to add an object with no properties to a Realm now throws.
```
tgoyne: "now throws rather than silently doing nothing." to make it clear that it didn't actually work before.

tgoyne:
👍

PR #3868, opened 2016-07-13
Add entry for fix changing `RLMResults` subscript return type to generic type

JadenGeller:
https://github.com/realm/realm-cocoa/pull/3848


JadenGeller:
@bdash

bdash:
👍

Issue #2840, opened 2015-11-11
Detect computed properties by checking for the existence of backing ivars


mrackwitz:
Assumingly, that this Swift code, won't synthesize an ivar for `bar`:

``` swift
class Foo : Object {
    dynamic var foo: String = "foo"
    var bar: String {
        get { return foo }
        set { foo = newValue }
    }
}
```

If we could detect by the absence of this ivar, that `bar` is actually a fully computed property, we could avoid having users to add them to `ignoredProperties` when defining their schema classes.

That would be helpful in cases like #2815.

michelalbers:
+1

dantheli:
+1

JadenGeller:
We can also do this in Objective-C, as the ivar is not automatically synthesized when both the getter and the setter are manually implemented. A property lacking  `V` attribute is lacking an ivar and thus computed. https://developer.apple.com/library/mac/documentation/Cocoa/Conceptual/ObjCRuntimeGuide/Articles/ocrtPropertyIntrospection.html

austinzheng:
We're now using `breaking` in this repo to indicate API changes that require a major version increment (per semver), so I'm going to untag this issue.

tgoyne:
Doing this would be an API change that requires a major version increment.

austinzheng:
This is in `master-3.0`.

markst:
What's the go if we *do not* wish a particular computed property to be ignored from Realm?

tgoyne:
What would you expect that to do? A property cannot be both computed and stored in Realm.

markst:
@tgoyne something along these lines as an example:
```swift
@objc dynamic var latestTest:Test? {
    get {
        if let latestTestId = self.latestTestId {
            return self.tests.filter("id == %@",latestTestId).first
        }
        return RLMDynamicGetByName(self, "latestTest", false) as? Test
    }
    set {
        self.latestTestId = newValue?.id
        RLMDynamicValidatedSet(self, "latestTest", newValue)
    }
}
```

austinzheng:
The most straightforward solution is to have a stored property and wrap it manually with a second, computed property you intend the user to use. The difference is purely aesthetic.

markst:
I've tried the following:
```swift
@objc private var test:Test?
@objc dynamic var latestTest:Test? {
    get {
        if let test = self.test {
            return test
        }
        if let latestTestId = self.latestTestId {
[...]

PR #1867, opened 2015-05-07
Have +[RLMObjectSchema schemaForObjectClass:] throw an exception when…

bdash:
… it sees the same property name multiple times for one class.

Fixes #1844.

/cc @segiddins @jpsim @tgoyne

Issue #3864, opened 2016-07-12
Eventually merge breaking changes into master


JadenGeller:
Since we don't want to introduce API-breaking changes except in large releases, we should track such issues or PRs here.
- [Annotate nullability for RLMRealm_Dynamic.h](https://github.com/realm/realm-cocoa/pull/3855)
- [Improve consistency add/delete methods on RLMReam](https://github.com/realm/realm-cocoa/pull/3836)
- [Automatically ignore computed properties when building schema](https://github.com/realm/realm-cocoa/pull/3870)

JadenGeller:
Please add new changes by editing the issue rather than by adding new comments.

austinzheng:
Closing, since all 3 linked PRs have long been merged.

austinzheng:
This was closed erroneously.

austinzheng:
All three of these changes have been merged into our `master-3.0` branch.

Better naming for API: https://github.com/realm/realm-cocoa/pull/5189
Detect computed properties: https://github.com/realm/realm-cocoa/pull/5160
Nullability on dynamic API: (already in)

PR #3870, opened 2016-07-14
Automatically ignore computed properties when building schema

JadenGeller:
Ignores properties not backed by ivars in both Swift and Objective-C (https://github.com/realm/realm-cocoa/issues/2840). Also makes testing for duplicate properties unnecessary since both cannot be ivar backed (https://github.com/realm/realm-cocoa/pull/1867).

//cc @bdash @tgoyne @mrackwitz @jpsim

### DIFF
--- a/Realm/RLMObjectSchema.mm
+++ b/Realm/RLMObjectSchema.mm
@@ -126,20 +126,6 @@ using namespace realm;
     }]];
     schema.computedProperties = computedProperties;
 
-    // verify that we didn't add any properties twice due to inheritance
-    if (allProperties.count != [NSSet setWithArray:[allProperties valueForKey:@"name"]].count) {
-        NSCountedSet *countedPropertyNames = [NSCountedSet setWithArray:[allProperties valueForKey:@"name"]];
-        NSSet *duplicatePropertyNames = [countedPropertyNames filteredSetUsingPredicate:[NSPredicate predicateWithBlock:^BOOL(id object, NSDictionary *) {
-            return [countedPropertyNames countForObject:object] > 1;
-        }]];
-
-        if (duplicatePropertyNames.count == 1) {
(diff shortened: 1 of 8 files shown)


JadenGeller:
Note that this will silently ignore properties marked as `@dynamic` (in Objective-C) or `@NSManaged` since they are not ivar backed. This should be considered a breaking change if we expect that users might be doing this. Perhaps they're only using their objects managed by Realm, and mark the properties `@dynamic` to reduce storage size for an instance of the object. To combat this, we might introduce an `managedProperties` method to parallel our `ignoredProperties` method?

tgoyne:
It's not just `@dynamic`. Currently there's nothing stopping you from having a property that's computed on unmanaged instances and persisted on managed instances, since we don't rely on the ivar existing for properties other than generic swift types. I'm not sure if there's any scenarios where that'd actually be *useful* or a good idea, but that doesn't mean no one is doing it...

JadenGeller:
@tgoyne I feel like it's reasonable to only manage properties with backing ivars by default. If a user creates an unsynthesized property that gets/sets an unrelated ivar, it think it's fair to treat this as a computed property unless they opt out. This is a very weird use case, and I think we should optimize for ease-of-use in the common case.

Just like we do not assume all properties are computed unless they explicit specify we should manage them, I don't think we should assume all properties should be managed unless they are ignored. It seems like a reasonable default would ignore properties without ivar backing.

Obviously this would be a breaking API change, so we couldn't merge this for a while, but I'd argue we should eventually transition to this behavior.

JadenGeller:
(Also, I definitely think it's surprising behavior that computed properties are (silently) backed by Realm by default. As a user, I would be very shocked by this.)

JadenGeller:
Closing for now, but will be eventually merged when we're prepared to make breaking changes https://github.com/realm/realm-cocoa/issues/3864

JadenGeller:
More motivation to ignore properties not backed by ivars by default: https://github.com/realm/realm-cocoa/issues/3994#issuecomment-241821129

PR #3874, opened 2016-07-14
Renaming a few Swift 3 methods

austinzheng:
@bdash @jadengeller @jpsim @tgoyne 

Changes:
- 'createObject(ofType:populatedWith:)' is now 'createObject(_:from:)'
- 'createDynamicObject(ofType:populatedWith:)' is now 'createDynamicObject(ofType:from:)'
- 'allObjects(ofType:)' is now 'objects(ofType:)'
- 'allDynamicObjects(ofType:)' is now 'dynamicObjects(ofType:)'

### DIFF
--- a/RealmSwift/Migration.swift
+++ b/RealmSwift/Migration.swift
@@ -127,7 +127,7 @@ public final class Migration {
     - returns: The created object.
     */
     @discardableResult
-    public func createObject(ofType typeName: String, populatedWith value: AnyObject = [:]) -> MigrationObject {
+    public func createObject(_ typeName: String, from value: AnyObject = [:]) -> MigrationObject {
         return unsafeBitCast(rlmMigration.createObject(typeName, withValue: value), to: MigrationObject.self)
     }
 
@@ -202,7 +202,7 @@ extension Migration {
     @available(*, unavailable, renamed:"enumerateObjects(ofType:_:)")
     public func enumerate(_ objectClassName: String, _ block: MigrationObjectEnumerateBlock) { }
 
-    @available(*, unavailable, renamed:"createObject(ofType:populatedWith:)")
+    @available(*, unavailable, renamed:"createObject(_:from:)")
     public func create(_ className: String, value: AnyObject = [:]) -> MigrationObject {
         fatalError()
     }
--- a/RealmSwift/Realm.swift
+++ b/RealmSwift/Realm.swift
@@ -263,7 +263,7 @@ public final class Realm {
     - returns: The created object.
     */
     @discardableResult
-    public func createObject<T: Object>(ofType type: T.Type, populatedWith value: AnyObject = [:], update: Bool = false) -> T {
+    public func create<T: Object>(_ type: T.Type, from value: AnyObject = [:], update: Bool = false) -> T {
         let typeName = (type as Object.Type).className()
         if update && schema[typeName]?.primaryKeyProperty == nil {
             throwRealmException("'\(typeName)' does not have a primary key and can not be updated")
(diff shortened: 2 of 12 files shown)


JadenGeller:
I understand that `ofType:` is necessary for dynamic objects because of the `String` argument and not necessary for normal objects, but I'd prefer consistency over slightly shortening the call site of `createObject`.

JadenGeller:
Have we ever considered doing some sort of user testing to determine optimal API naming? I wonder if `allObjects` is sufficiently more clear than `objects` that it's worth the verbose API.

austinzheng:
I'd also rather keep `ofType:` in the methods, but I guess I'll wait for some more folks to chime in.

JadenGeller:
If we're going to inconsistently use `ofType:`, we should at least *only* use it on `String` arguments.

JadenGeller:
Maybe this is crazy, but thoughts on
```swift
func create<T>(_ type: T.Type, from: AnyObject)
func create(_ type: DynamicType, from: AnyObject)
```
where `DynamicType` is a struct wrapping a `String`?

bdash:
`objects(…)`, rather than `allObjects(…)`, is consistent with the naming of `Dictionary.keys` / `Dictionary.values` in Swift. The Objective-C equivalents on `NSDictionary` have the "all" prefix but they're absent from the Swift API.

bdash:
Should this change also make the label name of `Object(value:)` consistent with the label used in `Realm.create`?

austinzheng:
@bdash Good catch on `Object(value:)`; they should all be consistent.

Would anyone object if I restored `ofType:` to the APIs?

@JadenGeller A struct wrapper is too heavy IMO, but a typealias would be almost as good (wouldn't stop you from doing stupid things, but it would let you see the intended purpose of the string at a glance).

bdash:
Is `realm.createObject(ofType: Person.self, from: […])` really so much clearer than `realm.create(Person.self, from: […])` that it justifies the extra overhead? "Create object of type person" doesn't seem like a big improvement over "create person" to me.

austinzheng:
It's not, but then you have `object(Foo.self)` instead of `object(ofType: Foo.self)`, which I think is a regression. But maybe not? Or maybe we can just keep `ofType:` for one but not the other? It wouldn't be a disaster if we got rid of `ofType:` altogether.

bdash:
`Realm.objects(Person.self)`, while what we currently have for Swift 2, doesn't read as naturally as some of our other method names. Including `ofType:` does help in this case. We could potentially omit it if we changed `objects` for a method name like `fetch` or `all`, but then the fact that type names are singular nouns starts feeling awkward (`Realm.all(Person.self)`).

austinzheng:
Maybe we should keep `ofType:` if the function name is a noun (e.g. `objects()`), and strike it if the function name is a verb (in which case the type is the implied grammatical direct object: "Realm, create a Person"). I think that would give us both consistency and a good user experience.

JadenGeller:
@austinzheng A typealias could be nice, but not a replacement for the argument label. Did you have thoughts on using an `RLMObjectSchema` instead of a `String` to specify the type?

piv199:
Well, as of Swift3 `createObject` should be replaced with `create`.

austinzheng:
@bdash @JadenGeller Changed `createObject(...)` to `create(...)`.

austinzheng:
@JadenGeller using the object schema instead of the string is an interesting idea, but should probably be discussed and implemented separately.

austinzheng:
Are these renamings still something we want to pursue? If so I'll rebase and merge, after making any further improvements that you might suggest.

jpsim:
This is still something that needs to happen, and the churn of the language is one reason I've been staying away from this, though supposedly now it's mostly settled?

JadenGeller:
I think it has mostly settled.

austinzheng:
Closed since the API changes are being made in #4078.

PR #3882, opened 2016-07-18
fix compilation with Swift 2.3 using Xcode 8 beta 2

### LINKED ISSUE #3749: Support Swift 2.3 in Xcode 8 beta 1
## Goals
Use Realm with Xcode 8 and Swift 2.3

## Actual Results
When using the current version I get the error message that the module was built with an old compiler.

When I try to build a new version of the framework with Xcode 8 I get the following error message:
`error: Swift 2.3 is not supported.`

## Steps to Reproduce
- Clone the repository
- set the selected Xcode version to Xcode 8 (`xcode-select -s <path-to-xcode-8>`)
- set the swift version to legacy in Xcode 8 for the realm project
- run `sh build.sh ios-swift`

Realm version: 1.0.1
Xcode version: 8.0
iOS/OSX version: iOS 10.0 / MacOS 10.11.5
Dependency manager + version:  Carthage 0.16.2

jpsim:
Closes #3749. Changelog entry not strictly necessary, I can remove that if requested. /cc @austinzheng @tgoyne

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ x.x.x Release notes (yyyy-MM-dd)
 ### Enhancements
 
 * Improve error message when using NSNumber incorrectly in Swift models.
+* Fix compilation with Swift 2.3 using Xcode 8 beta 2.
 
 ### Bugfixes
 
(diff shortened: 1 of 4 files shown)


tgoyne:
👍

austinzheng:
👍

PR #3883, opened 2016-07-18
[DynamicObjects] Fix to fail iterating Results of DynamicObject

kishikawakatsumi:
The following code will crash because cannot cast `Results<T>` to `Results<DynamicObject>`

```
let objects = try! Realm().dynamicObjects("SwiftObject")
for object in objects {
```

This PR to fix that to use `unsafeBitCast()` instead force cast.

Related https://github.com/realm/realm-cocoa/pull/2848

CC @bdash @mrackwitz

### DIFF
--- a/RealmSwift/RealmCollection.swift
+++ b/RealmSwift/RealmCollection.swift
@@ -34,7 +34,7 @@ public final class RLMIterator<T: Object>: IteratorProtocol {
 
     /// Advance to the next element and return it, or `nil` if no next element exists.
     public func next() -> T? { // swiftlint:disable:this valid_docs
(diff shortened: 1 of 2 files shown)


jpsim:
👍

jpsim:
@kishikawakatsumi I've tried this again and the `unsafeBitCast()` fails with Xcode 8.0 (I haven't tested with other versions). I've put this `S:On Hold` until you can update and fix. Thanks!

kishikawakatsumi:
@jpsim I've updated this PR to work on Xcode 8.x. Could you review again?

jpsim:
Thanks!

PR #3884, opened 2016-07-18
[DynamicObject] Fix to fail retrieving `List` from `DynamicObject`

kishikawakatsumi:
The following code will crash because cannot cast `List<T>` to `List<DynamicObject>`

```
let array = object["array"] as! List<DynamicObject>
```

results.rlmResults.objectSchema.accessorClass = DynamicObject.self
This PR to fix that to assign objectSchema of `Results` `accessorClass` to `DynamicObject.self`

Also subscripting, `first()`, and `last()` methods of `List<DynamicObject>` will fail with same cast error. This will be fixed to use unsafeBitCast() instead force cast.

Related #2848

CC @bdash @mrackwitz

### DIFF
--- a/RealmSwift/List.swift
+++ b/RealmSwift/List.swift
@@ -130,10 +130,10 @@ public final class List<T: Object>: ListBase {
     }
 
     /// Returns the first object in the list, or `nil` if the list is empty.
-    public var first: T? { return _rlmArray.firstObject() as! T? }
+    public var first: T? { return unsafeBitCast(_rlmArray.firstObject(), to: Optional<T>.self) }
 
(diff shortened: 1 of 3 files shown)


jpsim:
👍

jpsim:
@kishikawakatsumi I've tried this again and the `unsafeBitCast()` fails with Xcode 8.0 (I haven't tested with other versions). I've put this `S:On Hold` until you can update and fix. Thanks!

kishikawakatsumi:
@jpsim I've updated this PR to work on Xcode 8.x. Could you review this again?

jpsim:
Thanks!

PR #3886, opened 2016-07-18
Bring in recent changes from the object store

bdash:
This brings `Results` and `List` in sync with their upstream counterparts.


tgoyne:
👍

PR #3890, opened 2016-07-19
Pull in the object store code as a submodule

bdash:
We previously had included a copy of the code, but that has led to divergence and painful merging.

This is currently a WIP for testing that the approach works with CocoaPods and Carthage.

Remaining work:
- [x] Get CocoaPods working.
- [x] Get Carthage working.
- [x] Get CI working.


bdash:
CocoaPods to work as follows with submodules:
* When the Podfile specifies a released version of a pod, the submodules flag in podspecs `source` hash determines whether submodules will be checked out.
* When the Podfile specifies a git URL for a pod, the submodules flag on the pod in the Podfile determines whether submodules will be checked out.

This means that users that are building Realm with CocoaPods targeting a git branch or tag rather than a release will need to update their Podfile:
```diff
-  pod 'Realm', git: 'git@github.com:realm/realm-cocoa.git', :branch => "foo"
+  pod 'Realm', git: 'git@github.com:realm/realm-cocoa.git', :branch => "some-branch", :submodules => true
```

tgoyne:
I suppose we could init the submodule from the prepare command to make it just work automatically, but maybe there's some reason that cocoapods doesn't do that automatically?

bdash:
I'm not sure we should go out of our way when CocoaPods has a supported way to work with submodules. I've added some checking to the prepare command that instructs the user on how to update their Podfile instead.

bdash:
Use of submodules appears to be [incompatible with CocoaPods versions older than v1.0.1](https://github.com/CocoaPods/CocoaPods/issues/5539) if a recent version of Git is used (> v2.9).

bdash:
Submodules with Git 2.9.0 appears to be broken even with CocoaPods 1.0.1. Git 2.9.2 is fine.

bdash:
This is now ready to go.

austinzheng:
As far as I can tell it looks good. 👍 

May want to wait for someone with more CocoaPods and/or git expertise to provide feedback.

tgoyne:
👍

PR #3891, opened 2016-07-19
Look for the GitHub PR's source branch in a different environment variable

bdash:
We're now using a different GitHub plug-in for Jenkins and it exports different environment variables.


bdash:
Merging as this is necessary for CI to work.

PR #3896, opened 2016-07-20
Update to support Xcode 8 beta 3

bdash:
- [x] Accommodate changes to Objective-C method importing
- [x] Avoid warnings about missing `let` within `guard` statements.
- [x] Fix / remove tests related to the special handling of `is`-prefixed properties that has been removed.
- [x] Remove specification of development team on framework targets as it's no longer necessary. The team is still specified on test and app targets.


tgoyne:
31ed8282022145fbcd4661f38561bde032d2bec0 should probably be reverted entirely since the code change was more of a workaround than a change to make the introspection more correct.

bdash:
Ok, I can do that instead.

jpsim:
👍

PR #3897, opened 2016-07-20
use Yaml Axis Jenkins Plugin

jpsim:
this will allow us to:
- keep more Jenkins configuration in SCM
- use a nicer syntax to specify the matrix & exclusions
- test job changes in an isolated PR first before applying globally

unfortunately, the plugin doesn't support touchstone builds, but now that we have significantly more CI machines, I think it's a worthwhile tradeoff.

/cc @bdash @tgoyne

### DIFF
--- a/.jenkins.yml
+++ b/.jenkins.yml
@@ -1,0 +1,37 @@
+# Yaml Axis Plugin
+# https://wiki.jenkins-ci.org/display/JENKINS/Yaml+Axis+Plugin
+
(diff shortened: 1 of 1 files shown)


bdash:
> unfortunately, the plugin doesn't support touchstone builds, but now that we have significantly more CI machines, I think it's a worthwhile tradeoff.

Seems ok to me.

bdash:
👍

PR #3898, opened 2016-07-20
Improvements to iOS simulator testing

bdash:
This removes a workaround for an Xcode bug that's since been fixed, and makes the 32-bit simulator tests compatible with Xcode 8.

I've tested these locally with great success. Hopefully CI will also approve of these changes!


bdash:
This doesn't appear to have broken anything. Moving to Review.

tgoyne:
I guess not switching simulators is a good way to avoid issues related to switching between simulators.

👍

PR #3901, opened 2016-07-21
Get build.sh's Swift targets working with Swift 2.3

bdash:
This gets `REALM_SWIFT_VERSION=2.3 sh build.sh ios-swift` and friends working. `REALM_SWIFT_VERSION` must be explicitly set to 2.3 for it to be used as Xcode 8 defaults to Swift 3.0.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -90,7 +90,7 @@ EOF
 
 xcode() {
     mkdir -p build/DerivedData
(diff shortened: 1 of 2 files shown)


austinzheng:
It looks good to me. Have you tried it with both Xcode 7 and 8?

bdash:
Yes.

jpsim:
also looks good to me

Issue #3822, opened 2016-07-04
NSUncaughtExceptionHandler warning when using Crashlytics


japhar81:
## Goals
Using Realm and Crashlytics in the same app

## Expected Results
Should work

## Actual Results
Code does seem to work, but throws [Crashlytics:Crash] Warning: NSUncaughtExceptionHandler is '_ZZ34RLMInstallUncaughtExceptionHandlervEN3$_08__invokeEP11NSException' on first access to Realm

## Steps to Reproduce
See code sample -- dropping that into AppDelegate throws the warning

## Code Sample
```
Fabric.with([Crashlytics.self])
let realm = try? Realm()
if let realm = realm {
    let test = realm.objects(TestEntity.self)
}
```

## Version of Realm and Tooling

Realm version: 1.0.1

Xcode version: 7.3.1

iOS/OSX version: iOS 9

Dependency manager + version: pod 1.0.0

mrackwitz:
I guess this is happening because Crashlytics checks whether the uncaught exception handler is overridden, because it is relying on that itself, but it is commonly misused for purposes where there would be less dangerous solutions. 🐉 

Realm is using this for good reasons: we need to tear down open write transactions. While we are doing that, we still ensure to call the previously configured exception handler as [you can see here](https://github.com/realm/realm-cocoa/blob/v1.0.1/Realm/RLMRealmUtil.mm#L67-L86). So Crashlytics won't loose it's ability to report any exceptions in your app.

tgoyne:
I don't think we actually need the uncaught exception handler anymore, as it was just a partial workaround for not having robust mutexes. Crashing while holding a write transaction shouldn't cause any problems now.

TimOliver:
I think this is sufficiently resolved. While Crashlytics does throw a warning, Realm makes sure it passes on previously set exception handlers, so there's nothing broken here.

I'm closing this issue now. Feel free to re-open it if you want to discuss it further.

mrackwitz:
I think we should remove the uncaught exception handler if we don't need it anymore. So let's keep track of that by this issue. Promoted to `T:Enhancement`.

PR #3914, opened 2016-07-26
stop handling uncaught exceptions

jpsim:
Addresses #3822.

This used to be necessary to work around the lack of robust mutexes, causing a write lock to be held indefinitely. This is no longer useful.

/cc @tgoyne @bdash


bdash:
👍

austinzheng:
👍 

I can't speak to whether or not this isn't actually being used for any other purpose..

PR #3934, opened 2016-08-01
Support Xcode 8 beta 4

bdash:
This is almost entirely Swift changes. The changes break down into a couple of categories:

* `NSPredicate` and `NSRegularExpression` regained their `NS` prefix.
* `FooLiteralConvertible` protocols are now `ExpressibleByFooLiteral`.
* `URL.path` is no longer nullable.
* SE-0112 has changed how `NSError` is bridged.

To accommodate SE-0112 I've changed how our Objective-C framework defines `RLMError` to add the `ns_error_domain` attribute. This results in Swift importing it per the policy described in SE-0112, removing the need to implement our own Swift wrapper for it.

### DIFF
--- a/Realm/RLMConstants.h
+++ b/Realm/RLMConstants.h
@@ -29,8 +29,10 @@
 
 #if __has_attribute(ns_error_domain)
 #define RLM_ERROR_ENUM(type, name, domain) \
-    enum name : type name; \
-    enum __attribute__((ns_error_domain(domain))) name : type
+    _Pragma("clang diagnostic push") \
+    _Pragma("clang diagnostic ignored \"-Wignored-attributes\"") \
+    NS_ENUM(type, __attribute__((ns_error_domain(domain))) name) \
+    _Pragma("clang diagnostic pop")
 #else
 #define RLM_ERROR_ENUM(type, name, domain) NS_ENUM(type, name)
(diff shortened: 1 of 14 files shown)


austinzheng:
Looks good. Just two questions:

- Do there need to be any changes to `RLMSupport.swift` to get the Realm Objective-C Swift tests running?

- When I tried making the error change (#3835), it was mentioned that losing the Swift `Error` documentation might be an issue. Is this something we are still concerned about, and if so are there any mitigating steps?

bdash:
> When I tried making the error change (#3835), it was mentioned that losing the Swift Error documentation might be an issue. Is this something we are still concerned about, and if so are there any mitigating steps?

Hmm. Type aliases do appear in the documentation, but they only include any top-level comment associated with them. We can't really get away with just linking to the documentation of `RLMError` either since it doesn't show how they're translated to Swift. Ugh.

bdash:
I removed the aliasing of `Error` to `RLMError` in 0d7e7a1  due to the concern about documentation. Performing the bridging manually requires conforming to a couple of private protocols, plus a lot of boilerplate around the error code declarations. It'd be awesome if we could find a way to avoid this going forwards.

jpsim:
:+1:

PR #3937, opened 2016-08-02
Update to ObjectStore master

tgoyne:
Should be no functional changes other than that the result of setting both `dynamic` and `customSchema` on the configuration has changed (it used to ignore `dynamic`; now it ignores the custom schema) and a few error messages have changed.


bdash:
👍

bdash:
I think you'll also want to change the branch field in `.gitmodules` to master so `git submodule update --remote` will update to the tip of object store master in the future, rather than cocoa.

Issue #3854, opened 2016-07-11
Renaming a property does weird things to the target schema


tgoyne:
Calling `renamePropertyForClass:` replaces the object schema with one read from the group, which makes it possible to access all of the other deleted properties. An example test that currently passes, but shouldn't:

    - (void)testAccessDeletedPropetyViaNewObject {
        RLMObjectSchema *schema = [RLMObjectSchema schemaForObjectClass:StringObject.class];

        schema.properties = @[[schema.properties.firstObject copyWithNewName:@"stringCol0"],
                              [schema.properties.firstObject copyWithNewName:@"stringCol1"]];
        [self createTestRealmWithSchema:@[schema] block:^(RLMRealm *realm) {
            [StringObject createInRealm:realm withValue:@[@"0", @"1"]];
        }];

        schema.properties = @[[schema.properties.firstObject copyWithNewName:@"stringCol"]];
        RLMRealmConfiguration *config = [RLMRealmConfiguration new];
        config.fileURL = RLMTestRealmURL();
        config.customSchema = [self schemaWithObjects:@[schema]];
        config.schemaVersion = 2;
        config.migrationBlock = ^(RLMMigration *migration, uint64_t) {
            // Reading stringCol0 initially fails because it's not in the target schema
            [migration enumerateObjects:StringObject.className block:^(RLMObject *, RLMObject *newObject) {
                RLMAssertThrowsWithReasonMatching(newObject[@"stringCol0"], @"Invalid property name");
            }];

            // Renaming stringCol1 (and not touching stringCol0) makes it so that
            // stringCol0 can be read
            [migration renamePropertyForClass:StringObject.className oldName:@"stringCol1" newName:@"stringCol"];
            [migration enumerateObjects:StringObject.className block:^(RLMObject *, RLMObject *newObject) {
                XCTAssertEqualObjects(newObject[@"stringCol0"], @"0");
                XCTAssertEqualObjects(newObject[@"stringCol"], @"1");
            }];
        };

        XCTAssertTrue([RLMRealm performMigrationForConfiguration:config error:nil]);
    }

This is mostly harmless, but it means that the reported target schema is not what schema the Realm will have after the migration, and it exposes some useful functionality in a *really weird* way.

[...]

PR #3941, opened 2016-08-03
Redesign table column handling to support runtime schema changes

tgoyne:
No user-visible changes here as none of the functionality is surfaced in the API yet.

The new high-level design is:

 * RLMSchema and its children no longer contains any Realm-instance specific data (no table columns, no cached table accessors, no KVO tracking), and so we no longer need to make a shallow copy of it per RLMRealm. It does still store the accessor classes, and as a result the separate accessor cache is no longer needed.
 * All of the per-RLMRealm data that we need in addition to what the ObjectStore Schema stores is tracked in a new RLMObjectInfo class, which stores the table accessors, KVO info, and provides an efficient lookup of table columns from property indexes.
 * Everything that previously stored a table column now stores a property index and looks up the table column on each use. This has a small (~1-2%) negative impact on accessor performance.

@bdash @jpsim

### DIFF
--- a/Realm/RLMAccessor.h
+++ b/Realm/RLMAccessor.h
@@ -44,11 +44,11 @@ bool RLMIsGeneratedClass(Class cls);
 // Dynamic getters/setters
 //
 FOUNDATION_EXTERN void RLMDynamicValidatedSet(RLMObjectBase *obj, NSString *propName, id __nullable val);
-FOUNDATION_EXTERN RLMProperty *RLMValidatedGetProperty(RLMObjectBase *obj, NSString *propName);
 FOUNDATION_EXTERN id __nullable RLMDynamicGet(RLMObjectBase *obj, RLMProperty *prop);
+FOUNDATION_EXTERN id __nullable RLMDynamicGetByName(RLMObjectBase *obj, NSString *propName, bool asList);
 
 // by property/column
-FOUNDATION_EXTERN void RLMDynamicSet(RLMObjectBase *obj, RLMProperty *prop, id val, RLMCreationOptions options);
+void RLMDynamicSet(RLMObjectBase *obj, RLMProperty *prop, id val, RLMCreationOptions options);
 
 //
 // Class modification
--- a/Realm/RLMAccessor.mm
+++ b/Realm/RLMAccessor.mm
@@ -214,7 +214,7 @@ static inline RLMObjectBase *RLMGetLink(__unsafe_unretained RLMObjectBase *const
         return nil;
     }
     NSUInteger index = obj->_row.get_link(col);
-    return RLMCreateObjectAccessor(obj->_realm, obj->_realm->_info[obj->_objectSchema.properties[colIndex].objectClassName], index);
+    return RLMCreateObjectAccessor(obj->_realm, obj->_info->linkTargetType(colIndex), index);
 }
 
 static inline void RLMSetValue(__unsafe_unretained RLMObjectBase *const obj, NSUInteger colIndex,
--- a/Realm/RLMAccessor.mm
+++ b/Realm/RLMAccessor.mm
@@ -19,10 +19,11 @@
 #import "RLMAccessor.h"
 
 #import "RLMArray_Private.hpp"
-#import "RLMObservation.hpp"
+#import "RLMListBase.h"
 #import "RLMObjectSchema_Private.hpp"
 #import "RLMObjectStore.h"
 #import "RLMObject_Private.hpp"
+#import "RLMObservation.hpp"
 #import "RLMProperty_Private.h"
 #import "RLMRealm_Private.hpp"
 #import "RLMResults_Private.h"
@@ -869,15 +870,7 @@ void RLMDynamicSet(__unsafe_unretained RLMObjectBase *const obj, __unsafe_unreta
     });
 }
 
-RLMProperty *RLMValidatedGetProperty(__unsafe_unretained RLMObjectBase *const obj, __unsafe_unretained NSString *const propName) {
-    RLMProperty *prop = obj->_objectSchema[propName];
-    if (!prop) {
-        @throw RLMException(@"Invalid property name '%@' for class '%@'.", propName, obj->_objectSchema.className);
-    }
-    return prop;
-}
-
-id RLMDynamicGet(__unsafe_unretained RLMObjectBase *obj, __unsafe_unretained RLMProperty *prop) {
+id RLMDynamicGet(__unsafe_unretained RLMObjectBase *const obj, __unsafe_unretained RLMProperty *const prop) {
     NSUInteger col = prop.column;
     switch (accessorCodeForType(prop.objcType, prop.type)) {
         case RLMAccessorCodeByte:         return @((char)RLMGetLong(obj, col));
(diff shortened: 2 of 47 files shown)


RealmSwift/List.swift
```diff
@@ -626,6 +626,10 @@ public final class List<T: Object>: ListBase {
         super.init(array: RLMArray(objectClassName: (T.self as Object.Type).className()))
     }
 
+    internal init(rlmArray: RLMArray) {
+        super.init(array: rlmArray)
```
bdash: Does this change need to be made in both Swift 2.x and 3.0 portions of the file?

Realm/Tests/MigrationTests.mm
```diff
@@ -572,7 +575,7 @@ - (void)testMismatchedIndexAllowedForReadOnly {
     config.readOnly = true;
     RLMRealm *realm = [RLMRealm realmWithConfiguration:config error:nil];
     objectSchema = realm.schema[@"StringObject"];
-    XCTAssertTrue(objectSchema.table->has_search_index([objectSchema.properties[0] column]));
+//    XCTAssertTrue(objectSchema.table->has_search_index([objectSchema.properties[0] column]));
```
bdash: Not sure what's going on here?

Realm/Tests/MigrationTests.mm
```diff
@@ -1381,7 +1335,7 @@ - (void)testMultipleMigrationRenameProperty {
             [migration enumerateObjects:StringObject.className block:^(RLMObject *oldObject, RLMObject *newObject) {
                 oldValue = oldObject[@"stringCol0"];
                 XCTAssertNotNil(oldValue);
-                XCTAssertEqualObjects(newObject[@"stringCol1"], oldValue);
+//                XCTAssertEqualObjects(newObject[@"stringCol1"], oldValue);
```
bdash: Not sure what's going on here?

tgoyne: https://github.com/realm/realm-cocoa/issues/3854

Realm/RLMObjectInfo.hpp
```diff
@@ -0,0 +1,102 @@
...
+private:
+    mutable realm::Table *_Nullable m_table = nullptr;
+    std::vector<RLMObjectInfo *> m_linkTargets;
+};
+
+// A per-RLMRealm object schema map which stores RLMObjectInfo keyed on the name
+class RLMSchemaInfo {
+    using impl = std::unordered_map<NSString *, RLMObjectInfo>;
+public:
+    void init(RLMRealm *realm, RLMSchema *rlmSchema, realm::Schema const& schema);
```
bdash: I assume this uses an explicit `init` method to accommodate being used as an Objective-C ivar?

Realm/RLMAccessor.mm
```diff
@@ -56,11 +58,24 @@ typedef NS_ENUM(char, RLMAccessorCode) {
     RLMAccessorCodeBoolObject,
 };
 
-// long getter/setter
-static inline long long RLMGetLong(__unsafe_unretained RLMObjectBase *const obj, NSUInteger colIndex) {
+template<typename T>
+static T get(__unsafe_unretained RLMObjectBase *const obj, NSUInteger index) {
```
JadenGeller: Why are we dropping the `RLM` prefix?

tgoyne: It's only needed for exported functions.

Realm/RLMObjectInfo.hpp
tgoyne: An incomplete refactoring from when `init` was a method on RLMRealm. Should probably just make `init` a constructor and move it into the ivar.

Realm/RLMAccessor.mm
JadenGeller: What do you mean by that? `RLMGetLong` isn't visible to users of our API as far as I understand.

tgoyne: Yes, that's why it shouldn't have the prefix.

JadenGeller: Should we also get rid of these other prefixes (in a separate PR) then?

Realm/RLMObjectInfo.hpp
```diff
@@ -0,0 +1,102 @@
...
+template<> struct equal_to<NSString *> {
+    bool operator()(__unsafe_unretained NSString * lhs, __unsafe_unretained NSString *rhs) const {
+        return [lhs isEqualToString:rhs];
+    }
+};
+}
+
+// The per-RLMRealm object schema information which stores the cached table
+// reference, handles table column lookups, and tracks observed objects
+struct RLMObjectInfo {
```
bdash: This naming kinda implies that the class stores information about a particular `RLMObject`, when it's really information about the object type / schema.

Realm/RLMObjectSchema.mm
```diff
@@ -104,7 +102,7 @@ + (instancetype)schemaForObjectClass:(Class)objectClass {
     }
     schema.className = className;
     schema.objectClass = objectClass;
-    schema.accessorClass = RLMDynamicObject.class;
+    schema.accessorClass = objectClass;
```
bdash: Any idea why this was using `RLMDynamicObject` for the accessor class before?

tgoyne: I suspect it was to ensure that an accessor class wouldn't get created for the object schema, back from when we only supported a single accessor class per name.

Realm/RLMObjectInfo.hpp
tgoyne: Any suggestions for better names? `RLMObjectSchemaInfo` is the only obvious one, but it feels kinda gross to me.

bdash: `RLMObjectSchemaInfo` and `RLMClassInfo` are the two ideas that come to mind. Neither is great.

Realm/RLMObjectStore.mm
```diff
@@ -175,28 +157,24 @@ void RLMAddObjectToRealm(__unsafe_unretained RLMObjectBase *const object,
...
+    object->_row = (*info.table())[RLMCreateOrGetRowForObject(info, primaryGetter, createOrUpdate, created)];
 
     RLMCreationOptions creationOptions = RLMCreationOptionsPromoteUnmanaged;
     if (createOrUpdate) {
         creationOptions |= RLMCreationOptionsCreateOrUpdate;
     }
 
     // populate all properties
-    for (RLMProperty *prop in schema.properties) {
+    size_t i = 0;
```
bdash: This appears to be unused?

Realm/Tests/MigrationTests.mm
bdash: Ok. I think we should either add a FIXME, or remove it completely.

bdash:
👍, modulo a few small points of feedback.

Realm/RLMAccessor.h
```diff
@@ -44,11 +44,11 @@ bool RLMIsGeneratedClass(Class cls);
 // Dynamic getters/setters
 //
 FOUNDATION_EXTERN void RLMDynamicValidatedSet(RLMObjectBase *obj, NSString *propName, id __nullable val);
-FOUNDATION_EXTERN RLMProperty *RLMValidatedGetProperty(RLMObjectBase *obj, NSString *propName);
 FOUNDATION_EXTERN id __nullable RLMDynamicGet(RLMObjectBase *obj, RLMProperty *prop);
+FOUNDATION_EXTERN id __nullable RLMDynamicGetByName(RLMObjectBase *obj, NSString *propName, bool asList);
```
raphaelcruzeiro: This broke RealmSwift (Object.swift line 252).
(Not sure if outsiders are supposed to post comments on PRs so I apologize if I'm overstepping a boundary here).

PR #3946, opened 2016-08-08
Expose public thread handover APIs

JadenGeller:
A versatile async handover API is introduced using the `Realm.exportThreadHandover(containing:)` and `ThreadHandover.importOnCurrentThread()` APIs.

``` swift
let handover = realm.exportThreadHandover(containing: [foo, bar])
queue.async {
    let (realm, objects) = try! handover.importOnCurrentThread()
    let (foo, bar) = (objects[0], objects[1])
    try! realm.write {
        // do stuff
    }
}
```

The Objective-C API is nearly identical to the Swift API except that `[RLMThreadHandover importOnCurrentThread]` returns a `RLMThreadImport` instead of a tuple.

This implementation of handover has been build on top of the newly introduced Object Store support for handover https://github.com/realm/realm-object-store/pull/116 so this PR only deals with binding-specific code.

Unlike https://github.com/realm/realm-cocoa/pull/3828, this implementation does not end the read transaction on handover import. Instead, it will advance the transaction version or the handover version such that they are compatible. Because of this, `ThreadHandover.importOnCurrentThread()` refreshes the Realm and no longer invalidates accessors.

Note that this PR leaves out any convenience API since there was too much disagreement among the team as to how it should be structured. This API will likely be added in a future PR.


.gitmodules
```diff
@@ -1,4 +1,4 @@
 [submodule "Realm/ObjectStore"]
 	path = Realm/ObjectStore
 	url = https://github.com/realm/realm-object-store.git
-	branch = master
+	branch = jg-handover
```
austinzheng: Should this be changed back to `master`?

JadenGeller: Yes, I plan to update the PR once https://github.com/realm/realm-object-store/pull/116 is merged.

CHANGELOG.md
```diff
@@ -12,6 +12,7 @@ x.x.x Release notes (yyyy-MM-dd)
 
 ### Enhancements
 
+* Introduce API for safely handing objects between threads with convenience async write functions.
```
jpsim: this note is incorrect if we're not adding the convenience API.

JadenGeller: Good catch.

Realm/RLMObjectBase.mm
```diff
@@ -445,3 +448,31 @@ + (NSArray *)requiredPropertiesForClass:(Class)cls {
...
 
 @end
+
+@interface RLMObjectBase (Handover) <RLMThreadConfined_Private>
+@end
+
+@implementation RLMObjectBase (Handover)
+
+- (realm::AnyThreadConfined)rlm_handoverData {
+    // FIXME: Once merged, should be easier to get schema ref (https://github.com/realm/realm-cocoa/pull/3941)
```
jpsim: this was merged

JadenGeller: Yup, already working on rebasing over that! It requires rather significant changes to this PR 😝

JadenGeller:
The object store handover has been merged https://github.com/realm/realm-object-store/pull/116 🎉 and this has been rebased on top of master (which now includes https://github.com/realm/realm-cocoa/pull/3941)💃, so I think it's ready for review! @jpsim @tgoyne @bdash @austinzheng

JadenGeller:
Since this introduces new user-facing APIs, please take a careful look at the doc-comments.

Realm/RLMArrayLinkView.mm
```diff
@@ -46,19 +47,32 @@ @implementation RLMArrayLinkView {
     std::unique_ptr<RLMObservationInfo> _observationInfo;
 }
 
-- (RLMArrayLinkView *)initWithParent:(__unsafe_unretained RLMObjectBase *const)parentObject
-                            property:(__unsafe_unretained RLMProperty *const)property {
+- (RLMArrayLinkView *)initWithList:(realm::List)list
+                             realm:(__unsafe_unretained RLMRealm * const)realm
```
austinzheng: No space after `*`.

Realm/RLMObjectBase.mm
bdash: Have you completed this rebase? Should this FIXME go away?

Realm/RLMObjectBase.mm
```diff
@@ -441,3 +444,30 @@ + (NSArray *)requiredPropertiesForClass:(Class)cls {
...
+    auto className = RLMStringDataWithNSString(self.objectSchema.className);
+    auto& schema = _realm->_realm->schema();
+    auto objectSchema = schema.find(std::move(className));
+    REALM_ASSERT(objectSchema != schema.end());
+    return AnyThreadConfined(Object(_realm->_realm, *objectSchema, _row));
+}
+
+- (NSNull *)rlm_handoverMetadata {
+    return [NSNull null];
```
bdash: Presumably this is `NSNull` because the caller is going to stuff it into a collection of some sort. Seems a bit weird for this method to care about that.

Realm/RLMResults.mm
```diff
@@ -427,3 +428,25 @@ - (BOOL)isAttached
...
+@end
+
+@implementation RLMResults (Handover)
+
+- (realm::AnyThreadConfined)rlm_handoverData {
+    return translateErrors([&] { return AnyThreadConfined(_results); }); // TODO: Do we need more translate errors thing?
+}
+
+- (NSString *)rlm_handoverMetadata {
+    return self.objectSchema.className;
```
bdash: Why is passing this extra data necessary? `Results::get_object_type` on the handed-over instance should give you the same information.

RealmSwift/LinkingObjects.swift
```diff
@@ -469,7 +509,7 @@ public class LinkingObjectsBase: NSObject, NSFastEnumeration {
     internal let objectClassName: String
     internal let propertyName: String
 
-    private var cachedRLMResults: RLMResults?
+    fileprivate var cachedRLMResults: RLMResults?
```
bdash: This looks suspicious. `fileprivate` is new in Swift 3.

Realm.xcodeproj/project.pbxproj
```diff
@@ -787,8 +799,8 @@
 				3FE556431B9A43E5002A1129 /* schema.hpp */,
 				3FAE25531B8CEBBE00D01405 /* shared_realm.cpp */,
 				3FAE25541B8CEBBE00D01405 /* shared_realm.hpp */,
-				3F6864EA1D5B8272000024C3 /* thread_confined.cpp */,
-				3F6864EB1D5B8272000024C3 /* thread_confined.hpp */,
+				D4CD5ED01D594A0400FA8415 /* thread_confined.cpp */,
+				D4CD5ED11D594A0400FA8415 /* thread_confined.hpp */,
```
bdash: Weird.

Realm/RLMResults.mm
```diff
@@ -427,3 +428,25 @@ - (BOOL)isAttached
...
 @implementation RLMLinkingObjects
 @end
+
+@interface RLMResults (Handover) <RLMThreadConfined_Private>
+@end
+
+@implementation RLMResults (Handover)
+
+- (realm::AnyThreadConfined)rlm_handoverData {
+    return translateErrors([&] { return AnyThreadConfined(_results); }); // TODO: Do we need more translate errors thing?
```
bdash: Did you arrive at any conclusions related to this TODO?

Realm/Tests/ConformanceTests.mm
```diff
@@ -0,0 +1,60 @@
+////////////////////////////////////////////////////////////////////////////
+//
+// Copyright 2015 Realm Inc.
```
bdash: 2016

Realm/Tests/RLMTestCase.m
```diff
@@ -198,11 +198,15 @@ - (void)waitForNotification:(NSString *)expectedNote realm:(RLMRealm *)realm blo
...
-- (void)dispatchAsync:(dispatch_block_t)block {
+- (dispatch_queue_t)bgQueue {
     if (!_bgQueue) {
         _bgQueue = dispatch_queue_create("test background queue", 0);
     }
-    dispatch_async(_bgQueue, ^{
+    return _bgQueue;
+}
+
+- (void)dispatchAsync:(dispatch_block_t)block {
```
bdash: What's this change about?

Realm/Tests/RealmTests.mm
```diff
@@ -1,3 +1,4 @@
+
```
bdash: Looks bogus.

RealmSwift/LinkingObjects.swift
```diff
@@ -422,6 +429,39 @@ extension LinkingObjects : RealmCollection {
     }
 }
 
+private struct LinkingObjectsHandoverMetadata {
+    var propertyName: String
+    var property: RLMProperty?
```
bdash: If I'm understanding this correctly, `property` is only relevant if the `RLMResults` instance that backs the `LinkingObjects` has yet to be populated, and in that case it must be non-nil. Using an enum rather than a struct for this metadata type may allow us to communicate that relationship more clearly.

Realm/RLMObjectBase.mm
JadenGeller: Yes, good catch.

Realm/RLMObjectBase.mm
JadenGeller: It doesn't seem that odd to me. The type of `rlm_handoverMetadata` is `() -> Any`, so this function must return a non-nil value. I could change the protocol such that this return value is nullable, but that clearly better to me.

Realm/RLMResults.mm
JadenGeller: I actually forgot about this. Will fix.

RealmSwift/LinkingObjects.swift
JadenGeller: Not exactly. `private` is new in Swift 3, and Swift 2's `private` was renamed `fileprivate`.

bdash: `fileprivate` as a keyword is new in Swift 3. This code is in the Swift 2.2 `#else` block. That's what's suspicious about it 😛

jpsim: wat. Seriously? I'm fairly certain this isn't the case. That Swift 2.3 is _nearly_ the same compiler as Swift 2.2, but with accommodations for changes to Apple SDKs.

I think what you actually mean is that Swift 3's `fileprivate` is equivalent to Swift 2's `private`.

bdash: Note that [CI is complaining about the code as-is](https://ci.realm.io/job/objc_pr/configuration=Debug,swift_version=2.2,target=osx-swift/3865/console):
```
/Users/realm/workspace/objc_pr/configuration/Debug/swift_version/2.2/target/osx-swift/RealmSwift/LinkingObjects.swift:512:5: error: expected declaration
    fileprivate var cachedRLMResults: RLMResults?
    ^
/Users/realm/workspace/objc_pr/configuration/Debug/swift_version/2.2/target/osx-swift/RealmSwift/LinkingObjects.swift:512:16: error: consecutive declarations on a line must be separated by ';'
    fileprivate var cachedRLMResults: RLMResults?
               ^
               ;
```

Realm/Tests/RLMTestCase.m
JadenGeller: I guess this isn't really relevant to this PR, so I'll remove it. It got added to make the the Objective-C and Swift testing classes more consistent.

RealmSwift/LinkingObjects.swift
JadenGeller: It would, great point. I must've been scared away from tagged unions after the C++ changes…

RealmSwift/LinkingObjects.swift
JadenGeller: @jpsim Yes, that's what I meant. I meant "renamed" from Swift 2 to Swift 3, not that they changed it in Swift 2.

JadenGeller: @bdash Oh. Well then, I'll fix that…

RealmSwift/LinkingObjects.swift
JadenGeller: Thinking about this more, it's not clear to me that an `enum` is better. I think it would be better if `RLMObjectBase` also used and enum to represent this state, but it uses optional values that mostly mirror these. If we used an enum with a `cached` and `uncached` case, would we check if `property` exists or if `cachedRLMResults` exists when deciding which case to use? (They ought to give the same results, but this is a weird check.) After rewriting it with an enum, I think that the complexity it add is not worth it since it doesn't align with `LinkingObjectsBase`.

Realm/Tests/RLMTestCase.m
JadenGeller: Never mind, this is relevant since it fixes a potential bug in `dispatchAsyncAndWait:`, which I use in the tests. If `dispatchAsync:` is not called before `dispatchAsyncAndWait:`, then `_bgQueue` will not exist. This fixes that by making both call the property to lazily create the queue.

bdash: But `-dispatchAsyncAndWait:` calls `-dispatchAsync:` internally, ensuring that `_bgQueue` is initialized?

JadenGeller: Good point. I'm not sure why I made these changes anymore then. I'll revert this file.

JadenGeller:
@bdash This is ready for re-review whenever you're ready.

.gitmodules
bdash: This still needs to be addressed.

Realm/RLMHandover.h
```diff
@@ -0,0 +1,67 @@
...
+@property (nonatomic, readonly, nullable) RLMRealm *realm;
+
+// Runtime-enforced requirement that type also conforms to `RLMThreadConfined_Private`
+
+@end
+
+/// An object containing the data imported from handover
+@interface RLMThreadImport : NSObject
+
+/// The `RLMRealm` from which the `objects` were handed over
```
bdash: The way this is phrased it sounds like this is the `RLMRealm` from the original thread, but isn't it really the Realm they were imported into?

Realm/RLMHandover.h
```diff
@@ -0,0 +1,67 @@
...
+/// Objects equivalent to those handed over but associated with this thread's `realm`
+@property (nonatomic, readonly) NSArray<id<RLMThreadConfined>> *objects;
+
+@end
+
+/// An object intended to be passed between threads containing information about objects being handed over
+@interface RLMThreadHandover : NSObject
+
+/**
+ Imports the handover package, creating an instance of the realm and objects on the current thread.
```
bdash: s/realm/Realm/

Realm/RLMHandover.h
```diff
@@ -0,0 +1,67 @@
...
+
+ This method may be not be called more than once on a given handover package. The realm version will
```
bdash: s/realm/Realm/

Realm/RLMObjectBase.mm
bdash: `ram_handoverMetadata` only has a non-nullable return type because you chose to mark it as such. IMO it's clearer to mark the result as nullable and return `nil` than it is to claim it's non-nullable and just use a different value for null. `NSNull` is typically only used in places where `nil` is not valid.

Realm/RLMHandover.h
JadenGeller: Yes, good point.

Realm/RLMHandover.h
JadenGeller: Where? If you're talking about the `objects` comment, the `realm` mentioned in the name of the property. Though I could capitalize it to refer to the type and convey nearly the same thing.

Realm/RLMHandover.h
JadenGeller: Not sure what this one refers to. I only see it lowercase once, unless you want the property name capitalized?

bdash: "the realm version"

Realm/RLMHandover.h
bdash: "creating an instance of the realm and objects on the current thread"

Realm/RLMObjectBase.mm
JadenGeller: I guess that's reasonable. Sure.

RealmSwift/Handover.swift
```diff
@@ -0,0 +1,181 @@
...
+// limitations under the License.
+//
+////////////////////////////////////////////////////////////////////////////
+
+import Realm
+
+#if swift(>=3.0)
+    
+/// A Realm-bound object that can only be passed between threads by exporting for handover
+@objc public protocol ThreadConfined {
```
bdash: Why is this `@objc`?

Realm/RLMHandover.h
JadenGeller: Ah I see. Th last like of the diff above is the one commented on. Didn't know GitHub did it like that. Odd.

RealmSwift/Handover.swift
JadenGeller: So the existential `ThreadConfined` conforms to `ThreadConfined`, allowing usage such as `[foo, bar, baz] as [ThreadConfined]` for non-heterogenous types.

bdash: Huh. Any idea why `@objc` is required for that?

JadenGeller: Due to a current limitation in Swift's type system, Swift existentials do not conform to their protocol. They plan to resolve this eventually. I can search for an explanation from Swift Evolution if you'd like.

bdash: If you can easily find an explanation it could be worth linking to it from an explanatory comment here. Not a big deal if you can't, since a future reader will hopefully be able to track down this PR anyway.

JadenGeller: From a Swift Evolution [thread](https://lists.swift.org/pipermail/swift-evolution/Week-of-Mon-20160815/026366.html),

> The implementation for self-conforming opaque protocols is not difficult, because the value itself can already be of any size, so it’s really not a problem to have an existential in there. In theory, someone could cook it up in a week or so.
> 
> For class protocols, I don’t know how to do it without an efficiency hit unfortunately.
>
> Consider these two functions, taking a homogeneous and heterogeneous array of a class-bound protocol type:
> 
> protocol P : class {}
> 
> func f<T : P>(array: [T]) {} // this takes an array of pointers to T, because there’s only one witness table for all of them
> func ff(array: [P]) {} // this takes an array of <T, witness table> pairs, two pointers each, because each element can be a different concrete type
> 
> What you’re saying is that f() should in fact allow both representations, because you’ll be able to call f() with a value of type [P]. Right now, if we know a generic parameter is class-constrained, we use a much more efficient representation for values of that type, that is known to be fixed size in the LLVM IR. We would have to give that up to allow class-constrained existentials to self-conform, since now a class-constrained parameter can be an existential with any number of witness tables.
> 
> There might be some trick for doing this efficiently, but I don’t know of one yet.
> 
> Of course, we can just say that class-constrained protocols never self-conform, unless they’re @objc. That seems like a hell of an esoteric restriction though (can you imagine trying to come up with a clear phrasing for *that* diagnostic?)
> 
> And if you’re wondering, the reason that @objc protocols self-conform in Swift today, is because they their existentials don’t have *any* witness tables — @objc protocol method bodies are found by looking inside the instance itself.
> 
> AnyObject is the other kind of protocol that self-conforms — you can use it both as a generic constraint, and as a concrete type bound to a generic parameter, and it ‘just works’, because again it doesn’t have a witness table.

-Slava Pestov

RealmSwift/Handover.swift
```diff
@@ -0,0 +1,181 @@
...
+    private let package: RLMThreadHandover
+
+    internal init(realm: Realm, objects: [T]) {
+        self.metadata = objects.map { $0._private.bridgedMetadata }
+        self.types = objects.map { $0.dynamicType }
+        self.package = realm.rlmRealm.exportThreadHandover(containing: objects.map { $0._private.bridgedData })
+    }
+
+    /**
+     Imports the handover package, creating an instance of the realm and objects on the current thread.
```
bdash: "creating an instance of the ***r***ealm and objects"

RealmSwift/Handover.swift
```diff
@@ -0,0 +1,181 @@
...
+
+     - warning: This method may be not be called more than once on a given handover package. The realm version will
```
bdash: "The ***r***ealm version"

RealmSwift/Handover.swift
JadenGeller: I could add a FIXME to the codebase explaining this so that it can be removed once Swift has the necessary features. (Some tests will not compile with `@objc` removed, so there's no danger in prematurely removing it.)

RealmSwift/List.swift
```diff
@@ -69,6 +70,10 @@ public final class List<T: Object>: ListBase {
 
     // MARK: Initializers
 
+    fileprivate override init!(array: RLMArray<RLMObject>!) {
```
bdash: I guess the IUO is only due to some of our internal headers lacking nullability annotations?

bdash:
👍

RealmSwift/List.swift
JadenGeller: Yes.

Realm/RLMArray.mm
```diff
@@ -437,3 +438,23 @@ - (instancetype)reversedSortDescriptor {
...
 
 @end
+
+@interface RLMArray (Handover) <RLMThreadConfined_Private>
+@end
+
+@implementation RLMArray (Handover)
+
+- (realm::AnyThreadConfined)rlm_handoverData {
+    @throw RLMException(@"Unexpected handover of unmanaged `RLMArray`");
```
tgoyne: Does this happen due to a bug in our code, or due to API misuse? If it's the former it should be an assertion, and if it's the latter the error message needs to do a better job of explaining to the user what they did wrong.

Realm/RLMCollection.h
```diff
@@ -18,6 +18,9 @@
 
 #import <Foundation/Foundation.h>
 
+#import <Realm/RLMRealm.h>
```
tgoyne: This doesn't appear to need RLMRealm.h

Realm/RLMHandover.h
```diff
@@ -0,0 +1,67 @@
...
+#import <Foundation/Foundation.h>
+
+@class RLMRealm;
+
+NS_ASSUME_NONNULL_BEGIN
+
+/// A Realm-bound object that can only be passed between threads by exporting for handover
+@protocol RLMThreadConfined <NSObject>
+
+/// The destination `RLMRealm` the object was imported into
```
tgoyne: This is misleading, as it suggests that it would be `nil` for objects which have not been handed over.

Realm/RLMHandover.h
```diff
@@ -0,0 +1,67 @@
...
+//
+////////////////////////////////////////////////////////////////////////////
+
+#import <Foundation/Foundation.h>
+
+@class RLMRealm;
+
+NS_ASSUME_NONNULL_BEGIN
+
+/// A Realm-bound object that can only be passed between threads by exporting for handover
```
tgoyne: An object which is bound to a thread-specific `RLMRealm` instance, and so cannot be passed between threads.

Objects of classes conforming to this protocol can be packaged for transport between threads by calling `-[RLMRealm exportThreadHandoverWithObjects:]`. Note that only types defined by Realm can meaningfully conform to this protocol, and defining new classes which attempt to conform to it will not make them work with `exportThreadHandoverWithObjects`.

Realm/RLMRealm.h
```diff
@@ -18,8 +18,9 @@
 
 #import <Foundation/Foundation.h>
 #import "RLMConstants.h"
+#import "RLMHandover.h"
 
-@class RLMRealmConfiguration, RLMObject, RLMSchema, RLMMigration, RLMNotificationToken;
+@class RLMRealmConfiguration, RLMObject, RLMSchema, RLMMigration, RLMNotificationToken, RLMRealm;
```
tgoyne: ???

Realm/Tests/ConformanceTests.mm
```diff
@@ -0,0 +1,60 @@
...
+    [super setUp];
+
+    _classList = objc_copyClassList(&_count);
+}
+
+- (void)tearDown {
+    free(_classList);
+}
+
+- (void)testThreadConfinedPrivateConformance {
```
tgoyne: What exactly is this testing? That we haven't accidentally declared any extra classes as conforming to RLMThreadConfined? That doesn't really seem worth checking.

Realm/RLMHandover.mm
```diff
@@ -0,0 +1,155 @@
...
+                if ([object conformsToProtocol: @protocol(RLMThreadConfined)]) {
+                    @throw RLMException(@"Illegal custom conformance to `RLMThreadConfined` by `%@`", [object class]);
+                }
+                else {
+                    @throw RLMException(@"Unexpected `%@` in array of expected `RLMThreadConfined` objects", [object class]);
+                }
+            }
+            if (realm != object.realm) {
+                if (object.realm == nil) {
+                    @throw RLMException(@"Can only hand over objects that are mangaged by a Realm");
```
tgoyne: This should mention that unmanaged objects don't need to be handed over.

Realm/RLMRealm.h
```diff
@@ -495,6 +496,22 @@ __deprecated_msg("Use `performMigrationForConfiguration:error:`") NS_REFINED_FOR
...
+
+ @warning Note that the returned `ThreadHandover` must be imported at most once, and that the current version of the
+ Realm will remain pinned until this package is imported or deallocated.
+
+ @param objects The objects to pass between threads.
+ @return A package that can be imported on a separate thread.
+
+ @see RLMThreadHandover
+ */
+- (RLMThreadHandover *)exportThreadHandoverWithObjects:(NSArray<id<RLMThreadConfined>> *)objects
```
tgoyne: This name doesn't fit the Cocoa naming scheme; that would be "threadHandoverWithObjects" or "exportThreadHandoverForObjects". I also think we should try to avoid using the word "handover" in the public API. While the concept of handing objects over to other threads is generally understood, the compound noun form of it is basically internal jargon, and perhaps more importantly what we offer *isn't* transferring an existing object to a different thread. Another issue with this name is that it suggests that it's more restricted than it actually is. The other places in our API where we refer to "Objects" we specifically mean RLMObject/RealmSwift.Objects (e.g. `addObjects:`).

Realm/RLMHandover.h
JadenGeller: This comment is actually entirely incorrect. I modified the wrong one in response to @bdash's comment. Thanks for catching that.

Realm/Tests/ConformanceTests.mm
JadenGeller: It ensures that all classes that conform to `RLMThreadConfined` also conform to `RLMThreadConfined_Private`—why isn't this worth checking?

Realm/RLMRealm.h
```diff
@@ -495,6 +496,22 @@ __deprecated_msg("Use `performMigrationForConfiguration:error:`") NS_REFINED_FOR
  */
 + (BOOL)performMigrationForConfiguration:(RLMRealmConfiguration *)configuration error:(NSError **)error;
 
+/**
+ Creates a `RLMThreadHandover` from the passed in objects that can be safely imported on a separate thread.
+
+ Used to pass objects between threads in a thread-safe manner.
```
tgoyne: This needs to say a lot more about what it does, how it's used, and what you can pass in to it.

Realm/Tests/ConformanceTests.mm
tgoyne: The interesting case to be testing is that everything fails gracefully if we are handed an object with an illegal conformance, not that we personally haven't declared any such things.

Realm/RLMRealm.h
JadenGeller: That's reasonable. What do you suggest we use instead, @tgoyne?

`RLMThreadHandover` could instead be called `RLMThreadExport`, but I don't know how to drop "objects" from the method name (since things like `arrayWithObjects:` also uses that same jargon). Instead of `threadExportWithObject:`, we could do `threadExportWithArray:` or `threadExportPackaging:`, but it's not clear that's better.

tgoyne: I don't have any great ideas. I think it ideally should contain the word "thread" for the sake of people ctrl-fing the docs, should somehow indicate that it gives new references to the objects, should indicate that you can pass multiple types of things into it.

Some assorted bad ones that may inspire something:

  * `threadSafeCopy:`
  * `threadSafePackagedCopy:`
  * `threadSafePackage:`
  * `unThreadConfinedCopy:` (ugh...)
  * `exportThreadSafeReferences:`

Realm/RLMArray.mm
JadenGeller: I believe that an `RLMArray` property of a currently unmanned object is of this type rather than the `RLMArrayLinkView` subclass. I guess the handover code should first notice that `realm` is `nil`, and thus not call this code, so it would be a bug in our code. I'll replace with assertions.

JadenGeller: Well, I would like to call a no-return function here. Any suggestions if not throwing an exception? @tgoyne

tgoyne: `REALM_UNREACHABLE();` calls terminate with an appropriate message and should be marked as noreturn.

Realm/Tests/ConformanceTests.mm
JadenGeller: It seems to me that both are useful to do. I don't think I do the former, so it might be worth adding a test case for this.

JadenGeller: Essentially, this test dynamically verifies that `RLMThreadConfined` implies `RLMThreadConfined_Private` in our codebase since Objective-C's type system is not capable of doing so.

Realm/RLMArray.mm
JadenGeller: Could I define a macro `REALM_ASSERTION_FAILURE()` that unconditionally calls `REALM_UNREACHABLE()` after printing a message? It would seem odd to simply call unreachable here.

Realm/Tests/ConformanceTests.mm
JadenGeller: Apparently these macros are defined in Core, so I guess I'll hold off.

tgoyne: > unless we wrote handover test cases that utilize it

"We won't catch this bug unless we write even the most minimal test of newly added functionality" is not something I consider a meaningful problem.

Realm/RLMHandover.mm
JadenGeller: "Illegal thread export of unmanaged object, which doesn't require export"

Realm/RLMRealm.h
JadenGeller: More variations:
* `exportThreadUnconfinedCopies:`
* `exportThreadPassableCopies:`
* `exportThreadTransferableCopies:`
A problem with these, IMO, is that it seems to imply that the argument to the function is thread-passable, while it is actually the return value that is.

jpsim: Here are alternative names for the `ThreadHandover` type:

* Thread{Box,Wrapper,Package,Container}
* CrossThread{Box,Wrapper,Package,Container}
* ThreadTransfer{Box,Wrapper,Package,Container}
* ThreadSafe{Box,Wrapper,Package,Container}

And instead of export/import methods, you'd just have an `init(items:)` initializer and an `importItems()`/`unboxItems()` throwing function.

For example:

```swift
class ThreadBox<T: ThreadConfined> {
  init(items: [T])
  func unboxItems() -> throws [T]
}
```

JadenGeller: Completely ignoring naming altogether, we have quite a few choices for how we want to structure this API.

## Export
### Package Initializer
```swift
let package = ThreadSafePackage(exporting: [foo, bar], from: realm)
```
```swift
let package = ThreadSafePackage(items: [foo, bar]) // infer Realm from items
```

### Realm Method
```swift
let package = realm.exportThreadSafePackage(containing: [foo, bar])
```
```swift
let package = realm.export(items: [foo, bar])
```
```swift
let package = realm.package(items: [foo, bar])
```

## Import
### Throwing Method Returning Named Realm-Items Tuple
```swift
let (realm, items) = try! package.importOnCurrentThread() // or `import()` or `unpack()`
try! realm.write {
    for x in items {
        realm.delete(x)
    }
}
```

### Auto-Importing Convenience Properties
```swift
// power user feature for handling errors
// otherwise, implicitly binds to current thread on first access, crashing on error
try! package.bindToCurrentThread()

try! package.realm.write {
    for x in package.items {
        package.realm.delete(x)
    }
}
```

### Realm Initializer
```swift
// implicitly binds package to current thread on init, allowing handling of failure
let realm = try! Realm(from: package)
try! realm.write {
    for x in package.items {
        realm.delete(x)
    }
}
```

JadenGeller: I personally think that specifying the `Realm` on export makes it more clear that it is an active operation that must be done on the source thread. It also allows for more clear diagnostics when the user passes multiple objects associated with different realms. Finally, all enables passing zero objects (just opening the same `Realm` on another thread).

I also feel that the throwing, tuple-returning import has the most straightforward control flow. It avoid implicit state changes and always allows error handling. Further, it makes it easy to work with the `Realm` and items since they're declared as local variables rather than properties on a struct.

JadenGeller: As a side note, `ThreadBox`/`ThreadPackage`/etc. all sound much weirder than `ThreadHandover` in the uncommon case of handing over zero items.

JadenGeller: One particularly nice feature about using a `Realm` initializer for import is that, as handover currently works, the only way it can fail is failure to open the `Realm`. The actual object import does not need to throw.

That said, I still don't feel great about that solution. Unlike the tuple-returning method, it doesn't enforce that `items` is not accessed before import. Also, we might introduce new ways this can fail in the future.

JadenGeller: I discussed the API over DM with @bdash. Here are some takeaways:

1) There is no need for the `ThreadHandover` to contain a `Realm.Configuration` and no reason for the import function to produce a `Realm`. The API is mostly designed as such for historical reasons (`realm.asyncWrite { realm, objects in … }`) where the `Realm` had to be created on the new thread. Given that we're making the API "less convenient", it might make sense to no longer pass the configuration. Anyway, in the common case, creating a `Realm` with the proper configuration is as simple as `let realm = try! Realm()`.
```swift
let package = realm.exportThreadSafePackage(containing: [foo, bar, baz])
queue.async {
    let realm = try! Realm()
    let items = realm.importOnCurrentThread(package)
}
```

2) If we _really_ want to rethink this API, we might structure it so that items can be exported and imported individually. We could probably optimize it so that it doesn't require pinning and unpinning the version for each item, but it might be tricky. We would however gain some type safety and it might be easier for users to think about the API this way. This implementation would not rely much on the current Object Store handover implementation (which was built for opaque handover packages), but it might still be worthwhile.
```swift
let fooRef = foo.threadSafeReference
let barRef = bar.threadSafeReference
queue.async {
    let realm = try! Realm()
    let foo = realm.threadSafeLookup(of: fooRef)
    let bar = realm.threadSafeLookup(of: barRef)
    // do stuff
}
```
and
```swift
let references = [foo, bar, baz].map { $0.threadSafeReference }
queue.async {
    let realm = try! Realm()
    let items = references.map(realm.threadSafeLookup)
    // do stuff
}
```

jpsim: I like the API rethink, renaming `Realm.threadSafeLookup(_:)` to `ThreadSafeReference.resolve(in:)`.

```swift
let fooRef = foo.threadSafeReference
let barRef = bar.threadSafeReference
queue.async {
    let realm = try! Realm()
    let foo = fooRef.resolve(in: realm) // correct type!
    let bar = barRef.resolve(in: realm) // correct type!
    // do stuff
}
```

```swift
let references = [foo, bar, baz].map { $0.threadSafeReference }
queue.async {
    let realm = try! Realm()
    let items = references.map { $0.resolve(in: realm) }
    // do stuff
}
```

JadenGeller: A few modifications after further discussion:
The resolve function ought to be defined on `Realm` rather than the `ThreadSafeReference` since it must refresh on lookup, and it's unprecedented to refresh a `Realm` via a method on another object. Also, the resolve function ought to return `nil` if the object no longer exists (instead of returning an invalidated accessor) since it's similar to looking something up by primary key.

```swift
let fooRef = foo.threadSafeReference
let barRef = bar.threadSafeReference
queue.async {
    let realm = try! Realm()
    let foo = realm.resolve(reference: fooRef)! // correct type!
    let bar = realm.resolve(reference: barRef)! // correct type!
    // do stuff
}
```

astigsen: How is the lifetime of the threadSafeReference managed? Since it has to hold a pin on the version, we don't want to stay along for longer than absolutely necessary. Ideally it should be released the moment the handover of the objects has completed, as you don't know how long the async process will stay around (could be a very long time if it does blocking operations).

JadenGeller: @astigsen The version will be unpinned once all thread safe references that refer to a given version are either advanced (which is done on `resolve`) or destroyed. Note that a `ThreadSafeReference` can only be resolved once given current handover constraints.

astigsen: ok :-)

JadenGeller: I've realized that the export API would better be expressed as a initializer than a property. Though the latter seems more lightweight, it's misleading. Since a reference is consumed at the time of import, we shouldn't pretend it's an actual property of the object.

```swift
let ref = ThreadSafeReference(to: foo)
```
```objc
RLMThreadSafeReference *ref = [RLMThreadSafeReference referenceToThreadConfined:foo];
```

JadenGeller: FYI, since Objective-C doesn't support generic methods, this API would have to return `id`:
```objc
[realm resolveThreadSafeReference:ref]
```

JadenGeller: If we were to use `[ref resolveInRealm:realm]`, I think we could get back that type safety, but it still feels weird for a method _not_ on `RLMRealm` to refresh the realm.

RashmiZK:
It would be great to know -  by when can we expect Async handover in Realm (Objective C)?

jpsim:
@RashmiZK I can't provide an estimate at this time, as everyone on the team is busy with other things.

jpsim:
@bdash @tgoyne I've revived this PR and addressed a number of little issues here and there but I'd love to get your eyes on this again.

JadenGeller:
@jpsim Isn't this the older version of the API? I think the one we settled on is here. https://github.com/realm/realm-cocoa/pull/4114 I must have overlooked closing this PR.

JadenGeller:
Apologies for any duplication of work this may have caused 😣

jpsim:
Closing in favor of #4114.

jpsim:
Going to keep the branch alive for a bit in case we need to refer back to this implementation for some reason.

PR #3952, opened 2016-08-10
Update to core 1.5.0

tgoyne:


jpsim:
The CocoaPods CI job complained of not being able to find `handover.hpp`: https://ci.realm.io/job/objc_pr/3832/configuration=Release,swift_version=2.2,target=cocoapods/console

It's possible `handover.hpp` and `thread_confined.hpp` need to be added to the podspec.

Same thing goes for #3946 /cc @JadenGeller

tgoyne:
The non-public headers thankfully don't need to be explicitly listed in the podspec anymore; the problem was just that it was relying on Xcode's extra-forgiving logic to let you include files anywhere in the project by just the bare name.

jpsim:
Ah yes, that makes more sense. Thanks for clarifying.

bdash:
👍

Issue #3931, opened 2016-08-01
Crash when using Results<?>.maximumValue(ofProperty:) with Int64 property


yangmeyer:
## Goals

I want to compare a file’s modification timestamp with that of a Realm model object in order to see whether the database knows about the latest version of the file.

To do this, I ask the `RealmSwift.Results` for its `maximumValue`. The fileModificationTime property is modelled as an Int64. This is the line that crashes:

`let maxTime = results.maximumValue(ofProperty: "fileModificationTime") as Int64?`

## Expected Results

To get the maximum value.

## Actual Results

A crash. Console output: `Could not cast value of type '__NSCFNumber' (0x123a4b608) to 'Swift.Int64' (0x113849298).`

```
(lldb) thread backtrace
* thread #1: tid = 0xd37b, 0x0000000117eabf06 libsystem_kernel.dylib`__pthread_kill + 10, queue = 'com.apple.main-thread', stop reason = signal SIGABRT
    frame #0: 0x0000000117eabf06 libsystem_kernel.dylib`__pthread_kill + 10
    frame #1: 0x0000000117f184ec libsystem_pthread.dylib`pthread_kill + 90
    frame #2: 0x0000000117c2b087 libsystem_c.dylib`abort + 129
    frame #3: 0x00000001137df995 libswiftCore.dylib`swift::fatalError(unsigned int, char const*, ...) + 149
    frame #4: 0x00000001137c0303 libswiftCore.dylib`swift::swift_dynamicCastFailure(void const*, char const*, void const*, char const*, char const*) + 83
    frame #5: 0x00000001137c038f libswiftCore.dylib`swift::swift_dynamicCastFailure(swift::TargetMetadata<swift::InProcess> const*, swift::TargetMetadata<swift::InProcess> const*, char const*) + 127
    frame #6: 0x00000001137c2569 libswiftCore.dylib`swift_dynamicCast + 2825
    frame #7: 0x00000001137c2eeb libswiftCore.dylib`_dynamicCastFromExistential(swift::OpaqueValue*, swift::OpaqueValue*, swift::TargetExistentialTypeMetadata<swift::InProcess> const*, swift::TargetMetadata<swift::InProcess> const*, swift::DynamicCastFlags) + 91
    frame #8: 0x00000001137c23fd libswiftCore.dylib`swift_dynamicCast + 2461
[...]

PR #3955, opened 2016-08-12
Fix crash when obtaining Int8/Int16/Int32 value from max/min/average/sum

JadenGeller:
Fixes https://github.com/realm/realm-cocoa/issues/3931 by using appropriate getter property on `NSNumber` instead of erroneously trying to force cast to the correct type (since not all integer types are bridged to `NSNumber`).

//cc @jpsim @austinzheng

### DIFF
--- a/RealmSwift/LinkingObjects.swift
+++ b/RealmSwift/LinkingObjects.swift
@@ -274,7 +274,8 @@ public final class LinkingObjects<T: Object>: LinkingObjectsBase {
        is empty.
      */
     public func minimumValue<U: MinMaxType>(ofProperty property: String) -> U? {
-        return rlmResults.min(ofProperty: property) as! U?
+        guard let value = rlmResults.min(ofProperty: property) else { return nil }
+        return U.bridging(value)
     }
 
     /**
@@ -288,7 +289,8 @@ public final class LinkingObjects<T: Object>: LinkingObjectsBase {
        is empty.
      */
     public func maximumValue<U: MinMaxType>(ofProperty property: String) -> U? {
-        return rlmResults.max(ofProperty: property) as! U?
+        guard let value = rlmResults.max(ofProperty: property) else { return nil }
+        return U.bridging(value)
     }
 
     /**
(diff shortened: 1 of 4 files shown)


Realm.xcodeproj/project.pbxproj
```diff
@@ -1946,7 +1946,6 @@
 				E81A1FE11955FE0100FDED82 /* ObjectInterfaceTests.m in Sources */,
 				021A88361AAFB5CD00EEAC84 /* ObjectSchemaTests.m in Sources */,
 				E81A1FE31955FE0100FDED82 /* ObjectTests.m in Sources */,
-				3F13FA701D343DE100201E9B /* PerformanceTests.m in Sources */,
```
bdash: ?

JadenGeller: Uhhh not sure what happened there, good catch.

RealmSwift/LinkingObjects.swift
```diff
@@ -274,7 +274,8 @@ public final class LinkingObjects<T: Object>: LinkingObjectsBase {
        is empty.
      */
     public func minimumValue<U: MinMaxType>(ofProperty property: String) -> U? {
-        return rlmResults.min(ofProperty: property) as! U?
+        guard let value = rlmResults.min(ofProperty: property) else { return nil }
```
jpsim: Probably a matter of preference, but rather than all the guard's, else's and dual returns, this pattern could be replaced by `map` throughout:

```swift
return rlmResults.min(ofProperty: property).map(U.bridging)
```

jpsim:
Looks good to me, but missing a changelog entry. 👍

RealmSwift/LinkingObjects.swift
JadenGeller: I'm not sure if I find that definitely clearer. Particularly, it's weird that in the optional cases we'll have `U.bridging` at the end but in the non-optional case (`sum`) we'll have
```swift
return U.bridging(rlmResults.sum(ofProperty: property)`
```

JadenGeller: I don't feel strongly about this though, so I'll switch to the `map` pattern since higher order functions generally are more readable.

PR #3967, opened 2016-08-16
Fix build with Swift 2.3 in Xcode 8 beta 6

bdash:
The constant `RLMRealmDidChangeNotification` is now imported differently.


austinzheng:
👍 

I think it may have been broken in previous betas as well.

jpsim:
👍

Issue #3968, opened 2016-08-17
Support Swift native `Date` and `Data` types


JadenGeller:
Requires parsing schemas that use `Date` and `Data` types. Also should ensure we have tests that use both these and their NS variants. We also want to make sure they're supported in both Realm Swift and Realm Objective-C from Swift.

jpsim:
Done in #3965 for both Realm Swift & Realm Objective-C, @JadenGeller?

JadenGeller:
Nope, I'm working on this now.

JadenGeller:
Done in https://github.com/realm/realm-cocoa/pull/3979

PR #3979, opened 2016-08-18
Use `Date` and `Data`

JadenGeller:
Well, https://github.com/realm/realm-cocoa/issues/3968 was much easier than expected. Literally no changes were required, so I just updated our tests to use `Date` and `Data` and updated doc comments.

### DIFF
--- a/Realm/Tests/Swift/SwiftPropertyTypeTest.swift
+++ b/Realm/Tests/Swift/SwiftPropertyTypeTest.swift
@@ -121,6 +121,25 @@ class SwiftPropertyTypeTest: RLMTestCase {
         }
         XCTAssertNotNil(succeeded, "Writing an object with an ignored lazy property should work.")
     }
+
+    func testObjectiveCTypeProperties() {
+        let realm = realmWithTestPath()
+        var object: SwiftObjectiveCTypesObject!
+        let now = NSDate()
+        let data = "fizzbuzz".data(using: .utf8)! as Data as NSData
+        try! realm.transaction {
+            object = SwiftObjectiveCTypesObject()
+            realm.add(object)
+            object.stringCol = "Hello world!"
+            object.dateCol = now
+            object.dataCol = data
+            object.numCol = 42
+        }
+        XCTAssertEqual("Hello world!", object.stringCol)
+        XCTAssertEqual(now, object.dateCol)
+        XCTAssertEqual(data, object.dataCol)
+        XCTAssertEqual(42, object.numCol)
+    }
 }
 
 #else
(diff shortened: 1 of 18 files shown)


JadenGeller:
@bdash

bdash:
Do we still have tests that ensure that `NSData` and `NSDate` work? A quick scan over the changes suggest you changed all instances to `Data` and `Date`.

JadenGeller:
I can add some. The Objective-C tests kind of ensure that though.

bdash:
The Obj-C tests don't test handling of `Object` subclasses with `NSDate` properties declared.

JadenGeller:
@bdash Tests added.

Realm/Tests/Swift/SwiftPropertyTypeTest.swift
```diff
@@ -121,6 +121,25 @@ class SwiftPropertyTypeTest: RLMTestCase {
         }
         XCTAssertNotNil(succeeded, "Writing an object with an ignored lazy property should work.")
     }
+
+    func testObjectiveCTypeProperties() {
+        let realm = realmWithTestPath()
+        var object: SwiftObjectiveCTypesObject!
+        let now = NSDate()
+        let data = "fizzbuzz".data(using: .utf8)! as Data as NSData
```
jpsim: Doesn't `String.data(using:)` already return a `Data`? Why the `as Data` here?

bdash:
@JadenGeller, the tests you added are for `NSDate` and friends on a Swift subclass of `RLMObject`. I don't see any for `NSDate` on a subclass of `Object`?

JadenGeller:
@bdash That's true. We also don't have any tests of `NSString` on a subclass of `Object`, so I followed that lead. Are `RLMObject` vs `Object` models parsed differently?

Realm/Tests/Swift/SwiftPropertyTypeTest.swift
JadenGeller: Swift bug. Should I add a FIXME everywhere we have to workaround this? https://bugs.swift.org/browse/SR-2365

jpsim: No, I suppose not.

bdash:
I see a few different object classes in SwiftTestObjects.swift that have `NSString` properties.

JadenGeller:
@bdash You're right, I remembered incorrectly. I added a similar test to Realm Swift. Let me know if this is sufficient.

bdash:
👍

jpsim:
:+1:

PR #3980, opened 2016-08-18
Support `NSDate` and `NSNumber` in aggregates

JadenGeller:
Support things like `results.sum(ofProperty: "foo") as NSNumber`.

### DIFF
--- a/RealmSwift/Results.swift
+++ b/RealmSwift/Results.swift
@@ -26,6 +26,16 @@ import Realm
 // Used for conversion from Objective-C types to Swift types
 private protocol Bridgable  { static func bridging(_ value: Any) -> Self }
 
+private func forceCastTrampoline<T, U>(_ x: T) -> U {
+    return x as! U
+}
+
+extension NSNumber: Bridgable {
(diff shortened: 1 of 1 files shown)


JadenGeller:
@bdash

bdash:
Needs tests.

JadenGeller:
@bdash Added tests.

RealmSwift/Results.swift
```diff
@@ -26,6 +26,17 @@ import Realm
 // Used for conversion from Objective-C types to Swift types
 private protocol Bridgable  { static func bridging(_ value: Any) -> Self }
 
+// FIXME: Remove once Swift supports `as! Self` casts
+private func forceCastTrampoline<T, U>(_ x: T) -> U {
```
tgoyne: This function isn't a trampoline.

jpsim: Name isn't terribly important here. `forceCastHelper` would be fine. Additional context is given by the fixme above.

JadenGeller: @tgoyne You're right, I'll rename.

PR #3983, opened 2016-08-18
Fix a crash when an observed object is deallocated from within an observation

### LINKED ISSUE #2549: Crash in RLMObservationInfo::didChange(NSString*, NSKeyValueChange, NSIndexSet*) const (RLMObservation.hpp:116)
Hi!

I'm currently experiencing the following crash when trying to KVO observe objects invalidated property in my Realm. All the registering for the observations is done on the main thread. Any idea what could be wrong?

The crash specifically seems to occur when calling RLMRealm.defaultRealm().refresh() on the main thread and only in Release builds. Debug is running fine!

```
Thread : Crashed: com.apple.main-thread
0  MyApp                       0x0000000100718fc4 RLMObservationInfo::didChange(NSString*, NSKeyValueChange, NSIndexSet*) const (RLMObservation.hpp:116)
1  MyApp                       0x000000010071adb8 (anonymous namespace)::TransactLogHandler::notifyObservers() (RLMObservation.mm:473)
2  MyApp                       0x0000000100719e6c RLMAdvanceRead(realm::SharedGroup&, realm::History&, RLMSchema*) (RLMObservation.mm:503)
3  MyApp                       0x0000000100743d74 -[RLMRealm handleExternalCommit] (RLMRealm.mm:738)
4  MyApp                       0x00000001007488f0 -[RLMNotifier listen]::$_1::__invoke(void*) (RLMRealmUtil.mm:249)
5  CoreFoundation                 0x0000000182753f8c __CFRUNLOOP_IS_CALLING_OUT_TO_A_SOURCE0_PERFORM_FUNCTION__ + 24
6  CoreFoundation                 0x0000000182753230 __CFRunLoopDoSources0 + 264
7  CoreFoundation                 0x00000001827512e0 __CFRunLoopRun + 712
8  CoreFoundation                 0x000000018267cf74 CFRunLoopRunSpecific + 396
9  GraphicsServices               0x000000018c0d76fc GSEventRunModal + 168
10 UIKit                          0x000000018727ed94 UIApplicationMain + 1488
11 MyApp                       0x000000010014a398 main (main.m:33)
12 libdyld.dylib                  0x0000000194b1aa08 start + 4
```

tgoyne:
Retain the object while we're calling observers in case the user releases their last reference to it from within the observer.

Fixes #2549.

### DIFF
--- a/Realm/RLMObservation.hpp
+++ b/Realm/RLMObservation.hpp
@@ -114,10 +114,19 @@ private:
 
     template<typename F>
     void forEach(F&& f) const {
-        for (auto info = prev; info; info = info->prev)
+        // The user's observation handler may release their last reference to
+        // the object being observed, which will result in the RLMObservationInfo
+        // being destroyed. As a result, we need to retain the object which owns
+        // both `this` and the current info we're looking at.
+        __attribute__((objc_precise_lifetime)) id self = object, current;
+        for (auto info = prev; info; info = info->prev) {
+            current = info->object;
             f(info->object);
-        for (auto info = this; info; info = info->next)
+        }
+        for (auto info = this; info; info = info->next) {
+            current = info->object;
             f(info->object);
+        }
     }
 
     // Default move/copy constructors don't work due to the intrusive linked
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,8 @@ x.x.x Release notes (yyyy-MM-dd)
 * Fix crash when an aggregate is accessed as an `Int8`, `Int16`, `Int32`, or `Int64`.
 * Fix a race condition that could lead to a crash if an RLMArray or List was
   deallocated on a different thread than it was created on.
+* Fix a crash when the last reference to an observed object is released from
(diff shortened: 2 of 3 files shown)


Realm/RLMObservation.hpp
```diff
@@ -114,6 +114,9 @@ class RLMObservationInfo {
 
     template<typename F>
     void forEach(F&& f) const {
+        // Ensure that `this` continues to exist until the end of the function
+        // by retaining its owner
+        __attribute__((objc_precise_lifetime)) id self = object;
```
bdash: We could use Foundation's `NS_VALID_UNTIL_END_OF_SCOPE` instead of using the attribute directly.

bdash:
👍

Realm/RLMObservation.hpp
tgoyne: We use the attribute directly everywhere else, and in this case the macro doesn't seem any clearer to me.

jpsim:
👍

PR #3984, opened 2016-08-18
Use `ObjectiveCBridgeable ` for `RealmOptional` conversions

JadenGeller:
Moves `ObjectiveCBridgeable` protocol to `Util.swift` and uses the protocol for `RealmOptional` versions. Additionally, adds "FIXME" message that explains when the `ObjectiveCBridgeable ` protocol will no longer be necessary.

### DIFF
--- a/RealmSwift/Util.swift
+++ b/RealmSwift/Util.swift
@@ -47,6 +47,110 @@ internal func gsub(pattern: String, template: String, string: String, error: NSE
                                            withTemplate: template)
 }
 
+// MARK: Bridgable
+
+// Used for conversion from Objective-C types to Swift types
+internal protocol Bridgable  {
+    /* FIXME: Remove protocol once SR-2393 bridges all integer types to `NSNumber`
+     *        Instead, use `as AnyObject` and `as! [SwiftType]` to cast between. */
+    static func bridging(objCValue: Any) -> Self
+    var objCValue: Any { get }
+}
+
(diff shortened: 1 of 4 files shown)


JadenGeller:
@bdash

RealmSwift/Util.swift
```diff
@@ -47,6 +47,110 @@ internal func gsub(pattern: String, template: String, string: String, error: NSE
                                            withTemplate: template)
 }
 
+// MARK: ObjectiveCBridgeable
+
+// Used for conversion from Objective-C types to Swift types
+internal protocol ObjectiveCBridgeable  {
+    /* FIXME: Remove protocol once SR-2393 bridges all integer types to `NSNumber`
```
jpsim: Doesn't look like the Swift core team ever intends on changing this behavior, so the fixme doesn't really apply.

jpsim:
👍

RealmSwift/Util.swift
JadenGeller: That wasn't the impression I got talking with Joe today. Let me follow up before we change or merge this.

JadenGeller: A wild [proposal](https://github.com/jckarter/swift-evolution/blob/be49e08f56450ffea394306198bcd25f58915e30/proposals/XXXX-bridge-nsnumber-and-nsvalue.md) appeared—looks likely we'll get a fix! 😎

PR #3992, opened 2016-08-23
Simplifies the `CustomObjectiveCBrideable` protocol by removing unnecessary conformances

JadenGeller:
There's no need for implicitly bridged types to conform to the protocol, so we can remove their conformances and use traditional bridging for non-conforming types.


JadenGeller:
I'm not entirely sure this is an improvement given we lose exhaustivity checks, so I'm going to close this.

JadenGeller:
(For context, this change seemed to be the groundwork to fixing a bug, but I late realized that this would not be sufficient.)

Issue #3998, opened 2016-08-23
Update primary key implementation to use `SetUnique` instruction


jpsim:
See realm/realm-core#2026, realm/realm-object-store-private#12 and realm/realm-object-store#137 for existing related work on this.

bigfish24:
Is this in the object store.... if not any reason why this code isn't there?

jpsim:
> if not any reason why this code isn't there?

Yes, the pk functionality in the object store is entangled with its `object` accessor class, which cocoa doesn't use due to incompatible designs. There isn't an appreciable amount of common code that could be consolidated at this time.

tgoyne:
The overwhelming majority of the object accessor code in every binding has been very specific to the target language, with only the most trivial details being the same between them.

bigfish24:
ok just curious, seems like there can be clean up later but that it would slow things down right now

jpsim:
No, there's no clean up to be had. That's what @tgoyne and I are trying to say.

bigfish24:
fair enough but your comment "cocoa doesn't use due to incompatible designs" seemed to imply that this means the object store could be adapted to support the necessary design. Just didn't understand that, and at any rate, its not worth debating

jpsim:
Done in #4003.

PR #4003, opened 2016-08-24
Use `set_int_unique` and `set_string_unique` for primary keys

JadenGeller:
Note that Core doesn't yet support `set_null_unique` which is required for continued support of null integer primary keys. https://github.com/realm/realm-cocoa/issues/3998


JadenGeller:
@jpsim I incorporated the fix suggested by @simonask in https://github.com/realm/realm-sync/issues/673#issuecomment-242439412, so I think it's ready for review 🙂

jpsim:
LGTM 👍

JadenGeller:
To elaborate on why this change was necessary, if we try to `find_first_int` with an argument `0`, we'll get the newly created row that has no yet had its value changed from the default. As such, we want to make sure that we still perform the set operation in these cases.

This would not be necessary if we moved this logic into the row creation function and always checked for uniqueness before creating the row.

Issue #4007, opened 2016-08-25
Support lookup of non-bridged numeric primary keys


JadenGeller:
In Swift 2, users could not look up objects by primary key with `objectForPrimaryKey:` using `Int8`/`Int16`/`Int32`/`Int64` types. Though this was [inconvenient](https://github.com/realm/realm-cocoa/issues/2503#issuecomment-242356497), this was not a major problem since the compiler game a compile-time warning on any attempt. This is no longer true since we've replaced `AnyObject` with `Any` in our Swift codebase. Instead, this causes silent crashes. We should manually bridge these types within the implementation of this method.

JadenGeller:
As a a part of this issue, we should also vet the codebase for other problematic uses of `Any` and ensure we bridge in these cases as well. Another place that'll crash is `filter`, for example. Luckily the Swift team is looking to bridge these types in a [future version](https://github.com/jckarter/swift-evolution/blob/be49e08f56450ffea394306198bcd25f58915e30/proposals/XXXX-bridge-nsnumber-and-nsvalue.md), but we'll have to be very careful about bridging until then.

JadenGeller:
We also ought to be very careful about writing test code that utilizes these types. These crashes flew under the radar because we were using `Int` for *all* test lookups regardless of the primary key type.

JadenGeller:
Fixed in https://github.com/realm/realm-cocoa/pull/4010

PR #4008, opened 2016-08-25
Only conform types in need of custom bridging to `CustomObjectiveCBridgeable`

JadenGeller:
It doesn't make sense to keep adding no-op conformances to the type when we want to use it in different contexts. For example, fixing `filter` to support bridging (https://github.com/realm/realm-cocoa/issues/4007) would have required conforming *all* other types that should *not* be bridged to the protocol. Instead, this PR introduces a `dynamicBridgeCast` function that will bridge if a conformance exists and no-op otherwise.


JadenGeller:
@jpsim

jpsim:
👍

PR #4010, opened 2016-08-26
Fix crash when looking up by non-bridged primary key

JadenGeller:
Calls `dynamicBridgeCast` before passing primary key value into Objective-C to support `Int8`/`Int16`/`Int32`/`Int64` lookups. Note that most of the changes here are improvements of the test structure designed to make it possible to test lookups with the same type as the primary key type (rather than always using `Int`). https://github.com/realm/realm-cocoa/issues/4007

### DIFF
--- a/RealmSwift/Tests/RealmTests.swift
+++ b/RealmSwift/Tests/RealmTests.swift
@@ -497,18 +497,74 @@ class RealmTests: TestCase {
         XCTAssertEqual((object["optObjectCol"] as? SwiftBoolObject)?.boolCol, true)
     }
 
-    func testObjectForPrimaryKey() {
-        let intTypes: [Object.Type] = [SwiftPrimaryIntObject.self,
-                                       SwiftPrimaryInt8Object.self,
-                                       SwiftPrimaryInt16Object.self,
-                                       SwiftPrimaryInt32Object.self,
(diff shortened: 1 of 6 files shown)


JadenGeller:
@jpsim

RealmSwift/Tests/RealmTests.swift
```diff
@@ -497,92 +497,91 @@ class RealmTests: TestCase {
...
+        let realm = try! Realm()
+        try! realm.write {
+            realm.createObject(ofType: SwiftPrimaryStringObject.self, populatedWith: ["a", 1])
+            realm.createObject(ofType: SwiftPrimaryStringObject.self, populatedWith: ["b", 2])
         }
 
-        func assertOptionalIntObject(_ objectType: Object.Type) {
-            let object1 = realm.object(ofType: objectType, forPrimaryKey: NSNull())
-            XCTAssertNotNil(object1)
+        // When this is directly inside the XCTAssertNotNil, it doesn't work
```
jpsim: is there a Swift bug or radar for this?

JadenGeller: I didn't introduce this comment, so no clue. I could test and see if this is still and issue. Maybe we can get rid of it.

jpsim:
👍

Issue #4022, opened 2016-08-30
Changing the primary key of an attached `RealmOptional` fails to throw an exception


JadenGeller:
We're using the [unverified] `RLMDynamicSet` function to set the value of an `RealmOptional` instance, so it currently allows changes to the primary key.

PR #4023, opened 2016-08-30
Enforce uniqueness for `RealmOptional` primary keys

JadenGeller:
Fixes https://github.com/realm/realm-cocoa/issues/4022

@jpsim


jpsim:
👍

Issue #4017, opened 2016-08-26
Set primary key before all other properties


JadenGeller:
A new restriction in Core requires the primary key to be set before all other properties. We should make sure that we do this consistently.

PR #4024, opened 2016-08-30
Set primary key before all other properties

JadenGeller:
Refactors row creation code to always set the primary key on row creation so that the primary key is set before all other properties. This also simplifies the logic of populating properties (now skip all primary key properties rather than just primary key properties of not newly created objects) and removes the need to always emit a `SetUnique` instruction for rows set to the default value (https://github.com/realm/realm-cocoa/pull/4003#issuecomment-243322162).

Fixes https://github.com/realm/realm-cocoa/issues/4017.

@jpsim @bdash @tgoyne

### DIFF
--- a/Realm/RLMObjectStore.mm
+++ b/Realm/RLMObjectStore.mm
@@ -193,13 +193,13 @@ static NSUInteger createRowForObjectWithPrimaryKey(RLMClassInfo const& info, id 
         switch (primaryProperty.type) {
             case RLMPropertyTypeString:
                 REALM_ASSERT_DEBUG(!primaryValue || [primaryValue isKindOfClass:NSString.class]);
-                row.set_string_unique(primaryColumnIndex, RLMStringDataWithNSString((NSString *)primaryValue));
+                row.set_string_unique(primaryColumnIndex, RLMStringDataWithNSString(primaryValue));
                 break;
 
             case RLMPropertyTypeInt:
                 if (primaryValue) {
                     REALM_ASSERT_DEBUG([primaryValue isKindOfClass:NSNumber.class]);
-                    row.set_int_unique(primaryColumnIndex, ((NSNumber *)primaryValue).longLongValue);
+                    row.set_int_unique(primaryColumnIndex, [primaryValue longLongValue]);
                 } else {
                     row.set_null(primaryColumnIndex); // FIXME: Use `set_null_unique` once Core supports it
                 }
--- a/Realm/RLMObjectStore.mm
+++ b/Realm/RLMObjectStore.mm
@@ -183,23 +183,28 @@ static NSUInteger createRowForObjectWithPrimaryKey(RLMClassInfo const& info, id 
     validateValueForProperty(primaryValue, primaryProperty);
     primaryValue = RLMCoerceToNil(primaryValue);
 
-    switch (primaryProperty.type) {
-        case RLMPropertyTypeString:
-            REALM_ASSERT_DEBUG(!primaryValue || [primaryValue isKindOfClass:NSString.class]);
-            row.set_string_unique(primaryColumnIndex, RLMStringDataWithNSString(primaryValue));
-            break;
+    try {
+        switch (primaryProperty.type) {
+            case RLMPropertyTypeString:
+                REALM_ASSERT_DEBUG(!primaryValue || [primaryValue isKindOfClass:NSString.class]);
+                row.set_string_unique(primaryColumnIndex, RLMStringDataWithNSString(primaryValue));
+                break;
 
-        case RLMPropertyTypeInt:
-            if (primaryValue) {
-                REALM_ASSERT_DEBUG([primaryValue isKindOfClass:NSNumber.class]);
-                row.set_int_unique(primaryColumnIndex, [primaryValue longLongValue]);
-            } else {
-                row.set_null(primaryColumnIndex); // FIXME: Use `set_null_unique` once Core supports it
-            }
-            break;
-            
-        default:
-            REALM_UNREACHABLE();
+            case RLMPropertyTypeInt:
+                if (primaryValue) {
+                    REALM_ASSERT_DEBUG([primaryValue isKindOfClass:NSNumber.class]);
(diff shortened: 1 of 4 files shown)


Realm/RLMObjectStore.mm
```diff
@@ -95,45 +95,178 @@ void RLMInitializeSwiftAccessorGenerics(__unsafe_unretained RLMObjectBase *const
...
-        }
-        
-        // search for existing object based on primary key type
-        if (primaryProperty.type == RLMPropertyTypeString) {
-            rowIndex = table.find_first_string(info.tableColumn(primaryProperty), RLMStringDataWithNSString(primaryValue));
-        }
-        else {
-            rowIndex = table.find_first_int(info.tableColumn(primaryProperty), [primaryValue longLongValue]);
+static void validateValueForProperty(__unsafe_unretained id const obj,
+                                        __unsafe_unretained RLMProperty *const prop) {
```
tgoyne: indentation

Realm/RLMObjectStore.mm
```diff
@@ -95,45 +95,178 @@ void RLMInitializeSwiftAccessorGenerics(__unsafe_unretained RLMObjectBase *const
...
+                    return info.table()->find_first_null(primaryPropertyColumn);
+                }
+
+            default:
+                REALM_UNREACHABLE();
         }
-        created = YES;
+
+    }
+    catch (std::exception const& e) {
```
tgoyne: What exceptions is this catching? I don't think any of the things called here can throw.

Realm/RLMObjectStore.mm
```diff
@@ -209,6 +336,12 @@ void RLMAddObjectToRealm(__unsafe_unretained RLMObjectBase *const object,
                 ((void(*)(id, SEL, id))objc_msgSend)(object, prop.setterSel, nil);
             }
         }
+
+        // skip primary key when updating since it doesn't change
+        if (prop.isPrimary) continue;
```
tgoyne: newline before body of if

tgoyne:
👍 

Looks great other than the trivial bits I commented on.

Realm/RLMObjectStore.mm
```diff
@@ -95,45 +95,178 @@ void RLMInitializeSwiftAccessorGenerics(__unsafe_unretained RLMObjectBase *const
...
+
+    // set value for primary key
+    validateValueForProperty(primaryValue, primaryProperty);
+    primaryValue = RLMCoerceToNil(primaryValue);
+
+    try {
+        switch (primaryProperty.type) {
+            case RLMPropertyTypeString:
+                REALM_ASSERT_DEBUG(!primaryValue || [primaryValue isKindOfClass:NSString.class]);
+                row.set_string_unique(primaryColumnIndex, RLMStringDataWithNSString((NSString *)primaryValue));
```
bdash: Is the cast necessary here? Similar code in `getRowForObjectWithPrimaryKey` doesn't have one.

Realm/RLMObjectStore.mm
```diff
@@ -95,45 +95,178 @@ void RLMInitializeSwiftAccessorGenerics(__unsafe_unretained RLMObjectBase *const
...
+                break;
+
+            case RLMPropertyTypeInt:
+                if (primaryValue) {
+                    REALM_ASSERT_DEBUG([primaryValue isKindOfClass:NSNumber.class]);
+                    row.set_int_unique(primaryColumnIndex, ((NSNumber *)primaryValue).longLongValue);
```
bdash: Ditto about the cast here. `[primaryValue longLongValue]` should work without a cast.

Realm/RLMObjectStore.mm
JadenGeller: You're probably right. How do I determine if a Object Store or Core function can throw without reading through all the source? Do we consistently comment the ones that can throw?

What if we change the Object Store or Core to cause these to throw without changing Cocoa? Maybe we should just wrap all calls into our C++ code in this.

Realm/RLMObjectStore.mm
JadenGeller: It's unnecessary if we use message passing syntax instead of dot syntax. Is avoiding explicit static casts really that much of a goal for us?

bdash: Avoiding casts isn't a goal as such, but IMO it's pointless to cast just to use dot syntax. `getRowForObjectWithPrimaryKey` does a message send in the context of virtually identical code.

Realm/RLMObjectStore.mm
tgoyne: If a function isn't marked noexcept and doesn't have a comment specifying what it throws, the default assumption is that it only throws for out-of-memory, precondition violations, and generally unexpected things, none of which are particularly useful to translate to obj-c exceptions; in the first case because allocating new heap objects is pretty much the last thing you want to be doing, and in the others because the translation probably isn't going to actually help the user figure out what's wrong (and may make it harder, since you lose the original stack trace).

JadenGeller: "in the first case because allocating new heap objects is pretty much the last thing you want to be doing" LOL, good point 😛

jpsim:
👍

Realm/RLMObjectStore.mm
JadenGeller: @tgoyne It's true that the `set_`* variants also don't throw, correct?

tgoyne: set_string and set_binary throw with a useful message if the string is too long. I think that's it.

Realm/RLMObjectStore.mm
JadenGeller: I guess. It seems safer to me to *always* cast away from `id`, otherwise you don't have compile-time verification that the object can respond to the message you are sending. I could've easily called a similarly named method that's defined on another type, not `NSNumber`.

Realm/RLMObjectStore.mm
JadenGeller: Oh okay. They aren't documented indicating they throw. I wish C++ required annotations… 😞

Issue #4009, opened 2016-08-26
Fix incorrect force cast to `NSURL` in `TestCast.swift`


JadenGeller:

JadenGeller:
Fixed in https://github.com/realm/realm-cocoa/pull/4025

PR #4025, opened 2016-08-30
Fix force cast to incorrect type `URL` instead of `NSString`

JadenGeller:
Addresses https://github.com/realm/realm-cocoa/issues/4009, which was causing crashes when some tests fail. This used to have no cast when it was imported as `AnyObject`, but the Swift 3 changes importing it as `Any` changed this. In migrating that code, I incorrectly thought this should be cast to `URL` but the actual correct type is `NSString`. I missed that this was incorrect because this code is not called when tests pass.

@jpsim

### DIFF
--- a/RealmSwift/Tests/TestCase.swift
+++ b/RealmSwift/Tests/TestCase.swift
@@ -96,7 +96,7 @@ class TestCase: XCTestCase {
         // Verify that there are no remaining realm files after the test
         let parentDir = (testDir as NSString).deletingLastPathComponent
         for url in FileManager().enumerator(atPath: parentDir)! {
-            let url = url as! URL
+            let url = url as! NSString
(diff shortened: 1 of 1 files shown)


RealmSwift/Tests/TestCase.swift
```diff
@@ -96,7 +96,7 @@ class TestCase: XCTestCase {
         // Verify that there are no remaining realm files after the test
         let parentDir = (testDir as NSString).deletingLastPathComponent
         for url in FileManager().enumerator(atPath: parentDir)! {
```
jpsim: shouldn't we be using `FileManager.default` here?

jpsim:
👍

RealmSwift/Tests/TestCase.swift
JadenGeller: Probably.

Issue #3971, opened 2016-08-17
Find better way to specify `NSNumber` type from Swift when using Realm Objective-C


JadenGeller:
Starting in Swift 3, `Float` and `Double` no longer implicitly bridge to `NSNumber`. This means that model declarations now must write `dynamic var num: NSNumber? = 1.0 as Float as NSNumber` which is far from ideal. We should investigate if there is a better way to do this.

JadenGeller:
There's also no way to create an `NSNumber` model property with a `nil` default value in Swift for any type other than the default.

Issue #4026, opened 2016-08-30
Impossible to create `NSNumber` property with `nil` default value in Swift


JadenGeller:
We require a default value in order to specify the column type.

JadenGeller:
Minimal solution: Introduce `RLMNumberDefault` subclass of `NSNumber` that is *only* used in setting default values. It would have initializers for each numeric type that also take a `nil` value.
* `RLMNumberDefault.init(float: Float?)`
* `RLMNumberDefault.init(double: Double?)`
* `RLMNumberDefault.init(int: Int?)`
* etc.

This wouldn't be a breaking change because we could continue to support setting default values by simply assigning a non-null value to the property, but this would be the preferred way.

JadenGeller:
We could hide that fact that this subclass is used by providing static factory methods on `NSNumber` in an extension.
```swift
class TestObject: RLMObject {
    var a: NSNumber? = .float(nil)
    var b: NSNumber? = .int(nil)
    var c: NSNumber? = .int(100)
}
```

PR #4027, opened 2016-08-30
Allow explicitly specifying `NSNumber` property types in Swift

JadenGeller:
Adds static properties to `NSNumber` that allow users to explicitly specify the type of an `NSNumber` default value, including a `nil` default value fixing https://github.com/realm/realm-cocoa/issues/4026.

Also, this enhancement fixes https://github.com/realm/realm-cocoa/issues/3971, a regression due to the removal of implicit bridging that required `NSNumber<RLMFloat>` properties in Swift to be written `dynamic var floatCol: NSNumber? = 3.14 as Float as NSNumber`.

These changes live in a new public class `RLMNumericNull` and in `RLMSupport.swift`. Though Objective-C users can use `[RLMNumericNull nullFloat]` and friends as default values, there's no reason to recommend or document this due to the superior protocol approach. Swift users can use the `NSNumber` extensions define in `RLMSupport.swift`.

``` swift
class SwiftNilDefaultOptionalNumberObject: RLMObject {
    dynamic var intCol: NSNumber? = .int(nil)
    dynamic var floatCol: NSNumber? = .float(nil)
    dynamic var doubleCol: NSNumber? = .double(nil)
    dynamic var boolCol: NSNumber? = .bool(nil)
}
```

Note that these changes don't expose the extensions to Realm Swift. If we decide this is an API change we want to make, I'll copy the extension to the Realm Swift target as well.

### DIFF
--- a/Realm/RLMSchema.mm
+++ b/Realm/RLMSchema.mm
@@ -231,6 +231,9 @@ static void RLMRegisterClassLocalNames(Class *classes, NSUInteger count) {
             }
 
             [s_localNameToClass enumerateKeysAndObjectsUsingBlock:^(NSString *, Class cls, BOOL *) {
+                if ([NSStringFromClass(cls) isEqualToString:@"Tests.SwiftNilDefaultOptionalNumberObject"]) {
+                    ;
+                }
                 RLMRegisterClass(cls);
             }];
         }
--- a/Realm/RLMSchema.mm
+++ b/Realm/RLMSchema.mm
@@ -231,9 +231,6 @@ static void RLMRegisterClassLocalNames(Class *classes, NSUInteger count) {
             }
 
             [s_localNameToClass enumerateKeysAndObjectsUsingBlock:^(NSString *, Class cls, BOOL *) {
-                if ([NSStringFromClass(cls) isEqualToString:@"Tests.SwiftNilDefaultOptionalNumberObject"]) {
-                    ;
-                }
                 RLMRegisterClass(cls);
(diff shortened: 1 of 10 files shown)


JadenGeller:
@jpsim @austinzheng @bdash

jpsim:
CocoaPods job is failing because it can't find `RLMNumericNull.h`.

Realm/RLMSchema.mm
```diff
@@ -231,6 +231,9 @@ + (instancetype)sharedSchema {
             }
 
             [s_localNameToClass enumerateKeysAndObjectsUsingBlock:^(NSString *, Class cls, BOOL *) {
+                if ([NSStringFromClass(cls) isEqualToString:@"Tests.SwiftNilDefaultOptionalNumberObject"]) {
```
jpsim: debug code?

JadenGeller: Yes, sorry.

jpsim:
I like this, but it certainly does feel like an odd hack to be exposing in our public API. I'd want buy-in from @bdash and @tgoyne before we merge this.

JadenGeller:
I agree. I'm not 100% sold on it, but we also don't really have an alternative for setting a `nil` default value, and the `as Float as NSNumber` method of setting a non-`nil` default is also pretty gross.

jpsim:
Closing this as it's now outdated and it's unclear whether or not it's something we want in its current form.

Issue #3940, opened 2016-08-03
Throw exception when setting both deleteRealmIfMigrationNeeded and readOnly


AlexLardschneider:
Hello,
after I updated my project to Xcode 8 beta 4, my readOnly realm stopped working.
The cause of the crash is a 'Operation not permitted error' as soon as I try to open the realm.

## Goals

What do you want to achieve?

I want to use realm like I did before updating to Xcode 8 beta 4

## Expected Results

What did you expected to happen?

I can use realm like I could before the update, e.g. no crashes when trying to open it.

## Actual Results

What did happened instead?  

My app crashes as soon as I try to open the realm.

```
Could not open realm: Error Domain=io.realm Code=2 "Unable to open a realm at path '/var/containers/Bundle/Application/FC3109B8-F418-4886-98B1-E32EE4854212/test.app/db.realm.management': make_dir() failed: Operation not permitted." UserInfo={Error Code=2, NSFilePath=/var/containers/Bundle/Application/FC3109B8-F418-4886-98B1-E32EE4854212/test.app/db.realm.management, Underlying=make_dir() failed: Operation not permitted, NSLocalizedDescription=Unable to open a realm at path '/var/containers/Bundle/Application/FC3109B8-F418-4886-98B1-E32EE4854212/test.app/db.realm.management': make_dir() failed: Operation not permitted.}
```

## Steps to Reproduce

What are steps we can follow to reproduce this issue?

Happened after I updated to Xcode 8 beta 4, no code changes made, except updating realm to the latest commit to get support for beta 4.

## Code Sample

```

static var CONFIG: Realm.Configuration!

func bundleURL(name: String) -> URL? {
    return Bundle.main.url(forResource: name, withExtension: "realm")
}
    
static func setup() {
    let url = bundleURL(name: "db")
        
    print(url)
        
    CONFIG = Realm.Configuration()
    CONFIG.fileURL = url
    CONFIG.readOnly = true
    CONFIG.deleteRealmIfMigrationNeeded = true
    CONFIG.schemaVersion = 1
        
    do {
       _ = try Realm(configuration: CONFIG)
    } catch let error {
        print("Could not open realm: \(error)")
    }
}

```

## Version of Realm and Tooling

Realm version: ?

Commit '05ca7aad2b74ab9a5e1de11e0f9cbd4598f05458'

Xcode version: ?

Xcode 8 beta 4

iOS/OSX version: ?

10.0

Dependency manager + version: ?

[...]

PR #4031, opened 2016-08-31
Throw when `deleteRealmIfMigrationNeeded` and `readOnly` are both set

JadenGeller:
Fixes https://github.com/realm/realm-cocoa/issues/3940.

Note that this doesn't yet any changes to Swift's configuration, but a similar exception will be thrown when a `Realm.Configuration` is converted internally to an `RLMRealmConfiguration`. It does seem weird to throw exceptions on property set in Objective-C but when creating a `Realm` in Swift. I would appreciate feedback on how to make this more consistent (which way is better). @bdash

### DIFF
--- a/Realm/RLMRealmConfiguration.mm
+++ b/Realm/RLMRealmConfiguration.mm
@@ -217,9 +217,12 @@ static void RLMNSStringToStdString(std::string &out, NSString *in) {
 
 - (void)setReadOnly:(BOOL)readOnly {
     if (readOnly) {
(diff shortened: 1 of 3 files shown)


jpsim:
Could you please add Swift tests? I'm ok with this throwing an ObjC exception on `Realm.init(configuration:)` in Swift.

JadenGeller:
Done. Ready again for review @jpsim @bdash

jpsim:
👍 assuming CI agrees

bdash:
👍

PR #4032, opened 2016-08-31
Annotate nullability in all private Objective-C only headers

JadenGeller:
I audited our private `.h` files for nullability.

@jpsim

### DIFF
--- a/Realm/RLMRealm_Private.h
+++ b/Realm/RLMRealm_Private.h
@@ -20,10 +20,12 @@
 
 @class RLMFastEnumerator;
 
+NS_ASSUME_NONNULL_BEGIN
+
 // Disable syncing files to disk. Cannot be re-enabled. Use only for tests.
 FOUNDATION_EXTERN void RLMDisableSyncToDisk();
 
-FOUNDATION_EXTERN NSData *RLMRealmValidatedEncryptionKey(NSData *key);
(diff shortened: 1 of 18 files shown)


Realm/RLMRealm_Private.h
```diff
@@ -33,7 +35,7 @@ void RLMRealmTranslateException(NSError **error);
 @interface RLMRealm ()
 
 @property (nonatomic, readonly) BOOL dynamic;
-@property (nonatomic, readwrite) RLMSchema *schema;
+@property (nonatomic, readwrite, null_unspecified) RLMSchema *schema;
```
jpsim: I don't think this can ever be null.

jpsim:
👍

Realm/RLMRealm_Private.h
JadenGeller: I was mirroring how it is annotated in the public header.

jpsim: Ah, d80e58b indicates why it has to be `null_unspecified`. 👍

JadenGeller: Looks like it was annotated as such due to a [bug](https://github.com/realm/realm-cocoa/commit/d80e58bd262e) @bdash found. I'll change it if the tests pass.

JadenGeller: Oh, you found it too! (Wish GitHub comments refreshed live 😝)

PR #4033, opened 2016-08-31
Fix nullability annotations in Swift 3 Realm Objective-C tests

JadenGeller:
Broken by incomplete testing of #4032.

I accidentally introduced build errors into master.


jpsim:
👍

PR #4034, opened 2016-08-31
Fix inconsistencies between subscripting managed and unmanaged objects

JadenGeller:
Currently, subscripting a managed object looks up the managed properties value while subscripting an unmanaged object uses `valueForKey:` (meaning it also works with unmanaged properties). This PR ensures that the key corresponds to a property in the schema before calling `valueForKey:`.

@bdash @jpsim

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,8 @@ x.x.x Release notes (yyyy-MM-dd)
 * Fix a crash when `initWithValue:` is used to create a nested object for a class
   with an uninitialized schema.
 * Enforce uniqueness for `RealmOptional` primary keys when using the `value` setter.
+* Fix incorrect behavior where subscripting an unmanaged object can also set properties
+  that are not managed, inconsitent with subscripting a managed object.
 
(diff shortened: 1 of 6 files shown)


jpsim:
Looks like `ObjectTests.testUnretainedRealmObjectUnknownKey` is failing consistently: https://ci.realm.io/job/objc_pr/configuration=Debug,swift_version=2.2,target=ios-dynamic/3907/

JadenGeller:
@jpsim Good catch. Looks like it's failing because the exception message changed. I'll update the test.

CHANGELOG.md
```diff
@@ -27,6 +27,8 @@ x.x.x Release notes (yyyy-MM-dd)
 * Fix a crash when `initWithValue:` is used to create a nested object for a class
   with an uninitialized schema.
 * Enforce uniqueness for `RealmOptional` primary keys when using the `value` setter.
+* Fix incorrect behavior where subscripting an unmanaged object can also set properties
+  that are not managed, inconsitent with subscripting a managed object.
```
jpsim: "inconsistent"

Issue #3970, opened 2016-08-17
Replace `unsafeBitCast` usage with `unsafeVendCast`


JadenGeller:
Our usage of `unsafeBitCast` led to crashes after migration to Swift 3. We use this function to vend a superclass API as if it were a given subclass (that it is not). Because our codebase assumed that we already had an instance of that class (`AnyObject`), it bit casted `Any` to a pointer, which doesn't make sense. We should make our intent more clear by defining a new function that specifies the intermediate cast to the base class and whose name clarifies the purpose of such a cast.

Ideally, we should eventually find a way to remove `unsafeBitCast` from out codebase. It is not a good idea to lie about the type of an object, it can come back to bite the user in many cases. For example, when an `NSArray` is passed into Swift code, element access asserts the type (since Objective-C generics are loosey-goosy 🐦), and this check will fail.

PR #4035, opened 2016-08-31
Introduce `unsafeCastToRLMObject` method on `Object`

JadenGeller:
We expose the Realm Objective-C API operating on `RLMObject`, but many of the methods can operate on any subclass of `RLMObjectBase`. As such, we currently bitcast `Object` to `RLMObject` to call these methods. This commit introduces an `unsafeAsObjectiveCAccessor` function on `Object` to clarify the intent of the cast, to prevent mistakes (such as bitcasting `Any` to `RLMObject` without first force casting to `Object`), and to discourage raw usage of `unsafeBitCast` in our codebase.

Closes https://github.com/realm/realm-cocoa/issues/3970.

### DIFF
--- a/RealmSwift/Util.swift
+++ b/RealmSwift/Util.swift
@@ -47,6 +47,14 @@ internal func gsub(pattern: String, template: String, string: String, error: NSE
                                            withTemplate: template)
 }
 
+extension Object {
+    // Must *only* be used to call Realm Objective-C APIs that are exposed on `RLMObject`
+    // but actually operator on `RLMObjectBase`. Do not expose casted value to user.
+    internal func unsafeAsObjectiveCAccessor() -> RLMObject {
+        return unsafeBitCast(self, to: RLMObject.self)
+    }
+}
+
 // MARK: CustomObjectiveCBridgeable
 
 internal func dynamicBridgeCast<T>(fromObjectiveC x: Any) -> T {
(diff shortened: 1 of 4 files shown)


JadenGeller:
Note that all remaining raw uses of `unsafeBitCast` are especially problematic since they're essentially lying about a type that's vended to the user. This PR does not address this because fixing this requires reworking `DynamicObject`s design: [We cannot bitcast `Object` subclasses to `DynamicObject` since they are often not a `DynamicObject`.](https://github.com/realm/realm-cocoa/issues/4036)

RealmSwift/Util.swift
```diff
@@ -47,6 +47,14 @@ internal func gsub(pattern: String, template: String, string: String, error: NSE
                                            withTemplate: template)
 }
 
+extension Object {
+    // Must *only* be used to call Realm Objective-C APIs that are exposed on `RLMObject`
+    // but actually operator on `RLMObjectBase`. Do not expose casted value to user.
```
jpsim: s/operator/operate/

RealmSwift/Util.swift
```diff
@@ -125,6 +133,14 @@ internal func gsub(pattern: String, template: String, string: String, error: NSE
                                                    withTemplate: template)
 }
 
+extension Object {
+    // Must *only* be used to call Realm Objective-C APIs that are exposed on `RLMObject`
+    // but actually operator on `RLMObjectBase`. Do not expose casted value to user.
```
jpsim: s/operator/operate/

RealmSwift/Util.swift
```diff
@@ -47,6 +47,14 @@ internal func gsub(pattern: String, template: String, string: String, error: NSE
                                            withTemplate: template)
 }
 
+extension Object {
+    // Must *only* be used to call Realm Objective-C APIs that are exposed on `RLMObject`
+    // but actually operator on `RLMObjectBase`. Do not expose casted value to user.
+    internal func unsafeAsObjectiveCAccessor() -> RLMObject {
```
jpsim: How about `unsafeCastToRLMObject`?

jpsim:
👍

PR #4037, opened 2016-09-01
change .travis.yml to upgrade swiftlint rather than install

jpsim:
since the osx image now includes the formula by default.


jpsim:
I'll defer to fixing the line length issues at a later time.

Issue #3690, opened 2016-06-01
Using initWithValue with relationships before initializing the shared schema throws an exception


irishman1921:
What do you want to achieve?

I try create RLMObject initWithValue (dictionary).

What did you expected to happen?

Create new RLMObject.

What did happened instead?  

Crash with log error:

> "Terminating app due to uncaught exception 'RLMException', reason: 'Object type 'DBImage' not persisted in Realm".

When i checked persisted model (created DBImage (initWithValue) before my general object, code working fine. DBImage objects is one-to-many relationships in general model.

What are steps we can follow to reproduce this issue?

Generate dictionary general model -> generate DBImage dictionary (check or uncheck DBImage model persisted(create DBItem model with initWithValue)) -> add all DBImage model to array in general model dictionary -> create general model with initWithValue -> crash (else don't init DBImage before init general model).

Realm version: 1.0

Xcode version: 7.3.1

iOS/OSX version: 9.3.2

Dependency manager + version: cocoaPods + 0.39

irishman1921:
Hm, all work fine, when i call [RLMRealm defaultRealm] before my operation. 
Apparently, realm create realmScheme.

jpsim:
Thanks for reporting this @irishman1921, I think I know what's happening here.

Depending on how we choose to fix this, it's likely that it'll also fix #2377, but that's a completely symptom, so I'm keeping both these issues open.

JadenGeller:
I've been trying very hard to recreate this, but I'm not sure what the issue is. Can you elaborate @jpsim?

jpsim:
@JadenGeller here's a minimal repro case:

```objc
#import "AppDelegate.h"
#import <Realm/Realm.h>

@interface Model1 : RLMObject
@property NSString *prop;
@end
@implementation Model1
@end

RLM_ARRAY_TYPE(Model1)

@interface Model2 : RLMObject
@property RLMArray<Model1> *array;
@end
@implementation Model2
@end

@implementation AppDelegate
- (void)applicationDidFinishLaunching:(NSNotification *)aNotification {
    NSLog(@"Throws 'Object type 'Model1' not managed by the Realm' exception: %@", [[Model2 alloc] initWithValue:@{@"array": @[@[@""]]}]);
}
@end
```

I think the exception message has changed since this issue was first filed.

PR #4039, opened 2016-09-01
Fix crash when `initWithValue:` nested object w/ uninitialized schema

JadenGeller:
Fixes https://github.com/realm/realm-cocoa/issues/3690. @jpsim @tgoyne


JadenGeller:
Will this break for Swift objects (with mangled names)? I guess I ought to write tests to find out...

JadenGeller:
I think fixing this in Swift also will require changing our property parsing code to also store the property `Class` instead of just the class name. Any reason we shouldn't do this?

tgoyne:
There's already some [similar tests in SchemaTests.mm](https://github.com/realm/realm-cocoa/blob/master/Realm/Tests/SchemaTests.mm#L673) which take the approach of running the test in a pristine child process rather than trying to modify a bunch of global state and then restore it afterwards.

Realm/Tests/PreinitializationTests.mm
```diff
@@ -0,0 +1,60 @@
...
+- (void)testDisallowRealmConstruction {
+    RLMAssertThrowsWithReasonMatching(RLMRealm.defaultRealm, @"Illegal to initialize a Realm in this test case.");
+}
+
+- (void)testCreateUnmanagedObject {
+    XCTAssertNoThrow([[IntObject alloc] initWithValue:@[@0]]);
+}
+
+- (void)testCreateUnmanagedObjectWithNestedObject {
+    id value = @[@0, @[@[@0]]]; // on separate line due to parsing bug
```
tgoyne: C preprocessor only groups on parentheses, so any other comma-separated list used within a macro needs to be surrounded with them.

JadenGeller:
@tgoyne It's unfortunate that this makes debugging with breakpoints much more difficult though.

Realm/RLMSchema_Private.h
```diff
@@ -51,4 +51,7 @@ NS_ASSUME_NONNULL_BEGIN
 
 @end
 
+// Used in `PreinitializationTests` to reset the schema before each test case
+void RLMUnsafeResetSchema();
```
bdash: Can this be removed now?

JadenGeller: Yes.

jpsim:
This is breaking `RealmSwift_Tests.SchemaTests.testNoSchemaForUnmanagedObjectClasses`:

> XCTAssertNil failed: throwing "+[RLMObjectBase ignoredProperties]: unrecognized selector sent to class 0x10a33acb8" -

JadenGeller:
Looks like this is currently passing Swift tests, so I think it's done. 👌

Realm/Tests/Swift/SwiftSchemaTests.swift
```diff
@@ -106,6 +106,16 @@ class SwiftSchemaTests: RLMMultiProcessTestCase {
         let _ = SwiftCompanyObject()
     }
 
+    func testCreateUnmanagedObjectWhichCreatesAnotherClassViaInitWithValueDuringSchemaInit() {
```
jpsim: missing Swift 2.x version

jpsim:
Missing a changelog entry and a Swift 2.x version of the test. Otherwise 👍

JadenGeller:
One thing that's bugging me about this: The contents of the `objectSchemas` array is NOT the same as the schemas returned by `schemaForClassName:`. I think the solution to this is that accessing `objectSchemas` should FULLY initialize the object schema. (Also, I think that it's incorrect behavior that `schemaForClassName:` looks up unknown object schemas for any arbitrary `RLMSchema` instance—it should only do this for the shared schema).

PR #4044, opened 2016-09-01
add support for Swift 3 to GettingStarted.xcplayground

jpsim:
/cc @bdash


JadenGeller:
This is in the `realm-cocoa/examples/ios/swift-2.2/GettingStarted.playground/Contents.swift` directory. Seems a bit weird to support Swift 3 here. Also, given that this is a getting started document, maybe it'd make sense to have separate ones for Swift 2 and Swift 3? The `#if` blocks make the code a lot less friendly to beginners.

examples/ios/swift-2.2/GettingStarted.playground/Contents.swift
```diff
@@ -63,9 +67,15 @@ try! realm.write {
 
 let favorites = ["Jennifer"]
 
+#if swift(>=3.0)
+let favoritePeopleWithSpousesAndCars = realm.allObjects(ofType: Person.self)
+    .filter(using: "cars.@count > 1 && spouse != nil && name IN %@", favorites)
+    .sorted(onProperty: "age")
```
bdash: Comparing this Swift 3 block with the Swift 2 API really highlights how much verbosity has been introduced. The only additions that feel like they add value, IMO, is the introduction of the `onProperty` label for `sorted()`.

jpsim: yes, the Swift 3 API really needs work.

JadenGeller: Agreed. I think we ought to discuss the renaming again before Swift 3's final release.

austinzheng: I had a PR to fix some of the most egregious regressions but it got shunted aside. We can revisit it, or just wait until the GM.

bdash:
@JadenGeller, I'm not sure there's a good way to have separate Swift 2.x and Swift 3 example code, particularly if we want it to work with Xcode 8. The example code workspace pulls in the Realm Xcode project, which will defaults to building everything with Swift 3. This means that the Swift 2.x example code would be broken out of the box on Xcode 8. Conditional compilation at least means that the example code would compile, but the user would need to tweak the Xcode setting to build the Swift 2.3 code path.

JadenGeller:
@bdash I see. We could split the _entire_ file into Swift 2 and Swift 3 sections (like our codebase), but it's not clear to me that that is better. :/

jpsim:
> This is in the `realm-cocoa/examples/ios/swift-2.2/` directory. Seems a bit weird to support Swift 3 here. Also, given that this is a getting started document, maybe it'd make sense to have separate ones for Swift 2 and Swift 3? The #if blocks make the code a lot less friendly to beginners.

Yes. To do this right, we should duplicate the `swift-2.2` into a `swift-3.0` directory and update all the examples to Swift 3 in there. Also probably rename `swift-2.2` to `swift-2.x` since it should support 2.3 too.

Despite @bdash's last comment, I do think that we'd need to split it up, assuming we intend to include both Swift 2 & Swift 3 binaries in our release zips when Xcode 8 GM comes out.

jpsim:
Fixed in #4230.

Issue #3779, opened 2016-06-20
Get CI to pass


jpsim:
Specifically, device tests aren't even running at the moment, and #3622.

austinzheng:
This ticket is too broad to be of use anymore.

Issue #3622, opened 2016-05-19
Some interprocess tests on OS X are flaky


jpsim:
testShareInMemoryRealm & testChangeInChildTriggersNotificationInParent often fail when testing OS X in Debug mode. Probably ~30% of the time.

See https://ci.realm.io/job/objc_pr/configuration=Debug,swift_version=2.2,target=osx/buildTimeTrend & https://ci.realm.io/job/objc_pr/configuration=Debug,swift_version=2.2,target=osx-encryption/buildTimeTrend

PR #4045, opened 2016-09-01
disable unreliable tests

jpsim:
/cc @JadenGeller @bdash @tgoyne

### DIFF
--- a/.jenkins.yml
+++ b/.jenkins.yml
@@ -5,8 +5,6 @@ swift_version:
   - 2.2
 target:
   - osx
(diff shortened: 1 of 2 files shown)


JadenGeller:
+1, it's definitely too difficult right now to identity _actual_ failures. Should we create an issue on the backlog for fixing and reenabling them?

jpsim:
> Should we create an issue on the backlog for fixing and reenabling them?

That's covered by #3622 and #3779.

tgoyne:
👍

PR #4047, opened 2016-09-03
Fix behavior of `schemaForClassName:` and `objectSchema`

JadenGeller:
Addresses https://github.com/realm/realm-cocoa/pull/4039#issuecomment-244513503.
@jpsim

### DIFF
--- a/Realm/RLMSchema.mm
+++ b/Realm/RLMSchema.mm
@@ -40,12 +40,12 @@ const uint64_t RLMNotVersioned = realm::ObjectStore::NotVersioned;
 
 // RLMSchema private properties
 @interface RLMSchema ()
-@property (nonatomic, readwrite) NSMutableDictionary *objectSchemaByName;
+@property (nonatomic, readwrite) NSMutableDictionary<NSString *, RLMObjectSchema *> *objectSchemaByName;
 @end
 
 static RLMSchema *s_sharedSchema = [[RLMSchema alloc] init];
-static NSMutableDictionary *s_classByName = [[NSMutableDictionary alloc] init];
-static NSMutableDictionary *s_objectSchemaByName = [[NSMutableDictionary alloc] init];
+static NSMutableDictionary<NSString *, Class> *s_classByName = [[NSMutableDictionary alloc] init];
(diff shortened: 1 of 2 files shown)


jpsim:
This looks ok to me, but I'd feel more comfortable if @tgoyne signed off.

Realm/RLMSchema.mm
```diff
@@ -126,12 +126,20 @@ - (instancetype)init {
 
 - (NSArray *)objectSchema {
     if (!_objectSchema) {
+        if (self == s_sharedSchema) {
```
tgoyne: This check is only safe if `s_sharedSchema` is never assigned to after startup, so the variable should be made `const` (which requires putting the `const` to the right of the `*` in the declaration).

Realm/RLMSchema.mm
```diff
@@ -126,12 +126,20 @@ - (instancetype)init {
...
+            // Ensure the schema is fully initialized
+            [RLMSchema sharedSchema];
```
tgoyne: This initialization is incorrect. We specifically go out of our way to avoid fully initializing the shared schema when it's obtained via `partialSharedSchema` because on watchOS calling `objc_copyClassList()` can easily exceed the memory limit.

JadenGeller: The schema is only fully initialized when the array `objectSchema` is accessed. If a scheme is looked up by name, it doesn't load the entire schema.

JadenGeller: I wouldn't be opposed to adding a private `partialObjectSchema` property to the type, but I don't think we should pretend that it is the full object schema.

tgoyne: The only way for this code to do anything is if `objectSchema` is called on the shared schema obtained via `partialSharedSchema`. If that's something that can never actually happen, then it's dead code.

tgoyne: We're not pretending it's the full object schema. The entire point of `partialSharedSchema` is that it gives you the schema loaded so far and not the full one.

jpsim:
@tgoyne what do you think we should do with this?

jpsim:
Bump @tgoyne.

jpsim:
Closing as it's not clear we actually want this.

PR #4059, opened 2016-09-06
update for Query::find always returning indexes in table

jpsim:
/cc @tgoyne


tgoyne:
👍

Realm/RLMArrayLinkView.mm
```diff
@@ -375,7 +375,16 @@ - (RLMResults *)objectsWithPredicate:(NSPredicate *)predicate {
 - (NSUInteger)indexOfObjectWithPredicate:(NSPredicate *)predicate {
     auto query = translateErrors([&] { return _backingList.get_query(); });
     query.and_query(RLMPredicateToQuery(predicate, _objectInfo->rlmObjectSchema, _realm.schema, _realm.group));
+#if REALM_VER_MAJOR >= 2
```
bmunkholm: It's annoying to have to support 2 different versions - we should have updated master of core as well. If you want that there should not be any reason to not release a 1.x version with this change.

bdash: The behavior change was made on master but I don't think a core release has been made since it was landed.

PR #4062, opened 2016-09-07
Use protocol for primary key specification

JadenGeller:
This is an idea for an API breaking change we might want to make for Swift eventually. We introduce a protocol for models with a primary key that includes an associated type for the primary key. This allows us to provide type-safety in functions like `object(ofType:forPrimaryKey:)` and also allows us to, at compile-time, reduce the change they're using an invalid primary key type.

``` swift
class Person: Object, PrimaryKeyed {
  dynamic var id = 0
  dynamic var name = ""

  typealias PrimaryKey = Int
  static let primaryKey = "id"
}
```

### DIFF
--- a/Realm/RLMObjectSchema.mm
+++ b/Realm/RLMObjectSchema.mm
@@ -138,20 +138,24 @@ using namespace realm;
         }
     }
 
-    if (NSString *primaryKey = [objectClass primaryKey]) {
-        for (RLMProperty *prop in schema.properties) {
-            if ([primaryKey isEqualToString:prop.name]) {
-                prop.indexed = YES;
-                schema.primaryKeyProperty = prop;
-                break;
+    // Swift classes without primary keys don't respond
+    if ([objectClass respondsToSelector:@selector(primaryKey)]) {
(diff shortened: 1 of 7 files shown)


JadenGeller:
@austinzheng and @jpsim are probably the right people to take a look at this, but this is super low priority so do it whenever you feel like it.

JadenGeller:
I don't believe Swift's reflection mechanism exposes typealiases, so it's not possible to verify at runtime that the user used the property type. As such, it might make sense to delay this until we can provide additional runtime verification (or maybe even custom attributes to sidestep all of this).

jpsim:
@mrackwitz had a PR doing something similar a few months back, but we rejected it, I don't remember why.

JadenGeller:
Found it: https://github.com/realm/realm-cocoa/pull/3357#issuecomment-202671465. A small difference here is that this protocol uses an associated type as well, but otherwise it's nearly the same.

JadenGeller:
@jpsim Feel free to close this if you still feel the same.

jpsim:
Closing due to severe age. Tracked in #5030.

PR #4063, opened 2016-09-07
Fix `@escaping` compilation errors introduced in GM

JadenGeller:
Swift 3 GM changed where `@escaping` can be placed (now only in argument positions, not typealiases) and fixed bugs related to its enforcement.

@jpsim @austinzheng @bdash

### DIFF
--- a/Realm/RLMMigration.h
+++ b/Realm/RLMMigration.h
@@ -71,7 +71,7 @@ typedef void (^RLMObjectMigrationBlock)(RLMObject * __nullable oldObject, RLMObj
             to `className`. Instead, treat them as `RLMObject`s and use keyed subscripting to access
             properties.
  */
-- (void)enumerateObjects:(NSString *)className block:(RLMObjectMigrationBlock)block;
+- (void)enumerateObjects:(NSString *)className block:(__attribute__((noescape)) RLMObjectMigrationBlock)block;
 
 /**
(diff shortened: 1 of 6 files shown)


tgoyne:
👍

jpsim:
👍

hsandhu:
I am installing using the source (branch: master) and still get the error related to selecting the Swift version (see screenshot). 
screen shot 2016-09-07 at 3 27 25 pm

jpsim:
@hsandhu `SWIFT_VERSION` _is_ set on `master` ([see Base.xcconfig](https://github.com/realm/realm-cocoa/blob/72a7123ec22677bba077fe9f5f6f2b467f821edd/Configuration/Base.xcconfig#L52)), are you sure you're actually targeting the tip of that branch, and that you don't have any stale built artifacts that might be confusing Xcode?

hsandhu:
Ah, so you are converting the project to Swift 3.0 and also the Realm Swift code to 3.0?

jpsim:
We've supported the Xcode 8 betas since WWDC, though the API has yet to be finalized, it should compile and run (pass our test suite).

KelvinJin:
Hmm, this didn't really fix all the `escaping` errors  I think. I still get errors. Such as the one here. https://github.com/realm/realm-cocoa/blob/4a18f33b153b45adc1a6edc2abdf2c200c80e1df/RealmSwift/Migration.swift#L110

PR #4068, opened 2016-09-08
Final implementation of Swift3/Xcode8 compatible code

vincentsaluzzo:
Make some changes on branch of @jpsim to be compatible with Xcode 8 GM.


bdash:
Thanks for the pull request, @vincentsaluzzo. The changes to RealmSwift itself were already made in #4063. The remaining changes are similar to what is in #4044, which will need to be revised before it is landed.

PR #4078, opened 2016-09-12
Review diffs between Swift 2 --> preliminary Swift 3 API

austinzheng:
@jpsim @bdash @tgoyne @stel @mrackwitz @kishikawakatsumi @TimOliver @JadenGeller


docs/swift-3-api.diff
```diff
@@ -0,0 +1,399 @@
...
+-    public var invalidated: Bool 
++    public var isInvalidated: Bool
+     public var count: Int 
+     public init(fromType type: T.Type, property propertyName: String)
+     public override var description: String
+-    public func indexOf(object: T) -> Int?
+-    public func indexOf(predicate: NSPredicate) -> Int?
+-    public func indexOf(predicateFormat: String, _ args: AnyObject...) -> Int?
++    public func index(of object: T) -> Int?
++    public func indexOfObject(for predicate: NSPredicate) -> Int?
```
bdash: `for` feels odd here. It reads as "index of object for predicate". Something like "index of object matching predicate" would be clearer.

docs/swift-3-api.diff
```diff
@@ -0,0 +1,399 @@
...
+-    public func objects<T: Object>(type: T.Type) -> Results<T>
+-    public func dynamicObjects(className: String) -> Results<DynamicObject>
+-    public func objectForPrimaryKey<T: Object>(type: T.Type, key: AnyObject?) -> T?
+-    public func dynamicObjectForPrimaryKey(className: String, key: AnyObject?) -> DynamicObject?
+-    public func addNotificationBlock(block: NotificationBlock) -> NotificationToken
+-    public var autorefresh: Bool
++    public var isInWriteTransaction: Bool
++    public func add(_ object: Object, update: Bool = false)
++    public func add<S: Sequence>(_ objects: S, update: Bool = false) where S.Iterator.Element: Object
++    public func createObject<T: Object>(ofType type: T.Type, populatedWith value: Any = [:], update: Bool = false) -> T
```
austinzheng: `ofType` should go away, and the first parameter should not be labeled:

`myRealm.createObject(Person.self, populatedWith: [...])`

docs/swift-3-api.diff
```diff
@@ -0,0 +1,399 @@
...
+-    public func min<U: MinMaxType>(property: String) -> U?
+-    public func max<U: MinMaxType>(property: String) -> U?
+-    public func sum<U: AddableType>(property: String) -> U
+-    public func average<U: AddableType>(property: String) -> U?
+-    public func addNotificationBlock(block: (RealmCollectionChange<LinkingObjects> -> Void)) -> NotificationToken
+-    public func generate() -> RLMGenerator<T>
++    public override func value(forKey key: String) -> Any?
++    public override func value(forKeyPath keyPath: String) -> Any?
++    public override func setValue(_ value: Any?, forKey key: String)
++    public func filter(using predicateFormat: String, _ args: Any...) -> Results<T>
```
bdash: The API guidelines mention that weakly-typed parameters should include labels that describe their role. That suggests the `using` label should be something like `usingPredicate` for the string-based `filter` method.

That said, `filter` is used so frequently that the presence of labels feels somewhat noisy to me. `filter(_:)` in Swift's standard library gets away without using any labels, so perhaps we can too?

docs/swift-3-api.diff
```diff
@@ -0,0 +1,399 @@
...
++    public func filter(using predicate: NSPredicate) -> Results<T>
++    public func sorted(onProperty property: String, ascending: Bool = true) -> Results<T>
++    public func sorted<S: Sequence>(with sortDescriptors: S) -> Results<T> where S.Iterator.Element == SortDescriptor
```
bdash: `onProperty` vs `with` feels somewhat inconsistent. I can see why `with` was used as sorting _on_ a sort descriptor doesn't seem grammatically correct. Perhaps this should be "sorted _by_ property" / "sorted _by_ sort descriptors" instead, with `byProperty` and `by` being the labels?

docs/swift-3-api.diff
bdash: This makes creating an object quite verbose.
Before:
```swift
realm.createObject(ofType: Person.self, populatedWith: […])
```

After:
```swift
realm.create(Person.self, value: […])
```

docs/swift-3-api.diff
austinzheng: Yeah, I like that a lot better. In any case "sorted on X" sounds unnatural to me.

docs/swift-3-api.diff
```diff
@@ -0,0 +1,399 @@
...
++    public func add(_ object: Object, update: Bool = false)
++    public func add<S: Sequence>(_ objects: S, update: Bool = false) where S.Iterator.Element: Object
++    public func createObject<T: Object>(ofType type: T.Type, populatedWith value: Any = [:], update: Bool = false) -> T
++    public func createDynamicObject(ofType typeName: String, populatedWith value: Any = [:], update: Bool = false) -> DynamicObject
++    public func delete(_ object: Object)
++    public func delete<S: Sequence>(_ objects: S) where S.Iterator.Element: Object
++    public func delete<T: Object>(_ objects: List<T>)
++    public func delete<T: Object>(_ objects: Results<T>)
++    public func deleteAllObjects()
++    public func allObjects<T: Object>(ofType type: T.Type) -> Results<T>
```
bdash: Ditto about being much more wordy:
```swift
realm.objects(Person.self)
```
vs
```swift
realm.allObjects(ofType: Person.self)
```

docs/swift-3-api.diff
bdash: `populatedWith` doesn't seem like an improvement over `value`.

docs/swift-3-api.diff
tgoyne: This is one of the places I'm torn on. It's clear that there are a lot of users who don't realize that the argument to filter is an NSPredicate format string, so something that makes that obvious could be very beneficial. The problem is that I don't know how to communicate that without making it incredibly unwieldy, and I'm not sure what percentage of those people also don't know what a predicate format string is at all. All the people who use string interpolation or `stringWithFormat:` to build a string to pass to `predicateWithFormat:` obviously won't be helped by knowing that we pass the string to that.

docs/swift-3-api.diff
```diff
@@ -0,0 +1,399 @@
...
+-    public func dynamicObjects(className: String) -> Results<DynamicObject>
+-    public func objectForPrimaryKey<T: Object>(type: T.Type, key: AnyObject?) -> T?
+-    public func dynamicObjectForPrimaryKey(className: String, key: AnyObject?) -> DynamicObject?
+-    public func addNotificationBlock(block: NotificationBlock) -> NotificationToken
+-    public var autorefresh: Bool
++    public var isInWriteTransaction: Bool
++    public func add(_ object: Object, update: Bool = false)
++    public func add<S: Sequence>(_ objects: S, update: Bool = false) where S.Iterator.Element: Object
++    public func createObject<T: Object>(ofType type: T.Type, populatedWith value: Any = [:], update: Bool = false) -> T
++    public func createDynamicObject(ofType typeName: String, populatedWith value: Any = [:], update: Bool = false) -> DynamicObject
```
tgoyne: This was named `dynamicCreate` because we specifically don't want people discovering it via autocomplete unless they're specifically looking for the dynamic object functionality.

austinzheng: That makes sense. We should definitely change it back (and the other similarly structured APIs).

docs/swift-3-api.diff
jpsim: I'd propose `index(of object: T)` and `index(matching predicate: NSPredicate)`.

docs/swift-3-api.diff
jpsim: Agreed, and there's _value_ in keeping APIs as close as possible to their Swift 2.x version, avoiding unnecessary API changes or renames.

jpsim:
I have some recommendations: https://github.com/realm/realm-cocoa/compare/az/swift3apireview...jp/swift3apireview

I can make those edits directly on this branch if that's ok with @austinzheng.

austinzheng:
@jpsim That's absolutely fine with me.

jpsim:
@austinzheng merged, thanks.

jpsim:
I want to start applying these changes later today, so I encourage anyone who still has feedback to give on these APIs to do that soon.

tgoyne:
Everything looks reasonable to me now.

austinzheng:
Looks good to me as well.

Should `autorefresh` (a boolean property) be something like `willAutorefresh` or `isAutorefreshing`?

jpsim:
> Should `autorefresh` (a boolean property) be something like `willAutorefresh` or `isAutorefreshing`?

I'd say no, due to those being the wrong words. "Automatically Refresh" already accurately represents the behavior, but not worth expanding from the current shortened form of `autorefresh`. Given that we can't change the Objective-C property name, I'd prefer aligning the Swift one too.

jpsim:
I've applied the changes from the diff to our APIs.

jpsim:
2 CI failures are due to `Abort trap 6` when running `xcodebuild` and `Application "io.realm.TestHost" is unknown to FrontBoard.` in `CoreSimulator.log`.

The first one we could add to the "known Xcode error" detection in build.sh to rerun the job, the second one seems like something that could fixed by adapting `reset-simulators.sh` to launch the same simulator device as the job will use I think...

jpsim:
I'd like to merge this tomorrow, so I'd appreciate a quick review.

You'll find the completes docs diff with 1.0.2 & Swift 2 here: https://gist.github.com/jpsim/87002160be20193eb4b85a2ea1905231

Realm Swift 1.0.2 Swift 2.2 docs here: https://realm.io/docs/swift/latest/api
Realm Swift master Swift 3.0 docs here: https://static.realm.io/jazzy_demo/RealmSwift3

Looks like `Object` isn't documented at all for Swift 3, which I'll look into now.

austinzheng:
I think Jazzy doesn't (yet) know how to deal with declarations marked with `open`.

jpsim:
> I think Jazzy doesn't (yet) know how to deal with declarations marked with `open`.

I think you're right. It has _partial_ support for it, but not complete.

austinzheng:
👍 , based on the diff changes you made.

Should we remove the `.diff` file from the final PR?

RealmSwift/Realm.swift
```diff
@@ -300,7 +300,7 @@ public final class Realm {
     :nodoc:
     */
     @discardableResult
-    public func createDynamicObject(ofType typeName: String, populatedWith value: Any = [:], update: Bool = false) -> DynamicObject {
+    public func dynamicCreate(_ typeName: String, value: Any = [:], update: Bool = false) -> DynamicObject {
```
JadenGeller: It might be nice for this to share a prefix with `create`. For example, `create(typeNamed:value:update:)`.

RealmSwift/Realm.swift
```diff
@@ -391,13 +391,13 @@ public final class Realm {
...
 
-    - parameter className: The class name of the objects to be returned.
+    - parameter typeName: The class name of the objects to be returned.
 
     - returns: All objects for the given class name as dynamic objects
 
     :nodoc:
     */
-    public func allDynamicObjects(ofType typeName: String) -> Results<DynamicObject> {
+    public func dynamicObjects(_ typeName: String) -> Results<DynamicObject> {
```
JadenGeller: Same here; something like `objects(ofTypeName:)` or hopefully something more consistent with the above suggestion. It seems like our dynamic object API is weirdly distant from out non-dynamic API.

RealmSwift/Migration.swift
```diff
@@ -127,7 +127,7 @@ public final class Migration {
     - returns: The created object.
     */
     @discardableResult
-    public func createObject(ofType typeName: String, populatedWith value: Any = [:]) -> MigrationObject {
+    public func create(_ typeName: String, value: Any = [:]) -> MigrationObject {
```
JadenGeller: It's kind of unfortunate this doesn't read as a sentence anymore. `create(:with:)` or `create(:from:)` might be okay, though not great.

jpsim:
> Should we remove the `.diff` file from the final PR?

Yes of course. I was waiting for this to pass review before doing that so that we can double check that my changes are correct based on our discussions around the `.diff`.

> > I think Jazzy doesn't (yet) know how to deal with declarations marked with `open`.
> I think you're right. It has partial support for it, but not complete.

This was an easy fix in jazzy (https://github.com/realm/jazzy/pull/655). I've updated the gist and API docs I linked to above.

RealmSwift/Realm.swift
jpsim: This function's name _starts_ with `dynamic` to prevent it from showing up in autocorrect since it's not what users want 99% of the time. Ideally we'd have Clang submodules in Swift and expose it in there, but alas we don't, so we namespace it like this.

RealmSwift/Realm.swift
jpsim: Same reply as https://github.com/realm/realm-cocoa/pull/4078#discussion_r78673424

RealmSwift/Migration.swift
jpsim: Still complies with Swift 3 style without being too much of a breaking change for our existing Swift 2.2 users IMO.

jpsim:
@bdash is the change you made in 0d7e7a18 to make `RealmSwift.Error` a struct still necessary? It still seems better suited as an `enum` to me, and the fact that it's a struct with its members only documented indirectly via its `Code` equivalent member complicates the docs.

See http://static.realm.io/jazzy_demo/RealmSwift3/Structs/Error.html

image

---

Unless I'm missing something, I suggest we switch that back to an enum.

austinzheng:
If we want to not make adding new error enum cases a breaking change, it needs to be a struct. Otherwise, it can be an enum.

jpsim:
> If we want to not make adding new error enum cases a breaking change, it needs to be a struct. Otherwise, it can be an enum.

Good point. It'd still be nice to improve how the documentation for this is generated. Maybe moving the doc strings onto the Error members directly rather than the Code would be doable?

jpsim:
@austinzheng @tgoyne @bdash I think this is ready to land.

PR #4083, opened 2016-09-13
update build.sh docs generation for Swift 3 & Jazzy 0.7.1

jpsim:
Jazzy 0.7.1 changed the format of `undocumented.json` and added support for Swift 3 ACLs.

/cc @bdash


bdash:
:+1:

jpsim:
On an unrelated note, we should replace the usage of `jq` in `reset-simulators.sh` with the same `ruby -rjson -e '...'` trick used here.

PR #4094, opened 2016-09-14
Update packaging for Xcode 8

jpsim:
See individual commits for a detailed list of things that have changed in this PR.

### DIFF
--- a/scripts/package_examples.rb
+++ b/scripts/package_examples.rb
@@ -52,7 +52,7 @@ framework_directory_for_example = {
   'examples/osx/objc' => '../../../osx',
   'examples/tvos/objc' => '../../../tvos',
   'examples/ios/swift-2.2' => '../../../ios/swift-2.2',
-  'examples/tvos/swift-2.2' => '../../../tvos',
+  'examples/tvos/swift-2.2' => '../../../tvos/swift-2.2',
 }
 
 # Update the paths to the prebuilt frameworks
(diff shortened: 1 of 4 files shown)


jpsim:
Not sure who can review this. Maybe @tgoyne since you're familiar with the packaging process, and @austinzheng for high levels?

scripts/package_examples.rb
```diff
@@ -53,11 +50,13 @@ def replace_in_file(filepath, pattern, replacement)
 end
 
 framework_directory_for_example = {
-  'examples/ios/xcode-7/objc' => '../../../../ios/static/xcode-7',
+  'examples/ios/objc' => '../../../ios/static',
   'examples/osx/objc' => '../../../osx',
   'examples/tvos/objc' => '../../../tvos',
   'examples/ios/swift-2.2' => '../../../ios/swift-2.2',
-  'examples/tvos/swift-2.2' => '../../../tvos',
+  'examples/tvos/swift-2.2' => '../../../tvos/swift-2.2',
```
tgoyne: Is the lack of 2.3 for these because none of our examples are compatible with 2.3? If there's any non-packaging work needed for them then it seems fine to not bother (for now at least), but it does seem a little strange to be missing them.

tgoyne:
👍

scripts/package_examples.rb
jpsim: There's non-packaging work needed for them to build, and I don't think it's worth fixing that now. Essentially it's unclear how to tell `Realm.xcodeproj` to use Swift 2.3 from the Xcode GUI, without completely changing how it's built (i.e. building it via `build.sh` rather than Xcode like it is now).

PR #4114, opened 2016-09-19
Implement thread-safe references

JadenGeller:
This pull request is mostly finished, but there is no Swift 2 implementation and the test coverage could probably be improved. I'm _handing_ this _over_ to you, @jpsim 😝

### DIFF
--- a/Realm/RLMArrayLinkView.mm
+++ b/Realm/RLMArrayLinkView.mm
@@ -26,6 +26,7 @@
 #import "RLMQueryUtil.hpp"
 #import "RLMRealm_Private.hpp"
 #import "RLMSchema.h"
+#import "RLMThreadSafeReference_Private.hpp"
 #import "RLMUtil.hpp"
 
 #import "list.hpp"
@@ -46,17 +47,30 @@
     std::unique_ptr<RLMObservationInfo> _observationInfo;
 }
 
-- (RLMArrayLinkView *)initWithParent:(__unsafe_unretained RLMObjectBase *const)parentObject
-                            property:(__unsafe_unretained RLMProperty *const)property {
+- (RLMArrayLinkView *)initWithList:(realm::List)list
+                             realm:(__unsafe_unretained RLMRealm *const)realm
+                        parentInfo:(RLMClassInfo *)parentInfo
+                          property:(__unsafe_unretained RLMProperty *const)property {
     self = [self initWithObjectClassName:property.objectClassName];
     if (self) {
-        _realm = parentObject->_realm;
-        _backingList = realm::List(_realm->_realm, parentObject->_row.get_linklist(parentObject->_info->tableColumn(property)));
-        _objectInfo = &parentObject->_info->linkTargetType(property.index);
-        _ownerInfo = parentObject->_info;
+        _realm = realm;
+        REALM_ASSERT_DEBUG(list.get_realm() == realm->_realm);
+        _backingList = std::move(list);
+        _objectInfo = &parentInfo->linkTargetType(property.index);
+        _ownerInfo = parentInfo;
         _key = property.name;
     }
     return self;
+}
+
+- (RLMArrayLinkView *)initWithParent:(__unsafe_unretained RLMObjectBase *const)parentObject
+                            property:(__unsafe_unretained RLMProperty *const)property {
+    __unsafe_unretained RLMRealm *const realm = parentObject->_realm;
+    realm::List list(realm->_realm, parentObject->_row.get_linklist(parentObject->_info->tableColumn(property)));
+    return [self initWithList:std::move(list)
+                        realm:realm
+                   parentInfo:parentObject->_info
+                     property:property];
 }
 
 void RLMValidateArrayObservationKey(__unsafe_unretained NSString *const keyPath,
@@ -425,5 +439,48 @@ static void RLMInsertObject(RLMArrayLinkView *ar, RLMObject *object, NSUInteger 
     return RLMAddNotificationBlock(self, _backingList, block);
 }
 #pragma clang diagnostic pop
+
+@end
+
+@interface RLMArrayLinkViewHandoverMetadata : NSObject
+
+@property (nonatomic) NSString *parentClassName;
+@property (nonatomic) NSString *key;
+
+@end
+
+@implementation RLMArrayLinkViewHandoverMetadata
+@end
+
+@interface RLMArrayLinkView (ThreadConfined) <RLMThreadConfined_Private>
+@end
+
+@implementation RLMArrayLinkView (ThreadConfined)
+
+- (std::unique_ptr<realm::ThreadSafeReferenceBase>)rlm_newThreadSafeReference {
+    realm::ThreadSafeReference<realm::List> list_reference = _realm->_realm->obtain_thread_safe_reference(_backingList);
+    return std::make_unique<realm::ThreadSafeReference<realm::List>>(std::move(list_reference));
+}
+
+- (RLMArrayLinkViewHandoverMetadata *)rlm_objectiveCMetadata {
+    RLMArrayLinkViewHandoverMetadata *metadata = [[RLMArrayLinkViewHandoverMetadata alloc] init];
+    metadata.parentClassName = @(_ownerInfo->objectSchema->name.c_str());
+    metadata.key = _key;
+    return metadata;
+}
+
++ (instancetype)rlm_objectWithThreadSafeReference:(std::unique_ptr<realm::ThreadSafeReferenceBase>)reference
+                                         metadata:(RLMArrayLinkViewHandoverMetadata *)metadata
+                                            realm:(RLMRealm *)realm {
+    REALM_ASSERT_DEBUG(dynamic_cast<realm::ThreadSafeReference<realm::List> *>(reference.get()));
+    auto list_reference = static_cast<realm::ThreadSafeReference<realm::List> *>(reference.get());
+
+    realm::List list = realm->_realm->resolve_thread_safe_reference(std::move(*list_reference));
+
+    return [[RLMArrayLinkView alloc] initWithList:std::move(list)
+                                            realm:realm
+                                       parentInfo:&realm->_info[metadata.parentClassName]
+                                         property:realm.schema[metadata.parentClassName][metadata.key]];
+}
 
 @end
--- a/Realm/RLMArrayLinkView.mm
+++ b/Realm/RLMArrayLinkView.mm
@@ -476,7 +476,9 @@ static void RLMInsertObject(RLMArrayLinkView *ar, RLMObject *object, NSUInteger 
     auto list_reference = static_cast<realm::ThreadSafeReference<realm::List> *>(reference.get());
 
     realm::List list = realm->_realm->resolve_thread_safe_reference(std::move(*list_reference));
-
+    if (!list.is_valid()) {
+        return nil;
+    }
     return [[RLMArrayLinkView alloc] initWithList:std::move(list)
                                             realm:realm
                                        parentInfo:&realm->_info[metadata.parentClassName]
--- a/Realm/RLMArrayLinkView.mm
+++ b/Realm/RLMArrayLinkView.mm
@@ -35,6 +35,17 @@
 #import <realm/table_view.hpp>
 #import <objc/runtime.h>
 
+@interface RLMArrayLinkViewHandoverMetadata : NSObject
+@property (nonatomic) NSString *parentClassName;
+@property (nonatomic) NSString *key;
+@end
+
+@implementation RLMArrayLinkViewHandoverMetadata
+@end
+
+@interface RLMArrayLinkView () <RLMThreadConfined_Private>
+@end
+
 //
 // RLMArray implementation
 //
@@ -440,22 +451,7 @@ static void RLMInsertObject(RLMArrayLinkView *ar, RLMObject *object, NSUInteger 
 }
 #pragma clang diagnostic pop
 
-@end
-
-@interface RLMArrayLinkViewHandoverMetadata : NSObject
-
-@property (nonatomic) NSString *parentClassName;
-@property (nonatomic) NSString *key;
-
-@end
-
-@implementation RLMArrayLinkViewHandoverMetadata
-@end
-
-@interface RLMArrayLinkView (ThreadConfined) <RLMThreadConfined_Private>
-@end
-
-@implementation RLMArrayLinkView (ThreadConfined)
+#pragma mark - Thread Confined Protocol Conformance
 
 - (std::unique_ptr<realm::ThreadSafeReferenceBase>)rlm_newThreadSafeReference {
     realm::ThreadSafeReference<realm::List> list_reference = _realm->_realm->obtain_thread_safe_reference(_backingList);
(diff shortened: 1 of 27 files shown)


jpsim:
Thanks for your continued work on this @JadenGeller. Hope school is going well!

jpsim:
I've thought a bit more about potential convenience APIs, and the gains don't seem to justify the constraints at this point.

For example, here's a sample use case with the current full/flexible ("inconvenient") API:

```swift
let ref = ThreadSafeReference(to: obj)
Dispatch.main.async {
    let realm = try! Realm()
    guard let obj = realm.resolve(ref) else { return }
    try! realm.write {
        // do something with obj
    }
}
```

Let's look at the advantages of this API, from a usage and cognitive load standpoint:

1.

jpsim:
This is ready for review @austinzheng @bdash @tgoyne!

A few noteworthy design choices you should be aware of:

* I've opted to never return invalid accessors, but rather return `nil` instead.
* There might be cases in which it would be safe to recover from a failed attempt to resolve a reference, but `std::move`-ing the object store reference prevents this. Given that TSRs should be deallocated shortly after their use, I don't think avoiding the `move` would be harmful in any way, unless there's a valid reason to want to discourage error handling.
* This PR won't include a convenience API. Design discussions are under way in #4477.

austinzheng:
Is there a object store PR still associated with this PR? If not, perhaps it should be pointing at latest `master`.

jpsim:
There _was_ an object store PR, which was merged in realm/realm-object-store#157. This branch now points to _almost_ the tip of `master`, minus 4 commits: https://github.com/realm/realm-object-store/compare/1cb5d3e...master

Those 4 commits are breaking changes that I haven't adopted this branch to support in order to minimize the surface area of this PR.

Realm/RLMThreadSafeReference.h
```diff
@@ -0,0 +1,99 @@
...
+@end
+
+/**
+ An object intended to be passed between threads containing a thread-safe reference to its
+ thread-confined object.
+
+ To resolve a thread-safe reference on a target Realm on a different thread, pass to
+ `-[RLMRealm resolveThreadSafeReference:]`.
+
+ @warning Every `RLMThreadSafeReference` object created must be resolved exactly once.
```
austinzheng: 'exactly' or 'at most'?

Realm/RLMThreadSafeReference.mm
```diff
@@ -0,0 +1,84 @@
...
+    }
+
+    REALM_ASSERT_DEBUG([threadConfined conformsToProtocol:@protocol(RLMThreadConfined)]);
+    if (![threadConfined conformsToProtocol:@protocol(RLMThreadConfined_Private)]) {
+        @throw RLMException(@"Illegal custom conformance to `RLMThreadConfined` by `%@`", threadConfined.class);
+    } else if (threadConfined.invalidated) {
+        @throw RLMException(@"Cannot construct reference to invalidated object");
+    } else if (!threadConfined.realm) {
+        @throw RLMException(@"Cannot construct reference to unmanaged object, "
+                            "which can be passed across threads directly");
```
austinzheng: Should the second string be an `NSString` as well?

RealmSwift/ThreadSafeReference.swift
```diff
@@ -0,0 +1,156 @@
...
+}
+
+/**
+ An object intended to be passed between threads containing a thread-safe reference to its
+ thread-confined object.
+
+ To resolve a thread-safe reference on a target Realm on a different thread, pass to
+ `Realm.resolve(_:)`.
+
+ - warning: Every `ThreadSafeReference` object created must be resolved exactly once.
```
austinzheng: ditto

Realm/RLMThreadSafeReference.h
jpsim: "exactly once", otherwise the version will be pinned forever.

Realm/RLMThreadSafeReference.mm
jpsim: it's all an NSString, that's how you build a multiline string literal in Objective-C.

Realm/RLMThreadSafeReference.h
bdash: Surely it'd only be pinned until the reference is deallocated

jpsim: Correct, but is that a use case we want to encourage?

Realm/RLMThreadSafeReference.mm
```diff
@@ -0,0 +1,84 @@
...
+    }
+
+    translateErrors([&] {
+    //       Should we check the thread ourselves here?
```
bdash: Should we?

jpsim: Oops, no. I think this crept back in during rebasing.

Realm/RLMThreadSafeReference.h
bdash: I think it'd be preferable to be state why failing to resolve the reference could be problematic. Something like:

> A `RLMThreadSafeReference` object must be resolved at most once. Failing to resolve a `RLMThreadSafeReference` may result in the source version of the Realm being pinned until the reference is deallocated.

I don't think phrasing the warning like this encourages any undesirable behavior.

tgoyne: There should probably be another `@warning` advising users not to just outright store the thread-safe reference for an extended period of time.

Realm/RLMArrayLinkView.mm
```diff
@@ -427,3 +441,48 @@ - (RLMNotificationToken *)addNotificationBlock:(void (^)(RLMArray *, RLMCollecti
...
+
+@end
+
+@implementation RLMArrayLinkViewHandoverMetadata
+@end
+
+@interface RLMArrayLinkView (ThreadConfined) <RLMThreadConfined_Private>
+@end
+
+@implementation RLMArrayLinkView (ThreadConfined)
```
tgoyne: I'm not sure why this is a category. Splitting chunks of the implementation off into separate categories isn't something we do anywhere else in obj-c and it doesn't have obvious organizational benefits.

Realm/RLMArrayLinkView.mm
```diff
@@ -427,3 +441,48 @@ - (RLMNotificationToken *)addNotificationBlock:(void (^)(RLMArray *, RLMCollecti
...
+
+- (std::unique_ptr<realm::ThreadSafeReferenceBase>)rlm_newThreadSafeReference {
+    realm::ThreadSafeReference<realm::List> list_reference = _realm->_realm->obtain_thread_safe_reference(_backingList);
+    return std::make_unique<realm::ThreadSafeReference<realm::List>>(std::move(list_reference));
+}
+
+- (RLMArrayLinkViewHandoverMetadata *)rlm_objectiveCMetadata {
+    RLMArrayLinkViewHandoverMetadata *metadata = [[RLMArrayLinkViewHandoverMetadata alloc] init];
+    metadata.parentClassName = @(_ownerInfo->objectSchema->name.c_str());
```
tgoyne: `_ownerInfo->rlmObjectSchema.className`

Realm/RLMRealm.h
```diff
@@ -18,6 +18,7 @@
 
 #import <Foundation/Foundation.h>
 #import "RLMConstants.h"
+#import "RLMThreadSafeReference.h"
```
tgoyne: Don't need this import (can just forward-declare RLMThreadSafeReference).

Realm/RLMThreadSafeReference.mm
```diff
@@ -0,0 +1,83 @@
...
+        @throw RLMException(@"Can only resolve a thread safe reference once.");
+    }
+    return translateErrors([&] {
+        return [_type rlm_objectWithThreadSafeReference:std::move(_reference)
+                                               metadata:_metadata realm:realm];
+    });
+}
+
+- (BOOL)isInvalidated {
+    return !(bool)_reference;
```
tgoyne: The cast here is pointless.

Realm/RLMThreadSafeReference_Private.hpp
```diff
@@ -0,0 +1,45 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+////////////////////////////////////////////////////////////////////////////
+
+#import "RLMThreadSafeReference.h"
+#include "shared_realm.hpp"
+#include "thread_safe_reference.hpp"
```
tgoyne: Neither of these headers are needed (ThreadSafeReferenceBase can be forward-declared, and shared_realm.hpp isn't used at all).

jpsim: The `RLMThreadConfined` protocol is needed and can't be forward-declared.

Realm/RLMThreadSafeReference_Private.hpp
```diff
@@ -0,0 +1,44 @@
...
+
+#import "RLMThreadSafeReference.h"
+#include "thread_safe_reference.hpp"
+
+NS_ASSUME_NONNULL_BEGIN
+
+@protocol RLMThreadConfined_Private <NSObject>
+
+// Constructs a new `ThreadSafeReference`
+- (std::unique_ptr<realm::ThreadSafeReferenceBase>)rlm_newThreadSafeReference;
```
tgoyne: No reason to have the rlm_ prefix on these.

Realm/RLMThreadSafeReference_Private.hpp
```diff
@@ -0,0 +1,44 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+////////////////////////////////////////////////////////////////////////////
+
+#import "RLMThreadSafeReference.h"
+#include "thread_safe_reference.hpp"
```
tgoyne: #import

Realm/RLMArrayLinkView.mm
```diff
@@ -46,19 +58,32 @@ @implementation RLMArrayLinkView {
...
+                        parentInfo:(RLMClassInfo *)parentInfo
+                          property:(__unsafe_unretained RLMProperty *const)property {
     self = [self initWithObjectClassName:property.objectClassName];
     if (self) {
-        _realm = parentObject->_realm;
-        _backingList = realm::List(_realm->_realm, parentObject->_row.get_linklist(parentObject->_info->tableColumn(property)));
-        _objectInfo = &parentObject->_info->linkTargetType(property.index);
-        _ownerInfo = parentObject->_info;
+        _realm = realm;
+        REALM_ASSERT_DEBUG(list.get_realm() == realm->_realm);
```
tgoyne: Doesn't need to be a `_DEBUG()` assert (the intention for _DEBUG is for things we would like to always check, but are too slow to do in real builds).

Realm/RLMArrayLinkView.mm
```diff
@@ -426,4 +451,34 @@ - (RLMNotificationToken *)addNotificationBlock:(void (^)(RLMArray *, RLMCollecti
...
+    auto list_reference = static_cast<realm::ThreadSafeReference<realm::List> *>(reference.get());
+
+    realm::List list = realm->_realm->resolve_thread_safe_reference(std::move(*list_reference));
+    if (!list.is_valid()) {
+        return nil;
+    }
+    return [[RLMArrayLinkView alloc] initWithList:std::move(list)
+                                            realm:realm
+                                       parentInfo:&realm->_info[metadata.parentClassName]
+                                         property:realm.schema[metadata.parentClassName][metadata.key]];
```
tgoyne: Could do one fewer lookup here by getting the rlmobjectschema from the parentinfo.

Realm/RLMRealm.h
```diff
@@ -406,6 +406,26 @@ typedef void (^RLMNotificationBlock)(RLMNotification notification, RLMRealm *rea
 
 #pragma mark - Accessing Objects
 
+/**
+ Returns the same object as the one referenced when the `RLMThreadSafeReference` was first created,
+ but resolved for the current Realm for this thread. Returns `nil` if this object was deleted after
+ the reference was created, or if the operation failed.
```
tgoyne: I don't think the operation can fail?

Realm/RLMThreadSafeReference.h
```diff
@@ -0,0 +1,100 @@
...
+
+#import <Foundation/Foundation.h>
+
+@class RLMRealm;
+
+NS_ASSUME_NONNULL_BEGIN
+
+/**
+ An instance which is bound to a thread-specific `RLMRealm` instance, and so cannot be passed
+ between threads without being explicitly exported and imported.
```
tgoyne: This wording seems off. In particular, you can have objects of types which conform to `RLMThreadConfined` which are not actually confined to a thread and cannot be exported/imported (i.e. unmanaged objects). A stab at fixing this:

Objects of types which conform to `RLMThreadConfined` can be managed by a Realm, which will make them bound to a thread-specific `RLMRealm` instance. Managed objects must be explicitly exported and imported to be passed between threads.

Managed instances of objects conforming to this protocol can be converted to a thread-safe reference for transport between threads by passing to the `+[RLMThreadSafeReference referenceWithThreadConfined:]` constructor.

Realm/RLMThreadSafeReference.h
```diff
@@ -0,0 +1,100 @@
...
+ between threads by passing to the `+[RLMThreadSafeReference referenceWithThreadConfined:]`
+ constructor.
+
+ Note that only types defined by Realm can meaningfully conform to this protocol, and defining new
+ classes which attempt to conform to it will not make them work with `RLMThreadSafeReference`.
+ */
+@protocol RLMThreadConfined <NSObject>
+// Conformance to the `RLMThreadConfined_Private` protocol will be enforced at runtime.
+
+/// The Realm which manages the object, or `nil` if the object is unmanaged.
```
tgoyne: Maybe "Unmanaged objects are not confined to a thread and cannot be passed to methods expecting a `RLMThreadConfined` object."

Realm/RLMThreadSafeReference.mm
```diff
@@ -0,0 +1,81 @@
...
+    try {
+        return f();
+    }
+    catch (std::exception const& e) {
+        @throw RLMException(e);
+    }
+}
+
+@implementation RLMThreadSafeReference {
+@private
```
tgoyne: `@private` is pointless since these are only visible to the things in this file anyway.

RealmSwift/ThreadSafeReference.swift
```diff
@@ -0,0 +1,161 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+////////////////////////////////////////////////////////////////////////////
+
+import Realm
+
+/**
+ An instance which is bound to a thread-specific `Realm` instance, and so cannot be passed
```
tgoyne: Doc comments from obj-c apply here as well.

jpsim:
Thanks for the great feedback @tgoyne! I've addressed all your points here.

PR #4121, opened 2016-09-20
Allow throws within Realm.write(_:) block in Swift 3

### LINKED ISSUE #4120: Block of the write function is non-throwable in Swift 3
In Swift 3 version block of the write function is not marked as throwable, even though documentation states that it can throw. Swift 2 version is still implemented with a block that can throw.

```swift
// Declaration in Swift 3 version
public func write(_ block: () -> Swift.Void) throws

// Declaration in Swift 2 version
public func write(@noescape block: (() throws -> Void)) throws
```

jpsim:
Fixes #4120 /cc @austinzheng

### DIFF
--- a/RealmSwift/Realm.swift
+++ b/RealmSwift/Realm.swift
@@ -116,8 +116,15 @@ public final class Realm {
      - throws: An `NSError` if the transaction could not be completed successfully.
                If `block` throws, the function throws the propagated `ErrorType` instead.
      */
-    public func write(_ block: () -> Void) throws {
-        try rlmRealm.transaction(block)
+    public func write(_ block: (() throws -> Void)) throws {
+        beginWrite()
+        do {
+            try block()
+        } catch let error {
+            if isInWriteTransaction { cancelWrite() }
+            throw error
+        }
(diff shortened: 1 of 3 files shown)


RealmSwift/Realm.swift
```diff
@@ -116,8 +116,15 @@ public final class Realm {
      - throws: An `NSError` if the transaction could not be completed successfully.
                If `block` throws, the function throws the propagated `ErrorType` instead.
      */
-    public func write(_ block: () -> Void) throws {
-        try rlmRealm.transaction(block)
+    public func write(_ block: (() throws -> Void)) throws {
```
austinzheng: Could you mark the block as `throws`, the function as `rethrows`, and just keep the old implementation?

jpsim: No. `rethrows` doesn't apply since the function can throw even if the block doesn't.

jpsim:
Docs failure is unrelated. Was due to some CI workers not having jazzy 0.7.2. I updated the workers so that shouldn't happen again.

PR #4148, opened 2016-09-27
Improve object server test suite

austinzheng:
@jpsim 

Replaces https://github.com/realm/realm-cocoa-private/pull/369

Changes:
- Tests should no longer stall forever on semaphores
- 'Waiting for upload/download' API usage replaced by macros that assert on timeout
- Added tests for syncing multiple Realms
- Added more helper methods to make actual test code easier to read
- Moved responsibility for checking for errors when synchronously opening users and Realms for tests into the helper methods
- Organized and grouped new tests by purpose

### DIFF
--- a/RLMSyncManager+ObjectServerTests.h
+++ b/RLMSyncManager+ObjectServerTests.h
@@ -16,20 +16,10 @@
 //
 ////////////////////////////////////////////////////////////////////////////
 
-#import <Foundation/Foundation.h>
+#import "RLMSyncManager.h"
 
-@class RLMSyncUser;
+@interface RLMSyncManager (ObjectServerTests)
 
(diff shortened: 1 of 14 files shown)


jpsim:
Can we get the object server tests actually _running_ on CI before continuing to work on them?

austinzheng:
Are they not already? Do you want me to take that on?

jpsim:
The job runs on CI, but the tests don't run. They're hanging just like before: https://ci.realm.io/job/objc_pr/configuration=Release,swift_version=3.0,target=osx-object-server/4102/console

```
▸ Touching ObjectServerTests.xctest
Build timed out (after 25 minutes). Marking the build as aborted.
Build was aborted
```

Yes, I'd appreciate if you wanted to look into fixing that.

PR #4150, opened 2016-09-27
Add support for fetching Realm

austinzheng:
Needs: object server tests need to be updated.

Prereq: https://github.com/realm/realm-object-store/pull/174
Replaces: https://github.com/realm/realm-cocoa-private/pull/269

Changes:
- Session handles can now wait for uploads/downloads to complete
- Added API to fetch a Realm and expose it to the user upon download completion
- Fix Xcode project target membership for certain Sync source files
- Fix a threading bug
- Changed the `standaloneSession` BOOL argument some internal methods took into a `purpose` enum to more clearly distinguish between fetching Realms, opening Realms, and getting a standalone session


austinzheng:
I believe this feature is no longer desired? I am closing this PR, since even if we want it the implementation is no longer compatible with the latest code.

jpsim:
We definitely still need a "download Realm" API that only makes the Realm available when the entire Realm file has been downloaded, and without automatically creating the file if it doesn't already exist, but I agree that we'd be better off starting fresh to implement that, and I think we're still blocked on a few lower-level sync features to expose that (protocol changes, among other things).

PR #4156, opened 2016-09-28
Update to core 2.0.1

tgoyne:


dependencies.list
```diff
@@ -1,4 +1,4 @@
 VERSION=2.0.0
-REALM_CORE_VERSION=2.0.0
+REALM_CORE_VERSION=2.0.1
```
bdash: Hmm, I'm not sure this is even _used_ at this point. It looks like it's only used by `build.sh download-core`, and we've replaced all calls to this with `build.sh download-sync`.

jpsim: Correct.

tgoyne:
This now actually uses core 2.0.1. I think all of the failing jobs on CI are spurious.

tgoyne:
All of the failing jobs work for me locally, so just going to blame CI issues for them.

PR #4159, opened 2016-09-28
Fix object server test file settings

austinzheng:
@jpsim 

Changes:
- Object store multiprocess tests shouldn't share local Realm files anymore, meaning that they can actually test sync

This should fix the issue where the metadata Realm couldn't be opened by the child test processes, causing the tests to hang/fail.

### DIFF
--- a/Realm/ObjectServerTests/RLMObjectServerTests.m
+++ b/Realm/ObjectServerTests/RLMObjectServerTests.m
@@ -20,6 +20,20 @@
 
 #import "RLMSyncUser+ObjectServerTests.h"
 
+static NSURL *makeRealmURL(const char *function, NSString *identifier) {
+    // 'function' is expected to be an Objective-C method name: "[MyClass fooBarBaz]"
+    NSString *functionAsString = @(function);
+    NSString *reduced = [functionAsString substringWithRange:NSMakeRange(1, [functionAsString length] - 2)];
+    NSString *methodName = [reduced componentsSeparatedByCharactersInSet:[NSCharacterSet whitespaceCharacterSet]][1];
+    return [NSURL URLWithString:[NSString stringWithFormat:@"realm://localhost:9080/~/%@%@",
+                                 [methodName stringByAddingPercentEncodingWithAllowedCharacters:[NSCharacterSet alphanumericCharacterSet]],
+                                 identifier ?: @""]];
+}
+
+#define CUSTOM_REALM_URL(realm_identifier) makeRealmURL(__FUNCTION__, realm_identifier)
+
+#define REALM_URL() CUSTOM_REALM_URL(@"")
+
 @interface RLMObjectServerTests : RLMSyncTestCase
 @end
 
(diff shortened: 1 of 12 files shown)


Realm/RLMSyncFileManager.mm
```diff
@@ -39,7 +39,13 @@ + (NSURL *)_baseDirectory {
     dispatch_once(&onceToken, ^{
         // Create the path.
         NSFileManager *manager = [NSFileManager defaultManager];
-        NSURL *base = [NSURL fileURLWithPath:RLMDefaultDirectoryForBundleIdentifier(nil)];
+        NSString *bundleIdentifier = nil;
+        if (getenv("RLMProcessIsChild")) {
```
bdash: `RLMProcessIsChild` is an environment variable used by tests. Non-test code should never use it.

austinzheng: Is there a way to programmatically set a child process's bundle ID?

bdash: No. Bundle IDs are static.

tgoyne:
A private method which lets the test code override the location that the sync files are placed would be a cleaner way to achieve this IMO.

austinzheng:
I got rid of my original changes, so I just amended and force-pushed.

austinzheng:
Object server tests seem to finally be running with these changes: https://ci.realm.io/job/objc_pr/configuration=Release,swift_version=3.0,target=osx-object-server/4109/

Realm/ObjectServerTests/RLMObjectServerTests.m
```diff
@@ -20,6 +20,20 @@
 
 #import "RLMSyncUser+ObjectServerTests.h"
 
+static NSURL *makeRealmURL(const char *function, NSString *identifier) {
+    // 'function' is expected to be an Objective-C method name: "[MyClass fooBarBaz]"
```
bdash: Any reason to not do something like:
```swift
var components = URLComponents(string: "realm://localhost:9080/~/")!
components.path += functionName
return components.url
```

That'll take care of escaping any characters in `functionName` that need to be escaped. It drops the whole `identifier` thing as that doesn't appear to be used.

austinzheng: URLComponents seems like a good idea. I'd like to keep the `identifier` if possible, since I have some tests in the pipeline that would use it.

Realm/RLMSyncManager.mm
```diff
@@ -164,20 +171,25 @@ + (void)_resetStateForTesting {
         }
     }
     [[s_sharedManager.activeUsers allValues] makeObjectsPerformSelector:@selector(logOut)];
+    NSURL *metadataURL = [s_sharedManager.fileManager fileURLForMetadata];
 
     // Reset the singleton.
+    s_testBundleID = nil;
     s_onceToken = 0;
```
bdash: It's not sufficient to simply zero out a `dispatch_once_t` like this given how `dispatch_once` uses it.

austinzheng: I wonder if it makes sense of the tests to swizzle `[RLMSyncManager sharedManager]` and just have each test instantiate an instance of the manager. I don't think the singleton state of `RLMSyncManager` is tied to the singleton state of any of the object store singletons.

austinzheng:
Alright. This is ready for review @jpsim. Tests seem to be working reliably on CI now.

jpsim:
I'm rerunning the CI job as https://ci.realm.io/job/objc_pr/4136/ because the object server tests failed to run on your last commit due to a "Remote call" Jenkins error.

PR #4160, opened 2016-09-28
Use the Object Server from the public Realm Mobile Platform download

bdash:
This lets anyone run the object server tests, rather than only those
with access to a private S3 bucket.

/cc @jpsim @austinzheng

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -283,7 +283,7 @@ fi
 ######################################
 
 download_object_server() {
-    curl -C - -L -O "http://static.realm.io/downloads/mobile-platform/v$REALM_OBJECT_SERVER_VERSION/realm-mobile-platform.zip"
+    curl -C - -L -O "https://static.realm.io/downloads/mobile-platform/v$REALM_OBJECT_SERVER_VERSION/realm-mobile-platform.zip"
     rm -rf sync
     mkdir sync
     (cd sync && unzip ../realm-mobile-platform.zip)
--- a/build.sh
+++ b/build.sh
@@ -283,14 +283,14 @@ fi
 ######################################
 
 download_object_server() {
-    local archive_name="realm-object-server-bundled_node_darwin-$REALM_OBJECT_SERVER_VERSION.tar.gz"
-    /usr/local/bin/s3cmd get --force "s3://realm-ci-artifacts/services-bundle/$REALM_OBJECT_SERVER_VERSION/$archive_name"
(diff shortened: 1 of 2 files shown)


build.sh
```diff
@@ -283,14 +283,14 @@ fi
 ######################################
 
 download_object_server() {
-    local archive_name="realm-object-server-bundled_node_darwin-$REALM_OBJECT_SERVER_VERSION.tar.gz"
-    /usr/local/bin/s3cmd get --force "s3://realm-ci-artifacts/services-bundle/$REALM_OBJECT_SERVER_VERSION/$archive_name"
+    curl -C - -L -O "http://static.realm.io/downloads/mobile-platform/v$REALM_OBJECT_SERVER_VERSION/realm-mobile-platform.zip"
```
jpsim: can we use `https` here please?

build.sh
```diff
@@ -283,14 +283,14 @@ fi
...
     rm -rf sync
     mkdir sync
-    tar -C sync -xf $archive_name
-    rm  $archive_name
-    echo "\nenterprise:\n  skip_setup: true" >> "sync/object-server/configuration.yml"
-    touch "sync/object-server/do_not_open_browser"
+    (cd sync && unzip ../realm-mobile-platform.zip)
+    rm realm-mobile-platform.zip
+    echo "\nenterprise:\n  skip_setup: true" >> "sync/realm-object-server/object-server/configuration.yml"
+    sed -i '' -e "s/    listen_address: '0\.0\.0\.0'/    listen_address: '::'/" "sync/realm-object-server/object-server/configuration.yml"
```
jpsim: is this temporary until a new release of the object server is cut with this change baked in by default?

bdash: Yeah. It silences some distracting output when running the tests. I wrote up an issue about changing this to be the default.

radu-tutueanu:
A better solution would be publishing the ROS on static.realm.io . If we merge this, if there is a protocol change, we won't be able to release the RMP without a release of cocoa. And we cannot do a release of cocoa without the RMP.

radu-tutueanu:
It is available at `s3://static.realm.io/downloads/object-server/`  e.g. `https://static.realm.io/downloads/object-server/realm-object-server-bundled_node_darwin-1.0.0-BETA-1.1.tar.gz`

bdash:
@radu-tutueanu, I'm glad you're paying attention! I'll revise this as you suggest.

bdash:
Superseded by Thomas's change in #4173.

PR #4161, opened 2016-09-28
Enable building 100% from source via CocoaPods (with sync disabled)

jpsim:
This change makes it possible to build Realm Objective-C and/or Realm Swift 100% from source, although without any of the sync or Realm Mobile Platform features.

This requires (for now) a Podfile similar to this:

``` ruby
pod 'RealmCore', git: 'https://github.com/realm/realm-core.git', branch: 'v2.0.1'
pod 'Realm', git: 'https://github.com/realm/realm-cocoa.git', branch: 'jp-oss-podspecs', submodules: true
pod 'RealmSwift', git: 'https://github.com/realm/realm-cocoa.git', branch: 'jp-oss-podspecs'
```

And to install with the `REALM_DISABLE_SYNC` environment variable:

``` shell
$ REALM_DISABLE_SYNC=TRUE pod install
```

---

Up for discussion:
1. Should we do this via CocoaPods subspecs instead of env variables? If so, how? Seems better suited, but a bit more cumbersome to set up. /cc @mrackwitz 
2. If we keep the variable, should we rename it? e.g. `REALM_BUILD_FROM_SOURCE`, `REALM_SOURCE_CORE`, etc.

/cc @mrackwitz @tgoyne @bdash


Realm.podspec
```diff
@@ -1,89 +1,136 @@
...
+    s.compiler_flags       = '-D__ASSERTMACROS__',
+                             "-DREALM_COCOA_VERSION='@\"#{s.version}\"'",
+                             '-DREALM_ENABLE_ASSERTIONS',
+                             '-DREALM_ENABLE_ENCRYPTION'
+    s.pod_target_xcconfig  = { 'APPLICATION_EXTENSION_API_ONLY' => 'YES',
+                               'CLANG_CXX_LANGUAGE_STANDARD' => 'c++14',
+                               'HEADER_SEARCH_PATHS' => '"${PODS_ROOT}/RealmCore/src" "${PODS_ROOT}/Realm/Realm/ObjectStore/src"' }
+  else
+    s.libraries            = 'c++', 'z'
+    public_header_files    = 'include/**/RLMArray.h',
```
tgoyne: If we do go this route I'd rather have the non-sync files added unconditionally and have this just add more files rather than overriding the entire list.

jpsim:
I've redone this PR and it should now be ready for review.

This doesn't change the default behavior of `master` (other than fixing an issue where core would be downloaded/unpacked twice when installing Realm Swift.

To use this, you can specify the following in your Podfile:

```ruby
pod 'RealmCore', git: 'https://github.com/realm/realm-core.git', branch: 'v2.1.0'
pod 'Realm', git: 'https://github.com/realm/realm-cocoa.git', branch: 'jp-oss-podspecs', submodules: true
pod 'RealmSwift', git: 'https://github.com/realm/realm-cocoa.git', branch: 'jp-oss-podspecs'
```

And invoke `pod install` with the `REALM_DISABLE_SYNC` environment variable defined:

```
$ REALM_DISABLE_SYNC=1 pod install
```

/cc @mrackwitz @tgoyne @bdash

jpsim:
Bump @mrackwitz @tgoyne @bdash

mrackwitz:
>Should we do this via CocoaPods subspecs instead of env variables?

I think we should. Environment variables aren't really suitable for this purpose:

1. The presence of the environment variable won't be persisted in the `Podfile.lock` and so will effectively break the contract that this ensures a deterministically reproducible installation.
2. The environment variable is evaluated when the podspec is interpreted. Podspecs are submitted to trunk by transforming them into a declarative JSON representation. This will be derived from the Ruby podspec depending on the value of the environment variable when the podspec is submitted.

This would effectively require to use the following Podfile syntax to include the OSS version once this PR was merged to master:
```ruby
ENV['REALM_DISABLE_SYNC'] = 1
pod 'Realm', git: 'https://github.com/realm/realm-cocoa.git', branch: 'master', submodules: true
pod 'RealmSwift', git: 'https://github.com/realm/realm-cocoa.git', branch: 'master'
```

---

With a hypothetical solution using subspecs we would want to keep the current behavior, including sync by default. This matches the behavior of subspecs which are also included by default. But this can be overwritten by the podspec attribute `default_subspecs`. We would include a `Sync` subspec, including sync-related header and implementation files and relying on the presence of a prebuilt core. Additionally, we would add an alternative subspec to that e.g. `OSCore`, which depends on the dedicated RealmCore podspec and excludes sync-related files. This would not be part of the `default_subspecs`, so it can be chosen from the Podfile.

### Standard Installation

```ruby
pod 'Realm' #, subspecs: ['Sync']
pod 'RealmSwift' #, subspecs: ['Sync']
```

### From Source

```ruby
pod 'Realm', subspecs: ['OSCore']
pod 'RealmSwift', subspecs: []
```

jpsim:
I've switched this back to `S:In Progress` as I rewrite basically in its entirety to use CocoaPods subspecs.

jpsim:
Rewriting this from scratch has really not been a priority for me since I last touched this, so I no longer consider this to be "in progress", so I'll be closing this PR but keeping the branch alive so I can refer to it if ever I get back to this.

PR #4167, opened 2016-09-30
Fix Swift 2.3 and 3.0 installation examples

tgoyne:
None of them had been updated from Swift 2.2.

@jpsim

### DIFF
--- a/examples/installation/build.sh
+++ b/examples/installation/build.sh
@@ -7,6 +7,7 @@ Usage: sh $0 command [argument]
 command:
   test-all:                        tests all projects in this repo.
   test-xcode7:                     tests all Xcode 7 projects in this repo.
+  test-xcode8:                     tests all Xcode 8 projects in this repo.
 
   test-ios-objc-static:            tests iOS Objective-C static example.
   test-ios-objc-dynamic:           tests iOS Objective-C dynamic example.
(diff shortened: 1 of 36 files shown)


examples/installation/ios/swift-3.0/DynamicExample/DynamicExample/AppDelegate.swift
```diff
@@ -19,16 +19,10 @@
 import UIKit
 import RealmSwift
 
-public class MyModel: Object {}
+public class MyModel: RealmSwift.Object {}
```
jpsim: was this necessary to compile?

examples/installation/build.sh
```diff
@@ -6,7 +6,6 @@ Usage: sh $0 command [argument]
 
 command:
   test-all:                        tests all projects in this repo.
-  test-xcode6:                     tests all Xcode 6 projects in this repo.
   test-xcode7:                     tests all Xcode 7 projects in this repo.
```
jpsim: looks like this actually drops support for Xcode 7, which is fine, as long as these names are updated to refer to Xcode 8.

examples/installation/ios/swift-3.0/DynamicExample/DynamicExample/AppDelegate.swift
tgoyne: Yes. I'm not sure why, especially since none of the other did.

examples/installation/build.sh
tgoyne: I don't understand.

examples/installation/ios/swift-3.0/DynamicExample/DynamicExample/AppDelegate.swift
jpsim: That's... concerning.

examples/installation/build.sh
jpsim: Sorry, what I actually meant was that this adds support for Xcode 8 but there's no `test-xcode8` command added here.

jpsim: `test-xcode8` should also have an entry here.

PR #4174, opened 2016-10-03
Skip prelaunching the simulator for OS X PR jobs

tgoyne:
This takes ~40 seconds and is pointless for the OS X jobs.


austinzheng:
You might want to ask @jpsim; when I tried doing something similar he had qualms about the CI machines not being able to switch Xcode versions properly. Otherwise, +1.

jpsim:
In my limited tests, switching between Xcode 7.3.1 and 8.0, even with macOS-only builds/tests can cause CoreSimulator to crash `xcodebuild`. If that can somehow be avoided, then great! But we should make sure.

One option would be to refactor all our CI jobs in such a way that each VM would only ever use a single Xcode version but I don't think anyone has the time to do so at the moment.

PR #4175, opened 2016-10-03
Improve object server unit tests

### LINKED ISSUE #4151: Figure out why object server tests hang on CI
cc @jpsim

austinzheng:
@jpsim @bdash @tgoyne 

Supersedes: #4159, #4148 
Requires: https://github.com/realm/realm-object-store/pull/186
Fixes #4151 

Changes:
- Object server tests now each use unique URLs for synced Realms
- Object server multiprocess tests now use separate Realm files for child and server processes
- RLMFileManager API changed so that an instance can be created with a custom root directory path
- Private RLMSyncManager API added to specify a custom root directory path for testing purposes
- Cleaned up RLMSyncManager_Private.hpp to reduce number of imports
- Test code now swizzles `[RLMSyncManager defaultManager]` to return a special instance that is created anew for each test
- Moved test-specific APIs out of the main `RLMSyncManager` code
- `RLMRealm` is no longer responsible for resetting the sync manager before tests
- Tests should no longer stall forever on semaphores
- Tests should not continue after the first failure
- 'Waiting for upload/download' API usage replaced by macros that assert on timeout
- Added tests for syncing multiple Realms
- Added more helper methods to make actual test code easier to read
- Moved responsibility for checking for errors when synchronously opening users and Realms for tests into the helper methods
- Organized and grouped new tests by purpose

### DIFF
--- a/Realm/ObjectServerTests/RLMObjectServerTests.m
+++ b/Realm/ObjectServerTests/RLMObjectServerTests.m
@@ -20,6 +20,11 @@
 
 #import "RLMSyncUser+ObjectServerTests.h"
 
+#define ACCOUNT_NAME() NSStringFromSelector(_cmd)
+#define CUSTOM_REALM_URL(realm_identifier) \
+    [NSURL URLWithString:[NSString stringWithFormat:@"realm://localhost:9080/~/%@%@", ACCOUNT_NAME(), realm_identifier]]
+#define REALM_URL() CUSTOM_REALM_URL(@"")
+
 @interface RLMObjectServerTests : RLMSyncTestCase
 @end
 
@@ -30,16 +35,21 @@
 /// A valid username/password credential should be able to log in a user. Using the same credential should return the
 /// same user object.
 - (void)testUsernamePasswordAuthentication {
-    RLMSyncUser *firstUser = [self logInUserForCredential:[RLMSyncTestCase basicCredential:YES]
+    RLMSyncUser *firstUser = [self logInUserForCredential:[RLMSyncTestCase basicCredentialWithName:ACCOUNT_NAME()
+                                                                                     createAccount:YES]
                                                    server:[RLMSyncTestCase authServerURL]];
-    RLMSyncUser *secondUser = [self logInUserForCredential:[RLMSyncTestCase basicCredential:NO]
+    RLMSyncUser *secondUser = [self logInUserForCredential:[RLMSyncTestCase basicCredentialWithName:ACCOUNT_NAME()
+                                                                                      createAccount:NO]
                                                     server:[RLMSyncTestCase authServerURL]];
     // Logging in with equivalent credentials should return the same user object instance.
     XCTAssertEqual(firstUser, secondUser);
+    // Authentication server property should be properly set.
+    XCTAssertEqualObjects(firstUser.authenticationServer, [RLMSyncTestCase authServerURL]);
 
     // Trying to "create" a username/password account that already exists should cause an error.
     XCTestExpectation *expectation = [self expectationWithDescription:@""];
-    [RLMSyncUser authenticateWithCredential:[RLMObjectServerTests basicCredential:YES]
+    [RLMSyncUser authenticateWithCredential:[RLMObjectServerTests basicCredentialWithName:ACCOUNT_NAME()
+                                                                            createAccount:YES]
                               authServerURL:[RLMObjectServerTests authServerURL]
                                onCompletion:^(RLMSyncUser *user, NSError *error) {
         XCTAssertNil(user);
@@ -68,14 +78,15 @@
 /// `[RLMSyncUser all]` should be updated once a user is logged in.
 - (void)testBasicUserPersistence {
     XCTAssertEqual([[RLMSyncUser all] count], 0U);
-    RLMSyncUser *user = [self logInUserForCredential:[RLMObjectServerTests basicCredential:YES]
+    RLMSyncUser *user = [self logInUserForCredential:[RLMObjectServerTests basicCredentialWithName:ACCOUNT_NAME()
+                                                                                     createAccount:YES]
                                               server:[RLMObjectServerTests authServerURL]];
     XCTAssertNotNil(user);
     XCTAssertEqual([[RLMSyncUser all] count], 1U);
     XCTAssertTrue([[RLMSyncUser all] containsObject:user]);
 }
 
-#pragma mark - Sync
+#pragma mark - Basic Sync
 
 /// It should be possible to successfully open a Realm configured for sync with an access token.
 - (void)testOpenRealmWithAdminToken {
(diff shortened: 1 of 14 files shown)


RLMSyncManager+ObjectServerTests.h
```diff
@@ -0,0 +1,25 @@
+////////////////////////////////////////////////////////////////////////////
```
tgoyne: This file is in the root directory.

RLMSyncManager+ObjectServerTests.m
```diff
@@ -0,0 +1,76 @@
...
+@interface RLMSyncManager ()
+- (NSMutableDictionary<NSString *, RLMSyncUser *> *)activeUsers;
+- (RLMSyncFileManager *)fileManager;
+@end
+
+@implementation RLMSyncManager (ObjectServerTests)
+
++ (void)load {
+    static dispatch_once_t onceToken;
+    dispatch_once(&onceToken, ^{
```
tgoyne: Why is this wrapped in dispatch_once? The runtime only calls `load` once and locks around calls to it.

Realm/ObjectServerTests/RLMObjectServerTests.m
```diff
@@ -88,97 +104,200 @@ - (void)testOpenRealmWithAdminToken {
...
+
+/// If a client opens multiple Realms, there should be one session object for each Realm that was opened.
+- (void)testMultipleRealmsSessions {
+    NSURL *urlA = CUSTOM_REALM_URL(@"a");
+    NSURL *urlB = CUSTOM_REALM_URL(@"b");
+    NSURL *urlC = CUSTOM_REALM_URL(@"c");
+    RLMSyncUser *user = [self logInUserForCredential:[RLMObjectServerTests basicCredential:self.isParent]
+                                              server:[RLMObjectServerTests authServerURL]];
+    // Open three Realms.
+    __unused RLMRealm *realmA = [self openRealmForURL:urlA user:user];
```
tgoyne: If this is expecting the realms to stay alive until the end of the scope the variables need to be marked with `__attribute__((objc_precise_lifetime))`.

Realm/ObjectServerTests/RLMObjectServerTests.m
```diff
@@ -190,92 +309,104 @@ - (void)testUploadChangesWhenRealmOutOfScope {
...
     if (self.isParent) {
         [r beginWriteTransaction];
-        [r addObject:[[SyncObject alloc] initWithValue:@[@"parent-1"]]];
+        for (NSInteger i=0; i<OBJECT_COUNT; i++) {
+            [r addObject:[[SyncObject alloc] initWithValue:@[[NSString stringWithFormat:@"parent-%@", @(i+1)]]]];
+        }
         [r commitWriteTransaction];
+        [user logOut];
+        CHECK_COUNT(OBJECT_COUNT, SyncObject, r);
+        usleep(50000);
```
tgoyne: Why the sleep? If it's actually doing anything, there's a good chance this test will fail spuriously on CI machines under load.

Realm/ObjectServerTests/RLMSyncTestCase.h
```diff
@@ -32,29 +34,46 @@ NS_ASSUME_NONNULL_BEGIN
...
+                         realms:(NSArray<RLMRealm *> *)realms
+                      realmURLs:(NSArray<NSURL *> *)realmURLs
+                 expectedCounts:(NSArray<NSNumber *> *)counts;
+
 @end
 
 NS_ASSUME_NONNULL_END
 
+#define WAIT_FOR_UPLOAD(macro_user, macro_url) \
+XCTAssertTrue([macro_user waitForUploadToFinish:macro_url], @"Upload timed out for URL: %@", macro_url);
```
tgoyne: Indent the body of macros just like any other code.

austinzheng:
Apologies for the `sleep()`s. A lot of them were left over from when I was trying to get the tests working on CI, and should have been removed. The rest will be replaced by proper synchronization before I push another commit.

Realm/ObjectServerTests/RLMObjectServerTests.m
```diff
@@ -20,6 +20,25 @@
 
 #import "RLMSyncUser+ObjectServerTests.h"
 
+static NSString *reduceMethodName(const char *function) {
```
jpsim: doesn't seem like there's any reason to have this take `const char *` rather than `NSString *`, and just do the `@()` wrapping at the call site / macro if needed.

Realm/ObjectServerTests/RLMObjectServerTests.m
```diff
@@ -20,6 +20,25 @@
...
+    // 'function' is expected to be an Objective-C method name: "[MyClass fooBarBaz]"
+    NSString *functionAsString = @(function);
+    NSString *reduced = [functionAsString substringWithRange:NSMakeRange(1, [functionAsString length] - 2)];
+    return [reduced componentsSeparatedByCharactersInSet:[NSCharacterSet whitespaceCharacterSet]][1];
+}
+
+static NSURL *makeRealmURL(const char *function, NSString *identifier) {
```
jpsim: same here about taking `NSString *`

Realm/ObjectServerTests/RLMObjectServerTests.m
```diff
@@ -88,97 +113,206 @@ - (void)testOpenRealmWithAdminToken {
...
-    RLMSyncUser *user = [self logInUserForCredential:[RLMObjectServerTests basicCredential:self.isParent]
+- (void)testAddObjects {
+    NSURL *url = REALM_URL();
+    RLMSyncUser *user = [self logInUserForCredential:[RLMObjectServerTests basicCredentialWithName:ACCOUNT_NAME()
+                                                                                     createAccount:self.isParent]
                                               server:[RLMObjectServerTests authServerURL]];
-    NSError *error = nil;
-    RLMRealm *r = [self openRealmForURL:url user:user error:&error];
-    XCTAssertNil(error, @"Error when opening Realm: %@", error);
+    RLMRealm *r = [self openRealmForURL:url user:user];
```
jpsim: Calling this `r` hurts readability. `realm` isn't long

Realm/ObjectServerTests/RLMObjectServerTests.m
```diff
@@ -88,97 +113,206 @@ - (void)testOpenRealmWithAdminToken {
...
+    XCTAssert(user.allSessions.count == 3, @"Expected 3 sessions, but didn't get 3 sessions");
+    XCTAssertNotNil([user sessionForURL:urlA], @"Expected to get a session for URL A");
+    XCTAssertNotNil([user sessionForURL:urlB], @"Expected to get a session for URL B");
+    XCTAssertNotNil([user sessionForURL:urlC], @"Expected to get a session for URL C");
+    XCTAssertTrue([user sessionForURL:urlA].state == RLMSyncSessionStateActive, @"Expected active session for URL A");
+    XCTAssertTrue([user sessionForURL:urlB].state == RLMSyncSessionStateActive, @"Expected active session for URL B");
+    XCTAssertTrue([user sessionForURL:urlC].state == RLMSyncSessionStateActive, @"Expected active session for URL C");
+}
+/*
+/// A client should be able to open multiple Realms and add objects to each of them.
```
jpsim: I'd prefer prefixing disabled tests with `DISABLED_` rather than commenting them out to at least ensure they compile.

Realm/ObjectServerTests/RLMObjectServerTests.m
```diff
@@ -187,362 +321,279 @@ - (void)testUploadChangesWhenRealmOutOfScope {
...
-        // Wait for the sessions to asynchronously rebind
-        // FIXME: once new token system is in this will be unnecessary
-        usleep(500000);
-        [r beginWriteTransaction];
-        [r addObject:[[SyncObject alloc] initWithValue:@[@"parent-2"]]];
-        [r addObject:[[SyncObject alloc] initWithValue:@[@"parent-3"]]];
-        [r commitWriteTransaction];
-        [user waitForUploadToFinish:url];
+        // Wait for the Realm's session to be bound.
+        WAIT_FOR_SEMAPHORE(sema, 30);
```
jpsim: There's a real problem if this takes 30 seconds to happen. What's the lowest value that reliably passes?

Realm/ObjectServerTests/RLMSyncManager+ObjectServerTests.m
```diff
@@ -0,0 +1,73 @@
...
+    }
+}
+
++ (instancetype)ost_sharedManager {
+    return [RLMSyncTestCase managerForCurrentTest];
+}
+
+- (void)prepareForDestruction {
+    // Log out all the logged-in users. This will immediately kill any open sessions.
+    NSMutableArray<RLMSyncUser *> *buffer = [NSMutableArray array];
```
jpsim: Equivalent?

```objc
[self.activeUsers.allValues makeObjectsPerformSelector:@selector(logOut)];
```

Realm/ObjectServerTests/RLMSyncManager+ObjectServerTests.m
```diff
@@ -0,0 +1,73 @@
...
+    for (id key in [self activeUsers]) {
+        [buffer addObject:[self activeUsers][key]];
+    }
+    [[[self activeUsers] allValues] makeObjectsPerformSelector:@selector(logOut)];
+    NSURL *metadataURL = [self.fileManager fileURLForMetadata];
+
+    // Destroy the metadata Realm.
+    // FIXME: replace this with the appropriate call to `[RLMSyncFileManager deleteRealmAtPath:]` once that code is in.
```
jpsim: Can we instead reuse [`-[RLMTestCase deleteRealmFileAtURL:]`](https://github.com/realm/realm-cocoa/blob/v2.0.1/Realm/Tests/RLMTestCase.h#L45)?

build.sh
```diff
@@ -395,14 +401,18 @@ case "$COMMAND" in
...
-            kill $pid
-        done
+        kill_object_server
         ./sync/start-object-server.command
         exit 0
         ;;
 
+    "reset-object-server-between-tests")
+        kill_object_server
+        # Leave the server files alone to avoid 'bad_server_ident' errors
```
jpsim: Maybe add a fixme to remove when we target a release that includes realm/realm-sync#907? That may not be the same issue, I'm not sure.

build.sh
```diff
@@ -411,6 +421,8 @@ case "$COMMAND" in
                 break
             fi
         done
+        rm -rf "~/Library/Application Support/xctest"
```
jpsim: invoke `sh build.sh reset-object-server-between-tests` instead?

Realm/ObjectServerTests/RLMObjectServerTests.m
austinzheng: I just put 30 to be safe. I've never actually had problems with this timing out (or any of the other semaphores).

build.sh
austinzheng: I think that's a different issue. This issue is with sessions from older tests communicating with the server but finding out their server-side Realm files have been deleted from under them; I believe that issue is for multiple sessions being opened for the same file.

austinzheng:
I understand your concern with the high timeouts, but I'm afraid that "reasonable" timeouts will cause tests to sporadically fail when the CI machines are wedged or slow for whatever reason. What do you think?

build.sh
jpsim: Taking a step back, why is it necessary to kill and respawn the object server between each test now that credentials and realm URLs are unique for each test?

austinzheng: It's not necessary anymore. Do you want me to remove that?

jpsim: that would speed up the tests dramatically.

jpsim:
`testAddObjects` and `testLogBackInDeferredRealmDownload` take a disproportionately long time to run compared to other tests. Why do they take so long to run?

```
    ✓ testAddObjects (30.907 seconds)
    ✓ testAdminTokenAuthentication (0.081 seconds)
    ✓ testBasicUserPersistence (1.742 seconds)
    ✓ testDeleteObjects (3.588 seconds)
    ✓ testLogBackInDeferredRealmDownload (25.109 seconds)
    ✓ testLogBackInDeferredRealmUpload (5.101 seconds)
    ✓ testLogBackInOpenFirstTimePathDownload (5.336 seconds)
    ✓ testLogBackInOpenFirstTimePathUpload (5.087 seconds)
    ✓ testLogBackInReopenRealmDownload (5.158 seconds)
    ✓ testLogBackInReopenRealmUpload (5.170 seconds)
    ✓ testLogBackInSameRealmDownload (5.149 seconds)
    ✓ testLogBackInSameRealmUpload (5.024 seconds)
    ✓ testMultipleRealmsSessions (1.719 seconds)
    ✓ testOpenRealmWithAdminToken (0.086 seconds)
    ✓ testOpenRealmWithNormalCredential (1.667 seconds)
    ✓ testUsernamePasswordAuthentication (3.288 seconds)
```

austinzheng:
The first test always takes longer because it's responsible for setting up the sync server. I've seen the other test (and a few more) occasionally take ~20 seconds to run, I'm not sure what causes it and it's relatively rare. I've never seen that test actually fail since I made these changes, as well.

Realm/ObjectServerTests/RLMObjectServerTests.m
```diff
@@ -20,6 +20,23 @@
 
 #import "RLMSyncUser+ObjectServerTests.h"
 
+static NSString *reduceMethodName(NSString *fullName) {
```
jpsim: I think this is equivalent and a bit simpler:

```objc
#define ACCOUNT_NAME() NSStringFromSelector(_cmd)
#define CUSTOM_REALM_URL(realm_identifier) [NSURL URLWithString:[NSString stringWithFormat:@"realm://localhost:9080/~/%@%@", ACCOUNT_NAME(), realm_identifier]]
#define REALM_URL() CUSTOM_REALM_URL(@"")
```

build.sh
```diff
@@ -395,15 +401,19 @@ case "$COMMAND" in
...
-        # kill any object servers that are still running
-        (pgrep -f realm-object-server || true) | while read pid; do
-            kill $pid
-        done
+        kill_object_server
         ./sync/start-object-server.command
         exit 0
         ;;
 
+    "reset-object-server-between-tests")
```
jpsim: Why is this necessary now that each test runs in its own "isolated" environment? i.e. different Realms and user accounts.

build.sh
```diff
@@ -395,15 +401,19 @@ case "$COMMAND" in
...
+        # Leave the server files alone to avoid 'bad_server_ident' errors
+        rm -rf "~/Library/Application Support/xctest"
+        rm -rf "~/Library/Application Support/xctest-child"
+        exit 0
+        ;;
```
jpsim: missing newline after this

Realm/ObjectServerTests/RLMSyncTestCase.m
```diff
@@ -113,44 +183,67 @@ + (void)setUp {
...
+    }
 }
 
-- (void)tearDown {
++ (void)tearDown {
+    [s_task terminate];
+    [super tearDown];
+}
+
++ (void)runResetObjectServer {
```
jpsim: Why is this necessary now that each test runs in its own "isolated" environment? i.e. different Realms and user accounts.

build.sh
jpsim: Ok, that's fine given that this shouldn't be contributing to the long test run times.

build.sh
```diff
@@ -813,6 +825,7 @@ case "$COMMAND" in
 
     "verify-osx-object-server")
         sh build.sh download-object-server
+        sh build.sh reset-object-server
```
jpsim: I think this should be moved into `+[RLMSyncTestCase setUp]` because otherwise you won't kill previously running servers if you're running these tests via Xcode.

build.sh
```diff
@@ -284,6 +284,12 @@ fi
 # Downloading
 ######################################
 
+kill_object_server() {
+    (pgrep -f realm-object-server || true) | while read pid; do
+    kill $pid 2>/dev/null
```
jpsim: This should be indented. This is a while loop.

build.sh
```diff
@@ -284,6 +284,12 @@ fi
 # Downloading
 ######################################
 
+kill_object_server() {
+    (pgrep -f realm-object-server || true) | while read pid; do
```
jpsim: This won't kill `build.sh start-object-server` or `./sync/start-object-server.command`. Thoughts on changing this to `pgrep -f object-server`?

austinzheng:
The one test that failed has been flaky all day today, and has nothing to do with sync.

austinzheng:
After talking to @jpsim I'm going to squash and merge this PR. We can resolve any deficiencies in follow-up PRs.

PR #4176, opened 2016-10-03
Don't error on swiftlint already being up-to-date

tgoyne:
An example of a job that failed spuriously due to this: https://travis-ci.org/realm/realm-cocoa/jobs/164698561

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,7 +7,7 @@ matrix:
   include:
     - script: ./build.sh verify-swiftlint
       env: JOB=verify-swiftlint
-      before_install: brew update; brew upgrade swiftlint
+      before_install: brew update; brew upgrade swiftlint || true
     - script: ./build.sh verify-docs
       env: JOB=verify-docs
       before_install: gem install jazzy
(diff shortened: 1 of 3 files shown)


Realm/ObjectStore
```diff
@@ -1 +1 @@
-Subproject commit ed8813dc80eea2360f32f6a73893c1d296e03e61
```
austinzheng: This change looks spurious.

.travis.yml
```diff
@@ -7,7 +7,7 @@ matrix:
   include:
     - script: ./build.sh verify-swiftlint
       env: JOB=verify-swiftlint
-      before_install: brew update; brew upgrade swiftlint
+      before_install: brew update; brew upgrade swiftlint || true
```
bdash: We could instead only run `brew upgrade swiftlint` if it's outdated:

    brew update; brew outdated swiftlint || brew upgrade swiftlint

Realm/ObjectStore
```diff
@@ -1 +1 @@
...
+Subproject commit f2508684ce24d00321886ea3aa0a4d14b84a6ab8
```
bdash: This also appears to refer the object store to an older commit.

PR #4177, opened 2016-10-03
Fix a compilation error and warnings when building the examples

tgoyne:
Compiling RACTableView failed when using CocoaPods 1.1, seemingly related to changes between 0.39 and 1.1, which were fixed by a deintegrate and re-setup.

The examples had a bunch of assorted warnings that probably didn't cause any actual problems.

The main Realm xcodeproj had a bunch of headers marked as Private when they should have been just not included in any targets at all, and some of them were duplicated as well, which lead to some warnings when building the examples.


tgoyne:
@jpsim [objc_test_examples](https://ci.realm.io/job/objc_test_examples/1297/) now passes on this branch. The remaining CI failures for this PR are fixed by https://github.com/realm/realm-cocoa/pull/4176.

jpsim:
👍

PR #4188, opened 2016-10-05
Retry building installation examples if they hit an xcodebuild internal error

tgoyne:
Attempts to work around failures such as [this one](https://ci.realm.io/job/Cocoa%20Installation%20Examples/433/binding=objc,method=static,platform=ios/console) where it never actually starts to build the example.


tgoyne:
This didn't actually work; it just swallowed the failure rather than retrying.

jpsim:
Marked as `S:On Hold` until it can be fixed.

jpsim:
This never worked. We'll prioritize fixing if it becomes an issue again.

PR #4196, opened 2016-10-07
Fix some crashes when working with a RLMLinkingObjects property of an unmanaged object

### LINKED ISSUE #4195: Linking Object attributes don't exist when not in realm
Upgraded to Realm 2.0.1 from 1.X and been having some crashes when adding new objects.

It looks like an object with linking properties crashes if i try to access its linking objects before it's added to realm. I think in 1.X the behavior was to return an empty list.

Did this behavior change? What's expected here:?

bdash:
A `RLMLinkingObjects` property of an unmanaged object is the only case in which a `RLMResults` instance has a null `_info` pointer. We were not adequately ensuring that the `_info` pointer is not accessed in such cases.

Fixes #4195.

/cc @jpsim @tgoyne @austinzheng

### DIFF
--- a/Realm/RLMResults.mm
+++ b/Realm/RLMResults.mm
@@ -138,10 +138,6 @@ static inline void RLMResultsValidateInWriteTransaction(__unsafe_unretained RLMR
     return RLMStringDataToNSString(_results.get_object_type());
 }
 
-- (RLMObjectSchema *)objectSchema {
-    return _info->rlmObjectSchema;
-}
-
 - (RLMClassInfo *)objectInfo {
     return _info;
 }
@@ -149,6 +145,10 @@ static inline void RLMResultsValidateInWriteTransaction(__unsafe_unretained RLMR
 - (NSUInteger)countByEnumeratingWithState:(NSFastEnumerationState *)state
                                   objects:(__unused __unsafe_unretained id [])buffer
                                     count:(NSUInteger)len {
+    if (!_info) {
+        return 0;
+    }
+
     __autoreleasing RLMFastEnumerator *enumerator;
     if (state->state == 0) {
         enumerator = [[RLMFastEnumerator alloc] initWithCollection:self objectSchema:*_info];
@@ -376,6 +376,9 @@ static inline void RLMResultsValidateInWriteTransaction(__unsafe_unretained RLMR
 }
 
 - (id)aggregate:(NSString *)property method:(util::Optional<Mixed> (Results::*)(size_t))method methodName:(NSString *)methodName {
+    if (_results.get_mode() == Results::Mode::Empty) {
+        return nil;
+    }
     size_t column = _info->tableColumn(property);
     auto value = translateErrors([&] { return (_results.*method)(column); }, methodName);
     if (!value) {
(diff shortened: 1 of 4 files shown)


Realm/Tests/LinkingObjectsTests.mm
```diff
@@ -53,6 +53,37 @@ - (void)testBasics {
...
+        (void)parent;
+        XCTFail();
+    }
+
+    XCTAssertEqual(0u, [don.parents sortedResultsUsingProperty:@"age" ascending:YES].count);
+    XCTAssertEqual(0u, [don.parents objectsWhere:@"TRUEPREDICATE"].count);
+
+    XCTAssertNil([don.parents minOfProperty:@"age"]);
+    XCTAssertNil([don.parents maxOfProperty:@"age"]);
+    XCTAssertNil([don.parents sumOfProperty:@"age"]);
```
tgoyne: sumOfProperty: does not return a nullable value.

Realm/RLMResults.mm
```diff
@@ -376,6 +376,9 @@ - (id)objectAtIndexedSubscript:(NSUInteger)index {
 }
 
 - (id)aggregate:(NSString *)property method:(util::Optional<Mixed> (Results::*)(size_t))method methodName:(NSString *)methodName {
+    if (_results.get_mode() == Results::Mode::Empty) {
```
tgoyne: This is incorrect for `sum` (the sum of zero items is zero, not undefined), and it should just be checking for _info being null as that's the thing that actually needs to be guarded against, not calling the aggregate functions on an Empty Results.

bdash: You're right that sum should be returning zero in this case. It looks like `Results::aggregate`, which is what I looked at when checking what the behavior should be, also gets this wrong for the `Mode::Empty` case. It is not respecting the `return_none_for_empty` argument as it should be.

bdash: > it should just be checking for _info being null as that's the thing that actually needs to be guarded against

In a number of other functions we happen to not crash by virtue of a similar check for the results being in the `Empty` mode. It seemed cleanest to rely on the mode being `Empty` when `_info` is null, rather than checking both in all cases.

tgoyne: The places where we're currently checking the mode are where calling the functions on an Empty Results would be a precondition violation (e.g. get_query() requires that the Results be non-Empty rather than returning an Optional Query). In this case, calling the aggregate methods on an Empty results is perfectly valid, and is not the thing that needs to be guarded against and is merely correlated with the thing that needs to be guarded against.

Functionally it doesn't matter, but my reaction to seeing a check for Empty was to go look at the Results code  to see why calling the functions on an Empty results was incorrect, while checking `_info` would have made it immediately obvious why the check was being performed.

bdash: `-indexOfObjectWithPredicate:` having the check to avoid calling `Results::get_query` seems reasonable enough.

`-objectsWithPredicate:` and `-sortedResultsUsingDescriptors:` aren't consistent with that logic though. It should be perfectly valid to call `Results::filter` or `Results::sort` on an empty `Results`, yet they first check for `Mode::Empty`. ISTM that we should fix `Results::sort` and `Results::filter` to work when empty (presumably by returning another empty `Results`), and then switch `-objectsWithPredicate:` and `-sortedResultsUsingDescriptors:` to checking for `_info` being null instead.

bdash: Ah, the two methods in question deal with `Mode::Empty` because they need to work with a table prior to calling into `Results` (to prepare the query and sort descriptors respectively), and that table may not exist. So they are consistent with the logic you outlined, but just for a slightly different reason.

jpsim:
I'll rebase this and merge unless CI flags something wrong.

Issue #4205, opened 2016-10-11
Warnings when building with static Realm.framework


zpapp:
## Goals

I'd like to build my iOS app with the static Objective-C Realm framework.

## Expected Results

Build succeeds without warnings.

## Actual Results

I get the following warnings:

```
while processing /Users/zpapp/tmp.d/StaticRealmWarnings/StaticRealmWarnings/Realm.framework/Realm(RLMAuthResponseModel.o):
warning: /Users/realm/workspace/objc_ios/tightdb_objc/build/DerivedData/ModuleCache/1SLMU8SXHKS7C/Foundation-3JWVXIADENRHH.pcm: No such file or directory
note: Linking a static library that was built with -gmodules, but the module cache was not found.  Redistributable static libraries should never be built with module debugging enabled.  The debug experience will be degraded due to incomplete debug information.
while processing /Users/zpapp/tmp.d/StaticRealmWarnings/StaticRealmWarnings/Realm.framework/Realm(RLMAuthResponseModel.o):
warning: /Users/realm/workspace/objc_ios/tightdb_objc/build/DerivedData/ModuleCache/1SLMU8SXHKS7C/ObjectiveC-26VMO97172RKZ.pcm: No such file or directory
while processing /Users/zpapp/tmp.d/StaticRealmWarnings/StaticRealmWarnings/Realm.framework/Realm(RLMAuthResponseModel.o):
warning: /Users/realm/workspace/objc_ios/tightdb_objc/build/DerivedData/ModuleCache/JWC8IYTTLHHV/Foundation-3JWVXIADENRHH.pcm: No such file or directory
note: Linking a static library that was built with -gmodules, but the module cache was not found.  Redistributable static libraries should never be built with module debugging enabled.  The debug experience will be degraded due to incomplete debug information.
while processing /Users/zpapp/tmp.d/StaticRealmWarnings/StaticRealmWarnings/Realm.framework/Realm(RLMAuthResponseModel.o):
warning: /Users/realm/workspace/objc_ios/tightdb_objc/build/DerivedData/ModuleCache/JWC8IYTTLHHV/ObjectiveC-26VMO97172RKZ.pcm: No such file or directory
```

Note: Issue #2642 is similar to this. That issue was closed because the problem went away. That was true at the time; I am pretty sure that for example Realm version 1.0.1 did not have this problem.

## Steps to Reproduce

1. Create a new iOS project with Xcode 8. Select Objective-C for language.

[...]

PR #4209, opened 2016-10-11
Remove header files from target membership

austinzheng:
@jpsim 

The only private header files that should be added to the `Realm` and `Realm iOS static` targets are those used by the private module. This PR removes private header files not used by the module from those targets.


austinzheng:
@jpsim Disabling module debugging for the static library (like the error message recommends) fixed the problem. (This isn't disabling debugging, period; it's just changing the way debug information is emitted. See: http://stackoverflow.com/questions/36265490/turning-off-clang-enable-module-debugging-consequences)

Fixes: #4205.

PR #4210, opened 2016-10-11
recover from "operation never finished bootstrapping" CI errors

jpsim:
Like the one that happened in https://ci.realm.io/job/objc_pr/configuration=Release,swift_version=3.0,target=ios-static/4193/console /cc @austinzheng 

```
2016-10-11 13:59:48.895 xcodebuild[20525:2232979] Error Domain=IDETestOperationsObserverErrorDomain Code=5 "Early unexpected exit, operation never finished bootstrapping - no restart will be attempted" UserInfo={NSLocalizedDescription=Early unexpected exit, operation never finished bootstrapping - no restart will be attempted}

Testing failed:
	Test target iOS static tests encountered an error (Early unexpected exit, operation never finished bootstrapping - no restart will be attempted)
** TEST FAILED **
```

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -1037,7 +1037,7 @@ EOM
 
             failed=0
             sh build.sh verify-$target 2>&1 | tee build/build.log | xcpretty -r junit -o build/reports/junit.xml || failed=1
-            if [ "$failed" = "1" ] && cat build/build.log | grep -E 'DTXProxyChannel|DTXChannel|out of date and needs to be rebuilt'; then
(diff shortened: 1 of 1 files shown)


jpsim:
> What does it do right now? Hang forever, or just complain and abort?

Currently if the tooling fails with that error message, CI treats it the same as our build or tests failing, even though it's not a failure of ours, but rather Xcode. Hence the failure in https://ci.realm.io/job/objc_pr/configuration=Release,swift_version=3.0,target=ios-static/4193/console

This PR makes CI re-run the job once if it failed for this reason, since it's not something we have any control over.

PR #4219, opened 2016-10-13
Refactor binding to move sync functionality into object store

austinzheng:
Depends upon https://github.com/realm/realm-object-store/pull/193

Changes:
- All non-trivial `RLMSync*` types are now thin wrappers around their new object store counterparts
- Removed the file manager
- Removed the session handles
- Added a 'refresh handle', a helper object managed by a `RLMSyncUser` that does nothing but periodically refresh a session as long as that session remains live
- Updated existing unit tests to work with the new code


build.sh
```diff
@@ -1004,6 +1004,8 @@ EOM
           mkdir -p include/util/apple
           cp Realm/*.hpp include
           cp Realm/ObjectStore/src/*.hpp include
+          cp Realm/ObjectStore/src/sync/*.hpp include/sync
```
jpsim: These two lines will fail because `include/sync` doesn't exist. The `mkdir -p` lines above can be combined into this, which will fix the issue: `mkdir -p include/impl/apple include/util/apple include/sync/impl`

Realm/ObjectServerTests/RLMObjectServerTests.m
```diff
@@ -100,7 +100,11 @@ - (void)testOpenRealmWithAdminToken {
     RLMSyncUser *user = [self logInUserForCredential:credential
                                               server:[RLMObjectServerTests authServerURL]];
     NSURL *url = [NSURL URLWithString:@"realm://localhost:9080/testSyncWithAdminToken"];
-    RLMRealm *realm = [self openRealmForURL:url user:user];
+    RLMRealmConfiguration *c = [[RLMRealmConfiguration defaultConfiguration] copy];
```
jpsim: `[RLMRealmConfiguration defaultConfiguration]` already returns a copy: https://github.com/realm/realm-cocoa/blob/v2.0.3/Realm/RLMRealmConfiguration.mm#L63

Realm/ObjectServerTests/RLMObjectServerTests.m
```diff
@@ -41,8 +41,8 @@ - (void)testUsernamePasswordAuthentication {
     RLMSyncUser *secondUser = [self logInUserForCredential:[RLMSyncTestCase basicCredentialWithName:ACCOUNT_NAME()
                                                                                            register:NO]
                                                     server:[RLMSyncTestCase authServerURL]];
-    // Logging in with equivalent credentials should return the same user object instance.
-    XCTAssertEqual(firstUser, secondUser);
+    // Two users created with the same credential should resolve to the same actual user.
+    XCTAssertTrue([firstUser.identity isEqualToString:secondUser.identity]);
```
jpsim: Should we override `-[RLMSyncUser isEqual:]` to do this?

Realm/RLMRealm.mm
```diff
@@ -204,6 +204,16 @@ REALM_NOINLINE void RLMRealmTranslateException(NSError **error) {
             case RealmFileException::Kind::Exists:
                 RLMSetErrorOrThrow(RLMMakeError(RLMErrorFileExists, ex), error);
                 break;
+            case RealmFileException::Kind::BadHistoryError: {
+                NSString *err = @"Realm file's history format is incompatible with the "
```
jpsim: why not set this error message in the object store?

jpsim: or even better, core?

Realm/RLMSyncConfiguration.mm
```diff
@@ -35,51 +49,50 @@ static BOOL isValidRealmURL(NSURL *url) {
...
+- (RLMSyncUser *)user {
+    return [[RLMSyncUser alloc] initWithSyncUser:_config->user];
+}
+
+- (RLMSyncStopPolicy)stopPolicy {
+    return translateStopPolicy(_config->stop_policy);
+}
+
+- (NSURL *)realmURL {
+    NSString *rawStringURL = @(_config->realm_url.c_str());
```
jpsim: minor quibble, but this intermediate variable really isn't necessary.

Realm/RLMSyncManager_Private.h
```diff
@@ -0,0 +1,54 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+////////////////////////////////////////////////////////////////////////////
+
+#import "RLMSyncManager.h"
+
+#import "RLMSyncUtil_Private.h"
+
+typedef NS_ENUM(NSUInteger, RLMSyncSessionErrorKind) {
```
jpsim: should we make this public?

jpsim:
👍 other than my current comments.

jpsim:
@tgoyne @bdash can one of you please review this too?

Realm/RLMRealmConfiguration+Sync.mm
```diff
@@ -43,11 +42,12 @@ - (void)setSyncConfiguration:(RLMSyncConfiguration *)syncConfiguration {
     // Ensure sync manager is initialized, if it hasn't already been.
     [RLMSyncManager sharedManager];
     NSAssert(user.identity, @"Cannot call this method on a user that doesn't have an identity.");
-    NSURL *localFileURL = [[[RLMSyncManager sharedManager] fileManager] fileURLForRawRealmURL:realmURL user:user];
     if (syncConfiguration.customFileURL) {
-        localFileURL = syncConfiguration.customFileURL;
+        self.config.path = syncConfiguration.customFileURL.path.UTF8String;
+    } else {
+        self.config.path = std::move(SyncManager::shared().path_for_realm([user.identity UTF8String],
```
bdash: It's not necessary to `std::move` a temporary.

Realm/RLMSyncConfiguration.mm
```diff
@@ -35,51 +49,50 @@ static BOOL isValidRealmURL(NSURL *url) {
 }
 
 @interface RLMSyncConfiguration () {
-    std::function<realm::SyncSessionErrorHandler> _error_handler;
+    std::unique_ptr<realm::SyncConfig> _config;
```
bdash: Why heap allocate the configuration?

Realm/RLMSyncSession.mm
```diff
@@ -16,212 +16,107 @@
...
-                            user:(RLMSyncUser *)user
-                          handle:(RLMSyncSessionHandle *)handle {
-    self.parentUser = user;
-    self.accessToken = token;
-    self.accessTokenExpiry = expiry;
-    self.sessionHandle = handle;
-    [self _scheduleRefreshTimer];
+    auto this_ptr = _session.lock();
+    auto that_ptr = ((RLMSyncSession *)object)->_session.lock();
+    return this_ptr && that_ptr && (this_ptr == that_ptr);
```
bdash: This logic is a little odd. If you have two `RLMSyncSession`s that wrap the same session they will compare equal, but if the underlying session goes away the `RLMSyncSession`s will no longer compare equal?

Realm/RLMSyncSession.mm
```diff
@@ -16,212 +16,107 @@
...
-    else {
-        [self _invalidate];
+- (BOOL)waitForUploadCompletionOnQueue:(dispatch_queue_t)queue callback:(void(^)(void))callback {
+    if (auto session = _session.lock()) {
+        if (!session->is_valid()) {
+            return NO;
+        }
+        queue = queue ?: dispatch_get_main_queue();
+        session->wait_for_upload_completion([=](){
+            dispatch_async(queue, ^{
```
tgoyne: There's no reason to wrap the block in another block here.

jpsim: flattened

jpsim:
@bdash do you have any further requests for modifications to this PR?

Issue #1120, opened 2014-11-12
Support storing arrays of primitives (string, int, etc.) as RLMObject properties


andrewjl:
Reading over issue #1028 it looks like RLMObject doesn't support properties that are arrays of non-RLMObjects. Instead the individual strings each have to be declared as a property of an RLMObject and then stored in an RLMArray.

I'm curious to understand why your support for primitives doesn't extend to collections of objects of those types. As far as Realm is concerned, what's the difference between storing a string or an array of strings? And would you ever add this support to a future release?

---

Subtasks:
- [x] Non-breaking changes for arrays of primitives (#5148)
- [x] Breaking API changes for arrays of primitives (#5149)
- [x] Basic support for arrays of primitives (#5274)
- [ ] Query support for arrays of primitives

michaelforrest:
+1

andrewjl:
@jpsim based on our chat in person (at the carthage event) it sounds like the best solution would be to do what Realm already requires the user to do automatically, and transparently.

Seems like the best place to do this would be in `RLMProperty`, inside of `setTypeFromRawType`. Instead of throwing an exception when `RLMIsObjectSubclass` returns false, we instead check the `_objectClassName` against a white list of class names (Just `NSString`,and `NSNumber` would be a good start) and then do what Realm would usually do to create with an `RLMObject` the user would provide as the inner class of the array.

Let me know if you think this is in the right direction. Also happy to provide more details if this is too high level.

andrewjl:
More thoughts... when we detect a white listed inner `RLMArray` class we can handle it in one of two ways:

* Use the objc runtime to create an `RLMObject` subclass with a single property of the white listed inner class type. The upside is that we can then pass this dynamically generated class into Realm and it will do the right thing. The downside is that this won't work for Swift (as I don't believe it's possible to create classes on the fly in Swift) and also doing fancy things with the runtime may cause breakages elsewhere. (And is in general not recommended.)

* We create a new table inside the schema that will store the array's values and write all the querying, updating, and other logic needed to manipulate the table outside of `RLMObject`. Not sure how much work this entails, but it sounds safer than the first approach.

startupthekid:
I'm a little late to the party but I'd like to offer some thoughts.

[...]

PR #4227, opened 2016-10-16
RM-1: Reworked flow of adding object a bit, added wrappers around primitives

IljaKosynkin:
In my recent project in which my team is using Realm, I've meet few things that seems to be unintuitive for me. I tried my best to make it more clean and intuitive and hope that my idea will pass into project at some point :)
First of all, I believe way in which function `realm.add(object, update: false)` is a bit misleading. When I saw this function for first time, I thought that it's actually skipping object, if exists, without any changes, while `realm.add(object, update: true)` is updating existing object. Later I realised that `realm.add(object, update: false)` is actually throwing exception if object with such primary key already exists, what was a bit frustrating for me. The first change that I want to provide is to make `realm.add(object, update: false)` just skip object instead of throwing error if exists. It's quite critical for my project, since user of application can fire multiple requests at time and objects in result can already exists in database, but we don't want to update them, since it could overwrite local changes. 
Second thing: wrappers around primitive types (Int, Double and String for now) for List<T> to provide builtin possibility of saving list of primitive types  in Realm. Yes, it's not a big deal to write such thing for particular project, but I feel that it would be nice to have such possibility out of the box :)
I hope that I correctly read down the code and documentation and didn't duplicate features that already exists somewhere in Realm :)

### DIFF
--- a/Realm.xcodeproj/project.pbxproj
+++ b/Realm.xcodeproj/project.pbxproj
@@ -413,6 +413,8 @@
 		5DD755CA1BE056DE002800DA /* schema.hpp in Headers */ = {isa = PBXBuildFile; fileRef = 3FE556431B9A43E5002A1129 /* schema.hpp */; };
 		5DD755CB1BE056DE002800DA /* shared_realm.hpp in Headers */ = {isa = PBXBuildFile; fileRef = 3FAE25541B8CEBBE00D01405 /* shared_realm.hpp */; };
 		5DD755E21BE05DAF002800DA /* Realm.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = 5DD755CF1BE056DE002800DA /* Realm.framework */; };
+		5F3E8AB91DB20F0700B525CB /* Primitive.swift in Sources */ = {isa = PBXBuildFile; fileRef = 5F3E8AB81DB20F0700B525CB /* Primitive.swift */; };
+		5F8BB3EA1DB3A76C006FAADD /* SwiftObjectAddingTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 5F8BB3E81DB3A76C006FAADD /* SwiftObjectAddingTests.swift */; };
 		C042A48D1B7522A900771ED2 /* RealmConfigurationTests.mm in Sources */ = {isa = PBXBuildFile; fileRef = C042A48C1B7522A900771ED2 /* RealmConfigurationTests.mm */; };
 		C042A48E1B7522A900771ED2 /* RealmConfigurationTests.mm in Sources */ = {isa = PBXBuildFile; fileRef = C042A48C1B7522A900771ED2 /* RealmConfigurationTests.mm */; };
 		C0CDC0821B38DABA00C5716D /* UtilTests.mm in Sources */ = {isa = PBXBuildFile; fileRef = 021A88311AAFB5BE00EEAC84 /* UtilTests.mm */; };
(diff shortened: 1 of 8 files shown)


jpsim:
Hi @IljaKosynkin. We really appreciate the PR and the time you've put in to preparing this.

Regarding your first change: updating the behavior of `Realm.add(...)`, for starters this is a breaking change in semantics that could break lots of existing apps out there.

For example, `Realm.add(_:update: false)` currently always adds a new object. Some apps rely on this invariant behavior of Realm. For every user that would prefer `realm.add(_:update: false)` just be a no-op if the object already exists, there's another that relies on the opposite being the case.

We take making breaking changes very seriously, and therefore this is something we're unlikely to change. Especially given that it's trivial to obtain the behavior you'd like in your code by combining two very concise existing Realm APIs:

```swift
extension Realm {
  func addIfNonExistent<T: Object>(object: T) {
    if object(ofType: T.Type, forPrimaryKey: object.primaryKey) == nil {
      add(object)
    }
  }
}
```

As for your second proposal, adding a Realm model for each primitive type supported by Realm is also something we've thought about and ultimately rejected, because it has unfortunate consequences. For example, every app that uses Realm would now have these tables in the database, regardless of whether or not they use the functionality enabled by these wrappers. Also, these wrappers are trivial to implement in your own apps. Finally, it's a temporary workaround for a limitation, mostly preventing lists of primitives (#1120). We'd much rather remove the underlying limitation rather than introduce a temporary type. Adding these wrappers as a temporary measure would also require us to have a clear path to deprecation and migration to the "fixed" model once it's available. For all these reasons, we've decided not to move forward with that approach.

---

In the future, if you'd like to propose changes, I suggest you comment on an existing open issue if a relevant one already exists, or open a new one otherwise, where we can discuss the implications of those changes before you devote time to implementing them. Our team might share some edge cases you may not have considered, and conversely, you might open our eyes to an approach we haven't yet considered.

I hope the reasons I've provided here help clarify why we won't be accepting these changes and I sincerely thank you for your interest in contributing to Realm!

IljaKosynkin:
Hi @jpsim.
It's a bit strange for me, since I spent quite a bit of time thinking about providing such changes and didn't found any breaking changes in it. Let me clarify my point with a table:  

Method | Current behaviour | Behaviour in my PR
--------|--------------------|-------------------
add(Object) (not exists) | Adds object to database | Adds object to database
add(Object) (exists) | Throws exception | Throws exception
add(Object, update: true) (not exists) | Adds object to database | Adds object to database
add(Object, update: true) (exists) | Updates object in database | Updates object in database 
add(Object, update: false) (not exists) | Adds object to database | Adds object to database
add(Object, update: false) (exists) | Throws exception | **Skipping object with no changes and no exception thrown**

As you can see, changes was made for one particular use case and it's hard for me to imagine any app that is hardly rely on a fact that add(Object, update: false) will throw an exception if object exists. I will be very grateful, if you can provide me real-world use case, when my change will really broke down something fatally. 
About your solution with checking if object exists - yes, it's not hard to make such change, but it will provide additional layer above database, as well as give overhead in object searching. As I can see in Realm code - when somebody wants to add object, Realm is performing search of row by ID. With your code such search will be performed twice: in my code at first and then internally in Realm. May be it doesn't seems to be a big deal, but I don't see any point in doubling of operations either, especially since our application is supposed to download thousands of objects from our API and write it to the database. 

As about wrappers - well, I can see your point. However, I feel like such limitation on List<> was here for quite a bit of time, so may be it's not a bad idea to give users of library a built-in solution, while you guys are working on correct solution? Also, I think it's possible to map Int to RealmInt internally, i.e. List<Int> would be mapped to List<RealmInt> inside Realm and problem is basically resolved, no? Yes, once again it's not hard to implement in application all this wrappers, but I personally hate boiler-plate code and I'm trying to avoid it as much as possible. I will not feel happy if I will be forced to repeat wrappers over and over again in every project that is using Realm or write separate library for this, which will require additional time spent on its maintenance. 
Hopefully, I made my point clear. I completely understand you concerns about my PR, but I doesn't feel like it's really something big :) Thank you for spending time on reviewing my code and I hope  you would think over about it once again :)

jpsim:
@IljaKosynkin just FYI, I haven't forgotten you! Reviewing and processing your last comment has been on my todo list since you posted this. I'll get to this... eventually.

PR #4233, opened 2016-10-18
Reenable some disabled object server tests

austinzheng:
@jpsim 

Kicking this off to see what CI thinks...

Changes:
- Re-enabled Swift tests
- Re-enabled some disabled Objective-C tests
- Base test case class no longer terminates server process upon tearDown

### DIFF
--- a/Realm/ObjectServerTests/RLMSyncTestCase.m
+++ b/Realm/ObjectServerTests/RLMSyncTestCase.m
@@ -97,14 +97,18 @@ static NSURL *syncDirectoryForChildProcess() {
     NSAssert(realms.count == counts.count && realms.count == realmURLs.count,
              @"Test logic error: all array arguments must be the same size.");
     XCTestExpectation *checkCountExpectation = [self expectationWithDescription:@"Downloads should complete"];
-    for (NSUInteger i = 0; i < realms.count; i++) {
-        WAIT_FOR_DOWNLOAD(user, realmURLs[i]);
-    }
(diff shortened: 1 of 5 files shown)


Realm/ObjectServerTests/RLMSyncTestCase.m
```diff
@@ -97,14 +97,16 @@ - (void)waitForDownloadsForUser:(RLMSyncUser *)user
...
-    for (NSUInteger i = 0; i < realms.count; i++) {
-        WAIT_FOR_DOWNLOAD(user, realmURLs[i]);
-    }
     dispatch_async(dispatch_get_main_queue(), ^{
         for (NSUInteger i = 0; i < realms.count; i++) {
-            CHECK_COUNT([counts[i] integerValue], SyncObject, realms[i]);
+            WAIT_FOR_DOWNLOAD(user, realmURLs[i]);
         }
-        [checkCountExpectation fulfill];
+        dispatch_async(dispatch_get_main_queue(), ^{
```
jpsim: How does this `dispatch_async` help? Seems like refreshing the realms that need an update would be a more appropriate way to do this?

austinzheng: Maybe. I'm kind of grasping at straws, since the failure has been impossible to reproduce outside of the Jenkins CI environment :/.

jpsim: Can you at least add a FIXME or file an issue?

PR #4253, opened 2016-10-24
Add support for suppressing specific notifications from writes

### LINKED ISSUE #3665: Add write transactions which do not produce local notifications
The current fine-grained notifications approach makes it challenging to support scenarios where the UI is updated before the data model, rather than the reactive approach of the model being the canonical source of truth that the UI merely displays. The simplest motivating use-case is UITableView reordering; when using the built-in functionality for dragging rows around the delegate is merely notified of changes already displayed in the UI, and re-applying those changes in the collection notification block after the Realm is committed will corrupt the order to the UITV. Because notifications are coalesced, users can't reasonably filter out these redundant updates themselves. and even if they could it would be very easy to get wrong.

We should expose a variant of `Realm.write(_:)` that would specify that the transaction should not produce notifications on the current thread, as the things they want to update with the notifications are already up-to-date. In addition, this variant of write transactions would block until all async notifications are ready and delivers them, since otherwise the UI could easily get "out of sync".

An alternative approach would be to instead let the user pass in the tokens for notification blocks which they want to suppress. This is more flexible and makes it work with commits on a different thread from the notified thread, but maybe would make the API even more complex.

@tgoyne has implemented a proof of concept of this in #3439, and there's more discussion in that thread, but that work is currently on hold.

tgoyne:
Obj-c/Swift side of https://github.com/realm/realm-object-store/pull/200. This is mostly just tests and doc updates since everything interesting is done in the object store.

Closes #3665.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,11 +7,18 @@ x.x.x Release notes (yyyy-MM-dd)
 
 ### Enhancements
 
-* None.
+* Add the ability to skip calling specific notification blocks when committing
+  a write transaction.
 
 ### Bugfixes
 
-* None.
+* Deliver collection notifications when beginning a write transaction which
+  advances the read version of a Realm (previously only Realm-level
+  notifications were sent).
+* Fix some scenarios which would lead to inconsistent states when using
+  collection notifications.
+* Fix several race conditions in the notification functionality.
+* Don't send Realm change notifications when canceling a write transaction.
 
 2.0.4 Release notes (2016-11-14)
 =============================================================
--- a/Realm/RLMRealm.h
+++ b/Realm/RLMRealm.h
@@ -34,7 +34,7 @@ NS_ASSUME_NONNULL_BEGIN
  objects (for example, by using the same path or identifier) multiple times on a single thread
  within a single iteration of the run loop will normally return the same
  `RLMRealm` object.
- 
+
  If you specifically want to ensure an `RLMRealm` instance is
  destroyed (for example, if you wish to open a Realm, check some property, and
  then possibly delete the Realm file and re-open it), place the code which uses
@@ -61,7 +61,7 @@ NS_ASSUME_NONNULL_BEGIN
  default Realm is persisted as *default.realm* under the *Documents* directory of
  your Application on iOS, and in your application's *Application Support*
  directory on OS X.
- 
+
  The default Realm is created using the default `RLMRealmConfiguration`, which
  can be changed via `+[RLMRealmConfiguration setDefaultConfiguration:]`.
 
@@ -118,7 +118,7 @@ NS_ASSUME_NONNULL_BEGIN
 
 /**
  The type of a block to run whenever the data within the Realm is modified.
- 
+
  @see `-[RLMRealm addNotificationBlock:]`
  */
 typedef void (^RLMNotificationBlock)(RLMNotification notification, RLMRealm *realm);
@@ -130,10 +130,11 @@ typedef void (^RLMNotificationBlock)(RLMNotification notification, RLMRealm *rea
 
  Notification handlers are called after each write transaction is committed,
  either on the current thread or other threads.
- 
- Handler blocks are called on the same thread that they were added on, and may only be added on threads which are
- currently within a run loop. Unless you are specifically creating and running a run loop on a background thread, this
- will normally only be the main thread.
+
+ Handler blocks are called on the same thread that they were added on, and may
+ only be added on threads which are currently within a run loop. Unless you are
+ specifically creating and running a run loop on a background thread, this will
+ normally only be the main thread.
 
  The block has the following definition:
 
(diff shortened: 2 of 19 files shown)


RealmSwift/Tests/RealmCollectionTypeTests.swift
```diff
@@ -499,7 +499,24 @@ class RealmCollectionTypeTests: TestCase {
...
+        let token2 = collection.addNotificationBlock { (changes: RealmCollectionChange) in
+            theExpectation.fulfill()
+        }
+        waitForExpectations(timeout: 1, handler: nil)
+
+        // make a write and implicitly verify that only the unskipped
+        // notification is called (the first would error on .update)
+        theExpectation = expectation(description: "")
+        let realm = realmWithTestPath()
+        realm.beginWrite();
```
jpsim: this and the 3 other `beginWrite()`s have trailing semicolons

jpsim:
Are any of the failing CI jobs relevant?

tgoyne:
The interprocess test failure might be spurious or might be related to these changes.

CHANGELOG.md
```diff
@@ -61,7 +61,8 @@ GM seed.
 
 ### Enhancements
 
-* None.
+* Add the ability to skip calling specific notification blocks when committing
```
jpsim: Wrong changelog section

tgoyne:
The remaining CI failures here are unrelated (SwiftLint errors are for files not modified in this PR, one job had a test failure that has rarely occurred in the past without these changes, and other two were just the simulator failing to boot).

jpsim:
> SwiftLint errors are for files not modified in this PR

There are no SwiftLint errors according to Jenkins: https://ci.realm.io/job/objc_pr/4338/configuration=Release,swift_version=3.0.1,target=swiftlint/console

CHANGELOG.md
jpsim: Fixed.

CHANGELOG.md
```diff
@@ -75,6 +76,13 @@ GM seed.
   of process termination.
 * Fix a crash that could occur when working with a `RLMLinkingObject` property
   of an unmanaged object.
+* Deliver collection notifications when beginning a write transaction which
```
jpsim: These are still in the wrong changelog section though.

tgoyne:
The [Travis Swiftlint check](https://travis-ci.org/realm/realm-cocoa/jobs/176548751) failed. I assume it's running a different version of swiftlint.

jpsim:
> The Travis Swiftlint check failed. I assume it's running a different version of swiftlint.

I see! No, it's actually running with Swift 2.2 whereas Jenkins just runs SwiftLint on 3.0.1.

Realm/RLMRealm.h
```diff
@@ -202,6 +224,40 @@ typedef void (^RLMNotificationBlock)(RLMNotification notification, RLMRealm *rea
 - (BOOL)commitWriteTransaction:(NSError **)error;
 
 /**
+ Commits all write operations in the current write transaction, without
+ notifying specific callbacks of the changes.
```
jpsim: "callbacks" -> "notification blocks"

Realm/RLMRealm.h
```diff
@@ -258,35 +318,42 @@ typedef void (^RLMNotificationBlock)(RLMNotification notification, RLMRealm *rea
...
  even if `autorefresh` is set to `YES`.
 
- Even with this property enabled, you can still call `-refresh` at any time to update the
- Realm before the automatic refresh would occur.
+ Even with this property enabled, you can still call `-refresh` at any time to
+ update the Realm before the automatic refresh would occur.
 
- Notifications are sent when a write transaction is committed whether or not
- automatic refreshing is enabled.
+ Write transaction will still always advance a Realm to the latest version and
```
jpsim: "Write transaction" -> "Write transactions"

Realm/RLMRealm.mm
```diff
@@ -70,6 +70,13 @@ - (void)stop {
     _block = nil;
 }
 
+- (void)suppressNextNotification {
+    auto block = _block;
```
jpsim: A brief comment or different variable name might help explain what this is doing.

It wasn't immediately clear to me that this approach to suppressing notifications completely bypasses the Object Store's mechanism for suppressing notifications.

Realm/Tests/NotificationTests.m
```diff
@@ -132,6 +132,66 @@ - (void)testNonMatchingObjectMovedToIndexOfMatchingRowAndMadeMatching {
...
+    RLMRealm *realm = [RLMRealm defaultRealm];
+    RLMNotificationToken *token = [realm addNotificationBlock:^(__unused RLMNotification notification, __unused RLMRealm *realm) {
+        XCTFail(@"should not have been called");
+    }];
+
+    [realm beginWriteTransaction];
+    [realm deleteAllObjects];
+    [realm commitWriteTransactionWithoutNotifying:@[token] error:nil];
+
+    // local realm notifications are called synchronously so no need to wait for anything
```
jpsim: `[token stop]` to prevent the deallocation warning from being logged?

Realm/Tests/NotificationTests.m
```diff
@@ -132,6 +132,66 @@ - (void)testNonMatchingObjectMovedToIndexOfMatchingRowAndMadeMatching {
...
+- (void)testSuppressRealmNotificationForWrongRealm {
+    RLMRealm *otherRealm = [self realmWithTestPath];
+    RLMNotificationToken *token = [otherRealm addNotificationBlock:^(__unused RLMNotification notification, __unused RLMRealm *realm) {
+        XCTFail(@"should not have been called");
+    }];
+
+    RLMRealm *realm = [RLMRealm defaultRealm];
+    [realm beginWriteTransaction];
+    XCTAssertThrows([realm commitWriteTransactionWithoutNotifying:@[token] error:nil]);
+    [realm cancelWriteTransaction];
```
jpsim: `[token stop]` to prevent the deallocation warning from being logged?

Realm/Tests/NotificationTests.m
```diff
@@ -132,6 +132,66 @@ - (void)testNonMatchingObjectMovedToIndexOfMatchingRowAndMadeMatching {
...
+    RLMRealm *otherRealm = [self realmWithTestPath];
+    [otherRealm beginWriteTransaction];
+    XCTAssertThrows([otherRealm commitWriteTransactionWithoutNotifying:@[_token] error:nil]);
+
+    // and in a write transaction
+    RLMRealm *realm = [RLMRealm defaultRealm];
+    [realm beginWriteTransaction];
+    XCTAssertThrows([otherRealm commitWriteTransactionWithoutNotifying:@[_token] error:nil]);
+    [realm cancelWriteTransaction];
+    [otherRealm cancelWriteTransaction];
```
jpsim: `[token stop]` to prevent the deallocation warning from being logged?

RealmSwift/Realm.swift
```diff
@@ -719,21 +753,25 @@ public final class Realm {
...
 
-     If the block throws an error, the transaction will be canceled, reverting any writes made in the block before
-     the error was thrown.
+     If the block throws an error, the transaction will be canceled and any
+     changes made before the error will be rolled back.
 
-     Write transactions cannot be nested, and trying to execute a write transaction on a Realm which is already
-     participating in a write transaction will throw an error. Calls to `write` from `Realm` instances in other threads
-     will block until the current write transaction completes.
+     Only one write transaction can be open at a time for each Realm file. Write
```
jpsim: "for each Realm."

tgoyne: Just plain "for each Realm" is misleading, as to me it sounds like it's referring to `Realm` instances.

jpsim: yeah, ok. Hard to come up with phrasing that is generalized, accurate and not misleading. Fine to keep as-is.

chrisamanse:
Is there a way to use closures for this? i.e.:

```swift
try realm.write(withoutNotifying: tokens) {
  store.accounts.move(from: source, to: destination)
}
```

Right now, I'm using:

```swift
realm.beginWrite()

store.accounts.move(from: source, to: destination)

try realm.commitWrite(withoutNotifying: tokens)
```

jpsim:
No, there's no closure version of this API yet. We've considered that exact API that you've proposed in your comment, and might end up adding that in the future.

PR #4265, opened 2016-10-26
Update packaging to include Swift 3.0.1

jpsim:
/cc @tgoyne


tgoyne:
Can't we just symlink the examples? There shouldn't be any code changes needed.

jpsim:
> Can't we just symlink the examples? There shouldn't be any code changes needed.

Yes, but we'd need to update `package_examples.rb` to realize those symlinks into concrete copies. I'll work on that, and identifying the `KVOTests.testAllPropertyTypesStandalone()` failures.

jpsim:
3.0.1 examples are now symlinks and I've fixed the failing `KVOTests.testAllPropertyTypesStandalone()` test. Ready for review again @tgoyne

Issue #4232, opened 2016-10-18
ObjC/Swift interoperability


danielgindi:
## Goals

Interoperability with Swift & ObjC libraries.

Specifically, https://github.com/danielgindi/Charts.

We want to be able to provide both `Results` and `RLMResults` constructors.

## How to achieve this?

I can think of two easy ways:

1. Turn the `internal` constructors into `public`, so we can init with an `RLMResults` or an `RLMRealm`
2. Expose the internal properties for those.

Making even one of those subtle and easy changes - will allow having the parallel constructors, but either creating a `Results` from passed `RLMResults` and always working with the `Results` object in our code, or accessing the (currently) internal `RLMResults` of the `Results` object, and always working with the `RLMResults` object in our code.

Making both of those changes - will give great flexibility. Maybe even for other cases that we haven't considered.

## Background info

**Charts** is a crossplatform charting library. We worked with Adam Fish in the past for a Realm layer, and we stumbled upon the ObjC vs Swift situation, which we've put aside for some time. Now there's demand for a solution.

jpsim:
Hi @danielgindi, we're already tracking this as #2920, so I'll be closing this as a duplicate.

Our team would happily accept a PR that implemented this, if you're interesting in contributing this to Realm.

Thanks for the Charts project, looks pretty cool!

danielgindi:
I'll try to find the time and make a PR. This should be easy, a few lines of code...

PR #4293, opened 2016-11-04
Added ObjectivC interoperability feature (Closes #2920, #4232)

### LINKED ISSUE #2920: Provide access to ObjC objects from Swift
Creating a new issue based off the discussion in this PR: https://github.com/realm/realm-cocoa/pull/2906.

For component work, it is helpful to have a single code-base that will work with Realm objects. To facilitate that there needs to be an interface that makes available the underlying ObjC objects within the Swift objects.

It was suggested adding:

``` swift
@objc public class ObjectiveCCompability {
  public func rlmResults(results: Results) -> RLMResults { ... }
  public func rlmArray(list: List) -> RLMArray { ... }
  public func rlmCollection(collection: RealmCollectionType) -> RLMCollection { ... }
  ...
}
```

danielgindi:


danielgindi:
Trying to see the failing details, it gives me "danielgindi is missing the Overall/Read permission"

jpsim:
Thanks for the PR, @danielgindi! Sorry you can't see our Jenkins site, that's private, but also why we mirror a subset of our CI setup on Travis. The only job that failed on Jenkins is the one generating and verifying the documentation. The equivalent Travis job is here: https://travis-ci.org/realm/realm-cocoa/jobs/173215193

The issue is that there are undocumented declarations:

```json
{
  "warnings": [
    {
      "file": "/Users/realm/workspace/objc_pr/configuration/Release/swift_version/3.0.1/target/docs/RealmSwift/ObjectiveCSupport.swift",
      "line": 23,
      "symbol": "ObjectiveCSupport",
      "symbol_kind": "source.lang.swift.decl.class",
      "warning": "undocumented"
    },
    {
      "file": "/Users/realm/workspace/objc_pr/configuration/Release/swift_version/3.0.1/target/docs/RealmSwift/ObjectiveCSupport.swift",
      "line": 27,
      "symbol": "ObjectiveCSupport.rlmResults(results:)",
      "symbol_kind": "source.lang.swift.decl.function.method.instance",
      "warning": "undocumented"
    },
    {
      "file": "/Users/realm/workspace/objc_pr/configuration/Release/swift_version/3.0.1/target/docs/RealmSwift/ObjectiveCSupport.swift",
      "line": 31,
      "symbol": "ObjectiveCSupport.rlmArray(list:)",
      "symbol_kind": "source.lang.swift.decl.function.method.instance",
      "warning": "undocumented"
    },
    {
      "file": "/Users/realm/workspace/objc_pr/configuration/Release/swift_version/3.0.1/target/docs/RealmSwift/ObjectiveCSupport.swift",
      "line": 35,
      "symbol": "ObjectiveCSupport.results(rlmResults:)",
      "symbol_kind": "source.lang.swift.decl.function.method.instance",
      "warning": "undocumented"
    },
    {
      "file": "/Users/realm/workspace/objc_pr/configuration/Release/swift_version/3.0.1/target/docs/RealmSwift/ObjectiveCSupport.swift",
      "line": 39,
      "symbol": "ObjectiveCSupport.list(rlmArray:)",
      "symbol_kind": "source.lang.swift.decl.function.method.instance",
      "warning": "undocumented"
    }
  ],
  "source_directory": "/Users/realm/workspace/objc_pr/configuration/Release/swift_version/3.0.1/target/docs"
}
```

We require 100% documentation coverage for the public API, so these will need docstrings. Do you think you can add these?

You can also run this test locally if you have Xcode 8.1 and [jazzy](https://github.com/realm/jazzy) installed by running `REALM_SWIFT_VERSION=3.0.1 ./build.sh verify-docs`.

I haven't actually reviewed this PR, just letting you know some minor things that will certainly need addressing before we can review and hopefully this!

Thanks again for your contribution!

danielgindi:
Okay thanks for the info! I've rebased with the docs update.

danielgindi:
Still failing, although the docs are 100% , and it passes locally

danielgindi:
Seems likes there's also "Trailing Whitespace Violation: Lines should not have trailing whitespace. (trailing_whitespace)", on empty lines. Which Xcode adds automatically.

jpsim:
This looks pretty good to me, @danielgindi. Could you also please add some minimal unit tests? A changelog entry would also be nice.

danielgindi:
I'm not really sure what that should test. These are simple functions returning the internal properties...

jpsim:
I've made some suggested changes in a [jp-objc-support](https://github.com/realm/realm-cocoa/tree/jp-objc-support) branch to make `ObjectiveCSupport` methods more comprehensive in the types it can convert, more consistent in its naming, add a changelog entry, fix ordering in the Xcode project, and a few other small things.

I'd like you to consider adopting some of those suggested changes.

As for tests, I'd recommend you test round-tripping all types to make sure the conversion methods all work.

danielgindi:
I see that you've added the converters for the other object types as well, I didn't want to touch them because of being private...
Anyway, I pulled all your commits to this PR

jpsim:
> I see that you've added the converters for the other object types as well, I didn't want to touch them because of being private...

These other types actually weren't private, but a small number did have private initializers/converters that I've now made `internal` to make them accessible from `ObjectiveCSupport.swift`, which won't affect the public interface.

This branch now has merge conflicts with `master`, and we'll still need some minimal tests for this functionality before we can merge (see my previous suggestion about having round-trip tests).

danielgindi:
Yeah GitHub is seeing conflicts where local `git` does not.
I merged `master` to the PR branch, and it went flawlessly. I guess they have to update their `git` ;-)

Anyway, we have a problem now due to the generics. `Cannot invoke 'convert' with an argument list of type SwiftStringObject`. The function accepts `Results<Object>`, where it should actually be something like `Results<T>`, but I don't think that Swift can do that.

danielgindi:
Seems like it can. I'll just have to modify the function signatures again - to make them generic. I'll commit that with the tests.

danielgindi:
Okay so about tests:
I would test something like this:
```
        XCTAssertEqual(results,
                       ObjectiveCSupport.convert(
                        object: ObjectiveCSupport.convert(object: results)))
```

But it will fail, as the conversion from ObjC returns a new swift `Results` wrapping the `RLMResults`.

I could also try
```
        XCTAssertEqual(results.rlmResults,
                       ObjectiveCSupport.convert(object: results))
```

But that fails to - as `rlmResults` is internal.

danielgindi:
I just found the magic of `@testable`!

danielgindi:
Ok I've created some tests. Also weirdly Xcode says I don't have permissions to run these test.

jpsim:
@danielgindi please ping me once you've had to resolve all the failing tests and I'll take another look.

danielgindi:
@jpsim Do you have any idea why Xcode doesn't like your project? It does not allow running any tests in your project. Says it's a permissions thing. It obviously isn't.

I had to create the tests blindly. 

I also can't fix the tests based on your CI log, as it's private.

I think that the ball is in your court now.

danielgindi:
@jpsim The tests are passing now

danielgindi:
@jpsim I've had to rebase everything. It got too messy with the merge conflicts.

danielgindi:
@jpsim Any news on this?

RealmSwift/Tests/ObjectiveCSupportTests.swift
```diff
@@ -0,0 +1,243 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+////////////////////////////////////////////////////////////////////////////
+
+import XCTest
+import Realm
+@testable import RealmSwift
```
jpsim: The `RealmSwift` module isn't compiled with testability enabled. You'll either need to update the `xcconfig` files to compile with `ENABLE_TESTABILITY` enabled, or to refactor these tests to avoid needing to access `internal` members.

We've avoided enabling testability so far to prevent accidentally shipping compiled frameworks with extra debug information, or lower optimizations, so those concerns would need to be addressed if you end up wanting to go that route.

jpsim:
> Do you have any idea why Xcode doesn't like your project? It does not allow running any tests in your project. Says it's a permissions thing. It obviously isn't.

What makes you think that Xcode doesn't like our project? We perform most of our development in Xcode using this project, and use all sorts of features without issue: unit testing, `xcodebuild`, Profiling in Instruments, using lldb, etc.

Unless you're referring to why you weren't able to run the tests that you've added, which require that the `RealmSwift` module be built with testability enabled, as I've described in my previous comment.

---

Thanks for your continued work on this!

RealmSwift/Tests/ObjectiveCSupportTests.swift
danielgindi: Well as Apple configures `ENABLE_TESTABILITY` by default only for Debug targets, I don't see a problem.

So my understanding is that you are running tests in Release mode?
We can't really perform round-trip tests without `@testable`...

danielgindi:
I'm saying it "doesn't like" it, because it won't run tests. I can only run the tests from the command line.
I've seen projects that give that "permissions" error when the project or target configuration is a little too complex or too weird for Xcode.

It's a dumb bug of Xcode, but I've tried on more than one macOS instances with latest Xcode, and it just keeps bugging me with the permissions error.

Anyway I just run the tests from the CLI. They **do** pass locally, but it seems like Travis is running tests in Release mode?

jpsim:
We run tests in debug and release mode (and a total of about 40 different configurations).

I use ⌘+U to run the tests all day every day, so I'm not sure why you're getting a "permission error". Perhaps you could open a new issue detailing how you're getting that?

I'll talk to the team about our concerns with `ENABLE_TESTABILITY`. For the time being, you can add that to [`RealmSwift.xcconfig`](https://github.com/realm/realm-cocoa/blob/master/Configuration/RealmSwift/RealmSwift.xcconfig) in your PR, and hopefully we can sort out whether or not that's something we're comfortable with on our end.

jpsim:
After bringing this up with a few members of our team, we're not comfortable enabling testability.

Mostly because we're not familiar enough with its implementation to know if there's any risk of modifying the framework binary in any way: exposing non-public members, growing the binary or bitcode size, preventing certain optimizations from being applied, etc.

This is compounded by the fact that we ship the binaries that we build for testing. We do this so that we're 100% sure that the binaries we're shipping have passed all our tests, which wouldn't be as certain if we built it again for distribution.

So I'd recommend that you refactor the tests to avoid using those internal members. These tests will mostly act as a sanity check to make sure the methods aren't fundamentally broken. Here's an example of one of the tests that could be refactored:

```diff
-let results = realm.objects(SwiftObject.self)
-XCTAssertEqual(results.rlmResults,
-               ObjectiveCSupport.convert(object: results),
-               "RLMResults must stay the same")
+let results = realm.objects(SwiftObject.self)
+let rlmResults = ObjectiveCSupport.convert(object: results)
+XCTAssert(rlmResults.isKindOfClass(RLMResults.self))
+XCTAssertEqual(rlmResults.count, 1)
+XCTAssertEqual(rlmResults.firstObject!.intCol, 123)
```

danielgindi:
I'll change the tests.

About ENABLE_TESTABILITY, Apple says that it's meant for exposing the internal stuff so the tests can access them. I'm guessing that when they are not exposed, they get extra optimizations.
Optimizations did break compilation a few times in older Xcode versions, as we've experienced with Charts a few times. Where we had to just separate a line of code to two to make the library not crash.

So I get what you're saying. We can't trust the optimizer.

But this means that we can't run deeper tests with `@testable`, which is a very neat feature of Xcode.
Maybe in the future you can setup the configuration to run certain tests only with ENABLE_TESTABILITY enabled for Debug & Release, while running the rest of the tests with real Release binaries.

jpsim:
Great work @danielgindi! 👏 

I really appreciate your patience as you worked with us to land this. I hope you'll contribute again in the future 😊 .

I have a few minor edits that I'll do in a follow-up PR.

jpsim:
Follow-up edits: #4361.

jpsim:
> Maybe in the future you can setup the configuration to run certain tests only with ENABLE_TESTABILITY enabled for Debug & Release, while running the rest of the tests with real Release binaries.

This is a great idea, by the way, but as you can tell, not exactly a high priority.

danielgindi:
Yeah. I guess when you'll have a really-important-test-that-requires-testability, you'll create the appropriate solution...

PR #4304, opened 2016-11-07
replace AuthenticationActions with `register` boolean

### LINKED ISSUE #4303: Remove AuthenticationActions and replace with `register` boolean on username/password credential constructor
See realm/realm-mobile-platform#36.

jpsim:
Fixes #4303. /cc @austinzheng


jpsim:
Merged in #4311.

PR #4312, opened 2016-11-08
Various ObjC/Swift language convention API tweaks

jpsim:
As tracked in realm/realm-cocoa-private#353. Depends on #4304 and #4311. /cc @austinzheng


jpsim:
> Just out of curiosity, do we want to continue to provide Swift 2.3 users with Swift 2 naming conventions? It looks like Xcode 8.2 will be the last version to support Swift 2.3, although I'm sure some users will remain on older Xcodes to avoid having to migrate.

We still have ~33% of our active users from the last two weeks on Swift 2.x, so yeah, we should continue making a best effort to support them.

PR #4313, opened 2016-11-09
Add missing Error cases into switch case

RafaelPlantard:
To avoid Xcode warnings around it.


jpsim:
@RafaelPlantard thanks for the effort, but this has already been addressed in `realm/realm-object-store@master`, so the right way to address this would be to update the submodule to point to the head of that branch.

RafaelPlantard:
Thanks for the tip @jpsim, but how can I update the Realm submodule when using CocoaPods?

jpsim:
You could point CocoaPods to a branch of this repo that includes those updates to the object store submodule, such as #4219:

```ruby
pod 'Realm', git: 'https://github.com/realm/realm-cocoa.git', branch: 'az/refactor-sync-os', submodules: true
pod 'RealmSwift', git: 'https://github.com/realm/realm-cocoa.git', branch: 'az/refactor-sync-os', submodules: true
```

RafaelPlantard:
Thanks @jpsim...

PR #4320, opened 2016-11-10
change +[RLMSyncUser allUsers] to return a dictionary

jpsim:
As described in realm/realm-cocoa-private#376. /cc @austinzheng


jpsim:
Now rebased on top of #4322 for simplicity, as I intend to merge that once it passes CI, then this.

PR #4275, opened 2016-10-28
Add support for the client permission change

kishikawakatsumi:
Related: https://github.com/realm/realm-sync-services/pull/236

Add support for the client permission change.
This PR add:
- Data schema for the client permission (`PermissionChange` and a base class)
- Add convenience method for retrieving `__management.realm`
- Unit tests

The permission management requires the same table in both Realm and RealmSwift. I created `PermissionChange` class in both Realm and RealmSwift. (Then add `@compatibility_alias RLMSyncPermissionChange` for the Cocoa naming conventions. But maybe it is unnecessary.)

Unit tests will fail due to the same class exists in Realm and RealmSwift.

```
realm-cocoa/Realm/ObjectServerTests/SwiftSyncTestCase.swift:55: error: -[ObjectServerTests.SwiftObjectServerTests testBasicSwiftSync] : failed: caught "RLMException", "RLMObject subclasses with the same name cannot be included twice in the same target. Please make sure 'PermissionOffer' is only linked once to your current target."
```

Do you have any ideas avoiding it? @mrackwitz @austinzheng @jpsim 

For unit tests, it needs to add the following configurations to `configuration.yml` before tests.

```
auth:
  sync_hosts:
    - 'localhost:9080'
```

How should I achieve it? @austinzheng 

CC @mrackwitz @cmelchior @bigfish24

PR #4324, opened 2016-11-11
Implement PermissionChange

jpsim:
Continued from #4275. Creating a new PR to not lose the work that @kishikawakatsumi did in case we want to revert to that approach.


jpsim:
This is now ready for review /cc @kishikawakatsumi @mrackwitz @bigfish24.

Realm/RLMSyncUtil.mm
```diff
@@ -19,6 +19,25 @@
 #import <Foundation/Foundation.h>
 
 #import "RLMSyncUtil_Private.hpp"
+#import "RLMSyncUser_Private.hpp"
+#import "RLMRealmConfiguration+Sync.h"
+#import "RLMRealmConfiguration_Private.hpp"
+#import "RLMSyncPermissionChange.h"
+
+@implementation RLMRealmConfiguration (RealmSync)
++ (instancetype)configurationForUser:(RLMSyncUser *)user {
```
mrackwitz: The name of this method should indicate that this will return a configuration for a management-Realm, given that there will be other special purpose user-specific Realms.

jpsim:
> The original intention behind the `RLMSyncPermissionBaseObject` was to have a common ancestor for the different classes in the management-Realm and share properties between those. This would be good to re-add once we want to implement them in the Cocoa SDK. But from your commits, it seems as that's only a temporary workaround.

I didn't feel the complexity that having this base class was justified by the advantages of sharing a few properties between them.

Given that jazzy doesn't handle inherited docs very well, these properties would need to be overridden and redeclared in their subclasses anyway, not actually saving any code duplication, but rather causing more of it.

In my commit messages, the "temporary hack" I was referring to was the hardcoded `RLMSyncPermissionChange`/`SyncPermissionChange` class to `PermissionChange` table mapping that was hardcoded in the object store, which I've since generalized in the ability to map model classes to custom table names.

mrackwitz:
>I didn't feel the complexity that having this base class was justified by the advantages of sharing a few properties between them.
>
>Given that jazzy doesn't handle inherited docs very well, these properties would need to be overridden and redeclared in their subclasses anyway, not actually saving any code duplication, but rather causing more of it.

Given these complexities, I didn't had in mind anymore, I'd agree with this assessment. The only minor drawback now is that there is no common base type anymore for objects which could be potentially handled in the same way by user-side logic. This could be mitigated though by implementing a protocol defining the common properties, if this is any better than a parent class.

jpsim:
> The only minor drawback now is that there is no common base type anymore for objects which could be potentially handled in the same way by user-side logic.

What do you mean by this? The APIs from Realm Objective-C and Realm Swift aren't design to interoperate in any way, and doing so is unsupported. We state this explicitly in our docs:

> If you’re looking to use Realm from Objective‑C, or from mixed Objective‑C & Swift apps please see Realm Objective‑C instead. The Realm Objective‑C and Realm Swift APIs are not interoperable and using them together is not supported.

mrackwitz:
I meant once there is not only `PermissionChange`, but also `PermissionOffer` and `PermissionRequest`.

jpsim:
> I meant once there is not only `PermissionChange`, but also `PermissionOffer` and `PermissionRequest`.

I see your point, I'm not sure I agree at the moment, but I'm ok with reconsidering this once we have these additional models in the future.

jpsim:
I've reworked the implementation again to not require any changes to the object store, reporting a custom object name to the object store rather than storing a separate "table name" member in `realm::ObjectSchema`.

@tgoyne could you please take a look?

Realm/RLMSyncPermissionChange.h
```diff
@@ -0,0 +1,107 @@
...
+ @param userID    The user or users who should be granted these permission changes.
+                  Use `*` to change the permissions for all users.
+ @param mayRead   Define read access. Set to `YES` or `NO` to update this value.
+                  Leave unset to preserve the existing setting.
+ @param mayWrite  Define write access. Set to `YES` or `NO` to update this value.
+                  Leave unset to preserve the existing setting.
+ @param mayManage Define management access. Set to `YES` or `NO` to update this value.
+                  Leave unset to preserve the existing setting.
+ */
++ (instancetype)permissionChangeForRealmURL:(NSString *)realmURL
```
tgoyne: Initializers and factory methods normally use `With` here rather than `For`.

Realm/RLMSyncUtil.mm
```diff
@@ -19,6 +19,29 @@
...
+    } else {
+        components.scheme = @"realm";
+    }
+    components.path = @"/~/__management";
+    NSURL *managementRealmURL = components.URL;
+    RLMSyncConfiguration *syncConfig = [[RLMSyncConfiguration alloc] initWithUser:user realmURL:managementRealmURL];
+    RLMRealmConfiguration *config = [RLMRealmConfiguration new];
+    config.syncConfiguration = syncConfig;
+    config.objectClasses = @[RLMSyncPermissionChange.class];
+    config.schemaMode = realm::SchemaMode::Additive;
```
tgoyne: This is set automatically when a sync config is set (the private setter is just there for testing it without using sync).

kishikawakatsumi:
Looks good to me! Thanks.

PR #4341, opened 2016-11-17
Speed up the object server tests

tgoyne:
Reduce the number of iterations used to hash the password to 1, as that's where most of the run time of the tests was spent. Cuts the total run time from 1 minute to 25 seconds on my machine.

This exposes some race conditions in the tests which are currently just worked around with some `sleep()` calls until the object store bug is fixed.


tgoyne:
Speedup on the CI machine was similar (112 seconds -> 35 seconds).

PR #4362, opened 2016-11-24
realm iClould identity provider fix

vinnyt:
this wasn't working when it was iCloud.


jpsim:
@mrackwitz @TimOliver since the two of you built the iCloud/CloudKit integration, which term is more appropriate for this auth provider? If it's CloudKit, then we should probably merge this and later also rename the constructors, constants, documentation, etc to reflect that.

If iCloud is more appropriate, then we should take the opportunity of RMP being in beta to change the identifier before it's too late to make breaking changes like these.

vinnyt:
thanks for looking into this one really appreciate it!

jpsim:
Speaking with @mrackwitz "offline", his thoughts on this are that there's no such thing as "iCloud authentication", that the APIs we use to implement this are all in CloudKit, and so we should consistently use that naming.

So we'll merge this PR and file a new issue to update our API and documentation nomenclature to replace uses of "iCloud" with "CloudKit".

jpsim:
Thanks for the PR!

mrackwitz:
Thanks for bringing this up! :+1:

Issue #4338, opened 2016-11-17
Update SyncUser.logIn to have a callbackQueue optional defaulted parameter


austinzheng:
The sync login method callback runs on a background queue, which might be surprising behavior to those who expect it to run on the main queue. It might be a good idea to document this and/or allow the queue to be optionally specified.

stel:
I like the idea of specifying the the queue explicitly (with default value).

Reposting @jpsim 's suggestion from #4364 here:
> What could be done, however, is providing an optional default argument of callbackQueue: DispatchQueue? = .main, since the vast majority of users will want to handle the callback on the main queue, but this still leaves the flexibility for users wanting to avoid that overhead to pass in callbackQueue: nil.

austinzheng:
I think for our next breaking release, we should change the callback to just run on the main queue. Most of the time, the developer simply wants to immediately use the logged-in Realm user passed into the callback to configure their synced Realms. There are a number of features that require a run loop to function: auto-updating Realms, notification blocks, permission results, and we should make those use cases as foolproof as possible. In addition, this allows users to update their UI directly from the callback.

Given how infrequent a user log in should be, I don't think there is any real reason to try to optimize performance by allowing a queue to be specified.

austinzheng:
In `master-3.0`.

PR #4364, opened 2016-11-24
Fix dispatch SyncUser.logIn completion callback to the main thread

stel:
Initially this issue was addressed in https://github.com/realm/realm-cocoa-private/issues/186, so I've added a test to handle this in the future.

/cc @austinzheng, @jpsim

### DIFF
--- a/Realm/ObjectServerTests/RLMObjectServerTests.m
+++ b/Realm/ObjectServerTests/RLMObjectServerTests.m
@@ -32,6 +32,35 @@
 
 #pragma mark - Authentication
 
+/// Log in callback should be dispatched to the main thread
+- (void)testAuthenticationCallbackThread {
+    RLMSyncCredentials *credentials = [RLMObjectServerTests basicCredentialsWithName:ACCOUNT_NAME() register:YES];
+
+    // Successful callback should be dispatched to the main thread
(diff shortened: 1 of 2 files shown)


jpsim:
We discussed this IRL, but will share my thinking here again.

There's overhead to dispatching to a different queue. With this change, that overhead is applied to all our users, regardless of whether or not they need to process the results of the callback on the main queue or a background queue.

By not making any assumptions regarding what queue our users want to handle the results, we can avoid that overhead, and finally, there's very little boilerplate to dispatching to the main queue with GCD: `DispatchQueue.main.async { ... }`.

---

What _could_ be done, however, is providing an optional default argument of `callbackQueue: DispatchQueue? = .main`, since the vast majority of users will want to handle the callback on the main queue, but this still leaves the flexibility for users wanting to avoid that overhead to pass in `callbackQueue: nil`.

austinzheng:
I'm in favor of the optional argument with default of `.main` approach. Would solve the most common case transparently, and allow flexibility for testing (where we can't dispatch to the main queue with the way our Cocoa ROS tests are currently structured) or custom behavior.

stel:
Didn't notice https://github.com/realm/realm-cocoa/issues/4338, looks like the best place to track this.

jpsim:
@stel please delete this branch if you're done with it.

PR #4374, opened 2016-11-26
Do not dispatch logInWithCredentials callback to the main thread in case of error or access token credentials

stel:
This PR is mutually exclusive to https://github.com/realm/realm-cocoa/pull/4364. It fixes dispatching of the `+[RLMSyncUser logInWithCredentials:]` callback to the background thread in case of error or when using accessToken for authentication.

/cc @austinzheng, @jpsim

### DIFF
--- a/Realm/RLMSyncUser.mm
+++ b/Realm/RLMSyncUser.mm
@@ -261,17 +261,9 @@ using namespace realm;
                authServerURL:(NSURL *)authServerURL
                      timeout:(NSTimeInterval)timeout
              completionBlock:(RLMUserCompletionBlock)completion {
-    // Wrap the completion block.
-    RLMUserCompletionBlock theBlock = ^(RLMSyncUser *user, NSError *error){
-        if (!completion) { return; }
-        dispatch_async(dispatch_get_main_queue(), ^{
-            completion(user, error);
-        });
-    };
-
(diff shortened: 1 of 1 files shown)


austinzheng:
I've marked this as "On Hold" since we're currently examining the other approach. If we abandon that approach I'll update the labels.

jpsim:
Instead of this _or_ #4364, can we go with door number 3 and implement the API proposed in this comment? https://github.com/realm/realm-cocoa/pull/4364#issuecomment-263658980

stel:
I like the "door number 3"! I think we it should be tracked in https://github.com/realm/realm-cocoa/issues/4338

jpsim:
@stel please delete this branch if you're done with it.

stel:
@jpsim I actually think that it could be merged until we come up with the another approach. This PR aligns the behavior for all cases (error and access token auth) and dispatches the callback to the background thread as stated in the changelog.

jpsim:
You're right. On second thought, let's merge this.

jpsim:
There's no need for a changelog entry as this is really just fixing the implementation to match this previous changelog entry: https://github.com/realm/realm-cocoa/blame/v2.1.1/CHANGELOG.md#L82-L83

PR #4381, opened 2016-11-28
fix issue preventing using synchronized Realms in Xcode Playgrounds

jpsim:
fixes realm/realm-mobile-platform#57 /cc @austinzheng


DanielAsher:
Thanks so much!

PR #4391, opened 2016-11-30
Update binding to use latest object store

austinzheng:
@jpsim @TimOliver 

Changes:
- Session error callback no longer looks up its session, it gets the session directly (see https://github.com/realm/realm-object-store/pull/265)


TimOliver:
Looks good to me, but I'll wait and see what @jpsim says! :)

PR #4392, opened 2016-12-01
Revert "Make the 'Object Server Tests' scheme buildable directly"

### LINKED ISSUE #4343: Carthage builds 'Object Server Tests'
I have no clue why it's building a unit test bundle, but there it is:

```
*** Building scheme "Object Server Tests" in Realm.xcworkspace
```

This is with Carthage 0.18.1 and Realm 2.1.0.

See https://ci.realm.io/job/create_github_release_cocoa/60/console

jpsim:
This reverts commit 0b7efa72137cb761312b444cb188e352edc5cdbc.

By reverting this commit, Xcode will no longer guess that the "Object Server Tests" scheme is a framework, and therefore Carthage will no longer attempt to build it as if it was a framework.

Fixes #4343.

This has the unfortunate effect of 1) not doing something sensible when hitting ⌘+B in Xcode and 2) not being able to run the unit tests with Instruments.app.

See [rdar://29407087](http://www.openradar.me/29407087) for details.

/cc @tgoyne

### DIFF
--- a/Realm.xcodeproj/xcshareddata/xcschemes/Object Server Tests.xcscheme
+++ b/Realm.xcodeproj/xcshareddata/xcschemes/Object Server Tests.xcscheme
@@ -1,40 +1,10 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <Scheme
    LastUpgradeVersion = "0800"
-   version = "1.8">
+   version = "1.3">
    <BuildAction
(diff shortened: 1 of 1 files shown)


jpsim:
@tgoyne would be nice to get this in before release 2.1.1 so that Carthage users get this fixed.

PR #4398, opened 2016-12-02
Update code based on recent object store changes

austinzheng:
@jpsim 

Changes for compatibility with https://github.com/realm/realm-object-store/pull/270, and some other recent object store changes. All sync tests pass.


jpsim:
Some of this has now landed from #4404. This needs to be rebased.

austinzheng:
The underlying PRs this PR refers to are no longer valid.

PR #4410, opened 2016-12-05
Added support for 'LIKE' queries

astigsen:
Allow you to do NSPredicate wildcard queries like `"value LIKE '?bc*'"` (matches strings like "abcde" and "cbc").

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,7 +12,7 @@ x.x.x Release notes (yyyy-MM-dd)
 
 ### Enhancements
 
-* None.
+* Add support for 'LIKE' queries (wildcard matching).
 
 ### Bugfixes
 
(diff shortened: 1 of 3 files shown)


tgoyne:
Everything is `QueryTests` that's checked with `RLMAssertCount()` is also tested on a handed-over async query, so that's covered. The only thing I can see that's untested is `nil`; `testQueryOnNullableStringColumn` and `testQueryingOnLinkToNullableStringColumn` should add a `LIKE` test, and something should verify that `("col LIKE %@", nil)` fails sensibly.

jpsim:
There'd also be the option to test more edge cases, like multi-byte characters and non-ASCII unicode, but core's tests seem to cover that partly, and if we were to add more they should probably go there first.

bdash:
It'd also be worth having test coverage for `column LIKE other_colum` and `"literal" LIKE column`, even though they're not particularly likely to be used.

jpsim:
@astigsen let us know if you'd like us to move this forward and address feedback in your place, considering that you have more than just this PR on your plate 😉

astigsen:
I will be fine with you just moving ahead and expanding the tests appropriately.

PR #4246, opened 2016-10-21
Handle authentication errors in network client

stel:
Currently only errors that could be returned from `SyncUser.authenticate...` are handled:
- `RLMSyncAuthErrorInvalidCredential`
- `RLMSyncAuthErrorUserDoesNotExist`
- `RLMSyncAuthErrorUserAlreadyExists`

Any other non-successful HTTP codes are still treated as `RLMSyncErrorHTTPStatusCodeError`, but may contain specific error messages from server. This restriction could be removed by defining additional error codes.

I think error reporting needs to be discussed, my idea is to wrap all the network, parsing, etc errors into sync-specific ones, it doesn't make much sense for users to get `NSCocoaErrorDomain` errors from sync api calls in most cases, but this errors still could be included by `NSUnderlyingErrorKey` into `RLMSyncErrorDomain` errors. Currently it's not a part of this PR.

/cc @austinzheng, @jpsim

PR #4418, opened 2016-12-06
Fix a few warnings when building Realm with -Wall or -Wundef

### LINKED ISSUE #4409: Compile error after upgrading Realm to 2.1.1 with Carthage
## Actual Results

XCode report compile error on this line:

`#if __has_attribute(ns_error_domain) && (!__cplusplus || __cplusplus >= 201103L)`

in `RLMConstant.h` with error message:

`__cplusplus is not defined, evaluates to 0`

Realm version: 2.1.1 (Realm Objective-C)
Xcode 8.0: Build version 8A218a
Swift Version: 2.3
Carthage version: 0.15.2

jpsim:
Fixes #4409. /cc @bdash @tgoyne

### DIFF
--- a/Realm/ObjectServerTests/RLMObjectServerTests.m
+++ b/Realm/ObjectServerTests/RLMObjectServerTests.m
@@ -78,7 +78,7 @@
         XCTAssertEqual(error.domain, RLMSyncErrorDomain);
         XCTAssertEqual(error.code, RLMSyncAuthErrorInvalidCredential);
         XCTAssertNotNil(error.localizedDescription);
-        
+
         [expectation fulfill];
     }];
     [self waitForExpectationsWithTimeout:2.0 handler:nil];
@@ -96,7 +96,7 @@
         XCTAssertNil(user);
         XCTAssertNotNil(error);
         XCTAssertEqual(error.domain, RLMSyncErrorDomain);
-        XCTAssertEqual(error.code, RLMSyncAuthErrorUserDoesNotExist);
+        XCTAssertEqual(error.code, RLMSyncAuthErrorInvalidCredential);
         XCTAssertNotNil(error.localizedDescription);
 
         [expectation fulfill];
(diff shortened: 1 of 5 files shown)


jpsim:
Marking as in-progress as I haven't incorporated other changes in Object Store `master` yet.

xiaodao:
I would like to know when will this change be released.

Thanks

jpsim:
@xiaodao you'll be notified of any activity in this thread via GitHub notifications. FYI in the future, it's not necessary to comment on a thread to be notified of updates, you can simply click the "Subscribe" button on the right.

image

xiaodao:
@jpsim once this pull request gets merged, can I run `carthage update` to get the latest build?

jpsim:
I've decided to revert the object store changes in this PR to isolate the work that needs to happen to update the object store from the simple warning fixes in cocoa.

The object store changes either mean that we'll need to change the `sessionForURL:` implementation or change the public API, perhaps to a non-thread-confined `Realm.session` property. I think both cases are easy to implement, but need to be discussed and aligned across SDKs. In any case, I'm specifically reverting the object store update in this PR to avoid needing to decide on that here.

Can you review again @tgoyne or @bdash?

bdash:
Isn't keeping `-sessionForURL:` working as-is just a matter of:
```diff
-    return [[RLMSyncSession alloc] initWithSyncSession:_user->session_for_url([[url absoluteString] UTF8String])];
+    auto path = SyncManager::shared().path_for_realm(_user->identity(), [url.absoluteString UTF8String]);
+    return [[RLMSyncSession alloc] initWithSyncSession:_user->session_for_on_disk_path(path)];
```

jpsim:
Yes, but what I was saying in my previous comment was that it may not be the API we want to expose. I suppose we can stick with that for now and revisit later...

jpsim:
Looks like the error codes have changed, which is significant. I'll take a closer look tomorrow.

jpsim:
Looks like #4246 introduced incorrect error codes. That shouldn't have passed review...

I've fixed the expected error codes in the tests and filed realm/realm-sync-services#638 to report.

@bdash @tgoyne this is ready for review once more.

PR #4446, opened 2016-12-14
update ObjC model template comment to use ObjC generics

jpsim:
Since we no longer support Xcode 6, might as well encourage the use of ObjC generics.


colasbd:
:+1:

PR #4453, opened 2016-12-15
[Draw, iOS] Specify custom credentials if logging in failed

stel:
No need to recompile the app anymore. The alert is show only if the logging in with the default credentials fails.

screen shot 2016-12-15 at 2 37 10 pm

@TimOliver, @bigfish24


jpsim:
Going to let @TimOliver review this one.

stel:
Unfortunately until we have log out and client reset we still need to reset the local data, so this functionality if only helpful until the first successful login.

jpsim:
@TimOliver @stel @dhmspector  do we still need this given LoginKit and the current state of demos?

TimOliver:
No, I don't think we need this anymore.

In fact, we have a repo for Draw in the Realm Demos repo now. We can probably remove this copy from Realm Cocoa now.

PR #4457, opened 2016-12-16
Support Swift files with overridden @objc names

bnickel:
Previously, Realm would assert when creating a schema for this object
because it only considered files with modular class names to be Swift
files.

```
@objc(Foo) class Foo: Object {}
```

This is problematic for projects that do all manipulation of an
instance in Swift but still need to pass it between Objective-C classes.

Now we check for a dot in the name or inheriting from RealmSwiftObject.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,8 @@ x.x.x Release notes (yyyy-MM-dd)
 
 ### Bugfixes
 
-* None.
+* Support Realm model classes defined in Swift with overridden Objective-C
+  names (e.g. `@objc(Foo) class SwiftFoo: Object {}`).
 
 2.6.2 Release notes (2017-04-21)
 =============================================================
--- a/Realm/RLMObjectSchema.mm
+++ b/Realm/RLMObjectSchema.mm
@@ -96,10 +96,14 @@ using namespace realm;
 
     // determine classname from objectclass as className method has not yet been updated
     NSString *className = NSStringFromClass(objectClass);
-    bool isSwift = [RLMSwiftSupport isSwiftClassName:className];
-    if (isSwift) {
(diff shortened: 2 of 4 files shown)


jpsim:
Thanks for the PR, @bnickel! Always lovely to see contributions from the community.

Could you please add some Swift 2 and Swift 3 tests, as well as a changelog entry?

bnickel:
@jpsim Sure, I'll probably get that in on Monday.

As far as a test goes, we're moving from a state where schema generation went from asserting in `doesNotRecognizeSelector:` to running.  Is it sufficient to provide a test that would have crashed but now doesn't?

jpsim:
> Is it sufficient to provide a test that would have crashed but now doesn't?

That's not only sufficient, but ideal!

bnickel:
I've added tests and signed the CLA.

jpsim:
> That's not only sufficient, but ideal!

Sorry, thinking about this more, it'd be nice to also have tests confirming that an object of this type can be created and added to a Realm.

jpsim:
@bnickel do you still intend to continue this?

bnickel:
Yeah, I totally forgot about this.  I'll add the tests on Monday.

bnickel:
Rebased and tests added.

jpsim:
@bnickel thanks for adding the test. Does this also work if the ObjC name doesn't match the Swift one? If so, could you please add a test for that?

This will need an entry in the "Bugfixes" section of the changelog too. Something along the lines of:

```md
* Support Realm model classes defined in Swift with overridden Objective-C
  names (e.g. `@objc(Foo) class SwiftFoo: Object {}`).
```

If ObjC names that differ from Swift ones don't work, that should be explicitly noted in the changelog entry too.

bnickel:
I've added the tests.  Arbitrary names work and the changelog entry above was perfect so I just used it.

PR #4467, opened 2016-12-19
return class name from `Object.className` instance method/var

### LINKED ISSUE #4456: Object's className var differs from Object's className() static func
## Goals

I have a simple Object subclass named `Model`, and I want to use the `className` property on `Model`'s instances.

## Expected Results

Expected to get the same result as calling `Model.className()`

## Actual Results

Getting an empty string `""`

## Steps to Reproduce

1. Create an `Object` subclass
2. Create an instance of that subclass
3. Try using `instance.className` and `Class.className()` and notice that they differ

## Code Sample

```swift
// Model.swift

import Realm
import RealmSwift

final class Model: Object {
    dynamic var uuid: String = ""

        return #keyPath(Model.uuid)
    }
}
```

```swift
// FailingSpec.swift

class RealmClassNameTests: XCTestCase {
    func testClassName() {
        let uuid = UUID().uuidString
        let model = Model(value: [#keyPath(Model.uuid): uuid])

        XCTAssertNotNil(model) // OK
        XCTAssertEqual(uuid, model.uuid) // OK
        XCTAssertEqual(Model.className(), "Model") // OK
        XCTAssertEqual(model.className, "Model") // 😿
    }
}
```

```
ProductName:  Mac OS X
ProductVersion: 10.12.2
BuildVersion: 16C67

Xcode 8.2
Build version 8C38

/Users/nemesis/.rvm/gems/ruby-2.2.2/bin/pod
CocoaPods 1.1.1
Realm (2.1.1)
RealmSwift (2.1.1)

/bin/bash

/usr/local/bin/git
git version 2.10.2
```

jpsim:
The intention when these were added was to prevent them from being overridden in user models by marking these overrides as `final`, since users were trying to change the name of the table in the Realm by returning a different value than the default.

However, the `static` keyword was omitted by accident, causing this to not work as intended at all.

So now that we need to keep this around to avoid a breaking change to conform to SemVer, might as well have it return the correct values.

Fixes #4456. /cc @tgoyne @bdash

### DIFF
--- a/RealmSwift/Object.swift
+++ b/RealmSwift/Object.swift
@@ -131,10 +131,10 @@ open class Object: RLMObjectBase {
 
     #if os(OSX)
     /// Helper to return the class name for an Object subclass.
-    public final override var className: String { return "" }
+    public final override var className: String { return objectSchema.className }
     #else
     /// Helper to return the class name for an Object subclass.
-    public final var className: String { return "" }
+    public final var className: String { return objectSchema.className }
     #endif
 
     /**
@@ -487,10 +487,10 @@ public class Object: RLMObjectBase {
 
     #if os(OSX)
     /// A helper property that returns the class name for an `Object` subclass.
-    public final override var className: String { return "" }
+    public final override var className: String { return objectSchema.className }
     #else
     /// A helper property that returns the class name for an `Object` subclass.
-    public final var className: String { return "" }
+    public final var className: String { return objectSchema.className }
     #endif
 
     /**


jpsim:
Next steps:

1. Override [`+[RLMObjectBase className]`](https://github.com/realm/realm-cocoa/blob/v2.8.1/Realm/RLMObjectBase.h#L34) to make it `final` because some users think overriding it does something useful, which it does not since we replace it at runtime.
2. Remove this instance property for cocoa 3.0.

PR #4365, opened 2016-11-24
Add PermissionOffer and PermissionOfferResponse

kishikawakatsumi:
Add RLMSyncPermissionOffer and RLMSyncPermissionOfferResponse.
Move `RLMSyncManagementObjectStatus` and function to `RLMSyncUtil`.

Unit tests are split another PR because tests don't work until new ROS will be released.

CC @mrackwitz @jpsim

PR #4482, opened 2016-12-23
Make PermissionChange's `id` property declared as primary key

kishikawakatsumi:
To align Java APIs. This also ensures that `id` is a unique value. 

CC @mrackwitz @jpsim @bigfish24 @cmelchior


kishikawakatsumi:
https://github.com/realm/realm-cocoa/pull/4365 needs This change. I cherry-picked this to https://github.com/realm/realm-cocoa/pull/4365

PR #4496, opened 2016-12-28
Removed Optional for 'realm' Property on Swift 'Results'

AnthonyMDev:
There was no reason to be casting the `realm` property as an optional on `Results`. 

The `rlmResults.realm` is strongly retained and the internal initializer is neither throwing or failable, so there is no way for this to be `nil`.

The documentation even said: `Note that this property will never return nil`.

### DIFF
--- a/RealmSwift/Results.swift
+++ b/RealmSwift/Results.swift
@@ -100,8 +100,8 @@ public final class Results<T: Object>: NSObject, NSFastEnumeration {
 
     // MARK: Properties
 
-    /// The Realm which manages this results. Note that this property will never return `nil`.
-    public var realm: Realm? { return Realm(rlmResults.realm) }
+    /// The Realm which manages this results.
+    public var realm: Realm { return Realm(rlmResults.realm) }
 
     /**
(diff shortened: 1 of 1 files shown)


bdash:
Thanks for the pull request, @AnthonyMDev. Sadly it fails to compile:
```
RealmSwift/Results.swift:365:1: error: type 'Results<T>' does not conform to protocol 'RealmCollection'
extension Results: RealmCollection {
^
RealmSwift/RealmCollection.swift:136:9: note: protocol requires property 'realm' with type 'Realm?'; do you want to add a stub?
    var realm: Realm? { get }
        ^
RealmSwift/Results.swift:104:16: note: candidate has non-matching type 'Realm'
    public var realm: Realm { return Realm(rlmResults.realm) }
```

As you can see the `realm` property on `Results` is optional for sake of conformance with `RealmCollection`.

AnthonyMDev:
I was confused because it was building locally for me. Then I realized I was building the `Realm` target, not `RealmSwift`.

Thanks for the explanation.

PR #4505, opened 2017-01-03
Add support for sorting based on simple key paths

bdash:
This allows for sorting over to-one relationships. Key paths that include to-many relationships, collection operators, or non-persisted properties are not supported.

### DIFF
--- a/Realm/RLMQueryUtil.mm
+++ b/Realm/RLMQueryUtil.mm
@@ -860,7 +860,7 @@ struct KeyPath {
     bool containsToManyRelationship;
 };
 
-KeyPath key_path_from_string(RLMSchema* schema, RLMObjectSchema *objectSchema, NSString *keyPath)
+KeyPath key_path_from_string(RLMSchema *schema, RLMObjectSchema *objectSchema, NSString *keyPath)
 {
     RLMProperty *property;
     std::vector<RLMProperty *> links;
--- a/Realm/RLMQueryUtil.mm
+++ b/Realm/RLMQueryUtil.mm
@@ -854,7 +854,13 @@ void QueryBuilder::add_constraint(RLMPropertyType type, NSPredicateOperatorType 
     }
 }
 
-ColumnReference QueryBuilder::column_reference_from_key_path(RLMObjectSchema *objectSchema, NSString *keyPath, bool isAggregate)
+struct KeyPath {
+    std::vector<RLMProperty *> links;
+    RLMProperty *property;
+    bool containsToManyRelationship;
+};
+
+KeyPath key_path_from_string(RLMSchema* schema, RLMObjectSchema *objectSchema, NSString *keyPath)
 {
     RLMProperty *property;
     std::vector<RLMProperty *> links;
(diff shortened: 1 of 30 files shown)


Realm/Tests/QueryTests.m
```diff
@@ -594,12 +594,12 @@ - (void)testQuerySorting
     [self verifySort:realm column:@"stringCol" ascending:NO expected:@"cc"];
 
     // sort invalid name
-    RLMAssertThrowsWithReasonMatching([[AllTypesObject allObjects] sortedResultsUsingProperty:@"invalidCol" ascending:YES], @"'invalidCol'.* 'AllTypesObject'.* not found");
-    XCTAssertThrows([arrayOfAll.array sortedResultsUsingProperty:@"invalidCol" ascending:NO]);
+    RLMAssertThrowsWithReasonMatching([[AllTypesObject allObjects] sortedResultsUsingKeyPath:@"invalidCol" ascending:YES], @"'invalidCol'.* not found .*'AllTypesObject'");
+    XCTAssertThrows([arrayOfAll.array sortedResultsUsingKeyPath:@"invalidCol" ascending:NO]);
 
     // sort on key path
```
jpsim: Change this comment to "sort on key path with collection operator"? Or remove given the newly added `testSortByUnspportedKeyPath` test.

RealmSwift/LinkingObjects.swift
```diff
@@ -220,6 +220,23 @@ public final class LinkingObjects<T: Object>: LinkingObjectsBase {
     /**
      Returns a `Results` containing all the linking objects, but sorted.
 
+     Objects are sorted based on the values of the given key path. For example, to sort a collection of `Student`s from
+     youngest to oldest based on their `age` property, you might call
+     `students.sorted(byKeyPath: "age", ascending: true)`.
+
+     - warning: Collections may only be sorted by properties of boolean, `Date`, `NSDate`, single and double-precision
```
jpsim: This warning should be updated to reflect that key paths of object properties can also be used to sort.

RealmSwift/LinkingObjects.swift
```diff
@@ -220,6 +220,23 @@ public final class LinkingObjects<T: Object>: LinkingObjectsBase {
     /**
      Returns a `Results` containing all the linking objects, but sorted.
 
+     Objects are sorted based on the values of the given key path. For example, to sort a collection of `Student`s from
+     youngest to oldest based on their `age` property, you might call
+     `students.sorted(byKeyPath: "age", ascending: true)`.
```
jpsim: Perhaps a key path example could be added here to highlight that it's now possible to do?

> Or to sort a collection of `Dog`s alphabetically based on their `owner` property's `name` value, you could call `dogs.sorted(byKeyPath: "owner.name", ascending: true)`.

RealmSwift/SortDescriptor.swift
```diff
@@ -22,43 +22,51 @@ import Realm
...
+
     // MARK: Functions
 
     /// Returns a copy of the sort descriptor with the sort order reversed.
     public func reversed() -> SortDescriptor {
-        return SortDescriptor(property: property, ascending: !ascending)
+        return SortDescriptor(keyPath: keyPath, ascending: !ascending)
     }
+
+    @available(*, unavailable, renamed: "keyPath")
```
jpsim: `deprecated`, not `unavailable`, otherwise it'll break existing code.

RealmSwift/SortDescriptor.swift
```diff
@@ -129,36 +137,44 @@ public struct SortDescriptor {
...
+
     // MARK: Functions
 
     /// Returns a copy of the sort descriptor with the sort order reversed.
     public func reversed() -> SortDescriptor {
-        return SortDescriptor(property: property, ascending: !ascending)
+        return SortDescriptor(keyPath: keyPath, ascending: !ascending)
     }
+
+    @available(*, unavailable, renamed="keyPath")
```
jpsim: ditto

Realm/RLMQueryUtil.mm
```diff
@@ -854,7 +854,13 @@ bool is_nsnull(T) {
...
 }
 
-ColumnReference QueryBuilder::column_reference_from_key_path(RLMObjectSchema *objectSchema, NSString *keyPath, bool isAggregate)
+struct KeyPath {
+    std::vector<RLMProperty *> links;
+    RLMProperty *property;
+    bool containsToManyRelationship;
+};
+
+KeyPath key_path_from_string(RLMSchema* schema, RLMObjectSchema *objectSchema, NSString *keyPath)
```
jpsim: minor style: `RLMSchema *schema`

bdash:
> There are a few things in RLMQueryUtil.mm that might make sense to move to the object store, but that seems unnecessary until support for key path sorting in the parser is added.

We need to find a way to share the query building code between Cocoa and JS. Until then, it's more trouble than it's worth to try and share bits and pieces of it.

RealmSwift/LinkingObjects.swift
bdash: This warning doesn't seem like it needs to be updated to me? The data types that we can sort on hasn't changed.

jpsim: Maybe just changing "properties" to "key paths" should be sufficient then.

bdash: I dunno. The property is what has a type.

jpsim: It just seems like this warning lists the ways you can sort, this PR adds a way, but it's not reflected in the list. Seems like there ought to be a way to express that here.

bdash:
The failing CI jobs appear to be issues with CI.

PR #4507, opened 2017-01-04
Add support for client reset

austinzheng:
Depends upon https://github.com/realm/realm-object-store/pull/259

Changes:
- Added an error key representing client reset errors
- Client reset errors provide the location of the backed-up Realm file, as well as a block that can optionally be called in order to immediately carry out the reset process


CHANGELOG.md
```diff
@@ -3,7 +3,10 @@ x.x.x Release notes (yyyy-MM-dd)
 
 ### Sync Breaking Changes (In Beta)
 
-* None.
+* Underlying sync engine upgraded to version BETA-6.3.
```
austinzheng: Reminder: this needs to be updated to whatever actual version ends up being pulled into this commit.

PR #4478, opened 2016-12-22
Update binding in response to object store updates

austinzheng:
Depends upon: https://github.com/realm/realm-object-store/pull/303

Will unmark as 'in progress' once CI passes.

PR #4509, opened 2017-01-04
Add support for progress notifiers

austinzheng:
Depends upon https://github.com/realm/realm-cocoa/pull/4478, https://github.com/realm/realm-object-store/pull/310


Realm/RLMSyncSession.mm
```diff
@@ -24,10 +24,38 @@
...
+@implementation RLMProgressNotificationToken
+
+- (void)stop {
+    if (auto session = _session.lock()) {
+        session->unregister_progress_notifier(_token);
+    }
+}
+
+- (void)dealloc {
+    [self stop];
```
jpsim: This behavior is different than `RLMNotificationToken`:

```objc
- (void)dealloc {
    if (_realm || _block) {
        NSLog(@"RLMNotificationToken released without unregistering a notification. You must hold "
              @"on to the RLMNotificationToken returned from addNotificationBlock and call "
              @"-[RLMNotificationToken stop] when you no longer wish to receive RLMRealm notifications.");
    }
}
```

Realm/RLMSyncSession.mm
```diff
@@ -24,10 +24,42 @@
...
+
+@implementation RLMProgressNotificationToken
+
+- (void)stop {
+    if (auto session = _session.lock()) {
+        session->unregister_progress_notifier(_token);
+    }
+}
 
+- (void)dealloc {
```
jpsim: This behavior is different than `RLMNotificationToken`:

```objc
- (void)dealloc {
    if (_realm || _block) {
        NSLog(@"RLMNotificationToken released without unregistering a notification. You must hold "
              @"on to the RLMNotificationToken returned from addNotificationBlock and call "
              @"-[RLMNotificationToken stop] when you no longer wish to receive RLMRealm notifications.");
    }
}
```

austinzheng: So the expected behavior is, upon dealloc, if `-stop` hasn't been called previously a `NSLog` warning message is fired?

jpsim: I'm not saying that should be the expected behavior, only that it currently diverges from `RLMNotificationToken`.

austinzheng: I changed it so it matches. Consistency is a good thing here.

austinzheng:
I removed the use of `Optional<T>`, since @morten-krogh said he would change sync to call the notifier immediately after registration, meaning that we should always know the downloadable and uploadable bytes counts. (https://github.com/realm/product/issues/40#issuecomment-271144959)

PR #4531, opened 2017-01-11
Use static func in protocol Equatable

sora0077:


jpsim:
CI failures are unrelated.

PR #4533, opened 2017-01-11
Add support for client reset

austinzheng:
Changes:
- Updated version of object store used by Cocoa binding
- Updated version of sync used by Cocoa binding to beta-7.0
- Added an error key representing client reset errors
- Client reset errors provide the location of the backed-up Realm file, as well as a block that can optionally be called in order to immediately carry out the reset process

HOW IT WORKS:

If the object store receives a "this Realm needs to be reset" error, it is propagated to the binding and eventually packaged as an `NSError` for the app.

The error's code is `RLMSyncErrorClientResetError`, and it contains two items in its user info dictionary: one is the path where the original Realm file will be backed up to upon client reset. The other item is an opaque block that the user can call to initiate the client reset process if they can guarantee they've invalidated all instances of the offending Realm. Otherwise, they can wait until the next time the app is launched, at which point the process will automatically run.

TODO:
- Tests
- Discuss whether this API is good or bad

Replaces https://github.com/realm/realm-cocoa/pull/4507.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,10 +24,10 @@ x.x.x Release notes (yyyy-MM-dd)
 
 ### Sync Breaking Changes (In Beta)
 
-* Underlying sync engine upgraded to version BETA-6.5.
 * Sync-related error reporting behavior has been changed. Errors not related
   to a particular user or session are only reported if they are classed as
   'fatal' by the underlying sync engine.
+* Added `RLMSyncErrorClientResetError` to `RLMSyncError` enum.
 
 ### API Breaking Changes
 
--- a/Realm/RLMSyncConfiguration.mm
+++ b/Realm/RLMSyncConfiguration.mm
@@ -35,7 +35,9 @@ namespace {
 using ProtocolError = realm::sync::ProtocolError;
 
 RLMSyncSystemErrorKind errorKindForSyncError(SyncError error) {
-    if (error.error_code == ProtocolError::bad_authentication) {
+    if (error.is_client_reset_requested()) {
+        return RLMSyncSystemErrorKindClientReset;
+    } else if (error.error_code == ProtocolError::bad_authentication) {
         return RLMSyncSystemErrorKindUser;
     } else if (error.is_session_level_protocol_error()) {
         return RLMSyncSystemErrorKindSession;
@@ -140,11 +142,16 @@ static BOOL isValidRealmURL(NSURL *url) {
             errorHandler = [=](std::shared_ptr<SyncSession> errored_session,
                                SyncError error) {
                 RLMSyncSession *session = [[RLMSyncSession alloc] initWithSyncSession:errored_session];
+                NSMutableDictionary *userInfo = [NSMutableDictionary dictionaryWithCapacity:error.user_info.size()];
+                for (auto& pair : error.user_info) {
+                    userInfo[@(pair.first.c_str())] = @(pair.second.c_str());
+                }
                 // FIXME: how should the binding respond if the `is_fatal` bool is true?
                 [[RLMSyncManager sharedManager] _fireErrorWithCode:error.error_code.value()
                                                            message:@(error.message.c_str())
                                                            isFatal:error.is_fatal
                                                            session:session
+                                                          userInfo:userInfo
                                                         errorClass:errorKindForSyncError(error)];
             };
         }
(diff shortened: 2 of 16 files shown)


austinzheng:
New dependency: https://github.com/realm/realm-object-store/pull/351

Realm/RLMSyncConfiguration.mm
```diff
@@ -140,11 +142,16 @@ - (instancetype)initWithUser:(RLMSyncUser *)user
             errorHandler = [=](std::shared_ptr<SyncSession> errored_session,
                                SyncError error) {
                 RLMSyncSession *session = [[RLMSyncSession alloc] initWithSyncSession:errored_session];
+                NSMutableDictionary *userInfo = [@{} mutableCopy];
```
tgoyne: This is kinda needlessly clever (and since we know the size of the dictionary could be insignificantly faster as `[NSMutableDictionary dictionaryWithCapacity:error.user_info.size()]`).

Realm/RLMSyncUtil.h
```diff
@@ -59,6 +65,16 @@ typedef RLM_ERROR_ENUM(NSInteger, RLMSyncError, RLMSyncErrorDomain) {
 
     /// An error that indicates an internal, unrecoverable error with the underlying synchronization engine.
     RLMSyncErrorClientInternalError     = 6,
+
+    /// An error that indicates the Realm needs to be reset.
```
tgoyne: Doc comments should have an empty line between the summary and extended description (but still with the /// prefix).

Realm/RLMSyncUtil.h
```diff
@@ -59,6 +65,16 @@ typedef RLM_ERROR_ENUM(NSInteger, RLMSyncError, RLMSyncErrorDomain) {
...
+    /// There are two keys in the `userInfo` dictionary of particular interest.
+    /// `kRLMSyncRecoveredRealmLocationPathKey` describes the path of the recovered copy of the Realm. It will not
+    /// actually exist until the app is restarted or the recovery block is called, whichever happens first.
+    /// `kRLMSyncImmediatelyRecoverRealmBlockKey` contains a recovery block that takes no arguments and returns nothing.
+    /// If your app can guarantee that all instances of the offending Realm can be closed and invalidated, you may call
+    /// this block to immediately reset the Realm. This creates a recovery copy and deletes the original files, allowing
+    /// the Realm to be re-opened and a fresh copy to be re-downloaded.
```
tgoyne: What happens if you don't immediately call the block or if you call the block some time later is unclear. A brief mention of why a Realm would need to be reset could help as well.

CHANGELOG.md
```diff
@@ -25,9 +25,11 @@ x.x.x Release notes (yyyy-MM-dd)
 ### Sync Breaking Changes (In Beta)
 
 * Underlying sync engine upgraded to version BETA-6.5.
+* Underlying sync engine upgraded to version BETA-7.0.
```
austinzheng: This needs to be removed.

RealmSwift/Sync.swift
```diff
@@ -99,6 +99,21 @@ public typealias SyncManagementObjectStatus = RLMSyncManagementObjectStatus
  */
 public typealias Provider = RLMIdentityProvider
 
+/// A namespace for utility methods related to the Realm Object Server.
+public struct SyncUtil {
+    /// Given a client reset error, extract and return the recovery file path and the reset closure.
+    public static func clientResetInfoFrom(error: SyncError) -> (String, () -> Void)? {
```
austinzheng: I think this needs to be just `clientResetInfo(from:)`.

Realm/RLMSyncSession.mm
```diff
@@ -24,10 +24,7 @@
 
 using namespace realm;
 
-@interface RLMSyncSession () {
-    std::weak_ptr<SyncSession> _session;
-}
-
+@interface RLMSyncSession ()
```
tgoyne: Might as well delete this entirely.

tgoyne:
Putting the helper function in an extension on SyncError rather than a static struct sounds good to me.

austinzheng:
It looks like CI's hit our GitHub rate limits for today.

austinzheng:
`verify-osx` and `verify-osx-object-server` both passed, so I'm going to merge.

PR #4535, opened 2017-01-12
Add support for progress notifiers

austinzheng:
Replaces https://github.com/realm/realm-cocoa/pull/4509.

TODO:
- API discussion
- Binding-level tests


austinzheng:
I added tests for the "streaming" notifiers.

Unfortunately, the data transfer rate between the app and the prepackaged realm object server is way too fast to test the non-streaming notifiers. Testing those requires being able to register a notifier at a point in time when sync has begun, but not yet finished, an upload or download. This is impossible to do reliably, as far as I can tell, and we have no APIs that would facilitate that task.

tgoyne:
PF can be used to throttle network connections (including on the loopback interface) so in theory we could do something like limit the connection to 1 KB/sec and then make a several KB commit and then have a several-second window where syncing should be in progress. Might be worth doing manually once to at least minimally verify that things work, but figuring out what testing hooks we need in the sync library would be a better use of time for building something robust that'd live in the actual test suite.

Realm/ObjectServerTests/RLMSyncTestCase.m
```diff
@@ -40,6 +40,21 @@ @interface RLMSyncTestCase ()
 @implementation SyncObject
 @end
 
+@implementation HugeSyncObject
+
++ (instancetype)object  {
+    const NSInteger fakeDataSize = 1000000;
+    HugeSyncObject *object = [[self alloc] init];
+    char fakeData[fakeDataSize];
+    for (NSInteger i=0; i<fakeDataSize; i++) {
```
tgoyne: `memset(fakeData, sizeof(fakeData), 16);`

Realm/RLMSyncSession.h
```diff
@@ -30,11 +30,57 @@ typedef NS_ENUM(NSUInteger, RLMSyncSessionState) {
...
+ */
+typedef void(^RLMProgressNotificationBlock)(NSUInteger transferred_bytes, NSUInteger transferrable_bytes);
+
 NS_ASSUME_NONNULL_BEGIN
 
 /**
+ A token object corresponding to a progress notifier block on a `RLMSyncSession`. To stop notifications manually,
+ destroy the token or call `-stop` on it.
+ */
+@interface RLMProgressNotificationToken : NSObject
```
tgoyne: This should just be a private subclass of `RLMNotifcationToken`.

Realm/RLMSyncSession.h
```diff
@@ -56,6 +102,32 @@ NS_ASSUME_NONNULL_BEGIN
...
-NS_ASSUME_NONNULL_END
+/**
+ Register a progress notification block. Multiple blocks can be registered on the same session at once. All blocks
+ will be dispatched to the main queue.
+
+ The token returned by this method must be retained as long as progress notifications are desired, and the `-stop`
+ method should be called on it when notifications are no longer needed.
+
+ If no token is returned, the session was not in a state where it could accept progress notifiers, or the notifier was
+ not a streaming notifier, was called immediately, and will not be called again since there is no additional progress
```
tgoyne: What is a "streaming notifier"? Why would the block be called immediately? Why would the session not be in a state where it can accept progress notifiers?

Realm/RLMSyncSession.h
```diff
@@ -56,6 +102,32 @@ NS_ASSUME_NONNULL_BEGIN
 /// session.
 - (nullable RLMSyncConfiguration *)configuration;
 
-NS_ASSUME_NONNULL_END
+/**
+ Register a progress notification block. Multiple blocks can be registered on the same session at once. All blocks
+ will be dispatched to the main queue.
```
tgoyne: Dispatching notifications to a queue rather than the thread the notification block was added on is inconsistent with all of our other notifications.

austinzheng: I'll rework how the progress notifications are stored to make this work the same way.

Realm/RLMSyncSession.h
austinzheng: @jpsim originally proposed making it not a subclass of `RLMNotificationToken`, with the justification that it would prevent users from passing in this token into methods which take a `RLMNotificationToken` but couldn't do anything meaningful with this token type. Do you have any thoughts on changing this token to be a subclass of `RLMNotificationToken`, JP?

tgoyne: I don't think there's really a big gain there (IMO there's nothing unclear about the semantics of passing a sync progress notification cancelation token to beginWrite, even if it's pointless), and I think having a single cancelation token type used for absolutely everything makes our API more approachable. It's nice to have the property of that a user never has to think about what sort of token this notification method will return or what they can do with it.

austinzheng:
@bdash, if you could take a look at my usage of the `CFRunLoop` APIs I would appreciate it. I had a bit of trouble getting the two sync notifier tests to work properly, so I might have been doing something incorrectly.

Realm/ObjectServerTests/RLMSyncTestCase.m
```diff
@@ -45,11 +45,10 @@ @implementation HugeSyncObject
 + (instancetype)object  {
     const NSInteger fakeDataSize = 1000000;
     HugeSyncObject *object = [[self alloc] init];
-    char fakeData[fakeDataSize];
-    for (NSInteger i=0; i<fakeDataSize; i++) {
-        fakeData[i] = 16;
-    }
-    object.dataProp = [NSData dataWithBytes:fakeData length:fakeDataSize];
+    char *fakeData = malloc(fakeDataSize);
```
bdash: No reason to use malloc here.

Realm/RLMSyncSession.mm
```diff
@@ -149,25 +150,24 @@ - (BOOL)waitForDownloadCompletionOnQueue:(dispatch_queue_t)queue callback:(void(
...
-                                                     direction:(RLMSyncNotifierDirection)direction
-                                                          mode:(RLMSyncNotifierMode)mode
-                                                         queue:(dispatch_queue_t)queue {
     if (auto session = _session.lock()) {
         if (session->state() == SyncSession::PublicState::Error) {
             return nil;
         }
+        // Get the current runloop.
+        CFRunLoopRef currentRunLoop = CFRunLoopGetCurrent();
+        if (!currentRunLoop) {
```
bdash: I don't think it's possible for `CFRunLoopGetCurrent` to return nil. `CFRunLoopGetCurrent` creates a runloop if one does not already exist.

Realm/ObjectServerTests/RLMObjectServerTests.m
```diff
@@ -735,31 +739,37 @@ - (void)testStreamingUploadNotifier {
...
-        transferred = xfr;
-        transferrable = xfb;
-        callCount++;
-        [lock unlock];
-    } direction:RLMSyncNotifierDirectionUpload mode:RLMSyncNotifierModeAlwaysReportLatest queue:queue];
+    __block RLMProgressNotificationToken *token;
+    XCTestExpectation *ex = [self expectationWithDescription:@"token expectation"];
+    dispatch_async(dispatch_get_main_queue(), ^{
+        token = [session addProgressNotificationBlock:^(NSUInteger xfr, NSUInteger xfb) {
+            [lock lockBeforeDate:[NSDate distantFuture]];
```
bdash: You can just use `[lock lock];`

bdash: Then again, this block should be invoked on the main thread so the locking looks to be unnecessary.

Realm/ObjectServerTests/RLMObjectServerTests.m
```diff
@@ -735,31 +739,37 @@ - (void)testStreamingUploadNotifier {
...
-        XCTAssert(xfr >= transferred);
-        XCTAssert(xfb >= transferrable);
-        transferred = xfr;
-        transferrable = xfb;
-        callCount++;
-        [lock unlock];
-    } direction:RLMSyncNotifierDirectionUpload mode:RLMSyncNotifierModeAlwaysReportLatest queue:queue];
+    __block RLMProgressNotificationToken *token;
+    XCTestExpectation *ex = [self expectationWithDescription:@"token expectation"];
+    dispatch_async(dispatch_get_main_queue(), ^{
```
bdash: Why is this necessary?

Realm/ObjectServerTests/RLMObjectServerTests.m
```diff
@@ -735,31 +739,37 @@ - (void)testStreamingUploadNotifier {
...
+    [self waitForExpectationsWithTimeout:5.0 handler:nil];
     // Upload lots of data
     [realm beginWriteTransaction];
     for (NSInteger i=0; i<NUMBER_OF_BIG_OBJECTS; i++) {
         [realm addObject:[HugeSyncObject object]];
     }
     [realm commitWriteTransaction];
     // Wait for upload to begin and finish
-    sleep(2);
+    usleep(500000);
```
bdash: Why is this sleep necessary?

Realm/ObjectServerTests/RLMObjectServerTests.m
```diff
@@ -735,31 +739,37 @@ - (void)testStreamingUploadNotifier {
...
+            XCTAssert(xfb >= transferrable);
+            transferred = xfr;
+            transferrable = xfb;
+            callCount++;
+            [lock unlock];
+        } direction:RLMSyncNotifierDirectionUpload mode:RLMSyncNotifierModeAlwaysReportLatest];
+        CFRunLoopRun();
+        [ex fulfill];
+    });
+    [self waitForExpectationsWithTimeout:5.0 handler:nil];
```
bdash: What purpose does this expectation serve?

Realm/RLMSyncSession.h
```diff
@@ -30,11 +30,57 @@ typedef NS_ENUM(NSUInteger, RLMSyncSessionState) {
...
 @class RLMSyncUser, RLMSyncConfiguration;
 
+/**
+ The type of a block intended for reporting a session's network activity to the user.
+
+ `transferred_bytes` refers to the number of bytes that have been uploaded or downloaded.
+
+ `transferrable_bytes` refers to the number of total bytes to be uploaded or downloaded.
+ */
+typedef void(^RLMProgressNotificationBlock)(NSUInteger transferred_bytes, NSUInteger transferrable_bytes);
```
bdash: Objective-C naming typically doesn't use snake_case.

Realm/RLMSyncSession.h
bdash: We need to decide on an approach here. @jpsim, can you please weigh in?

Realm/RLMSyncSession.mm
```diff
@@ -24,10 +24,49 @@
...
     std::weak_ptr<SyncSession> _session;
 }
+@end
+
+@implementation RLMProgressNotificationToken
+
+- (void)stop {
+    if (auto session = _session.lock()) {
+        session->unregister_progress_notifier(_token);
+        _session = {};
```
bdash: `.reset()` would probably be clearer.

Realm/RLMSyncSession.h
```diff
@@ -30,11 +30,57 @@ typedef NS_ENUM(NSUInteger, RLMSyncSessionState) {
...
+    /// The block will be called forever, or until the user unregisters it.
+    /// It will always report the latest number of transferred bytes, and the
+    /// latest number of total bytes to be transferred.
+    RLMSyncNotifierModeAlwaysReportLatest,
+    /// The block will, upon registration, store the total number of bytes
+    /// to be transferred. When invoked, it will always report the latest number
+    /// of transferred bytes out of that original number of bytes to be
+    /// transferred. When the number of transferred bytes reaches or exceeds the
+    /// number of bytes to be transferred, the block will be unregistered.
+    RLMSyncNotifierModeProgressIndicator,
```
bdash: The naming of these enum cases doesn't seem great. It's also different than the terminology used in the Swift API.

austinzheng:
Any ideas for better case names than `RLMSyncNotifierModeAlwaysReportLatest` and `RLMSyncNotifierModeProgressIndicator` are very welcome.

Realm/RLMSyncSession.h
```diff
@@ -30,11 +30,57 @@ typedef NS_ENUM(NSUInteger, RLMSyncSessionState) {
...
+ */
+typedef NS_ENUM(NSUInteger, RLMSyncNotifierDirection) {
+    /// For monitoring upload progress.
+    RLMSyncNotifierDirectionUpload,
+    /// For monitoring download progress.
+    RLMSyncNotifierDirectionDownload,
+};
+
+/**
+ The desired behavior of a particular sync session notifier block.
```
bdash: It's a little strange to me that this refers to things as a "sync session notifier block" when the API is for a "progress notification block".

Realm/RLMSyncSession.h
austinzheng: I'll let him comment in more detail, but JP proposed we have the token inherit from `RLMNotificationToken`, and have the writes ignoring notifications API throw an exception if you pass in the sync progress notifier token.

Realm/RLMSyncSession.h
bdash: It seems to me that the distinction is between reporting the instantaneous status indefinitely vs reporting progress for the currently outstanding work. "Always report latest" somewhat captures the behavior of the former, but "progress indicator" doesn't at all hint at the semantics of the latter.

Realm/RLMSyncSession.h
jpsim: Yeah, I basically agree with Thomas.

austinzheng:
This PR depends on https://github.com/realm/realm-object-store/pull/332, which should be going in soon.

Realm/RLMSyncSession.h
austinzheng: How about `RLMSyncNotifierModeAlwaysReportLatestValues` and `RLMSyncNotifierModeReportRelativeToTransferrableValueAtTimeOfRegistration`?

Realm/RLMSyncSession.h
```diff
@@ -30,11 +32,59 @@ typedef NS_ENUM(NSUInteger, RLMSyncSessionState) {
     RLMSyncSessionStateInvalid
 };
 
+/**
+ The type of data transfer that a particular progress notification block will track.
+
+ Progress notification blocks can be registered on sessions if your app wishes to be informed
+ how many bytes have been uploaded or downloaded, for example to show progress indicator UIs.
+ */
+typedef NS_ENUM(NSUInteger, RLMSyncNotifierDirection) {
```
bdash: The "sync notifier" terminology remained here. Perhaps `RLMSyncProgressDirection`?

Realm/RLMSyncSession.h
```diff
@@ -56,6 +106,25 @@ NS_ASSUME_NONNULL_BEGIN
...
+ Register a progress notification block. Multiple blocks can be registered on the same session at once. All blocks
+ will be run on the same thread that they were registered on. The progress notification block will always be called
+ once immediately upon registration, in order to provide up-to-date information.
+
+ The token returned by this method must be retained as long as progress notifications are desired, and the `-stop`
+ method should be called on it when notifications are no longer needed.
+
+ If no token is returned, the notifier will not be ever be called again. There are a number of reasons this might
+ be true. If the session has previously experienced a fatal error it will not accept progress notifiers.
+ If the notifier was configured in the `RLMSyncNotifierModeProgressIndicator` mode but there is no additional
```
bdash: Old name for the enum member here.

Realm/RLMSyncSession.h
```diff
@@ -56,6 +106,25 @@ NS_ASSUME_NONNULL_BEGIN
...
+ progress to report (for example, the number of transferrable bytes and transferred bytes are equal), the notifier
+ will not be called again.
 
+ @see: `RLMSyncNotifierDirection`, `RLMSyncNotifierMode`, `RLMProgressNotificationBlock`
```
bdash: `RLMSyncNotifier` terminology remains here, too.

Realm/RLMSyncSession.h
```diff
@@ -30,11 +32,59 @@ typedef NS_ENUM(NSUInteger, RLMSyncSessionState) {
...
+ `transferred_bytes` refers to the number of bytes that have been uploaded or downloaded.
+
+ `transferrable_bytes` refers to the number of total bytes to be uploaded or downloaded.
+ */
+typedef void(^RLMProgressNotificationBlock)(NSUInteger transferredBytes, NSUInteger transferrableBytes);
+
 NS_ASSUME_NONNULL_BEGIN
 
 /**
+ A token object corresponding to a progress notifier block on a `RLMSyncSession`. To stop notifications manually,
```
bdash: progress _notification_ block?

Realm/RLMSyncSession.h
```diff
@@ -56,6 +106,25 @@ NS_ASSUME_NONNULL_BEGIN
...
-NS_ASSUME_NONNULL_END
+/**
+ Register a progress notification block. Multiple blocks can be registered on the same session at once. All blocks
+ will be run on the same thread that they were registered on. The progress notification block will always be called
+ once immediately upon registration, in order to provide up-to-date information.
+
+ The token returned by this method must be retained as long as progress notifications are desired, and the `-stop`
+ method should be called on it when notifications are no longer needed.
+
+ If no token is returned, the notifier will not be ever be called again. There are a number of reasons this might
```
bdash: Should "notifier" be "notification block" in this paragraph? "Notifier" would seem like the thing doing the notifying, not the thing asking to be notified.

RealmSwift/Sync.swift
```diff
@@ -469,6 +469,97 @@ fileprivate extension SyncManagementObjectStatus {
     }
 }
 
+public extension SyncSession {
+    /// Whether a given notifier block will report the upload status or the download status of its session.
```
bdash: s/notifier/notification/?

RealmSwift/Sync.swift
```diff
@@ -469,6 +469,97 @@ fileprivate extension SyncManagementObjectStatus {
...
+    public enum ProgressDirection {
+        /// :nodoc:
+        case upload
+        /// :nodoc:
+        case download
+    }
+    /// How the notifier block will behave.
```
bdash: s/notifier/notification/?

RealmSwift/Sync.swift
```diff
@@ -469,6 +469,97 @@ fileprivate extension SyncManagementObjectStatus {
...
+
+        /// :nodoc:
+        fileprivate init(transferred: UInt, pending: UInt) {
+            bytesTransferred = transferred
+            bytesPending = pending
+        }
+    }
+
+    /**
+     Register a progress notifier to the session. This progress notifier allows your app to be
```
bdash: s/notifier/notification block/?

"Register … to the session" should probably be "with the session".

Realm/RLMSyncSession.h
```diff
@@ -30,11 +32,59 @@ typedef NS_ENUM(NSUInteger, RLMSyncSessionState) {
...
+};
+
 @class RLMSyncUser, RLMSyncConfiguration;
 
+/**
+ The type of a progress notification block intended for reporting a session's network activity to the user.
+
+ `transferred_bytes` refers to the number of bytes that have been uploaded or downloaded.
+
+ `transferrable_bytes` refers to the number of total bytes to be uploaded or downloaded.
```
bdash: The variable names in this documentation comment need updated to match their names in the declaration below.

Realm/RLMSyncSession.h
```diff
@@ -56,6 +106,25 @@ NS_ASSUME_NONNULL_BEGIN
 /// session.
 - (nullable RLMSyncConfiguration *)configuration;
 
-NS_ASSUME_NONNULL_END
+/**
+ Register a progress notification block. Multiple blocks can be registered with the same session at once. All blocks
+ will be run on the same thread that they were registered on. The progress notification block will always be called
```
bdash: "All blocks will be run on the same thread that they were registered on" -> "Each block will be invoked from the runloop of the thread on which it was registered". The existing text threw me off because it starts by saying "all blocks will be run on the same thread".

RealmSwift/Sync.swift
```diff
@@ -469,6 +469,97 @@ fileprivate extension SyncManagementObjectStatus {
...
+
+    /**
+     A struct encapsulating progress information, as well as useful helper methods.
+     */
+    public struct Progress {
+        /// The number of bytes that have been transferred.
+        public let bytesTransferred: UInt
+        /// The total number of bytes that have been transferred, and are pending transfer.
+        /// If the notification block is tracking downloads, this number represents the size of the
+        /// changesets generated by all other clients using the database.
```
bdash: s/database/Realm/?

RealmSwift/Sync.swift
```diff
@@ -469,6 +469,97 @@ fileprivate extension SyncManagementObjectStatus {
...
+     */
+    public typealias NotificationToken = RLMProgressNotificationToken
+
+    /**
+     A struct encapsulating progress information, as well as useful helper methods.
+     */
+    public struct Progress {
+        /// The number of bytes that have been transferred.
+        public let bytesTransferred: UInt
+        /// The total number of bytes that have been transferred, and are pending transfer.
```
bdash: I'm a bit confused by this `bytesPending` property. I'd expect `bytesPending` to decrease as data as uploaded, but it looks from how it's used within `fractionTransferred` that this value does not change as data is uploaded. Am I missing something?

austinzheng: No, you're right. The naming is bad.

Realm/RLMSyncSession.h
```diff
@@ -56,6 +105,26 @@ NS_ASSUME_NONNULL_BEGIN
 /// session.
 - (nullable RLMSyncConfiguration *)configuration;
 
-NS_ASSUME_NONNULL_END
+/**
+ Register a progress notification block. Multiple blocks can be registered with the same
```
bdash: The first sentence should probably stand alone as the one-line summary as the method, with the remainder of the paragraph being moved to a second paragraph.

Realm/RLMSyncSession.h
```diff
@@ -56,6 +105,26 @@ NS_ASSUME_NONNULL_BEGIN
...
+
+ If no token is returned, the notification block will never be called again. There are a number of reasons this might
+ be true. If the session has previously experienced a fatal error it will not accept progress notification blocks.
+ If the block was configured in the `RLMSyncProgressForCurrentlyOutstandingWork` mode but there is no additional
+ progress to report (for example, the number of transferrable bytes and transferred bytes are equal), the block
+ will not be called again.
 
+ @see: `RLMSyncProgressDirection`, `RLMSyncProgress`, `RLMProgressNotificationBlock`, `RLMProgressNotificationToken`
+ */
+- (nullable RLMProgressNotificationToken *)addProgressNotificationBlock:(RLMProgressNotificationBlock)block
```
bdash: Hmm, I just realized that this API doesn't follow the standard practice of the block being the last parameter to make it easier to use inline blocks. Can we we reorder things so that the block is last?

austinzheng:
CI is still problematic. OS X tests passed locally.

Issue #4534, opened 2017-01-11
`swift-version.sh` hanging on Spotlight call


larkost:
There have been a couple of jobs recently that have failed with timeouts (35 minutes in this case), and the message `No version of Xcode found that supports Swift 2.2` in their logs (version of Swift does change). The line that is causing the problem is:
```mdfind "kMDItemCFBundleIdentifier == 'com.apple.dt.Xcode'"```
... and that will hang when you run it manually in an VM that gets into this state. There is no real reason that this situation occurs, and nothing obvious in the log about what is going on. In one case I was able to recover by re-indexing the `/Applications` folder, but in other cases a reboot has been required.

In the latter case all calls to `mdfind` or `mdimport`hung the same way, and calls to `mdutil` reported a communications error and then told me that 

The most straightforward way of handling this is to first use a `glob` to find Xcode versions by name in `/Applications`, and then if what we are looking for is not there, then fall-back to Spotlight. In the case of CI, then we are always putting them there, and the fallback will work for other users if they choose to put versions elsewhere.

larkost:
Fixed in 942fd88.

Issue #4503, opened 2017-01-03
make Xcode selection for CI deterministic


larkost:
In CI runs the script `./scripts/swift-version.sh` to select the version of Xcode used for swift compiles. For most situations it is near-deterministic, but the listing order depends on Spotlights ordering (so a black box to us), as well as what the last version of Xcode selected on the system was.

We should make this completely deterministic, either by being explicit on the Swift->Xcode mappings, or by skipping the currently selected check and making the ordering completely deterministic. I lean to making this explicit, as it will better point out failures in the setup.

larkost:
Fixed in 942fd88.

PR #4536, opened 2017-01-12
fix for #4534, find xcode without spotlight

larkost:
This avoids using spotlight to find Xcode if it can, avoiding some times when Spotlight has been unreliable on the CO machines. It also solves #4503 by making the selection of Xcode on CI machines deterministic (given identical installs of Xcode).

### DIFF
--- a/scripts/swift-version.sh
+++ b/scripts/swift-version.sh
@@ -1,37 +1,66 @@
+#!/usr/bin/env bash
+
 get_swift_version() {
-    swift_command=$@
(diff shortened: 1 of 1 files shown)


jpsim:
This doesn't seem to have had the intended effect.

image

larkost:
There was a debugging line that got left in, and that turned into the first line in the output when `build.sh get-version` is called by `generate-rlmversion.sh`, so was being put into `RLMVersion.h`.

jpsim:
Thanks @larkost !

PR #4537, opened 2017-01-12
Add object-level notifications with information about what properties changed

tgoyne:
~~The Swift 2 version of the code currently doesn't compile (I need to reinstall Xcode 7) and~~ there's probably a few more things that could use testing, but otherwise I think this is all ready for review. The Swift API should be exactly what was originally proposed (with parts 1-4 implemented (i.e. everything but keypath filtering)), with the obj-c API "inspired" by that and the exist collection notifications.

### DIFF
--- a/Realm/RLMObject.h
+++ b/Realm/RLMObject.h
@@ -23,10 +23,12 @@
 
 NS_ASSUME_NONNULL_BEGIN
 
+@class RLMNotificationToken;
+@class RLMObjectSchema;
+@class RLMPropertyChange;
 @class RLMPropertyDescriptor;
 @class RLMRealm;
 @class RLMResults;
-@class RLMObjectSchema;
 
 /**
  `RLMObject` is a base class for model objects representing data stored in Realms.
@@ -39,13 +41,13 @@ NS_ASSUME_NONNULL_BEGIN
      @property NSString *name;
      @property BOOL      adopted;
      @end
- 
+
      // Dog.m
      @implementation Dog
      @end //none needed
- 
+
  ### Supported property types
- 
+
  - `NSString`
  - `NSInteger`, `int`, `long`, `float`, and `double`
  - `BOOL` or `bool`
@@ -56,17 +58,17 @@ NS_ASSUME_NONNULL_BEGIN
  - `RLMArray<X>`, where `X` is an `RLMObject` subclass, to model many-to-many relationships.
 
  ### Querying
- 
+
  You can initiate queries directly via the class methods: `allObjects`, `objectsWhere:`, and `objectsWithPredicate:`.
  These methods allow you to easily query a custom subclass for instances of that class in the default Realm.
- 
+
  To search in a Realm other than the default Realm, use the `allObjectsInRealm:`, `objectsInRealm:where:`,
  and `objectsInRealm:withPredicate:` class methods.
- 
+
  @see `RLMRealm`
- 
+
  ### Relationships
- 
+
  See our [Cocoa guide](https://realm.io/docs/objc/latest#relationships) for more details.
 
  ### Key-Value Observing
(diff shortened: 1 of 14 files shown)


Realm/RLMObject.h
```diff
@@ -428,13 +482,41 @@ NS_ASSUME_NONNULL_BEGIN
...
 
+/**
+ Information about a specific property which changed in an RLMObject change notification.
+ */
+@interface RLMPropertyChange : NSObject
+
+/**
+ The name of the property which changed.
+ */
+@property (nonatomic, strong) NSString *name;
```
bdash: Should these properties be `readonly`?

Realm/RLMObject.h
```diff
@@ -428,13 +482,41 @@ NS_ASSUME_NONNULL_BEGIN
 
 @end
 
+/**
+ Information about a specific property which changed in an RLMObject change notification.
```
bdash: It looks like we wrap type names in backticks elsewhere in our documentation comments.

Realm/RLMObject.h
```diff
@@ -428,13 +482,41 @@ NS_ASSUME_NONNULL_BEGIN
...
+ The name of the property which changed.
+ */
+@property (nonatomic, readonly, strong) NSString *name;
+
+/**
+ The value of the property before the change occurred. This will always be `nil`
+ if the change happened on the same thread as the notification and for `RLMArray`
+ properties.
+
+ For object properties this will give the object which was previously linked to,
```
bdash: Does this apply to the Swift API as well? It's not mentioned in the comment for the corresponding Swift property.

Realm/RLMObject.h
```diff
@@ -403,12 +405,64 @@ NS_ASSUME_NONNULL_BEGIN
...
+
+ If the object is modified, the block will be called with `deleted` set to
+ `false`, a `nil` `error, and an array of `RLMPropertyChange` objects which
+ indicate which properties of the objects were modified.
+
+ If an error occurs, `deleted` will be false, `changes` will be `nil`, and
+ `error` will include information about the error. The block will never be
+ called again after an error occurs.
+ */
+typedef void (^RLMObjectChangeBlock)(bool deleted, NSArray<RLMPropertyChange *> *_Nullable changes, NSError *_Nullable error);
```
jpsim: is there a reason not to use an objc `BOOL` here?

Realm/RLMObject.h
```diff
@@ -403,12 +405,64 @@ NS_ASSUME_NONNULL_BEGIN
...
+ notifications can't be delivered instantly, multiple notifications may be
+ coalesced into a single notification.
+
+ Unlike with `RLMArray` and `RLMResults`, there is no "initial" callback made
+ after you add a new notification block.
+
+ You must retain the returned token for as long as you want updates to continue
+ to be sent to the block. To stop receiving updates, call `-stop` on the token.
+
+ It is safe to capture a strong reference to the observed object within the
```
jpsim: nice that you documented this!

Realm/RLMObject_Private.h
```diff
@@ -75,6 +75,12 @@ FOUNDATION_EXTERN id _Nullable RLMValidatedValueForProperty(id object, NSString
 // Compare two RLObjectBases
 FOUNDATION_EXTERN BOOL RLMObjectBaseAreEqual(RLMObjectBase * _Nullable o1, RLMObjectBase * _Nullable o2);
 
+typedef void (^RLMObjectNotificationCallback)(NSArray<NSString *> *_Nullable propertyNames,
```
jpsim: Why not use `RLMObjectChangeBlock` here?

RealmSwift/Object.swift
```diff
@@ -193,6 +193,50 @@ open class Object: RLMObjectBase, ThreadConfined {
         }
     }
 
+    // MARK: Notifications
+
+    /**
+     Observe this object for changes, including those made on other threads and
```
jpsim: The wording differs between the objc and swift docs and could probably mostly be realigned.

Realm/RLMObject_Private.h
tgoyne: Initially it was different because I thought `RealmOptional` needed some special handling, and then I didn't bother changing it because I don't think it'd actually meaningfully simply or shorten the code.

PR #4545, opened 2017-01-13
Fix an issue that crashing by failing to cast `RLMError.Code` when bridging `RLMError` to `Realm.Error`

kishikawakatsumi:
Fix an issue that crashing by failing to cast `RLMError.Code` when bridging `RLMError` to `Realm.Error`

Currently, the following code crashes 

```swift
 let realm: Realm?
 do {
     realm = try Realm()
 } catch Realm.Error.incompatibleLockFile {
     print("Realm Browser app may be attached to Realm on device?")
 }
```

with the error below.
```
Could not cast value of type '__C.RLMError.Code' (0x101b501b8) to 'RealmSwift.Realm.Error.Code' (0x101e78e00).
```

CC @jpsim @austinzheng

### DIFF
--- a/RealmSwift/Error.swift
+++ b/RealmSwift/Error.swift
@@ -28,50 +28,49 @@ extension Realm {
      ```swift
      let realm: Realm?
      do {
-         realm = Realm()
+         realm = try Realm()
      } catch Realm.Error.incompatibleLockFile {
          print("Realm Browser app may be attached to Realm on device?")
      }
      ```
     */
     public struct Error {
-        public typealias Code = RLMError.Code
-
         /// Error thrown by Realm if no other specific error is returned when a realm is opened.
-        public static let fail: Code = .fail
+        public static let fail = RLMError.fail
 
         /// Error thrown by Realm for any I/O related exception scenarios when a realm is opened.
-        public static let fileAccess: Code = .fileAccess
+        public static let fileAccess = RLMError.fileAccess
 
         /// Error thrown by Realm if the user does not have permission to open or create
         /// the specified file in the specified access mode when the realm is opened.
-        public static let filePermissionDenied: Code = .filePermissionDenied
+        public static let filePermissionDenied = RLMError.filePermissionDenied
 
         /// Error thrown by Realm if the file already exists when a copy should be written.
-        public static let fileExists: Code = .fileExists
+        public static let fileExists = RLMError.fileExists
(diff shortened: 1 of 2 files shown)


RealmSwift/Error.swift
```diff
@@ -274,14 +219,4 @@ public func == (lhs: ErrorType, rhs: ErrorType) -> Bool { // swiftlint:disable:t
         && lhs._domain == rhs._domain
 }
 
-// MARK: Pattern Matching
-
-/**
- Pattern matching matching for `Realm.Error`, so that the instances can be used with Swift's
- `do { ... } catch { ... }` syntax.
-*/
-public func ~= (lhs: Error, rhs: ErrorType) -> Bool { // swiftlint:disable:this valid_docs
```
jpsim: This is a breaking change since we're removing a public API.

kishikawakatsumi: @jpsim Hmm, but the pattern matching operator has never been called.

kishikawakatsumi: Actually, I mistook the lines to delete. The code I really wanted to delete is Swift 3's code.

jpsim:
The problem with this @kishikawakatsumi is that it breaks the public API, which means that even if this is something that's good to merge, we can't do so until Realm 3.x.

Is there a way to fix this without breaking the public API?

kishikawakatsumi:
@jpsim Removing pattern matching operator `~= (lhs: Realm.Error, rhs: Error) -> Bool` doesn't break anything. None of the codes below invokes the operator.

```swift
do {
    let realm = try Realm()
} catch Realm.Error.fail {
    ...
} catch {
    ...
}
```

```swift
do {
    let realm = try Realm()
} catch Realm.Error.fail {
    ...
} catch {
    ...
}
```

```swift
let error = Realm.Error.fail
switch error {
case .fail:
    ...
default:
    ...
}
```

```swift
Realm.Error.fail ~= Realm.Error.incompatibleLockFile
Realm.Error.fail ~= userDefinedError
```

kishikawakatsumi:
I reverted removing pattern matching operator. It isn't needed to fix the crash.

kishikawakatsumi:
This crash was fixed by https://github.com/realm/realm-cocoa/commit/433d41641fa30fa1e81106030cc25e6c545c4987

PR #4546, opened 2017-01-13
Delete custom pattern matching operator for `RealmSwift.Error` in Swift 2.3

kishikawakatsumi:
Custom pattern matching operator for `RealmSwift.Error` seems not working.
The following code doesn't work as expected:

```swift
let realm: Realm?
do {
    realm = try Realm()
} catch RealmSwift.Error.IncompatibleLockFile() {
    print("Incompatible lock file. The Realm Browser app might be attached to a Realm on the device.")
} catch  {
    // Always enter here 
}
```

It always enter `catch {` instead `} catch RealmSwift.Error.IncompatibleLockFile {` 

CC @jpsim @austinzheng


kishikawakatsumi:
No longer support Swift 2.x

PR #4563, opened 2017-01-19
Expose Permission Realm

mrackwitz:
Hey, it's been a while here! 👋 

This exposes the server-side API (which wasn't publicly launched yet 🤐) in the client by defining the model and convenience methods.

* [x] Expose models
* [x] Convenience getters to access permission-Realm on `*SyncUser`
* [x] Changelog entry
* [x] Tests

Along the way:
* Removes a superfluous forward declaration
* Refers consistently to the "Management Realm"

/c @jpsim @kishikawakatsumi @teotwaki


kishikawakatsumi:
@mrackwitz Is it intentional that `createdAt` property was removed from the `Permission` schema?

mrackwitz:
@kishikawakatsumi: Yes, because this isn't tracked in the admin-Realm.

kishikawakatsumi:
The Permission object seems to be synchronized only in the Professional Edition. I will investigate it.

austinzheng:
Is this just blocked on the issue of the Permissions Realm being synchronized only on PE?

kishikawakatsumi:
@austinzheng Yes. To use Professional Edition, it works fine and pass unit tests.

Realm/RLMSyncPermission.h
```diff
@@ -0,0 +1,54 @@
...
+ See https://realm.io/docs/realm-object-server/#permissions for general
+ documentation.
+ */
+@interface RLMSyncPermission : RLMObject
+
+/// The date this object was last modified.
+@property (readonly) NSDate *updatedAt;
+
+/// The identity of a user affected by this permission change.
+@property (readonly) NSString *userId;
```
mrackwitz: No, as the permission-Realm reflects as well the permissions of Realms you own and have shared with others.

Realm/RLMSyncPermission.h
```diff
@@ -0,0 +1,54 @@
...
+
+ See https://realm.io/docs/realm-object-server/#permissions for general
+ documentation.
+ */
+@interface RLMSyncPermission : RLMObject
+
+/// The date this object was last modified.
+@property (readonly) NSDate *updatedAt;
+
+/// The identity of a user affected by this permission change.
```
mrackwitz: This should say instead `The identity of a user affected by this permission.`

Realm/RLMSyncPermission.h
nirinchev: Oh, so that's the user id of the owner of the Realm? I assumed that it's the id of the grantee (i.e. the user whose permission realm we're inspecting).

mrackwitz: No, it's the ID of the user who is affected by the given permission, that can be either the owner or someone else who got access to this Realm.

ianpward:
@bmunkholm any update on this?

jpsim:
@ianpward see https://github.com/realm/product/issues/214#issuecomment-284591352

kishikawakatsumi:
Using ROS 1.2.1, now unit tests are passed. Get ready to merge @jpsim @ianpward

jpsim:
I've confirmed the failed CI jobs passed locally or they're pre-existing occasional failures. Merging this. Thanks @kishikawakatsumi @mrackwitz! 👏

Issue #4569, opened 2017-01-20
Fix sync token refresh logic


austinzheng:
There are at least two problems with the token refresh logic, which is designed to pre-emptively refresh a sync token before it expires.

1. The `RLMSyncSessionRefreshHandle` class is an object that encapsulates the idea of periodically refreshing a session's token, preferably before it expires. It uses a `NSTimer`. Unfortunately, this class is always constructed on `NSURLSession`'s callback worker thread. As a result, the timer is never fired. The most straightforward fix is to add the timer to the main queue's runloop instead.

2. When the `RLMSyncSessionRefreshHandle` refreshes the token, it currently passes up the entire Realm URL (e.g. `realm://192.168.1.180:9080/f8ea086a230d1f4aab8b75071b55c9e5/realmtasks`). This is rejected by the authentication server. Using the path (`/~/realmtasks`) works. This needs to be fixed in the client.

PR #4570, opened 2017-01-20
Fix issues with automatic sync token refresh subsystem

austinzheng:
Changes:
- Sync token refresh timers will now be registered upon the main queue
- Sync token refresh handles will now be properly created with the path to the Realm (instead of the full URL)
- Made the logic to calculate the time at which the token is refreshed more resilient to bad values

Fixes https://github.com/realm/realm-cocoa/issues/4569. Related to https://github.com/realm/realm-sync/issues/1114 and https://github.com/realm/RealmTasks/issues/384.

### DIFF
--- a/Realm/ObjectServerTests/RLMSyncSessionRefreshHandle+ObjectServerTests.h
+++ b/Realm/ObjectServerTests/RLMSyncSessionRefreshHandle+ObjectServerTests.h
@@ -1,6 +1,6 @@
 ////////////////////////////////////////////////////////////////////////////
 //
-// Copyright 2016 Realm Inc.
+// Copyright 2017 Realm Inc.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
@@ -18,17 +18,8 @@
 
 #import "RLMSyncSessionRefreshHandle.h"
 
-#import <memory>
+@interface RLMSyncSessionRefreshHandle (ObjectServerTests)
 
-namespace realm {
-class SyncSession;
-}
++ (void)calculateFireDateUsingTestLogic:(BOOL)forTest blockOnRefreshCompletion:(void(^)(BOOL))block;
 
-@class RLMSyncUser;
-
-@interface RLMSyncSessionRefreshHandle ()
-
-- (instancetype)initWithPathToRealm:(NSString *)path
-                               user:(RLMSyncUser *)user
-                            session:(std::shared_ptr<realm::SyncSession>)session;
 @end
--- a/Realm/ObjectServerTests/RLMTestUtils.h
+++ b/Realm/ObjectServerTests/RLMTestUtils.h
@@ -16,19 +16,5 @@
 //
 ////////////////////////////////////////////////////////////////////////////
 
-#import "RLMSyncSessionRefreshHandle.h"
-
-#import <memory>
-
-namespace realm {
-class SyncSession;
-}
-
-@class RLMSyncUser;
-
-@interface RLMSyncSessionRefreshHandle ()
-
-- (instancetype)initWithPathToRealm:(NSString *)path
-                               user:(RLMSyncUser *)user
-                            session:(std::shared_ptr<realm::SyncSession>)session;
-@end
+void RLMSwapOutClassMethod(id classObject, SEL original, SEL swizzled);
+void RLMSwapOutInstanceMethod(id classObject, SEL original, SEL swizzled);
--- a/Realm/ObjectServerTests/RLMTestUtils.m
+++ b/Realm/ObjectServerTests/RLMTestUtils.m
@@ -1,0 +1,53 @@
+////////////////////////////////////////////////////////////////////////////
+//
+// Copyright 2016 Realm Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+////////////////////////////////////////////////////////////////////////////
+
+#import <Foundation/Foundation.h>
+
+#import "RLMTestUtils.h"
+
+#import <objc/runtime.h>
+
+static void RLMSwapOutMethod(Class class,
+                             SEL original, Method originalMethod,
+                             SEL swizzled, Method swizzledMethod) {
+    if (class_addMethod(class,
+                        original,
+                        method_getImplementation(swizzledMethod),
+                        method_getTypeEncoding(swizzledMethod))) {
+        class_replaceMethod(class,
+                            swizzled,
+                            method_getImplementation(originalMethod),
+                            method_getTypeEncoding(originalMethod));
+    } else {
+        method_exchangeImplementations(originalMethod, swizzledMethod);
+    }
+}
+
+void RLMSwapOutClassMethod(id classObject, SEL original, SEL swizzled) {
+    Class class = object_getClass((id)classObject);
+    RLMSwapOutMethod(class,
+                     original, class_getClassMethod(class, original),
+                     swizzled, class_getClassMethod(class, swizzled));
+}
+
+void RLMSwapOutInstanceMethod(id classObject, SEL original, SEL swizzled) {
+    Class class = [classObject class];
+    RLMSwapOutMethod(class,
+                     original, class_getInstanceMethod(class, original),
+                     swizzled, class_getInstanceMethod(class, swizzled));
+}
--- a/Realm/RLMSyncSessionRefreshHandle.h
+++ b/Realm/RLMSyncSessionRefreshHandle.h
@@ -1,6 +1,6 @@
 ////////////////////////////////////////////////////////////////////////////
 //
-// Copyright 2016 Realm Inc.
+// Copyright 2017 Realm Inc.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
@@ -16,19 +16,15 @@
 //
 ////////////////////////////////////////////////////////////////////////////
 
-#import "RLMSyncSessionRefreshHandle.h"
-
-#import <memory>
-
-namespace realm {
-class SyncSession;
-}
+#import <Foundation/Foundation.h>
 
 @class RLMSyncUser;
 
-@interface RLMSyncSessionRefreshHandle ()
+/// An object that handles refreshing a session's token periodically, as long
+/// as the session remains live and valid.
+@interface RLMSyncSessionRefreshHandle : NSObject
 
-- (instancetype)initWithPathToRealm:(NSString *)path
-                               user:(RLMSyncUser *)user
-                            session:(std::shared_ptr<realm::SyncSession>)session;
+- (void)scheduleRefreshTimer:(NSDate *)dateWhenTokenExpires;
+- (void)invalidate;
+
 @end
--- a/Realm/RLMSyncSessionRefreshHandle.mm
+++ b/Realm/RLMSyncSessionRefreshHandle.mm
@@ -33,40 +33,137 @@ using namespace realm;
 }
 
 @property (nonatomic, weak) RLMSyncUser *user;
-@property (nonatomic, strong) NSString *fullURLPath;
+@property (nonatomic, strong) NSString *pathToRealm;
 @property (nonatomic) NSTimer *timer;
 
 @end
 
 @implementation RLMSyncSessionRefreshHandle
 
-- (instancetype)initWithFullURLPath:(NSString *)urlPath
+- (instancetype)initWithPathToRealm:(NSString *)path
                                user:(RLMSyncUser *)user
                             session:(std::shared_ptr<realm::SyncSession>)session {
     if (self = [super init]) {
-        self.fullURLPath = urlPath;
+        self.pathToRealm = path;
         self.user = user;
-        _session = session;
+        _session = std::move(session);
         return self;
     }
     return nil;
 }
 
-- (void)invalidate {
+- (void)dealloc {
     [self.timer invalidate];
-    self.user = nil;
 }
 
-- (void)scheduleRefreshTimer:(NSTimeInterval)fireTime {
-    static const NSInteger refreshBuffer = 10;
+- (void)invalidate {
     [self.timer invalidate];
-    self.timer = [[NSTimer alloc] initWithFireDate:[NSDate dateWithTimeIntervalSince1970:(fireTime - refreshBuffer)]
-                                          interval:0
-                                            target:self
-                                          selector:@selector(timerFired:)
-                                          userInfo:nil
-                                           repeats:NO];
-    [[NSRunLoop currentRunLoop] addTimer:self.timer forMode:NSDefaultRunLoopMode];
+}
+
++ (NSDate *)fireDateForTokenExpirationDate:(NSDate *)date {
+    static const NSTimeInterval refreshBuffer = 10;
+    NSDate *fireDate = [date dateByAddingTimeInterval:-refreshBuffer];
+    // Only fire times in the future are valid.
+    return ([fireDate compare:[NSDate date]] != NSOrderedDescending ? fireDate : nil);
+}
+
+- (void)scheduleRefreshTimer:(NSDate *)dateWhenTokenExpires {
+    // Schedule the timer on the main queue.
+    // It's very likely that this method will be run on a side thread, for example
+    // on the thread that runs `NSURLSession`'s completion blocks. We can't be
+    // guaranteed that there's an existing runloop on those threads, and we don't want
+    // to create and start a new one if one doesn't already exist.
+    dispatch_async(dispatch_get_main_queue(), ^{
+        [self.timer invalidate];
+        NSDate *fireDate = [RLMSyncSessionRefreshHandle fireDateForTokenExpirationDate:dateWhenTokenExpires];
+        if (!fireDate) {
+            [self.user _unregisterRefreshHandleForURLPath:self.pathToRealm];
+            return;
+        }
+        self.timer = [[NSTimer alloc] initWithFireDate:fireDate
+                                              interval:0
+                                                target:self
+                                              selector:@selector(timerFired:)
+                                              userInfo:nil
+                                               repeats:NO];
+        [[NSRunLoop currentRunLoop] addTimer:self.timer forMode:NSDefaultRunLoopMode];
+    });
+}
+
+/// Handler for network requests whose responses successfully parse into an auth response model.
+- (BOOL)_handleSuccessfulRequest:(RLMAuthResponseModel *)model strongUser:(RLMSyncUser *)user {
+    // Success
+    if (auto session = _session.lock()) {
+        if (session->state() != SyncSession::PublicState::Error) {
+            session->refresh_access_token([model.accessToken.token UTF8String], none);
+            NSDate *expiration = [NSDate dateWithTimeIntervalSince1970:model.accessToken.tokenData.expires];
+            [self scheduleRefreshTimer:expiration];
+            return YES;
+        }
+    }
+    // The session is dead or in a fatal error state.
+    [user _unregisterRefreshHandleForURLPath:self.pathToRealm];
+    [self.timer invalidate];
+    return NO;
+}
+
+/// Handler for network requests that failed before the JSON parsing stage.
+- (BOOL)_handleFailedRequest:(NSError *)error strongUser:(RLMSyncUser *)user {
+    // Something else went wrong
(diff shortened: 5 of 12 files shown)


Realm/RLMSyncSessionRefreshHandle.mm
```diff
@@ -53,20 +53,38 @@ - (instancetype)initWithFullURLPath:(NSString *)urlPath
...
-    static const NSInteger refreshBuffer = 10;
-    [self.timer invalidate];
-    self.timer = [[NSTimer alloc] initWithFireDate:[NSDate dateWithTimeIntervalSince1970:(fireTime - refreshBuffer)]
-                                          interval:0
-                                            target:self
-                                          selector:@selector(timerFired:)
-                                          userInfo:nil
-                                           repeats:NO];
-    [[NSRunLoop currentRunLoop] addTimer:self.timer forMode:NSDefaultRunLoopMode];
+    constexpr NSInteger REFRESH_BUFFER = 10;
```
bdash: This naming style is typically reserved for macros. Shouldn't this be a `NSTimeInterval`?

Realm/RLMSyncSessionRefreshHandle.mm
```diff
@@ -53,20 +53,38 @@ - (instancetype)initWithFullURLPath:(NSString *)urlPath
...
 - (void)invalidate {
-    [self.timer invalidate];
-    self.user = nil;
+    dispatch_async(dispatch_get_main_queue(), ^{
+        [self.timer invalidate];
+        self.user = nil;
+    });
 }
 
 - (void)scheduleRefreshTimer:(NSTimeInterval)fireTime {
```
bdash: From the rest of this code `fireTime` appears to contain a time interval from 1970 representing the time at which the token will expire? It'd be a lot clearer to convert this to only expose this from `RLMTokenDataModel` as an `NSDate` so we don't have to reason here about how to interpret this value.

Realm/RLMSyncSessionRefreshHandle.mm
```diff
@@ -53,20 +53,38 @@ - (instancetype)initWithFullURLPath:(NSString *)urlPath
...
+    constexpr NSInteger REFRESH_BUFFER = 10;
+    // Schedule the timer on the main queue.
+    // It's very likely that this method will be run on a side thread, for example
+    // on the thread that runs `NSURLSession`'s completion blocks. We can't be
+    // guaranteed that there's an existing runloop on those threads, and we don't want
+    // to create and start a new one if one doesn't already exist.
+    dispatch_async(dispatch_get_main_queue(), ^{
+        [self.timer invalidate];
+        // The fire time is `REFRESH_BUFFER` seconds before the token expires, but it also
+        // must be at least `REFRESH_BUFFER` seconds in the future from now.
```
bdash: I can see why we want to schedule the refresh 10 seconds before the token expires. Why do we not want to schedule it if we're already within 10 seconds of it expiring?

Realm/RLMSyncSessionRefreshHandle.mm
```diff
@@ -33,18 +33,18 @@ @interface RLMSyncSessionRefreshHandle () {
...
 
-- (instancetype)initWithFullURLPath:(NSString *)urlPath
+- (instancetype)initWithPathToRealm:(NSString *)path
                                user:(RLMSyncUser *)user
                             session:(std::shared_ptr<realm::SyncSession>)session {
     if (self = [super init]) {
-        self.fullURLPath = urlPath;
+        self.pathToRealm = path;
         self.user = user;
         _session = session;
```
bdash: We may as well move `session` here.

Realm/RLMSyncSessionRefreshHandle.mm
```diff
@@ -53,20 +53,38 @@ - (instancetype)initWithFullURLPath:(NSString *)urlPath
...
-                                          selector:@selector(timerFired:)
-                                          userInfo:nil
-                                           repeats:NO];
-    [[NSRunLoop currentRunLoop] addTimer:self.timer forMode:NSDefaultRunLoopMode];
+    constexpr NSInteger REFRESH_BUFFER = 10;
+    // Schedule the timer on the main queue.
+    // It's very likely that this method will be run on a side thread, for example
+    // on the thread that runs `NSURLSession`'s completion blocks. We can't be
+    // guaranteed that there's an existing runloop on those threads, and we don't want
+    // to create and start a new one if one doesn't already exist.
```
bdash: Another option would be to use a dispatch timer source instead.

Realm/RLMSyncSessionRefreshHandle.mm
```diff
@@ -53,20 +53,38 @@ - (instancetype)initWithFullURLPath:(NSString *)urlPath
 }
 
 - (void)invalidate {
-    [self.timer invalidate];
-    self.user = nil;
+    dispatch_async(dispatch_get_main_queue(), ^{
```
bdash: Do we also need to invalidate the timer in `-dealloc`?

Realm/RLMSyncSessionRefreshHandle.mm
austinzheng: My reasoning was that, if the server was misconfigured and sending tokens that expired e.g. only after 5 seconds, this check would prevent clients from hammering the server with requests.

Thinking about it more, though, the server would still be hammered; the codepath would just be different (5 seconds after a token is expired the object store gets the "not logged in" error and the client still tries to refresh the token). I'll remove it.

Realm/RLMSyncSessionRefreshHandle.mm
austinzheng: Yes it should, and it shouldn't be a constexpr. Will fix.

Realm/RLMSyncSessionRefreshHandle.mm
austinzheng: Good idea.

Realm/RLMSyncSessionRefreshHandle.mm
austinzheng: I'm inclined to leave it as-is for now, but I'll read up on dispatch sources and dispatch timers and we can refactor it if this becomes a problem.

Realm/RLMSyncSessionRefreshHandle.mm
bdash: It's not necessary to `dispatch_async` here in order to call `-invalidate`. `NSTimer` is one of the classes listed as being thread safe.

austinzheng:
I tried adding a test, but adding needed support files completely broke my test setup, so that'll have to wait until Monday.

austinzheng:
I added an end-to-end test for the subsystem this PR fixes. Any further problems should be caught by that test. CI is still broken.

radu-tutueanu:
I tested this branch with the RealmTasks iOS app and 1.0.0 ROS with the expiration time set to 60 sec.

It does not look like the timer gets fired(I set a breakpoint in `timerFired` ). The app refreshes the token only after the server disconnects the current session.

App logs:
```
RealmTasks[8183:109523] Sync: Connection[1]: Received: ERROR(error_code=202, message_size=20, try_again=0, session_ident=1)
2017-01-23 14:54:34.544125 RealmTasks[8183:109523] Sync: Connection[1]: Connection closed due to error reported by server: Access token expired (202)
2017-01-23 14:54:53.345471 RealmTasks[8183:109523] Sync: Connection[1]: Resolving '127.0.0.1:9080'
2017-01-23 14:54:53.345935 RealmTasks[8183:109523] Sync: Connection[1]: Connecting to endpoint '127.0.0.1:9080' (1/1)
2017-01-23 14:54:53.346163 RealmTasks[8183:109523] Sync: Connection[1]: Connected to endpoint '127.0.0.1:9080' (from '127.0.0.1:54093')
2017-01-23 14:54:53.351409 RealmTasks[8183:109523] Sync: Connection[1]: Session[1]: Sending: BIND(server_path='/bc51c3d96ef1ec9cf4260e51e60138bb/realmtasks', signed_user_token_size=613, need_file_ident_pair=0)
2017-01-23 14:54:53.351553 RealmTasks[8183:109523] Sync: Connection[1]: Session[1]: Sending: IDENT(server_file_ident=4336788893534473065, client_file_ident=2, client_file_ident_secret=7897167207874186617, scan_server_version=31, scan_client_version=56, latest_server_version=31, latest_server_session_ident=4762414798728238705)
```

austinzheng:
I talked to Radu, and it looks like CocoaPods has some sort of issue pulling in this PR's branch, since the code files it changed weren't touched. I wonder if the `/` in the name is confusing CocoaPods.

austinzheng:
Trying to pull in the branch using the commit SHA also doesn't seem to work. 🤔  Also see: https://github.com/CocoaPods/CocoaPods/issues/2386

austinzheng:
@bigfish24 wants this merged soon, so any final reviews would be appreciated.

Realm/ObjectServerTests/RLMSyncSessionRefreshHandle+ObjectServerTests.h
```diff
@@ -0,0 +1,26 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+////////////////////////////////////////////////////////////////////////////
+
+#import "RLMSyncSessionRefreshHandle.h"
+
+@interface RLMSyncSessionRefreshHandle (ObjectServerTests)
+
++ (void)rlmTestUtils_calculateFireDateUsingTestLogic:(BOOL)forTest
```
jpsim: there's no need to prefix this method like this given that it's defined and implemented in our unit tests only. Could just be `+calculateFireDateUsingTestLogic:blockOnRefreshCompletion:`

Realm/RLMSyncSessionRefreshHandle.h
```diff
@@ -0,0 +1,30 @@
...
+// limitations under the License.
+//
+////////////////////////////////////////////////////////////////////////////
+
+#import <Foundation/Foundation.h>
+
+@class RLMSyncUser;
+
+/// An object that handles refreshing a session's auth token periodically, as long as that session remains viable.
+/// Intended for easy removal once the new auth system is in place.
```
jpsim: Is there further discussion somewhere on this "new auth system"? Would be good to link to it here for context.

austinzheng: I'll delete this. It was a reference to the new auth system Adam was talking about for some time a few months ago. I don't think that new auth system is relevant anymore to our discussions.

austinzheng:
Squashed into one commit, addressed feedback. Hopefully a new test job will be triggered.

Realm/ObjectServerTests/RLMTestUtils.m
```diff
@@ -0,0 +1,55 @@
...
+
+#import <Foundation/Foundation.h>
+
+#import "RLMTestUtils.h"
+
+#import <objc/runtime.h>
+
+void rlmTestUtils_swapOutMethod(id classObject, SEL original, SEL swizzled, BOOL isClass);
+
+void rlmTestUtils_swapOutMethod(id classObject, SEL original, SEL swizzled, BOOL isClass) {
```
tgoyne: This function should be `static` and not have a declaration.

Realm/ObjectServerTests/RLMTestUtils.m
```diff
@@ -0,0 +1,55 @@
...
+    Class class = isClass ? object_getClass((id)classObject) : [classObject class];
+    Method originalMethod = (isClass
+                             ? class_getClassMethod(class, original)
+                             : class_getInstanceMethod(class, original));
+    Method swizzledMethod = (isClass
```
tgoyne: It seems like it'd be simpler to just have the wrapper functions pass these in rather than switching on the bool.

Realm/ObjectServerTests/RLMTestUtils.h
```diff
@@ -0,0 +1,20 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+////////////////////////////////////////////////////////////////////////////
+
+void rlmTestUtils_swapOutClassMethod(id classObject, SEL original, SEL swizzled);
```
tgoyne: All of our other test helpers just have the normal RLM prefix.

Realm/RLMSyncSessionRefreshHandle.mm
```diff
@@ -33,40 +33,132 @@ @interface RLMSyncSessionRefreshHandle () {
...
+
+- (void)scheduleRefreshTimer:(NSDate *)dateWhenTokenExpires {
+    // Schedule the timer on the main queue.
+    // It's very likely that this method will be run on a side thread, for example
+    // on the thread that runs `NSURLSession`'s completion blocks. We can't be
+    // guaranteed that there's an existing runloop on those threads, and we don't want
+    // to create and start a new one if one doesn't already exist.
+    dispatch_async(dispatch_get_main_queue(), ^{
+        [self.timer invalidate];
+        // The fire time is `self.refreshBuffer` seconds before the token expires.
```
bdash: This comment no longer makes sense in this context.

Realm/RLMSyncSessionRefreshHandle.mm
```diff
@@ -33,40 +33,132 @@ @interface RLMSyncSessionRefreshHandle () {
...
+        NSDate *fireDate = [RLMSyncSessionRefreshHandle fireDateForTokenExpirationDate:dateWhenTokenExpires];
+        if (!fireDate) {
+            [self.user _unregisterRefreshHandleForURLPath:self.pathToRealm];
+            return;
+        }
+        NSTimer *t = [[NSTimer alloc] initWithFireDate:fireDate
```
bdash: Assign directly to `self.timer`?

Realm/RLMSyncSessionRefreshHandle.mm
```diff
@@ -33,40 +33,132 @@ @interface RLMSyncSessionRefreshHandle () {
...
+                                                target:self
+                                              selector:@selector(timerFired:)
+                                              userInfo:nil
+                                               repeats:NO];
+        [[NSRunLoop currentRunLoop] addTimer:t forMode:NSDefaultRunLoopMode];
+        self.timer = t;
+    });
+}
+
+- (BOOL)onRefreshCompletionWithError:(NSError *)error json:(NSDictionary *)json {
```
bdash: Why not split success and error cases into two separate methods? It looks like there's no shared code here, so we can easily move the `if (error)` into the caller and have it choose between a success and error method.

Realm/RLMSyncSessionRefreshHandle.mm
```diff
@@ -33,40 +33,132 @@ @interface RLMSyncSessionRefreshHandle () {
...
-                                          interval:0
-                                            target:self
-                                          selector:@selector(timerFired:)
-                                          userInfo:nil
-                                           repeats:NO];
-    [[NSRunLoop currentRunLoop] addTimer:self.timer forMode:NSDefaultRunLoopMode];
+}
+
++ (NSDate *)fireDateForTokenExpirationDate:(NSDate *)date {
+    static NSTimeInterval refreshBuffer = 10;
```
tgoyne: `static const`

Realm/RLMSyncSessionRefreshHandle.mm
```diff
@@ -33,40 +33,132 @@ @interface RLMSyncSessionRefreshHandle () {
...
+            [self.user _unregisterRefreshHandleForURLPath:self.pathToRealm];
+            return;
+        }
+        NSTimer *t = [[NSTimer alloc] initWithFireDate:fireDate
+                                              interval:0
+                                                target:self
+                                              selector:@selector(timerFired:)
+                                              userInfo:nil
+                                               repeats:NO];
+        [[NSRunLoop currentRunLoop] addTimer:t forMode:NSDefaultRunLoopMode];
```
tgoyne: Unless `_unregisterRefreshHandleForURLPath` needs to be run on the main thread, this can skip the dispatch to the main thread and just do `[[NSRunLoop mainRunLoop] addTimer:...`

Realm/RLMSyncSessionRefreshHandle.mm
```diff
@@ -33,40 +33,132 @@ @interface RLMSyncSessionRefreshHandle () {
...
+                                             code:RLMSyncErrorBadResponse
+                                         userInfo:@{kRLMSyncUnderlyingErrorKey: error}];
+    [[RLMSyncManager sharedManager] _fireError:syncError];
+    NSDate *nextFireDate = nil;
+    // Certain errors should trigger a retry.
+    if (error.domain == NSURLErrorDomain) {
+        switch (error.code) {
+            case NSURLErrorCannotConnectToHost:
+                // FIXME: 120 seconds is an arbitrarily chosen value, consider rationalizing it.
+                nextFireDate = [NSDate dateWithTimeIntervalSinceNow:120];
```
bdash: Why does this deserve a longer delay than the others?

Realm/RLMSyncSessionRefreshHandle.mm
```diff
@@ -33,40 +33,132 @@ @interface RLMSyncSessionRefreshHandle () {
...
+                break;
+            case NSURLErrorNotConnectedToInternet:
+            case NSURLErrorNetworkConnectionLost:
+            case NSURLErrorTimedOut:
+            case NSURLErrorDNSLookupFailed:
+            case NSURLErrorCannotFindHost:
+                // FIXME: 30 seconds is an arbitrarily chosen value, consider rationalizing it.
```
bdash: 30 seconds seems like quite a long time?

Realm/RLMSyncSessionRefreshHandle.mm
```diff
@@ -33,40 +33,132 @@ @interface RLMSyncSessionRefreshHandle () {
...
+                nextFireDate = [NSDate dateWithTimeIntervalSinceNow:30];
+                break;
+            default:
+                break;
```
bdash: Is doing nothing really the appropriate behavior for all other error types? How will apps recover from other errors?

Realm/RLMSyncSessionRefreshHandle.mm
```diff
@@ -33,40 +33,132 @@ @interface RLMSyncSessionRefreshHandle () {
...
+                                               repeats:NO];
+        [[NSRunLoop currentRunLoop] addTimer:t forMode:NSDefaultRunLoopMode];
+        self.timer = t;
+    });
+}
+
+- (BOOL)onRefreshCompletionWithError:(NSError *)error json:(NSDictionary *)json {
+    RLMSyncUser *user = self.user;
+    if (!user) {
+        return NO;
```
bdash: What's the significance of the return value. The caller doesn't seem to use it.

Realm/RLMSyncSessionRefreshHandle.mm
```diff
@@ -33,40 +33,132 @@ @interface RLMSyncSessionRefreshHandle () {
...
+    }
+    if (json && !error) {
+        RLMAuthResponseModel *model = [[RLMAuthResponseModel alloc] initWithDictionary:json
+                                                                    requireAccessToken:YES
+                                                                   requireRefreshToken:NO];
```
bdash: Alignment is a bit out of whack here.

Realm/RLMSyncSessionRefreshHandle.mm
bdash: `NSRunLoop` is documented as not being thread-safe.

Realm/RLMSyncSessionRefreshHandle.mm
austinzheng: The return value is checked by the new test case.

Realm/RLMSyncSessionRefreshHandle.mm
tgoyne: Huh, you can use most of the CFRunLoop functions from other threads but not the NSRunLoop ones. That's a bit odd.

Realm/RLMSyncSessionRefreshHandle.mm
austinzheng: The original idea was that we didn't want to overwhelm the host. I don't know how well-supported that assumption is, though. I think a delay of 10 seconds for all the network failure cases is probably good enough.

Realm/RLMSyncSessionRefreshHandle.mm
austinzheng: Apologies, I don't see it. All the colons are lined up, it seems?

bdash: Oh, yeah, that's how Xcode does things. My mistake.

Realm/RLMSyncSessionRefreshHandle.mm
austinzheng: I was going to do that in my follow-up PR, but I'll do it here instead.

Realm/RLMSyncSessionRefreshHandle.mm
austinzheng: The only errors that can occur are network errors, or errors I consider fatal (e.g. malformed JSON response, internal server error). That being said, my follow-up PR has some error recovery code.

austinzheng:
One of these test failures is because the build timed out after 35 minutes, the other is a known transient failure unrelated to sync.

PR #4576, opened 2017-01-23
Fix error behavior when binding session for the first time

austinzheng:
~Changes:~
- ~If the initial bind-session network request fails due to a network error, the request will be automatically retried. (This is how the pre-emptive token refreshing system works already.)~
- ~Changed some internal APIs to remove unused arguments and make implementing the above easier~

Fix error behavior when binding session for the first time

Changes:
- Improved the refresh handle logic to handle both initial binds and refreshing
- Removed the code in `RLMSyncUser.mm` to handle initial binds
- This means that initial binds that fail because of network conditions will use the same retry logic as refresh attempts

### DIFF
--- a/Realm/RLMSyncConfiguration.mm
+++ b/Realm/RLMSyncConfiguration.mm
@@ -129,14 +129,12 @@ static BOOL isValidRealmURL(NSURL *url) {
         if (!isValidRealmURL(url)) {
             @throw RLMException(@"The provided URL (%@) was not a valid Realm URL.", [url absoluteString]);
         }
-        auto bindHandler = [=](const std::string& path,
+        auto bindHandler = [=](auto&,
                                const SyncConfig& config,
                                const std::shared_ptr<SyncSession>& session) {
-            [user _bindSessionWithPath:path
-                                config:config
-                               session:session
-                            completion:[RLMSyncManager sharedManager].sessionCompletionNotifier
-                          isStandalone:NO];
+            [user _bindSessionWithConfig:config
+                                 session:session
+                              completion:[RLMSyncManager sharedManager].sessionCompletionNotifier];
         };
         if (!errorHandler) {
             errorHandler = [=](std::shared_ptr<SyncSession> errored_session,
--- a/Realm/RLMSyncSessionRefreshHandle.mm
+++ b/Realm/RLMSyncSessionRefreshHandle.mm
@@ -23,40 +23,62 @@
 #import "RLMSyncManager_Private.h"
 #import "RLMSyncUser_Private.hpp"
 #import "RLMTokenModels.h"
+#import "RLMUtil.hpp"
 
 #import "sync/sync_session.hpp"
+
+typedef NS_ENUM(NSUInteger, RLMSyncHandleMode) {
+    RLMSyncHandleModeBind,          // Initial `bind` of a session
+    RLMSyncHandleModeRetryBind,     // Retrying the initial `bind` of a session
+    RLMSyncHandleModeRefresh,       // Preemptive refreshing of a session
+};
 
 using namespace realm;
 
 @interface RLMSyncSessionRefreshHandle () {
     std::weak_ptr<SyncSession> _session;
+    std::shared_ptr<SyncSession> _strongSession;
 }
 
 @property (nonatomic, weak) RLMSyncUser *user;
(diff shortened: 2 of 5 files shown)


austinzheng:
I need to think about this a bit more.

austinzheng:
Okay, I think this can be reviewed (after #4570 is approved). It's 26 more lines of code, but the code is easier to understand now and logic isn't duplicated anymore. Let me know if there are any questions.

Realm/RLMSyncConfiguration.mm
```diff
@@ -129,14 +129,12 @@ - (instancetype)initWithUser:(RLMSyncUser *)user
         if (!isValidRealmURL(url)) {
             @throw RLMException(@"The provided URL (%@) was not a valid Realm URL.", [url absoluteString]);
         }
-        auto bindHandler = [=](const std::string& path,
+        auto bindHandler = [=](__unused const std::string& path,
```
bdash: In cases like this it's often simpler to just omit the parameter name instead.

austinzheng: Oh, right. This is C++. 🤦‍♂️

Realm/RLMSyncSessionRefreshHandle.mm
```diff
@@ -23,40 +23,62 @@
...
+    RLMSyncHandleModeBind,          // Initial `bind` of a session
+    RLMSyncHandleModeRetryBind,     // Retrying the initial `bind` of a session
+    RLMSyncHandleModeRefresh,       // Preemptive refreshing of a session
+};
+
 using namespace realm;
 
 @interface RLMSyncSessionRefreshHandle () {
     std::weak_ptr<SyncSession> _session;
+    std::shared_ptr<SyncSession> _strong_session;
```
bdash: Using `snake_case` here is inconsistent with the other ivars of this class.

Realm/RLMSyncSessionRefreshHandle.mm
```diff
@@ -125,21 +190,34 @@ - (BOOL)_handleFailedRequest:(NSError *)error strongUser:(RLMSyncUser *)user {
...
+    if (!nextTryDate) {
+        // This error isn't a network failure error. Just invalidate the refresh handle and stop.
+        [user _unregisterRefreshHandleForURLPath:self.pathToRealm];
+        [self invalidate];
+        return NO;
+    }
+    switch (self.mode) {
+        case RLMSyncHandleModeBind:
+            // A failed initial bind always results in retry binds.
+            _session = std::move(_strongSession);
```
tgoyne: Moving from a shared_ptr to a weak_ptr doesn't do anything.

Realm/RLMSyncSessionRefreshHandle.mm
```diff
@@ -23,40 +23,62 @@
...
+        _strongSession = std::move(session);
+        // Immediately fire off the network request.
+        [self _timerFired:nil];
         return self;
     }
     return nil;
 }
 
 - (void)dealloc {
+    _strongSession = nullptr;
```
tgoyne: This is unnecessary.

austinzheng:
@tgoyne is right: the state stuff is unnecessary. I've changed the code a bit to reflect that.

Realm/RLMSyncSessionRefreshHandle.mm
```diff
@@ -93,29 +107,60 @@ - (void)scheduleRefreshTimer:(NSDate *)dateWhenTokenExpires {
 /// Handler for network requests whose responses successfully parse into an auth response model.
 - (BOOL)_handleSuccessfulRequest:(RLMAuthResponseModel *)model strongUser:(RLMSyncUser *)user {
     // Success
-    if (auto session = _session.lock()) {
-        if (session->state() != SyncSession::PublicState::Error) {
-            session->refresh_access_token([model.accessToken.token UTF8String], none);
-            NSDate *expiration = [NSDate dateWithTimeIntervalSince1970:model.accessToken.tokenData.expires];
-            [self scheduleRefreshTimer:expiration];
-            return YES;
+    std::shared_ptr<SyncSession> session = _strongSession ?: _session.lock();
```
austinzheng: I think this can just be `session = _session.lock()`, since `_strongSession` only exists to ensure the session remains alive until we enter this method (or the corresponding failure method).

PR #4580, opened 2017-01-24
Use sync error domain errors from network client

stel:
Network client handles some sync-related (e.g. authentication) errors that should be reported by SyncManager directly.

Note that currently ROS never reports `RLMSyncAuthErrorUserAlreadyExists`, `RLMSyncAuthErrorUserDoesNotExist` codes and I'm not sure if that's changed. So probably this codes should be removed and `InvalidCredentials` could be moved to `RLMSyncError` enum.

@austinzheng, @jpsim let me know what you think.


stel:
Sorry for a delay on this. I've updated the PR and added a corresponding test to check the reported error domain.

jpsim:
> Note that currently ROS never reports `RLMSyncAuthErrorUserAlreadyExists`, `RLMSyncAuthErrorUserDoesNotExist` codes and I'm not sure if that's changed. So probably this codes should be removed and `InvalidCredentials` could be moved to `RLMSyncError` enum.

I don't know if this comment is still accurate after your latest changes, but that's not quite right, as "user already exists" could be reported if the current user is an admin.

stel:
It doesn't look like "user already exists" can be reported in any cases at all anymore (@mrackwitz please correct me if I'm wrong) while "user doesn't exists" can be reported when working with permissions, so it's not really a "SyncDomain" error.

I think if nothing is planned to be changed back on a server-side we should deprecate `RLMSyncAuthErrorUserAlreadyExists` and `RLMSyncAuthErrorUserDoesNotExist`.

PR #4581, opened 2017-01-24
Fix a typo, reformat docstrings for easier reading

### LINKED ISSUE #4579: Typo/repetition in the inline docs `addProgressNotificationForDirection:mode:block:`
The [inline docs state](https://github.com/realm/realm-cocoa/blob/master/Realm/RLMSyncSession.h#L120) (emphasis mine):
> The progress notification block will always be called once immediately upon registration **in order to** provide up-to-date registration **in order to** provide the latest available status information.

It feels like we should only keep one of the statements.

austinzheng:
Add line breaks to make things easier to read, and fix a typo. No functional change.

Fixes #4579.


jpsim:
Most of these changes need to be mirrored in the Swift docs too.

PR #4595, opened 2017-01-27
Change the way a private header is used by the object server test code

austinzheng:
Fixes an issue where the object server tests target refused to build after JP's module changes.


jpsim:
All CI failures are unrelated and it passed all the interesting stuff. Merging. Thanks again Austin.

PR #4604, opened 2017-01-30
remove reset simulators script

jpsim:
it's possible that this workaround is no longer needed now that Realm doesn't support building with Xcode 7 any more. Marked as "in progress" because it's possible a CI run will indicate that it is in fact still needed.


jpsim:
I'll see if I can test this locally rather than test on CI.

PR #4609, opened 2017-01-30
Clarify documentation around `delete()` method

### LINKED ISSUE #4591: Exception when trying to delete suffix of Results
Hi! I experienced the following bug:

1. **Goals:** get first 100 objects from realm and then delete them
2. **Expected results**: objects deleted
3. **Actual results**: Received exception: `*** Terminating app due to uncaught exception 'RLMException', reason: 'Index 950 is out of bounds (must be less than 950)'`
4. **Steps to reproduce**: populate Realm with 1000 elements and then run the following code:

```swift
let realm = try! Realm()
let results = realm.objects(Model.self).suffix(100)

try! realm.write {
    realm.delete(results)
}
```
Realm version: v2.3.0
Xcode version: 8.2.1
iOS/OSX version: iOS 10
Dependency manager + version: Carthage

austinzheng:
Fixes #4591
NFC

### DIFF
--- a/RealmSwift/Realm.swift
+++ b/RealmSwift/Realm.swift
@@ -359,10 +359,17 @@ public final class Realm {
     /**
      Deletes zero or more objects from the Realm.
 
+     Do not pass in a slice to a `Results` or any other auto-updating Realm collection
+     type (for example, the type returned by the Swift `suffix(_:)` standard library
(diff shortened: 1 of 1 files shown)


austinzheng:
If it does become possible, it would be a nice non-breaking enhancement.

Issue #1385, opened 2015-01-21
Support queries from within migrations


alazier:
This could perhaps be superseded by full dynamic api support.

mkolodny:
+1

jdanthinne:
Any update on this?
I'd like to be able to delete some data from a Class, ie. 
```swift
let deleteFilter = NSPredicate(format: "title = %@", "Delete me")
migration.enumerateObjects(ofType: Ad.self, filter: deleteFilter) { oldObject, newObject in
    migration.delete(oldObject)
}
```
or
```swift
migration.deleteObjects(realm.objects(Message.self).filter(format: "title = %@", "Delete me"))
```
or something like that…

jpsim:
This still hasn't been built. @jdanthinne if you're looking for a way to contribute to Realm, this seems like a great place to start!

jdanthinne:
I must admit I'm not fluent with ObjC, but here's my best try…

PR #4611, opened 2017-01-31
#1385: allow enumeration with a filter

jdanthinne:


jpsim:
Looks like this would be genuinely useful. It'd be nice to add tests for this, and probably add the filter as an optional default argument in Swift.

jpsim:
@jdanthinne please reopen a new PR if this is something you intend to complete. Thanks!

PR #4619, opened 2017-02-01
[WIP] Add wrapper API for sync permissions

austinzheng:


bigfish24:
I like it. @dhmspector what do you think from demo app perspective? @cmelchior @kristiandupont @jpsim will need other binding input as well.

austinzheng:
I'll note I didn't use the `PermissionManager` type in the Java proposal since this implementation seemed more idiomatic for Cocoa, but that's not to say other bindings shouldn't use it if it makes sense.

Realm/RLMSyncUser.h
```diff
@@ -113,8 +140,118 @@ NS_SWIFT_UNAVAILABLE("Use the full version of this API.");
...
+                        which a notification token should be passed out.
+ @param callback        A block through which the status of the permission
+                        offer creation process should be reported.
+ @return    An identifier for the newly-created permission offer. This
+            identifier can be passed into any method taking an `offerKey`
+            parameter. Do not confuse this with the permission offer's
+            token, which is passed back asynchronously through the callback
+            once the server has processed the request.
+ */
+- (NSString *)offerPermissionsForURL:(NSURL *)realmURL
```
bigfish24: This method was quite confusing for me. The passing in of a pointer to an `RLMNotificationToken` isn't ideal. Then it took me a few paths over the docs to understand what the key is in regards to the token in the `RLMSyncOfferChangeBlock` is compared to the `token` parameter.

My recommendation would be to have this method return an `RLMNotificationToken` just like the `modifyRealmPermissionsForURL:` method and get rid of the concept of a key. The permission offer's token should be the way to identify it and there is no reason to use the other methods until you have the token anyway.

Realm/RLMSyncUser.h
```diff
@@ -113,8 +140,118 @@ NS_SWIFT_UNAVAILABLE("Use the full version of this API.");
...
+ Revoke a permission offer which was previously generated.
+
+ @param offerKey        The identifier of the previously-generated permission
+                        offer (not the token).
+ @param callback        An optional callback block, to asynchronously report
+                        the progress of the operation.
+ @return    If a callback was specified, this will be a Realm notification
+            token. Call `-stop` if you want to deactivate the callback.
+ */
+- (nullable RLMNotificationToken *)deleteOffer:(NSString *)offerKey
```
bigfish24: Per my comment above, I think this should accept a permission offer token instead of a key.

Realm/RLMSyncUser.h
```diff
@@ -113,8 +140,118 @@ NS_SWIFT_UNAVAILABLE("Use the full version of this API.");
...
+ in a Realm), and then retrieved the next time the app is launched.
+
+ @param offerKey        The identifier of the previously-generated permission
+                        offer (not the token).
+ @param callback        A callback block, to asynchronously report
+                        the progress of the operation.
+ @return    A Realm notification token. Call `-stop` if you want to deactivate
+            the callback.
+ */
+- (nullable RLMNotificationToken *)addCallbackToExistingOffer:(NSString *)offerKey
```
bigfish24: I think it is odd we have a method to assist across restarts for permission offer but not change... however, in general I don't like the key concept at all. If users really want to work with permissions in an offline scenario, they should skip the convenience methods and just work with the objects in the `management.realm` directly. This is an advanced scenario and I think fits going to the docs to learn how to do this.

austinzheng: To be honest, I really don't like this "convenience methods versus working directly with the objects" dichotomy. Objects are a great way to build the underpinnings of an API. I think they are a horrible, developer-hostile way to implement the user-facing part of an API. We should look towards building an API which can eventually replace direct manipulation of the objects, or not bother at all.

bigfish24: I strongly disagree. It is a fundamental piece of using RMP so I find it bizarre that it would be described as "hostile" to an advanced user.

austinzheng: When you use RMP you define your own objects and your own semantics for them. That's not the case for the Management or Permission Realms. A good API is one that performs exactly one clearly defined operation. Giving a consumer (an app developer) a Realm with a whole bunch of knobs, most of which are completely irrelevant to whatever specific operation they want to perform, is error-prone and difficult to understand.

bigfish24: I still strongly disagree. It isn't different semantics, it is the same thing if I wanted to create an object to trigger some other server process and then update the object in response. This is central to using RMP beyond basic device to device sync or data backup.

austinzheng: I'd like to think what the other binding devs think. Did the original Realm Object based design even go through API review?

Realm/RLMSyncUser.h
austinzheng: I agree. Passing multiple return values through pointer arguments is inelegant, especially if both values are equal in importance (note that the Cocoa convention is pretty much to use it just for `NSError`).

AndyDentFree:
General opinions - when we have a moderately complex data model:

**Pros**

1. If they have it exposed as a Realm then they have idiomatic ways to work with it in their language.
1. An API is visible in Intellisense/code-completion in an IDE.
1. A Realm can be inspected thoroughly and (presumably) triggering established to warn of change.

**Cons**

1. Any time we expose something as a Realm we have to be prepared for all valid Realm operations. If the model does not allow for that, _with the current semantic expressiveness of Realm_ then we are actually **confusing** the user by having hidden semantics. (eg: say if we relied on cascading delete in the permissions model but don't yet have it Realm).
1. If we expose a complex model through an API we have to think up semantics for it in each language. It adds design, documentation and testing load.
1. An API is an invisible model with possible sampling through query functions.

I think it would help the debate to review a list of all the **forthcoming** command/control and status-query features yet to be designed (I assume we already _have_ a list) and quickly mark candidates to be exposed as Realms. It will be easier to learn if this approach is going to be used readily.

Realm/RLMSyncUser.h
cmelchior: Given that we are a offline-first database, it seems natural to me to have this behavior as a part of the convenience API, but instead of having a separate method for it. Would it make sense to overload the other methods with one that just accept the token and the callback?

Realm/RLMSyncUser.h
austinzheng: Actually, there is still a problem here.

No matter what, you need to pass back two values:
* Something representing the Offer itself, whether it be the offer object or a token.
* The `RLMNotificationToken`, which you hold on to in order for the callback to be called.

Any ideas other than passing back through a pointer parameter?

bigfish24: Do we want to use `RLMNotificationToken` or just have the `PermissionManager` concept as a singleton that holds the references?

bigfish24: Alternatively, why not just pass the object/token in the callback? This would mirror the `modify` API in the sense it isn't really designed for offline-use because while underneath it is creating a `PermissionChange` object, unless the callback fires, you would have no idea that object exists.

Realm/RLMSyncUser.h
```diff
@@ -113,8 +140,118 @@ NS_SWIFT_UNAVAILABLE("Use the full version of this API.");
...
+                        the Realm at `realmURL` should be modified, or
+                        `nil`.
+ @param permissions     The new permissions to be set.
+ @param callback        An optional block through which the progress of
+                        the permission change operation can be reported.
+ @returns   If `callback` was specified, a notification token. Hold onto
+            the token for as long as callback status updates are desired.
+            Call `-stop` on the token before destroying it.
+ */
+- (nullable RLMNotificationToken *)modifyRealmPermissionsForURL:(nullable NSURL *)realmURL
```
bigfish24: As per my comment below, it isn't clear to me why we are making the `offerPermissionForURL:` method "offline-first" by passing immediately an object/token/key and this method we are not?

Realm/RLMSyncUser.h
austinzheng: I was actually thinking that, instead of having the app developer hold on to the token, we just take care of holding on to it for him transparently, and we can stop it ourselves once there are logically no more progress to report (e.g. the operation succeeded). If we are going with the easy APIs versus advanced Realms dichotomy we might as well optimize the APIs for the common case as well as we can. Would that be a decent solution?

(I think the problem with passing in the callback is that we'd need to keep the token alive ourselves anyways until the first time the callback is called, and then if it was called later it would have a useless empty token argument.)

Realm/RLMSyncUser.h
```diff
@@ -32,9 +34,34 @@ typedef NS_ENUM(NSUInteger, RLMSyncUserState) {
...
+    RLMSyncRealmPermissionWrite,
+    /// The Realm can be managed by a target user.
+    RLMSyncRealmPermissionManage,
+};
+
 /// A block type used for APIs which asynchronously vend an `RLMSyncUser`.
 typedef void(^RLMUserCompletionBlock)(RLMSyncUser * _Nullable, NSError * _Nullable);
 
+/// A block type used to asynchronously report permission changes.
+typedef void(^RLMSyncPermissionChangeBlock)(RLMSyncManagementObjectStatus, NSError * _Nullable);
```
alazier: I think it will be cumbersome to use these callbacks. If I am making a permission change or offer, I want to get a response for my action specific to that action. With this design all responses are grouped together and dispatched though these two callbacks. This means that if a user wants to handle a specific change or offer they need to do their own complicated bookkeeping.

Realm/RLMSyncUser.h
```diff
@@ -113,8 +140,118 @@ NS_SWIFT_UNAVAILABLE("Use the full version of this API.");
...
+ @param callback        An optional block through which the progress of
+                        the permission change operation can be reported.
+ @returns   If `callback` was specified, a notification token. Hold onto
+            the token for as long as callback status updates are desired.
+            Call `-stop` on the token before destroying it.
+ */
+- (nullable RLMNotificationToken *)modifyRealmPermissionsForURL:(nullable NSURL *)realmURL
+                                                   userIdentity:(nullable NSString *)identity
+                                                    permissions:(RLMSyncRealmPermission)permissions
+                                                       callback:(nullable RLMSyncPermissionChangeBlock)callback
```
alazier: Disregard my last comment - this is what I was looking for.

Realm/RLMSyncUser.h
alazier: I don't think we should require a notification token. This is a case where the callback should get called without the need for a token.

Realm/RLMSyncUser.h
alazier: Get rid of RLMNotificationToken or call it something else, like request. If it were a request it could have a cancel method to allow cancelation but only if the user wants this. For cases where you don't need to cancel you can just not store the token/request.

Realm/RLMSyncUser.h
```diff
@@ -113,8 +140,118 @@ NS_SWIFT_UNAVAILABLE("Use the full version of this API.");
...
+NS_REFINED_FOR_SWIFT;
+
+/**
+ Create a permission offer.
+
+ A permission offer is repesented by a portable string token which can be
```
alazier: These concepts are themselves confusing. I have not spent much time with the permission system but why does this need to be so complex? Why can't I just give someone permission to something without them having to accept?

alazier:
So the apis I don't see which I would want is a general callback to monitor all permission changes, and an easy way to simply access/look at current permissions for a realm or all of those for a user.

alazier:
For the offer functionality why have static methods rather than just making an `Offer` class with member functions. Would simplify the top level api footprint considerably.

austinzheng:
My latest commit has the most minimal API possible. It's definitely preliminary pending feedback. Depending on what people think, I might incorporate @alazier's concept of a `Permission` object that is returned by the API so further actions on it are possible.

nirinchev:
I like the bare minimum proposal - one concern though is that we're losing the ability to set permissions in "additive" mode (by supplying `nil` for `canXXX`). Now we must supply the exact permission set we want the user to end up with.

austinzheng:
I think maybe there can be a corresponding "addPermissions" API that takes in the same bitmask, but only changes permissions (and requires an explicit user + Realm to avoid misuse).

Or we can have an enum (cases `Grant`, `Deny`, `LeaveUnchanged`), and then have three arguments in the method, one for each permission type (manage, read, write).

alazier:
@austinzheng that was the motivation behind the `setPermissions` method on the `Permissions` object. Could also make it a static method but seems nice to me as a member method if you are changing existing permissions. I would imagine making subsequent requests to create permissions for the same user/url combo would overwrite the previous value.

austinzheng:
Yeah. Maybe it should be a two-step process like you said: get a permissions object for a given Realm + user, and then read or set individual permissions on it.

I think setting permissions on all Realms and/or all users should still use the existing API, and should not be additive, given how dangerous it is if done by accident.

Thoughts?

alazier:
I don't think we should allow you to edit/create multiple permissions with the convenience api.

austinzheng:
Here's my strawman:

```objc
@implementation RLMSyncUser
// On RLMSyncUser. Get a permissions object representing the permissions for a given user to access a given Realm.
// That given Realm must be a Realm managed by this user, or else we return nil.
- (RLMSyncPermission *)permissionsObjectForRealmAtURL:(NSURL *)realmURL user:(NSString *)identity;
@end

@interface RLMSyncPermission : ???
@property (nonatomic, readonly) BOOL canRead;
@property (nonatomic, readonly) BOOL canWrite;
@property (nonatomic, readonly) BOOL canManage;

- (void)setReadPermission:(BOOL)canRead callback:(RLMSyncPermissionChangeBlock)callback;
- (void)setWritePermission:(BOOL)canRead callback:(RLMSyncPermissionChangeBlock)callback;
- (void)setManagePermission:(BOOL)canRead callback:(RLMSyncPermissionChangeBlock)callback;
// Or maybe just
// - (void)setPermissions:(RLMSyncPermissionOptions)permissions callback:(RLMSyncPermissionChangeBlock)callback;

- (void)addReadPermission:(BOOL)canRead callback:(RLMSyncPermissionChangeBlock)callback;
- (void)addWritePermission:(BOOL)canRead callback:(RLMSyncPermissionChangeBlock)callback;
- (void)addManagePermission:(BOOL)canRead callback:(RLMSyncPermissionChangeBlock)callback;
// Or maybe just
// - (void)addPermissions:(RLMSyncPermissionOptions)permissions callback:(RLMSyncPermissionChangeBlock)callback;
@end
```

nirinchev:
I can't say I like it :) Introducing some intermediate object to avoid using the models is no better than what we currently have.

The more I think about it, the more I believe a simple convenience api that wraps getting the management realm, adding the permission object, subscribing for notifications and then invoking a callback would be just fine. Something like (excuse my syntax mistakes):
```objc
RLMSyncPermissionChange *permissionChange = 
    [RLMSyncPermissionChange permissionChangeWithRealmURL:realmURL
                             userID:anotherUser
                             read:@YES
                             write:@YES
                             manage:@NO];

[user setPermissions: permissionChange
      callback: ^(NSError *error) {
}];
```

I know it feels like going full circle, but at the end of the day it feels as simple as the current approach. I know Obj C doesn't have default parameters, but languages that do can benefit from making the `mayXXX` arguments in the `PermissionChange` constructor optional to make it even simpler.

alazier:
Not sure why we need two methods, one to create a `PermissionChange` object and another to set it. It isn't clear from the naming if it is creating a permission, or editing an existing one. Also am not a fan of passing in three booleans to specify permissions (prefer the bitmask). Otherwise looks fine!

bigfish24:
This is interesting. Is the idea we would unify the representation of the permission state and making changes on it? Because today the permission state is represented in `Permission` objects in a user's `Permission.realm` whereas mutations occur in the `Management.realm`. The simplification of these together is appealing, since you don't ever have the state of another users permissions, just mutations and your own state (this likely needs different backend support but it is a simpler model). Note, we should still log the mutations in a way that is auditable since I can see value in that for customers (this is inherent with the `Permission.realm`).

Also I am assuming there would be an API to retrieve a `Results` collection of "my permissions" similar to the `Permission.realm` today, but that these are the same `SyncPermission` objects? 

Also, one thing to keep in mind is that we do want to support offline capabilities, could this mean that we have:
```swift
pendingChanges: Results<SyncPermission>
```
such that an app can retrieve a `SyncPermission` across app restarts from this collection and we would have a way to reattach a callback?

Overall, it took me a bit to wrap my head around this, but I like the simpler model of working with permission state objects that allow mutations instead of state and mutations as separate entities.

What's unclear is how this API utilizes the existing infrastructure underneath or requires new functionality (today how can you get the state for the permission of a User on a realm? That is represented in a `Permission` in the `Permission.realm` of a given use).

bigfish24:
If we are unifying the state and the mutations, note the state needs to a be live Realm object since it, at least for the user of the app, changes to the state are important for the app to act on (I.e. Remove a Slack channel from sidebar). For a server-side client you might want to be tracking changes for lots of users, so there would need to be a realm that is specific for each user that the global notifier can listen to and react to permission changes.

Given my comment above of needing a collection of "my permissions":
```swift 
`Sync.User.myPermissions: Results<SyncPermission>`
```

This would be backed by a Realm (do we unify `Management.realm` and `Permission.realm`?) which would allow the server side client to listen all of these users' Realms to react to permission state changes.

alazier:
In my proposal on slack I suggested exactly this - a method to get a Results object to vendor permissions objects. That and the ability to create/modify permissions through a `RLMPermission` object is all I think you need for most operations. That was the point of the proposal. To get notified of permission changes you can either register a notifier on the `Results` object or we can add a separate method for this.

bigfish24:
I like this, simplifies into 1 Object that includes state and provides mutation methods with callbacks. To me this is simpler and actually offers functionality we can't do today which is: seeing the state of another user's permission for a Realm you manage. I could see where an app would have a manager view allowing he/she can verify user X has write access and then choose revoke it. Today, the verification isn't possible.

Definitely do need to add the offline support. If we unify the mutation to just be 1 method:
`- (void)setPermissions:(RLMSyncPermissionOptions)permissions callback:(RLMSyncPermissionChangeBlock)callback;`
(Which I agree I like better than the others)

Then you could add:
`reattachCallback:(RLMSyncPermissionChangeBlock)callback;`

after retrieving the `SyncPermission` from the `pendingChanges` collection. The Object would be removed from `pendingChanges` once the mutation is accepted.

Also means a `cancelChanges()` on the `SyncPermission` makes sense.

nirinchev:
> Not sure why we would need two methods, one to create a PermissionChange object and another to set it. It isn't clear from the naming if it is creating a permission, or editing an existing one. Also am not a fan of passing in three booleans to specify permissions (prefer the bitmask). Otherwise looks fine!

- I went with the 3 nullable booleans, because I wanted to handle the case where we can leave certain fields unmodified. Alternatively we could have `setPermissions` and `addPermissions` methods or just another parameter `mode: Additive/Absolute` and go with the bitmask.
- I split it into two methods, because the number of arguments was rather large, but if we go with a bitmask, it's perfectly fine to combine in one method (yielding essentially what @austinzheng is currently proposing).
- I don't think people should care if it's creating a new permission or editing an existing one as long as it does what it's supposed to do :)

In my opinion, we should try to avoid making backend changes and work with the existing API instead. The reasons are that 1) we already shipped the existing API, so that should be supported, regardless of the changes we make and 2) it will require a lot more effort to change the backend in return for very little gain.

bigfish24:
>In my opinion, we should try to avoid making backend changes and work with the existing API instead. The reasons are that 1) we already shipped the existing API, so that should be supported, regardless of the changes we make and 2) it will require a lot more effort to change the backend in return for very little gain.

My guess on the implementation for @austinzheng proposal https://github.com/realm/realm-cocoa/pull/4619#issuecomment-280171907

Is that it would still use `PermissionChange` underneath the hood, but that:
```objc
- (RLMSyncPermission *)permissionsObjectForRealmAtURL:(NSURL *)realmURL user:(NSString *)identity;
```
will require a new API to retrieve the permission state for another user, but I think it is a good idea.

I am more in favor of Austin's approach for the reasons listed above. It simplifies the concept of permission state and mutations, while all my additional comments show how it can be used in an offline scenario which a basic wrapper on the models fails to do so.

nirinchev:
It can be achieved by using a combination of Permission and Management realm information, but it feels like a lot of effort and quite a bit of magic (e.g. calling `setXXX` will create a new PermissionChange object). Also, the `RLMSyncPermission` returned from `permissionsObjectForRealmAtURL` will not be an actual RealmObject, so keeping that up-to-date seems like an unpleasant exercise.

bigfish24:
>Also, the RLMSyncPermission returned from permissionsObjectForRealmAtURL will not be an actual RealmObject, so keeping that up-to-date seems like an unpleasant exercise.

I am not sure why you have to keep it up-to-date.... isn't it something you use situationally? Instead, the permission state we do need to keep up to date is the user's permissions:
`Sync.User.myPermissions: Results<SyncPermission>`

austinzheng:
I don't think we need any back-end changes, and I don't think other approaches are any better. Let me try to explain more thoroughly.

`Permission` represents the set of permissions a single user has for a single Realm. The user and Realm involved are immutable. Only a user with permission to manage a Realm can vend out `Permission` objects that involve that Realm.

Because `Permission` represents a single set of permissions for a given user-Realm 2-ple, it becomes the way the API provides to _set_ those permissions and to _get_ the values of those permissions.

The thing is, it doesn't store any state itself. It sets permissions by creating `PermissionChange` objects under the hood, registering callbacks on them, and storing the tokens until the change completes or fails. It gets permissions by querying the permissions Realm for the right information.

I think it's quite a bit more intuitive to have the concept of a `Permission` object (permissions are nouns) rather than to expect users to work directly with objects representing permission changes (a noun standing in for a verb; a gerund I guess?). But the object that represents a permission can implement itself using our existing machinery; I don't see why backend changes are necessary.

Does this explanation help at all?

bigfish24:
@austinzheng how do you get the permission state for another user? Right now a user has his/her own permission realm for state and management realm for mutations. I would imagine the API to get the permission state for another user would need to be a new API that retrieves this from the admin realm. Also it seems unlikely that another user's state could be a Realm object that live updates since we already have some performance issues with keeping the user's permission state in the permission realm up to date, not to mention, I am not sure how that would work, where does this live object live?

From an API perspective I like it but after talking with @nirinchev last week I think we should be more pragmatic and just stick to the wrapper APIs on top of the current system, and focus on shipping since the permission realm is still not shipped for example.

austinzheng:
Alright, let's go with the simple API in this PR (not the one I sketched out in my comment above), with possibly an additional simple API to additively set permissions. Is that okay with everyone? @bigfish24 @nirinchev @alazier

austinzheng:
I added a second API that allows permissions to be set additively.

nirinchev:
Looks good to me. Just some clarification - why are the parameters (user id/realm url) nullable in `setPermissions` and non-nullable in `addPermissions`?

austinzheng:
It's to discourage people from making additive changes to anything except a single user/Realm combo. Should those types be nullable so that the two APIs work exactly the same way?

bigfish24:
I would recommend a single API, not sure I understand the value of set and add. Seems more confusing on face value.

austinzheng:
How should this API distinguish between setting and adding permissions? (nullable `NSNumber` arguments are not a good way to do this)

nirinchev:
I don't have a strong opinion on nullability - just feels a bit awkward that they don't match. @bigfish24 the second method was added to allow scenarios where you want to grant read permissions without checking if the user already has write/manage permissions.

bigfish24:
I thought a set with just read and nullable for write and manage is additive? Sorry still confused on why distinction matters.

nirinchev:
With the new API, there's no longer a triple `bool? mayRead, bool? mayWrite, bool? mayManage` and instead there's a flag enum [`RLMSyncRealmPermission`](https://github.com/realm/realm-cocoa/pull/4619/files#diff-37be3042befd72f8aea6cca929dbac8cR41) that you pass as a single argument. Therefore, you can't really make a distinction between additive and absolute permission change.

bigfish24:
Ah, I see, sorry for mix up. I understand and seems reasonable unless others feel strongly for the `additive` parameter to be exposed unifying into 1 API.

austinzheng:
I changed this some more based on @bigfish24 and @alazier feedback. If nobody objects within the next two hours I'm going to build out the rest of this API and tests.

austinzheng:
Superseded by #4799.

Issue #4620, opened 2017-02-01
specify both Xcode and Swift version on CI


jpsim:
As tracked in [rdar://30215450](http://www.openradar.me/30215450), Xcode 8.3 beta 1 is reporting its Swift version as "3.0", not "3.1" as it should.

This means that rather than running on Xcode 8.0, Swift 3.0 CI jobs are actually being run on Xcode 8.3 beta 1.

More generically, multiple Xcode versions may support compiling the same Swift version, such as when Xcode 9 comes out and has a Swift 3 compilation mode, where we'll want to specify an Xcode version _and_ a Swift version when running jobs.

PR #4621, opened 2017-02-01
specify Xcode version in Jenkins jobs

jpsim:
Addresses #4620

### DIFF
--- a/scripts/swift-version.sh
+++ b/scripts/swift-version.sh
@@ -3,6 +3,42 @@ get_swift_version() {
     $swift_command --version 2>/dev/null | sed -ne 's/^Apple Swift version \([^\b ]*\).*/\1/p'
 }
 
+get_xcode_version() {
+    xcodebuild_command=$@
+    $xcodebuild_command -version 2>/dev/null | sed -ne 's/^Xcode \([^\b ]*\).*/\1/p'
+}
+
+find_xcode_with_version() {
+    # First check if the currently active one is fine
+    version="$(get_xcode_version xcodebuild)"
+    if [[ "$version" = "$1" ]]; then
+        export DEVELOPER_DIR="$(xcode-select -p)"
+        export REALM_SWIFT_VERSION=$(get_swift_version xcrun swift)
+        return 0
+    fi
+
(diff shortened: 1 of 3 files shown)


jpsim:
This will need to be amended once we want to resume running jobs with non-default Swift toolchains again (such as when Xcode 9 comes out with a Swift 3 compilation mode) but for now this should be sufficient.

tgoyne:
Do the release scripts need to be updated as well so that we don't accidentally cut a release using the 8.3 beta?

jpsim:
You're right @tgoyne, done in ee64c1f.

scripts/swift-version.sh
```diff
@@ -3,6 +3,42 @@ get_swift_version() {
...
 }
 
+get_xcode_version() {
+    xcodebuild_command=$@
+    $xcodebuild_command -version 2>/dev/null | sed -ne 's/^Xcode \([^\b ]*\).*/\1/p'
+}
+
+find_xcode_with_version() {
+    # First check if the currently active one is fine
+    version="$(get_xcode_version xcodebuild)"
```
larkost: The quotes here are superfluous, but not harmful.

scripts/swift-version.sh
```diff
@@ -3,6 +3,42 @@ get_swift_version() {
...
+    if [[ "$version" = "$1" ]]; then
+        export DEVELOPER_DIR="$(xcode-select -p)"
+        export REALM_SWIFT_VERSION=$(get_swift_version xcrun swift)
+        return 0
+    fi
+
+    local xcodes dev_dir version
```
larkost: This should probably move up before line 13.

scripts/swift-version.sh
```diff
@@ -34,7 +70,9 @@ find_xcode_for_swift() {
     return 1
 }
 
-if [[ "$REALM_SWIFT_VERSION" ]]; then
+if [[ "$REALM_XCODE_VERSION" ]]; then
```
larkost: I would generally use `-n` to be certain, but I think that this will work in any event.

PR #4622, opened 2017-02-01
test only ios-device

larkost:
Test branch for signing problem


larkost:
closing this as the test is complete

Issue #4628, opened 2017-02-03
Prevent throwing RLMException "Object has been deleted or invalidated". Return nil instead.


rokgregoric:
## Goals

The problem I am facing happens when a deleted object is accessed by the UI (fetched form the DB and referenced before being deleted).
Realm throws an exception (RLMException - "Object has been deleted or invalidated").

Checking the object if `isInvalidated` or using unmanaged (detached) objects - `Object(value: ferchedObject)` is an option, but makes the code super ugly with tons of checks all over the place and very hard to manage.

When using unmanaged objects all references (links) to other objects have to be checked if `isInvalidated` otherwise the exception is thrown again.

So instead of throwing an exception I would like realm to just `return nil` as if the object or property was released.

## Expected Results

I would like realm to just `return nil` as if the object or property was released.

## Actual Results

Realm throws an exception (RLMException - "Object has been deleted or invalidated").

## Steps to Reproduce

That's very simple to reproduce:
* just create an object
* keep a reference to it
* delete the object
* try to access the deleted object

## Version of Realm and Tooling

Realm version: 2.4.2

Xcode version: 8.2.1

iOS/OSX version: iOS 10.2

Dependency manager + version: Cocoapods 1.2.0

istx25:
Thanks for filing this, @rokgregoric. Someone will take a look at your PR soon. :)

jpsim:
There's a significant difference between returning `nil` when accessing a valid object's property and accessing a deleted object. This type of ambiguous behavior, such as messaging `nil` in the Objective-C runtime, has [historically shown](http://blog.ioactive.com/2012/11/ios-security-objective-c-and-nil.html) to be [a source of bugs](http://vgable.com/blog/2008/05/31/messages-to-nowhere/).

That's why Realm distinguishes between both cases. For example:

```swift
class MyModel: Object {
    dynamic var property: String? = ""
}
let realm = try Realm()
realm.beginWrite()
let object = MyModel()
object.property = nil
_ = object.property // nil means one thing here

realm.cancelWrite()
```

We define any access to a deleted object as a usage error (by throwing that exception). Thankfully, Realm objects can only be invalidated at very specific points in time, so you shouldn't have to write as defensively as you would to avoid "faults" present in other frameworks. For example, when a Realm is refreshed, or when a deletion has happened earlier in that same thread.

[...]

PR #4629, opened 2017-02-03
Prevent throwing RLMException "Object has been deleted or invalidated". Return nil instead.

rokgregoric:
## Goals

The problem I am facing happens when a deleted object is accessed by the UI (fetched form the DB and referenced before being deleted).
Realm throws an exception (RLMException - "Object has been deleted or invalidated").

Checking the object if `isInvalidated` or using unmanaged (detached) objects - `Object(value: ferchedObject)` is an option, but makes the code super ugly with tons of checks all over the place and very hard to manage.

When using unmanaged objects all references (links) to other objects have to be checked if `isInvalidated` otherwise the exception is thrown again.

So instead of throwing an exception I would like realm to just `return nil` as if the object or property was released.

## Expected Results

I would like realm to just `return nil` as if the object or property was released.

## Actual Results

Realm throws an exception (RLMException - "Object has been deleted or invalidated").

## Steps to Reproduce

That's very simple to reproduce:
* just create an object
* keep a reference to it
* delete the object
* try to access the deleted object

## Related Issues

Addresses https://github.com/realm/realm-cocoa/issues/4628.

### DIFF
--- a/Realm/RLMAccessor.mm
+++ b/Realm/RLMAccessor.mm
@@ -60,12 +60,14 @@ typedef NS_ENUM(char, RLMAccessorCode) {
 
 template<typename T>
 static T get(__unsafe_unretained RLMObjectBase *const obj, NSUInteger index) {
+    if (!obj->_row.is_attached()) { return T(); }
     RLMVerifyAttached(obj);
     return obj->_row.get_table()->get<T>(obj->_info->objectSchema->persisted_properties[index].table_column, obj->_row.get_index());
 }
 
 template<typename T>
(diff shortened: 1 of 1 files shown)


rokgregoric:
Maybe it could be disabled with an option if you are building a framework that really need fine grained control over invalidated accesses.

jpsim:
Thanks for taking the time to make a PR and contribute to Realm @rokgregoric! However, for the reasons I've shared in https://github.com/realm/realm-cocoa/issues/4628#issuecomment-279767278, we can't accept this change. We hope you'll contribute again in the future.

PR #4632, opened 2017-02-03
Fix issue where non-RLMObject could be passed to `RLMArray` method

### LINKED ISSUE #4335: Unable to use setValue for List
## Goals

Able to use `setValue` for List Types

## Expected Results

It should work

## Actual Results

Get Object Type errors

## Steps to Reproduce

run the sample code

## Code Sample

```swift
class BaseEntity: Object {
    dynamic var simpleProp: String?
    var listProp = List<ListEntity>()
}

class ListEntity: Object {
    dynamic var simpleProp: String?
}

let json: [String: Any] = ["simpleProp": "Wayne", "listProp": [["simpleProp": "Bruce"]]]

// this works
let entity1 = BaseEntity(value: json)
print(entity1)

// this does not
let entity = BaseEntity()

json.keys.forEach { key in
    entity.setValue(json[key], forKey: key)
}

print(entity)
```

```
ProductName:	Mac OS X
ProductVersion:	10.12.2
BuildVersion:	16C41b

Xcode 8.1
Build version 8B62

/usr/local/var/rbenv/shims/pod
1.1.1
Realm (2.0.3)
RealmSwift (2.0.3)

/usr/local/bin/bash
GNU bash, version 4.4.0(1)-release (x86_64-apple-darwin16.0.0)

/usr/local/bin/carthage
0.18.1
(not in use here)

git: aliased to hub
git version 2.10.2
hub version 2.2.9
```

austinzheng:
Fixes #4335 

In some cases a non-`RLMObject` object could be passed down into methods expecting a `RLMObject`, which causes issues when ivars on that object are accessed. This requires an object to be validated before it can be passed further down the chain.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,9 @@ x.x.x Release notes (yyyy-MM-dd)
 
 * Fix synchronized Realms not downloading remote changes when an access token
   expires and there are no local changes to upload.
+* Fix an issue where values set on a Realm object using `setValue(value:, forKey:)`
+  that were not themselves Realm objects were not properly converted into Realm
+  objects or checked for validity.
 
 2.4.2 Release notes (2017-01-30)
 =============================================================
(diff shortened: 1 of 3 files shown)


tgoyne:
Need a changelog entry.

PR #4633, opened 2017-02-03
Add validation for Swift property names

tgoyne:
Swift properties which fall in to any of the ARC method families don't work properly via KVC, so we can't support them and will either leak memory or crash when trying to use them (currently we crash).

Trying to use properties with these names is a compilation error in Objective-C, so it's not an issue there.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,8 @@ x.x.x Release notes (yyyy-MM-dd)
 * Fix incorrect sync progress notification values for Realms originally created
   using a version of Realm prior to 2.3.0.
 * Fix LLDB integration to be able to display summaries of `RLMResults` once more.
+* Reject Swift properties with names which cause them to fall in to ARC method
+  families rather than crashing when they are accessed.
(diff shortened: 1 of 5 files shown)


czjeep:
dd

czjeep:
this happen to me when i update realmswift from 2.0.* to 3.0.0. I rename my model's property name and migration schemaVersion from 21 to 22, and it solved. but i have a question. eg: i rename a property from "newRegistWithFCode" to "newregistWithFCode",and the migration is: migration.renameProperty(onType: UserInfo.className(), from: "newRegistWithFCode", to: "newRegistWithFCode"). everything is ok. but focus on the two 'to name', R and r is different, and why this could be ok?

czjeep:
i think , in the migration, the toname should only be "newregistWithFCode" ,not the "newRegistWithFCode"

austinzheng:
Please create a new ticket and fill out the Issue Template, rather than commenting on closed issues or merged PRs. Thanks.

Issue #4586, opened 2017-01-25
Support Xcode 8.3 beta 1


TimOliver:
iOS 10.3 beta 1 was released earlier today, along with the first beta of Xcode 8.3. This version of Xcode introduced Swift 3.1, for which Realm Swift will need to be updated.

## Goals

To be able to use Realm Swift in Xcode 8.3 without any issues.

## Expected Results

To be able to succesffuly build a project with Realm Swift.

## Actual Results

When building the Realm Swift build target itself in Xcode 8.3 beta 1:

<img width="958" alt="screen_shot_2017-01-25_at_8 23 20_pm" src="https://cloud.githubusercontent.com/assets/429119/22294044/c484d58c-e34c-11e6-909e-cb8c523b5740.png">
(Attempting to do the same in Xcode 8.2 builds successfully with no issues)

When bringing Realm Swift in as a CocoaPods dependency for an iOS app:

![screen shot 2017-01-25 at 10 19 00 pm](https://cloud.githubusercontent.com/assets/429119/22294087/f0fed202-e34c-11e6-8367-9d1b75b6672e.png)

(Again, opening the same project in Xcode 8.2 will build and deploy with no issues).

## Steps to Reproduce

1. [Download Xcode 8.3 beta 1](https://developer.apple.com/download/)
2. Open Realm.xcodeproj, select the Realm Swift build target, and hit 'Build'

## Version of Realm and Tooling

Realm version: 2.3.0

Xcode version: 8.3 beta 1

iOS/OSX version: 10.3

Dependency manager + version: CocoaPods 1.1.1

jpsim:
Note that it's not just Realm Swift that's affected, but also Realm Objective-C (the first error you report).

jpsim:
Filed a radar: http://www.openradar.me/30194818

But I'm continuing to investigate.

jpsim:
Swift 3.1 is supposed to be source-compatible with 3.0.x, so I'm:

1) finding workarounds so our users can build with 8.3 despite these compiler regressions
2) checking if apple/swift#7017 helps
3) filing radars/SRs if it doesn't

jpsim:
No luck. Filed [rdar://30228957](http://www.openradar.me/30228957).

istx25:
Would we consider this blocked?

woodland021:
same issue :(

RustyKnight:
Should be filing our radars (basically copying yours) to get this more notice from Apple?

[...]

PR #4640, opened 2017-02-06
Fix compilation issues with Xcode 8.3 beta 2

jpsim:
Addresses #4586. Doesn't look like apple/swift#7136 made it into Xcode 8.3 beta 2, but at least [rdar://30215450](http://www.openradar.me/30215450) was resolved so we can conditionally fix Swift 3.1.

Thomas: do you have any ideas for a more correct fix for db13fba?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ x.x.x Release notes (yyyy-MM-dd)
 * Fix an issue where values set on a Realm object using `setValue(value:, forKey:)`
   that were not themselves Realm objects were not properly converted into Realm
   objects or checked for validity.
+* Fix compilation issues with Xcode 8.3 beta 2.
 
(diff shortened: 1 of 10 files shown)


tgoyne:
Running the tests with asan hit use-after-frees before even getting to testDeleteResults, so I assume more things are broken and the change in https://github.com/realm/realm-cocoa/commit/db13fba1baadd8e5ea98bd97a21c36d6e21164f4 is only making it not crash by coincidence.

tgoyne:
The use-after-free is unrelated to the crash in `testDeleteResults`.

Changing `RLMTrackDeletions(_realm, ^{ _results.clear(); })` to `RLMTrackDeletions(_realm, [&] { _results.clear(); })` also eliminates the crash. Might be a clang bug?

jpsim:
Thanks for filing that radar!

PR #4645, opened 2017-02-08
Fix some logic and test bugs

austinzheng:
Changes:
- Fixed an issue where a non-nil `RLMSyncSession` object with an nil underlying pointer could be returned when looking up a session for an invalid synced Realm URL
- Wrote two object server tests to verify fix
- Changed the way upload/download object server tests work to prevent them from sporadically failing

The upload/download object server tests sometimes fail because the upload or download completion callback occurs before the notification that indicates completion can be delivered. (Upon completion of the callback, the token is immediately stopped, and no further notifications will be delivered.) The fix in this PR prevents that spurious failure from occurring.


austinzheng:
The seven test failures all seem to have to do with the simulator not booting up after 35 minutes. This is a known issue with some of the CI machines.

PR #4651, opened 2017-02-09
update core to 2.3.1 and sync to 1.0.4

jpsim:
Core changelog here: https://github.com/realm/realm-core/blame/v2.3.1/CHANGELOG.md#L1-L31

I don't think anything else from core applies to our end-users, nor does anything from realm-sync: https://github.com/realm/realm-sync/compare/v1.0.3...v1.0.4


tgoyne:
The async commit daemon change cuts the number of files we open per thread in half on macOS (but not any other cocoa platforms).

jpsim:
Filed rdar://30446115 about clang crash from the [ios-static debug Xcode 8.2 job](https://ci.realm.io/job/objc_pr/4745/configuration=Debug,target=ios-static,xcode_version=8.2/console).

jpsim:
I couldn't reproduce the clang crash locally. The other CI failures are unrelated.

jpsim:
Response from Apple on the radar I had filed:

> We can’t reproduce the reported issue and the provided crashlogs look fairly random. We would advise testing your machine for faulty memory, the compiler is often one of the first victims of bad RAM.

PR #4652, opened 2017-02-09
find Xcode preferentially without spotlight (v2)

larkost:


larkost:
Note that this solves #4503 and #4534.

build.sh
```diff
@@ -1134,8 +1139,11 @@ EOM
     "package-ios-swift")
         cd tightdb_objc
         for version in 8.0 8.1 8.2; do
-            REALM_XCODE_VERSION="$version" sh build.sh prelaunch-simulator
-            REALM_XCODE_VERSION="$version" sh build.sh ios-swift
+            REALM_XCODE_VERSION=$version
+            REALM_SWIFT_VERSION=
+            set_xcode_and_swift_versions
```
jpsim: `build.sh` now calls `set_xcode_and_swift_versions` unconditionally, so it's superfluous to call it in all these `package-` commands. Also, `REALM_SWIFT_VERSION=` could be moved outside these for loops.

```diff
-for version in 8.0 8.1 8.2; do
-REALM_XCODE_VERSION=$version
-REALM_SWIFT_VERSION=
-set_xcode_and_swift_versions
+REALM_SWIFT_VERSION=
+for REALM_XCODE_VERSION in 8.0 8.1 8.2; do
```

larkost: Both are actually needed, since `set_xcode_and_swift_versions` is the only one `export`ing it (so it can go to a child process), and it is also always setting `REALM_SWIFT_VERSION` so without that being inside the loop you would get it set to the value for the first one.

jpsim: You should still be able to do this, cutting down on some boilerplate:

```diff
-for version in 8.0 8.1 8.2; do
-  REALM_XCODE_VERSION=$version
-  REALM_SWIFT_VERSION=
-  set_xcode_and_swift_versions
+REALM_SWIFT_VERSION=
+for REALM_XCODE_VERSION in 8.0 8.1 8.2; do
+  set_xcode_and_swift_versions
```

larkost: No, doing that would fail with a complaint that later versions of Xcode do not support the selected (one for Xcode 8.0) version of Swift.

jpsim:
This has been causing the `cocoa_docs` PR to fail in the `cocoa-pipeline` Jenkins job since it was merged. https://ci.realm.io/job/cocoa_docs/

@larkost could you please fix?

PR #4662, opened 2017-02-13
update sync client to 1.1.0

jpsim:


jpsim:
Fixes realm/realm-mobile-platform#119.

jpsim:
Lots of unrelated CI issues. See realm/ci#85

Issue #4626, opened 2017-02-02
Rework how sync progress notifications are dispatched


austinzheng:
I talked to @bdash about this on Slack, but wanted to solicit more opinions, especially from @tgoyne and @jpsim.

Right now, we (attempt to) dispatch our sync progress notifications onto the same thread upon which they were registered. We do this to conform to the behavior of our Realm, collection, and object notifications. This works great if the user registers their sync progress notifications on the main thread, and falls down if they try to do it on any other thread.

We wanted to revisit this design decision so we can fix the feature and its semantics. First, I think it might be okay for these types of notifications to be dispatched to a thread other than the one the notification was registered on. Our other notifications all deal specifically with changes to thread-confined objects, which is why it's important to user experience that they run on the same thread they were registered on. However, sync progress notifications don't deal directly with any Realm objects. The sync objects they deal with are all thread safe. More importantly, the biggest use case of a sync progress notification block is to update UI.

With this being said, there are at least three options we can take to fix how this feature works:

1. Dispatch notification blocks to the main queue. The advantage is that users can call UIKit methods directly from their notification blocks. The disadvantage is that, given how frequently the notifications are fired, we risk degrading app performance by flooding the main queue.

2. Dispatch notification blocks to a serial queue we create for this purpose. The advantage is that we don't flood the main queue, and we can configure the queue for our needs. The disadvantage is that we need to call out in our documentation that the app is responsible for throttling and dispatching to the main thread to modify UI.

3. Implement object store throttling, and dispatch notification blocks to the main queue. Object store throttling, in the absence of timers, would work by using the C++ time APIs to avoid forwarding any notifications to the binding during a cooldown period. The biggest disadvantage of this method is that the bindings might be stuck with stale date (e.g. if the last notifications come in during cooldown, but after cooldown ends network traffic stops and no more notifications come in).

austinzheng:
Closed because of #4663.

PR #4663, opened 2017-02-13
Change how progress notifications are dispatched

### LINKED ISSUE #4623: SyncSession.addProgressNotification does not work
## Goals

Get notifications for sync upload and download progress.

## Expected Results

Get sync progress notifications.

## Actual Results

Sync progress notifications are never delivered.

## Steps to Reproduce

## Code Sample

```
import UIKit
import RealmSwift

class O: Object {

    dynamic var data = Data()

}

@UIApplicationMain
class AppDelegate: UIResponder, UIApplicationDelegate {

    var window: UIWindow?
    var token: NotificationToken?
    
    var realm: Realm!
    var downloadToken: SyncSession.ProgressNotificationToken?
    var uploadToken: SyncSession.ProgressNotificationToken?
    
    func application(_ application: UIApplication, didFinishLaunchingWithOptions launchOptions: [UIApplicationLaunchOptionsKey: Any]?) -> Bool {
        
        let serverURL = URL(string: "http://localhost:9080")!
        
        SyncUser.logIn(with: .usernamePassword(username: "username", password: "password"), server: serverURL) {
            assert($1 == nil)
            self.loginCompleted(user: $0!)
        }
        
        return true
    }
    
    func loginCompleted(user: SyncUser) {
        let realmURL = URL(string: "realm://localhost:9080/~/10k_data_multiple_writes")!
        var configuration = Realm.Configuration.defaultConfiguration
        configuration.syncConfiguration = SyncConfiguration(user: user, realmURL: realmURL)
        realm = try! Realm(configuration: configuration)
        
        let session = user.session(for: realmURL)!
        downloadToken = session.addProgressNotification(for: .download, mode: .reportIndefinitely) {
            print("download progress: \($0)") // This is never called.
        }
        uploadToken = session.addProgressNotification(for: .upload, mode: .reportIndefinitely) {
            print("upload progress: \($0)") // This is never called.
        }
        
        for _ in 0..<10_000 {
            try! realm.write {
                let o = O()
                o.data = "this is a test with some data...".data(using: .utf8)!
                realm.add(o)
            }
        }
    }
    
}
```

Realm version: 2.4.2 / Realm Mobile Platform 1.0.0-2

Xcode version: 8.2.1

austinzheng:
Based on one possible interpretation of https://github.com/realm/realm-cocoa/issues/4626. See that ticket for discussion and rationale.

Fixes #4623.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,9 @@ x.x.x Release notes (yyyy-MM-dd)
   write does not actually change existing values.
 * Improve performance of deleting all objects in an RLMResults.
 * Reduce the number of files opened per thread-specific Realm on macOS.
+* Sync progress notifications are now dispatched to a side queue, and a
+  convenience API has been added to allow notifications to be directly dispatched
+  to the main thread instead.
 
 ### Bugfixes
 
@@ -23,6 +26,8 @@ x.x.x Release notes (yyyy-MM-dd)
   non-nil value when passed in an invalid URL.
 * `SyncSession.Progress.fractionTransferred` now returns 1 if there are no
   transferrable bytes.
+* Change the way sync progress notifications are dispatched to fix broken
+  behavior when they aren't registered from the main thread.
 * Fix compilation issues with Xcode 8.3 beta 2.
 * Fix incorrect sync progress notification values for Realms originally created
   using a version of Realm prior to 2.3.0.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,9 +11,6 @@ x.x.x Release notes (yyyy-MM-dd)
   write does not actually change existing values.
 * Improve performance of deleting all objects in an RLMResults.
 * Reduce the number of files opened per thread-specific Realm on macOS.
-* Sync progress notifications are now dispatched to a side queue, and a
-  convenience API has been added to allow notifications to be directly dispatched
-  to the main thread instead.
 
 ### Bugfixes
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,8 +23,8 @@ x.x.x Release notes (yyyy-MM-dd)
   non-nil value when passed in an invalid URL.
 * `SyncSession.Progress.fractionTransferred` now returns 1 if there are no
   transferrable bytes.
-* Change the way sync progress notifications are dispatched to fix broken
-  behavior when they aren't registered from the main thread.
+* Fix sync progress notifications registered on background threads by always
+  dispatching on a dedicated background queue.
 * Fix compilation issues with Xcode 8.3 beta 2.
 * Fix incorrect sync progress notification values for Realms originally created
   using a version of Realm prior to 2.3.0.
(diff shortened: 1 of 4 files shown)


austinzheng:
Adding reviewers to open up discussion. (We might not want to go with this approach.)

Realm/RLMSyncSession.h
```diff
@@ -150,8 +152,20 @@ NS_ASSUME_NONNULL_BEGIN
  */
 - (nullable RLMProgressNotificationToken *)addProgressNotificationForDirection:(RLMSyncProgressDirection)direction
                                                                           mode:(RLMSyncProgress)mode
+                                                           dispatchToMainQueue:(BOOL)dispatchToMainQueue
```
bdash: This is really gross.

austinzheng: Would it be preferable to just have the user explicitly specify a queue?

bdash: What do we gain from the user explicitly specifying a queue? Why is that preferable to having them dispatch to whichever queue they care about from within the callback?

austinzheng: If we're okay with telling users they must explicitly dispatch onto whatever queue they want, sure. (We might also not want users to 'accidentally' run expensive work in their callbacks without dispatching, since IIRC the callback is run on the sync client's thread.)

bdash: But after the rest of this PR it's _not_ run on the sync client thread, it's run on a separate notification queue. The only thing they'll delay by doing expensive work is further progress callbacks.

austinzheng: Oh, I see what you mean now. In that case, I agree. We specify that there's a notification queue, and if they want the work to be done elsewhere they can dispatch it to whatever queue they want. Is that satisfactory?

bdash: That would be my inclination.

CHANGELOG.md
```diff
@@ -23,6 +23,8 @@ x.x.x Release notes (yyyy-MM-dd)
   non-nil value when passed in an invalid URL.
 * `SyncSession.Progress.fractionTransferred` now returns 1 if there are no
   transferrable bytes.
+* Change the way sync progress notifications are dispatched to fix broken
```
jpsim: This doesn't announce that they're no longer dispatched on the same thread on which they were registered, which is technically a breaking change, but we've always allowed ourselves to make breaking changes to already broken implementations, so that's fine.

However, I'd reword this:

> * Fix sync progress notifications registered on background threads by always dispatching on a dedicated background queue.

austinzheng: Yeah, good call. There was a bit more about this in the "Enhancements" section originally, but I deleted that part. Thanks for pointing out the omission.

> but we've always allowed ourselves to make breaking changes to already broken implementations

That was my rationale as well.

jpsim:
Fixes #4623.

austinzheng:
One of the builds failed because the compiler segfaulted, the other three failed because of various CI issues.

PR #4669, opened 2017-02-14
insert missing set_xcode_and_swift_versions

larkost:


larkost:
The two failures look unrelated (test failures), so I think this one can go in.

PR #4675, opened 2017-02-17
Fix sorting by key path when the property index doesn't match the column index

### LINKED ISSUE #4592: Object type '(null)' is not managed by the Realm when using sorting on OneToOne child property
## Goals

Sorting using Results.sorted() on a child property of the current Realm Object. (eg. "shop.name") 
Shop is a Realm object and name a String

## Expected Results

It works on a clean install of our app; eg. empty Realm db file. It doesn't work if we're upgrading from our production App (using existing Realm db and migration script)

What did you expected to happen?

It does work on clean install but also when upgrading from an older Realm version and Realm Migration script.

## Actual Results

We receive the following crash when we upgrade our app:

Terminating app due to uncaught exception 'RLMException', reason: 'Object type '(null)' is not managed by the Realm. If using a custom `objectClasses` / `objectTypes` array in your configuration, add `(null)` to the list of `objectClasses` / `objectTypes`.'

translateErrors is called in the following code:

```
- (RLMResults *)sortedResultsUsingDescriptors:(NSArray<RLMSortDescriptor *> *)properties {
    if (properties.count == 0) {
        return self;
    }
    return translateErrors([&] {
        if (_results.get_mode() == Results::Mode::Empty) {
            return self;
        }

        return [RLMResults resultsWithObjectInfo:*_info results:_results.sort(RLMSortDescriptorFromDescriptors(*_info, properties))];
    });
}
```

## Steps to Reproduce

Realm 2.2.0 and 2.3.0 - Upgrading from 2.1

Xcode version: 8

iOS/OSX version: 8.4

Dependency manager + version: Cocoapods 1.1.1

bdash:
Fixes #4592.
- [x] Bug fix
- [x] Tests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,6 +31,9 @@ x.x.x Release notes (yyyy-MM-dd)
 * Fix LLDB integration to be able to display summaries of `RLMResults` once more.
 * Reject Swift properties with names which cause them to fall in to ARC method
   families rather than crashing when they are accessed.
+* Fix sorting by key path when the declared property order doesn't match the order
+  of properties in the Realm file, which can happen when properties are added in
+  different schema versions.
 
 2.4.2 Release notes (2017-01-30)
 =============================================================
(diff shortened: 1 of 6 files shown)


bdash:
Tests added, so this is now ready for review.

PR #4679, opened 2017-02-20
Fix initialization of dictionary

austinzheng:
Sorry for not catching this earlier.


austinzheng:
There was a piece of code keeping the session alive for longer than it should have been, so I fixed that. I verified the test I originally wrote to check this case failed after I made that fix and removed the dictionary initialization, and passed once I added the dictionary initialization back in.

austinzheng:
The one failing CI test is just more Jenkins sadness.

PR #4681, opened 2017-02-21
Add download Realm functionality

austinzheng:
This PR does a few things:

- It adds properties on `RLMSyncConfiguration` to determine whether the synced Realm should download all outstanding changesets at time of opening when opened (`waitForServerChanges`), ~and if so how long it should wait (if desired)~.
- ~When opening a Realm with the `waitForServerChanges` flag set on its configuration, the Realm initializer call will block until the Realm has downloaded, or throw an exception if opened on the main thread.~
- It adds an API on `RLMRealm` to asynchronously open a Realm and pass ~its configuration~ it into a callback block once any work required to open it (see point 2 above) has completed.

It needs:

- [x] Prerequisite: https://github.com/realm/realm-object-store/pull/385 merged
- [ ] TODOs fixed
- [ ] Design review, which might result in this design discarded in favor of a different one
- [ ] Tests
- [ ] Swift implementation


austinzheng:
I pushed a revised version of this API + code, based on discussions I had with @bigfish24 this morning. Comments would be welcome. (In particular, there is no more blocking anywhere in the design anymore.)

austinzheng:
Any further thoughts on this PR before I move forward with it? @bigfish24

jpsim:
Continued in #4859. Thanks @austinzheng!

PR #4698, opened 2017-02-24
Add aggregate functions to RLMArray and RLMCollection

tgoyne:
List and RealmCollection had these, but in obj-c only RLMResults did, which was oddly inconsistent.

Also add support for them on unmanaged arrays, which previously did not work for List.


RealmSwift/Tests/RealmCollectionTypeTests.swift
```diff
@@ -904,90 +904,6 @@ class ListStandaloneRealmCollectionTypeTests: ListRealmCollectionTypeTests {
         assertThrows(collection.filter("ANY stringListCol == %@", CTTStringObjectWithLink()))
     }
 
-    override func testMin() {
```
jpsim: These tests were removed because this is now tested in Objective-C? I still think they're valuable here IMO.

tgoyne: No, they're no longer being overridden from the base class because aggregates are now supported for unmanaged objects.

jpsim: Ah, I see. All good!

CHANGELOG.md
```diff
@@ -10,6 +10,8 @@ x.x.x Release notes (yyyy-MM-dd)
 * Add a `{RLM}SyncUser.isAdmin` property indicating whether a user is a Realm
   Object Server administrator.
 * Speed up case-insensitive queries on indexed string properties.
+* Add RLMResults's collection aggregate methods to RLMArray.
```
jpsim: > RLMResults'

tgoyne: RLMResults is used as a proper noun that happens to end in s and not as a plural, so 's is correct.

jpsim: Yay English.

PR #4699, opened 2017-02-25
Add `isAdmin` flag to `RLMSyncUser`

austinzheng:
Changes:
- Sync users now expose a property indicating whether or not they are Realm Object Server administrator users

Requires: https://github.com/realm/realm-object-store/pull/396

To do:
- [x] Verify it works with a version of ROS with the corresponding server change
- [x] Write test(s)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,7 +12,7 @@ x.x.x Release notes (yyyy-MM-dd)
 
 ### Bugfixes
 
-* Fix an issue where the project would fail to compile with clang 4.0.
+* None.
 
 2.5.0 Release notes (2017-03-28)
 =============================================================
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,11 +7,12 @@ x.x.x Release notes (yyyy-MM-dd)
 
 ### Enhancements
 
-* None.
+* Add an `isAdmin` property to `RLMSyncUser` indicating whether a user is a
+  Realm Object Server administrator.
 
 ### Bugfixes
 
-* None.
+* Fix an issue where the project would fail to compile with clang 4.0.
 
 2.5.0 Release notes (2017-03-28)
 =============================================================
--- a/Realm/RLMSyncUser.h
+++ b/Realm/RLMSyncUser.h
@@ -70,6 +70,11 @@ NS_ASSUME_NONNULL_BEGIN
 @property (nullable, nonatomic, readonly) NSURL *authenticationServer;
 
 /**
+ Whether the user is an Realm Object Server administrator user.
+ */
+@property (nonatomic, readonly) BOOL isAdmin;
+
+/**
  The current state of the user.
  */
 @property (nonatomic, readonly) RLMSyncUserState state;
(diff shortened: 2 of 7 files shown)


austinzheng:
I've manually verified that this works, but without a way to programmatically create admin vs non-admin users we can't write a test for it. I've added a FIXME; maybe this is simple enough that we can put off writing the test.

Realm/RLMSyncUser.mm
```diff
@@ -253,6 +254,7 @@ + (void)_performLogInForUser:(RLMSyncUser *)user
                     return;
                 }
                 user->_user = sync_user;
+                user.isAdmin = model.refreshToken.tokenData.isAdmin;
```
bdash: Is this state that should be tracked on object store's `SyncUser` class? I see that it already has  `SyncUser::is_admin()`, which represents something different than this flag.

austinzheng: Yes. There are actually two issues here:

- One is that, as you said, the flag should be tracked on the object store class instead. If the user fetches another `RLMSyncUser` object pointing to the same user, it won't have the proper flag set.
- The other is that the terminology is really confusing. We have the notion of a "client side admin user", which is a user who uses the admin token to open Realms, and a "server side admin user", which is a user flagged by ROS as an 'admin'. Perhaps we should rename the public API to "isObjectServerAdmin" or something similar.

bdash: ISTM that the concepts are really:
* A user can be an admin or a normal user. Distinguishing those is what this PR is about.
* The initial admin user (I think? Or is it more than just the initial user?) can authenticate via a special access token, bypassing the normal authentication flow. This is what `SyncUser`'s `m_is_admin` currently tracks. This concept should probably be renamed to make it clear that it's about how the authentication is handled.

austinzheng:
I rebased and merged onto latest master. This can probably be reviewed now.

One thing I noticed was that git's conflict resolution algorithm might move changelog entries to random places in the changelog without complaining about merge conflicts. Something to watch out for (I suspect this may have happened to me in the past).

Realm/RLMSyncUser.h
```diff
@@ -70,6 +70,11 @@ NS_ASSUME_NONNULL_BEGIN
 @property (nullable, nonatomic, readonly) NSURL *authenticationServer;
 
 /**
+ Whether the user is an Realm Object Server administrator user.
```
bdash: The extra "user" at the end feels redundant.

CHANGELOG.md
```diff
@@ -7,11 +7,12 @@ x.x.x Release notes (yyyy-MM-dd)
...
 ### Enhancements
 
-* None.
+* Add an `isAdmin` property to `RLMSyncUser` indicating whether a user is a
+  Realm Object Server administrator.
 
 ### Bugfixes
 
-* None.
+* Fix an issue where the project would fail to compile with clang 4.0.
```
bdash: This isn't relevant to Cocoa at the moment since no Xcode version exists that uses Clang 4.0.

austinzheng:
Object server tests pass and I verified things work manually as well. I'll merge it when CI finishes.

Issue #4702, opened 2017-02-27
mutex lock failed: Invalid argument on `[RLMRealm realmWithConfiguration:error:]`


kanekin:
## Goals

Crash free Relam instantiation

## Expected Results

Getting an Realm instance without c++ exception

## Actual Results

```

Fatal Exception: std::__1::system_error: mutex lock failed: Invalid argument


#0. com.apple.main-thread
0  CortexFramework                0x1006d6b58 __llvm_gcov_indirect_counter_increment + 34
1  CortexFramework                0x100bfb898 llvm_writeout_files + 36
2  libsystem_c.dylib              0x18a898000 __cxa_finalize_ranges + 408
3  libsystem_c.dylib              0x18a841390 exit + 24
4  UIKit                          0x1918967fc +[_UIAlertManager hideAlertsForTermination] + 490
5  UIKit                          0x191a987cc __102-[UIApplication _handleApplicationDeactivationWithScene:shouldForceExit:transitionContext:completion:]_block_invoke.2100 + 792
6  UIKit                          0x191a9bfdc _runAfterCACommitDeferredBlocks + 292
7  UIKit                          0x191a8dd50 _cleanUpAfterCAFlushAndRunDeferredBlocks + 560
8  UIKit                          0x1917fd0b4 _afterCACommitHandler + 168
9  CoreFoundation                 0x18b9020c0 __CFRUNLOOP_IS_CALLING_OUT_TO_AN_OBSERVER_CALLBACK_FUNCTION__ + 32
10 CoreFoundation                 0x18b8ffcf0 __CFRunLoopDoObservers + 372
11 CoreFoundation                 0x18b900180 __CFRunLoopRun + 1024
12 CoreFoundation                 0x18b82e2b8 CFRunLoopRunSpecific + 444
13 GraphicsServices               0x18d2e2198 GSEventRunModal + 180
14 UIKit                          0x1918757fc -[UIApplication _run] + 684
15 UIKit                          0x191870534 UIApplicationMain + 208
16 Goalie                         0x10004aacc main (main.m:16)
17 libdispatch.dylib              0x18a8115b8 (Missing)

#1. com.apple.uikit.eventfetch-thread
0  libsystem_kernel.dylib         0x18a905188 mach_msg_trap + 8
1  libsystem_kernel.dylib         0x18a904ff8 mach_msg + 72
2  CoreFoundation                 0x18b9025d0 __CFRunLoopServiceMachPort + 192
[...]

PR #4703, opened 2017-02-28
Heap-allocate two locks, to avoid spurious errors when app is terminated

austinzheng:


Realm/RLMRealm.mm
```diff
@@ -287,8 +287,12 @@ + (instancetype)realmWithConfiguration:(RLMRealmConfiguration *)configuration er
     realm->_dynamic = dynamic;
 
     // protects the realm cache and accessors cache
-    static std::mutex initLock;
-    std::lock_guard<std::mutex> lock(initLock);
+    static std::mutex *initLock;
+    static dispatch_once_t onceToken;
+    dispatch_once(&onceToken, ^{
+        initLock = new std::mutex();
+    });
```
bdash: This can all be simplified to `static std::mutex& initLock = *new std::mutex;`

Realm/RLMRealmUtil.mm
```diff
@@ -36,11 +36,15 @@
 #import <unistd.h>
 
 // Global realm state
-static std::mutex s_realmCacheMutex;
+static dispatch_once_t s_realmCacheMutexDispatchToken;
+static std::mutex *s_realmCacheMutex;
```
bdash: Ditto about this being `static std::mutex& s_realmCacheMutex = *new std::mutex;`.

Note that heap-allocating this mutex without doing similarly for the data it protects (`s_realmsPerPath`) won't be super helpful.

Realm/RLMRealm.mm
austinzheng: Oh, this is good to know. Thanks!

umarniz:
Hi, I am coming from this issue:

https://github.com/realm/realm-cocoa/issues/4702

That references to this patch being a possible solution so firstly, thank you for the patch :).

Just wanted to point out that in the patch you are allocating the pointer on the heap to a static object but not freeing it anywhere and this would cause a memory leak, albeit only a tiny one (single mutex object = 64 bytes), as the heap is never de-allocated.

I could live with that as long as it resolves the realm crash I mentioned above as it is affecting ~50% of our users :(

bdash:
There is no leak here. The heap-allocated objects are always reachable and in use until the point the application exits.

umarniz:
@bdash sorry you are correct I just re-evaluated the condition and the heap allocated for the static object would be freed up with the application heap on exit.

I was using [leaker](http://left404.com/leaker/) to do leak tests but it would report this condition as a memory leak as we don't free the memory directly but it does free itself correctly for a static object when the application exits.

bdash:
Correct, all memory within the application's address space is discarded when the process exits. Tools that attempt to detect leaks by matching `free` with `malloc` will report false-positive in cases like this.

CHANGELOG.md
```diff
@@ -11,7 +11,8 @@ x.x.x Release notes (yyyy-MM-dd)
 
 ### Bugfixes
 
-* None.
+* Fix an issue where spurious crashes might occur when the host application is
```
bdash: It's not accurate to say the crashes are spurious. They're real crashes.

I'd phase this more like:
> Fix a crash that could occur if new Realm instances were created while the application was exiting.

PR #4715, opened 2017-03-07
Update object store submodule pointer

austinzheng:
Would appreciate any corrections or additions required to the changelog notes.


dependencies.list
```diff
@@ -1,4 +1,4 @@
 VERSION=2.4.3
-REALM_CORE_VERSION=2.3.1
-REALM_SYNC_VERSION=1.1.0
+REALM_CORE_VERSION=2.3.2
+REALM_SYNC_VERSION=1.3.0
```
bdash: Are there any other bug fixes in these two components that we should call out in our change log?

austinzheng: Probably. I'll go through the changelogs when I get home and update the PR.

austinzheng: It actually looks like there are no public-facing changes for sync, and the core changes are both for Android issues.

bdash: realm/realm-core#2465 doesn't appear to be an Android-specific fix?

austinzheng: That should probably be mentioned.

PR #4726, opened 2017-03-09
Pull in 'schema move' object store fix

austinzheng:
The object store also now contains some updates for compatibility with the Java binding, but as far as I can tell those don't affect the user-facing functionality of the Cocoa binding in any way.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,6 +32,8 @@ x.x.x Release notes (yyyy-MM-dd)
 * Fix an issue where certain `LIKE` queries could hang.
 * Fix an issue where `-[RLMRealm writeCopyToURL:encryptionKey:error]` could create
   a corrupt Realm file.
+* Fix an issue where incrementing a Realm's schema version without actually
+  changing the schema could cause a crash.
 
 2.4.3 Release notes (2017-02-20)
 =============================================================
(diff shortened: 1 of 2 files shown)


austinzheng:
🤦‍♂️

CHANGELOG.md
```diff
@@ -32,6 +32,8 @@ x.x.x Release notes (yyyy-MM-dd)
 * Fix an issue where certain `LIKE` queries could hang.
 * Fix an issue where `-[RLMRealm writeCopyToURL:encryptionKey:error]` could create
   a corrupt Realm file.
+* Fix an issue where incrementing a Realm's schema version without actually
```
bdash: I believe the bug was specific to synced Realms?

austinzheng: From the logic of the fix, it seemed to be the sort of thing that could happen to any Realm as long as it was in the `Additive` mode. I don't think there is anything sync-related on that codepath.

tgoyne: The additive schema mode isn't conceptually tied to sync, but it's currently not exposed as an option for non-sync realms.

austinzheng: Ah, that makes sense. I'll change the entry.

Issue #4724, opened 2017-03-09
Mac unit tests crash after updating to core v2.3.3 when run in the Release configuration


bdash:
After updating to core v2.3.3 in #4723, the Release configuration of the Mac unit tests is now crashing. The backtrace looks like so:

```
* thread #1: tid = 0x85e70f, 0x0000000104a3c7f3 Realm`realm::util::InterprocessCondVar::set_shared_part(realm::util::InterprocessCondVar::SharedPart&, std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> >, std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> >, std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> >) + 355, queue = 'com.apple.main-thread', stop reason = EXC_BAD_ACCESS (code=1, address=0x0)
    frame #0: 0x0000000104a3c7f3 Realm`realm::util::InterprocessCondVar::set_shared_part(realm::util::InterprocessCondVar::SharedPart&, std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> >, std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> >, std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> >) + 355
    frame #1: 0x0000000104b27c45 Realm`realm::SharedGroup::do_open(std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > const&, bool, bool, realm::SharedGroupOptions) + 2901
    frame #2: 0x00000001049e1582 Realm`realm::SharedGroup::open(realm::Replication&, realm::SharedGroupOptions) + 226
    frame #3: 0x00000001049e1187 Realm`realm::SharedGroup::SharedGroup(realm::Replication&, realm::SharedGroupOptions) + 1527
[...]

PR #4723, opened 2017-03-08
Update to core v2.3.3

bdash:
Fixes #4690.

PR #4727, opened 2017-03-09
Revert to core v2.3.2.

bdash:
Reverts #4723 due to #4724.


bdash:
CI is completely busted at the moment, but there's no reason to think that reverting this will be a problem, so I'm going to go ahead and merge it.

PR #4742, opened 2017-03-16
Support the diacritic-insensitive modifier for string comparisons

bdash:
This adds support for the diacritic-insensitive modifier, `[d]`, to all supported string comparison operators except `LIKE`.

The performance is significantly worse than for our existing diacritic-sensitive comparisons, but is comparable to CoreData's performance on equivalent queries.

### DIFF
--- a/Realm/RLMQueryUtil.mm
+++ b/Realm/RLMQueryUtil.mm
@@ -88,6 +88,7 @@ BOOL RLMPropertyTypeIsNumeric(RLMPropertyType propertyType) {
 
 // Declare an overload set using lambdas or other function objects.
 // A minimal version of C++ Library Evolution Working Group proposal P0051R2.
+// FIXME: Switch to realm::util::overload once https://github.com/realm/realm-core/pull/2539 is in a core release.
 
 template <typename Fn, typename... Fns>
 struct Overloaded : Fn, Overloaded<Fns...> {
--- a/Realm/RLMQueryUtil.mm
+++ b/Realm/RLMQueryUtil.mm
@@ -146,7 +146,21 @@ struct FalseExpression : realm::Expression {
 // Equal and ContainsSubstring are used by QueryBuilder::add_string_constraint as the comparator
 // for performing diacritic-insensitive comparisons.
 
-template <NSUInteger options>
+bool equal(NSStringCompareOptions options, StringData v1, StringData v2)
+{
+    if (v1.is_null() || v2.is_null()) {
+        return v1.is_null() == v2.is_null();
+    }
+
+    @autoreleasepool {
+        NSString *s1 = [[NSString alloc] initWithBytesNoCopy:(void*)v1.data() length:v1.size() encoding:NSUTF8StringEncoding freeWhenDone:NO];
+        NSString *s2 = [[NSString alloc] initWithBytesNoCopy:(void*)v2.data() length:v2.size() encoding:NSUTF8StringEncoding freeWhenDone:NO];
+
+        return [s1 compare:s2 options:options] == NSOrderedSame;
+    }
+}
+
+template <NSStringCompareOptions options>
 struct Equal {
     using CaseSensitive = Equal<options & ~NSCaseInsensitiveSearch>;
     using CaseInsensitive = Equal<options | NSCaseInsensitiveSearch>;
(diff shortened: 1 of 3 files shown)


austinzheng:
All 3 job failures (so far) are CI issues.

Realm/RLMQueryUtil.mm
```diff
@@ -85,6 +85,34 @@ BOOL RLMPropertyTypeIsNumeric(RLMPropertyType propertyType) {
...
+template <typename Fn>
+struct Overloaded<Fn> : Fn {
+    template <typename U>
+    Overloaded(U&& fn) : Fn(std::forward<U>(fn)) { }
+
+    using Fn::operator();
+};
+
+template <typename... Fns>
+Overloaded<Fns...> overload(Fns&&... f)
```
bdash: There's probably a better home for this. Not sure where, though. Object store's `util` directory, perhaps?

Realm/RLMQueryUtil.mm
```diff
@@ -113,6 +141,68 @@ void set_base_table(const Table*) override {}
...
+};
+
+template <NSUInteger options>
+struct ContainsSubstring {
+    using CaseSensitive = ContainsSubstring<options & ~NSCaseInsensitiveSearch>;
+    using CaseInsensitive = ContainsSubstring<options | NSCaseInsensitiveSearch>;
+
+    bool operator()(StringData v1, StringData v2, bool v1_null, bool v2_null) const
+    {
+        REALM_ASSERT_DEBUG(v1_null == v1.is_null());
```
tgoyne: Could avoid emitting a whole bunch of nearly identical copies of this function by moving the body to a function which takes `options` as a non-template parameter.

Realm/RLMQueryUtil.mm
```diff
@@ -113,6 +141,68 @@ void set_base_table(const Table*) override {}
     }
 };
 
+
+// Equal and ContainsSubstring are used by QueryBuilder::add_string_constraint as the comparator
+// for performing diacritic-insensitive comparisons.
+
+template <NSUInteger options>
+struct Equal {
+    using CaseSensitive = Equal<options & ~NSCaseInsensitiveSearch>;
```
tgoyne: I'm not sure these aliases are actually useful. They look weird on their own (my initial thought before seeing how they're used was "why would you do `Equal<NSDiacriticInsensitivePredicateOption>::CaseInsensitive` rather than `Equal<NSDiacriticInsensitivePredicateOption|NSCaseInsensitiveSearch>`") and t he fact that they're defined in two places but only used in one means they aren't reducing duplication

Realm/Tests/QueryTests.m
```diff
@@ -895,6 +899,11 @@ - (void)testStringBeginsWith
     RLMAssertCount(StringObject, 1U, @"stringCol BEGINSWITH[c] 'a'");
     RLMAssertCount(StringObject, 1U, @"stringCol BEGINSWITH[c] 'A'");
 
+    RLMAssertCount(StringObject, 1U, @"stringCol BEGINSWITH 'u'");
+    RLMAssertCount(StringObject, 1U, @"stringCol BEGINSWITH[c] 'U'");
+    RLMAssertCount(StringObject, 3U, @"stringCol BEGINSWITH[d] 'u'");
+    RLMAssertCount(StringObject, 3U, @"stringCol BEGINSWITH[cd] 'U'");
```
tgoyne: `stringCol BEGINSWITH 'ü'` etc. seems worth testing as well.

bdash:
Retest this please

Realm/RLMQueryUtil.mm
bdash: It allows for moving the handling of case-sensitive vs case-insensitive inside the `add_constraint` function rather than repeating it for each operator type. `add_constraint` is invoked with the operator type, and then it uses these type aliases to compute the case-sensitive / case-insensitive variants of that operator type.

Realm/RLMQueryUtil.mm
bdash: realm/realm-core#2539, and I added a FIXME to make it clear that we can just remove this code when we have a new core release.

Realm/RLMQueryUtil.mm
```diff
@@ -113,6 +141,68 @@ void set_base_table(const Table*) override {}
...
+    {
+        REALM_ASSERT_DEBUG(v1_null == v1.is_null());
+        REALM_ASSERT_DEBUG(v2_null == v2.is_null());
+
+        if (v1_null || v2_null) {
+            return v1_null == v2_null;
+        }
+
+        @autoreleasepool {
+            NSString *s1 = [[NSString alloc] initWithBytesNoCopy:(void*)v1.data() length:v1.size() encoding:NSUTF8StringEncoding freeWhenDone:NO];
```
bdash: I had compared the two back when I first started looking at it, and the differences appeared to be within the noise (a rather large amount of noise at the time). I just compared again with a better test scenario (larger table, more iterations of the test) and there's a measurable ~15% win from using the CF APIs.

PR #4744, opened 2017-03-16
Add support for counters

austinzheng:
PROVISIONAL: Add support for the new `RLMInteger` and `RLMNullableInteger` types, which act as normal integers and contain API methods that allow them to take advantage of additional Realm functionality.

The basic functionality is working and there are a few tests.

To do (add items as necessary)
- [ ] Review the API proposed in this PR
- [x] Should the new types inherit from `NSNumber`? (seems like no)
- [x] Compatibility with KVO
- [x] Compatibility with notifications
- [x] Compatibility with migrations
- [x] Compatibility with queries
- [x] Compatibility with initializing an object using a value
- [x] Swift support
- [x] Figure out how this works with the dynamic API
- [x] Adapt this to use Thomas's accessor changes once they're finished
- [x] Fix everything marked in the code with a `TODO` flag
- [x] Tests to check what happens if you init a required property with a Realm int set to nil.
- [ ] Make any changes necessary for this to work with arrays of primitives.

Bugs:
- [x]  Need to figure out what to do about object store-level schemas, which have no knowledge about the difference between counters and normal nullable/non-nullable integers. Right now we set the property on the schema as a nullable int, but it's being read back out as a non-nullable int, causing a test to fail (and other problems).


emanuelez:
Retest this please

austinzheng:
I don't expect the tests to pass yet, there are some outstanding issues.

emanuelez:
Sorry for the disruption @austinzheng. This PR just happened to be the first one I found during my tests earlier today.

austinzheng:
No worries @emanuelez, I just didn't want you to be misled because of all the failures :).

austinzheng:
I'm still working on this, but since the remaining work is mostly more tests and the Swift API, I've added some reviewers.

larkost:
The current failures are primarily due to a pair of tests failing:
[CompactionTests.testReturnNoCompactOnLaunch](https://ci.realm.io/job/objc_pr/configuration=Release,target=ios-device-objc,xcode_version=8.3.1/5016/testReport/junit/(root)/CompactionTests/testReturnNoCompactOnLaunch/)
[CompactionTests.testSuccessfulCompactOnLaunch]([https://ci.realm.io/job/objc_pr/configuration=Release,target=ios-device-objc,xcode_version=8.3.1/5016/testReport/junit/(root)/CompactionTests/testSuccessfulCompactOnLaunch/)

jpsim:
Those tests are way too strict. I'll take care of relaxing their constraints in a PR shortly.

Realm/RLMAccessor.mm
```diff
@@ -419,15 +419,22 @@ static id makeSetter(__unsafe_unretained RLMProperty *const prop) {
     };
 }
 
+static inline NSNumber<RLMInt> *RLMCoerceToNSNumber(__unsafe_unretained id obj) {
+    if ([obj conformsToProtocol:@protocol(RLMIntegerProtocol)]) {
+        return [obj boxedValue];
+    }
+    REALM_ASSERT_DEBUG(obj == nil || RLMDynamicCast<NSNumber>(obj));
```
austinzheng: This shouldn't be a `REALM_ASSERT`, since user error can trigger this.

austinzheng:
I feel like the correct thing to do is to rewrite this PR to use `RLMProxyBase` (or whatever it's called) as the base for `RLMInteger`, and get rid of the two different types (nullable and non-nullable).

austinzheng:
I don't think we have any spare bandwidth, but whenever we do this is probably ready for a first pass.

austinzheng:
Not an employee anymore, but would love to get this in...eventually. If there are plans to review and merge this in the future, I'm happy to keep it up-to-date and on latest `master` (cc @bdash).

PR #4746, opened 2017-03-17
Change our approach to resetting iOS simulators

### LINKED ISSUE #4095: Rewrite reset-simulators.sh in ruby
Rather than going back and forth between ruby and bash as it currently does.

bdash:
We used to do the following:
1. Kill any running simulator processes.
2. Shut down any running simulator devices.
3. Erase the contents of all simulator devices.
4. Launch Simulator.app.
5. Wait for Simulator.app to boot a simulator device.

Steps 3, 4 and 5 were somewhat problematic as there's no guarantee that the simulator devices will exist and be in a valid state that will allow Simulator.app to launch them. This often resulted in step 5, waiting for the simulator devices to boot, timing out.

We now do the following:
1. Kill any running simulator processes.
2. Delete any unavailable simulator devices.
3. Shut down any running simulator devices.
4. Delete all remaining simulator devices.
5. Create a fresh simulator device for each valid platform / device type
   combination.

Fixes #4095.

### DIFF
--- a/scripts/reset-simulators.rb
+++ b/scripts/reset-simulators.rb
@@ -79,12 +79,16 @@ begin
     runtimes_by_platform[platform_for_runtime(runtime)] << runtime
   end
 
-  puts 'Creating fresh simulators...'
+  print 'Creating fresh simulators...'
   device_types.each do |device_type|
     platform = platform_for_device_type(device_type)
     runtimes_by_platform[platform].each do |runtime|
       output = `xcrun simctl create '#{device_type['name']}' '#{device_type['identifier']}' '#{runtime['identifier']}' 2>&1`
       next if $? == 0
+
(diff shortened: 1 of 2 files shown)


jpsim:
This looks off to a great start, also addresses #4095. Lots of little nitpicks that could be addressed by running `rubocop -a scripts/reset-simulators.rb` and `rubocop scripts/reset-simulators.rb` (though in my experience that applies corrections that aren't actually improvements, so requires human vetting, yay linters!).

bdash:
This PR does appear to have helped. https://ci.realm.io/job/objc_pr/4838/ is the most recent build of this PR, and it hit zero issues related to the simulator. It was built immediately after https://ci.realm.io/job/objc_pr/4837/, in which the simulator failed to boot in half a dozen or more of the jobs.

bdash:
This is now ready for review.

Issue #3289, opened 2016-03-04
add RLMRealmConfiguration.compactOnLaunch


jpsim:
It's common for users with large Realm files with significant amounts of unused space to want to compact them on a regular basis. The code users typically write to accomplish this is error prone due to challenges dealing with Realm instance life cycles.

Unfortunately, if this option existed, it would likely be abused and used in cases in which compaction wouldn't actually reclaim much disk space. If core had a way to get a ballpark estimate of how much space could be reclaimed, this could be used to more intelligently perform compactions.

tgoyne:
Approximating the amount of space that will be freed up by compact() is pretty fast and simple (just add up the size of everything in the free list). Rather than just a binary "always compact on open", we could maybe expose "compact if live data is less than X% of the file"?

cmelchior:
It sounds weird to expose such an option, everybody would just enable it I think?

jpsim:
> > we could maybe expose "compact if live data is less than X% of the file"?
>
> It sounds weird to expose such an option, everybody would just enable it I think?

Maybe having a `RLMRealmConfiguration.compactOnUsageRatio` with a default value of `0.5` could work? Naming needs work, obviously.

jasper-ch-chan:
Any chance this feature can be moved into a P1 backlog?

jpsim:
I don't think so. [Our current P1s](https://github.com/realm/realm-cocoa/issues?q=is%3Aopen+is%3Aissue+label%3A%22S%3AP1+Backlog%22) are all higher priority.

jpsim:
Merged in #4755 and available in today's 2.6 release: https://realm.io/news/realm-objc-swift-2.6

PR #4755, opened 2017-03-21
Expose a block to determine if the Realm should be compacted on launch

jpsim:
Addresses #3289.

### DIFF
--- a/Realm/ObjectServerTests/RLMObjectServerTests.mm
+++ b/Realm/ObjectServerTests/RLMObjectServerTests.mm
@@ -1719,9 +1719,6 @@
 #pragma mark - Validation
 
 - (void)testCompactOnLaunchCannotBeSet {
-    // This test just needs a single process, so skip child.
-    if (!self.isParent) { return; }
-
     RLMSyncUser *user = [self logInUserForCredentials:[RLMObjectServerTests basicCredentialsWithName:ACCOUNT_NAME()
                                                                                             register:YES]
                                                server:[RLMObjectServerTests authServerURL]];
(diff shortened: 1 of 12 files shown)


jpsim:
This is now ready for review.

Realm/ObjectServerTests/RLMObjectServerTests.mm
```diff
@@ -1716,4 +1716,21 @@ - (void)testPermissionOfferResponseTokenNotExist {
     [token stop];
 }
 
+#pragma mark - Validation
+
+- (void)testCompactOnLaunchCannotBeSet {
+    // This test just needs a single process, so skip child.
+    if (!self.isParent) { return; }
```
tgoyne: The child process is created by `RLMRunChildAndWait()`, so there's no need to do anything other than just not call that if only one process is needed.

Realm/RLMRealmConfiguration.mm
```diff
@@ -275,4 +279,21 @@ - (NSString *)pathOnDisk {
     return @(_config.path.c_str());
 }
 
+- (void)setShouldCompactOnLaunch:(RLMShouldCompactOnLaunchBlock)shouldCompactOnLaunch {
+    _shouldCompactOnLaunch = shouldCompactOnLaunch;
```
tgoyne: Should set this after validation so that the object isn't left in an inconsistent state when validation fails.

RealmSwift/Tests/CompactionTests.swift
```diff
@@ -0,0 +1,86 @@
+////////////////////////////////////////////////////////////////////////////
+//
+// Copyright 2015 Realm Inc.
```
austinzheng: 2017

PR #4756, opened 2017-03-21
delete derived data if pcm file needs to be rebuilt

jpsim:
For errors such as:

> module file '.../build/DerivedData/.../Module.pcm' is out of date
> and needs to be rebuilt

We should actually delete the derived data directory in order to
rebuild the pcm file

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -1049,6 +1049,9 @@ EOM
             sh build.sh verify-$target 2>&1 | tee build/build.log | xcpretty -r junit -o build/reports/junit.xml || failed=1
             if [ "$failed" = "1" ] && cat build/build.log | grep -E 'DTXProxyChannel|DTXChannel|out of date and needs to be rebuilt|operation never finished bootstrapping'; then
                 echo "Known Xcode error detected. Running job again."
+                if cat build/build.log | grep -E 'out of date and needs to be rebuilt'; then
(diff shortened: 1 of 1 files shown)


jpsim:
For example: https://ci.realm.io/job/objc_pr/4860/configuration=Release,target=osx,xcode_version=8.1/console

build.sh
```diff
@@ -1049,6 +1049,9 @@ EOM
             sh build.sh verify-$target 2>&1 | tee build/build.log | xcpretty -r junit -o build/reports/junit.xml || failed=1
             if [ "$failed" = "1" ] && cat build/build.log | grep -E 'DTXProxyChannel|DTXChannel|out of date and needs to be rebuilt|operation never finished bootstrapping'; then
                 echo "Known Xcode error detected. Running job again."
+                if cat build/build.log | grep -E 'out of date and needs to be rebuilt'; then
```
bdash: I assume that we start off with an empty build directory, so how do we end up with an out of date module file? Do we think this error is indicative of a Xcode / compiler bug?

jpsim: We do start off with an empty build directory. I don't know what causes the pcm file to appear out of date, but I can say for sure that our current approach of just re-running xcodebuild will never succeed without first deleting the outdated file.

PR #4757, opened 2017-03-22
Improve the obj-c object tests and fix some minor bugs

tgoyne:
The tests have been restructured to put everything related to init/add/create in ObjectCreationTests, things directly related to getting an object schema from a class in ObjectSchemaTests, and tests about the generated property getters/setters in ObjectInterfaceTests. All the assorted other things remain in ObjectTests. The object creation tests now actually cover 100% of the code which can be covered by obj-c object creation tests, and a lot of assorted bits and pieces that were previously hit by tests but not meaningfully tested (in that they could be broken without any tests failing) are now actually verified.

initWithValue: and createInRealm:withValue: pointlessly differed on what input they accepted. They now share more of their validation logic, and in some cases the validation logic has been loosened so that values which previously could only be passed to one can now be passed to both.

All of the object tests which test that an exception is thrown now also check that the exception message is something reasonable. This revealed a few C++ exceptions that weren't being converted to NSException, and some error cases that just didn't have very good error messages.


Realm/Tests/ObjectCreationTests.mm
```diff
@@ -0,0 +1,1231 @@
...
+    auto ca = [[CustomAccessorsObject alloc] initWithValue:@[@"a", @1]];
+    XCTAssertEqualObjects(ca.name, @"a");
+    XCTAssertEqual(ca.age, 1);
+
+    // Create with dictionary
+    ca = [[CustomAccessorsObject alloc] initWithValue:@{@"name": @"b", @"age": @2}];
+    XCTAssertEqualObjects(ca.name, @"b");
+    XCTAssertEqual(ca.age, 2);
+
+    // Create with KVO-compatible object
```
jpsim: "KVC-compatible"

Realm/RLMUtil.mm
```diff
@@ -169,6 +170,38 @@ BOOL RLMIsObjectValidForProperty(__unsafe_unretained id const obj,
...
+        case RLMPropertyTypeData:
+            if (!RLMIsObjectValidForProperty(obj, prop)) {
+                @throw RLMException(@"Invalid value '%@' for property '%@'", obj, prop.name);
+            }
+            break;
+        case RLMPropertyTypeObject:
+            break;
+        case RLMPropertyTypeArray: {
+            if (obj != nil && obj != NSNull.null) {
+                if (![obj conformsToProtocol:@protocol(NSFastEnumeration)]) {
```
jpsim: this condition could be folded into its parent

Realm/RLMObjectStore.mm
```diff
@@ -352,7 +322,10 @@ void RLMAddObjectToRealm(__unsafe_unretained RLMObjectBase *const object,
     RLMInitializeSwiftAccessorGenerics(object);
 }
 
-RLMObjectBase *RLMCreateObjectInRealmWithValue(RLMRealm *realm, NSString *className, id value, bool createOrUpdate = false) {
+RLMObjectBase *RLMCreateObjectInRealmWithValue(RLMRealm *realm, NSString *className,
+                                               id value, bool createOrUpdate = false) {
+    RLMVerifyInWriteTransaction(realm);
```
jpsim: I look forward to seeing what invalid code this breaks 😅

Realm/RLMObjectStore.mm
```diff
@@ -361,8 +334,9 @@ void RLMAddObjectToRealm(__unsafe_unretained RLMObjectBase *const object,
         }
     }
 
-    // verify writable
-    RLMVerifyInWriteTransaction(realm);
+    if (!value) {
```
jpsim: should this also check against `NSNull.null`?

Realm/RLMRealm.mm
```diff
@@ -142,7 +142,12 @@ - (BOOL)isEmpty {
 }
 
 - (void)verifyThread {
-    _realm->verify_thread();
+    try {
+        _realm->verify_thread();
+    }
+    catch (std::exception const& e) {
+        @throw RLMException(@"%s", e.what());
```
jpsim: we usually use `@throw RLMException(e);` in cases like these

Realm/Tests/PropertyTests.m
```diff
@@ -139,4 +144,18 @@ - (void)testSwiftPropertyNameValidation {
     XCTAssertNoThrow(RLMValidateSwiftPropertyName(@"__alloc"));
 }
 
+- (void)testTypeToString {
```
jpsim: 1. Do we want this now that you've added `static_asserts` testing the same thing?
2. This is missing "linking objects".

tgoyne: This test is pretty dubious since it's not really going to catch anything other than a typo in RLMTypeToString, but it is sorta testing something.

tgoyne:
Only failure is tvOS timing out when launching the tests on the device and it passed on a previous run.

Issue #4730, opened 2017-03-13
Xcode plugin in Swift/Objc docs


dakeshi:
Hi. I find out two issues in the Xcode plugin section in the latest Swift docs.

In the documentation, it seems that we can install RealmPlugin through Alcatraz.
> The easiest way to install the Realm Xcode plugin is through Alcatraz under the name “RealmPlugin”. 

But, as it mentioned in #4357, we can not use Alcatraz anymore in Xcode 8 later without unsigned or resigned Xcode.  (refer to https://github.com/alcatraz/Alcatraz/issues/475, https://github.com/inket/update_xcode_plugins, https://github.com/XVimProject/XVim/blob/3167408ade82cfef87acc704822da61af69688f8/INSTALL_Xcode8.md). 
I think that removing contents about Alcatraz will be helpful to avoid confusion.
Are there some reasons to remain this description for Alcatraz? 

Another issue is the README.md for RealmPlugin which downloads from the release.zip link in docs.
> 3. A menu item in Xcode's 'File' menu to quickly launch the Realm Browser.

I install RealmPlugin with RealmPlugin.xcodeproj but I can not see Realm Browser menu in File menu. Is it normal? or wrong info in README.md?

jpsim:
Thanks for asking these questions, @dakeshi! We should certainly update our docs to reflect that Alcatraz requires resigning Xcode.

As for your second question, the "Open Realm..." menu item is only added if the Realm Browser is installed on your Mac. https://github.com/realm/realm-cocoa/blob/v2.4.3/plugin/RealmPlugin/RLMPRealmPlugin.m#L87-L108

dakeshi:
Thanks for your response @jpsim 

>We should certainly update our docs to reflect that Alcatraz requires resigning Xcode.

I’m not sure that we need to mention how to resign or unsign Xcode 8 to use Alcatraz. 
Without Alcatraz support, we can still install RealmPlugin in Xcode 8(with release.zip)
Actually, it is not the RealmPlugin issue but 3rd party plugin issue.
Using Alcatraz is more convenient but it is not the mandatory to use RealmPlugin, I think. 
If some users want to install RealmPlugin using Alcatraz tools, 
they could get the solution from the Alcatraz repo.

[...]

PR #4763, opened 2017-03-23
add note about plugin menu item not being displayed in Xcode 8

jpsim:
There was some user confusion in https://github.com/realm/realm-cocoa/issues/4730.


bdash:
Is it even worth keeping this plug-in around at this point? It won't be loaded by current versions of Xcode, and the installation of the templates and LLDB script could be accomplished in other ways.

jpsim:
Until we migrate to a different installation mechanism for the templates and LLDB script, better document the current behavior?

But yeah, just having a `install_realm_addons.sh` script in the release zips is probably best moving forward.

PR #4778, opened 2017-03-27
Hide comments in issues

bmunkholm:


jpsim:
I don't think the previous version was causing problems, nor do I think this will have an effect on the quality of reports we receive, but I don't have a strong opinion, so let's just merge this.

Issue #4585, opened 2017-01-24
Intermittent `SchemaTests` failure


austinzheng:
I occasionally see the following test failure on CI jobs running for PRs:

```
-[SchemaTests testMultipleProcessesTryingToInitializeSchema] : Asynchronous wait failed: Exceeded timeout of 10 seconds, with unfulfilled expectations: "notification fired".
```

Not sure what is causing it, why it's sporadic, etc, but filing a ticket to track.

Example:
https://ci.realm.io/job/objc_pr/configuration=Debug,swift_version=3.0.2,target=osx/4651/console

jpsim:
Seen again in [this run](https://ci.realm.io/job/objc_pr/configuration=Debug,target=osx,xcode_version=8.0/4901/testReport/(root)/SchemaTests/testMultipleProcessesTryingToInitializeSchema/). @tgoyne would you object to disabling this test in debug configurations?

austinzheng:
According to https://github.com/realm/realm-cocoa/pull/4789#pullrequestreview-29581582, https://github.com/realm/realm-core/pull/2552 may fix the underlying issue.

jpsim:
Unfortunately saw this again in a PR job from #4755, which included core 2.6: https://ci.realm.io/job/objc_pr/5006/configuration=Debug,target=osx,xcode_version=8.0/

PR #4781, opened 2017-03-27
disable testMultipleProcessesTryingToInitializeSchema in debug mode

jpsim:
since it occasionally fails on CI and its failure doesn't indicate something inherently broken with the patch being tested.

Addresses https://github.com/realm/realm-cocoa/issues/4585


jpsim:
The failed CocoaPods job was running successfully, but very slowly, so it was aborted by Jenkins' 35 minute timeout.

jpsim:
Let's wait and see what comes of https://github.com/realm/realm-cocoa/issues/4585#issuecomment-289914342

PR #4782, opened 2017-03-27
Support encrypted synchronized Realms

jpsim:
Still missing changelog entries for this and
other changes that happen to have been pulled in due to the
core, sync and object store updates


Realm/RLMQueryUtil.mm
```diff
@@ -126,6 +126,7 @@ size_t find_first(size_t start, size_t end) const override
         return realm::not_found;
     }
     void set_base_table(const Table*) override {}
+    void verify_column() const override {}
```
austinzheng: I ran into this issue as well. We should make sure that an empty override is the right thing to do in this case.

Realm/RLMRealm.mm
```diff
@@ -289,6 +290,12 @@ + (instancetype)realmWithConfiguration:(RLMRealmConfiguration *)configuration er
     configuration = [configuration copy];
     Realm::Config& config = configuration.config;
 
+    if (configuration.encryptionKey && config.sync_config) {
+        std::array<char, 64> sync_encryption_key;
+        std::copy_n(config.encryption_key.begin(), 64, sync_encryption_key.begin());
+        config.sync_config.get()->realm_encryption_key = std::move(sync_encryption_key);
```
bdash: `std::move` isn't useful here as the data is stored directly within `std::array` and inherently has to be copied. It'd probably be simplest to just populate the `realm_encryption_key` member directly rather than going via a temporary.

Realm/RLMQueryUtil.mm
bdash: It is. `TrueExpression` / `FalseExpression` don't have columns so there's nothing to verify.

Realm/RLMRealm.mm
austinzheng: My bad. I suggested this approach.

austinzheng:
Hmm...tests are failing with `fatal error: 'sync_config.hpp' file not found: #include "sync_config.hpp"`. Can you check the target membership settings on the object store files in Xcode?

Realm/RLMRealm.mm
```diff
@@ -41,6 +41,7 @@
 #include "object_store.hpp"
 #include "schema.hpp"
 #include "shared_realm.hpp"
+#include "sync_config.hpp"
```
tgoyne: sync/sync_config.hpp

Realm/RLMRealm.mm
tgoyne: Setting the encryption key on the SyncConfig object should be done in RLMRealmConfiguration like all of the other fields.

Realm/ObjectServerTests/RLMObjectServerTests.m
```diff
@@ -318,6 +319,67 @@ - (void)testDeleteObjects {
...
+        NSString *path;
+        @autoreleasepool {
+            RLMRealm *realm = [self openRealmForURL:url user:user encryptionKey:RLMGenerateKey()
+                                         stopPolicy:RLMSyncStopPolicyImmediately immediatelyBlock:nil];
+            path = realm.configuration.pathOnDisk;
+            CHECK_COUNT(0, SyncObject, realm);
+            RLMRunChildAndWait();
+            [self waitForDownloadsForUser:user realms:@[realm] realmURLs:@[url] expectedCounts:@[@3]];
+        }
+        RLMRealmConfiguration *c = [RLMRealmConfiguration defaultConfiguration];
```
tgoyne: This test should also test re-opening it as a synced Realm with the wrong encryption key.

jpsim: I'd like to do that, but the sync client is getting in the way by throwing an exception in its own thread, even with the sync policy mode set to "immediately". Do you have any ideas on how to catch that? Or make the object store / cocoa exception happen before that point?

Realm/RLMRealm.mm
jpsim: That complicates things unnecessarily. What if the encryption key is set before the sync configuration is set?

tgoyne: Having a weird special case in some unrelated code is much worse than an extra few lines of code.

jpsim: Done.

Realm/RLMRealmConfiguration+Sync.mm
```diff
@@ -51,6 +51,10 @@ - (void)setSyncConfiguration:(RLMSyncConfiguration *)syncConfiguration {
     self.config.in_memory = false;
     self.config.sync_config = std::make_shared<realm::SyncConfig>([syncConfiguration rawConfiguration]);
     self.config.schema_mode = realm::SchemaMode::Additive;
+    if (!self.config.encryption_key.empty()) {
+        self.config.sync_config.get()->realm_encryption_key = std::array<char, 64>();
```
bdash: `.get()->` on a `unique_ptr` / `shared_ptr` is equivalent to just `->`.

I'd probably add `auto& sync_encryption_key = self.config.sync_config->realm_encryption_key;` here so that this line and the following can refer to it as `sync_encryption_key`.

Realm/RLMRealmConfiguration.mm
```diff
@@ -25,6 +25,9 @@
 
 #import "schema.hpp"
 #import "shared_realm.hpp"
+#import "sync/sync_config.hpp"
+
+#include <realm/util/optional.hpp>
```
bdash: This doesn't appear to be necessary. `SyncConfig` has a `util::Optional` as a member so is required to have included this header.

Realm/RLMRealmConfiguration.mm
```diff
@@ -176,9 +177,16 @@ - (void)setEncryptionKey:(NSData * __nullable)encryptionKey {
...
+        if (_config.sync_config) {
+            auto& sync_encryption_key = self.config.sync_config->realm_encryption_key;
+            sync_encryption_key = std::array<char, 64>();
+            std::copy_n(_config.encryption_key.begin(), 64, sync_encryption_key->begin());
+        }
     }
     else {
         _config.encryption_key.clear();
+        if (_config.sync_config)
+            _config.sync_config.get()->realm_encryption_key = realm::util::none;
```
bdash: You overlooked this `.get()->`

dependencies.list
```diff
@@ -1,4 +1,4 @@
 VERSION=2.4.4
-REALM_CORE_VERSION=2.3.3
-REALM_SYNC_VERSION=1.3.1
-REALM_OBJECT_SERVER_VERSION=1.2.1
+REALM_CORE_VERSION=2.4.0
+REALM_SYNC_VERSION=1.4.2
```
bdash: Is there a reason we're going to 1.4.2 here rather than 1.5.0?

jpsim: Not that I can think of, no. I'll update.

jpsim:
I've added some changelog entries. I don't think realm/realm-core#2415 was ever triggerable by our end users, so I didn't include that in the changelog entries, but I'd appreciate if someone could confirm.

tgoyne:
realm/realm-core#2415 is only relevant with a true dynamic schema, which we don't support.

Realm/ObjectServerTests/RLMObjectServerTests.m
tgoyne: I suppose it's fine for now if we need to get this out, but crashing in an awkward place when given an incorrect encryption key sounds like a problem that needs to be addressed.

jpsim: Is this a deal breaker for you to get this in?

tgoyne: I suppose it's fine for now if we need to get this out.

Fixing it probably requires changes on the sync side to report the error in a better way.

jpsim:
Addressed all feedback so far, have changelog entries, updated to core 2.5.1 and sync 1.5.0. Object store is at realm/realm-object-store#420, which I'll update to point to `master` once that PR passes CI and is merged.

CHANGELOG.md
```diff
@@ -11,6 +17,8 @@ x.x.x Release notes (yyyy-MM-dd)
 * Add support for the `[d]` modifier on string comparison operators to perform
   diacritic-insensitive comparisons.
 * Explicitly mark `[[RLMRealm alloc] init]` as unavailable.
+* Don't unnecessarily advance the Realm transaction when resolving a
```
bdash: This reads more like a bug fix than an enhancement?

CHANGELOG.md
```diff
@@ -1,6 +1,12 @@
 x.x.x Release notes (yyyy-MM-dd)
 =============================================================
 
+This release requires all processes accessing a Realm file concurrently to be on
+this version or later. This includes iOS extensions and the Realm Browser.
```
tgoyne: I believe older versions will fail to open files written by this version even if it's non-concurrent.

CHANGELOG.md
tgoyne: I'm not sure what change this is referring to? 7cf4c6f is the only TSR-related change that I see, and that's something significantly different that doesn't apply to cocoa.

jpsim: Yes, that's the commit that prompted me to write this entry, but you're right that it doesn't apply. Removed.

PR #4789, opened 2017-03-28
update CI job, packaging scripts and directory layout for Swift 3.1

jpsim:


tgoyne:
Oh, actually those look like real test failures...

jpsim:
Yeah, the crash in `testDefaultValuesFromNSNull` is concerning, and the linker error for the static framework. I think the latter might be caused by Xcode 8.3 no longer supporting iOS 7 as a minimum target.

tgoyne:
I can't reproduce the testDefaultValuesFromNSNull failure locally, and it passes when building with address sanitizer. It sure would be nice if xctest didn't swallow the error...

jpsim:
I've updated the matrix to skip ios-static on Xcode 8.3 and changed the packaging script to always compile the ObjC framework with Xcode 8.2. @tgoyne @bdash could one of you please take another look?

build.sh
```diff
@@ -174,6 +181,13 @@ build_combined() {
     if [[ "$destination" != "" && "$config" == "Release" ]]; then
         sh build.sh binary-has-bitcode "$LIPO_OUTPUT"
     fi
+
+    if [[ "$module_name" == "Realm" ]]; then
```
tgoyne: Don't actually need this check.

build.sh
```diff
@@ -130,6 +130,13 @@ build_combined() {
     local scope_suffix="$5"
     local version_suffix="$6"
     local config="$CONFIGURATION"
+    local previous_realm_xcode_version="$REALM_XCODE_VERSION"
+    local previous_realm_swift_version="$REALM_SWIFT_VERSION"
+
+    if [[ "$module_name" == "Realm" ]]; then
```
tgoyne: I'd be inclined to push this check inside workaround_rdar_31302382.

jpsim: it's a local variable, so you're suggesting passing the module name to the function?

jpsim: actually, I was considering just inlining `force_xcode82` and `workaround_rdar_31302382` into here since we only invoke it once now.

PR #4790, opened 2017-03-28
reset simulators when forcing xcode 8.2

jpsim:


bdash:
The two `ios-static` failures look like they may be real problems

jpsim:
> The two `ios-static` failures look like they may be real problems

Yes. I think I've addressed that now.

PR #4793, opened 2017-03-29
CocoaPods `target has libraries with conflicting names ` error

OliverLetterer:
Hello,

I think found a problem when integrating Realm with CocoaPods and using multiple platforms in a Podfile while integrating with CocoaPods static libraries instead of CocoaPods dynamic frameworks. The following Podfile is an example of the use case that does not work:

```
abstract_target "iOS" do
    platform :ios, "9.0"
    pod "Realm"
    target "iOSApp"
end

abstract_target "tvOS" do
    platform :tvos, "9.0"
    pod "Realm"
    target "tvApp"
end
```

The problem here is that CocoaPods exits with an error `[!] The 'Pods-iOS-XXX' target has libraries with conflicting names: librealm-ios.a`. This is due to the fact that

1. CocoaPods creates a target `libRealm-iOS.a`, scoped for the current platform
2. realm-cocoa ships the static library `core/librealm-ios.a`

Both have the same lowercase case which results in the conflicting names error. My suggestion for a fix is renaming `librealm-*.a` to `librealmcore-*.a` in the `cocoapods-setup` build phase while updating `vendored_library` in `Realm.podspec`. I'm a first time contributor, so I don't know the codebase nor the build process very well.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,7 @@ x.x.x Release notes (yyyy-MM-dd)
 
 ### Bugfixes
 
-* None.
+* Fixing CocoaPods installation with static libraries and multiple platforms.
 
 2.5.0 Release notes (2017-03-28)
 =============================================================
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,7 @@ x.x.x Release notes (yyyy-MM-dd)
 
 ### Bugfixes
 
-* Fixing CocoaPods installation with static libraries and multiple platforms.
+* Fix CocoaPods installation with static libraries and multiple platforms.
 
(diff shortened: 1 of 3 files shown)


jpsim:
@OliverLetterer thanks for the PR! Could you please add a changelog entry in the "Bugfixes" section?

This looks fine to me, actually, although our CocoaPods CI jobs don't run on PRs from forks, so this change isn't being tested right now as well as it should. I'll file another PR with these changes to get that to run.

I'd like @mrackwitz to sign off on it too, if you're available.

OliverLetterer:
@jpsim Thanks for the feedback :) A changelog entry has just been added.

jpsim:
Thanks @OliverLetterer!

PR #4794, opened 2017-03-29
Use reachability support from ObjectStore

stel:
This PR adds linking against `SystemConfiguration.framework` in order to support https://github.com/realm/realm-object-store/pull/289. It's not available on `watchOS`.


Configuration/Realm/Realm.xcconfig
```diff
@@ -27,11 +27,11 @@ LD_RUNPATH_SEARCH_PATHS[sdk=iphone*] = $(inherited) @executable_path/Frameworks
 LD_RUNPATH_SEARCH_PATHS[sdk=watch*] = $(inherited) @executable_path/Frameworks @loader_path/Frameworks;
 LD_RUNPATH_SEARCH_PATHS[sdk=appletv*] = $(inherited) @executable_path/Frameworks @loader_path/Frameworks;
 
-OTHER_LDFLAGS[sdk=macosx*] = -lrealm$(REALM_LIBRARY_SUFFIX);
+OTHER_LDFLAGS[sdk=macosx*] = -lrealm$(REALM_LIBRARY_SUFFIX) -framework SystemConfiguration;
```
bdash: Is there a reason to add these via `OTHER_LDFLAGS` rather than linking to them in the normal way (adding the framework to the Linked Libraries section of the target in the Xcode project)?

One thing that comes to mind is that this means that anyone using the static framework on iOS will have to link to this framework when they add Realm to their project. Perhaps we should be loading this framework at runtime when needed, rather than linking to it statically?

stel: We don't have separate targets for different platforms in Xcode project and watchOS doesn't have this framework so I don't see a way how to specify it there.

Good point about static framework, let me think a bit more about that.

bdash: My bad, I'd forgotten that watchOS is missing reachability.

stel: Related to dynamic loading: according to docs `dlopen` seems to work on iOS8+ while we still support iOS7. I didn't find if it works for system frameworks though. I've also noticed some SO question saying `dlopen` is not supported in app extensions but didn't find any proofs in docs.
Do you know something about that? As a second option we can mention `SystemConfiguration` in docs the same way as `libc++`.

jpsim: You can definitely use `dlopen` with iOS 7, it's just that you can't include dynamic frameworks in your app, which isn't the case here.

stel: Alright, thanks! Then I'll add it to https://github.com/realm/realm-object-store/pull/289 for all supported platforms, so will keep it opened until then.

CHANGELOG.md
```diff
@@ -7,7 +7,8 @@ x.x.x Release notes (yyyy-MM-dd)
 
 ### Enhancements
 
-* None.
+* Use reachability API to minimize the reconnection delay if the network
+connection was lost.
```
jpsim: nitpick: we usually align subsequent lines for lists in the changelog:

```md
   connection was lost.
```

jpsim:
Will merge now to avoid merge conflicts from sneaking up on us. Thanks @stel!

austinzheng:
I need to hook Reachability up into the token refresh subsystem. Will do that shortly...

Issue #4251, opened 2016-10-24
valueForKeyPath returns incorrect results for List properties


jpsim:
``` swift
// Models
class ModelA: Object {
    let bObjects = List<ModelB>()
}
class ModelB: Object {
    dynamic var prop = 0
}
// Failing Test
let realm = try! Realm()
realm.beginWrite()
for value in [1, 2] {
    realm.create(ModelA.self, value: [[[value]]])
}
let aObjects = realm.objects(ModelA.self)
let props = Array(aObjects.flatMap({ $0.bObjects.map({ $0.prop }) }))
let listsOfBObjects = aObjects.value(forKeyPath: "bObjects") as! [List<ModelB>]
let kvcProps = Array(listsOfBObjects.flatMap({ $0.map({ $0.prop }) }))
XCTAssertEqual(props, kvcProps) // fails [1, 2] != [2, 2]
```

tgoyne:
`RLMCollectionValueForKey()` uses a single accessor object and just updates the row. This works fine for obj-c objects and non-generic properties on Swift objects, but for generic properties on Swift objects this results in an array where every entry is the same object as they aren't lazily created like everything else is.

jpsim:
Thanks for clarifying @tgoyne.

hartbit:
I'm affected by this. Took me a while to figure out it was Realm's fault :)

PR #4796, opened 2017-03-29
Fix `valueForKey:`/`valueForKeyPath:` returns incorrect results for `List` properties

kishikawakatsumi:
Fixes https://github.com/realm/realm-cocoa/issues/4251

Modified to create different accessors for each element in the List.

CC @tgoyne @bdash @jpsim

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,7 @@ x.x.x Release notes (yyyy-MM-dd)
 ### Bugfixes
 
 * Fix a deadlock when multiple processes open a Realm at the same time.
+* Fix `value(forKey:)`/`value(forKeyPath:)` returning incorrect values for `List` properties.
 
 2.5.1 Release notes (2017-04-05)
 =============================================================
--- a/Realm/RLMCollection.mm
+++ b/Realm/RLMCollection.mm
@@ -149,12 +149,9 @@ NSArray *RLMCollectionValueForKey(id<RLMFastEnumerable> collection, NSString *ke
         return results;
     }
 
-    RLMObject *accessor = RLMCreateManagedAccessor(info->rlmObjectSchema.accessorClass, realm, info);
-    realm::Table *table = info->table();
     for (size_t i = 0; i < count; i++) {
         size_t rowIndex = [collection indexInSource:i];
-        accessor->_row = (*table)[rowIndex];
-        RLMInitializeSwiftAccessorGenerics(accessor);
+        RLMObjectBase *accessor = RLMCreateObjectAccessor(realm, *info, rowIndex);
         [results addObject:[accessor valueForKey:key] ?: NSNull.null];
     }
 
--- a/Realm/RLMCollection.mm
+++ b/Realm/RLMCollection.mm
@@ -22,6 +22,7 @@
 #import "RLMObjectSchema_Private.hpp"
 #import "RLMObjectStore.h"
 #import "RLMObject_Private.hpp"
+#import "RLMProperty_Private.h"
 
 #import "collection_notifications.hpp"
 #import "list.hpp"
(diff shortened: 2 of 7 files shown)


RealmSwift/Tests/ListTests.swift
```diff
@@ -358,6 +358,25 @@ class ListTests: TestCase {
...
+                object.intArray.append(SwiftIntObject(value: [value]))
+                realm.add(object)
+            }
+        }
+
+        let objects = realm.objects(SwiftArrayPropertyObject.self)
+
+        let properties = Array(objects.flatMap({ $0.intArray.map({ $0.intCol }) }))
+        let listsOfObjects = objects.value(forKeyPath: "intArray") as! [List<SwiftIntObject>]
+        let kvcPropertiess = Array(listsOfObjects.flatMap({ $0.map({ $0.intCol }) }))
```
jpsim: `kvcPropertiess` -> `kvcProperties`

jpsim:
I forgot to report back that I couldn't get the fix while preserving the optimization. Actually, the current code on `master` works as-is on macOS in Debug mode, but not in release mode and never on iOS IIRC.

tgoyne:
Creating a new accessor object for each value is needed if and only if the property we're getting the value of has a non-nil `swiftIvar`. This is because for those properties `valueForKey:` simply returns  an object with a reference to the parent object, while `valueForKey:` on other properties returns a new object which contains the value directly.

I'd be inclined to split these two cases into different loops (like how the `self` case is already a different loop). Note that in the accessor-reuse version calling `RLMInitializeSwiftAccessorGenerics()` is unnecessary, as the point of it is to initialize the properties which accessor reuse doesn't work for.

kishikawakatsumi:
@tgoyne Thanks. I've updated this PR to split into two loops for Swift generic properties and for other properties. `RealmOptional<T>` properties seem to be able to use reused version. In that case, calling `RLMInitializeSwiftAccessorGenerics()` seems necessary. If it doesn't call it, `valueForKey:` for `RealmOptional<T>` properties returns `nil`.

CHANGELOG.md
```diff
@@ -12,7 +12,7 @@ x.x.x Release notes (yyyy-MM-dd)
 
 ### Bugfixes
 
-* None.
+* Fix `valueForKey:`/`valueForKeyPath:` returns incorrect results for `List` properties.
```
jpsim: > Fix `value(forKey:)`/`value(forKeyPath:)` returning incorrect values for `List` properties.

Realm/RLMCollection.mm
```diff
@@ -149,6 +150,17 @@ - (NSUInteger)countByEnumeratingWithState:(NSFastEnumerationState *)state
         return results;
     }
 
+    for (RLMProperty *prop in info->rlmObjectSchema.swiftGenericProperties) {
+        if ([prop.name isEqual:key] && prop.swiftIvar && !prop.optional) {
```
jpsim: Why the optional check? `value(forKey:)` should work with optional properties, and you handle this case a little later with the `?: NSNull.null`.

Also shouldn't need to check for `prop.swiftIvar` since `swiftGenericProperties` checks for the swift ivars already.

In other words, this implementation should be equivalent:

```objc
for (RLMProperty *prop in info->rlmObjectSchema.swiftGenericProperties) {
    if (![prop.name isEqual:key])
        continue;

    for (size_t i = 0; i < count; i++) {
        size_t rowIndex = [collection indexInSource:i];
        RLMObjectBase *accessor = RLMCreateObjectAccessor(realm, *info, rowIndex);
        [results addObject:[accessor valueForKey:key] ?: NSNull.null];
    }
    return results;
}
```

jpsim:
Also, ideally there'd be tests for all sub-property types, not just Int.

Realm/RLMCollection.mm
kishikawakatsumi: Because Swift optional properties can be used accessor reused version. I'd like to use creating a new accessor object for each value version only for `List` and `LinkingObjects`. It is two times slower than the accessor reused version.

kishikawakatsumi: @jpsim Checking `prop.swiftIvar`is unnecessary, I removed it.

jpsim:
This is failing on Xcode 8. I'm installing it locally to debug. I think I can carry this the last mile. Thanks @kishikawakatsumi!

PR #4606, opened 2017-01-30
Improve handling of auth errors

austinzheng:
Changes:
- Added support for multiple auth errors that weren't being handled before
- Fixed nullability annotations on an internal model object
- Auth errors now include the hint in `userInfo`, if one is provided
- Added another test

To do:
- [x] Handle 615 error by logging out user and asking user to refresh.

PR #4799, opened 2017-03-30
Add support for new permissions API

austinzheng:
Changes:
- Add new permissions APIs
- Add tests for new permissions APIs

To do:
- [x] Figure out how to present errors to the user in a less ad-hoc way
- [x] Swift API
- [x] Add even more tests
- [x] Finish docstrings


austinzheng:
Not sure why the sync test is failing on CI. Can't yet reproduce the issue on my machine...

austinzheng:
I got the test failure to reproduce on a machine running Xcode 8.3. Will continue to investigate...

austinzheng:
I think the issue I was seeing on CI was a logic bug with my tests. I fixed the bug and now the test that crashes on CI fails "gracefully" on my machine. I'll continue debugging.

Realm/RLMSyncPermissionResults.h
```diff
@@ -0,0 +1,70 @@
...
+@interface RLMSyncPermissionResultsToken : RLMNotificationToken
+@end
+
+/**
+ An object representing the results of a permissions query.
+
+ Permissions results objects are thread-confined, and should not be shared across
+ threads.
+ */
+@interface RLMSyncPermissionResults : NSObject
```
tgoyne: This class is missing a lot of methods. I would expect it to have all the same functionality as RLMResults other than the aggregate functions.

Realm/RLMSyncPermissionResults.mm
```diff
@@ -0,0 +1,111 @@
...
+#import "RLMUtil.hpp"
+
+using namespace realm;
+
+@interface RLMSyncPermissionResultsToken() {
+    std::unique_ptr<NotificationToken> _token;
+}
+@end
+
+@implementation RLMSyncPermissionResultsToken
```
tgoyne: This subclass seems pointless. It could just use `RLMCancellationToken`.

Realm/RLMSyncPermissionResults.mm
```diff
@@ -0,0 +1,111 @@
...
+        _token = std::make_unique<NotificationToken>(std::move(token));
+        return self;
+    }
+    return nil;
+}
+
+@end
+
+@interface RLMSyncPermissionResults () {
+    std::unique_ptr<PermissionResults> _results;
```
tgoyne: Why is this a `unique_ptr` rather than just storing the PermissionResults directly?

Realm/RLMSyncPermissionResults.mm
```diff
@@ -0,0 +1,111 @@
...
+
+- (RLMSyncPermissionResultsToken *)addNotificationBlock:(RLMPermissionStatusBlock)block {
+    REALM_ASSERT_DEBUG(_results);
+    auto token = _results->async(RLMWrapPermissionStatusCallback(block));
+    return [[RLMSyncPermissionResultsToken alloc] initWithToken:std::move(token)];
+}
+
+- (RLMSyncPermissionValue *)permissionAtIndex:(NSInteger)index {
+    REALM_ASSERT_DEBUG(_results);
+    if (index < 0 || (size_t)index >= _results->size()) {
```
tgoyne: `_results->get()` is already doing this check, and the exception thrown by that should just be translated to an obj-c exception both for consistency and because there are a bunch of other exceptions that could be thrown which also need to be translated.

Realm/RLMSyncPermissionResults.mm
```diff
@@ -0,0 +1,111 @@
...
+
+- (instancetype)initWithResults:(std::unique_ptr<PermissionResults>)results {
+    if (self = [super init]) {
+        REALM_ASSERT_DEBUG(results);
+        _results = std::move(results);
+    }
+    return self;
+}
+
+- (NSString *)description {
```
tgoyne: This code is whacky and different from the existing collection description methods for no apparent reason.

Realm/RLMSyncPermissions.h
```diff
@@ -0,0 +1,108 @@
...
+- (instancetype)initWithRealmPath:(NSString *)path
+                           userID:(NSString *)userID
+                      accessLevel:(RLMSyncAccessLevel)accessLevel;
+
+/**
+ Create a new sync permission value, for use with permission APIs.
+
+ @param path            The path to the Realm whose permission should be modified (e.g. "/path/to/realm")
+ @param key             A key describing a property used to evaluate the set of users to whom the permission applies
+ @param value           A filtering criteria on the `key` property determining who the permission applies to
```
tgoyne: After reading all of the documentation here on KeyValue permissions I still don't have the slightest clue what they are, what objects the properties are on that the keys apparently apply to or what sort of values are valid "filtering criteria".

Realm/RLMSyncPermissions.mm
```diff
@@ -0,0 +1,307 @@
...
+        return *permission;
+    }
+    return Permission{
+        [_path UTF8String],
+        accessLevelForObjcAccessLevel(_accessLevel),
+        Permission::Condition([_userID UTF8String])
+    };
+}
+
+- (BOOL)isEqual:(id)object {
```
tgoyne: `hash` must always be overridden when `isEqual:` is.

Realm/RLMSyncUser_Private.hpp
```diff
@@ -46,4 +46,12 @@ NS_ASSUME_NONNULL_BEGIN
 
 @end
 
+namespace realm {
```
tgoyne: Why is this in namespace realm? None of the cocoa code should be.

Realm/RLMSyncUser.mm
```diff
@@ -31,9 +35,67 @@
 #import "sync/sync_user.hpp"
 
 using namespace realm;
+using ConfigMaker = std::function<Realm::Config(std::shared_ptr<SyncUser>, std::string)>;
+using PermissionGetCallback = std::function<void(std::unique_ptr<PermissionResults>, std::exception_ptr)>;
+
+namespace {
+
+NSError *RLMTranslateExceptionPtrToError(std::exception_ptr ptr, bool get)
```
tgoyne: No RLM prefix for local functions, and no newline before {.

Realm/RLMSyncPermissions.h
austinzheng: This was an oversight on my part. We had designed some object store APIs to be implemented in the future, and these Cocoa APIs should never have made it into the review.

Realm/RLMSyncPermissionResults.h
```diff
@@ -0,0 +1,103 @@
...
+ @note Valid properties to filter on are `path` and `userId`, as well as
+       the boolean properties `mayRead`, `mayWrite`, and `mayManage`.
+ */
+- (RLMSyncPermissionResults *)objectsWithPredicate:(NSPredicate *)predicate;
+
+/**
+ Return a sorted `RLMSyncPermissionResults` from the collection, sorted based on
+ the given property.
+ */
+- (RLMSyncPermissionResults *)sortedResultsUsingProperty:(RLMSyncPermissionResultsSortProperty)property
```
austinzheng: Suggestions for improving this API welcome.

Realm/RLMSyncPermissionResults.h
```diff
@@ -0,0 +1,106 @@
...
+
+@class RLMSyncPermissionValue;
+
+/**
+ An object representing the results of a permissions query.
+
+ Permissions results objects are thread-confined, and should not be shared across
+ threads.
+ */
+@interface RLMSyncPermissionResults : NSObject<NSFastEnumeration>
```
jpsim: This needs `Sequence` support in `RLMSupport.swift`

Realm/RLMSyncUser.h
```diff
@@ -118,6 +131,39 @@ NS_SWIFT_UNAVAILABLE("Use the full version of this API.");
...
+
+ The operation will take place asynchronously, and the callback will be used to report whether
+ the permission change succeeded or failed. The user upon which this method is called must have
+ the right to grant the given permission, or else the oepration will fail.
+
+ @see `RLMSyncPermissionValue`
+ */
+- (void)revokePermission:(RLMSyncPermissionValue *)permission callback:(RLMPermissionStatusBlock)callback;
+
+#pragma mark - Permissions (old)
```
jpsim: We should come up with better wording for deprecated permissions APIs than simply "old".

jpsim: We should also update the documentation of the APIs that we wish to deprecate.

Realm/RLMSyncPermissionResults.h
```diff
@@ -0,0 +1,106 @@
...
+    /// Sort by the path of the Realm the permission pertains to.
+    RLMSyncPermissionResultsSortPropertyPath,
+    /// Sort by the user ID of the user the permission applies to.
+    RLMSyncPermissionResultsSortPropertyUserID,
+} RLMSyncPermissionResultsSortProperty;
+
+@class RLMSyncPermissionValue;
+
+/**
+ An object representing the results of a permissions query.
```
jpsim: We should state that `RLMSyncPermissionResults` is an auto-updating collection but its values are static.

Realm/RLMSyncPermissions_Private.hpp
```diff
@@ -0,0 +1,41 @@
...
+
+- (instancetype)initWithPermission:(realm::Permission)permission;
+
+- (realm::Permission)rawPermission;
+
+@end
+
+// `RLMSyncPermissionValue` is subclassed so that future types of
+// permission values can be implemented in the future.
+@interface RLMSyncUserIDPermissionValue : RLMSyncPermissionValue
```
jpsim: Do we really need to subclass `RLMSyncPermissionValue`? I don't see what that gains us.

jpsim:
Also: since `RLMSyncPermissionResults` is thread-confined, should we support wrapping it in a `RLMThreadSafeReference`?

austinzheng:
I also reworked the way sync error handling works.

Now errors are constructed in a much less ad-hoc way, and the keys they use don't differ across error types. I plan to pull in some of the changes that were introduced in #4606 in a follow-up PR; that will complete the sync error refactoring.

This is sort of a breaking change, but given that the error handling system was more or less useless before (and the errors that users are intended to interact with haven't changed), I think it's fine. (Same reasoning as described in #4606.)

austinzheng:
My inclination is that we probably don't need `RLMThreadSafeReference` support for the initial version.

Marshalling and unmarshalling permission results into thread safe references to move them across thread boundaries seems like a lot of work when the sync user (which can be passed across threads by default) can simply be passed and the permissions obtained from the destination thread. If it turns out people do want this capability (and this API is the correct long term direction), we can include support in a completely additive manner.

Realm/ObjectServerTests/RLMObjectServerTests.mm
```diff
@@ -17,17 +17,13 @@
...
 
 #import "RLMSyncTestCase.h"
+#import "RLMTestUtils.h"
 #import "RLMSyncSessionRefreshHandle+ObjectServerTests.h"
 #import "RLMSyncUser+ObjectServerTests.h"
 #import "RLMSyncUtil_Private.h"
 #import "RLMRealmConfiguration_Private.h"
 #import "RLMRealmUtil.hpp"
 
-#define ACCOUNT_NAME() NSStringFromSelector(_cmd)
```
jpsim: Why remove this helper? It hurts code readability and makes this patch larger for no apparent reason.

CHANGELOG.md
```diff
@@ -9,6 +9,8 @@ x.x.x Release notes (yyyy-MM-dd)
 
 * Use reachability API to minimize the reconnection delay if the network
   connection was lost.
+* Add a new API for changing and retrieving permissions. This API can be used
```
jpsim: If the intention is for these APIs to replace the existing ones, we should use stronger wording:

```md
* Add a new API for changing and retrieving permissions for synchronized Realms.
  This API is intended to replace the existing Realm Object-based permissions
  system.
```

Realm/RLMSyncPermissionResults.h
```diff
@@ -0,0 +1,109 @@
...
+    /// Sort by the path of the Realm the permission pertains to.
+    RLMSyncPermissionResultsSortPropertyPath,
+    /// Sort by the user ID of the user the permission applies to.
+    RLMSyncPermissionResultsSortPropertyUserID,
+};
+
+@class RLMSyncPermissionValue;
+
+/**
+ An object representing the results of a permissions query.
```
jpsim:
```objc
/**
 A collection object representing the results of a permissions query.

 This collection will automatically update its contents at the start of each runloop
 iteration, but the objects it vends are immutable and will themselves not change.

 Permission results objects are thread-confined, and should not be shared across
 threads.
 */
```

Realm/Swift/RLMSupport.swift
```diff
@@ -80,3 +80,22 @@ extension RLMCollection {
         return objects(with: NSPredicate(format: predicateFormat, arguments: getVaList(args)))
     }
 }
+
+// Sequence conformance for RLMSyncPermissionResults.
+extension RLMSyncPermissionResults: Sequence {
+    public struct Iterator: IteratorProtocol {
+        let iteratorBase: NSFastEnumerationIterator
```
jpsim: please use an explicit ACL keyword: `private let iteratorBase`.

Realm/Swift/RLMSupport.swift
```diff
@@ -80,3 +80,22 @@ extension RLMCollection {
...
+
+        init(results: RLMSyncPermissionResults) {
```
jpsim: please use an explicit ACL keyword: `fileprivate init(results:...)`.

Realm/ObjectServerTests/RLMObjectServerTests.mm
```diff
@@ -823,8 +819,8 @@ - (void)testLogBackInReopenRealmDownload {
 /// Ensure that a client reset error is propagated up to the binding successfully.
 - (void)testClientReset {
     NSURL *url = REALM_URL();
-    NSString *sessionName = ACCOUNT_NAME();
-    RLMSyncUser *user = [self logInUserForCredentials:[RLMObjectServerTests basicCredentialsWithName:ACCOUNT_NAME()
+    NSString *sessionName = NSStringFromSelector(_cmd);
+    RLMSyncUser *user = [self logInUserForCredentials:[RLMObjectServerTests basicCredentialsWithName:NSStringFromSelector(_cmd)
```
jpsim: reuse `sessionName` here?

Realm/ObjectServerTests/RLMObjectServerTests.mm
```diff
@@ -852,8 +848,8 @@ - (void)testClientReset {
 /// Test manually initiating client reset.
 - (void)testClientResetManualInitiation {
     NSURL *url = REALM_URL();
-    NSString *sessionName = ACCOUNT_NAME();
-    RLMSyncUser *user = [self logInUserForCredentials:[RLMObjectServerTests basicCredentialsWithName:ACCOUNT_NAME()
+    NSString *sessionName = NSStringFromSelector(_cmd);
+    RLMSyncUser *user = [self logInUserForCredentials:[RLMObjectServerTests basicCredentialsWithName:NSStringFromSelector(_cmd)
```
jpsim: reuse `sessionName` here?

Realm/RLMSyncUser.mm
```diff
@@ -73,6 +127,13 @@ - (instancetype)initWithAuthServer:(nullable NSURL *)authServer {
     if (self = [super init]) {
         self.authenticationServer = authServer;
         self.refreshHandles = [NSMutableDictionary dictionary];
+        _configMaker = std::make_unique<ConfigMaker>([](std::shared_ptr<SyncUser> user, std::string url) {
+            RLMRealmConfiguration *config = [[RLMRealmConfiguration defaultConfiguration] copy];
```
jpsim: `[RLMRealmConfiguration defaultConfiguration]` already returns a copy.

Realm/RLMSyncPermissionValue.mm
```diff
@@ -0,0 +1,203 @@
...
+        _path = path;
+        _userID = userID;
+    }
+    return self;
+}
+
+- (instancetype)initWithPermission:(Permission)permission {
+    switch (permission.condition.type) {
+        case ConditionType::UserId:
+            self = [[RLMSyncPermissionValue alloc] initPrivate];
```
jpsim: Can't use you just call `self = [super init]` here and get rid of `initPrivate` altogether?

Realm/ObjectServerTests/RLMObjectServerTests.mm
austinzheng: Thomas expressed the opinion that the macro was needlessly obfuscating in his earlier review. Happy to change it back if you think the readability benefit makes it worth keeping.

jpsim: no strong preference I guess.

Realm/RLMSyncPermissionValue.mm
austinzheng: Yeah, that was cruft left over from when the subclass still existed. Thanks for pointing it out.

Realm/RLMSyncUser.h
```diff
@@ -124,6 +137,46 @@ NS_SWIFT_UNAVAILABLE("Use the full version of this API.");
...
+
+ The results will be returned through the callback block, or an error if the operation fails.
+ */
+- (void)retrievePermissionsWithCallback:(RLMPermissionResultsBlock)callback;
+
+/**
+ Apply a given permission.
+
+ The operation will take place asynchronously, and the callback will be used to report whether
+ the permission change succeeded or failed. The user upon which this method is called must have
```
jpsim: "The user calling this method must have..."

Realm/RLMSyncUser.h
```diff
@@ -124,6 +137,46 @@ NS_SWIFT_UNAVAILABLE("Use the full version of this API.");
...
+ the right to grant the given permission, or else the oepration will fail.
```
jpsim: "operation"

Realm/RLMSyncUser.h
```diff
@@ -124,6 +137,46 @@ NS_SWIFT_UNAVAILABLE("Use the full version of this API.");
...
+ @see `RLMSyncPermissionValue`
+ */
+- (void)applyPermission:(RLMSyncPermissionValue *)permission callback:(RLMPermissionStatusBlock)callback;
+
+/**
+ Revoke a given permission.
+
+ The operation will take place asynchronously, and the callback will be used to report whether
+ the permission change succeeded or failed. The user upon which this method is called must have
+ the right to grant the given permission, or else the oepration will fail.
```
jpsim: "operation"

Realm/RLMSyncUser.h
```diff
@@ -124,6 +137,46 @@ NS_SWIFT_UNAVAILABLE("Use the full version of this API.");
...
+
+ @see `RLMSyncPermissionValue`
+ */
+- (void)applyPermission:(RLMSyncPermissionValue *)permission callback:(RLMPermissionStatusBlock)callback;
+
+/**
+ Revoke a given permission.
+
+ The operation will take place asynchronously, and the callback will be used to report whether
+ the permission change succeeded or failed. The user upon which this method is called must have
```
jpsim: "The user calling this method must have..."

Realm/RLMSyncUser.h
```diff
@@ -124,6 +137,46 @@ NS_SWIFT_UNAVAILABLE("Use the full version of this API.");
...
+ the right to grant the given permission, or else the oepration will fail.
+
+ @see `RLMSyncPermissionValue`
+ */
+- (void)revokePermission:(RLMSyncPermissionValue *)permission callback:(RLMPermissionStatusBlock)callback;
+
+// This set of permissions APIs uses a set of Realms and Realm model objects
```
jpsim: "These permissions APIs access Realms and Realm model objects"

CHANGELOG.md
jpsim: If we do this, the API should reflect the deprecations.

austinzheng: Do we want these APIs to replace the existing ones? I don't feel comfortable making a unilateral decision about whether we should commit to deprecation.

jpsim: It's the whole motivation behind these new APIs.

RealmSwift/Sync.swift
```diff
@@ -276,6 +276,47 @@ extension SyncUser {
...
+public typealias SyncAccessLevel = RLMSyncAccessLevel
+
+/**
+ A sequence of results that contains the permissions for Realms
+
+ - see: `RLMSyncPermissionResults`
+ */
+public typealias SyncPermissionResults = RLMSyncPermissionResults
+
+extension SyncPermissionResults: Sequence {
```
jpsim: Can you add conformance to the `Collection` protocol?

Realm/RLMSyncUtil_Private.h
```diff
@@ -22,6 +22,17 @@
 #import <Realm/RLMRealmConfiguration.h>
 #import <Realm/RLMSyncCredentials.h>
 
+typedef NS_ENUM(NSUInteger, RLMSyncSystemErrorKind) {
```
jpsim: Since some errors with codes in this enum can be reported to users via the public API, does it make sense to make this enum public as well?

Realm/RLMSyncUtil.mm
```diff
@@ -112,8 +130,81 @@ RLMSyncStopPolicy translateStopPolicy(SyncSessionStopPolicy stop_policy)
...
+    return make_permission_error(description, std::move(code), false);
+}
+
+NSError *make_sync_error(RLMSyncSystemErrorKind kind, NSString *description, NSInteger code, NSDictionary *custom) {
+    NSDictionary *userInfo;
+    if (custom) {
+        NSMutableDictionary *buffer = [custom mutableCopy];
+        buffer[NSLocalizedDescriptionKey] = description;
+        buffer[kRLMSyncErrorStatusCodeKey] = @(code);
+        userInfo = buffer;
```
jpsim: `userInfo = [buffer copy];`?

Realm/RLMSyncUtil.h
```diff
@@ -29,34 +29,30 @@ extern NSString *const kRLMSyncPathOfRealmBackupCopyKey;
...
+ */
+extern NSString *const RLMSyncPermissionErrorDomain;
+
+/**
+ An error related to a problem that might be reported by the synchronization manager
+ error handler, or a callback on a sync-related API that performs asynchronous work.
+ */
 typedef RLM_ERROR_ENUM(NSInteger, RLMSyncError, RLMSyncErrorDomain) {
-    /// An error that indicates that the response received from the authentication server was malformed.
-    RLMSyncErrorBadResponse             = 1,
```
jpsim: Perhaps it'd be worth keeping this member around, but deprecating it, as to avoid unnecessary code breakages?

Realm/RLMSyncPermissionResults.mm
```diff
@@ -0,0 +1,160 @@
...
+    auto sorted_results = std::make_unique<PermissionResults>(_results->sort({
+        table, {{ col_idx }}, { static_cast<bool>(ascending) }
+    }));
+    return [[RLMSyncPermissionResults alloc] initWithResults:std::move(sorted_results)];
+}
+
+- (NSUInteger)countByEnumeratingWithState:(NSFastEnumerationState *)state
+                                  objects:(id __unsafe_unretained [])buffer
+                                    count:(NSUInteger)len {
+    NSUInteger thisSize = self.count;
```
jpsim: This fast enumeration logic is pretty hairy. Is there any value in sharing some of this code with the RLMArray/RLMResults/RLMCollection/RLMArrayLinkView implementations?

Realm/RLMSyncPermissionResults.h
```diff
@@ -0,0 +1,109 @@
...
+ */
+@interface RLMSyncPermissionResults : NSObject<NSFastEnumeration>
+
+/// The number of results contained within the object.
+@property (nonatomic, readonly) NSInteger count;
+
+/**
+ Return the first permission, or nil if the collection is empty.
+ */
+- (nullable RLMSyncPermissionValue *)firstObject;
```
jpsim: Since this returns a "value type" and not a Realm Object, is `firstObject` the right name for this method? Same goes for `lastObject` and other methods on this type with the word "object" in it.

Also, we translate `firstObject`/`lastObject` to simply `first`/`last` in Swift, and I don't think that's being done here.

jpsim: BTW, I personally think it's fine to refer to these as "objects", but it seemed relevant to bring up.

tgoyne: The terminology overloading here is a little unfortunate, but I don't think it's worth breaking with NSArray/etc.'s naming.

Realm/RLMSyncPermissionResults.mm
austinzheng: That was my initial goal, but the existing enumeration machinery and helper class are too tightly integrated into `RLMObject` to make it worthwhile.

Realm/RLMSyncUtil.mm
austinzheng: Good call.

Realm/RLMSyncUtil_Private.h
austinzheng: I'm pretty sure that this enum is internal only. There should be explicit codes for the various kinds of errors (we don't map precisely from the object store categories, which is what this enum represents, to the user-facing ones for a couple of reasons).

dhmspector:
Apropos of working with the API with @austinzheng  earlier:

1. Need to make sure the requirement for main thread execution of the `retrievePermissions()` call is clearly documented (took about 10 mins of Austin's time before it was clear why no results came back when using this inside the `SyncUser.logIn` callback).
2. The fact that a user has access to 1 or more realms via a wildcard permission setting needs to be return as part of the retrievePermissions()`  result set... right now no information about such realms is returned.
3. The linear nature of the permission structure needs to be very clearly documented and explained - many Unix developers will expect that these to function like Unix style permissions when they do not.

Realm/RLMSyncUtil_Private.h
jpsim: `asyncOpen` can yield an error with an `RLMSyncSystemErrorKind` if the error is produced by the sync subsystem.

jpsim:
Thanks for the feedback @dhmspector!

> 1. Need to make sure the requirement for main thread execution of the `retrievePermissions()` call is clearly documented (took about 10 mins of Austin's time before it was clear why no results came back when using this inside the `SyncUser.logIn` callback).

Seems like we should adapt this so you can _call_ this function from any thread, and the implementation would call out to `Permissions::get_permissions` and the callback block would be invoked on the main thread. @austinzheng what do you think?

> 2. The fact that a user has access to 1 or more realms via a wildcard permission setting needs to be return as part of the `retrievePermissions()`  result set... right now no information about such realms is returned.

This scenario will need to be spelled out for me. Are you saying that the Permission Value returned will have a `realmUrl` value of `*` but you'd like a bunch of values to be returned, one for each `realmUrl`?

> 3. The linear nature of the permission structure needs to be very clearly documented and explained - many Unix developers will expect that these to function like Unix style permissions when they do not. 

A comment is definitely needed, @austinzheng you can pull from the discussions linked to from realm/product#300.

dhmspector:
>> 2. The fact that a user has access to 1 or more realms via a wildcard permission setting needs to be return as part of the `retrievePermissions()`  result set... right now no information about such realms is returned.

>This scenario will need to be spelled out for me. Are you saying that the Permission Value returned will have a `realmUrl` value of `*` but you'd like a bunch of values to be returned, one for each `realmUrl`?


Other way around - for any Realms for which there is wildcard user access I would like to know that I have access to those Realms and what permissions those wildcards contain.  Basically right now nothing is returned for any realms you have access to where a wildcard user is set - so yes an enumeration of the permissions for each realm that you have wildcard access to....

So if a permission exists like: 

wilcardperms

I would except to see something like...
```
Results: RLMSyncPermissionResults <0x60800000ad70> (
	[0] <RLMSyncPermissionValue> user ID: 6abcee96b1335b690234e95e6ae7f975, path: /Teamwork-MR/ManagerRealm, access level: read
	[1] <RLMSyncPermissionValue> user ID: 6abcee96b1335b690234e95e6ae7f975, path: /Teamwork-MR/CommonRealm, access level: read
	[2] <RLMSyncPermissionValue> user ID: 6abcee96b1335b690234e95e6ae7f975, path: /Teamwork-MR/CommonRealm, access level: write
)
```

Is that helpful?

austinzheng:
> Seems like we should adapt this so you can call this function from any thread, and the implementation would call out to Permissions::get_permissions and the callback block would be invoked on the main thread.

Because of the way the object store implementation works, we'd have to just dispatch onto the main thread within `retrievePermissionsWithCallback:` itself (since the callback itself can't be re-dispatched onto a different thread, given that it's passed in something that depends on a thread-confined `Results`). It's also pretty different from how both notifications and results work right now: the former expects the caller to take care of ensuring a runloop; the latter presents the results on the same thread as the caller.

I honestly think the better solution at this point is to change `logIn` to dispatch its block back onto the main thread. This is what users want 99% of the time, it fixes a bunch of other issues that manifest if you ignore or accidentally overlook the documentation, and if they really want a background queue they can re-dispatch manually. The performance argument is probably spurious since a user log-in event is extremely infrequent.

Realm/RLMSyncUtil_Private.h
austinzheng: I just double-checked `asyncOpen`. It can only either return a regular failed-to-open-Realm error, or a sync error created by my `make_sync_error()` helper. We do pass in the error kind, but the function translates it and does other stuff, and doesn't put it into the final `NSError`.

jpsim:
> I honestly think the better solution at this point is to change logIn to dispatch its block back onto the main thread. This is what users want 99% of the time, it fixes a bunch of other issues that manifest if you ignore or accidentally overlook the documentation, and if they really want a background queue they can re-dispatch manually. The performance argument is probably spurious since a user log-in event is extremely infrequent.

See #4338, feel free to extend on that. We have to consider changing what queue/thread invokes a callback as a breaking change, so please keep that in mind.

Realm/RLMSyncManager.mm
```diff
@@ -150,47 +151,31 @@ - (void)_fireErrorWithCode:(int)errorCode
...
-    mutableUserInfo[@"underlying_class"] = @(errorClass);
-
+    BOOL shouldMakeError = YES;
+    NSDictionary *custom = nil;
     switch (errorClass) {
         case RLMSyncSystemErrorKindClientReset: {
-            // Client reset is a special case; the application can respond to it to a greater degree than
-            // it can for most other errors.
-            mutableUserInfo[kRLMSyncPathOfRealmBackupCopyKey] = userInfo[@(realm::SyncError::c_recovery_file_path_key)];
+            // Client reset is a special case; the application
```
jpsim: Wording is a bit awkward here. Consider rewording:

```
// Users can respond to "client reset" errors to a
// greater degree than it can for most other errors.
```

Realm/RLMSyncPermissionValue.h
```diff
@@ -0,0 +1,97 @@
...
+    /// granting permissions to other users.
+    RLMSyncAccessLevelAdmin,
+};
+
+NS_ASSUME_NONNULL_BEGIN
+
+/**
+ A value representing a permission granted to one or more users with respect to a particular Realm.
+
+ `RLMSyncPermissionValue` is immutable and thread-safe.
```
jpsim: Should we also link to general permissions docs like we do for `RLMSyncPermission`?

> See https://realm.io/docs/realm-object-server/#permissions for general documentation.

Realm/RLMSyncPermissionValue.h
```diff
@@ -0,0 +1,97 @@
...
+/// Whether the access level allows the user to write to the Realm.
+@property (nonatomic, readonly) BOOL mayWrite;
+
+/// Whether the access level allows the user to administer the Realm.
+@property (nonatomic, readonly) BOOL mayManage;
+
+/**
+ Create a new sync permission value, for use with permission APIs.
+
+ @param path            The path to the Realm whose permission should be modified (e.g. "/path/to/realm")
```
jpsim: "The Realm Object Server path..."?

Realm/RLMSyncPermissionValue.h
```diff
@@ -0,0 +1,97 @@
...
+ @param userID          The user ID of the user who should be granted access to the Realm at `path`
```
jpsim: We generally terminate `@param` text with periods.

Realm/RLMSyncPermissionValue.h
```diff
@@ -0,0 +1,97 @@
...
+ @param accessLevel     What access level to grant
```
jpsim: "The access level to grant"?

Realm/RLMSyncPermissionValue.h
```diff
@@ -0,0 +1,97 @@
...
+/// Whether the access level allows the user to read from the Realm.
+@property (nonatomic, readonly) BOOL mayRead;
+
+/// Whether the access level allows the user to write to the Realm.
+@property (nonatomic, readonly) BOOL mayWrite;
+
+/// Whether the access level allows the user to administer the Realm.
+@property (nonatomic, readonly) BOOL mayManage;
+
+/**
```
jpsim: Suggested edits:

```objc
/**
 Create a new sync permission value, for use with permission APIs.

 @param path        The Realm Object Server path to the Realm whose permission should be modified
                    (e.g. `/path/to/realm`). Pass `*` to apply to all Realms managed by the server.
 @param userID      The identity of the user who should be granted access to the Realm at `path`.
                    Pass `*` to apply to all users managed by the server.
 @param accessLevel The access level to grant.
 */
```

Realm/RLMSyncPermissionResults.h
```diff
@@ -0,0 +1,112 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+////////////////////////////////////////////////////////////////////////////
+
+#import <Foundation/Foundation.h>
+
+#import "RLMSyncUser.h"
+#import "RLMRealm.h"
```
jpsim: Drop this import and forward-declare `@class RLMNotificationToken`.

Realm/RLMSyncPermissionResults.h
```diff
@@ -0,0 +1,112 @@
...
+
+NS_ASSUME_NONNULL_BEGIN
+
+/// Properties which a sync permission results collection can be sorted by.
+typedef NS_ENUM(NSUInteger, RLMSyncPermissionResultsSortProperty) {
+    /// Sort by the path of the Realm the permission pertains to.
```
jpsim: "Sort by the Realm Object Server path to the Realm to which the permission applies."?

Realm/RLMSyncPermissionResults.h
```diff
@@ -0,0 +1,112 @@
...
+    RLMSyncPermissionResultsSortPropertyPath,
+    /// Sort by the user ID of the user the permission applies to.
```
jpsim: "Sort by the identity of the user to whom the permission applies."?

Realm/RLMSyncPermissionResults.h
```diff
@@ -0,0 +1,112 @@
...
+    RLMSyncPermissionResultsSortPropertyUserID,
+};
+
+@class RLMSyncPermissionValue;
+
+/**
+ A collection object representing the results of a permissions query.
+
+ This collection will automatically update its contents at the start of each runloop
+ iteration, but the objects it vends are immutable and will themselves not change.
```
jpsim: Drop " and will themselves not change" please.

Realm/RLMSyncPermissionResults.h
```diff
@@ -0,0 +1,112 @@
...
+ */
+@interface RLMSyncPermissionResults : NSObject<NSFastEnumeration>
+
+/// The number of results contained within the object.
+@property (nonatomic, readonly) NSInteger count;
+
+/**
+ Return the first permission, or nil if the collection is empty.
+ */
+- (nullable RLMSyncPermissionValue *)firstObject NS_SWIFT_UNAVAILABLE("Use the `first` property.");
```
jpsim: Since these are marked as `NS_SWIFT_UNAVAILABLE`, maybe we should re-expose them in `RLMSupport.swift`? This applies to several of our existing APIs btw.

Realm/RLMSyncPermissionResults.h
```diff
@@ -0,0 +1,112 @@
...
+- (RLMSyncPermissionValue *)objectAtIndex:(NSInteger)index;
+
+/**
+ Returns the index of the permission in the collection, or `NSNotFound` if the permission
+ is not found in the collection.
+ */
+- (NSInteger)indexOfObject:(RLMSyncPermissionValue *)object;
+
+/**
+ Register a notification block upon the results object. The block will be called
```
jpsim: "Register to be notified when the contents of the results object change."?

Realm/RLMSyncPermissionValue.h
```diff
@@ -0,0 +1,97 @@
...
+};
+
+NS_ASSUME_NONNULL_BEGIN
+
+/**
+ A value representing a permission granted to one or more users with respect to a particular Realm.
+
+ `RLMSyncPermissionValue` is immutable and thread-safe.
+ */
+@interface RLMSyncPermissionValue : NSObject
```
jpsim: `RLMSyncPermission` exposes an `updatedAt` date property. This is useful to filter or sort by, for example. Why is that not exposed here?

Realm/RLMSyncPermissionResults.mm
```diff
@@ -0,0 +1,160 @@
...
+                                               ascending:(BOOL)ascending {
+    REALM_ASSERT_DEBUG(_results);
+    auto& results = _results->results();
+    std::string property_name;
+    switch (property) {
+        case RLMSyncPermissionResultsSortPropertyPath:
+            property_name = "path";
+            break;
+        case RLMSyncPermissionResultsSortPropertyUserID:
+            property_name = "user_id";
```
jpsim: are you sure this is `user_id`? Looks like it's actually `userId` according to this: https://github.com/realm/realm-object-store/blob/956bc716aab38719f5a1410a535f14eebf4e7038/src/sync/sync_permission.cpp#L234

Realm/RLMSyncPermissionValue.h
```diff
@@ -0,0 +1,97 @@
...
+
+/**
+ The path to the Realm to which this permission applies.
```
jpsim: > The Realm Object Server path to the Realm to which this permission applies (e.g. `/path/to/realm`).
> `*` if this permission applies to all Realms managed by the server.

Realm/RLMSyncPermissionValue.h
```diff
@@ -0,0 +1,97 @@
...
+ @param path            The path to the Realm whose permission should be modified (e.g. "/path/to/realm")
+ @param userID          The user ID of the user who should be granted access to the Realm at `path`
+ @param accessLevel     What access level to grant
+ */
+- (instancetype)initWithRealmPath:(NSString *)path
+                           userID:(NSString *)userID
+                      accessLevel:(RLMSyncAccessLevel)accessLevel;
+
+/**
+ The user ID of the user to whom these permissions are granted.
```
jpsim: > The identity of the user to whom this permission is granted.
> `*` if all users are granted this permission.

Realm/RLMSyncUser.h
```diff
@@ -35,6 +35,15 @@ typedef NS_ENUM(NSUInteger, RLMSyncUserState) {
 /// A block type used for APIs which asynchronously vend an `RLMSyncUser`.
 typedef void(^RLMUserCompletionBlock)(RLMSyncUser * _Nullable, NSError * _Nullable);
 
+/// A block type used to report the progress of a permissions set operation.
```
jpsim: "A block type used to report the progress of a permission apply or revoke operation."?

Realm/RLMSyncUser.h
```diff
@@ -124,21 +137,71 @@ NS_SWIFT_UNAVAILABLE("Use the full version of this API.");
  */
 - (NSArray<RLMSyncSession *> *)allSessions;
 
+// This set of permissions APIs uses immutable `RLMSyncPermissionValue` objects to
+// get and set permissions. It is intended to replace the set of APIs which directly
```
jpsim: "retrieve and apply permissions"?

Realm/RLMSyncUser.h
```diff
@@ -124,21 +137,71 @@ NS_SWIFT_UNAVAILABLE("Use the full version of this API.");
...
+// access Realms and Realm model objects to work with permissions.
+#pragma mark - Value-based Permissions API
+
+/**
+ Asynchronously retrieve all permissions associated with the user calling this method.
+
+ The results will be returned through the callback block, or an error if the operation fails.
```
jpsim: stick with the same tense: "failed".

Realm/RLMSyncUser.h
```diff
@@ -124,21 +137,71 @@ NS_SWIFT_UNAVAILABLE("Use the full version of this API.");
...
+ The callback block will be run on the same thread the method was called on.
+
+ @warning This method must be called upon a thread with a currently active run loop. Unless
```
jpsim: "from a thread with an active run loop".

Realm/RLMSyncPermissionResults.h
austinzheng: The `first` and `last` APIs are given to free to anyone who conforms to `Collection`.

Realm/RLMSyncUtil.mm
```diff
@@ -112,8 +132,81 @@ RLMSyncStopPolicy translateStopPolicy(SyncSessionStopPolicy stop_policy)
...
+NSError *make_permission_error_get(NSString *description, util::Optional<NSInteger> code) {
+    return make_permission_error(description, std::move(code), true);
+}
+
+NSError *make_permission_error_change(NSString *description, util::Optional<NSInteger> code) {
+    return make_permission_error(description, std::move(code), false);
+}
+
+NSError *make_sync_error(RLMSyncSystemErrorKind kind, NSString *description, NSInteger code, NSDictionary *custom) {
+    NSDictionary *userInfo;
```
austinzheng: Yes, there's really nothing that `kind` gets you that passing the underlying error message and string don't do better.

jpsim: Ok. I think it's fair to assume no one was relying on that.

Realm/RLMSyncUser.h
```diff
@@ -124,21 +137,71 @@ NS_SWIFT_UNAVAILABLE("Use the full version of this API.");
...
+#pragma mark - Realm Object-based Permissions API
+
 /**
  Returns an instance of the Management Realm owned by the user.
 
  This Realm can be used to control access permissions for Realms managed by the user.
  This includes granting other users access to Realms.
  */
-- (RLMRealm *)managementRealmWithError:(NSError **)error NS_REFINED_FOR_SWIFT;
+- (RLMRealm *)managementRealmWithError:(NSError **)error __deprecated_msg("Only use the management Realm for permission offers and responses.") NS_REFINED_FOR_SWIFT;
```
austinzheng: This is sort of a hack to get Realm Swift to compile without warnings, since the Swift method to open the management Realm accesses the deprecated classes when setting the schema (and hence, annoyingly, propagates the deprecated warning). I can remove it if you prefer, though.

jpsim: I think we can use `NSClassFromString` in the Swift implementation to avoid the deprecation warnings and therefore drop the deprecation annotations in both the Objc & Swift versions of this method.

Realm/RLMSyncUser.h
jpsim: Actually, "progress" is the wrong word here as the operation has always _completed_ when the callback is invoked, it just may or may not have succeeded. So instead:

"A block type used to report the status of a permissions set operation."

Realm/RLMSyncPermissionResults.mm
austinzheng: Good catch.

Realm/RLMSyncPermissionValue.mm
```diff
@@ -0,0 +1,201 @@
...
+            return RLMSyncAccessLevelRead;
+        case Permission::AccessLevel::Write:
+            return RLMSyncAccessLevelWrite;
+        case Permission::AccessLevel::Admin:
+            return RLMSyncAccessLevelAdmin;
+    }
+    REALM_UNREACHABLE();
+}
+
+NSString *descriptionForAccessLevel(RLMSyncAccessLevel level) {
```
austinzheng: I would say yes for the first two, no for the second two (since Cocoa needs to compare both "output" permissions (those backed by the object store) and "input" permissions (those whose state is set using ivars)).

austinzheng:
To do:
- [ ] Address https://github.com/realm/realm-cocoa/pull/4799#discussion_r113789817
- [ ] Address https://github.com/realm/realm-cocoa/pull/4799#discussion_r113793406

austinzheng:
All outstanding issues have now been addressed. Awaiting final review early next week...

Realm/RLMSyncPermissionResults.h
```diff
@@ -0,0 +1,112 @@
...
+
+#pragma mark - Queries
+
+/**
+ Return all permissions matching the given predicate in the collection.
+
+ @note Valid properties to filter on are `path` and `userId`, as well as
+       the boolean properties `mayRead`, `mayWrite`, and `mayManage`.
+ */
+- (RLMSyncPermissionResults *)objectsWithPredicate:(NSPredicate *)predicate;
```
cmelchior: Doesn't this mean you expose the underlying properties of `mayRead`, `mayWrite` and `mayManage` instead of `accessLevel`? I thought the whole purpose of this exercise was to hide those properties?

Realm/RLMSyncPermissionResults.h
cmelchior: Just to make to sure I understand this correctly: The result can only be sorted using the fields defined in `RLMSyncPermissionResultsSortProperty`. Specifically, this means you cannot sort on `accessLevel`?

Realm/RLMSyncPermissionResults.h
austinzheng: Yes. Unfortunately this is one of the places where the abstraction leaks through.

CHANGELOG.md
```diff
@@ -9,6 +9,9 @@ x.x.x Release notes (yyyy-MM-dd)
 
 * Use reachability API to minimize the reconnection delay if the network
   connection was lost.
+* Add a new API for changing and retrieving permissions for synchronized Realms.
```
jpsim: I'd make this plural to indicate that we didn't just add a single method, but a family of types & methods:

```md
* Add new APIs for changing and retrieving permissions for synchronized Realms.
  These API are intended to replace the existing Realm Object-based permissions
  system.
```

Realm/ObjectServerTests/SwiftPermissionsAPITests.swift
```diff
@@ -0,0 +1,240 @@
...
+    private func ensureAbsence(of permission: SyncPermissionValue,
+                               from results: SyncPermissionResults,
+                               after wait: Double = 0.5,
+                               file: StaticString = #file,
+                               line: UInt = #line) {
+        let ex = expectation(description: "Looking for permission")
+        var isPresent = false
+        let token = results.addNotificationBlock { (error) in
+            XCTAssertNil(error, "Notification returned error '\(error!)' when running test at \(file):\(line)")
+            for result in results where result == permission {
```
jpsim: This is just

```diff
-for result in results where result == permission {
-    isPresent = true
-    return
-}
+isPresent = results.contains(permission)
```

Realm/RLMSyncPermissionResults.h
jpsim: Only Swift 3.1 conforms to `Collection`

austinzheng:
Requires: https://github.com/realm/realm-object-store/pull/441. Cancelling CI.

Realm/ObjectServerTests/SwiftPermissionsAPITests.swift
```diff
@@ -88,11 +88,9 @@ class SwiftPermissionsAPITests: SwiftSyncTestCase {
         var isPresent = false
         let token = results.addNotificationBlock { (error) in
             XCTAssertNil(error, "Notification returned error '\(error!)' when running test at \(file):\(line)")
-            for result in results where result == permission {
-                isPresent = results.contains(permission)
-                if (isPresent) {
-                    return
-                }
+            isPresent = results.contains(permission)
+            if isPresent {
```
jpsim: Calling `return` in a block just avoids executing the rest of the block, which is nothing in this case, so just remove the whole check:

```swift
let token = results.addNotificationBlock { error in
    XCTAssertNil(error, "Notification returned error '\(error!)' when running test at \(file):\(line)")
    isPresent = results.contains(permission)
}

austinzheng:
The two CI jobs that failed both failed because of CI issues. For the object server job (the one that matters more), all the tests in the suite executed successfully before CI broke. Merging.

PR #4812, opened 2017-04-03
Disable the ObjectStore Realm caching

tgoyne:
Might fix https://secure.helpscout.net/conversation/343482861/0/?folderId=366246.

Since we still have the cache of obj-c RLMRealm instances the ObjectStore cache is at best a harmless waste of time, but if the Realm instance outlives the owning RLMRealm (perhaps due to a collection notifier for that Realm being in the middle of running when the RLMRealm is deallocaed?) the existing Realm instance may be handed to a new RLMRealm, which is not handled by the existing code and may cause more odd problems.

### DIFF
--- a/Realm/RLMRealmConfiguration.mm
+++ b/Realm/RLMRealmConfiguration.mm
@@ -94,6 +94,8 @@ NSString *RLMRealmPathForFile(NSString *fileName) {
         static NSURL *defaultRealmURL = [NSURL fileURLWithPath:RLMRealmPathForFile(c_defaultRealmFileName)];
         self.fileURL = defaultRealmURL;
         self.schemaVersion = 0;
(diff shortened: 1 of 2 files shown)


Realm/RLMRealmConfiguration.mm
```diff
@@ -94,6 +94,8 @@ - (instancetype)init {
         static NSURL *defaultRealmURL = [NSURL fileURLWithPath:RLMRealmPathForFile(c_defaultRealmFileName)];
         self.fileURL = defaultRealmURL;
         self.schemaVersion = 0;
+        self.cache = YES;
+        _config.cache = false;
```
jpsim: A comment explaining _why_ we opt out of OS caching would be useful here.

Issue #4768, opened 2017-03-24
Deadlock after receiving a "N5realm10LogicErrorE: Bad version number "


tbaranes:
May be a duplicate / follow up of #4428 and #4649

## Goals

Just using Realm.

## Expected Results

No bugs, no freeze, no warnings.

## Actual Results

I'm getting this using Realm v2.4.4, but I got a customer getting that same freeze using v2.0.4, so I was wrong when I was saying that issue was introduced *after* 2.0.4 (cf the previous issue).

The app is freezing after logging "N5realm10LogicErrorE: Bad version number".
Looks like a deadlock (since the audio is still playing, just the interface won't respond anymore)

Here the full stack trace from Xcode when the app has entered in the _deadlock_:

```
* thread #1: tid = 0x14ac3, 0x000000018f6cae1c libsystem_kernel.dylib`__psynch_cvwait + 8, queue = 'com.apple.main-thread', stop reason = signal SIGSTOP
  * frame #0: 0x000000018f6cae1c libsystem_kernel.dylib`__psynch_cvwait + 8
    frame #1: 0x000000018f7909c0 libsystem_pthread.dylib`_pthread_cond_wait + 640
    frame #2: 0x000000018f0b93ec libc++.1.dylib`std::__1::condition_variable::wait(std::__1::unique_lock<std::__1::mutex>&) + 56
    frame #3: 0x0000000101e629f4 Realm`realm::_impl::NotifierPackage::package_and_wait(realm::util::Optional<unsigned long long>) [inlined] void std::__1::condition_variable::wait<std::__1::unique_lock<std::__1::mutex> realm::_impl::RealmCoordinator::wait_for_notifiers<realm::_impl::NotifierPackage::package_and_wait(realm::util::Optional<unsigned long long>)::$_12>(realm::_impl::NotifierPackage::package_and_wait(realm::util::Optional<unsigned long long>)::$_12&&)::'lambda'()>(this=<unavailable>, __pred=<unavailable>)::$_12) + 204 at __mutex_base:345 [opt]
[...]

PR #4813, opened 2017-04-03
update Object Store

### LINKED ISSUE #4802: Bad version number in Realm 2.5.0 (Swift)
## Goals
Got "Bad version number" again in v2.5.0

## Expected Results
This issue happened in v2.1.0, but it seems to be fixed (https://github.com/realm/realm-cocoa/issues/4422), and now came back.

## Actual Results
uncaught exception in notifier thread: N5realm10LogicErrorE: Bad version number

## Steps to Reproduce
This warning shows up randomly, no specific step to reproduce.

Realm version: 2.5.0

Xcode version: 8.3

iOS/OSX version: 10.2

Dependency manager + version: Swift 3.1.0

jpsim:
fixes #4802 and #4768

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,8 @@ x.x.x Release notes (yyyy-MM-dd)
 ### Bugfixes
 
 * Fix CocoaPods installation with static libraries and multiple platforms.
+* Fix uncaught "Bad version number" exceptions when refreshing a Realm with no
+  notification registered blocks.
 
 2.5.0 Release notes (2017-03-28)
 =============================================================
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -13,6 +13,8 @@ x.x.x Release notes (yyyy-MM-dd)
 ### Bugfixes
 
 * Fix CocoaPods installation with static libraries and multiple platforms.
+* Fix uncaught "Bad version number" exceptions when refreshing a Realm with no
+  notification registered blocks.
```
tgoyne: I think these words are in the wrong order, and it's possible (although harder) to hit the problematic case even when using explicit notification blocks, so I'd just say something like "Fix uncaught "Bad version number" exceptions on the notification worker thread followed by deadlocks when Realms refresh".

PR #4851, opened 2017-04-11
Support building Realm 100% from source with CocoaPods

jpsim:
This is #4161 but with CocoaPods subspecs instead of environment variables.

This doesn't affect existing users since the specs default to having sync enabled (i.e. the `Realm/Sync` & `RealmSwift/Sync` subspecs).

Once this is released, usage in the `Podfile` will look like this:

```ruby
pod 'Realm/OSSCore'
# or
pod 'RealmSwift/OSSCore'
```

However, today to try this out, you'll need to specify this in your `Podfile`:

```ruby
pod 'RealmCore', git: 'https://github.com/realm/realm-core.git', branch: 'v2.5.1'
pod 'Realm', subspecs: ['OSSCore'], git: 'https://github.com/realm/realm-cocoa.git', branch: 'jp-oss-podspecs-v2', submodules: true
pod 'RealmSwift', subspecs: ['OSSCore'], git: 'https://github.com/realm/realm-cocoa.git', branch: 'jp-oss-podspecs-v2'
```


Realm/Realm.modulemap
```diff
@@ -18,8 +18,6 @@ framework module Realm {
         header "RLMRealm_Private.h"
         header "RLMResults_Private.h"
         header "RLMSchema_Private.h"
-        header "RLMSyncConfiguration_Private.h"
-        header "RLMSyncUtil_Private.h"
```
mrackwitz: It might make sense to declare here `config_macros REALM_ENABLE_SYNC` (as seen in the [Clang docs](https://clang.llvm.org/docs/Modules.html#configuration-macros-declaration)), so that the compiler knows that the Clang module API looks different on base of that and can use that to invalidate the cache. That might help to avoid problems where derived data would have to be cleared when switching from one variant to the other.

jpsim:
At the moment, you need to specify the `REALM_DISABLE_SYNC` env variable, but I'd like to contribute a patch to CocoaPods that will pass the specs currently being built to the `prepare_command` as an environment variable so we'll be able to tell if the `OSSCore` subspec was specified, which should allow us to do this without requiring the user to specify an environment variable.

build.sh
```diff
@@ -1016,38 +1022,35 @@ their entries in your Podfile.
...
+        cp Realm/ObjectStore/src/*.hpp include
+        cp Realm/ObjectStore/src/sync/*.hpp include/sync
+        cp Realm/ObjectStore/src/sync/impl/*.hpp include/sync/impl
+        cp Realm/ObjectStore/src/impl/*.hpp include/impl
+        cp Realm/ObjectStore/src/impl/apple/*.hpp include/impl/apple
+        cp Realm/ObjectStore/src/util/*.hpp include/util
+        cp Realm/ObjectStore/src/util/apple/*.hpp include/util/apple
+
+        touch Realm/RLMPlatform.h
+        if [ -n "$COCOAPODS_VERSION" ]; then
```
mrackwitz: It just occurred to me that changes to the podspec are likely to break backwards-compatibility to versions of CocoaPods before 1.0, as [documented in this commit](https://github.com/realm/realm-cocoa/commit/71bb89389d1784d1130f001a5986a0735cb896c8). That seems acceptable at this point, so that we could also drop this if branch.

jpsim: I'm fine to require CocoaPods 1.0 or later. It'll have been out for over a year by the time this is released.

austinzheng:
Is this still in progress, or on hold?

bdash:
Closing due to this being very out of date, and not something we intend to actively work on in the near future.

PR #4859, opened 2017-04-13
Add method to open Realm asynchronously

jpsim:
Continued from @austinzheng's initial work in #4681.

Design goals for this were to:

- Work with both local and synced Realms
- Not block current thread until migration completes for local Realms
- Not invoke callback until full Realm contents at the moment the operation was kicked off is fully available locally
- Avoid eagerly writing the schema or other data to synced Realms, causing unnecessary transactions for the Realm Object Server to merge
- Not leak Realms or pin their versions outside the scope of the callback block


Realm/RLMRealm.mm
```diff
@@ -180,6 +185,77 @@ + (instancetype)realmWithURL:(NSURL *)fileURL {
     configuration.fileURL = fileURL;
     return [RLMRealm realmWithConfiguration:configuration error:nil];
 }
+
++ (dispatch_queue_t)asyncOpenDispatchQueue {
+    static dispatch_queue_t queue;
```
tgoyne: This can just be `static dispatch_queue_t queue = dispatch_queue_create("io.realm.asyncOpenDispatchQueue", DISPATCH_QUEUE_CONCURRENT); return queue;`.

Or just forgo creating the queue entirely.

Realm/RLMRealm.mm
```diff
@@ -180,6 +185,77 @@ + (instancetype)realmWithURL:(NSURL *)fileURL {
...
+                callback(nil, error);
+            });
+            return;
+        }
+    }
+    dispatch_async(self.asyncOpenDispatchQueue, ^{
+        (void)realmStrongRef;
+        @autoreleasepool {
+            RLMRealm *realm = [RLMRealm realmWithConfiguration:configuration error:&error];
+            if (realm && !error) {
```
tgoyne: This would benefit a lot from an early-return for the error case.

Realm/RLMRealm.mm
```diff
@@ -180,6 +185,77 @@ + (instancetype)realmWithURL:(NSURL *)fileURL {
...
+                    dispatch_async(callbackQueue, ^{
+                        @autoreleasepool {
+                            NSError *error = nil;
+                            RLMRealm *localRealm = [RLMRealm realmWithConfiguration:configuration error:&error];
+                            callback(localRealm, error);
+                        }
+                    });
+                    return;
+                }
+                session->wait_for_download_completion([=,session=session](std::error_code error) {
```
tgoyne: `session=session` is the same thing as just `session`.

Realm/RLMRealm.h
```diff
@@ -91,6 +93,26 @@ NS_ASSUME_NONNULL_BEGIN
 + (instancetype)realmWithURL:(NSURL *)fileURL;
 
 /**
+ Open a Realm asynchronously on a background queue, and then invoke a callback
```
tgoyne: <s>"background queue" suggests that the main queue isn't a legal argument for this.</s>

I misunderstood what this was even referring to, and it's not wrong but maybe misleading? Maybe something like "Asynchronously open a Realm and deliver it to a block on the given queue"?

Realm/RLMRealm.h
```diff
@@ -19,7 +19,9 @@
 #import <Foundation/Foundation.h>
 #import "RLMConstants.h"
 
-@class RLMRealmConfiguration, RLMObject, RLMSchema, RLMMigration, RLMNotificationToken, RLMThreadSafeReference;
+@class RLMRealmConfiguration, RLMRealm, RLMObject, RLMSchema, RLMMigration, RLMNotificationToken, RLMThreadSafeReference;
+
+typedef void(^RLMAsynchronouslyOpenRealmCallback)(RLMRealm * _Nullable realm, NSError * _Nullable error);
```
tgoyne: "Asynchronous Open" or "Open Asynchronously" (adverbs normally go after the thing they modify).

Realm/RLMRealm.h
```diff
@@ -91,6 +93,26 @@ NS_ASSUME_NONNULL_BEGIN
...
+ its remote content available at the time the operation began.
+
+ @param configuration A configuration object to use when opening the Realm.
+ @param callbackQueue The dispatch queue on which the callback should be run.
+ @param callback      A callback block. If the Realm was successfully opened,
+                      it will be passed in as an argument.
+                      Otherwise, an `NSError` describing what went wrong will be
+                      passed to the block instead.
+ */
++ (void)openAsynchronouslyWithConfiguration:(RLMRealmConfiguration *)configuration
```
tgoyne: How about just `openOnQueue:configuration:callback:` or `openAsyncOnQueue:configuration:callback:`.

Realm/RLMRealm.h
```diff
@@ -91,6 +93,26 @@ NS_ASSUME_NONNULL_BEGIN
 + (instancetype)realmWithURL:(NSURL *)fileURL;
 
 /**
+ Open a Realm asynchronously on a background queue, and then invoke a callback
+ on success or failure.
+
+ Opening a Realm asynchronously may be useful if, for example, a time-consuming
```
tgoyne: "may be useful if, for example," is needlessly hesitant about what this is for. A rough stab at better wording:

Opening a Realm asynchronously will perform all work needed to get the Realm to a usable state (such as running potentially time-consuming migrations) on a background thread before dispatching to the given queue. In addition, synchronized Realms wait for all remote content available at the time the operation began to be downloaded.

Realm/RLMRealm.h
jpsim: Neither of those are entirely accurate either, since the _initial_ open doesn't happen on the provided queue. I'm thinking abbreviating to `asyncOpenWithConfiguration:callbackQueue:callback:`. There's not much value in having the unabridged word, especially if the Swift API uses `asyncOpen` too.

Realm/RLMRealm.mm
jpsim: That doesn't work, but apparently just `[=]` is sufficient.

jpsim:
Some of the checks to verify that the file hasn't yet been created are racy and occasionally fail if the async dispatch is invoked quickly enough. For example, see this failure in [testOpenAsync on ios-static Xcode 8.0](https://ci.realm.io/job/objc_pr/4983/configuration=Release,target=ios-static,xcode_version=8.0/).

Any ideas to make this more reliable without introducing an artificial delay in the `asyncOpen` implementation?

tgoyne:
You could create the lock file, acquire an exclusive file lock on it, call asyncOpen, and then release the lock. This will block the background thread from creating the realm file until you are ready for it to (and make it deadlock if it tries to open the realm on the same thread, which is not exactly the *ideal* way to fail...).

Realm/RLMRealm.h
```diff
@@ -91,6 +98,26 @@ NS_ASSUME_NONNULL_BEGIN
...
+ operation began to be downloaded and available locally.
+
+ @param configuration A configuration object to use when opening the Realm.
+ @param callbackQueue The dispatch queue on which the callback should be run.
+ @param callback      A callback block. If the Realm was successfully opened,
+                      it will be passed in as an argument.
+                      Otherwise, an `NSError` describing what went wrong will be
+                      passed to the block instead.
+ */
++ (void)asyncOpenWithConfiguration:(RLMRealmConfiguration *)configuration
```
tgoyne: The docs should mention that the Realm is still thread-confined and not queue-confined (and thus should not be held on to past the callback for anything but the main queue) as otherwise I can guarantee we'll eventually get a question from someone expecting it to be bound to that queue.

Realm/RLMRealm.h
```diff
@@ -19,7 +19,14 @@
...
 
-@class RLMRealmConfiguration, RLMObject, RLMSchema, RLMMigration, RLMNotificationToken, RLMThreadSafeReference;
+@class RLMRealmConfiguration, RLMRealm, RLMObject, RLMSchema, RLMMigration, RLMNotificationToken, RLMThreadSafeReference;
+
+/**
+ A callback block for opening Realms asynchronously.
+
+ Returns the Realm if the open was successful, or an error otherwise.
+ */
+typedef void(^RLMAsynchronouslyOpenRealmCallback)(RLMRealm * _Nullable realm, NSError * _Nullable error);
```
tgoyne: RLMOpenAsyncCallback to go with the new method name?

Realm/Tests/MigrationTests.mm
```diff
@@ -652,6 +653,33 @@ - (void)testMigrationBlockCalledWhenSchemaVersionHasChanged {
...
+    XCTestExpectation *ex = [self expectationWithDescription:@"async-migration"];
+    __block bool migrationCalled = false;
+    c.schemaVersion = 2;
+    c.migrationBlock = ^(__unused RLMMigration *migration, __unused uint64_t oldSchemaVersion) {
+        migrationCalled = true;
+    };
+    [RLMRealm asyncOpenWithConfiguration:c
+                           callbackQueue:dispatch_get_main_queue()
+                                 callback:^(RLMRealm * _Nullable realm, NSError * _Nullable error) {
+        [ex fulfill];
```
tgoyne: Fulfilling the expectation should always be the last thing done in the block when the block isn't running on the queue owned by `RLMTestCase`, as otherwise the following assertions in the block could be checked after the test case has completed and been marked as passing (there's a bunch of tests in this PR that do this).

Realm/Tests/MigrationTests.mm
```diff
@@ -652,6 +653,33 @@ - (void)testMigrationBlockCalledWhenSchemaVersionHasChanged {
...
+        XCTAssertNil(error);
+        XCTAssertNotNil(realm);
```
tgoyne: Should also test `migrationCalled` within the block.

Realm/RLMRealm.h
jpsim: Well, it'd be `RLMAsyncOpenRealmCallback` since the new method name is "async open" not "open async". But sure.

PR #4860, opened 2017-04-13
retry downloading core and sync binaries

larkost:


jpsim:
Is this ready for review and can you please share examples of CI failures that this should fix?

larkost:
I am trying to puzzle out what is happening here, since the `tvos Release 8.0` is showing the exact error that this was trying to avoid... in a way that should no longer be possible (the message has changed). So... not yet, and the example: https://ci.realm.io/job/objc_pr/configuration=Release,target=tvos,xcode_version=8.0/4981/console

jpsim:
Alright. I'll mark this as `S:In Progress` then to reflect its current state.

I'd appreciate if you could read our [GitHub Issues Usage](https://github.com/realm/realm-wiki/wiki/Realm-GitHub-Issues-Usage) wiki article which outlines how we use labels as part of our engineering management process.

larkost:
This time it tested the correct version, so things look ready to go.

jpsim:
> This time it tested the correct version, so things look ready to go.

As indicated in [GitHub Issues Usage](https://github.com/realm/realm-wiki/wiki/Realm-GitHub-Issues-Usage) wiki article, this is when you'd change the label from `S:In Progress` to `S:Review`. I've done that this time and am reviewing this now.

build.sh
```diff
@@ -321,66 +321,61 @@ download_object_server() {
...
-    if [ ! -f "${CORE_TAR}" ]; then
-        local CORE_URL="https://static.realm.io/downloads/core/realm-core-${REALM_CORE_VERSION}.tar.xz"
-        set +e # temporarily disable immediate exit
-        local ERROR # sweeps the exit code unless declared separately
-        ERROR=$(curl --fail --silent --show-error --location "$CORE_URL" --output "${CORE_TMP_TAR}" 2>&1 >/dev/null)
-        if [[ $? -ne 0 ]]; then
-            echo "Downloading core failed:\n${ERROR}"
-            exit 1
+download_common() {
+    local download_type=$1 tries_left=3 version core_url error temp_dir temp_path tar_path
```
jpsim: `core_url` should be `url`

build.sh
```diff
@@ -321,66 +321,61 @@ download_object_server() {
...
+        echo "Unknown dowload_type: $download_type"
+        exit 1
+    fi
+    
+    echo "Downloading dependency: ${download_type} ${version}"
+    
+    if [ -z "$TMPDIR" ]; then
+        TMPDIR='/tmp'
+    fi
+    temp_dir=$(dirname "$TMPDIR/waste")/core_bin
```
jpsim:
```bash
temp_dir="$(dirname "$TMPDIR/waste")/${download_type}_bin"
```

build.sh
```diff
@@ -321,66 +321,61 @@ download_object_server() {
...
+    mkdir -p "$temp_dir"
+    temp_path="${temp_dir}/${download_type}-${version}.tar.xz.tmp"
```
jpsim: Flip these?

```bash
tar_path="${temp_dir}/${download_type}-${version}.tar.xz"
temp_path="${tar_path}.tmp"
```

build.sh
```diff
@@ -321,66 +321,61 @@ download_object_server() {
...
+    tar_path="${temp_dir}/${download_type}-${version}.tar.xz"
+        
+    while [ 0 -lt $tries_left ] && [ ! -f "$tar_path" ]; do
+        if ! error=$(curl --fail --silent --show-error --location "$url" --output "$temp_path" 2>&1); then
+            tries_left=$[$tries_left-1]
```
jpsim: according to [shellcheck](https://www.shellcheck.net), this should be `tries_left=$($tries_left-1)`

build.sh
```diff
@@ -321,66 +321,61 @@ download_object_server() {
...
+        else
+            mv "$temp_path" "$tar_path"
         fi
-        set -e # re-enable flag
-        mv "${CORE_TMP_TAR}" "${CORE_TAR}"
+    done
+    
+    if [ ! -f $tar_path ]; then
```
jpsim: should be quoted: `"$tar_path"`

build.sh
```diff
@@ -321,66 +321,61 @@ download_object_server() {
...
+        printf "Downloading ${download_type} failed:\n\t$url\n\t$error\n"
```
jpsim: according to [shellcheck](https://www.shellcheck.net), format strings are preferred with `printf`:

```bash
printf "Downloading %s failed:\n\t%s\n\t%s\n" "$download_type" "$url" "$error"
```

build.sh
```diff
@@ -321,66 +321,61 @@ download_object_server() {
...
     fi
-
+    
     (
-        cd "${TMP_DIR}"
-        rm -rf core
-        tar xf "${CORE_TAR}" --xz
-        mv core core-${REALM_CORE_VERSION}
+        cd "$temp_dir"
+        rm -rf $download_type
```
jpsim: fine today, but might as well stay safe in the future by quoting this: `rm -rf "$download_type"`

build.sh
larkost: That is both harder to read in this case, and gets us into potential problems between GNU and BSD versions of bash (the former requires commas, the latter confusingly prints the commas). In theory using the format string would avoid some injection-style stacks, but I don't think we are worried about those here.

build.sh
```diff
@@ -340,27 +340,27 @@ download_common() {
...
         if ! error=$(curl --fail --silent --show-error --location "$url" --output "$temp_path" 2>&1); then
-            tries_left=$[$tries_left-1]
+            tries_left=$($tries_left-1)
         else
             mv "$temp_path" "$tar_path"
         fi
     done
     
-    if [ ! -f $tar_path ]; then
+    if [ ! -f '$tar_path" ]; then
```
jpsim: can't mix single quotes with double quotes

Issue #3390, opened 2016-03-29
Can't enumerate objects in migration after deleting


mladjan:
I'm doing simple STLUser migration because we added new fields in STLUser model.
We are doing this like that: http://hastebin.com/iduyemuxol

Problem happens when we migrate from v7 to v8.

The problem is, that on some cases to some users (not all), migration fails: http://cloud.blgrd.co/413q3r45180T

The reason is that newObjects array has two more objects, and I don't have idea why.
Few more screenshots:
http://cloud.blgrd.co/2b3F44451v2g
http://cloud.blgrd.co/0v1a03320D0i

Has anybody clue why this is happening?

jpsim:
Sorry you're affected by this issue, this is the first report of unmatched old/new migration objects we've ever seen. Could you please share a bit more information so that we can better identify what may have caused this? Please share:

1. The versions of Realm shipped with your app alongside schema versions 7 & 8.
2. Your `STLUser` model definitions for schema versions 7 & 8.
3. Any other relevant information that could help identify why this happened to a subset of users. What's unique (if anything) about the users affected compared to those unaffected?
  1. iOS version.
  2. iOS device.
  3. Number of `STLUser` objects in both the old and new Realms.

The best way to make progress on identifying what may have caused this would be to attempt to reproduce the configuration that lead to this issue, so if you manage to find and share reproducible steps with us, that would greatly accelerate the process.

Thanks!

mladjan:
Hi guys,
1. Both versions have 0.97 ObjC
2. Changes in STLUser model are: http://cloud.blgrd.co/3J3a3h3a2A1X
In v8 model STLUser: http://cloud.blgrd.co/2e0Y1G1U3T3a
In v7 model STLUser: http://cloud.blgrd.co/0Y3h3y0F2h0p
3.1. iOS 9.2
3.2. iPhone 6+
3.3. Based on that screenshots we can see: http://cloud.blgrd.co/413q3r45180T 
old objects: 545
new objects: 542

Note: its happening only for some users, not all of them.

jpsim:
Reading your migration code again, the error is clear: objects from the new Realm are being deleted and then old & new objects are being enumerated:

```objc
if (oldSchemaVersion < 6) {
    [migration enumerateObjects:STLUser.className block:^(RLMObject * _Nullable oldObject, RLMObject * _Nullable newObject) {
        STLUser *oldUser = (STLUser *)oldObject;
[...]

PR #4864, opened 2017-04-14
Fix not being able to enumerate objects in migration after deleting

kishikawakatsumi:
Fixes https://github.com/realm/realm-cocoa/issues/3390

The cause of the issue is that when deleting objects, the number of objects does not match between the new schema and the old schema.

In the current version, this case crashes, so it will not be a big problem.
However, If inserting the same number of objects as deleted objects, it will not crash and can not be enumerated correctly (the order will not match between the old schema and the new schema), the issue becomes serious.

To solve this problem, defer deletion during migration and actually delete the object when migration finished.

CC @jpsim @tgoyne

### DIFF
--- a/Realm/RLMMigration.mm
+++ b/Realm/RLMMigration.mm
@@ -20,6 +20,7 @@
 
 #import "RLMAccessor.h"
 #import "RLMObject_Private.h"
+#import "RLMObject_Private.hpp"
 #import "RLMObjectSchema_Private.hpp"
 #import "RLMObjectStore.h"
 #import "RLMProperty_Private.h"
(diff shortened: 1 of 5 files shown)


Realm/RLMMigration.mm
```diff
@@ -131,7 +144,17 @@ - (RLMObject *)createObject:(NSString *)className withObject:(id)object {
 }
 
 - (void)deleteObject:(RLMObject *)object {
-    [_realm deleteObject:object];
+    [deletedObjectIndices[object.objectSchema.className] addObject:@(object->_row.get_index())];
+}
+
+- (void)deleteObjectsActually {
```
jpsim: Could you please rename this to `deleteObjectsMarkedForDeletion`?

kishikawakatsumi:
@jpsim Thanks. I reflected your reviews.  I borrowed your sentence as-is for the changelog.

jpsim:
Thanks Katsumi!

PR #4866, opened 2017-04-15
Add predicate option to LinkingObjects initializer

AndrewBarba:
If you guys are okay with this API I can go ahead and add tests

### DIFF
--- a/RealmSwift/LinkingObjects.swift
+++ b/RealmSwift/LinkingObjects.swift
@@ -29,19 +29,26 @@ public class LinkingObjectsBase: NSObject, NSFastEnumeration {
     @objc fileprivate var object: RLMWeakObjectHandle?
     @objc fileprivate var property: RLMProperty?
     @objc fileprivate var predicate: NSPredicate?
+    @objc fileprivate var sortDescriptors: [RLMSortDescriptor]?
 
     internal var rlmResults: RLMResults<RLMObject> {
         if cachedRLMResults == nil {
             if let object = self.object, let property = self.property {
-                let results = RLMDynamicGet(object.object, property)! as? RLMResults
+                var results = RLMDynamicGet(object.object, property)! as? RLMResults
+                
                 if let predicate = self.predicate {
-                    self.cachedRLMResults = results?.objects(with: predicate)
-                } else {
-                    self.cachedRLMResults = results
+                    results = results?.objects(with: predicate)
                 }
+                
+                if let sortDescriptors = self.sortDescriptors {
+                    results = results?.sortedResults(using: sortDescriptors)
+                }
+                
+                self.cachedRLMResults = results
+                self.predicate = nil
                 self.object = nil
                 self.property = nil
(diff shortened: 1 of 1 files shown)


tgoyne:
This won't work correctly when querying based on the LinkingObjects property (it won't use the filter, since the query engine doesn't instantiate the Swift objects), and I think that'd be a requirement to merge this. I think it's conceptually possible to fix that, but it'd be fairly difficult as the relevant code is already dreadfully complicated.

AndrewBarba:
@tgoyne I wrote a simple test on my own code that worked fine, but I guess I don't really understand the internals here. The underlying `rlmResults` aren't created until someone tries to access the property, and then any filter off that property would obviously work fine. How is what I am going any different? The filter is applied at the same time user land code would apply it. The basic test I wrote looked something like this:

```
expect(user.profile.count).to(equal(0))
expect(user.firstName).toNot(beNil())
expect(user.lastName).toNot(beNil())
            
let realm = try! Realm()
    try! realm.write {
    let p1 = UserProfile()
    p1.deleted = false
    p1.user = user
                
    let p2 = UserProfile()
    p2.deleted = true
    p2.user = user
                
    realm.add(p1)
    realm.add(p2)
}
            
expect(user.profile.count).to(equal(1))
```

jpsim:
@AndrewBarba what Thomas is saying is that you can query linking objects like this:

```swift
class Person: Object {
  dynamic var name = ""
  dynamic var age = 0
  let children = List<Person>()
  // all parents
  let parents = LinkingObjects(fromType: Person.self, property: "children")
}

let realm = try! Realm()
let dianesKids = realm.objects(Person.self).filter("ANY parents.name == 'Diane'")
```

However, your PR doesn't update the query engine to incorporate the filter defined at property declaration:

```swift
class Person: Object {
  dynamic var name = ""
  dynamic var age = 0
  let children = List<Person>()
  // all parents whose name isn't 'Diane'
  let parents = LinkingObjects(fromType: Person.self, property: "children", with: NSPredicate(string: "name != 'Diane'"))
}

let realm = try! Realm()
let dianesKids = realm.objects(Person.self).filter("ANY parents.name == 'Diane'")
// ^still returns Diane's kids despite the 'parents' property filtering those out
```

AndrewBarba:
@jpsim Makes perfect sense, bummer :(

jpsim:
You can do what your PR does fairly easily today by using computed properties, btw:

```swift
class Person: Object {
  dynamic var name = ""
  dynamic var age = 0
  let children = List<Person>()
  private let allParents = LinkingObjects(fromType: Person.self, property: "children")
  // all parents whose name isn't 'Diane'
  var parents: Results<Person> { return allParents.filter("name != 'Diane'") }
}
```

I really should have come up with a better example since "parents who aren't Diane" doesn't make much sense, but you get the idea.

AndrewBarba:
Yup, great point, that's what I'll do. Feel free to close this and the issue, thanks guys

PR #4867, opened 2017-04-17
disable testMultipleProcessesTryingToInitializeSchema in debug mode

jpsim:
since it occasionally fails on CI and its failure doesn't indicate something inherently broken with the patch being tested.

Addresses #4585

This was originally filed as #4781, but we closed that in the hopes that https://github.com/realm/realm-core/pull/2552 would have addressed this occasional failure, but this was seen again after integrating that here: https://ci.realm.io/job/objc_pr/5006/configuration=Debug,target=osx,xcode_version=8.0/


jpsim:
Just had another two failures in #4920 ([Jenkins run](https://ci.realm.io/job/objc_pr/5077/)). Merging this as we just ignore those failures anyway.

PR #4883, opened 2017-04-20
fix sync session being closed

jpsim:
before `wait_for_download_completion` called back.

This fixes an error that @dhmspector was seeing in a demo app, and I've confirmed that this change works successfully in his app.


austinzheng:
`wait_for_download_completion()` doesn't currently do anything special to prolong the session's lifetime. If you want, I can change that code so that it does.

jpsim:
Other bindings might benefit from that.

Issue #4855, opened 2017-04-12
SyncUser.logout() doesn't work as specified


icanzilb:
## Goals
I want to log out a SyncUser as described in the docs: 
https://realm.io/docs/swift/latest/#logging-out

## Expected Results

The API will behave as described, namely "To log a user out of their account, call SyncUser.logOut(). Any pending local changes will continue to be uploaded until the Realm Object Server has been fully synchronized. Then, all of their local synced Realms will be deleted from their device."

## Actual Results

The realm files remain on the client device.

## Steps to Reproduce

Log a user in, then log them out.

## Code Sample

[LogOutTest.zip](https://github.com/realm/realm-cocoa/files/915726/LogOutTest.zip)

1) Unzip the folder, run `pod install`, then run the app and observe the folder in the Simulator.
2) Click on LogIn and note that the sync file appears in the Documents folder:

![image](https://cloud.githubusercontent.com/assets/633862/24946229/8b147a08-1f62-11e7-875a-b4b26c802f9f.png)

3) The app will move on to the next screen, there click on LogOut button.

The synced file remains on the clients device.

## Version of Realm and Tooling

```

RealmSwift: 2.5
ROS: 1.4.1 

ProductName:	Mac OS X
ProductVersion:	10.12.4
BuildVersion:	16E195

/Users/marin/Developer/Xcode 8.3.app/Contents/Developer
Xcode 8.3
Build version 8E162

/Users/marin/.rvm/gems/ruby-2.2.1/bin/pod
1.2.0
(not in use here)

/bin/bash
GNU bash, version 3.2.57(1)-release (x86_64-apple-darwin16)

/usr/local/bin/carthage
0.18.1
(not in use here)

/usr/local/bin/git
git version 2.8.2
```

jpsim:
The docs are a bit wrong there. The Realms are deleted on the next app launch. Not immediately. I'll update the docs.

icanzilb:
Isn't it rather inconvenient that there is no support for logout in the same app launch? what if it's sensitive data that people do not want to hang around?

austinzheng:
The problem is that simply deleting the Realm files immediately is going to crash the user's app unless they explicitly invalidate all their Realms and objects before calling logout. We could have an API that vends a block they could call once they've finished cleaning up to delete the Realm files immediately (like how client reset works right now), but that would be a feature request we'd need to prioritize along with everything else.

tgoyne:
[...]

PR #4890, opened 2017-04-21
update logOut docs to reflect that Realms are deleted on next app launch

jpsim:
Addresses #4855.


austinzheng:
Looks good.

You may or may not want to note that, if the user is logged back in, their files will not be removed upon next launch. This might be too obvious to be worth mentioning, though.

jpsim:
I think that's obvious enough.

PR #4902, opened 2017-04-25
fail the build at the first sign of trouble

larkost:


jpsim:
This "worked" in the sense that it failed early, but obviously the ideal would be for our VMs to not have so much difficulty cloning a git repo 😬 . https://ci.realm.io/job/Cocoa%20Installation%20Examples/599/binding=objc,method=cocoapods,platform=ios/console

Also, failing early means we lost the "failure indication" for this type of failure:

image

larkost:
I was waiting to see what the new failure looks like to put in the new indication (just done), and have been trying to figure out why the https://github.com/philsquared/Catch sub (sub) module is always the one that fails. It is tiny, so it would be odd if it was actually a network error. I can occasionally reproduce this on the remote machines, and am trying to narrow it down to an actual actionable problem.

jpsim:
Amazing

PR #4910, opened 2017-04-27
Fix async open interaction with read-only Realms

austinzheng:
Changes:
- async open no longer opens a synced Realm conventionally at the beginning, causing data to be written to the server

This change is verified through a unit test in an upcoming PR. (It is not included in this PR because it contains a large number of upcoming dependencies that would be impractical to port into this PR.)


Realm/RLMRealm.mm
```diff
@@ -187,30 +187,60 @@ + (instancetype)realmWithURL:(NSURL *)fileURL {
...
     if (hasSyncConfig) {
         realmStrongRef = [RLMRealm uncachedSchemalessRealmWithConfiguration:configuration error:&error];
-        if (error) {
-            dispatch_async(callbackQueue, ^{
-                callback(nil, error);
-            });
-            return;
-        }
+    }
+    if (error) {
```
jpsim: this check just needs to happen in the `hasSyncConfig` case.

PR #4912, opened 2017-04-29
Align KVC behavior for nil-values for Swift and ObjC

dimazen:
Currently passing `nil` via `-[RLMObject setValue:forKey:]` in ObjC removes all values from the Array property:
```objc
void RLMSetValue(RLMObjectBase *const obj, ..., id<NSFastEnumeration> const array, ...) {
    ...
    realm::LinkViewRef linkView = obj->_row.get_linklist(colIndex);
    linkView->clear();
    for (RLMObjectBase *link in array) {
        ...
    }
}
```
As you can see the `for` loop will skip its body for nil `array` which is ok. 

However in Swift `Realm.Object.setValue(_:forKey:)` with `nil` value doesn't do anything for Array properties. KVC fallthrough to the `-[RLMObjectBase setValue:forUndefinedKey:]` where we have the following code:
```objc
- (void)setValue:(id)value forUndefinedKey:(NSString *)key {
    RLMProperty *property = _objectSchema[key];
    if (Ivar ivar = property.swiftIvar) {
        if (property.type == RLMPropertyTypeArray && [value conformsToProtocol:@protocol(NSFastEnumeration)]) {
            RLMArray *array = [object_getIvar(self, ivar) _rlmArray];
            [array removeAllObjects];
            [array addObjects:validatedObjectForProperty(value, property, RLMSchema.partialSharedSchema)];
        }
       ....
```

Check ` [value conformsToProtocol:@protocol(NSFastEnumeration)]` fails and therefore property remains untouched. 

I'm not sure whether tests updated in correct place so please advise where it should be done.


dimazen:
Hm, part of the tests have failed. Can someone share details please? (I don't have access on Jenkins).

jpsim:
Hi @dimazen, sorry when I attempted to push my commits to your branch, it closed this PR. I've continued your work in #4917 and intend to merge that once CI is happy with it.

Many thanks for your work here!

dimazen:
@jpsim no worries. Thanks for merge! :) Looking forward for the new release!

PR #4915, opened 2017-05-01
Update realm-cocoa to use sync 1.7.0

austinzheng:
Changes
- SSL validation APIs removed from sync manager
- SSL validation APIs added to `RLMSyncConfiguration` and `SyncConfiguration`

Requires https://github.com/realm/realm-object-store/pull/442.

Note:
- This is a breaking change, although there is no good reason for anyone to be setting that flag to false in production. We may want to make it less of one.
- `RLMSyncConfiguration` becomes mutable since you can make one, and then set the SSL flag later.


Realm/RLMSyncManager.h
```diff
@@ -79,14 +79,6 @@ typedef void(^RLMSyncErrorReportingBlock)(NSError *, RLMSyncSession * _Nullable)
 @property (nonatomic, copy) NSString *appID;
 
 /**
- Whether SSL certificate validation should be disabled. SSL certificate validation is ON by default. Setting this
```
jpsim: We need to preserve this API to maintain source compatibility. It shouldn't be too hard to just store this value in an ivar-backed property and look it up when constructing `SyncConfig` instances.

jpsim:
`dependencies.list` needs to be updated (core, sync & object server would be good), we need to update the object store submodule commit, and this needs changelog entries for the fixes the dependency updates pull in. Among other changes being pulled in, there's the lifting of the 16MB transaction size limit:

```md
* Fix crash when committing a write transaction with more than 16MB of changes
  in a synchronized Realm. Write transactions are no longer constrained in
  size in any Realm.
```

Realm/RLMSyncManager.h
austinzheng: Would it be okay to keep the API, stick a deprecation notice on it, and have it do nothing?

jpsim: Deprecation notice is definitely a good idea, but the change I proposed doesn't seem to hard to do in the meantime to avoid entirely breaking existing uses of this in the wild?

austinzheng: They're not hard to make at all, but it would really be good to move users off this global property (and minimize the risk that they disable validation for their entire application unintentionally).

jpsim: So then we're in agreement that marking as deprecated, but mostly retaining its functionality until it's entirely removed, is the right move?

austinzheng: Okay, sure. If you don't set it we'll use the value from the configuration. If you do set it we'll log a warning to console and then use the global value instead.

jpsim: no need to log a warning to the console if we mark the API as deprecated in the interface, leading to a compile-time warning instead.

CHANGELOG.md
```diff
@@ -9,6 +9,12 @@ x.x.x Release notes (yyyy-MM-dd)
 
 * Use reachability API to minimize the reconnection delay if the network
   connection was lost.
+* `RLMSyncConfiguration` now has an `enableSSLValidation` property to allow
```
jpsim: These two items are really about the same thing, consider combining:

```md
* `{RLM}SyncConfiguration` now has an `enableSSLValidation` property
  (and default parameter in the Swift initializer) to allow SSL validation to be
  specified on a per-server basis.
```

RealmSwift/Sync.swift
```diff
@@ -137,28 +137,41 @@ public struct SyncConfiguration {
      */
     internal let stopPolicy: RLMSyncStopPolicy
 
+    internal let enableSSLValidation: Bool
```
jpsim: `public`?

PR #4916, opened 2017-05-01
Implement changePassword API

jpsim:
Addresses https://github.com/realm/product/issues/294.


jpsim:
Passed CI in https://ci.realm.io/job/objc_pr/5073/.

PR #4922, opened 2017-05-02
Update sync to 1.8.2 and core to 2.6.2

### LINKED ISSUE #4895: Infinite loop in db query [Realm Swift 2.6.2]
In the latest Realm update, when running a filter query involves case-insensitive string compare, it went into an infinite loop causing the phone heats up. 

## Steps to Reproduce

This is 100% reproducible with this query: `("displayName = [c]%@", id)`

It worked fine till the 2.6.1 update, and still happens in 2.6.2.

## Code Sample

Here is the code snippet:

```swift
static func getContactByDisplayName(_ displayName:String?) -> EdoContact? {
    if let id = displayName {
        return filter("displayName = [c]%@", id).first //case insensitive
    } else {
        return nil
    }
}
```

- Realm framework version: Realm Swift 2.6.1 & 2.6.2
- Realm Object Server version: N/A
- Xcode version: 8.3
- iOS/OSX version: 10.3
- Dependency manager + version: ?

jpsim:
Fixes #4895.


jpsim:
The 3 CI failures are unrelated.

CHANGELOG.md
```diff
@@ -31,6 +31,9 @@ x.x.x Release notes (yyyy-MM-dd)
 * Using KVC to set a `List` property to `nil` now clears it to match the
   behavior of `RLMArray` properties.
 * Fix crash from `!m_awaiting_pong` assertion failure when using synced Realms.
+* Fix for very computationally expensive case-insensitive searches on indexed
```
bdash: I'd rephrase this more like "Fix poor performance or hangs when performing case-insensitive queries on indexed string properties that contain many characters that don't differ between upper and lower case (e.g., numbers, punctuation)".

jpsim: Thanks.

Issue #4900, opened 2017-04-25
Use Network Link Conditioner in object server tests


jpsim:
The [Network Link Conditioner](http://nshipster.com/network-link-conditioner/) could be integrated into our Realm Object Server integration tests to simulate packet loss, slow network connections, connectivity loss. This would be useful to test the Reachability integration that was added in #4794.

PR #4923, opened 2017-05-03
Test ROS on a poor network connectivity condition 

kishikawakatsumi:
Fixes https://github.com/realm/realm-cocoa/issues/4900
This is a WIP, but I’d appreciate it if you could give me your feedback.

Add an ROS test that running on a slow network environment. It uses `pfctl`, ` dnctl` and `dummynet` to simulate a slow network. Network Link Conditioner probably uses the same mechanism. Since Network Link Conditioner doesn't has command line interface, it doesn't seem to affect local hosts.

@jpsim Does this fit your thoughts? These commands require `sudo`. How should we handle a password?

CC @bdash @stel

### DIFF
--- a/Realm/ObjectServerTests/RLMObjectServerTests.mm
+++ b/Realm/ObjectServerTests/RLMObjectServerTests.mm
@@ -1356,7 +1356,11 @@
             XCTAssertTrue([[RLMRealm realmWithConfiguration:c error:nil] isEmpty]);
         }
         XCTAssertNil(RLMGetAnyCachedRealmForPath(c.pathOnDisk.UTF8String));
-        [self waitForExpectationsWithTimeout:60.0 handler:nil];
+        if (simulateReconnection) {
+            [self waitForExpectationsWithTimeout:60.0 handler:nil];
+        } else {
+            [self waitForExpectationsWithTimeout:10.0 handler:nil];
+        }
         XCTAssertGreaterThan(fileSize(c.pathOnDisk), sizeBefore);
         XCTAssertNil(RLMGetAnyCachedRealmForPath(c.pathOnDisk.UTF8String));
     } else {
@@ -1374,7 +1378,7 @@
             [realm addObject:[HugeSyncObject object]];
         }
         [realm commitWriteTransaction];
-        [self waitForUploadsForUser:user url:url];
+        [self waitForUploadsForUser:user url:url timeout:simulateReconnection ? 20.0 : 60.0 error:nil];
         CHECK_COUNT(NUMBER_OF_BIG_OBJECTS, HugeSyncObject, realm);
     }
 }
--- a/Realm/ObjectServerTests/RLMObjectServerTests.mm
+++ b/Realm/ObjectServerTests/RLMObjectServerTests.mm
@@ -36,12 +36,22 @@
 /// Valid username/password credentials should be able to log in a user. Using the same credentials should return the
 /// same user object.
 - (void)testUsernamePasswordAuthentication {
-    RLMSyncUser *firstUser = [self logInUserForCredentials:[RLMSyncTestCase basicCredentialsWithName:NSStringFromSelector(_cmd)
+    [self verifyUsernamePasswordAuthenticationWithUsername:NSStringFromSelector(_cmd) simulateReconnection:NO];
+}
+
+- (void)testUsernamePasswordAuthenticationWithReconnection {
+    [self verifyUsernamePasswordAuthenticationWithUsername:NSStringFromSelector(_cmd) simulateReconnection:YES];
+}
+
+- (void)verifyUsernamePasswordAuthenticationWithUsername:(NSString *)username simulateReconnection:(BOOL)simulateReconnection  {
+    RLMSyncUser *firstUser = [self logInUserForCredentials:[RLMSyncTestCase basicCredentialsWithName:username
                                                                                             register:YES]
-                                                    server:[RLMSyncTestCase authServerURL]];
-    RLMSyncUser *secondUser = [self logInUserForCredentials:[RLMSyncTestCase basicCredentialsWithName:NSStringFromSelector(_cmd)
+                                                    server:[RLMSyncTestCase authServerURL]
+                                      simulateReconnection:simulateReconnection];
+    RLMSyncUser *secondUser = [self logInUserForCredentials:[RLMSyncTestCase basicCredentialsWithName:username
                                                                                              register:NO]
-                                                     server:[RLMSyncTestCase authServerURL]];
+                                                     server:[RLMSyncTestCase authServerURL]
+                                       simulateReconnection:simulateReconnection];
     // Two users created with the same credential should resolve to the same actual user.
     XCTAssertTrue([firstUser.identity isEqualToString:secondUser.identity]);
     // Authentication server property should be properly set.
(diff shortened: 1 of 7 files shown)


austinzheng:
This looks great. Is it possible to create a script that completely disables and re-enables the network connection as well?

kishikawakatsumi:
@austinzheng Added `verify-osx-object-server-no-network` that is completely disables network connection. I'm not sure if this is what you are saying. Is my understanding is correct?

@austinzheng @jpsim I also modified `testUsernamePasswordAuthentication()`, `testAddObjects()` and `testDownloadRealm()` to split original test cases and confirming reconnect issues. Probably those would succeed if https://github.com/realm/realm-cocoa/pull/4957 is merged.

bdash:
Should this be moved to `S:Review`?

Realm/ObjectServerTests/RLMObjectServerTests.mm
```diff
@@ -446,11 +456,30 @@ - (void)testOpenRealmWithNormalCredentials {
...
+                simulateReconnection:(BOOL)simulateReconnection {
+    NSLog(@"isParent: %d", self.isParent);
+    RLMSyncUser *user = [self logInUserForCredentials:[RLMObjectServerTests basicCredentialsWithName:username
                                                                                             register:self.isParent]
-                                              server:[RLMObjectServerTests authServerURL]];
-    RLMRealm *realm = [self openRealmForURL:url user:user];
+                                               server:[RLMObjectServerTests authServerURL]];
+    NSLog(@"Login succeeded: %@", user.identity);
+    if (!self.isParent && simulateReconnection) {
+        [self disableNetworking];
```
austinzheng: I don't think this will disable networking for just the child process. Perhaps I'm missing something, but there's no process-specific state passed down to the shell script that calls `pfctl`, nor any argument passed to `pfctl` that would allow it to distinguish one process from another. (If I overlooked something, please let me know.)

austinzheng: Given that in this test the child process runs before the parent process, I think instead of waiting 15 seconds to re-enable the network connectivity, maybe you should just re-enable the network connectivity right before the child version of the test returns.

kishikawakatsumi:
@austinzheng @jpsim It turns out that `ReachabilityStatus` does not change when the network is disabled by Network Link Conditioner or the server is down. So a session opened while the server is down (or packets are lost due to some network problem) will never reconnect again.

https://github.com/realm/realm-cocoa/issues/5104

kishikawakatsumi:
@austinzheng This PR is ready for review again. Could you please review this if you get a chance?

austinzheng:
This PR looks good. I've restarted the CI job to get the object server tests to run again (the last time they failed was on the "child process" issue in https://github.com/realm/ci/issues/145), and if everything looks good I'll approve.

build.sh
```diff
@@ -13,6 +13,8 @@
 set -o pipefail
 set -e
 
+source ./scripts/bandwidth_throttling.sh
```
jpsim: the `bandwidth_throttling.sh` script doesn't appear to be used anywhere in `build.sh`, so why are we sourcing it here?

Realm/ObjectServerTests/RLMSyncTestCase.mm
```diff
@@ -314,7 +324,7 @@ - (void)waitForDownloadsForUser:(RLMSyncUser *)user
         XCTFail(@"Download waiter did not queue; session was invalid or errored out.");
         return;
     }
-    [self waitForExpectations:@[ex] timeout:20.0];
+    [self waitForExpectations:@[ex] timeout:60.0];
```
jpsim: What's this PR's effect on the total runtime of the test suite? The ROS tests are already pretty lengthy. Perhaps we should be conditionally running tests with degraded network connectivity? Much like we do with `build.sh verify-osx-encryption`.

scripts/disable_networking.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
```
jpsim: Seems like this script and `enable_networking.sh` could just be run directly from `bandwidth_throttling.sh` as different commands.

build.sh
kishikawakatsumi: You're right. Probably I had used it for debug purpose and forgot to delete it.

scripts/disable_networking.sh
kishikawakatsumi: 👍

Realm/ObjectServerTests/RLMSyncTestCase.mm
kishikawakatsumi: I see. I split the test into two, (normal and poor network condition).

kishikawakatsumi:
@austinzheng I split the tests into normal condition tests and poor network condition tests. Could you please check just in case?

https://github.com/realm/realm-cocoa/pull/4923#discussion_r132041430

Realm/ObjectServerTests/RLMSyncTestCase.mm
```diff
@@ -362,6 +380,26 @@ - (void)primeSyncManagerWithSemaphore:(dispatch_semaphore_t)semaphore {
...
+    task.arguments = @[@"./scripts/disable_networking.sh"];
+    [task launch];
+    [task waitUntilExit];
+}
+
+- (void)enableNetworkingAfter:(NSTimeInterval)secs {
+    dispatch_after(dispatch_time(DISPATCH_TIME_NOW, (int64_t)(secs * NSEC_PER_SEC)), dispatch_get_global_queue(DISPATCH_QUEUE_PRIORITY_HIGH, 0), ^{
+        NSTask *task = [[NSTask alloc] init];
+        task.launchPath = @"/bin/bash";
+        task.currentDirectoryPath = NSProcessInfo.processInfo.environment[@"PWD"];
```
bdash: Should this instead use `-[NSFileManager currentDirectoryPath]` rather than using the value of the environment variable? I don't think there's any guarantee that `PWD` is set to a meaningful value.

build.sh
```diff
@@ -703,7 +703,18 @@ case "$COMMAND" in
         ;;
 
     "test-osx-object-server")
-        xc "-scheme 'Object Server Tests' -configuration $CONFIGURATION -sdk macosx test"
+        xc "-scheme 'Object Server Tests' -configuration $CONFIGURATION -sdk macosx test \
+            -skip-testing:ObjectServerTests/RLMObjectServerTests/testUsernamePasswordAuthenticationWithReconnection \
+            -skip-testing:ObjectServerTests/RLMObjectServerTests/testAddObjectsWithReconnection \
+            -skip-testing:ObjectServerTests/RLMObjectServerTests/testDownloadRealmWithReconnection"
```
bdash: What's the purpose of these flags? Why do we want to skip these tests only when testing via `build.sh`?

scripts/bandwidth_throttling.sh
```diff
@@ -0,0 +1,26 @@
...
+
+set -o pipefail
+set -e
+
+######################################
+# Simulating Poor Network Connectivity
+######################################
+enable_bandwidth_throttling() {
+    # run inside sudo
+    sudo sh <<SCRIPT
```
bdash: How will this work on the default macOS configuration that requires a password be entered to run anything via `sudo`? What will happen when I run the object server tests on my local machine?

Issue #4926, opened 2017-05-03
Realm.create(_:value:update:) doesn't set relationships to nil when passed an Object


bdash:
As mentioned in https://github.com/realm/realm-cocoa/issues/4882#issuecomment-298830486, a call to `Realm.create(_:value:update:)` that is passed an instance of an `Object` subclass with a relationship set to `nil` doesn't update an existing object's relationship to `nil`, instead leaving it set to whatever its previous value was. `Realm.add(_:update:)` does set the relationship to `nil` in this instance.

The following test case demonstrates the problem:
```swift
func testUpdateLinkWithNil() {
    let realm = try! Realm()
    var object1: SwiftLinkToPrimaryStringObject! = nil
    var object2: SwiftLinkToPrimaryStringObject! = nil
    var object3: SwiftLinkToPrimaryStringObject! = nil
    try! realm.write {
        object1 = realm.create(SwiftLinkToPrimaryStringObject.self, value: ["key1", ["key1", 1]])
        object2 = realm.create(SwiftLinkToPrimaryStringObject.self, value: ["key2", ["key2", 2]])
        object3 = realm.create(SwiftLinkToPrimaryStringObject.self, value: ["key3", ["key3", 3]])
    }
    XCTAssertNotNil(object1.object)
    XCTAssertNotNil(object2.object)
    XCTAssertNotNil(object3.object)

    try! realm.write {
        realm.create(SwiftLinkToPrimaryStringObject.self, value: ["key1", nil], update: true)
        realm.create(SwiftLinkToPrimaryStringObject.self, value: ["pk": "key2", "object": nil], update: true)
        realm.create(SwiftLinkToPrimaryStringObject.self,
                     value: SwiftLinkToPrimaryStringObject(value: ["key3", nil]), update: true)
    }

    XCTAssertNil(object1.object)
    XCTAssertNil(object2.object)
    XCTAssertNil(object3.object)
}
```

Output:
```
Test Case '-[RealmSwift_Tests.ObjectCreationTests testUpdateLinkWithNil]' started.
RealmSwift/Tests/ObjectCreationTests.swift:382: error: -[RealmSwift_Tests.ObjectCreationTests testUpdateLinkWithNil] : XCTAssertNil failed: "SwiftPrimaryStringObject {
	stringCol = key3;
	intCol = 3;
}" - 
Test Case '-[RealmSwift_Tests.ObjectCreationTests testUpdateLinkWithNil]' failed (0.121 seconds).
```

[...]

PR #4934, opened 2017-05-08
Adopt the object store object accessor functionality

tgoyne:
Should be no functional changes here other than that some error messages have changed a bit (hopefully not for the worse) and a single bug has been fixed (the `testAssignIncorrectType` test crashes on master rather than throwing an exception and compiles without casts or warnings if the generic type isn't specified).

Needs a new core/sync release so CI will fail for now, but the code is reviewable.


Realm/RLMAccessor.mm
```diff
@@ -692,3 +562,295 @@ id RLMDynamicGetByName(__unsafe_unretained RLMObjectBase *const obj, __unsafe_un
...
+        }
+        else {
+            @throw RLMException(@"Object has been invalidated");
+        }
+    }
+
+    if (![link->_objectSchema.className isEqualToString:_info.rlmObjectSchema.className]) {
+        if (create && !_promote_existing)
+            return RLMCreateObjectInRealmWithValue(_realm, _info.rlmObjectSchema.className, link, update)->_row;
+        return link->_row;
```
bdash: I'm a little confused by the logic here. If the type of the passed-in object doesn't match the expected type, what good is it to return the passed-in object's row?

Realm/RLMAccessor.hpp
```diff
@@ -0,0 +1,116 @@
...
+#import "object_accessor.hpp"
+
+#import "RLMUtil.hpp"
+
+@class RLMRealm;
+class RLMClassInfo;
+class RLMObservationInfo;
+
+// realm::util::Optional<id> doesn't work due to that explicitly invoking the
+// destructor isn't allowed under ARC, so this covers the subset of Optional
```
bdash: Isn't the immediate problem that Obj-C types cannot be members of a union under ARC? Explicitly invoking the destructor of a variable of type `id` appears to be allowed, but doesn't do anything useful since the destructor isn't what handles releasing the object that the variable points to.

Realm/RLMAccessor.mm
tgoyne: The code doing things with the row validates that the row belongs to the correct table.

tgoyne:
osx-object-server failure is due to that there isn't an object server release which works with sync 1.9 and that's the only sync release with core 2.7.

Realm/RLMAccessor.hpp
tgoyne: Yes. I was mislead by the fact that the thing it happens to complain about is the destructor call, but that's just because that's the first time the value in the union is used and it complains on use rather than when the union is defined.

tgoyne:
Now passes on CI and the dependencies have been merged/released.

Realm/RLMClassInfo.mm
```diff
@@ -78,6 +78,11 @@
     return *m_linkTargets[propertyIndex];
 }
 
+RLMClassInfo &RLMClassInfo::linkTargetType(realm::Property const& property) {
+    REALM_ASSERT(property.type == PropertyType::Object || property.type == PropertyType::Array);
```
jpsim: would there be any reason for this to support linking objects?

Realm/RLMAccessor.hpp
```diff
@@ -0,0 +1,116 @@
...
+
+#import "object_accessor.hpp"
+
+#import "RLMUtil.hpp"
+
+@class RLMRealm;
+class RLMClassInfo;
+class RLMObservationInfo;
+
+// realm::util::Optional<id> doesn't work due to that Objective-C types can't
```
jpsim: "due to that" is a bit awkward. maybe just "doesn't work because Objective-C types..."

Realm/RLMAccessor.hpp
```diff
@@ -0,0 +1,116 @@
...
+struct RLMOptionalId {
+    id value;
+    RLMOptionalId(id value) : value(value) { }
+    explicit operator bool() const noexcept { return value; }
+    id operator*() const noexcept { return value; }
+};
+
+class RLMAccessorContext {
+public:
+    // Accessor context interfact
```
jpsim: "interface"

Realm/RLMAccessor.hpp
```diff
@@ -0,0 +1,116 @@
...
+    // rather than copying them
+    bool _promote_existing = true;
+    // Parent object of the thing currently being processed, for KVO purposes
+    __unsafe_unretained RLMObjectBase *const _parentObject = nil;
+
+    // Cached default values dictionary to avoid having to call the class method
+    // for every property
+    NSDictionary *_defaultValues;
+
+    bool _nilHack = false;
```
jpsim: what does this do? a comment explaining that would be helpful.

Realm/RLMClassInfo.mm
tgoyne: None of the places where `linkTargetType()` is used are operations applicable to backlinks.

Realm/RLMAccessor.hpp
tgoyne: It's just a hack to temporarily preserve the current behavior for #4926.

jpsim: right, comment probably not necessary then given that we'll be addressing that shortly.

PR #4935, opened 2017-05-08
Clarify documentation for `Realm.create()` method

### LINKED ISSUE #3020: Clarify how createOrUpdate methods merge with existing data
It doesn't seem very obvious from the name and the method documentation how createOrUpdate methods update related objects, especially for list properties. We should point out that:
- New objects will be created and added.
- If there is a primary key on the related object, all existing objects are updated with new data and remain associated.
- If there is no primary key on the related object, all objects will be created from scratch.
- Old related objects are removed, but kept in their tables. No reference counting / cascading delete behavior is applied.

### LINKED ISSUE #4904: Documentation for create(_:value:update:) is ambiguous
As you will see here: https://realm.io/docs/swift/2.6.2/api/Classes/Realm.html#/s:FC10RealmSwift5Realm6createuRxCS_6ObjectrFTMx5valueP_6updateSb_x

The documentation sates that the update parameter works such that:

> If true, the Realm will try to find an existing copy of the object (with the same primary key), and update it. Otherwise, the object will be added.

However, this can be interpreted in one of two ways and the correct unambiguous interpretation should be added to the docs.

Is it:

> If true, the Realm will try to find an existing copy of the object (with the same primary key), and update it such that any properties not provided in the value dictionary are not modified. Otherwise, the object will be added.

Or is it:

> If true, the Realm will try to find an existing copy of the object (with the same primary key), set it's properties, and delete any properties not provided within the value dictionary. Otherwise, the object will be added.

Thanks!

austinzheng:
Fixes #4904 
Fixes #3020

### DIFF
--- a/Realm/RLMObject.h
+++ b/Realm/RLMObject.h
@@ -139,13 +139,8 @@ NS_ASSUME_NONNULL_BEGIN
  The `value` argument can be a key-value coding compliant object, an array or dictionary returned from the methods in
  `NSJSONSerialization`, or an array containing one element for each managed property.
  
- If the object is being created, an exception will be thrown if any required properties are
- not present and those properties were not defined with default values.
-
- If the object is being updated and the `value` argument is a dictionary or a key-value
- coding compliant object, any property on the object that doesn't have a corresponding
- entry in the `value` argument will remain untouched. Nullable properties on the object can
- be set to nil by using `NSNull` as the updated value.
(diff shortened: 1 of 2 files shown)


austinzheng:
I'm going to have to investigate that object server test failure. This is the second time it's happened in the past few weeks.

austinzheng:
I think the changes you made are good and strictly increase the clarity of the documentation. I'll merge this if you have no objections.

PR #4943, opened 2017-05-09
Add changePassword:forUserID:completion:

jpsim:
Do not merge until https://github.com/realm/realm-object-server/pull/1061 is included in a ROS release.

### DIFF
--- a/Realm/ObjectServerTests/RLMObjectServerTests.mm
+++ b/Realm/ObjectServerTests/RLMObjectServerTests.mm
@@ -23,6 +23,7 @@
 #import "RLMSyncUtil_Private.h"
 #import "RLMRealmConfiguration_Private.h"
 #import "RLMRealmUtil.hpp"
+#import "RLMRealm_Dynamic.h"
 
 @interface RLMObjectServerTests : RLMSyncTestCase
 @end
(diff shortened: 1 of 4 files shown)


jpsim:
Addresses https://github.com/realm/product/issues/294.

Realm/ObjectServerTests/RLMObjectServerTests.mm
```diff
@@ -298,6 +299,117 @@ - (void)testUserChangePassword {
...
+        RLMSyncUser *adminTokenUser = [self logInUserForCredentials:credentials server:[RLMObjectServerTests authServerURL]];
+
+        // Create admin/admin "normal" user. Won't have admin priviledges until further below.
+        RLMSyncCredentials *creds = [RLMSyncCredentials credentialsWithUsername:@"admin" password:@"admin" register:YES];
+        RLMSyncUser *adminUser = [self logInUserForCredentials:creds server:[RLMObjectServerTests authServerURL]];
+        XCTAssertFalse(adminUser.isAdmin);
+        [adminUser logOut];
+
+        // User should be created very quickly but not necessarily instantly.
+        [[NSRunLoop currentRunLoop] runUntilDate:[NSDate dateWithTimeIntervalSinceNow:1.0]];
```
austinzheng: Is there any way this test can be changed so that it doesn't depend strictly upon the timer?

jpsim: That'd be pretty difficult AFAICT because I think the reason it's not instantaneous is that the transaction log has been successfully received but not necessarily applied, or the node.js runloop hasn't yet advanced to the latest state. There's no convenient way to hook into that from a client.

austinzheng:
I'm very curious as to why `testLogBackInSameRealmUpload` failed. That test hasn't been problematic for many months...

Realm/ObjectServerTests/RLMObjectServerTests.mm
austinzheng: Alright, that makes sense. Not sure we can do any better with our current toolkit.

jpsim:
`testLogBackInSameRealmUpload` failed because a previous test crashed the server, leading to inconsistent state for the rest of the run.

austinzheng:
Oh, I thought that test came first. I wish there were a way to abort the entire test run upon initial failure.

jpsim:
This is ready for review again.

jpsim:
Object Server CI test failures appear unrelated.

Issue #4905, opened 2017-04-26
Hook reachability into token refresh subsystem


austinzheng:
Right now the system that refreshes tokens just attempts to retry failed network requests nonstop (with some backoff delay), depending on the specific error.

~If reachability notes that there is no network connectivity, that subsystem should stop refreshing tokens.~ (see below) When connectivity is restored, that subsystem should refresh any outstanding tokens.

bdash:
I'm not sure it's wise to halt attempts to refresh just because reachability says there may not be connectivity (we'd need to be certain that we're using reachability data for the specific host / port we need to talk to). It's definitely worth using the transition back to having connectivity as a trigger for refreshes, though.

PR #4957, opened 2017-05-16
Fix issue where sync does not properly reconnect

austinzheng:
Requires https://github.com/realm/realm-object-store/pull/451
Fixes: #4905 
~Pending testing~

Changes:
- Update sync dependency to 1.10.0
- Use object equality comparison to check the error domain
- Update object store to pull in related fixes at that layer
- Also, remove a spurious print statement


austinzheng:
Looks like a bunch of CI machines are indisposed.

jpsim:
> Disabled permission offer/response tests broken by the core/sync upgrade (unrelated to this change)

Actually, it's the added call to `SyncManager::shared().cancel_reconnect_delay();` in `RLMSyncSessionRefreshHandle.mm` that's causing the permission tests to fail. Removing it makes all object server tests pass.

austinzheng:
There's no conceivable good reason why the added call should be triggering the assert that causes the tests to fail. Anyways I won't be able to look at this any further until I get back.

austinzheng:
Sorry, that last comment was a bit snippy. When I get access to my work machine again I'll investigate the problem further and see if it's something related to the models, object store, or if I can file a bug against sync with useful reproduction data.

austinzheng:
Worked around the issue (and improved the fix overall) through the most recent commit + https://github.com/realm/realm-object-store/pull/454.

austinzheng:
The reason all the tests are failing is because there's an object store-binding dependencies mismatch, and there are some code changes required to get both in sync.

kishikawakatsumi:
@austinzheng Should `REALM_SYNC_VERSION` on `realm-cocoa`'s `dependencies.list` to `1.9.0` as well as `object-store`'s one?

PR #4964, opened 2017-05-17
Move ownership of refresh handles to `SyncUser`

austinzheng:
This fixes a potential issue where, if a `RLMSyncUser` representing a given user is instantiated (whether by logging in or getting the current user), used to open Realms, and then destroyed, the refresh handles for those Realms are destroyed along with it, even if other `RLMSyncUser` instances representing the same user are still around (and the Realms are still open).

### DIFF
--- a/Realm/RLMSyncSessionRefreshHandle.mm
+++ b/Realm/RLMSyncSessionRefreshHandle.mm
@@ -30,16 +30,30 @@
 
 using namespace realm;
 
+namespace {
+void runBlockForUserContext(const std::weak_ptr<SyncUser>& user,
+                                const std::string& path,
+                                std::function<void(CocoaSyncUserContext&, const std::string&)> block) {
+    if (auto strong_user = user.lock()) {
+        strong_user->run_block_with_context<CocoaSyncUserContext>([block=std::move(block), path](auto& context) {
+            block(context, path);
+        });
+    }
+}
+}
+
 @interface RLMSyncSessionRefreshHandle () {
+    std::weak_ptr<SyncUser> _user;
+    std::string _path;
     std::weak_ptr<SyncSession> _session;
     std::shared_ptr<SyncSession> _strongSession;
 }
 
-@property (nonatomic, weak) RLMSyncUser *user;
-@property (nonatomic, strong) NSString *pathToRealm;
 @property (nonatomic) NSTimer *timer;
+@property (nonatomic) NSString *identity;
 
 @property (nonatomic) NSURL *realmURL;
+@property (nonatomic) NSURL *authServerURL;
 @property (nonatomic, copy) RLMSyncBasicErrorReportingBlock completionBlock;
 
 @end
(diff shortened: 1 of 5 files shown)


Realm/RLMSyncUser.mm
```diff
@@ -300,8 +313,14 @@ - (void)revokePermission:(RLMSyncPermissionValue *)permission callback:(RLMPermi
...
 #pragma mark - Private API
 
-- (void)_unregisterRefreshHandleForURLPath:(NSString *)path {
-    [self.refreshHandles removeObjectForKey:path];
+- (CocoaSyncUserContext *)_context {
+    if (!_user) {
+        return nullptr;
+    }
+    if (!_user->context) {
+        _user->context = std::make_unique<CocoaSyncUserContext>();
```
austinzheng: This needs to be synchronized, I think. What if two different `RLMSyncUser`s try to do this at once?

Realm/RLMSyncUser_Private.hpp
```diff
@@ -21,17 +21,37 @@
...
 
 using namespace realm;
 
 typedef void(^RLMFetchedRealmCompletionBlock)(NSError * _Nullable, RLMRealm * _Nullable, BOOL * _Nonnull);
 
 NS_ASSUME_NONNULL_BEGIN
 
+class CocoaSyncUserContext : public SyncUserContext {
+public:
+    CocoaSyncUserContext() = default;
```
austinzheng: This is unnecessary, get rid of it

austinzheng:
CI is failing as a result of the mismatch between object store and Cocoa sync dependencies. I'll look into it after the object store PR is merged.

jpsim:
Please ping me when this is ready for review again @austinzheng

Realm/RLMSyncUser.mm
```diff
@@ -69,6 +75,34 @@ PermissionGetCallback RLMWrapPermissionResultsCallback(RLMPermissionResultsBlock
     };
 }
 
+void runBlockForUserContext(const std::shared_ptr<SyncUser>& user, std::function<void(CocoaSyncUserContext&)> block) {
```
bdash: I think it'd be clearer to have a method that simply returns a `std::shared_ptr<CocoaSyncUserContext>` and then let callers use it, rather than having this function-based pattern.

Realm/RLMSyncUser.mm
```diff
@@ -305,8 +330,8 @@ - (void)revokePermission:(RLMSyncPermissionValue *)permission callback:(RLMPermi
 
 #pragma mark - Private API
 
-- (void)_unregisterRefreshHandleForURLPath:(NSString *)path {
-    [self.refreshHandles removeObjectForKey:path];
++ (void)_setupBindingContextFactory {
```
bdash: `setup` -> `setUp`, since it's a verb.

Realm/RLMSyncSessionRefreshHandle.mm
```diff
@@ -30,16 +30,28 @@
 
 using namespace realm;
 
+namespace {
+void runBlockForUserContext(const std::weak_ptr<SyncUser>& user,
```
bdash: This is only ever used to call `unregister_refresh_handle`, so it should probably just encapsulate that call at this point.

Realm/RLMSyncUser_Private.hpp
```diff
@@ -21,17 +21,37 @@
...
+    /**
+     A map of paths to 'refresh handles'.
+
+     A refresh handle is an object that encapsulates the concept of periodically
+     refreshing the Realm's access token before it expires. Tokens are indexed by their
+     paths (e.g. `/~/path/to/realm`).
+     */
+    std::unordered_map<std::string, RLMSyncSessionRefreshHandle *> m_refresh_handles;
+
+    mutable std::mutex m_mutex;
```
bdash: I don't think this needs to be mutable given there are no `const` member functions.

PR #4979, opened 2017-05-29
fix typo

1amageek:


jpsim:
Thanks for the tip, but your commit actually omits a necessary word. Fixed in 1b349e434.

PR #4983, opened 2017-05-30
Add support for email permissions

austinzheng:
cc @alazier 

I'm actually not sure any other work needs to be done, besides pulling in the correct object store dependency. I need to figure out how to test this, which is equivalent to figuring out how to build a version of ROS that supports this new feature.

Also, ideas for email validation welcome. I figure I'll just add in a very lenient regex check and have the initializer throw an exception if the provided email fails even that.

### DIFF
--- a/Realm/RLMSyncPermissionValue.mm
+++ b/Realm/RLMSyncPermissionValue.mm
@@ -60,6 +60,8 @@ RLMSyncAccessLevel objCAccessLevelForAccessLevel(Permission::AccessLevel level) 
 @interface RLMSyncPermissionValue () {
 @private
     NSString *_userID;
+    NSString *_key;
+    NSString *_value;
     util::Optional<Permission> _underlying;
     RLMSyncAccessLevel _accessLevel;
     NSString *_path;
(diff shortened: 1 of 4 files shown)


nirinchev:
Latest ROS should support permission change by email: https://github.com/realm/realm-object-server/releases/tag/v1.6.0

I'm not sure we should add email validation (on .NET I decided against it, but am open to revisit if there's compelling arguments). My reasoning is that the server doesn't reject user id's that are not emails (we regularly use guids), so it would be weird to end up in a case where you can't change permissions because your user ids are not valid emails.

austinzheng:
@nirinchev That's a good point. Maybe I should rename the API to not mention emails, since it looks like any sort of valid user identifier (that might just happen to be an email, or not) can be used.

nirinchev:
Yes, that was a bit confusing for me - technically we're changing permissions based on user id in the password provider. My guess was that we didn't choose Id to avoid confusion with provider-specific Ids from the other auth providers. Maybe @alazier can shed some light.

Realm/RLMSyncPermissionValue.mm
```diff
@@ -76,6 +78,27 @@ - (instancetype)initWithRealmPath:(NSString *)path
...
         _userID = userID;
+        if (!_userID) {
+            @throw RLMException(@"A permission value cannot be created without a valid user ID");
+        }
+        _updatedAt = [NSDate date];
+    }
+    return self;
+}
+
+- (instancetype)initWithRealmPath:(NSString *)path
```
cmelchior: Doesn't this have the exact same signature as `initWithRealmPath` just above that accepts userID? Couldn't that be a source of errors?

nirinchev: With ObjC, the signature of the method includes the argument names

austinzheng: Yeah, the effective method name is different (`initWithRealmPath:user:...`) versus (`initWithRealmPath:email`).

austinzheng:
This is ready for review, assuming that no relevant CI tests fail.

austinzheng:
The object server test that failed doesn't have anything to do with this PR. I'll investigate.

Realm/RLMSyncPermissionValue.mm
```diff
@@ -76,27 +78,41 @@ - (instancetype)initWithRealmPath:(NSString *)path
         _accessLevel = accessLevel;
         _path = path;
         _userID = userID;
+        if (!_userID) {
```
tgoyne: What's the rationale for doing a runtime check for just this parameter? Both strings are declared as nonnull, so I would expect either both or neither to be checked.

I'd also be inclined to write this as argument validation done before the init stuff rather than validation of the object's initialized state, but I won't try to pretend that's anything other than personal preference.

austinzheng: I think the `nonnull` declaration should be good enough to demonstrate to users what this initializer requires. I'll remove the check.

> I'd also be inclined to write this as argument validation done before the init stuff rather than validation of the object's initialized state

Yeah, I agree this is cleaner. I'll do that in the future.

PR #4987, opened 2017-05-31
Add integration tests for wildcard permissions

austinzheng:
Add tests for wildcard permissions, including a test suite for a global Realm. Also, break out the code JP wrote earlier to create an admin user into its own helper method.

~This requires a working OS X version of ROS with the appropriate fix before it can be merged.~

### DIFF
--- a/Realm/ObjectServerTests/RLMObjectServerTests.mm
+++ b/Realm/ObjectServerTests/RLMObjectServerTests.mm
@@ -302,13 +302,14 @@
 /// A sync user should be able to successfully change their own password.
 - (void)testOtherUserChangePassword {
     // Create admin user.
+    NSString *adminUsername = [[NSUUID UUID] UUIDString];
     {
         NSURL *url = [RLMObjectServerTests authServerURL];
-        RLMSyncUser *adminUser = [self makeAdminUser:@"admin" password:@"admin" server:url];
+        RLMSyncUser *adminUser = [self makeAdminUser:adminUsername password:@"admin" server:url];
         [adminUser logOut];
 
-        // Confirm that admin/admin user has admin privileges.
-        RLMSyncCredentials *creds = [RLMSyncCredentials credentialsWithUsername:@"admin"
+        // Confirm that admin user has admin privileges.
+        RLMSyncCredentials *creds = [RLMSyncCredentials credentialsWithUsername:adminUsername
                                                                        password:@"admin"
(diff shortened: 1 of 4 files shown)


austinzheng:
Requires #4989.

austinzheng:
Needs to be changed so that the wildcard global Realm user A is an admin user. Otherwise user A can't actually give out permissions.

austinzheng:
This is ready for review, but since it's just an enhancement to the test suite it shouldn't take higher priority over other PRs.

austinzheng:
Increasing the tolerance of the timeout (as part of instrumenting the offending method) seems to have fixed the problem.

Realm/ObjectServerTests/RLMSyncTestCase.m
```diff
@@ -195,6 +196,51 @@ - (RLMSyncUser *)logInUserForCredentials:(RLMSyncCredentials *)credentials
...
+    XCTAssertNotNil(adminTokenUser);
+
+    // Create a new user, which starts off without admin privileges.
+    RLMSyncCredentials *creds = [RLMSyncCredentials credentialsWithUsername:userName password:password register:YES];
+    RLMSyncUser *adminUser = [self logInUserForCredentials:creds server:url];
+    XCTAssertFalse(adminUser.isAdmin);
+
+    // FIXME: is there any way to do this without waiting X seconds?
+    // User should be created very quickly but not necessarily instantly.
+    [[NSRunLoop currentRunLoop] runUntilDate:[NSDate dateWithTimeIntervalSinceNow:5.0]];
```
bdash: A 5 second delay every time this method is called isn't great. Can we instead wait for some event that signals the user has been created? (Doesn't the call to `-logInUserForCredentials:server:` already do that through its use of `waitForExpectationsWithTimeout:handler:` internally?)

austinzheng: Yeah, I agree. I brought up the same issue with JP in the PR that introduced this code, and he didn't have a good solution. But now that this is a helper function, I want to think about whether there's a better way to do this.

`logInUser...` does wait, but for some reason the test gets stuck if it immediately proceeds without the additional delay. I'll investigate further.

austinzheng:
I figured out how to get rid of the "wait 5 seconds" in the helper method. If CI passes, I'll merge.

PR #5005, opened 2017-06-06
fixed issue when compiling RealmSwift via carthage using swift4

### LINKED ISSUE #5003: Carthage + Xcode 9: Build error
## Goals

Build Realm with Carthage and Xcode 9

## Expected Results

It builds without error.

## Actual Results

It fails with this error:

`Carthage/Checkouts/realm-cocoa/RealmSwift/RealmCollection.swift:35:38: error: cannot use mutating member on immutable value: 'generatorBase' is a 'let' constant
        let accessor = unsafeBitCast(generatorBase.next() as! Object?, to: Optional<T>.self)`

Realm framework version: Newest Master commit: 778017d76fb61699ac6c02979a350f0539889f8b

Xcode version: 9 Beta1

iOS/OSX version: iOS 11

Dependency manager + version: Carthage 0.23

nahive:
fixes #5003

### DIFF
--- a/RealmSwift/RealmCollection.swift
+++ b/RealmSwift/RealmCollection.swift
@@ -24,7 +24,7 @@ import Realm
  */
 public final class RLMIterator<T: Object>: IteratorProtocol {
     private var i: UInt = 0
-    private let generatorBase: NSFastEnumerationIterator
+    private var generatorBase: NSFastEnumerationIterator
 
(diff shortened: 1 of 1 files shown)


austinzheng:
Subsumed by https://github.com/realm/realm-cocoa/pull/5006.

PR #5006, opened 2017-06-06
Add support for Xcode 9

tgoyne:
Compiles but doesn't actually work yet (the Float <-> NSNumber conversions crash even when building as Swift 3.2).


tgoyne:
Now successfully builds and passes tests with Xcode 8.3 and Xcode 9 in both swift 3 and swift 4 modes, so I guess it's ready for review.

tgoyne:
Going to file a radar about the `NSNumber as! Float`.

austinzheng:
It looks like some of the CI builds (tvOS and watchOS) are failing to compile with errors, while other builds are just causing the compiler to segfault.

jpsim:
Should we standardize on `@objc` rather than `@objc dynamic`?

tgoyne:
`@objc` without `dynamic` doesn't work.

jpsim:
That's very good to know (and surprising)

austinzheng:
My impression was that `dynamic` prevents the compiler from allowing the method to be invoked by anything other than dynamic dispatch, while `@objc` doesn't make that same guarantee if dynamic dispatch isn't required (e.g. if the method is not actually ever invoked from Objective-C code).

rkim-homma:
awesome pull request. Thanks for rolling it in.

tgoyne:
Remaining CI failures appear unrelated (test failure in the sync progress notifier and failing to do anything at all in the device test).

cnstoll:
Any update on this? Should we build from TG's branch for Xcode 9?

marcuswu0814:
@cnstoll, I build branch by self using cocoapod and it work.

kashifshaikh:
Does that mean with Swift 4 we're going to have to add both @objc along with dynamic to all our properties?

Yuck.

tgoyne:
Yes, with Swift 4, exposing things to the obj-c runtime has been decoupled from disabling static dispatch and we need both parts.

Property behaviors will *hopefully* make it possible for us to switch to a pure-Swift solution and not need `@objc` or `dynamic`, but those aren't coming in Swift 4.

RealmSwift/ThreadSafeReference.swift
```diff
@@ -84,7 +83,6 @@ public class ThreadSafeReference<Confined: ThreadConfined> {
     public init(to threadConfined: Confined) {
         let bridged = (threadConfined as! AssistedObjectiveCBridgeable).bridged
         swiftMetadata = bridged.metadata
-        type = type(of: threadConfined)
```
ddevjani: how to resolve error

PR #5024, opened 2017-06-12
Update object store and bump version to 2.8.1

tgoyne:


austinzheng:
I'm going to try to fix the problematic object server test today.

PR #5047, opened 2017-06-19
adding xcode 9 beta to the test mix

larkost:


larkost:
@tgoyne Could you look at the failures in this 9.0 job? I think the answer is just to add `@objc` to the qualifiers, but I don't know what other side effects that would have.

https://ci.realm.io/job/objc_pr/configuration=Release,target=tvos,xcode_version=9.0/5232/console

tgoyne:
Adding `@objc` to all the `dynamic` properties is correct. Probably a bunch of Swift examples that I forgot to check for that.

larkost:
Closing in favor of pull #5101 due to the rebases here messing GitHub up.

PR #5056, opened 2017-06-21
Change how synced Realm permission errors are handled

### LINKED ISSUE #5043: Remove local copies of synced Realms if 403
Doing things like trying to synchronously open a Realm that you only have read permissions for puts Realm in a bad state: there's a file on disk that Realm can't do anything with, even if you try to properly `asyncOpen()` the Realm later.

We should repurpose the client reset mechanism to allow such files to be deleted. cc @bigfish24

austinzheng:
Fixes #5043
Requires https://github.com/realm/realm-object-store/pull/476

Changes:
- Permission denied errors now have their own code, as well as a block for recovery purposes
- Sessions that encounter a permission denied error are now closed immediately

### DIFF
--- a/Realm/NSError+RLMSync.h
+++ b/Realm/NSError+RLMSync.h
@@ -32,6 +32,13 @@ NS_ASSUME_NONNULL_BEGIN
 - (nullable void(^)(void))rlmSync_clientResetBlock NS_REFINED_FOR_SWIFT;
 
 /**
+ Given a Realm Object Server permission denied error, return the block that
+ can be called to manually initiate or cancel the Realm file deletion process,
+ or nil if the error isn't a permission denied error.
+ */
+- (nullable void(^)(BOOL))rlmSync_deleteRealmBlock NS_REFINED_FOR_SWIFT;
+
+/**
  Given a Realm Object Server client reset error, return the path where the
  backup copy of the Realm will be placed once the client reset process is
  complete.
--- a/Realm/NSError+RLMSync.h
+++ b/Realm/NSError+RLMSync.h
@@ -35,6 +35,13 @@ NS_ASSUME_NONNULL_BEGIN
  Given a Realm Object Server permission denied error, return the block that
  can be called to manually initiate or cancel the Realm file deletion process,
  or nil if the error isn't a permission denied error.
+
+ The block itself takes a single boolean argument. Pass in YES to immediately
+ delete the files on disk (after all references to the Realm and objects in
+ the Realm have been invalidated). Pass in NO to never delete the Realm files.
+ The block can only be called once. If the block isn't called at all, the
+ Realm files will be deleted the next time your application is launched and the
+ sync subsystem is initialized.
  */
 - (nullable void(^)(BOOL))rlmSync_deleteRealmBlock NS_REFINED_FOR_SWIFT;
 
--- a/Realm/ObjectServerTests/RLMPermissionsAPITests.m
+++ b/Realm/ObjectServerTests/RLMPermissionsAPITests.m
@@ -198,9 +198,7 @@ static RLMSyncPermissionValue *makeExpectedPermission(RLMSyncPermissionValue *or
     workBlock = ^(NSError *err) {
         // Expect an error from the global error handler.
         XCTAssertNotNil(err);
-        XCTAssertEqual(err.code, RLMSyncErrorClientSessionError);
-        // 206 is a permission error. TODO: don't hardcode the literal.
-        XCTAssertEqual([err.userInfo[kRLMSyncErrorStatusCodeKey] integerValue], 206);
+        XCTAssertEqual(err.code, RLMSyncErrorPermissionDeniedError);
         [deniedEx fulfill];
     };
     [self addSyncObjectsToRealm:userBRealm descriptions:@[@"child-4", @"child-5", @"child-6"]];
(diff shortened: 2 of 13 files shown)


Realm/ObjectServerTests/RLMPermissionsAPITests.m
```diff
@@ -804,4 +802,103 @@ - (void)testSettingUnownedRealmPermission {
...
 }
 
+#pragma mark - Delete Realm upon permission denied
+
+/// A Realm which is opened improperly should report an error allowing the app to recover.
+- (void)testDeleteRealmUponPermissionDenied {
+    __block void(^workBlock)(NSError *, NSURL *url) = ^(NSError *err, __unused NSURL *url) {
+        XCTFail(@"Error handler should not be called unless explicitly expected. Error: %@", err);
+    };
+    [[RLMSyncManager sharedManager] setErrorHandler:^(NSError *error, RLMSyncSession *session) {
```
tgoyne: Why are these tests mutating a local variable captured by this block rather than just changing the errorHandler block?

tgoyne: This wrapper would make sense if it was instead something like the following:

```
__block void (^errorBlock)(NSError *, NSURL *url) = nil;
RLMSyncManager.sharedManager.errorHandler = ^(NSError *error, RLMSyncSession *session) {
    if (errorBlock) {
        errorBlock(error, session.realmURL);
        errorBlock = nil;
    }
    else {
        XCTFail(@"Error handler should not be called unless explicitly expected. Error: %@", err);
    }
};
```

This would remove the need to manually reset the block to the error case after each check.

Realm/NSError+RLMSync.h
```diff
@@ -32,6 +32,13 @@ NS_ASSUME_NONNULL_BEGIN
 - (nullable void(^)(void))rlmSync_clientResetBlock NS_REFINED_FOR_SWIFT;
 
 /**
+ Given a Realm Object Server permission denied error, return the block that
+ can be called to manually initiate or cancel the Realm file deletion process,
+ or nil if the error isn't a permission denied error.
+ */
```
tgoyne: This needs to say what the parameter to the block is.

Realm/ObjectServerTests/RLMPermissionsAPITests.m
```diff
@@ -804,4 +802,103 @@ - (void)testSettingUnownedRealmPermission {
...
+            XCTFail(@"Error handler should not be called unless explicitly expected. Error: %@", err);
+        };
+    }
+
+    // Check the error and perform the Realm deletion.
+    XCTAssertNotNil(theError);
+    XCTAssertNotNil([theError rlmSync_deleteRealmBlock]);
+    [theError rlmSync_deleteRealmBlock](YES);
+
+    // Incorrectly open the Realm for user B _again_.
```
tgoyne: Why? It's not obvious to me what doing it again is testing.

Realm/ObjectServerTests/RLMPermissionsAPITests.m
```diff
@@ -804,4 +802,103 @@ - (void)testSettingUnownedRealmPermission {
...
+        [self waitForExpectationsWithTimeout:10.0 handler:nil];
+        workBlock = ^(NSError *err, __unused NSURL *url) {
+            XCTFail(@"Error handler should not be called unless explicitly expected. Error: %@", err);
+        };
+    }
+
+    // Check the error and perform the Realm deletion.
+    XCTAssertNotNil(theError);
+    XCTAssertNotNil([theError rlmSync_deleteRealmBlock]);
+    [theError rlmSync_deleteRealmBlock](YES);
```
tgoyne: Verify that the file is actually gone from disk?

Realm/NSError+RLMSync.h
tgoyne: Attaching functionality to a NSError like this is really weird and unusual, but I guess since client reset is already done that way this should as well (at least for now).

Realm/ObjectServerTests/RLMPermissionsAPITests.m
tgoyne: A two iteration for-loop wrapping this would be better than duplicating it.

Realm/RLMSyncManager.mm
```diff
@@ -182,6 +182,27 @@ - (void)_fireErrorWithCode:(int)errorCode
...
+            custom = @{kRLMSyncInitiateDeleteRealmBlockKey: ^(BOOL deleteRealm) {
+                NSString *originalPath = userInfo[@(realm::SyncError::c_original_file_path_key)];
+                if (calledAlready) {
+                    @throw RLMException(@"The handler block for the Realm at '%@' has already been called once.",
+                                        originalPath);
+                }
+                calledAlready = YES;
+                std::string original_path = [originalPath UTF8String];
+                if (deleteRealm) {
+                    NSLog(@"Deleting Realm...");
```
tgoyne: We shouldn't log anything during normal operation.

RealmSwift/Sync.swift
```diff
@@ -122,6 +122,15 @@ public extension SyncError {
         }
         return nil
     }
+
+    /// Given a permission denied error, extract and return the reset closure.
+    public func deleteRealmUserInfo() -> ((Bool) -> Void)? {
+        if code == SyncError.permissionDeniedError,
+            let block = _nsError.__rlmSync_deleteRealmBlock() {
```
tgoyne: Both parts of this check seem unnecessary.

Realm/NSError+RLMSync.h
austinzheng: Yes, I hope we can revisit the design for 3.0.

Realm/ObjectServerTests/RLMPermissionsAPITests.m
austinzheng: Yes, testing the same thing twice doesn't make sense.

Realm/ObjectServerTests/RLMPermissionsAPITests.m
austinzheng: This suggestion makes far more sense than what I was doing before.

austinzheng:
Device job failure is a CI issue. Swift job failure is a single unrelated Swift test that has been sporadically failing across various PRs.

Realm/RLMSyncUtil.h
```diff
@@ -118,6 +121,30 @@ typedef RLM_ERROR_ENUM(NSInteger, RLMSyncError, RLMSyncErrorDomain) {
...
+     permissions to without using the `asyncOpen()` APIs.
+
+     A Realm that suffers a permission denied error is, by default, flagged so that its
+     local copy will be deleted the next time the application starts.
+     
+     The `userInfo` dictionary contains a block under the key
+     `kRLMSyncInitiateDeleteRealmBlockKey`. This block can be called with a single argument:
+     `YES` to immediately delete the Realm file, `NO` to not delete the file at all (either
+     now or upon restart). This block should only be called with `YES` if and when your app
+     closes and invalidates every instance of the offending Realm on all threads (note that
```
bdash: Some of the wording here is a little odd. Perhaps "should only be called … after your app has closed and invalidated every instance…"

Realm/RLMSyncUtil.h
```diff
@@ -118,6 +121,30 @@ typedef RLM_ERROR_ENUM(NSInteger, RLMSyncError, RLMSyncErrorDomain) {
...
+     This error may also occur if a user incorrectly opens a Realm they have read-only
+     permissions to without using the `asyncOpen()` APIs.
+
+     A Realm that suffers a permission denied error is, by default, flagged so that its
+     local copy will be deleted the next time the application starts.
+     
+     The `userInfo` dictionary contains a block under the key
+     `kRLMSyncInitiateDeleteRealmBlockKey`. This block can be called with a single argument:
+     `YES` to immediately delete the Realm file, `NO` to not delete the file at all (either
+     now or upon restart). This block should only be called with `YES` if and when your app
```
bdash: It seems like there's three possible actions that can be taken:
1. Delete the file now.
2. Delete the file on next launch.
3. Don't delete the file.

But we're using a boolean to select between only two of them?

It's also weird that calling the "initiate delete realm block" with `NO` doesn't initiate a deletion at all.

austinzheng: I think that I should just get rid of option 3, hence get rid of the boolean argument. Then the semantics will be exactly the same as with client reset.

PR #5065, opened 2017-06-23
Reapply #5056

austinzheng:
Apologies for the second PR. I merged #5056 by accident instead of the other PR I had intended, and had to back it out.

The first commit is simply #5056's changes, but with the proper object store pointer.

The second commit attempts to address @bdash's feedback [here](https://github.com/realm/realm-cocoa/pull/5056#pullrequestreview-45861924).

Let me know if there are any questions.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,11 @@ x.x.x Release notes (yyyy-MM-dd)
 
 ### Enhancements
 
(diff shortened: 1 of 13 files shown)


austinzheng:
The failing object server test is unrelated. I'm trying to reproduce locally so I can investigate further.

PR #5068, opened 2017-06-24
Support stable IDs

### LINKED ISSUE #4909: Support Stable IDs
Support stable IDs as provided by upstream. Unfortunately, we cannot wait for a proper release of upstream packages for work on this to begin.

There should be no change in user-facing APIs or behavior.

Corresponding issue in realm-dotnet: https://github.com/realm/realm-dotnet/issues/1371
Corresponding issue in realm-java: https://github.com/realm/realm-java/issues/4542
Corresponding issue in realm-js: https://github.com/realm/realm-js/issues/985

Upstream PR: https://github.com/realm/realm-sync/pull/1170

Object Store PR: https://github.com/realm/realm-object-store/pull/434

bdash:
This PR is currently for sake of having CI chew on the tests.

Fixes #4909.


bdash:
I'm seeing an occasional deadlock when running the object server tests locally. The main thread is waiting on notifications to be delivered as part of beginning a write transaction, but the notifier thread it is waiting on is itself trying to open a write transaction. I don't think we ever expect the notifier thread to create a write transaction. In this case, the write transaction is being created by core within `SharedGroup::upgrade_file_format`. I'm not sure why we're attempting to upgrade the file format of a Realm that has already been opened.

bdash:
Filed realm/realm-core#2724 about the deadlock.

PR #5080, opened 2017-06-28
Make `Object` comparison behavior match `RLMObject` comparison behavior

### LINKED ISSUE #4705: Conformance to Hashable is broken for objects without primary key
Classes deriving from `Object` appear to automatically inherit conformance to `Hashable` due to having `NSObject` as a base class. The implementation of `hashValue` comes from `-[RLMObjectBase hash]`, by way of `NSObject`'s `Hashable` conformance delegating to the object's `hash` property. `-[RLMObjectBase hash]` isn't correct with respect to `-[RLMObjectBase isEqual:]` for objects that don't have a primary key declared, `-hash` will return different values for two objects that compare as equal, which violates the assumptions about how `Hashable` objects are used.

austinzheng:
Breaking change for 3.0
Fixes #4705

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -54,6 +54,8 @@ x.x.x Release notes (yyyy-MM-dd)
 * Swift `Object` comparison and hashing behavior now works the same way as
   that of `RLMObject` (objects are now only considered equatable if their
   model class defines a primary key).
+* Fix the way the hash property works on `Object` when the object model has
+  no primary key.
 
 ### Enhancements
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -51,10 +51,14 @@ x.x.x Release notes (yyyy-MM-dd)
 * Fix empty strings, binary data, and null on the right side of `BEGINSWITH`,
   `ENDSWITH` and `CONTAINS` operators in predicates to match Foundation's
   semantics of never matching any strings or data.
+* Swift `Object` comparison and hashing behavior now works the same way as
+  that of `RLMObject` (objects are now only considered equatable if their
+  model class defines a primary key).
 
 ### Enhancements
 
-* None.
+* Add Swift `Object.isSameObject(as:_)` API to perform the same function as
+  the existing Objective-C API `-[RLMObject isEqualToObject:]`.
 
 ### Bugfixes
 
--- a/RealmSwift/Object.swift
+++ b/RealmSwift/Object.swift
@@ -266,17 +266,22 @@ open class Object: RLMObjectBase, ThreadConfined {
                                    to: List<DynamicObject>.self)
     }
 
-    // MARK: Equatable
-
+    // MARK: Comparison
     /**
-     Returns whether two Realm objects are equal.
+     Returns whether two Realm objects are the same.
 
-     Objects are considered equal if and only if they are both managed by the same Realm and point to the same
-     underlying object in the database.
+     Objects are considered the same if and only if they are both managed by the same
+     Realm and point to the same underlying object in the database.
+     
+     - note: Equality comparison is implemented by `isEqual(_:)`, but objects are only
+             considered equatable using `==` or `isEqual(_:)` if a primary key is
+             defined for the model type. This method can be used to carry out the same
+             underlying comparison regardless of whether the object types are defined
+             with primary keys or not.
 
      - parameter object: The object to compare the receiver to.
      */
-    open override func isEqual(_ object: Any?) -> Bool {
+    public func isSameObject(as object: Any?) -> Bool {
         return RLMObjectBaseAreEqual(self as RLMObjectBase?, object as? RLMObjectBase)
     }
 
@@ -492,3 +497,15 @@ extension Object: AssistedObjectiveCBridgeable {
         return (objectiveCValue: unsafeCastToRLMObject(), metadata: nil)
     }
 }
+
+// MARK: - Migration assistance
+
+#if os(OSX)
+#else
+extension Object {
+    /// :nodoc:
+    @available(*, unavailable, renamed: "isSameObject(as:)") public func isEqual(to object: Any?) -> Bool {
+        fatalError()
+    }
+}
+#endif
--- a/RealmSwift/Object.swift
+++ b/RealmSwift/Object.swift
@@ -273,11 +273,12 @@ open class Object: RLMObjectBase, ThreadConfined {
      Objects are considered the same if and only if they are both managed by the same
      Realm and point to the same underlying object in the database.
      
-     - note: Equality comparison is implemented by `isEqual(_:)`, but objects are only
-             considered equatable using `==` or `isEqual(_:)` if a primary key is
-             defined for the model type. This method can be used to carry out the same
-             underlying comparison regardless of whether the object types are defined
-             with primary keys or not.
+     - note: Equality comparison is implemented by `isEqual(_:)`. If the object type
+             is defined with a primary key, `isEqual(_:)` behaves identically to this
+             method. If the object type is not defined with a primary key,
+             `isEqual(_:)` uses the `NSObject` behavior of comparing object identity.
+             This method can be used to compare two objects for database equality
+             whether or not their object type defines a primary key.
 
      - parameter object: The object to compare the receiver to.
      */
--- a/RealmSwift/Object.swift
+++ b/RealmSwift/Object.swift
@@ -282,8 +282,8 @@ open class Object: RLMObjectBase, ThreadConfined {
 
      - parameter object: The object to compare the receiver to.
      */
-    public func isSameObject(as object: Any?) -> Bool {
-        return RLMObjectBaseAreEqual(self as RLMObjectBase?, object as? RLMObjectBase)
+    public func isSameObject(as object: Object?) -> Bool {
+        return RLMObjectBaseAreEqual(self, object)
     }
 
     // MARK: Private functions
--- a/RealmSwift/Tests/RealmCollectionTypeTests.swift
+++ b/RealmSwift/Tests/RealmCollectionTypeTests.swift
@@ -222,8 +222,8 @@ class RealmCollectionTypeTests: TestCase {
         guard let collection = collection else {
             fatalError("Test precondition failed")
         }
-        XCTAssertEqual(str1, collection[0])
-        XCTAssertEqual(str2, collection[1])
+        assertEqual(str1, collection[0])
+        assertEqual(str2, collection[1])
 
         assertThrows(collection[200])
         assertThrows(collection[-200])
@@ -233,8 +233,8 @@ class RealmCollectionTypeTests: TestCase {
         guard let collection = collection else {
             fatalError("Test precondition failed")
         }
-        XCTAssertEqual(str1, collection.first!)
-        XCTAssertEqual(str2, collection.filter("stringCol = '2'").first!)
+        assertEqual(str1, collection.first!)
+        assertEqual(str2, collection.filter("stringCol = '2'").first!)
         XCTAssertNil(collection.filter("stringCol = '3'").first)
     }
 
@@ -242,8 +242,8 @@ class RealmCollectionTypeTests: TestCase {
         guard let collection = collection else {
             fatalError("Test precondition failed")
         }
-        XCTAssertEqual(str2, collection.last!)
-        XCTAssertEqual(str2, collection.filter("stringCol = '2'").last!)
+        assertEqual(str2, collection.last!)
+        assertEqual(str2, collection.filter("stringCol = '2'").last!)
         XCTAssertNil(collection.filter("stringCol = '3'").last)
     }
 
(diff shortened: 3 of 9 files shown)


RealmSwift/Object.swift
```diff
@@ -273,12 +273,15 @@ open class Object: RLMObjectBase, ThreadConfined {
      Returns whether two Realm objects are equal.
 
      Objects are considered equal if and only if they are both managed by the same Realm and point to the same
-     underlying object in the database.
+     underlying object in the database. Object comparison only takes place if the object type has a primary key.
```
bdash: I'm not sure what this extra sentence is trying to say?

RealmSwift/Object.swift
```diff
@@ -273,12 +273,15 @@ open class Object: RLMObjectBase, ThreadConfined {
...
 
      - parameter object: The object to compare the receiver to.
      */
-    open override func isEqual(_ object: Any?) -> Bool {
-        return RLMObjectBaseAreEqual(self as RLMObjectBase?, object as? RLMObjectBase)
+    open override func isEqual(to object: Any?) -> Bool {
+        if let object = object as? Object {
+            return RLMObjectBaseAreEqual(self, object)
```
bdash: Shouldn't this only be calling `RLMObjectBaseAreEqual` if the object has a primary key?

The logic in `-[RLMObjectBase isEqual:]` looks like so:
```objc
- (BOOL)isEqual:(id)object {
    if (RLMObjectBase *other = RLMDynamicCast<RLMObjectBase>(object)) {
        if (_objectSchema.primaryKeyProperty) {
            return RLMObjectBaseAreEqual(self, other);
        }
    }
    return [super isEqual:object];
}
```

bdash: Maybe this should just be delegating to `-[RLMObjectBase isEqual:]` rather than reimplementing it?

tgoyne:
There needs to be tests which actually verify the new behavior.

RealmSwift/Object.swift
austinzheng: This is supposed to be equivalent to `-[RLMObject isEqualToObject:]`, which is defined explicitly for `RLMObject` and not at all for `RealmSwift.Object`.

bdash: But isn't `isEqual(to:)` what ends up being called by `Equatable`? Surely the semantics therefore need to match those of `-isEqual:`, not our own `-isEqualToObject:`?

bdash: Oh, I see, you've renamed this from `isEqual(_:)`, which was shadowing `RLMObjectBase`'s implementation, to `isEqual(to:)`, which is a _new_ method.

austinzheng: I'm checking it right now, but I'm almost certain `Equatable`'s `==` invokes `isEqual(_ to:)`, not `isEqual(to:)`. Horrendously confusing, unfortunately.

RealmSwift/Object.swift
bdash: This new sentence seems incorrect, given the implementation below. `RLMObjectBaseAreEqual` doesn't care if objects have primary keys.

RealmSwift/Object.swift
bdash: `-[RLMObject isEqualToObject:]` is implemented like so:
```objc
return [object isKindOfClass:RLMObject.class] && RLMObjectBaseAreEqual(self, object);
```

The new behavior doesn't appear to be equivalent to that (for one, it calls through to `super.isEqual()` if it's passed a non-`Object`). The old implementation that you're replacing looks like it provides the correct behavior?

austinzheng: I believe you are mostly right. Instead of deferring to `super.isEqual()`, it should just return false. However, the old implementation (e.g. the former implementation of `isEqual(_:)`) will throw an Obj-C exception if the other object isn't an `RLMObject`, rather than returning false. I'll fix that.

bdash: How so? The implementation passed `object as? RLMObjectBase` for the argument, which will return nil in that case.

RealmSwift/Object.swift
austinzheng: Yes, the sentence should just be excised.

RealmSwift/Object.swift
austinzheng: Oh, you're right. I didn't look at the casts inside the function call carefully enough.

RealmSwift/Object.swift
```diff
@@ -277,9 +277,15 @@ open class Object: RLMObjectBase, ThreadConfined {
 
      - parameter object: The object to compare the receiver to.
      */
-    open override func isEqual(_ object: Any?) -> Bool {
+#if os(macOS)
```
bdash: Hmm. Is this necessary due to the weird scripting category that adds methods to NSObject on Mac?

I wonder if we should give this a name that more clearly distinguishes it from `isEqual(_:)`?

RealmSwift/Tests/TestCase.swift
```diff
@@ -113,6 +113,24 @@ class TestCase: XCTestCase {
         queue.sync { }
     }
 
+    /// Check whether two test objects are equal (refer to the same row in the same Realm), even if their models
+    /// don't define a primary key.
+    func assertEqual(_ o1: Object?, _ o2: Object?, fileName: String = #file, lineNumber: UInt = #line) {
+        RLMAssertEqualTestObjects(self, o1, o2, fileName, lineNumber)
```
bdash: Couldn't this be implemented in terms of `Object.isEqual(to:)`, rather than calling into Obj-C and using `RLMObjectBaseAreEqual`?

RealmSwift/Object.swift
austinzheng: Yes, it's because of that horrible, horrible method + Swift's `override` annotation requirement.

I'd be alright with giving it a new name, as long as we're aware that doing so will break a bit of symmetry with `RLMObject` (not sure that is even an issue at this point).

RealmSwift/Tests/TestCase.swift
austinzheng: Yeah, that's a good idea.

RealmSwift/Tests/TestUtils.h
```diff
@@ -19,18 +19,24 @@
 #import <Foundation/Foundation.h>
 #import <XCTest/XCTestCase.h>
 
+@class RLMObjectBase;
```
austinzheng: Get rid of this forward declaration, and probably all the changes in this file.

austinzheng:
The first four job failures are known CI issues. The object server tests are failing on that execrable child process issue.

austinzheng:
This is ready for review.

RealmSwift/Tests/TestCase.swift
```diff
@@ -113,6 +113,29 @@ class TestCase: XCTestCase {
...
+            inFile: fileName, atLine: lineNumber, expected: false)
+    }
+
+    /// Check whether two collections containing Realm objects are equal.
+    func assertEqualObjectCollections<T: Collection, U: Collection>(_ c1: T, _ c2: U, fileName: String = #file, lineNumber: UInt = #line)
+        where T.Iterator.Element : Object,
+        U.Iterator.Element : Object,
+        T.IndexDistance : Equatable,
+        T.IndexDistance == U.IndexDistance {
+        XCTAssertEqual(c1.count, c2.count, "Collection counts were incorrect")
```
jpsim: this should forward the `fileName` and `lineNumber` parameters.

RealmSwift/Tests/TestCase.swift
```diff
@@ -113,6 +113,29 @@ class TestCase: XCTestCase {
         queue.sync { }
     }
 
+    /// Check whether two test objects are equal (refer to the same row in the same Realm), even if their models
+    /// don't define a primary key.
+    func assertEqual(_ o1: Object?, _ o2: Object?, fileName: String = #file, lineNumber: UInt = #line) {
+        if o1?.isSameObject(as: o2) ?? false {
```
jpsim: if `o1` is `nil` but `o2` is not, `assertEqual` won't trigger an XCTest failure.

CHANGELOG.md
```diff
@@ -33,10 +33,14 @@ x.x.x Release notes (yyyy-MM-dd)
 * The return types of the `SyncError.clientResetInfo()` and
   `SyncError.deleteRealmUserInfo()` APIs have been changed. They now return
   `RLMSyncErrorActionToken`s or `SyncError.ActionToken`s instead of closures.
+* Swift `Object` comparison and hashing behavior now works the same way as
+  that of `RLMObject` (objects are now only considered equatable if their
+  model class defines a primary key).
```
bdash: I'm a little confused by this. Objects without primary keys can still be compared, it's just based on object identity in that case.

austinzheng: Yes, this should be changed. NSObjects are always `Equatable` according to Swift; the change now is that we don't break the contract for hash values anymore.

RealmSwift/Object.swift
```diff
@@ -266,17 +266,22 @@ open class Object: RLMObjectBase, ThreadConfined {
...
-     Returns whether two Realm objects are equal.
+     Returns whether two Realm objects are the same.
 
-     Objects are considered equal if and only if they are both managed by the same Realm and point to the same
-     underlying object in the database.
+     Objects are considered the same if and only if they are both managed by the same
+     Realm and point to the same underlying object in the database.
+     
+     - note: Equality comparison is implemented by `isEqual(_:)`, but objects are only
+             considered equatable using `==` or `isEqual(_:)` if a primary key is
```
bdash: Again, this comment doesn't seem to match the implementation. Objects without a primary key can be compared, they're just compared using object identity rather than checking if they refer to the same underlying object in the Realm.

RealmSwift/Object.swift
```diff
@@ -266,17 +266,23 @@ open class Object: RLMObjectBase, ThreadConfined {
...
+             is defined with a primary key, `isEqual(_:)` behaves identically to this
+             method. If the object type is not defined with a primary key,
+             `isEqual(_:)` uses the `NSObject` behavior of comparing object identity.
+             This method can be used to compare two objects for database equality
+             whether or not their object type defines a primary key.
 
      - parameter object: The object to compare the receiver to.
      */
-    open override func isEqual(_ object: Any?) -> Bool {
+    public func isSameObject(as object: Any?) -> Bool {
```
jpsim: Should this take an `object: Self?` rather than `Any?`?

RealmSwift/Tests/RealmCollectionTypeTests.swift
```diff
@@ -689,7 +689,7 @@ class ResultsWithCustomInitializerTest: TestCase {
         let expected = Array(collection.map { $0.stringCol })
         let actual = collection.value(forKey: "stringCol") as! [String]!
         XCTAssertEqual(expected, actual!)
-        XCTAssertEqual(collection.map { $0 }, collection.value(forKey: "self") as! [CTTStringObjectWithLink])
+        assertEqualObjectCollections(collection.map { $0 }, collection.value(forKey: "self") as! [SwiftCustomInitializerObject])
```
jpsim: How was this test passing before?

Unrelated to this PR, but we should rename `ResultsWithCustomInitializerTest` to `ResultsWithCustomInitializerTests` to match our test case naming convention.

austinzheng: May as well take the opportunity to rename it now.

Not sure why that test didn't fail before. Probably something to do with how XCTAssertEqual is implemented? The more I look at the old version of this test case the more confused I am.

RealmSwift/Tests/TestCase.swift
```diff
@@ -113,6 +113,32 @@ class TestCase: XCTestCase {
         queue.sync { }
     }
 
+    /// Check whether two test objects are equal (refer to the same row in the same Realm), even if their models
+    /// don't define a primary key.
+    func assertEqual(_ o1: Object?, _ o2: Object?, fileName: StaticString = #file, lineNumber: UInt = #line) {
```
jpsim: Some suggestions:

```diff
-    func assertEqual(_ o1: Object?, _ o2: Object?, fileName: StaticString = #file, lineNumber: UInt = #line) {
+    func assertEqual<O: Object>(_ o1: O?, _ o2: O?, fileName: StaticString = #file, lineNumber: UInt = #line) {
         if o1 == nil && o2 == nil {
             return
         }
         if let o1 = o1, let o2 = o2, o1.isSameObject(as: o2) {
             return
         }
-        recordFailure(withDescription: "Objects expected to be equal, but weren't. First: \(o1?.description ?? "nil"), "
-            + "second: \(o2?.description ?? "nil")",
-            inFile: String(describing: fileName), atLine: lineNumber, expected: false)
+        XCTFail("Objects expected to be equal, but weren't. First: \(String(describing: o1)), "
+            + "second: \(String(describing: o2))", file: fileName, line: lineNumber)
     }
 
     /// Check whether two collections containing Realm objects are equal.
-    func assertEqualObjectCollections<T: Collection, U: Collection>(_ c1: T, _ c2: U, fileName: StaticString = #file, lineNumber: UInt = #line)
-        where T.Iterator.Element : Object,
-        U.Iterator.Element : Object,
-        T.IndexDistance : Equatable,
-        T.IndexDistance == U.IndexDistance {
+    func assertEqualObjectCollections<C: Collection>(_ c1: C, _ c2: C, fileName: StaticString = #file, lineNumber: UInt = #line)
+        where C.Iterator.Element: Object, C.IndexDistance: Equatable {
             XCTAssertEqual(c1.count, c2.count, "Collection counts were incorrect", file: fileName, line: lineNumber)
             for (o1, o2) in zip(c1, c2) {
                 assertEqual(o1, o2, fileName: fileName, lineNumber: lineNumber)
```

PR #5081, opened 2017-06-28
Add support for user info REST API

### LINKED ISSUE #5060: Expose User Lookup API
### Goal

Our users need a way to look up the random generated user IDs to connect those with their own user models of existing backends and can interact with user-specific Realms. We've an API which covers a minimum useful solution to achieve a user lookup. As this is subject to change, it would be desirable to wrap this API endpoint, so that our users can rely on a stable API and we can change the implementation details.

austinzheng:
Changes:
- Added a `RLMSyncUserInfo` class to encapsulate info retrieved from a server about a user
- Added an API for admin users to retrieve info about a user
- Refactored the internal JSON network client to enable the new endpoint
- Consolidated all the internal JSON model classes in the same set of source files
- Defined a few new internal key strings

Fixes #5060

TODO:
- [x] Resolve the TODO comments
- [x] Write tests (needs a ROS with this feature)
- [x] Changelog entry

This is _not_ a breaking change.


austinzheng:
This is ready for review. I know the test for the new feature passes but I'm still having trouble running the entire object server test suite on my machine, so I'll be closely watching the result of the equivalent CI job.

bdash:
@jpsim, can you please review this when you have a moment?

Realm/ObjectServerTests/RLMObjectServerTests.mm
```diff
@@ -388,7 +388,73 @@ - (void)testOtherUserChangePassword {
...
+                                   [ex2 fulfill];
+                               }];
+    [self waitForExpectationsWithTimeout:10 handler:nil];
+
+    // Get invalid user's info.
+    XCTestExpectation *ex3 = [self expectationWithDescription:@"should fail for non-existent user"];
+    [testUser retrieveUserInfoForProvider:RLMIdentityProviderUsernamePassword
+                         providerIdentity:@"invalid_user@realm.example.org"
+                               completion:^(RLMSyncUserInfo *info, NSError *err) {
+                                   XCTAssertNotNil(err);
```
jpsim: Should we validate that the error matches what we expect here? Code, message, etc?

Realm/ObjectServerTests/RLMObjectServerTests.mm
```diff
@@ -388,7 +388,73 @@ - (void)testOtherUserChangePassword {
...
+                                   [ex3 fulfill];
+                               }];
+    [self waitForExpectationsWithTimeout:10 handler:nil];
+
+    // Get info using user without admin privileges.
+    XCTestExpectation *ex4 = [self expectationWithDescription:@"should fail for user without admin privileges"];
+    [user1 retrieveUserInfoForProvider:RLMIdentityProviderUsernamePassword
+                         providerIdentity:username2
+                               completion:^(RLMSyncUserInfo *info, NSError *err) {
+                                   XCTAssertNotNil(err);
```
jpsim: Should we validate that the error matches what we expect here? Code, message, etc?

Realm/ObjectServerTests/RLMObjectServerTests.mm
```diff
@@ -388,7 +388,73 @@ - (void)testOtherUserChangePassword {
...
+/// A sync admin user should be able to retrieve information about other users.
+- (void)testRetrieveUserInfo {
+    NSString *username1 = @"meela@realm.example.org";
+    NSString *username2 = @"jyaku@realm.example.org";
+    NSString *pw = @"p";
+    NSURL *server = [RLMObjectServerTests authServerURL];
+
+    // Create a non-admin user.
+    RLMSyncCredentials *c1 = [RLMSyncCredentials credentialsWithUsername:username1 password:pw register:YES];
+    RLMSyncUser *user1 = [self logInUserForCredentials:c1 server:server];
```
jpsim: Can we use more descriptive variable names? `user1` -> `nonAdminUser`, `user2` -> `adminUser1`, `testUser` -> `adminUser2`.

Realm/RLMSyncUser.h
```diff
@@ -173,6 +179,22 @@ NS_SWIFT_UNAVAILABLE("Use the full version of this API.");
...
+
+ @param provider    The authentication provider that manages the user whose information is desired.
+ @param providerID  The identity of the user as issued by the authentication provider.
+ @param completion  Completion block invoked when request has completed or failed.
+                    The callback will be invoked on a background queue provided
+                    by `NSURLSession`.
+ */
+- (void)retrieveUserInfoForProvider:(RLMIdentityProvider)provider
+                   providerIdentity:(NSString *)providerID
+                         completion:(RLMRetrieveUserBlock)completion;
```
bdash: I'm not sure I like this naming. `providerIdentity` suggests to me that parameter is the identity of the provider, when in fact it's the provider's concept of the identity of the user. Perhaps `-retrieveInfoForUser:identityProvider:completion:`?

austinzheng: Yeah, that looks a lot more natural.

Realm/RLMSyncUser.h
```diff
@@ -245,6 +267,41 @@ NS_SWIFT_UNAVAILABLE("Use the full version of this API.");
...
+ */
+@property (nonatomic, readonly) RLMIdentityProvider provider;
+
+/**
+ The provider-based username of the user.
+ */
+@property (nonatomic, readonly) NSString *providerID;
+
+/**
+ The internal Realm Object Server identity of the user.
```
bdash: What do we mean by "internal" here? It sorta has the implication of "private", which obviously isn't right.

Realm/RLMSyncUser.h
```diff
@@ -245,6 +267,41 @@ NS_SWIFT_UNAVAILABLE("Use the full version of this API.");
...
+
+/**
+ The authentication provider which manages the user represented by this user info instance.
+ */
+@property (nonatomic, readonly) RLMIdentityProvider provider;
+
+/**
+ The provider-based username of the user.
+ */
+@property (nonatomic, readonly) NSString *providerID;
```
bdash: Is this a username or identity? The doc comment suggests a username, but the property name indicates otherwise. It'd also be good to be consistent about whether we use `identity` or `ID` in these property names. We're using both within this same class.

austinzheng: I'm just going to call it `username`, `identity` and related terms are too confusing.

bdash:
It looks like this needs further revision to conform to the latest API consensus.

austinzheng:
I'll fix it up first thing tomorrow and comment again when it's ready for (re-)review.

austinzheng:
I made the changes, squashed everything, and rebased onto latest master to get rid of GitHub's merge conflict complaints. It should be ready for final review.

bdash:
The API now looks consistent with the consensus we'd arrived at. 👍

@jpsim, can you please do another review pass on this?

austinzheng:
Fixed object server test failure. Rebased onto latest master and manually resolved merge conflicts.

RealmSwift/Sync.swift
```diff
@@ -23,11 +23,19 @@ import Foundation
 /**
  An object representing a Realm Object Server user.
 
- - see: `SyncUser`
+ - see: `RLMSyncUser`
```
jpsim: why this change?

PR #5085, opened 2017-06-29
Change `SyncUser.logIn()` to invoke its callback on the main thread

### LINKED ISSUE #4338: Update SyncUser.logIn to have a callbackQueue optional defaulted parameter

austinzheng:
Also rewrote the documentation to be clearer.

Fixes #4338. Breaking change.

I'm quite against offering the option of a queue to be specified as an argument. This would increase our API surface for negligible gain.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -38,6 +38,9 @@ x.x.x Release notes (yyyy-MM-dd)
   removed.
 * The class methods `Object.className()`, `Object.objectUtilClass()`, and
   the property `Object.isInvalidated` can no longer be overriden.
+* The callback which runs when a sync user login succeeds or fails
+  now runs on the main queue by default, or can be explicitly specified
+  by a new `callbackQueue` parameter on the `{RLM}SyncUser.logIn(...)` API.
 
 ### Enhancements
 
--- a/Realm/RLMSyncUser.h
+++ b/Realm/RLMSyncUser.h
@@ -103,18 +103,31 @@ NS_ASSUME_NONNULL_BEGIN
 #pragma mark - Lifecycle
 
 /**
- Create, log in, and asynchronously return a new user object, specifying a custom timeout for the network request.
- Credentials identifying the user must be passed in. The user becomes available in the completion block, at which point
- it is ready for use.
+ Create, log in, and asynchronously return a new user object, specifying a custom
+ timeout for the network request and a custom queue to run the callback upon.
+ Credentials identifying the user must be passed in. The user becomes available in
+ the completion block, at which point it is ready for use.
  */
 + (void)logInWithCredentials:(RLMSyncCredentials *)credentials
                authServerURL:(NSURL *)authServerURL
                      timeout:(NSTimeInterval)timeout
+               callbackQueue:(dispatch_queue_t)callbackQueue
                 onCompletion:(RLMUserCompletionBlock)completion NS_REFINED_FOR_SWIFT;
 
 /**
- Create, log in, and asynchronously return a new user object. Credentials identifying the user must be passed in. The
- user becomes available in the completion block, at which point it is ready for use.
+ Create, log in, and asynchronously return a new user object.
+
+ If the login completes successfully, the completion block will invoked with
+ a `RLMSyncUser` object representing the logged-in user. This object can be
+ used to open synchronized Realms. If the login fails, the completion block
+ will be invoked with an error.
+
+ The completion block always runs on the main queue.
+
+ @param credentials     A credentials value identifying the user to be logged in.
+ @param authServerURL   The URL of the authentication server (e.g. "http://realm.example.org:9080").
+ @param completion      A callback block that returns a user object or an error,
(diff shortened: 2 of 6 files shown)


austinzheng:
The object server test that failed is one that JP wrote for `asyncOpen`, and has nothing to do with this PR. It's been failing very occasionally; I'm going to investigate when I have time.

Realm/RLMSyncUser.h
```diff
@@ -107,8 +107,19 @@ NS_ASSUME_NONNULL_BEGIN
                 onCompletion:(RLMUserCompletionBlock)completion NS_REFINED_FOR_SWIFT;
 
 /**
- Create, log in, and asynchronously return a new user object. Credentials identifying the user must be passed in. The
- user becomes available in the completion block, at which point it is ready for use.
+ Create, log in, and asynchronously return a new user object.
+
+ If the log in completes successfully, the completion block will be called
```
jpsim: "login"

jpsim: "will be invoked with an `RLMSyncUser` object"

Realm/RLMSyncUser.h
```diff
@@ -107,8 +107,19 @@ NS_ASSUME_NONNULL_BEGIN
...
+ and passed in a `RLMSyncUser` object representing the logged-in user. This
+ object can be used to open synchronized Realms. If the log in fails, the
```
jpsim: "login"

Realm/RLMSyncUser.h
```diff
@@ -107,8 +107,19 @@ NS_ASSUME_NONNULL_BEGIN
...
+ completion block will be called and passed in an error.
```
jpsim: "will be invoked with an error."

Realm/RLMSyncUser.mm
```diff
@@ -413,30 +417,44 @@ + (void)_performLogInForUser:(RLMSyncUser *)user
...
+            });
         }
     };
     [RLMNetworkClient sendRequestToEndpoint:[RLMSyncAuthEndpoint endpoint]
                                      server:authServerURL
                                        JSON:json
                                     timeout:timeout
-                                 completion:handler];
+                                 completion:^(NSError *error, NSDictionary *dictionary) {
+                                     dispatch_async(dispatch_get_main_queue(), ^{
```
bdash: Should this be using `callbackQueue` rather than the main queue here? Should we have tests verifying that each of the possible callbacks are made on the requested queue?

CHANGELOG.md
```diff
@@ -33,6 +33,10 @@ x.x.x Release notes (yyyy-MM-dd)
 * The return types of the `SyncError.clientResetInfo()` and
   `SyncError.deleteRealmUserInfo()` APIs have been changed. They now return
   `RLMSyncErrorActionToken`s or `SyncError.ActionToken`s instead of closures.
+* The callback which runs when a sync user login succeeds or fails
```
jpsim: A single changelog entry should be sufficient.

> The callback which runs when a sync user login succeeds or fails
> now runs on the main queue by default, or can be explicitly specified
> by a new `callbackQueue` parameter on the `{RLM}SyncUser.logIn(...)` API.

RealmSwift/Sync.swift
```diff
@@ -300,16 +300,30 @@ extension RLMSyncCredentials {
...
+     `SyncUser` representing the logged-in user will be passed to it. This user object
+     can be used to open `Realm`s and retrieve `SyncSession`s. Otherwise, the
+     completion block will be called with an error.
+
+     The completion block always runs on the main queue.
+
+     - parameter credentials: A `SyncCredentials` object representing the user to log in.
+     - parameter authServerURL: The URL of the authentication server (e.g. "http://realm.example.org:9080").
+     - parameter timeout: How long the network client should wait, in seconds, before timing out.
+     - parameter queue: The dispatch queue upon which the callback should run. Defaults to the main queue.
```
jpsim: why not keep the name consistent with objc as `callbackQueue`?

RealmSwift/Sync.swift
```diff
@@ -300,16 +300,30 @@ extension RLMSyncCredentials {
...
-     Given credentials and a server URL, log in a user and asynchronously return a `SyncUser`
-     object which can be used to open `Realm`s and retrieve `SyncSession`s.
+     Log in a user and asynchronously retrieve a user object.
+
+     If the log in completes successfully, the completion block will be called, and a
+     `SyncUser` representing the logged-in user will be passed to it. This user object
+     can be used to open `Realm`s and retrieve `SyncSession`s. Otherwise, the
+     completion block will be called with an error.
+
+     The completion block always runs on the main queue.
```
bdash: This comment appears to be inaccurate?

austinzheng: Ah, yes. Thank you for catching this.

PR #5086, opened 2017-06-29
Remove workarounds for incorrect `create(_:, value:, update:)` behavior

### LINKED ISSUE #4926: Realm.create(_:value:update:) doesn't set relationships to nil when passed an Object

austinzheng:
Closes #4926.

This PR is nothing more than deleting some code marked by FIXMEs as to be removed for 3.0, and expanding on the documentation for the affected methods.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,7 +25,11 @@ Synchronized Realms require a server running Realm Object Server v2.0 or higher.
   `SortDescriptor.property`.
   These APIs have been superseded by equivalent APIs that take
   or return key paths instead of property names.
-  
+* The Objective-C and Swift `create(_:, value: update:)` APIs now
+  correctly nil out nullable properties when updating an existing
+  object when the `value` argument specifies nil or `NSNull` for
+  the property value.
+
 ### Enhancements
 
 * Add a new error code to denote 'permission denied' errors when working
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,10 +25,6 @@ Synchronized Realms require a server running Realm Object Server v2.0 or higher.
   `SortDescriptor.property`.
   These APIs have been superseded by equivalent APIs that take
   or return key paths instead of property names.
-* The Objective-C and Swift `create(_:, value: update:)` APIs now
-  correctly nil out nullable properties when updating an existing
-  object when the `value` argument specifies nil or `NSNull` for
-  the property value.
 
 ### Enhancements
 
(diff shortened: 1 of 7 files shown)


CHANGELOG.md
```diff
@@ -47,6 +47,10 @@ Synchronized Realms require a server running Realm Object Server v2.0 or higher.
 * Fix unmanaged object initialization when a nested property type returned
   `false` from `Object.shouldIncludeInDefaultSchema()`.
 * Don't clear RLMArrays on self-assignment.
+* The Objective-C and Swift `create(_:, value: update:)` APIs now
```
jpsim: This is technically a breaking change, so shouldn't it go in that changelog section?

austinzheng: It's a breaking change, but not a breaking API change. If you prefer it be in the first section, I can move it back.

jpsim: I'd say let's just call that section "Breaking Changes"

PR #5091, opened 2017-06-30
Refine `SyncPermissionResults` APIs to match other Realm collections

### LINKED ISSUE #5075: SyncPermissionResults's method names aren't consistent with RealmCollection
## Goals

The query and sorting capability on `SyncPermissionResults` is exposed using the Objective-C API's naming conventions rather than those used in the Swift API.

## Expected Results
```swift
// Not available!
permissions.filter(predicate)
permissions.sorted(byKeyPath: keyPath, ascending: true)

// Instead:
permissions.objects(with: predicate)
permissions.sortedResults(using: keyPath, ascending: true)
```

austinzheng:
Fixes #5075 
Breaking change

TODO:
- [ ] Any affordances needed for migration? This feature has only been around for a short while, and it's not clear whether there exists a way to provide a fixit for a non-refined API that becomes refined for Swift.

### DIFF
--- a/Realm/ObjectServerTests/SwiftPermissionsAPITests.swift
+++ b/Realm/ObjectServerTests/SwiftPermissionsAPITests.swift
@@ -151,9 +151,9 @@ class SwiftPermissionsAPITests: SwiftSyncTestCase {
         XCTAssertNotNil(finalValue, "Did not find the permission \(expectedPermission)")
 
         // Check getting permission by its index.
-        let index = results.index(ofObject: expectedPermission)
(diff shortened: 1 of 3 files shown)


tgoyne:
I think you could add definitions with the old unrefined names in the extension that are marked as `@available(renamed:)` to get fix-its to the new names. If that doesn't work for whatever reason it's not worth worrying about too much.

austinzheng:
Cool! I'll try that out.

austinzheng:
Superseded by #5168

PR #5092, opened 2017-06-30
Rework the 'interactive errors' API

austinzheng:
Changes:
- Client reset and access denied errors no longer pass back a block in the `userInfo` dictionary, but rather a token which can be passed into a separate API

TODO:
- [x] Discuss with team whether this is a good approach
- [x] Think about affordances for migration


austinzheng:
I've decided not to add migration aliases (e.g. the old methods signatures + `@available(*, unavailable, ...)` because the API changes only differ in terms of return value. The errors users get should be pretty straightforward, there is no trivial fixit migration path, and adding in aliases would break type inference for these methods.

austinzheng:
I updated and rebased this PR.

I'm wondering if it might be possible to stick `RLMSyncErrorActionToken` in an existing file to keep our file count down, especially since it has no publicly accessible API.

RealmSwift/Sync.swift
```diff
@@ -82,6 +82,41 @@ public typealias UserCompletionBlock = RLMUserCompletionBlock
...
+        if code == SyncError.clientResetError,
+            let recoveryPath = userInfo[kRLMSyncPathOfRealmBackupCopyKey] as? String,
+            let token = _nsError.__rlmSync_errorActionToken() {
+            return (recoveryPath, token)
+        }
+        return nil
+    }
+
+    /**
+     Given a permission denied error, extract and return the recovery file path
```
jpsim: no file path is returned here

austinzheng:
This is ready for a second pass.

PR #5114, opened 2017-07-11
don't test "compact on launch" if the Realm is open on another thread

### LINKED ISSUE #5010: Clarification shouldCompactOnLaunch documentation
The example in the documentation (https://realm.io/docs/swift/latest/#compacting-realms) is not very clear to me, as I don't know if the compaction could be called all the time during app use or only once at startup. Is the implementation below correct or would it be better to make a separate config including shouldCompactOnLaunch to call once on app launch. 

If I add shouldCompactOnLaunch to the default configuration I see the block being called every time I create a realm instance.

            Realm.Configuration.defaultConfiguration = Realm.Configuration(schemaVersion: schemaVersion, migrationBlock: migrationBlock,shouldCompactOnLaunch: { totalBytes, usedBytes in
            // totalBytes refers to the size of the file on disk in bytes (data + free space)
            // usedBytes refers to the number of bytes used by data in the file
            
            // Compact if the file is over 100MB in size and less than 50% 'used'
            let oneHundredMB = 100 * 1024 * 1024
            print ("totalbytes \(totalBytes)")
            print ("usedbytes \(usedBytes)")
            if (totalBytes > oneHundredMB) && (Double(usedBytes) / Double(totalBytes)) < 0.7{
                print("will compact realm")
            }
            return (totalBytes > oneHundredMB) && (Double(usedBytes) / Double(totalBytes)) < 0.7
        })
        do {
            // Realm is compacted on the first open if the configuration block conditions were met.
            _ = try Realm(configuration: config)
        } catch {
            // handle error compacting or opening Realm
        }

And one more thing would be interesting to me: What happens if the compaction fails? Too little storage would be a reason. Will I still be able to access the data and the compaction will just be skipped?

jpsim:
Fixes #5010. Depends on realm/realm-object-store#496.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,8 @@ x.x.x Release notes (yyyy-MM-dd)
 
 ### Bugfixes
 
-* None.
+* The `shouldCompactOnLaunch` block is no longer invoked if the Realm at that
+  path is already open on other threads.
 
 2.9.0 Release notes (2017-07-26)
 =============================================================
--- a/Realm/Tests/CompactionTests.m
+++ b/Realm/Tests/CompactionTests.m
@@ -117,7 +117,7 @@ NSUInteger count = 1000;
 }
 
 - (void)testCachedRealmCompactOnLaunch {
-    // Test that compact never gets called if there are cached Realms
+    // Test that the compaction block never gets called if there are cached Realms
     // Access Realm before opening it with a compaction block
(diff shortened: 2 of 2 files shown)


jpsim:
This is ready for review again, now that https://github.com/realm/realm-object-store/pull/496 has been merged.

jpsim:
Rebased to move the changelog entry to the appropriate section. This should be easy to review, it's just tests /cc @tgoyne

PR #5126, opened 2017-07-13
Pull in object store changes to sync metadata subsystem

austinzheng:
Requires https://github.com/realm/realm-object-store/pull/497. No user facing changes worth discussing.

### DIFF
--- a/Realm/RLMSyncUser.mm
+++ b/Realm/RLMSyncUser.mm
@@ -203,7 +203,7 @@ PermissionChangeCallback RLMWrapPermissionStatusCallback(RLMPermissionStatusBloc
     if (!_user) {
         return nil;
     }
-    auto path = SyncManager::shared().path_for_realm(_user->identity(), [url.absoluteString UTF8String]);
+    auto path = SyncManager::shared().path_for_realm(*_user, [url.absoluteString UTF8String]);
     if (auto session = _user->session_for_on_disk_path(path)) {
         return [[RLMSyncSession alloc] initWithSyncSession:session];
     }
@@ -252,7 +252,7 @@ PermissionChangeCallback RLMWrapPermissionStatusCallback(RLMPermissionStatusBloc
 }
 
 - (NSURL *)authenticationServer {
-    if (!_user) {
+    if (!_user || _user->token_type() == SyncUser::TokenType::Admin) {
         return nil;
     }
     return [NSURL URLWithString:@(_user->server_url().c_str())];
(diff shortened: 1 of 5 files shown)


Realm/RLMSyncUser.mm
```diff
@@ -252,7 +252,7 @@ - (RLMRealm *)permissionRealmWithError:(NSError **)error {
 }
 
 - (NSURL *)authenticationServer {
-    if (!_user) {
+    if (!_user || _user->token_type() == SyncUser::TokenType::Admin) {
```
austinzheng: This change isn't strictly necessary, but maybe it's good for documentation purposes?

bdash: Why is it unnecessary?

austinzheng: No, you're right. I was too hasty.

bdash: I wasn't saying you were wrong, I was just interested in the reasoning!

austinzheng: My original reasoning was that the code that method called into took care of returning null or nil or whatever for admin token users, but that doesn't appear to be true. (It might have been true before the object store changes this is trying to incorporate.)

Realm/RLMSyncUser.mm
```diff
@@ -401,13 +406,28 @@ + (void)_performLogInForUser:(RLMSyncUser *)user
...
+        if (identity) {
+            identity_string = util::make_optional<std::string>(identity.UTF8String);
+        }
+        sync_user = SyncManager::shared().get_admin_token_user([serverURL absoluteString].UTF8String,
+                                                               credentials.token.UTF8String,
+                                                               std::move(identity_string));
+    } else {
+        // Retrieve the user based on the identity.
+        if (!identity) {
+            @throw RLMException(@"A direct access credential must specify either/both an identity or a server URL.");
```
bdash: I think this message would be clearer if it ended "must specify either an identity, server URL, or both".

Realm/RLMSyncUser.mm
```diff
@@ -401,13 +406,28 @@ + (void)_performLogInForUser:(RLMSyncUser *)user
...
-    auto sync_user = SyncManager::shared().get_user([identity UTF8String],
-                                                    [credentials.token UTF8String],
-                                                    none,
-                                                    SyncUser::TokenType::Admin);
+    std::shared_ptr<SyncUser> sync_user;
+    if (serverURL) {
+        // Retrieve the user based on the auth server URL.
+        util::Optional<std::string> identity_string;
+        if (identity) {
+            identity_string = util::make_optional<std::string>(identity.UTF8String);
```
bdash: This can just use `std::string(identity.UTF8String)`.

PR #5136, opened 2017-07-17
disable update checker on master-3.0 branch

jpsim:
since we won't be updating the contents of `static.realm.io/update/cocoa` with 3.0 pre-releases, and we don't want to constantly prompt users of those pre-releases for updates.


jpsim:
Alright, I added the same check as #5137.

PR #5137, opened 2017-07-17
point podspec documentation URL to '/latest'

jpsim:
since we're not publishing dedicated docs for 3.0 pre-releases. I did this manually for the release of `3.0.0-beta` a few minutes ago (https://github.com/CocoaPods/Specs/commit/350f917949eb80336f100bf37634fec6f159fab5 & https://github.com/CocoaPods/Specs/commit/a8526da512820edb459e2590b5d5b6039a30f381)

### DIFF
--- a/Realm.podspec
+++ b/Realm.podspec
@@ -13,7 +13,8 @@ Pod::Spec.new do |s|
   s.library                 = 'c++', 'z'
   s.requires_arc            = true
   s.social_media_url        = 'https://twitter.com/realm'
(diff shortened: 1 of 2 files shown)


bdash:
Much like with #5136, I'd prefer to see this done in a way that doesn't require that we remember to revert it in the future.

jpsim:
Alright, done as a check for 'alpha', 'beta' or 'rc' in the version string.

PR #5096, opened 2017-07-05
Remove non-sync APIs marked `deprecated` and `unavailable`

austinzheng:
TODO:
- [x] Some of the deprecated Swift APIs removed in this commit should be brought back but marked `unavailable`

PR #5141, opened 2017-07-18
Add changelog entry for removed non-sync APIs

austinzheng:
Adds a changelog for #5096.


CHANGELOG.md
```diff
@@ -3,6 +3,23 @@
...
+  `+[RLMSortDescriptor sortDescriptorWithProperty:ascending:]`,
+  `RLMSortDescriptor.property`.
+* Remove the following deprecated Swift APIs:
+  `AnyRealmCollection.sorted(byProperty:, ascending:)`,
+  `LinkingObjects.sorted(byProperty:, ascending:)`,
+  `List.sorted(byProperty:, ascending:)`,
+  `Results.sorted(byProperty:, ascending:)`,
+  `SortDescriptor.init(property:, ascending:)`,
+  `SortDescriptor.property`.
+
```
bdash: It might be worth adding a note that users should switch to the equivalent APIs that work with key paths.

PR #5142, opened 2017-07-18
Rename 'API Breaking Changes' to 'Breaking Changes'

austinzheng:
Per @jpsim's suggestion.

Thoughts or objections?


austinzheng:
Do we want to change the name forever, or just for this major version increment? If the former, I'll update the script.

jpsim:
I don't see why we wouldn't want to call out breaking changes that didn't affect the API as breaking changes in the changelog.

Issue #5104, opened 2017-07-10
Should constantly try to reconnect regardless of ReachabilityStatus


kishikawakatsumi:
Because Reachability doesn't guarantee that a packet can actually reach the target host. It is about the local network configuration (interface up / down) and if it can send a packet. 

> A remote host is considered reachable when a data packet, sent by an application into the network stack, can leave the local device. Reachability does not guarantee that the data packet will actually be received by the host.

[SCNetworkReachability | Apple Developer Documentation](https://developer.apple.com/documentation/systemconfiguration/scnetworkreachability-g7d#//apple_ref/doc/uid/TP40007260)

> Important: The SCNetworkReachability API is not intended for use as a preflight mechanism for determining network connectivity. You determine network connectivity by attempting to connect. If the connection fails, consult the SCNetworkReachability API to help diagnose the cause of the failure.

[Design for Variable Network Interface Availability](https://developer.apple.com/library/content/documentation/NetworkingInternetWeb/Conceptual/NetworkingOverview/WhyNetworkingIsHard/WhyNetworkingIsHard.html#//apple_ref/doc/uid/TP40010220-CH13-SW3)

There are situations where we can not connect to ROS even if the device's network connection is fine. (e.g. ROS is down, 100% packet loss by Network Link Conditioner) In those cases, it will not reconnect forever since `ReachabilityStatus` does not change after failing connection once.

## Goals
Sync will resume when the server recovers even if the app launched while the server is down.

## Expected Results
An app tries to reconnect to ROS constantly.

## Actual Results
If an app fails first (ROS is down or 100% packet loss), never reconnect again.

## Steps to Reproduce
1. Launch sync example apps without ROS
2. Launch ROS after the app launched
3. Sync never happen

## Version of Realm and Tooling
Realm framework version: 2.8.3

Realm Object Server version: 1.8

Xcode version: 8.3

iOS/OSX version: 1.3

CC @austinzheng @jpsim @bdash

bdash:
As far as I'm aware, we'll always try to reconnect so long as we have a sync session active. It's just that we'll increase the interval between reconnection attempts when they fail. We use reachability changes as a trigger to reset the interval between reconnection attempts. When we fail to reconnect in this case, is it simply because the reconnect interval has grown very large? Or is something else at play?

austinzheng:
I agree with Mark. I'll have to check again, but as far as I remember reachability is only used to speed up reconnection, never to delay or stop it.

[...]

PR #5143, opened 2017-07-19
Extend the retry time to longer than `refreshBuffer`

kishikawakatsumi:
Fixes https://github.com/realm/realm-cocoa/issues/5104

Fix an issue that reconnection does not happen if initial session binding fails.
If initial session binding fails, retry time will be offset by `refreshBuffer`. Retry will not be registered in the timer.  In that case, there is no session, no error handler will be called anymore, so reconnecting will not happen again.

Extend the retry time to longer than `refreshBuffer`.

CC: @austinzheng @bdash

### DIFF
--- a/Realm/RLMSyncSessionRefreshHandle.mm
+++ b/Realm/RLMSyncSessionRefreshHandle.mm
@@ -40,6 +40,8 @@ void unregisterRefreshHandle(const std::weak_ptr<SyncUser>& user, const std::str
 
 }
 
+static const NSTimeInterval RefreshBuffer = 10;
+
 @interface RLMSyncSessionRefreshHandle () {
     std::weak_ptr<SyncUser> _user;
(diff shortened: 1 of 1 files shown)


Realm/RLMSyncSessionRefreshHandle.mm
```diff
@@ -40,6 +40,8 @@ void unregisterRefreshHandle(const std::weak_ptr<SyncUser>& user, const std::str
 
 }
 
+static const NSTimeInterval RefreshBuffer = 10;
```
austinzheng: Based on `RLMCollection.mm`, can you rename this `RLMRefreshBuffer` just so it's clear that it's a static constant?

kishikawakatsumi:
This PR changes only object server code. CI failures seems unrelated.

PR #5148, opened 2017-07-20
Make non-breaking changes to support arrays of primitives

tgoyne:
A bunch of individually not very interesting changes made to support arrays of primitives. Should have no functional changes other than some error messages being a bit different.

Pulls in https://github.com/realm/realm-object-store/pull/464 (and so depends on that getting merged).

### DIFF
--- a/Realm/RLMObjectSchema.mm
+++ b/Realm/RLMObjectSchema.mm
@@ -114,7 +114,8 @@ using namespace realm;
     Class superClass = class_getSuperclass(cls);
     NSArray *allProperties = @[];
     while (superClass && superClass != RLMObjectBase.class) {
-        allProperties = [[RLMObjectSchema propertiesForClass:cls isSwift:isSwift] arrayByAddingObjectsFromArray:allProperties];
+        allProperties = [[RLMObjectSchema propertiesForClass:cls isSwift:isSwift]
+                         arrayByAddingObjectsFromArray:allProperties];
         cls = superClass;
         superClass = class_getSuperclass(superClass);
     }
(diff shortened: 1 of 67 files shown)


Realm/RLMProperty.mm
```diff
@@ -354,42 +373,40 @@ - (instancetype)initSwiftPropertyWithName:(NSString *)name
 
     // convert array types to objc variant
     if ([rawType isEqualToString:@"@\"RLMArray\""]) {
-        rawType = [NSString stringWithFormat:@"@\"RLMArray<%@>\"", [propertyValue objectClassName]];
+        RLMArray *value = propertyValue;
+        _type = RLMPropertyTypeArray;
+        _objectClassName = value.objectClassName;
+        if (_objectClassName) {
+            // FIXME: validate
```
bdash: Is this an existing FIXME that was moved around, or something new that should be addressed?

Realm/Tests/DynamicTests.m
```diff
@@ -161,7 +161,7 @@ - (void)testDynamicTypes {
 
     // verify properties
     RLMRealm *dyrealm = [self realmWithTestPathAndSchema:nil];
-    RLMResults *results = [dyrealm allObjects:AllTypesObject.className];
+    RLMResults<RLMObject *> *results = [dyrealm allObjects:AllTypesObject.className];
```
bdash: Are these changes required by this PR, or are they being made preemptively in advance of #5149?

Realm/Tests/Swift/SwiftSchemaTests.swift
```diff
@@ -82,13 +82,13 @@ class SwiftRecursingSchemaTestObject : RLMObject {
 }
 
 class InitAppendsToArrayProperty : RLMObject {
-    @objc dynamic var propertyWithIllegalDefaultValue: RLMArray = {
+    @objc dynamic var propertyWithIllegalDefaultValue: RLMArray<SwiftIntObject> = {
```
bdash: Are these changes required by this PR, or are they being made preemptively in advance of #5149?

RealmSwift/Migration.swift
```diff
@@ -80,7 +80,7 @@ extension Realm {
  instance provides access to the old and new database schemas, the objects in the Realm, and provides functionality for
  modifying the Realm during the migration.
  */
-public final class Migration {
+public struct Migration {
```
bdash: I assume that changing from class to struct doesn't in any way change how users interact with these types (i.e., it's not a breaking change)?

Realm/Tests/DynamicTests.m
tgoyne: All of the tests build and pass with no changes to the tests other than updating what error message a few of the tests are looking for.

RealmSwift/Migration.swift
tgoyne: I don't see any way for it to break anything, since they're immutable types which the user can't construct directly. `Realm` should maybe also be a struct but that would actually be a breaking change.

austinzheng: Users can theoretically do stupid things like retroactively conforming `Migration` to a class-only protocol, but I don't think that's something we should worry about.

Realm/RLMObjectSchema.mm
```diff
@@ -114,7 +114,8 @@ + (instancetype)schemaForObjectClass:(Class)objectClass {
     Class superClass = class_getSuperclass(cls);
     NSArray *allProperties = @[];
     while (superClass && superClass != RLMObjectBase.class) {
-        allProperties = [[RLMObjectSchema propertiesForClass:cls isSwift:isSwift] arrayByAddingObjectsFromArray:allProperties];
+        allProperties = [[RLMObjectSchema propertiesForClass:cls isSwift:isSwift]
+                         arrayByAddingObjectsFromArray:allProperties];
```
jpsim: indentation

Realm/RLMObjectSchema.mm
```diff
@@ -217,87 +218,81 @@ + (NSArray *)propertiesForClass:(Class)objectClass isSwift:(bool)isSwiftClass {
...
+            if (!type) {
+                return;
+            }
+            if (property) {
+                property.type = RLMPropertyType(type.intValue);
+                return;
+            }
+
+            // Check to see if this optional property is an underlying storage property for a Swift lazy var.
+            // Managed lazy vars are't allowed.
```
jpsim: "aren't"

PR #5149, opened 2017-07-20
Make breaking changes needed to support arrays of primitives

tgoyne:
Everything but the last commit is just pulling in latest master and https://github.com/realm/realm-cocoa/pull/5148 on top of master-3.0.

This makes the two API changes needed for arrays of primitives: loosening the generic constraint to allow non-RLMObjects in RLMArray/RLMResults (but not yet actually supporting anything else), and changing RLMProperty/RLMPropertyType so that arrays of things other than objects can be expressed.


bdash:
Shouldn't we pull in the changes from master (and #5148, once it's merged to master) via a merge commit, rather than rebasing as you appear to have done here?

tgoyne:
Yes, but that merge should be from `master` so it needs #5148 to be merged first.

Realm/RLMUtil.mm
```diff
@@ -101,77 +97,82 @@ static inline bool nsnumber_is_like_double(__unsafe_unretained NSNumber *const o
...
-                                 __unsafe_unretained RLMProperty *const property) {
-    if (property.optional && !RLMCoerceToNil(obj)) {
+BOOL RLMValidateValue(id value, RLMPropertyType type, bool optional, bool array,
+                      NSString *objectClassName) {
+    if (optional && !RLMCoerceToNil(value)) {
         return YES;
     }
+    if (array) {
+        if (RLMArray *array = RLMDynamicCast<RLMArray>(value)) {
+            return [array.objectClassName isEqualToString:objectClassName];
```
jpsim: Doesn't this need to consider primitive types?

Realm/RLMUtil.mm
```diff
@@ -101,77 +97,82 @@ static inline bool nsnumber_is_like_double(__unsafe_unretained NSNumber *const o
...
+        }
+        if (RLMListBase *list = RLMDynamicCast<RLMListBase>(value)) {
+            return [list._rlmArray.objectClassName isEqualToString:objectClassName];
```
jpsim: Doesn't this need to consider primitive types?

Realm/RLMArray.mm
```diff
@@ -387,7 +435,7 @@ - (id)maxOfProperty:(NSString *)property {
 
 - (id)sumOfProperty:(NSString *)property {
     [self validateAggregateProperty:property method:_cmd allowDate:false];
-    return [_backingArray valueForKeyPath:[@"@sum." stringByAppendingString:property]];
+    return [_backingArray valueForKeyPath:[@"@sum." stringByAppendingString:property]] ?: @0;
```
tgoyne: We decided not to match Foundation's behavior because Foundation is wrong; the sum of an empty set isn't undefined.

Realm/RLMUtil.mm
tgoyne: No, this PR does not add support for arrays of primitives.

Realm/RLMArray.mm
jpsim: I vaguely remember this discussion, but not the outcome. thanks for clarifying.

RealmSwift/List.swift
```diff
@@ -118,28 +118,28 @@ public final class List<T: Object>: ListBase {
...
     // MARK: KVC
 
     /**
      Returns an `Array` containing the results of invoking `valueForKey(_:)` using `key` on each of the collection's
      objects.
-    */
-    public override func value(forKey key: String) -> Any? {
-        return value(forKeyPath: key)
+     */
+    @nonobjc public func value(forKey key: String) -> [AnyObject] {
```
bdash: Does `@nonobjc` mean that this cannot be invoked from Obj-C? I'd think that would give unexpected behavior from calls to `-valueForKeyPath:` with a key path that traverses into a `List` but where the `List` isn't the root object.

RealmSwift/RealmCollection.swift
```diff
@@ -123,6 +122,11 @@ private func forceCast<A, U>(_ from: A, to type: U.Type) -> U {
     return from as! U
 }
 
+/// A type which can be stored in a Realm List or Results
+public protocol RealmCollectionValue {
+    static func className() -> String
```
bdash: What will this return for primitive types?

RealmSwift/Tests/ListTests.swift
```diff
@@ -299,20 +299,20 @@ class ListTests: TestCase {
...
+        array.swap(1, 1)
         XCTAssertEqual(Int(2), array.count)
         assertEqual(str2, array[0])
         assertEqual(str1, array[1])
 
-        assertThrows(array.swap(index1: -1, 0))
-        assertThrows(array.swap(index1: 0, -1))
-        assertThrows(array.swap(index1: 1000, 0))
-        assertThrows(array.swap(index1: 0, 1000))
+        assertThrows(array.swap(-1, 0))
```
bdash: In Swift 4, the equivalent method on `Array` appears to be called [`swapAt`](https://developer.apple.com/documentation/swift/array/2893281-swapat).

RealmSwift/Tests/ObjectSchemaTests.swift
```diff
@@ -47,9 +47,9 @@ class ObjectSchemaTests: TestCase {
...
...\t}\n\tfloatCol {\n\t\ttype = float;\n\t\tobjectClassName = (null);\n\t\tlinkOriginPropertyName = (null);\n\t\tindexed = NO;\n\t\tisPrimary = NO;\n\t\toptional = NO;\n\t}\n\tdoubleCol {\n\t\ttype = double;\n\t\tobjectClassName = (null);\n\t\tlinkOriginPropertyName = (null);\n\t\tindexed = NO;\n\t\tisPrimary = NO;\n\t\toptional = NO;\n\t}\n\tstringCol {\n\t\ttype = string;\n\t\tobjectClassName = (null);\n\t\tlinkOriginPropertyName = (null);\n\t\tindexed = NO;\n\t\tisPrimary = NO;\n\t\toptional = NO;\n\t}\n\tbinaryCol {\n\t\ttype = data;\n\t\tobjectClassName = (null);\n\t\tlinkOriginPropertyName = (null);\n\t\tindexed = NO;\n\t\tisPrimary = NO;\n\t\toptional = NO;\n\t}\n\tdateCol {\n\t\ttype = date;\n\t\tobjectClassName = (null);\n\t\tlinkOriginPropertyName = (null);\n\t\tindexed = NO;\n\t\tisPrimary = NO;\n\t\toptional = NO;\n\t}\n\tobjectCol {\n\t\ttype = object;\n\t\tobjectClassName = SwiftBoolObject;\n\t\tlinkOriginPropertyName = (null);\n\t\tindexed = NO;\n\t\tisPrimary = NO;\n\t\toptional = YES;\n\t}\n\tarrayCol {\n\t\ttype = array;\n\t\tobjectClassName = SwiftBoolObject;\n\t\tlinkOriginPropertyName = (null);\n\t\tindexed = NO;\n\t\tisPrimary = NO;\n\t\toptional = NO;\n\t}\n}")
```
bdash: I assume this needs to be updated?

RealmSwift/List.swift
tgoyne: Invoking it from obj-c calls the one defined on the parent class which does the exact same thing. This function just refines the return type when calling it from from Swift to make the code using it a little less awkward.

RealmSwift/RealmCollection.swift
tgoyne: This function has actually gone away on the primitive arrays branch because different functions were needed.

CHANGELOG.md
```diff
@@ -99,6 +99,13 @@ x.x.x Release notes (yyyy-MM-dd)
   correctly nil out nullable properties when updating an existing
   object when the `value` argument specifies nil or `NSNull` for
   the property value.
+* Rename `List.remove(objectAtIndex:)` to `List.remove(at:)` to match the named
```
austinzheng: "name"

Realm/RLMObjectBase.mm
```diff
@@ -78,34 +78,28 @@ - (void)dealloc {
     _observationInfo = nullptr;
 }
 
+static id cooerceToObjectType(id obj, Class cls, RLMSchema *schema) {
```
austinzheng: "coerce"

RealmSwift/List.swift
```diff
@@ -314,14 +314,16 @@ public final class List<T: Object>: ListBase {
 
      - parameter index: The index at which to remove the object.
      */
-    public func remove(objectAtIndex index: Int) {
+    public func remove(at index: Int) {
```
austinzheng: The Swift standard library version of this method returns the removed element. I don't know if this is something we should replicate.

tgoyne: It's sort of gratuitous overhead for something that's at best vaguely convenient, but matching the standard library is probably the way to go.

tgoyne: It looks like `Array` returns the value that was there, but `Collection`/`RangeReplaceableCollection` don't and there's alternative functions on `Collection` for remove and return old (`popLast()` etc.), so I'm inclined to leave it as a void return.

RealmSwift/RealmCollection.swift
bdash: It looks like CI is failing due to this function being undocumented:
```
Undocumented Realm swift declarations:
{
  "warnings": [
    {
      "file": "/Users/realm/workspace/objc_pr/configuration/Release/target/docs/xcode_version/8.3.3/RealmSwift/RealmCollection.swift",
      "line": 127,
      "symbol": "RealmCollectionValue.className()",
      "symbol_kind": "source.lang.swift.decl.function.method.static",
      "warning": "undocumented"
    }
  ],
```

bdash:
The changelog entries will need to be moved into the section for the upcoming section now that v3.0.0-beta.3 has been released.

tgoyne:
Failures are both CI things that passed before updating the changelog.

PR #5155, opened 2017-07-21
Create SUPPORT.md

austinzheng:
Thoughts or comments appreciated.

### DIFF
--- a/SUPPORT.md
+++ b/SUPPORT.md
@@ -1,3 +1,5 @@
+# Support
+
 The Realm team is here to help you with your Realm-related issues!
 
 ## Documentation
--- a/SUPPORT.md
+++ b/SUPPORT.md
@@ -14,6 +14,6 @@ When asking questions on Stack Overflow, please keep in mind Stack Overflow's [q
 
 ## GitHub Issues
 
-If you are running into issues with Realm, including questions which are very specific to your particular application or potential bugs, we encourage you to file an issue on our [GitHub issue tracker](https://github.com/realm/realm-cocoa/issues). Please fill out the issue template in its entirety.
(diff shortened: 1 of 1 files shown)


jpsim:
From what I can tell from the [blog post](https://github.com/blog/2400-support-file-support) announcing this, the contents should be more concise than the multi-paragraph approach you've taken here. Care to trim that down to a 1-liner? /cc @aviwarner

bdash:
Why do you think that, @jpsim? The document in question is simply linked to at the top of the [New Issue page], much like our existing guidelines for contributing.

aviwarner:
This is great @austinzheng. I'll go ahead and replicate on our other public repos. I think we're OK with the multi-paragraph format due to how the support guidelines are presented (just clarified this with @jpsim).

austinzheng:
Sounds good. I'm wondering if I've missed anything, or if there are other resources we do (or specifically do not) want to mention.

SUPPORT.md
```diff
@@ -0,0 +1,19 @@
...
+
+## Stack Overflow
+
+If you have questions about Realm you can ask them on Stack Overflow. We continually monitor the [`realm` tag](https://stackoverflow.com/tags/realm). Please also tag your question with `ios`, `swift`, or other tags as appropriate.
+
+When asking questions on Stack Overflow, please keep in mind Stack Overflow's [question guidelines](https://stackoverflow.com/help/how-to-ask), and please use their search functionality to see if your question has been asked before.
+
+## GitHub Issues
+
+If you are running into issues with Realm, including questions which are very specific to your particular application or potential bugs, we encourage you to file an issue on our [GitHub issue tracker](https://github.com/realm/realm-cocoa/issues). Please fill out the issue template in its entirety.
```
bdash: I think It'd be worth pointing users to CONTRIBUTING.md in this section.

aviwarner:
@austinzheng there's some fine tuning to be done here, but I'm ok with this going up in the short term as a v1. I'll work on a rewrite just to help clarify what we do/don't support and where certain channels are more appropriate.

PR #5098, opened 2017-07-05
Remove deprecated sync-related APIs

austinzheng:
Changes:
- Removed deprecated APIs and definitions
- Removed the `RLMSyncPermission` and `RLMSyncPermissionChange` classes
- Removed now-irrelevant tests
- Moved sync offer and response tests into `RLMPermissionsAPITests.m`

PR #5160, opened 2017-07-24
Detect and automatically ignore computed properties

### LINKED ISSUE #2840: Detect computed properties by checking for the existence of backing ivars

austinzheng:
Replaces #3870. Is that PR, but manually refactored to work with our latest code and to accommodate lazy Swift properties.

I've left that PR alone so that the changes in this PR can be compared more easily against the original changes.

Fixes #2840.

### DIFF
--- a/Realm/RLMObjectSchema.mm
+++ b/Realm/RLMObjectSchema.mm
@@ -216,7 +216,7 @@ using namespace realm;
 
         if (prop) {
             [propArray addObject:prop];
-         }
+        }
     }
 
     auto existingPropertyIndex = [=](NSString *name) -> NSUInteger {
--- a/Realm/RLMProperty.mm
+++ b/Realm/RLMProperty.mm
@@ -74,7 +74,7 @@ void RLMValidateSwiftPropertyName(NSString *name) {
     }
 }
 
-static bool rawTypeIsComputedProperty(NSString *rawType) {
+static bool rawTypeShouldBeTreatedAsComputedProperty(NSString *rawType) {
     return [rawType isEqualToString:@"@\"RLMLinkingObjects\""] || [rawType hasPrefix:@"@\"RLMLinkingObjects<"];
 }
 
@@ -296,10 +296,14 @@ static realm::util::Optional<RLMPropertyType> typeFromProtocolString(const char 
     return YES;
 }
 
-- (void)parseObjcProperty:(objc_property_t)property readOnly:(bool *)readOnly rawType:(NSString **)rawType {
+- (void)parseObjcProperty:(objc_property_t)property
+                 readOnly:(bool *)readOnly
+                 computed:(bool *)computed
+                  rawType:(NSString **)rawType {
     unsigned int count;
     objc_property_attribute_t *attrs = property_copyAttributeList(property, &count);
 
+    *computed = true;
     for (size_t i = 0; i < count; ++i) {
         switch (*attrs[i].name) {
             case 'T':
@@ -319,6 +323,9 @@ static realm::util::Optional<RLMPropertyType> typeFromProtocolString(const char 
                 break;
             case 'S':
                 _setterName = @(attrs[i].value);
+                break;
+            case 'V': // backing ivar name
+                *computed = false;
                 break;
             default:
                 break;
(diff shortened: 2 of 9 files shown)


Realm/RLMObjectSchema.mm
```diff
@@ -128,20 +128,6 @@ + (instancetype)schemaForObjectClass:(Class)objectClass {
     }]];
     schema.computedProperties = computedProperties;
 
-    // verify that we didn't add any properties twice due to inheritance
```
tgoyne: Why is this (and the associated tests) being removed?

austinzheng: I believe the original rationale was that the new ivar-based property detection made this check redundant. That being said, it doesn't make sense to remove the tests and I'm not convinced the check is actually redundant. I'll restore both.

RealmSwift/Tests/SwiftTestObjects.swift
```diff
@@ -442,6 +442,18 @@ class SwiftObjectiveCTypesObject: Object {
     @objc dynamic var numCol: NSNumber? = 0
 }
 
+class SwiftComputedPropertyNotExplicitlyIgnoredObject: Object {
```
jpsim: Shouldn't there be a test that validates the schema for this model?

RealmSwift/Tests/SwiftTestObjects.swift
```diff
@@ -442,6 +442,18 @@ class SwiftObjectiveCTypesObject: Object {
...
+    dynamic var _urlBacking = ""
+    var url: URL? {
```
jpsim: Shouldn't this be marked as `dynamic`, as all other Realm properties that would otherwise have been considered "managed" be configured? Or perhaps it'd be wise to have both defined here to confirm that we do the right thing in all cases.

austinzheng: Sorry, do you mean `url` should be marked as `dynamic`, or that we should have two properties, one marked `dynamic` and one not?

austinzheng:
I rebased this PR, but I can't test it because every time I run the tests I run into the history upgrade core crash.

bdash:
Which history upgrade core crash is that?

austinzheng:
I appear to be seeing https://github.com/realm/realm-core/pull/2726 whenever I run the tests, although I want to verify that I didn't just screw up my local setup. JP earlier worked around this crash in some of our other tests, as shown in the comment at https://github.com/realm/realm-cocoa/pull/5098#pullrequestreview-51298908.

bdash:
That bug manifests as a deadlock. I introduced the workaround earlier.

austinzheng:
I'll continue investigating tomorrow. I assume that if I were correct, I would have seen it even before rebasing.

austinzheng:
I've addressed feedback and this is ready for another round of review.

austinzheng:
The object server test failure is unrelated to this change, and I think I know why it happens. Will open another PR to try to address it.

Realm/RLMProperty.mm
```diff
@@ -349,8 +356,17 @@ - (instancetype)initSwiftPropertyWithName:(NSString *)name
...
     NSString *rawType;
     bool readOnly = false;
-    [self parseObjcProperty:property readOnly:&readOnly rawType:&rawType];
-    if (readOnly) {
+    bool isComputed = false;
+    [self parseObjcProperty:property readOnly:&readOnly computed:&isComputed rawType:&rawType];
+    if (!readOnly && isComputed) {
+        // Check for lazy property.
+        NSString *backingPropertyName = [NSString stringWithFormat:@"%@.storage", name];
+        if (class_getInstanceVariable([obj class], backingPropertyName.UTF8String)) {
```
markst: Don't quite follow why checking for lazy property occurs if the property is computed?
or do lazy properties resolve as being computed when parsing?
also seems strange since lazy vars are not supported in Realm Swift object classes

image

markst: https://github.com/realm/realm-cocoa/issues/2840#issuecomment-356177886

PR #5164, opened 2017-07-25
Updates for Xcode 9 beta 4

jpsim:
unfortunately, due to [rdar://33514802](http://www.openradar.me/33514802), Realm Objective-C's Swift tests don't compile or run.


Realm/Tests/Swift/RLMTestCaseUtils.swift
```diff
@@ -22,3 +22,14 @@ extension RLMTestCase {
             RLMAssertThrowsWithReasonMatchingSwift(self, { _ = block() }, regexString, message, fileName, lineNumber)
     }
 }
+
+#if !swift(>=3.2)
```
austinzheng: Would this be clearer as `<`? Can you even use `<` in this directive?

tgoyne: No, `>=` is the only allowed comparison.

PR #5165, opened 2017-07-25
Update SUPPORT.md

aviwarner:
Made a couple small changes to clarify what channels are best for what kind of issue, and revised instructions for sharing code samples that require confidentiality.

### DIFF
--- a/SUPPORT.md
+++ b/SUPPORT.md
@@ -8,12 +8,12 @@ Before asking questions, please familiarize yourself with our [Swift](https://re
 
 ## Stack Overflow
 
-If you have questions about Realm you can ask them on Stack Overflow. We continually monitor the [`realm` tag](https://stackoverflow.com/tags/realm). Please also tag your question with `ios`, `swift`, or other tags as appropriate.
+If you have questions about configuring or using Realm you can ask them on Stack Overflow. We continually monitor the [`realm` tag](https://stackoverflow.com/tags/realm). Please also tag your question with `ios`, `swift`, or other tags as appropriate.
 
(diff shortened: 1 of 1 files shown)


SUPPORT.md
```diff
@@ -8,12 +8,12 @@ Before asking questions, please familiarize yourself with our [Swift](https://re
...
+If you are running into issues with Realm, including potential bugs or feature requests, we encourage you to file an issue on our [GitHub issue tracker](https://github.com/realm/realm-cocoa/issues). Please check out our [Contribution Guidelines](CONTRIBUTING.md) for information on how to properly file an issue.
 
-We greatly appreciate demonstration projects that we can run for ourselves in order to see issues or potential bugs; we prioritize clearly-written tickets that include reproduction cases. You may attach these to the ticket or send these to help@realm.io; all emails and code sent to that address will be kept confidential.
+We greatly appreciate demonstration projects that we can run for ourselves in order to see issues or potential bugs; we prioritize clearly-written tickets that include reproduction cases. You may attach these to the ticket; let us know if you need to share them confidentially, and we’ll provide instructions on how to do so.
```
jpsim: Why request that our users ask us how to share stuff with us rather than instruct them here directly, which is what the previous wording does? Seems like adding unnecessary steps...

aviwarner: We're trying to minimize issues that come into help@realm.io directly. We already request that users email help@ during the course of working on GitHub issues, and that works well. Given that it's a relatively rare occurrence, I'd prefer this document reflect 'top-level' channels and when to use them.

PR #5167, opened 2017-07-25
Enable running Swift sources of Realm Objective-C tests in Swift 4

jpsim:
by applying a workaround to [rdar://33514802](http://www.openradar.me/33514802) as suggested here: https://twitter.com/aikoniv/status/889961905464397824

### DIFF
--- a/Realm/Swift/RLMSupport.swift
+++ b/Realm/Swift/RLMSupport.swift
@@ -57,7 +57,7 @@ public struct RLMIterator<T>: IteratorProtocol {
         iteratorBase = NSFastEnumerationIterator(collection)
     }
 
-    public func next() -> T? {
+    public mutating func next() -> T? {
         return iteratorBase.next() as! T?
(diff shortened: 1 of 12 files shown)


jpsim:
I had to change this a bit to fix support for Xcode 8.x and fallout from #5148. @austinzheng do you mind confirming that this is alright by you?

austinzheng:
The changes look fine to me. I assume adding `mutating` won't break the Xcode 8 builds.

austinzheng:
It looks like these changes might still be problematic? Many of the CI jobs are failing because of compilation failures.

jpsim:
> It looks like these changes might still be problematic? Many of the CI jobs are failing because of compilation failures.

Digging through the CI logs, none of it made sense based on the contents of this PR, leading me to think that we were hit by realm/ci#109 again.

I reran the jobs and everything looks good now.

PR #5101, opened 2017-07-07
enable xcode 9 in CI

larkost:

PR #5170, opened 2017-07-26
update examples for Swift 4

jpsim:


jpsim:
Extracted out of #5101, which has been inactive for two weeks.

Really, we should remove the Swift version symlinks and just have a single version. We used to need symlinks because you couldn't conditionally compile different versions of Swift from the same source, but now all the Swift versions we support can do that. #5171

PR #5176, opened 2017-07-26
Merge master as of v2.9.0 into master-3.0

bdash:
The only interesting change _should_ be in the change log, where I replicated the entries from v2.9.0 that were new since v3.0.0-beta into the unreleased section. The only conflicts were in dependencies.list and the Realm framework's Info.plist, and both were resolved by keeping the version from master-3.0.


bdash:
My intention was for each of the 3.0.0-beta change log entries to contain information about the changes that are new in that release compared to the release on which it was based (i.e., v3.0.0-beta2's section mentions changes from v3.0.0-beta, even if they were present in v2.9.0). When we go to make our final 3.0.0 release, we'd have entries that summarize all changes that are new in the 3.0 release vs the most recent 2.x release.

PR #5177, opened 2017-07-26
mark the GitHub release as 'prerelease' if version matches criteria

jpsim:


scripts/github_release.rb
```diff
@@ -43,7 +43,8 @@ def release_notes(version)
 github.access_token = ENV['GITHUB_ACCESS_TOKEN']
 
 puts 'Creating GitHub release'
-response = github.create_release(REPOSITORY, RELEASE, name: RELEASE, body: RELEASE_NOTES)
+prelease = (VERSION =~ /alpha|beta|rc/) ? true : false
```
bdash: s/prelease/prerelease/, here and in two places on the next line.

jpsim: oops, nice catch 😅

jpsim:
FYI: http://octokit.github.io/octokit.rb/Octokit/Client/Releases.html#create_release-instance_method

PR #5178, opened 2017-07-26
Update object store to avoid making non-additive schema changes to synced Realms

bdash:
This pulls in the fix from realm/realm-object-store#504.


bdash:
I'm not sure there's a lot of value in duplicating tests in both object store and cocoa, particularly when the behavior is implemented entirely at the object store level and can be tested entirely at the object store level.

Issue #1103, opened 2014-11-07
distinct query in Realm database


alexwillrock:
hello, realm is super fast db
but very necessary distinct query from database e.g.  schedules = Schedules.objectsWhere("areas.id = '(idAreas)' distict")

ajimix:
+1 here :)

kunalsood:
+1
This would be especially useful when number of sections in a grouped tableView is dynamic, iterating over all objects to find distinct sectionIdentifier properties after every change notification seems extremely inefficient, especially for large datasets.

It would be even better if this could be built into RLMResults or another similar class, and potentially support change/fine grain notifications in the future.

gabrielnika:
yes, something like @distinctUnionOfObjects is badly needed

stsmurf:
+1 - 10,000+ records with a lot of duplicates, just need to know the distinct/unique values in a column!

anbarasu0504:
+1

marcoscurvello:
+1

HiromichiYamada:
+1

roughboy:
+1

mandrachek:
+1

kevinmlong:
For Swift 1.2 (due to the contains func) you could do something like this

```swift
import Foundation
import RealmSwift

extension Results {
    
    func uniqueValueForObject<U : Equatable>(objectKey: String, paramKey: String, type: U.Type)->[U]{
        
        var uniqueValues : [U] = [U]()
        
        for obj in self {
            let obj = obj as T
            if(obj.valueForKey(objectKey) != nil){
                let uniqueObj: Object = obj.valueForKey(objectKey) as! Object
                if(uniqueObj.valueForKey(paramKey) != nil){
                    let uniqueObjValue = uniqueObj.valueForKey(paramKey) as! U
                    if !contains(uniqueValues, uniqueObjValue) {
                        uniqueValues.append(uniqueObjValue)
                    }
                }
            }
            
        }
        return uniqueValues
    }
    
    func uniqueValue<U : Equatable>(paramKey: String, type: U.Type)->[U]{
        
        var uniqueValues : [U] = [U]()
        
        for obj in self {
            let obj = obj as T
            if(obj.valueForKey(paramKey) != nil){
                let uniqueValue = obj.valueForKey(paramKey) as! U
                if !contains(uniqueValues, uniqueValue) {
                    uniqueValues.append(uniqueValue)
                }
            }
            
        }
        return uniqueValues
    }
[...]

PR #5182, opened 2017-07-28
Expose distinct query via ObjC and Swift APIs

qubblr:
First I'd like to thank you guys for all the incredible work you are doing!

Ability to get distinct and sorted data is vital to my new project. According to [this PR](https://github.com/realm/realm-object-store/pull/467) the object store is ready so I followed @jpsim's [suggestion](https://github.com/realm/realm-cocoa/issues/1103#issuecomment-287484504) and exposed it via a new method on Results.

Hope this helps.

To do:
- [ ] tests
- [ ] documentation

### DIFF
--- a/Realm/RLMResults.h
+++ b/Realm/RLMResults.h
@@ -176,6 +176,8 @@ NS_ASSUME_NONNULL_BEGIN
  */
 - (RLMResults<RLMObjectType> *)sortedResultsUsingDescriptors:(NSArray<RLMSortDescriptor *> *)properties;
 
+- (RLMResults *)distinctResultsUsingProperty:(NSString *)property;
+
 #pragma mark - Notifications
 
 /**
--- a/Realm/RLMResults.mm
+++ b/Realm/RLMResults.mm
@@ -348,6 +348,27 @@ static inline void RLMResultsValidateInWriteTransaction(__unsafe_unretained RLMR
     });
 }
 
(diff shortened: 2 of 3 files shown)


jpsim:
Thanks for the PR and your interest in doing this @qubblr! I've marked this as "in-progress" until you address the tests and documentation items you mentioned. We might want to bikeshed on the naming a bit too, but what you have seems like a good starting point. We'll definitely want to refer to "key paths" and not "properties" though, since we moved our terminology to that a few months ago.

And if the object store implementation doesn't yet support nested key paths (it doesn't _appear_ to at quick glance) we'll need to update that as well.

bdash:
> And if the object store implementation doesn't yet support nested key paths (it doesn't appear to at quick glance) we'll need to update that as well.

It should, AFAICT, since it takes a `DistinctDescriptor`, which [contains a vector of column indexes](https://github.com/realm/realm-core/blob/develop/src/realm/views.hpp#L48-L54).

ppaulojr:
@qubblr I have interest in this feature as well. If you want I can help with tests and docs. Let me know.

bdash:
This was closed automatically due to my merging of `master-3.0` into `master`. Reopening.

FreudGit:
@bdash May I please ask what are the status.  Do we not have a way to filter distinct() ?  Thanks

gcox:
Can we go ahead and close this since it's handled in #5467

PR #5183, opened 2017-07-28
log error message when theer is no device

larkost:


larkost:
This appears to be working successfully:
https://ci.realm.io/job/objc_pr/configuration=Release,target=tvos-device,xcode_version=8.3.3/5366/console

This is a minimal fix, and I will be filing a more complete fix for later.

PR #5184, opened 2017-07-28
Try to fix sporadic CI test failure

austinzheng:
There is a family of CI object server test failures that all look something like this:

```
RLMRealm *realm = // ...open a Realm
doSomething();
CHECK_COUNT(3, MyTestObject, realm);   // this passes
doSomethingElse();
waitForDownloads();
CHECK_COUNT(5, MyTestObject, realm);   // this fails; count is still 3
```

They are all impossible to reproduce, happen in different tests, but all take that same basic structure.

My hypothesis is that, since Realms are not refreshed until the current runloop iteration is complete, that there exists a possibility where `waitForDownloads()` completes quickly enough the second `CHECK_COUNT()` call is made before the runloop has had a chance to complete and the Realm has a chance to refresh.

### DIFF
--- a/Realm/ObjectServerTests/RLMObjectServerTests.mm
+++ b/Realm/ObjectServerTests/RLMObjectServerTests.mm
@@ -1280,7 +1280,7 @@
         NSUInteger sizeBefore = fileSize(c.pathOnDisk);
         XCTAssertGreaterThan(sizeBefore, 0U);
         XCTAssertNotNil(RLMGetAnyCachedRealmForPath(c.pathOnDisk.UTF8String));
-        [self waitForExpectationsWithTimeout:15.0 handler:nil];
+        [self waitForExpectationsWithTimeout:10.0 handler:nil];
         XCTAssertGreaterThan(fileSize(c.pathOnDisk), sizeBefore);
         XCTAssertNotNil(RLMGetAnyCachedRealmForPath(c.pathOnDisk.UTF8String));
         CHECK_COUNT(NUMBER_OF_BIG_OBJECTS, HugeSyncObject, realm);
--- a/Realm/ObjectServerTests/RLMObjectServerTests.mm
+++ b/Realm/ObjectServerTests/RLMObjectServerTests.mm
@@ -1254,6 +1254,7 @@
         // Wait for the child process to upload everything.
         RLMRunChildAndWait();
         XCTestExpectation *ex = [self expectationWithDescription:@"download-realm"];
+        XCTestExpectation *ex2 = [self expectationWithDescription:@"wait for downloads after asyncOpen"];
         RLMRealmConfiguration *c = [RLMRealmConfiguration defaultConfiguration];
         RLMSyncConfiguration *syncConfig = [[RLMSyncConfiguration alloc] initWithUser:user realmURL:url];
         c.syncConfiguration = syncConfig;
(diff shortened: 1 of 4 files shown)


austinzheng:
I think my hypothesis was not correct, but I had another hypothesis, so I made more fixes. Will also run the tests locally to see if I can see any difference.

My new hypothesis is this: ROS takes some time to perform actions (e.g. process permission changes or get changes ready to be downloaded), but we don't account for that in our tests, so if ROS is slow the tests will fail.

austinzheng:
I rebased this onto `master` (from `master-3.0`), since it's not breaking and since once it's in we can easily fold it into the WIP branch (but not vice versa).

Realm/ObjectServerTests/RLMObjectServerTests.mm
```diff
@@ -1277,7 +1280,7 @@ - (void)testDownloadAlreadyOpenRealm {
         NSUInteger sizeBefore = fileSize(c.pathOnDisk);
         XCTAssertGreaterThan(sizeBefore, 0U);
         XCTAssertNotNil(RLMGetAnyCachedRealmForPath(c.pathOnDisk.UTF8String));
-        [self waitForExpectationsWithTimeout:10.0 handler:nil];
+        [self waitForExpectationsWithTimeout:15.0 handler:nil];
```
jpsim: 15 seconds seems very long, even for large objects... did you arrive at this through experimentation? I'm not opposed to leaving it in if lower values sometimes cause CI failures, but I want to make sure this isn't leftovers from your debugging.

Realm/ObjectServerTests/RLMSyncTestCase.mm
```diff
@@ -294,20 +294,27 @@ - (RLMSyncUser *)makeAdminUser:(NSString *)userName password:(NSString *)passwor
...
+    BOOL queued = [session waitForDownloadCompletionOnQueue:dispatch_get_global_queue(DISPATCH_QUEUE_PRIORITY_LOW, 0)
+                                                   callback:^(NSError *err){
+                                                       theError = err;
+                                                       [ex fulfill];
+                                                   }];
+    if (!queued) {
+        XCTFail(@"Download waiter did not queue; session was invalid or errored out.");
+        return;
+    }
+    [self waitForExpectations:@[ex] timeout:20.0];
```
jpsim: Same comment here regarding the very high timeout value.

Realm/ObjectServerTests/RLMSyncTestCase.mm
```diff
@@ -322,12 +329,16 @@ - (void)waitForUploadsForUser:(RLMSyncUser *)user url:(NSURL *)url error:(NSErro
...
+    BOOL queued = [session waitForUploadCompletionOnQueue:dispatch_get_global_queue(DISPATCH_QUEUE_PRIORITY_LOW, 0)
+                                                 callback:^(NSError *err){
+                                                     theError = err;
+                                                     [ex fulfill];
+                                                 }];
+    if (!queued) {
+        XCTFail(@"Upload waiter did not queue; session was invalid or errored out.");
+        return;
+    }
+    [self waitForExpectations:@[ex] timeout:20.0];
```
jpsim: Same comment here regarding the very high timeout value.

jpsim:
If the high timeout values are indeed required, we should still report this on the sync repo as an area worth optimizing.

Realm/ObjectServerTests/RLMObjectServerTests.mm
austinzheng: This particular timeout isn't problematic and I will remove it; like you said it was leftovers from experimentation.

Realm/ObjectServerTests/RLMSyncTestCase.mm
austinzheng: This one (and the one below) do seem to be making a difference. Unfortunately, everything here is guesswork since the tests never fail in this way on my MBP and they don't even fail all the time on our CI machines. I earlier made changes to try to narrow down the cause of this expectation failing and reached the conclusion that the issue simply was `waitForDownloads` not being given enough time to complete.

austinzheng:
I agree with filing an issue. I'd like to merge in this change and let it "bake" for a while (see if it results in a trend of improved CI performance), then file an issue if it does look like this was the culprit.

austinzheng:
Good news! The object server tests failed not because of anything I did, but because of the "child process" issue.

jpsim:
> Good news! The object server tests failed not because of anything I did, but because of the "child process" issue.

are we tracking that one?

austinzheng:
https://github.com/realm/ci/issues/145

PR #5189, opened 2017-07-31
Rename `addOrUpdateObjectsInArray:`

austinzheng:
Changes:
- Renamed `-[RLMRealm addOrUpdateObjectsInArray:]` to just `addOrUpdateObjects:` to match other APIs
- The three add/update/delete methods now actually check their collection elements to avoid unsafe memory access

Supersedes #3836.

### DIFF
--- a/Realm/RLMRealm.mm
+++ b/Realm/RLMRealm.mm
@@ -675,7 +675,7 @@ REALM_NOINLINE void RLMRealmTranslateException(NSError **error) {
 
 - (void)addObjects:(id<NSFastEnumeration>)objects {
     for (RLMObject *obj in objects) {
-        if (![obj isKindOfClass:[RLMObject class]]) {
+        if (![obj isKindOfClass:RLMObjectBase.class]) {
             @throw RLMException(@"Cannot insert objects of type %@ with addObjects:. Only RLMObjects are supported.",
                                 NSStringFromClass(obj.class));
         }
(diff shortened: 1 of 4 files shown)


Realm/RLMRealm.mm
```diff
@@ -702,22 +707,21 @@ - (void)deleteObject:(RLMObject *)object {
...
 }
 
-- (void)deleteObjects:(id)array {
-    if ([array respondsToSelector:@selector(realm)] && [array respondsToSelector:@selector(deleteObjectsFromRealm)]) {
-        if (self != (RLMRealm *)[array realm]) {
+- (void)deleteObjects:(id<NSFastEnumeration>)objects {
+    NSObject *collection = (NSObject *)objects;
+    if ([collection respondsToSelector:@selector(realm)]
+        && [collection respondsToSelector:@selector(deleteObjectsFromRealm)]) {
+        if (self != (RLMRealm *)[collection performSelector:@selector(realm)]) {
```
tgoyne: Just cast to `id` rather than using `performSelector`. It's less verbose and there's only a functional difference if there's multiple possible return types for that selector (with the cast to `id` it'll fail to compile; with `performSelector` it'll produce a warning and leak memory if you ignore that).

Realm/RLMRealm.mm
```diff
@@ -692,8 +692,13 @@ - (void)addOrUpdateObject:(RLMObject *)object {
     RLMAddObjectToRealm(object, self, true);
 }
 
-- (void)addOrUpdateObjectsFromArray:(id)array {
-    for (RLMObject *obj in array) {
+- (void)addOrUpdateObjects:(id<NSFastEnumeration>)objects {
+    for (RLMObject *obj in objects) {
+        if (![obj isKindOfClass:[RLMObject class]]) {
```
tgoyne: These checks (including the pre-existing one in addObjects:) should be for RLMObjectBase.

Realm/RLMRealm.mm
austinzheng: That's good to know. Thanks!

austinzheng:
Trying out a fresh CI build, with the most recent test fixes and build script fixes pulled in, just to see what happens.

PR #5190, opened 2017-07-31
fix bash scoping issue

larkost:


austinzheng:
The object server test failure is an example of https://github.com/realm/ci/issues/145 and unrelated to this commit.

larkost:
@austinzheng I am not sure what you are trying to say, but this one is straightforward. What is described in realm/ci#145 is not similar to this problem, or related to this fix.

austinzheng:
Right, I am saying you can ignore the object server failure because it's unrelated to the fix you're making here.

larkost:
Neither of these failures is related, so we are good here.

PR #5192, opened 2017-07-31
Fix the way generic Swift properties are added to schema

### LINKED ISSUE #5102: Realm Swift schema properties not in declared order in some cases
If a Realm Swift object is defined with non-generic properties following generic properties (e.g. `List<T>`), the order in the schema will not match the order the properties were declared in. This is because of how the schema creation logic works. There are also implications for object creation using arrays (since users expect to pass in an array whose objects correspond to the property values in the order they were declared).

austinzheng:
Changes:
- Generic Swift properties should now be added to schema in the order they are defined

Fixes #5102

### DIFF
--- a/RealmSwift/Object.swift
+++ b/RealmSwift/Object.swift
@@ -416,23 +416,43 @@ public class ObjectUtil: NSObject {
         return nil
     }
 
-    // Get the names of all properties in the object which are of type List<>.
-    @objc private class func getGenericListPropertyNames(_ object: Any) -> NSArray {
-        return Mirror(reflecting: object).children.filter { (prop: Mirror.Child) in
-            return type(of: prop.value) is RLMListBase.Type
-        }.flatMap { (prop: Mirror.Child) in
-            return prop.label
-        } as NSArray
+    // Reflect an object, and remove all the ignored stored properties from the
+    // children collection.
+    private static func getNonIgnoredMirrorChildren(for object: Any) -> [Mirror.Child] {
+        guard let realmObject = object as? Object else {
+            return Array(Mirror(reflecting: object).children)
+        }
+        let ignoredPropNames: Set<String> = Set(type(of: realmObject).ignoredProperties())
+        // No HKT in Swift, unfortunately
+        return Mirror(reflecting: realmObject).children.filter { (prop: Mirror.Child) -> Bool in
+            guard let label = prop.label else {
+                return false
+            }
+            return !ignoredPropNames.contains(label)
+        }
     }
 
-    // Return the indices of generic properties in the object.
-    @objc private class func getGenericListPropertyIndices(_ object: Any) -> NSArray {
-        let children = Mirror(reflecting: object).children
-        let indices = Array(0..<Int(children.count))
-        let mask = children.map { (prop: Mirror.Child) -> Bool in
-            return type(of: prop.value) is RLMListBase.Type
+    @objc private class func getListProperties(_ object: Any) -> [RLMListPropertyMetadata] {
+        var props: [RLMListPropertyMetadata] = []
+        for (idx, prop) in getNonIgnoredMirrorChildren(for: object).enumerated() {
+            if let value = prop.value as? RLMListBase {
+                props.append(RLMListPropertyMetadata(propertyName: prop.label!, index: idx))
+            }
         }
-        return zip(indices, mask).filter { $0.1 }.map { $0.0 } as NSArray
+        return props
+    }
+
+    @objc private class func getLinkingObjectsProperties(_ object: Any) -> [RLMLinkingObjectsPropertyMetadata] {
+        var props: [RLMLinkingObjectsPropertyMetadata] = []
+        for (idx, prop) in getNonIgnoredMirrorChildren(for: object).enumerated() {
+            if let value = prop.value as? LinkingObjectsBase {
+                props.append(RLMLinkingObjectsPropertyMetadata(propertyName: prop.label!,
+                                                               className: value.objectClassName,
+                                                               linkedPropertyName: value.propertyName,
+                                                               index: idx))
+            }
+        }
+        return props
     }
 
     // swiftlint:disable:next cyclomatic_complexity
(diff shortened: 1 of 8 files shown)


RealmSwift/Object.swift
```diff
@@ -419,6 +419,16 @@ public class ObjectUtil: NSObject {
         } as NSArray
     }
 
+    // Return the indices of generic properties in the object.
+    @objc private class func getGenericListPropertyIndices(_ object: Any) -> NSArray {
+        let children = Mirror(reflecting: object).children
+        let indices = Array(0..<Int(children.count))
```
austinzheng: Need to verify that `children` actually corresponds to properties on the object, and that (e.g.) child number 0 is actually the first property etc

austinzheng:
I'm 99% sure I'll need to go back and do more with this.

Children doesn't imply just stored properties. It _may_ include method names, computed properties, superclass members, and other things that would make this code incorrect as-is.

RealmSwift/Object.swift
```diff
@@ -410,13 +410,43 @@ public class ObjectUtil: NSObject {
...
-        } as NSArray
+    // Reflect an object, and remove all the ignored stored properties from the
+    // children collection.
+    private static func getNonIgnoredMirrorChildren(for object: Any) -> [Mirror.Child] {
+        guard let realmObject = object as? Object else {
+            return Array(Mirror(reflecting: object).children)
+        }
+        let ignoredPropNames: Set<String> = Set(type(of: realmObject).ignoredProperties())
+        // No HKT in Swift, unfortunately
+        return Mirror(reflecting: realmObject).children.filter { (prop: Mirror.Child) -> Bool in
```
tgoyne: This probably also needs to handle lazy properties in some way (they produce a stored property that we have to explicitly filter out in other places).

RealmSwift/Tests/SwiftTestObjects.swift
```diff
@@ -481,3 +481,36 @@ class SwiftCircleObject: Object {
...
+
+class SwiftGenericPropsOrderingParent: Object {
+    let parentFirstList = List<SwiftIntObject>()
+    @objc dynamic var parentFirstNumber = 0
+    func parentFunction() -> Int { return parentFirstNumber + 1 }
+    @objc dynamic var parentSecondNumber = 1
+    var parentComputedProp: String { return "hello world" }
+}
+
+class SwiftGenericPropsOrderingObject: SwiftGenericPropsOrderingParent {
```
tgoyne: Other things that would be good to test:

1. A `@objc dynamic` ignored property.
2. A `@objc dynamic` computed property.
3. A lazy property
4. A `RealmOptional` property

austinzheng:
Thanks for the review. Thanks to your suggestions I identified a number of additional potential issues and rewrote my code to address them.

RealmSwift/Object.swift
```diff
@@ -410,69 +410,109 @@ public class ObjectUtil: NSObject {
...
+            }
+            return true
         }
     }
 
-    @objc private class func requiredPropertiesForClass(_: Any) -> [String] {
-        return []
+    // Build optional property metadata for a given property.
+    // swiftlint:disable:next cyclomatic_complexity
+    private static func buildMetadata(for child: Mirror.Child, at index: Int) -> RLMGenericPropertyMetadata? {
```
jpsim: Maybe rename to `getSwiftOptionalProperties` for consistency with `getSwiftGenericProperties`?

RealmSwift/Object.swift
```diff
@@ -410,69 +410,109 @@ public class ObjectUtil: NSObject {
...
+            type is RealmOptional<Int32>.Type ||
+            type is RealmOptional<Int64>.Type {
+            code = .int
+        } else if type is RealmOptional<Float>.Type {
+            code = .float
+        } else if type is RealmOptional<Double>.Type {
+            code = .double
+        } else if type is RealmOptional<Bool>.Type {
+            code = .bool
+        } else if child.value as? RLMOptionalBase != nil {
```
jpsim: shouldn't this be `if child.value is RLMOptionalBase`?

RealmSwift/Object.swift
```diff
@@ -410,69 +410,109 @@ public class ObjectUtil: NSObject {
...
-        }.flatMap { (prop: Mirror.Child) in
-            (prop.label!, prop.value as! LinkingObjectsBase)
-        }
-        return properties.reduce([:]) { (dictionary, property) in
-            var d = dictionary
-            let (name, results) = property
-            d[name] = ["class": results.objectClassName, "property": results.propertyName]
-            return d
+    @objc private class func getSwiftGenericProperties(_ object: Any) -> [RLMGenericPropertyMetadata] {
+        var props: [RLMGenericPropertyMetadata] = []
```
jpsim: flatMap would be a good fit here:

```swift
return getNonIgnoredMirrorChildren(for: object).enumerated().flatMap { idx, prop in
    if let value = prop.value as? LinkingObjectsBase {
        return .init(forLinkingObjectsProperty: prop.label!,
                     className: value.objectClassName,
                     linkedPropertyName: value.propertyName,
                     index: idx)
    } else if prop.value is RLMListBase {
        return .init(forListProperty: prop.label!, index: idx)
    } else if let optional = buildMetadata(for: prop, at: idx) {
        return optional
    }
    return nil
}
```

RealmSwift/Object.swift
```diff
@@ -410,69 +410,109 @@ public class ObjectUtil: NSObject {
...
             }
-            return properties
+            if let lazyBaseName = baseName(forLazySwiftProperty: label) {
+                if ignoredPropNames.contains(lazyBaseName) {
+                    // Ignored lazy property.
+                    return false
+                }
+                // Managed lazy property; not currently supported.
+                // FIXME: revisit this once Swift gets property behaviors/property macros.
+                throwRealmException("Lazy managed property '\(lazyBaseName)' is not allowed on a Realm Swift object"
```
jpsim: could you write a test confirming that this exception is thrown for non-ignored lazy properties? Could use our `FakeObject` utilities like we do for other invalid schema tests.

RealmSwift/Object.swift
austinzheng: Good call.

RealmSwift/Object.swift
```diff
@@ -410,70 +410,111 @@ public class ObjectUtil: NSObject {
...
-        }.flatMap { (prop: Mirror.Child) in
-            return prop.label
-        } as NSArray
+    // If the property is a storage property for a lazy Swift property, return
+    // the base property name (e.g. `foo.storage` becomes `foo`). Otherwise, nil.
+    private static func baseName(forLazySwiftProperty name: String) -> String? {
+        // A Swift lazy var shows up as two separate children on the reflection tree:
+        // one named 'x', and another that is optional and is named 'x.storage'. Note
+        // that '.' is illegal in either a Swift or Objective-C property name.
+        let storageSuffix = ".storage"
```
jpsim: Avoid the copy & mutation:

```diff
-        let storageSuffix = ".storage"
-        if let storageRange = name.range(of: storageSuffix, options: [.anchored, .backwards]) {
-            var baseName = name
-            baseName.removeSubrange(storageRange)
-            return baseName
+        if let storageRange = name.range(of: ".storage", options: [.anchored, .backwards]) {
+            return name.substring(to: storageRange.lowerBound)
```

Issue #5025, opened 2017-06-12
Align notification API with Swift 4 KVO API


tgoyne:
Swift 3.2/4's new block-based API for KVO is similar enough to our existing API for notifications that we should probably make them the same whereever possible.

Some things we may want to do:

1. Rename `stop()` to `invalidate()`, with a deprecated alias for Swift <4. It's a little unfortunate that we already use `invalidate()` for something different, but it doesn't directly conflict.
2. Rename `addNotificationBlock` to `observe()`. I think this won't conflict with using the native `observe()` method on objects (the KVO one requires a keypath argument), but need to double-check. Does this have the potential to be confusing? We already see some confusion between our `filter()` method on `Results` and the Swift-native `filter()` method.
3. Remove `NotificationToken` from the API entirely and return `NSKeyValueObservation`. Probably a bad idea.

If we do rename things, should we also do the same for objective-c?

austinzheng:
I'm going to provisionally put the `breaking` tag on this, because it's conceivable we could make it as a breaking change for 3.0 if the timeframe works. If we think it's better to not break the API even across 2.* to 3.0, I'll remove the tag.

jpsim:
> Rename `stop()` to `invalidate()`

Sure.

> Rename `addNotificationBlock` to `observe()`

Definitely.

> Remove `NotificationToken` from the API [...] Probably a bad idea.

Agreed, let's not do that.

---

Let's keep this as `breaking`. The fewer ways to do the same thing, the better.

jpsim:
Done in #5193.

PR #5193, opened 2017-08-01
Align notification APIs with new KVO APIs

jpsim:
Addresses #5025. I opted to leave the Objective-C `addNotificationBlock` API as-is, because `observe` isn't a very idiomatic ObjC method name. However, the ObjC API _is_ affected by this (i.e. `-[RLMNotificationToken invalidate]`) in order to keep our terminology aligned between Swift & ObjC and the first argument of idiomatic naming doesn't apply here.

### DIFF
--- a/RealmSwift/LinkingObjects.swift
+++ b/RealmSwift/LinkingObjects.swift
@@ -322,7 +322,7 @@ public final class LinkingObjects<T: Object>: LinkingObjectsBase {
      ```swift
      let results = realm.objects(Dog.self)
      print("dogs.count: \(dogs?.count)") // => 0
-     let token = dogs.addNotificationBlock { changes in
+     let token = dogs.observe { changes in
          switch changes {
          case .initial(let dogs):
              // Will print "dogs.count: 1"
@@ -351,7 +351,7 @@ public final class LinkingObjects<T: Object>: LinkingObjectsBase {
      - parameter block: The block to be called whenever a change occurs.
      - returns: A token which must be held for as long as you want updates to be delivered.
      */
-    public func addNotificationBlock(_ block: @escaping (RealmCollectionChange<LinkingObjects>) -> Void) -> NotificationToken {
+    public func observe(_ block: @escaping (RealmCollectionChange<LinkingObjects>) -> Void) -> NotificationToken {
         return rlmResults.addNotificationBlock { _, change, error in
             block(RealmCollectionChange.fromObjc(value: self, change: change, error: error))
         }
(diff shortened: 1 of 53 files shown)


austinzheng:
It looks like the object server tests may need to be updated; they seem to not have built.

jpsim:
Fixed in 6e96202

austinzheng:
Thanks.

Sorry for making you have to do that; eventually we'll be getting rid of `RLMSyncPermissionsResults` as its own base type altogether (but we still need the changes to get your PR in; it shouldn't be blocked on my stuff).

jpsim:
I'll add the missing docs tomorrow morning.

tgoyne:
What impact does this have on using the native `observe()` function on `Object`? The method signatures are sufficiently different that I wouldn't expect any issues other than slightly worse autocomplete, but we don't have any tests which use it.

austinzheng:
It looks like there are still a bunch of KVO tests failing. Not sure if they're related.

jpsim:
Only failing on Xcode 8.0. I guess I have to download that now.

RealmSwift/LinkingObjects.swift
```diff
@@ -436,10 +436,3 @@ internal enum LinkingObjectsBridgingMetadata {
         }
     }
 }
-
-// MARK: Unavailable
```
austinzheng: Why was this removed? It provides Xcode fix-its for users who are migrating from Realm 2.* to 3 and still using the deprecated APIs.

RealmSwift/List.swift
```diff
@@ -524,10 +524,3 @@ extension List: AssistedObjectiveCBridgeable {
         return (objectiveCValue: _rlmArray, metadata: nil)
     }
 }
-
-// MARK: Unavailable
-
-extension List {
```
austinzheng: Ditto.

austinzheng:
Looks good, my comments notwithstanding. Looks like a rebase is needed.

RealmSwift/LinkingObjects.swift
jpsim: Because the migration assistance this provides is already provided by the more generic ones on `RealmCollection`, so it's redundant to add it here as well.

RealmSwift/List.swift
jpsim: Because the migration assistance this provides is already provided by the more generic ones on `RealmCollection`, so it's redundant to add it here as well.

PR #5197, opened 2017-08-02
Fix sorting collections on iOS 7

### LINKED ISSUE #5194: 2.9.0 crash with sortedResultsUsingKeyPath
## Goals
Upgraded to 2.9.0 so that he can run the 2.8.3 version of the code. No changes were made to any code

## Expected Results
Normal execution

## Actual Results
carsh

````
Terminating app due to uncaught exception 'NSInvalidArgumentException', reason: '-[__NSCFConstantString containsString:]: unrecognized selector sent to instance 0x8fdd90'
````

## Code Sample
````objective-c
    RLMResults<OddityChannel *> *results =  [[[OddityChannel allObjectsInRealm:realm] objectsWhere:@"is_children = false"] sortedResultsUsingKeyPath:@"orderindex" ascending:true];
    
````

Realm framework version: 2.9.0

Realm Object Server version:   Not in use

Xcode version: Version 8.3.3 (8E3004b)

iOS/OSX version:  7.1.2

Dependency manager + version: Cocoapods  1.3.0.rc.1

bdash:
`-[NSString containsString:]` wasn't added until iOS 8, so attempting to use it on iOS 7 would throw an unrecognized selector exception. Switch to using `-[NSString rangeOfString:]` for compatibility with iOS 7.

This also enables `-Wpartial-availability`, which results in warnings when using symbols that are not available in all OS versions that we're building for. It revealed a similar issue with our use of `-containsString:` in unit tests, and a false-positive within RLMPredicateUtil.mm.

I've not been able to run the tests on iOS 7 as I don't have any devices with such an old version of iOS. The iOS 7 simulator is only supported back on macOS 10.10, and no version of Xcode that will build Realm from source runs on that version of macOS either.

Fixes #5194.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,7 @@ x.x.x Release notes (yyyy-MM-dd)
 
 ### Bugfixes
 
-* None.
+* Sorting Realm collection types no longer throws an exception on iOS 7.
 
 2.9.1 Release notes (2017-08-01)
 =============================================================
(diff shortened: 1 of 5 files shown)


austinzheng:
Should fix #5194, or at least is a prerequisite for fixing it.

PR #5200, opened 2017-08-03
Make empty string matching match Foundation semantics

austinzheng:
Foundation never matches an empty or null string with `BEGINSWITH`, `ENDSWITH` and `CONTAINS`. Core's query engine treats them as always matching since they're logically a match at any character position within the string. To emulate Foundation's semantics we generate additional criteria on the core query.

Replaces #3613. I tried rebasing that PR but it's nigh-impossible.

To do:
- [x] Make the code less hideous
- [x] More tests for the diacritic-insensitive case and binary case
- [x] Implement support for binary data properties

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -48,6 +48,9 @@ x.x.x Release notes (yyyy-MM-dd)
 * Remove deprecated `{RLM}SyncManager.disableSSLValidation` property. Disable
   SSL validation on a per-Realm basis by setting the `enableSSLValidation`
   property on `{RLM}SyncConfiguration` instead.
+* Fix empty strings, binary data, and null on the right side of `BEGINSWITH`,
+  `ENDSWITH` and `CONTAINS` operators in predicates to match Foundation's
+  semantics of never matching any strings or data.
 
 ### Enhancements
 
--- a/Realm/RLMQueryUtil.mm
+++ b/Realm/RLMQueryUtil.mm
@@ -528,6 +528,12 @@ public:
     template <typename A, typename B>
     void add_bool_constraint(NSPredicateOperatorType operatorType, A lhs, B rhs);
 
+    void add_substring_constraint(null, Query condition);
+    template<typename T>
+    void add_substring_constraint(const T& value, Query condition);
+    template<typename T>
+    void add_substring_constraint(const Columns<T>& value, Query condition);
+
     template <typename T>
     void add_string_constraint(NSPredicateOperatorType operatorType,
                                NSComparisonPredicateOptions predicateOptions,
@@ -551,9 +557,9 @@ public:
 
     void add_between_constraint(const ColumnReference& column, id value);
 
-    template<typename T>
-    void add_binary_constraint(NSPredicateOperatorType operatorType, const ColumnReference& column, T value);
+    void add_binary_constraint(NSPredicateOperatorType operatorType, const ColumnReference& column, BinaryData value);
     void add_binary_constraint(NSPredicateOperatorType operatorType, const ColumnReference& column, id value);
+    void add_binary_constraint(NSPredicateOperatorType operatorType, const ColumnReference& column, null);
     void add_binary_constraint(NSPredicateOperatorType operatorType, id value, const ColumnReference& column);
     void add_binary_constraint(NSPredicateOperatorType, const ColumnReference&, const ColumnReference&);
 
@@ -606,7 +612,8 @@ void QueryBuilder::add_numeric_constraint(RLMPropertyType datatype,
             break;
         default:
             @throw RLMPredicateException(@"Invalid operator type",
-                                         @"Operator '%@' not supported for type %@", operatorName(operatorType), RLMTypeToString(datatype));
+                                         @"Operator '%@' not supported for type %@",
+                                         operatorName(operatorType), RLMTypeToString(datatype));
     }
 }
 
@@ -625,6 +632,27 @@ void QueryBuilder::add_bool_constraint(NSPredicateOperatorType operatorType, A l
     }
 }
 
+void QueryBuilder::add_substring_constraint(null, Query) {
+    // Foundation always returns false for substring operations with a RHS of null or "".
+    m_query.and_query(std::unique_ptr<Expression>(new FalseExpression));
+}
+
+template<typename T>
+void QueryBuilder::add_substring_constraint(const T& value, Query condition) {
+    // Foundation always returns false for substring operations with a RHS of null or "".
+    m_query.and_query(value.size()
+                      ? std::move(condition)
+                      : std::unique_ptr<Expression>(new FalseExpression));
+}
+
+template<typename T>
+void QueryBuilder::add_substring_constraint(const Columns<T>& value, Query condition) {
+    // Foundation always returns false for substring operations with a RHS of null or "".
+    // We don't need to concern ourselves with the possibility of value traversing a link list
+    // and producing multiple values per row as such expressions will have been rejected.
+    m_query.and_query(const_cast<Columns<String>&>(value).size() != 0 && std::move(condition));
+}
+
 template <typename T>
 void QueryBuilder::add_string_constraint(NSPredicateOperatorType operatorType,
                                          NSComparisonPredicateOptions predicateOptions,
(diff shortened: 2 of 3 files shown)


bdash:
This also needs updated to work for binary data properties.

austinzheng:
Perhaps I'm missing something, but it looks like we already did the "right" thing for binary properties before this PR? From an existing test:

```objc
NSData *data = [NSData data];
RLMAssertCount(BinaryObject, 0U, @"binaryCol BEGINSWITH %@", data);
RLMAssertCount(BinaryObject, 0U, @"binaryCol ENDSWITH %@", data);
RLMAssertCount(BinaryObject, 0U, @"binaryCol CONTAINS %@", data);
```

bdash:
What about null?

Realm/RLMQueryUtil.mm
```diff
@@ -625,6 +635,50 @@ void add_collection_operation_constraint(NSPredicateOperatorType operatorType,
...
+    // Foundation always returns false for substring operations with a RHS of null or "".
+    m_query.and_query(value.size()
+                      ? std::move(condition)
+                      : std::unique_ptr<Expression>(new FalseExpression));
+}
+
+// Diacritic-sensitive version.
+void QueryBuilder::add_substring_constraint(const Columns<String>& value, Query&& condition) {
+    // Foundation always returns false for substring operations with a RHS of null or "".
+    // We don't need to concern ourselves with the possibility of column2 traversing a link list
```
austinzheng: This should read `value`, not `column2`.

Realm/RLMQueryUtil.mm
```diff
@@ -762,6 +822,27 @@ void validate_and_extract_between_range(id value, RLMProperty *prop, id *from, i
     m_query.end_group();
 }
 
+void QueryBuilder::add_binary_substring_constraint(size_t index,
+                                                   StringData value,
+                                                   Query&(Query::*constraint_maker)(size_t, StringData, bool)) {
+    // Foundation always returns false for substring operations with a RHS of null or "".
+    if (value.size()) {
+        (m_query.*constraint_maker)(index, std::move(value), true);
```
austinzheng: Unfortunately, because of the differing arities of the string and binary versions of this `Query` API, I don't think there's a way to collapse these two overloads.

CHANGELOG.md
```diff
@@ -41,6 +41,9 @@ x.x.x Release notes (yyyy-MM-dd)
 * The callback which runs when a sync user login succeeds or fails
   now runs on the main queue by default, or can be explicitly specified
   by a new `callbackQueue` parameter on the `{RLM}SyncUser.logIn(...)` API.
+* Fix empty strings and null on the right side of `BEGINSWITH`, `ENDSWITH` and
+  `CONTAINS` operators in predicates to match Foundation's semantics of never
+  matching any strings.
```
bdash: This makes it sound like the change is specific to strings, while it also impacts `NSData` properties.

Realm/RLMQueryUtil.mm
```diff
@@ -625,6 +635,50 @@ void add_collection_operation_constraint(NSPredicateOperatorType operatorType,
...
+}
+
+// Diacritic-sensitive version.
+void QueryBuilder::add_substring_constraint(const Columns<String>& value, Query&& condition) {
+    // Foundation always returns false for substring operations with a RHS of null or "".
+    // We don't need to concern ourselves with the possibility of column2 traversing a link list
+    // and producing multiple values per row as such expressions will have been rejected.
+    m_query.group();
+    m_query.and_query(value != null() && value != "");
+    m_query.and_query(std::move(condition));
```
bdash: I think this could be:
```c++
m_query.group();
m_query.and_query(value.size() != 0 && condition);
m_query.end_group();
```

Realm/RLMQueryUtil.mm
```diff
@@ -625,6 +635,50 @@ void add_collection_operation_constraint(NSPredicateOperatorType operatorType,
...
+    m_query.end_group();
+}
+
+// Diacritic-insensitive version.
+template<typename ConstraintAdder, typename P>
+void QueryBuilder::add_substring_constraint(const StringData& value,
+                                            const P& comparator,
+                                            ConstraintAdder&& add_constraint) {
+    if (value.size()) {
+        add_constraint(comparator);
```
bdash: It seems surprising that this function takes `add_constraint` and expects it to mutate `m_query`. It'd make more sense for this to take the string comparison query as an argument and call `and_query` here to combine them instead.

I think that would also allow these functions to be merged with the diacritic-sensitive versions?

Realm/RLMQueryUtil.mm
bdash: Is this overload taking `StringData` even used? It looks like `QueryBuilder::add_substring_constraint(const StringData& value, Query&& condition)` is used for the `StringData` case, while `add_binary_substring_constraint` is specific to `BinaryData`s?

Instead of taking the function pointer and calling it here, we could have the caller produce a`Query` that we'd `and_query` into `m_query` here. I'd expect that to let us handle both binary and string in the same function template.

Realm/RLMQueryUtil.mm
austinzheng: I have to remove `value`'s `const` qualifier to use `size()`, unfortunately.

bdash: Hmm, we should fix `size()` to be const since there's no reason I can see for it to be non-const. I'd suggest doing that, and working around that for now using a `const_cast`.

austinzheng:
I attempted to address as many of @bdash's review items as possible.

Unfortunately, core throws an exception if the LHS of the predicate is a binary column, the RHS is`NULL`, and I try to eagerly evaluate the predicate using the string versions of the helpers, so the binary-specific helpers needed to be retained.

austinzheng:
Did some squashing and commit fixup. I think it's ready.

Realm/RLMQueryUtil.mm
```diff
@@ -636,13 +664,13 @@ void add_collection_operation_constraint(NSPredicateOperatorType operatorType,
     if (diacriticSensitive) {
         switch (operatorType) {
             case NSBeginsWithPredicateOperatorType:
-                m_query.and_query(column.begins_with(value, caseSensitive));
+                add_substring_constraint(std::forward<T>(value), column.begins_with(value, caseSensitive));
```
bdash: You want `std::move` rather than `std::forward` here, since `value` isn't a forwarding reference.

Realm/RLMQueryUtil.mm
```diff
@@ -762,22 +796,24 @@ void validate_and_extract_between_range(id value, RLMProperty *prop, id *from, i
...
-template<typename T>
 void QueryBuilder::add_binary_constraint(NSPredicateOperatorType operatorType,
                                          const ColumnReference& column,
-                                         T value) {
+                                         BinaryData value) {
     RLMPrecondition(!column.has_links(), @"Unsupported operator", @"NSData properties cannot be queried over an object link.");
 
     size_t index = column.index();
+    const Table& table = *m_query.get_table();
+    Query query(table);
```
bdash: Oh, I guess this can probably be `Query query = m_query.get_table().where()`?

Realm/RLMQueryUtil.mm
austinzheng: Yeah, I thought `value` was a `T&&` for some reason :(. In this case, though, I made a different mistake. I have to pass `value` just as `value`, since moving it here is going to screw up constructing the second parameter (which also uses `value`), and the first argument takes a `const T&` anyways.

bdash: Good catch.

Realm/RLMQueryUtil.mm
```diff
@@ -663,15 +663,18 @@ void add_collection_operation_constraint(NSPredicateOperatorType operatorType,
 
     if (diacriticSensitive) {
         switch (operatorType) {
-            case NSBeginsWithPredicateOperatorType:
-                add_substring_constraint(std::forward<T>(value), column.begins_with(value, caseSensitive));
+            case NSBeginsWithPredicateOperatorType: {
```
bdash: Why add braces?… Why add them to only a subset of the cases? 😅

austinzheng: Ahhh! I was originally going to construct the queries as temporary variables and then move them into the `add_substring_constraint` call, but then realized that was unnecessary.

austinzheng:
CI failures are jobs timing out.

PR #5208, opened 2017-08-07
Work around a hang seen when building Realm using Xcode 9b5 with Carthage

### LINKED ISSUE #5207: Realm Cocoa fails to build with Carthage under Xcode9beta5
## Goals
Build Realm cocoa with Xcode9 (beta 5) for use with iOS11 testing

## Expected Results
Carthage build of Realm Cocoa would complete successfully 

## Actual Results
Carthage buid fails with a timeout on RealmExamples.xcworkspace:

`xcodebuild timed out while trying to read RealmExamples.xcworkspace 😭`

## Steps to Reproduce

1. Download Xcode9beta5 (released 7-Aug-17)
2. Set it to be the default version of Xcode with xcode-select
3. Create a new single view iOS11 project
4. Create. new Cartfile in the project root directory for Realm, it should contain 1 line: `github "realm/realm-cocoa" "master"`
5. Attempt to build the Carthage artifacts with `carthage bootstrap` or `carthage update` from the command line
6. After 5-6 mins the build will fail with the following error:
`xcodebuild timed out while trying to read RealmExamples.xcworkspace 😭`

This doesn't fail with code 8.3.3 and did it fail with Xcode9beta4.

## Code Sample
N/A

Realm framework version: N/A - building from Realm master which as present is 

Realm Object Server version: N/A

Xcode version: ? 
Xcode9Beta5 (9M202q)

iOS/OSX version: ? 
10.12.6 (16G29)

Carthage v0.24.0

bdash:
Xcode 9 beta 5 appears to hang when computing the build settings for the PreloadedData scheme in the tvOS examples project. At the point of the hang, `xcodebuild` is initializing some portion of the code signing subsystem. The only thing interesting about the PreloadedData target is that it has a development team specified, unlike all of the other targets in our example projects. Removing the development team prevents the hang. There's no reason for a development team to be specified for that target anyway.

Fixes #5207 (works around the issue, really).

### DIFF
--- a/examples/tvos/swift-3.0/RealmExamples.xcodeproj/project.pbxproj
+++ b/examples/tvos/swift-3.0/RealmExamples.xcodeproj/project.pbxproj
@@ -219,7 +219,6 @@
 					};
 					14AACA851BFE7B740046BD85 = {
 						CreatedOnToolsVersion = 7.2;
-						DevelopmentTeam = QX5CR2FTN2;
 					};
 				};
 			};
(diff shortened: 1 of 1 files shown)


jpsim:
Does the fact that we also specify the team in `Realm.xcodeproj` also cause issues? https://github.com/realm/realm-cocoa/blob/v2.9.1/Realm.xcodeproj/project.pbxproj#L1922

bdash:
No, it doesn't appear to.

bdash:
`carthage build --no-skip-current` with this branch gets around to building the frameworks, while with master it errors out prior to building anything. I've not yet waited for it to finish building, since it takes so long for Carthage to build everything repeatedly.

PR #5209, opened 2017-08-07
reactivate ios10 device

larkost:


larkost:
None of the failures looks related at all, and the iOS 10 device succeeded. So I think we should re-enable this.

jpsim:
How come this didn't show up in the matrix?

image

larkost:
I have no idea why it did not show up in the matrix view. Going from the links on GitHub I can clearly see that it is part of this job, and it is clearly in the JSON API describing this job. The `.jenkins.yml` is also identical (per `git diff`) to the checkin immediately before this was disabled. I am not able to verify that that one showed it in the matrix view as our retention policy puts those a day or two outside the window.

So my best response unfortunately is: Jenkins glitch? We can watch it to make sure?

jpsim:
Probably a bug in the matrix plugin.

PR #5210, opened 2017-08-07
Update to latest object store

austinzheng:
Changes:
- Various internal compatibility changes to use latest object store

NOTE: There are no user-facing object store changes between the previous and current versions, AFAIK.

NOTE: Hold merging until CI completes successfully.

### DIFF
--- a/Realm/RLMSyncUser.mm
+++ b/Realm/RLMSyncUser.mm
@@ -36,7 +36,6 @@
 
 using namespace realm;
 using ConfigMaker = std::function<Realm::Config(std::shared_ptr<SyncUser>, std::string)>;
-using PermissionGetCallback = std::function<void(Results, std::exception_ptr)>;
 
 namespace {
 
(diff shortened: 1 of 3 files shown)


Realm/RLMSyncUser.mm
```diff
@@ -36,7 +36,7 @@
 
 using namespace realm;
 using ConfigMaker = std::function<Realm::Config(std::shared_ptr<SyncUser>, std::string)>;
-using PermissionGetCallback = std::function<void(std::unique_ptr<PermissionResults>, std::exception_ptr)>;
+using PermissionGetCallback = std::function<void(Results, std::exception_ptr)>;
```
bdash: Why do we redefine this typedef rather than using `Permission::PermissionResultsCallback`?

austinzheng: We should just be using the object store typedef here.

PR #5213, opened 2017-08-08
enable Xcode 9 in testing

larkost:


.jenkins.yml
```diff
@@ -46,6 +47,9 @@ configuration:
...
 # | 8.3.3 | Debug        | X   |      |            | X           | X         | X         | X       |           |           | X    |                |                   |                      |                       |                       |                        |             |
 # | 8.3.3 | Release      | X   | X    |            | X           | X         | X         | X       | X         | X         | X    | X              | X                 | X                    |                       | X                     |                        | X           |
+# | -------------------- | --- | ---- | ---------- | ----------- | --------- | --------- | ------- | --------- | --------- | ---- | -------------- | ----------------- | -------------------- | --------------------- | --------------------- | ---------------------- | ----------- |
+# | 9.0   | Debug        | X   |      | X          |             |           |           |         |           |           |      |                |                   |                      |                       |                       |                        |             |
```
jpsim: We can't run `ios-static` on Xcode 9 for the same reasons we can't run it in Xcode 8.3.3.

build.sh
```diff
@@ -197,7 +197,12 @@ test_ios_static() {
 
     destination="$1"
     xc "-scheme 'Realm iOS static' -configuration $CONFIGURATION -sdk iphonesimulator -destination '$destination' build"
-    xc "-scheme 'Realm iOS static' -configuration $CONFIGURATION -sdk iphonesimulator -destination '$destination' test 'ARCHS=\$(ARCHS_STANDARD_32_BIT)'"
+    if [[ `echo $REALM_XCODE_VERSION | awk '{tmp=match($0, /^[[:digit:]]+/); print substr($0, RSTART, RLENGTH)}'` -lt 9 ]]; then
```
bdash: It's hard to tell precisely what this is doing. My assumption is that it's extracting the first digit of the Xcode version number? If so, `${REALM_XCODE_VERSION%%.*}` would be a more concise way of going about it (that expands to `REALM_XCODE_VERSION` with the the longest substring matching `.*` removed from the end of it). I'd also be inclined to wrap this up in a helper function, since we appear to perform this same test elsewhere. That'd let us write this comparison more like:

```sh
if (( xcode_version_major < 9 )); then
fi
```

build.sh
```diff
@@ -197,7 +197,12 @@ test_ios_static() {
...
+        xc "-scheme 'Realm iOS static' -configuration $CONFIGURATION -sdk iphonesimulator -destination '$destination' test 'ARCHS=\$(ARCHS_STANDARD_32_BIT)'"
+    else
+        # in Xcode 9 iOS 11 is the standard, and it does not support 32-bit iOS configurations
+        xc "-scheme 'Realm iOS static' -configuration $CONFIGURATION -sdk iphonesimulator -destination '$destination' test"
```
bdash: We test the 64-bit configuration below (~line 213). This looks like it'll result in it being tested twice for Xcode 9.

build.sh
jpsim: can this check be extracted into a function since it's copied 3 times in here?

examples/installation/ios/swift-4.0
```diff
@@ -1 +0,0 @@
-swift-3.1
```
jpsim: what changed in all these symlinks?

scripts/reset-simulators.rb
```diff
@@ -2,6 +2,9 @@
 
 require 'json'
 
+# use default DEVELOPER_DIR if not already set
```
jpsim: There are many changes in here that I assume came out of problems with the previous implementation. It'd be nice to know why each change was made. It's challenging to review this without that context.

scripts/reset-simulators.rb
```diff
@@ -17,55 +20,70 @@ def platform_for_device_type(device_type)
...
-  running_devices(devices).each do |device|
-    puts "Shutting down simulator #{device['udid']}"
-    system("xcrun simctl shutdown #{device['udid']}") or puts "    Failed to shut down simulator #{device['udid']}"
-  end
-end
-
 begin
+  # - Check if the wrong version of the CoreSimulatorService is running, stop it if so
+  if ENV['JENKINS_HOME'] || ! /^\d+\s+#{ENV['DEVELOPER_DIR']}\/.+/ =~ `/usr/bin/pgrep -lf com.apple.CoreSimulator.CoreSimulatorService 2>/dev/null` ; then    
+    # Kill all the current simulator processes as they are from a different Xcode version
```
bdash: Can you please explain why you're taking a different approach to cleaning up simulator processes?

scripts/reset-simulators.rb
```diff
@@ -17,55 +20,70 @@ def platform_for_device_type(device_type)
...
-  # Shut down any simulators that need it.
-  running_devices(devices).each do |device|
-    puts "Shutting down simulator #{device['udid']}"
-    system("xcrun simctl shutdown #{device['udid']}") or puts "    Failed to shut down simulator #{device['udid']}"
-  end
-end
-
 begin
+  # - Check if the wrong version of the CoreSimulatorService is running, stop it if so
+  if ENV['JENKINS_HOME'] || ! /^\d+\s+#{ENV['DEVELOPER_DIR']}\/.+/ =~ `/usr/bin/pgrep -lf com.apple.CoreSimulator.CoreSimulatorService 2>/dev/null` ; then
```
bdash: Why are we doing this unconditionally in CI, but otherwise doing it conditionally? What's wrong with always doing it unconditionally, like we were in the past.

scripts/reset-simulators.rb
```diff
@@ -17,55 +20,70 @@ def platform_for_device_type(device_type)
...
+    # Get the list of non-shutdown devices
+    running_devices = all_available_devices.reject { |device| device['state'] == 'Shutdown' }
+    
+    break if running_devices.empty?
+    
+    # Shutdown the running devices
+    running_devices.each do |device|
+      puts "  Shutting down simulator #{device['udid']}"
+      system("xcrun simctl shutdown #{device['udid']}") or puts "    Failed to shut down simulator #{device['udid']}"
+    end
```
bdash: I'm not really clear why you've chosen to move code that was previously in functions and inline it here. It makes it harder to see what's changed.

scripts/reset-simulators.rb
```diff
@@ -17,55 +20,70 @@ def platform_for_device_type(device_type)
...
-
 begin
+  # - Check if the wrong version of the CoreSimulatorService is running, stop it if so
+  if ENV['JENKINS_HOME'] || ! /^\d+\s+#{ENV['DEVELOPER_DIR']}\/.+/ =~ `/usr/bin/pgrep -lf com.apple.CoreSimulator.CoreSimulatorService 2>/dev/null` ; then    
+    # Kill all the current simulator processes as they are from a different Xcode version
+    print "Killing running Simulator processes and replacing with one from #{ENV['DEVELOPER_DIR']}..."
+    
+    if ! `/usr/bin/pgrep -qx Simulator`.empty?
+      `/usr/bin/osascript -e 'tell app "Simulator" to quit without saving'`
+      sleep 0.2 # we are sending AppleEvents, not a kill signal
```
bdash: Why bother asking the simulator nicely to quit when we're just going to go ahead and kill it later on? What benefit does this provide?

examples/installation/ios/swift-4.0
larkost: They were all files when I checked out that version. So I just policed them all to be symlinks.

scripts/reset-simulators.rb
larkost: This offers the best chance that all of the supporting files will be in a good state.

bdash: What do you mean by that? Is this fixing a specific, concrete problem, or a hypothetical one?

If this is attempting to fix a real problem, surely we need to do better than hoping that 200ms is sufficient for it to quit.

scripts/reset-simulators.rb
larkost: 1. If we are testing in CI then we absolutely want to kill everything.
2. On a developer workstation we don't want to kill things unless we have to. I am waffling a little in shutting down all of the running simulators, but since we are not actively making sure that we use one simulator for the whole process, my hands are a bit tied there.
3. If the Xcode we are using is not the one that launched the Simulator background process, then we need to kill it. Otherwise things will likely go wrong.

scripts/reset-simulators.rb
larkost: Answer is one step down.

scripts/reset-simulators.rb
bdash: This appears to be an answer as to my other question as to why we're doing things conditionally, not to why we're changing the approach we're using to kill simulator-related processes.

I don't think 2 is a compelling reason. I've not heard any complaints from engineers on the Cocoa team about simulators being reset by this script. Our goal is to have local testing of changes via `build.sh` behave as closely as possible to testing in CI, and we should try and stick to that unless we have a good reason not to.

bdash:
Superseded by #5290 and #5291.

PR #5214, opened 2017-08-08
move SUPPORT.md to .github/

jpsim:
to avoid cluttering the root directory of the repo


bdash:
I think having it at the top level, alongside CONTRIBUTING.md, is fine.

jpsim:
Your call 🤷‍♂️

PR #5219, opened 2017-08-09
Add swift_objc_members attribute to RLMObjectBase

jpsim:
which will allow Swift 3.x syntax RLMObject/Object subclasses to continue working in Swift 4.x (i.e. without marking `dynamic` properties as also being `@objc`).

Depends on #5218.

### DIFF
--- a/Realm/ObjectServerTests/SwiftSyncTestCase.swift
+++ b/Realm/ObjectServerTests/SwiftSyncTestCase.swift
@@ -20,11 +20,11 @@ import XCTest
 import RealmSwift
 
 class SwiftSyncObject: Object {
-    @objc dynamic var stringProp: String = ""
+    dynamic var stringProp: String = ""
 }
 
 class SwiftHugeSyncObject: Object {
-    @objc dynamic var dataProp: NSData?
+    dynamic var dataProp: NSData?
 
     required init() {
         super.init()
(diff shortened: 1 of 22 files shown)


tgoyne:
Should we also remove `@objc` from all our tests and sync models if this makes it unnecessary?

bdash:
The downside of doing this is that it's not at all obvious to users that other properties and types in their classes are going to have subtly different semantics than if they didn't inherit from `Object`. Requiring `@objc` be specified on each property or `@objcMembers` be used makes the different behavior explicit.

jpsim:
> Should we also remove `@objc` from all our tests and sync models if this makes it unnecessary?

Sure, let's do that.

> their classes are going to have subtly different semantics

I don't think so, since it appears to only really affect properties/functions that are also marked as `dynamic`, and so far the `dynamic` keyword is only implemented in terms of objective-c runtime dynamic dispatch.

Do you have a concrete example in mind where you'd get different semantics?

bdash:
The examples in [SE-160](https://github.com/apple/swift-evolution/blob/master/proposals/0160-objc-inference.md) demonstrate a number of cases where you'll see different semantics.

Two cases stick out:
```swift
class MyClass : Object {
  func foo() { }
}
```
`MyClass.foo()` would not be callable from Objective-C in Swift 4 unless explicitly annotated as `@objc`, but is if `Object` is declared as `@objcMembers`.

```swift
class MyClass : Object {
  func foo(_ int: Int) { }
  func foo(_ double: Double) { } 
}
```
This would be valid in Swift 4 without `@objcMembers`, but is invalid with it due to the two functions having the same Obj-C selector.

jpsim:
Thanks for surfacing those examples. IMO that's reason enough to not do this. Closing.

PR #5220, opened 2017-08-09
Collect more information on sporadic test failure

austinzheng:
The good news is that certain types of object server test CI failures we used to get seem to be gone for good. The bad news is that there are other types of failures that continue to occasionally happen. I want to gather more information on exactly what is going on.


austinzheng:
After talking to the Java folks we're 99% sure the failure in question has to do with a ROS subsystem that is being rewritten in the near future anyways, so I'm going to close this.

PR #5225, opened 2017-08-10
Allow read-only permission Realms to be opened offline

### LINKED ISSUE #5186: Cannot open Read Only Synced Realms without internet connection
## Goals

Trying to use a read-only synced realm to propagate data to clients. This will require reliable opening of the realm, with (after sync) or without internet connection (using cached local copy) at the time of application startup.

## Expected Results

An API to reliably open a read only synced realm during application launch. 

## Actual Results

When online, Realm syncs and then opens with Async API (correct). Note: Read only synced realms can only be opened with AsyncOpen because conventional open will write data to the realm, according to other threads.

When offline, AsyncOpen will not return (to completion block) as it cannot connect to the server. Which is right, but leaves us with no way to open the realm.

RLMSyncManager error handler reports error with RLMSyncSession.

## Steps to Reproduce

Async open a read only synced realm. Try again when disabling Internet connection.

## Code Sample

```objc
NSURL *syncServerURL = [NSURL URLWithString: @"XXXXXXX"];

RLMRealmConfiguration *config = [RLMRealmConfiguration defaultConfiguration];

config.syncConfiguration = [[RLMSyncConfiguration alloc] initWithUser:loggedInUser realmURL:syncServerURL];


// Open the remote Realm
[RLMRealm asyncOpenWithConfiguration:config
                       callbackQueue:dispatch_get_main_queue()
                            callback:^(RLMRealm *realm, NSError *error)
 {
     if  (error)
     {
         NSLog(@"Error opening racing realm: %@", error);
         
     }
     else
     {
         if (realm)
         {
             NSLog(@"Opened Racing Realm");
             
         }
         else
         {
             NSLog(@"Error opening racing realm: Unspecified error");
             
         }
     }
 }];
```    

Realm framework version: 2.8.3

Realm Object Server version: 

Xcode version: 8.3

iOS/OSX version: 10

Dependency manager + version: CocoaPod

austinzheng:
Changes:
- Added a version of `asyncOpen(...)` that takes a flag indicating whether to wait for a synced Realm to connect and download all pending changes before invoking the completion callback

Fixes #5186 

API improvement suggestions are welcome.

Not sure how to test this. Maybe I can use Katsumi's network conditioner tests to completely remove the connection, then try opening a Realm.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,9 @@ x.x.x Release notes (yyyy-MM-dd)
   a Realm Object Server.
 * An error handler can now be registered on `{RLM}SyncUser`s in order to
   report authentication-related errors that affect the user.
+* Add a variant API to asynchronously open a synced Realm for which the user
+  has read-only permissions and immediately return its contents before the
+  latest updates have been 
 
 ### Bugfixes
 
--- a/Realm/RLMRealm.h
+++ b/Realm/RLMRealm.h
@@ -106,12 +106,21 @@ NS_ASSUME_NONNULL_BEGIN
  synchronized Realms wait for all remote content available at the time the
  operation began to be downloaded and available locally.
 
- @param configuration A configuration object to use when opening the Realm.
- @param callbackQueue The dispatch queue on which the callback should be run.
- @param callback      A callback block. If the Realm was successfully opened,
-                      it will be passed in as an argument.
-                      Otherwise, an `NSError` describing what went wrong will be
-                      passed to the block instead.
+ @param configuration   A configuration object to use when opening the Realm.
+ @param waitForDownload If the configuration describes a synchronized Realm, whether
+                        to immediately return the Realm or to only return the Realm in
+                        the callback once the connection with the Realm Object Server
+                        has been established, and all pending changes downloaded. If
+                        the configuration describes a non-synced Realm, this argument
+                        is ignored. Set this to NO if you are trying to open a synced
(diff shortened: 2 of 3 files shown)


jpsim:
Seems like what you really want here is a synchronous open that wouldn't write the schema to the file? Overloading `asyncOpen` for this use case seems like a poor fit.

austinzheng:
Sure, it would be preferable if users could just open their Realms using the normal means, but I am wondering how much work that would be and whether we can get such a solution out to users in the near future.

austinzheng:
There's a better way to do this, and it depends upon https://github.com/realm/realm-object-store/pull/511. Stay tuned...

astigsen:
>  takes a flag indicating whether to wait for a synced Realm to connect and download all pending changes before invoking the completion callback

Does this distinguish between the case where you have accessed the realm previously, so you have a previous state cached on (which may or may not be up-to-date, no way to know until you get a connection), which you can return immediately (even without connectivity), and opening the Realm for the first time, where if there is no connectivity you would have to return an empty Realm?

austinzheng:
It doesn't distinguish between those two cases; if the user opened the Realm previously they'll get back the Realm in that latest local state; if they never opened the Realm at all they'll get back an empty Realm that will fill with objects as sync occurs.

The use case that originally prompted this was bundling a copy of a read-only Realm with the application. The idea is that a user could open it with connectivity and get the latest updates from the server, or open it without connectivity and at least have the pre-bundled version of the data.

PR #5226, opened 2017-08-10
Move sync error JSON models into the shared models file

austinzheng:
No functional change.


austinzheng:
Both CI failures are spurious.

austinzheng:
All three CI job failures are timeout failures.

PR #5227, opened 2017-08-10
Specify Realm type using an enum

### LINKED ISSUE #4852: Realm Configuration Sanity Check
When modified via the `Configuration` object, `Realm` instances can represent one of three different Realm types.

* Local Realms
* In-memory Realms
* Synchronized Realms

These states are mutually exclusive and Realms created in one context can't be converted to another. They also each have specific properties in the `Configuration` object governing how the type of the Realm is controlled:

* `fileURL` for local Realms
* `inMemoryIdentifier` for in-memory Realms
* `syncConfiguration` for synchronized Realms

When creating Realms for one context, it doesn't make sense to also modify the properties for the other contexts (eg setting an `inMemoryIdentifier` value for a synchronized Realm). 

At the moment, Realm will accept a `Configuration` containing all of these properties and will set up a `Realm` instance in the same reliable internal logic, but to avoid user confusion, I'm wondering if it's worth adding a console warning to let users know that they shouldn't be setting these properties in specific circumstances.

Are there any opinions on this?

## Goals
Let Realm be more vocal and concise about what it's going to do with a Realm `Configuration` object when properties representing conflicting Realm types are modified.

## Expected Results
If modifying conflicting properties like `syncConfiguration` and `inMemoryIdentifier`, Realm to log a warning in the console letting the developer know that one property will be ignored, and it shouldn't be set.

## Actual Results
Absolutely nothing.

austinzheng:
Also make `SyncConfiguration` struct `Equatable`.

Fixes #4852.

### DIFF
--- a/RealmSwift/RealmConfiguration.swift
+++ b/RealmSwift/RealmConfiguration.swift
@@ -22,6 +22,43 @@ import Realm.Private
 
 extension Realm {
     /**
+     An enum that describes the different kinds of Realms that can be created.
+     */
+    public enum Kind: Equatable {
+        /**
+         Describes a non-synchronized Realm backed by a file saved to disk. The URL
+         is the local filesystem URL to the Realm file.
+         */
+        case file(URL)
+        /**
+         Describes a non-synchronized Realm backed by memory. The string is used to
+         identify the Realm.
+         */
+        case inMemory(String)
+        /**
+         Describes a synchronized Realm. The sync configuration is used to specify
+         additional configuration details.
+         
+         - see: `SyncConfiguration`
+         */
+        case synced(SyncConfiguration)
+
+        /// :nodoc:
+        public static func == (lhs: Kind, rhs: Kind) -> Bool {
+            switch (lhs, rhs) {
+            case let (.file(lhsURL), .file(rhsURL)):
+                return lhsURL == rhsURL
+            case let (.inMemory(lhsIdentifier), .inMemory(rhsIdentifier)):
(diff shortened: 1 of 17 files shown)


RealmSwift/RealmConfiguration.swift
```diff
@@ -269,3 +264,39 @@ extension Realm.Configuration: CustomStringConvertible {
                     string: rlmConfiguration.description) ?? ""
     }
 }
+
+// MARK: Migration assistance
+
+public extension Realm.Configuration {
+    @available(*, unavailable, message: "Set `Realm.Configuration.kind` instead.")
```
austinzheng: Wondering if, instead of making these unavailable, they should just be get-only properties.

austinzheng:
I told Mark I'd think about names but I haven't come up with anything yet. In the meantime I think it makes sense to expose `fileURL` as a read-only property.

austinzheng:
What about `identity`? The commonality between file paths, in-memory identifiers, and sync configurations is that they all uniquely identify a Realm database.

Option 1:
```swift
class Realm {
  var identity: Realm.Configuration.Kind // e.g. .file(someFileURL)
  var fileURL: URL? { get }
}
```

Option 2:
```swift
class Realm {
  var identity: Realm.Configuration.Identifier // same as above
  var fileURL: URL? { get }
}
```

austinzheng:
After talking to @bdash, we've decided we don't have the bandwidth to get this into Realm 3 while ensuring a high quality user experience. I'm closing this PR for now.

PR #5236, opened 2017-08-15
fix error message when device is missing

larkost:


bdash:
@larkost, can you please add the appropriate labels to this PR to communicate what state it is in? As it stands it's not clear what the next step is here.

bdash:
@larkost, can you please clarify what work remains here (i.e., why is it marked as "In Active Progress" vs "Review"?).

It'd also be beneficial to explain what this change is doing (what issue is it fixing, and why it's the right change). Typically the commit messages and/or PR description are used for that purpose.

austinzheng:
Not enough context right now to know what to do with this. We can revive this in the future if necessary.

PR #5247, opened 2017-08-21
Update object store

austinzheng:


Realm/RLMRealm.mm
```diff
@@ -460,8 +461,8 @@ + (void)resetRealmState {
 
 - (void)verifyNotificationsAreSupported {
     [self verifyThread];
-    if (_realm->config().read_only()) {
-        @throw RLMException(@"Read-only Realms do not change and do not have change notifications");
+    if (_realm->config().immutable()) {
+        @throw RLMException(@"Immutable Realms do not change and do not have change notifications");
```
bdash: The term we use in this error message needs to match the term used in the Cocoa API, which remains read-only.

Realm/RLMRealm.mm
```diff
@@ -488,7 +489,7 @@ - (RLMNotificationToken *)addNotificationBlock:(RLMNotificationBlock)block {
 }
 
 - (void)sendNotifications:(RLMNotification)notification {
-    NSAssert(!_realm->config().read_only(), @"Read-only realms do not have notifications");
+    NSAssert(!_realm->config().immutable(), @"Immutable realms do not have notifications");
```
bdash: Ditto about terminology.

Realm/RLMRealmConfiguration.mm
```diff
@@ -202,7 +202,8 @@ - (void)setEncryptionKey:(NSData * __nullable)encryptionKey {
 }
 
 - (BOOL)readOnly {
-    return _config.read_only();
+    // FIXME: rename this immutable for Realm 3
```
bdash: A comment isn't a great way to track this. An issue would be preferable.

CHANGELOG.md
```diff
@@ -20,6 +20,9 @@ x.x.x Release notes (yyyy-MM-dd)
   example, users who are authenticated using third-party credentials will find
   themselves logged out of the Realm Object Server if the third-party identity
   service indicates that their credential is no longer valid.
+* Search depth when reporting changes to observation blocks has been
```
bdash: I'd suggest rephrasing this to focus on the symptom that's being fixed, and to note that it's related to collection notifications. Perhaps:

> Fix high CPU usage and hangs when processing collection notifications in highly-connected object graphs.

I'm not sure that it's worth mentioning the change to maximum depth?

PR #5265, opened 2017-08-24
Support optionals when specifying predicate varargs

### LINKED ISSUE #4197: Can RealmCollection.filter do better with Optional<T>?
In Swift 3 we changed RealmCollection.filter to take variadic arguments of type `Any`. This allows passing `Optional<T>`s as arguments, but they make it down to Objective-C code that has no knowledge of that type. Can our Swift wrapper unwrap the optionals, replacing `none` with `NSNull`, and `some(x)` with `x`?

austinzheng:
Fixes #4197.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,9 +11,6 @@ x.x.x Release notes (yyyy-MM-dd)
   types.
 * Remove `RLMPropertyTypeArray` in favor of a separate bool `array` property on
   `RLMProperty`/`Property`.
-* Realm Swift APIs that filter or look up the index of an object based on a
-  format string now properly handle optional arguments in their variadic argument
-  list.
 
 ### Enhancements
 
@@ -21,7 +18,9 @@ x.x.x Release notes (yyyy-MM-dd)
 
 ### Bugfixes
 
-* None.
+* Realm Swift APIs that filter or look up the index of an object based on a
+  format string now properly handle optional arguments in their variadic argument
+  list.
 
 3.0.0-beta.3 Release notes (2017-08-23)
 =============================================================
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,9 @@ x.x.x Release notes (yyyy-MM-dd)
   types.
 * Remove `RLMPropertyTypeArray` in favor of a separate bool `array` property on
   `RLMProperty`/`Property`.
+* Realm Swift APIs that filter or look up the index of an object based on a
+  format string now properly handle optional arguments in their variadic argument
+  list.
 
 ### Enhancements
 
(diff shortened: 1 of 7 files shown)


RealmSwift/Util.swift
```diff
@@ -27,6 +27,15 @@ internal func noWarnUnsafeBitCast<T, U>(_ x: T, to type: U.Type) -> U {
     return unsafeBitCast(x, to: type)
 }
 
+internal func unwrap(arguments: [Any]) -> [Any] {
```
bdash: This seems like an awfully generic name for such a special-purpose function. Not sure how much that matters, though, given that it's for internal use only.

RealmSwift/Tests/RealmCollectionTypeTests.swift
```diff
@@ -280,6 +280,18 @@ class RealmCollectionTypeTests: TestCase {
...
 
+    func testFilterWithAnyVarags() {
+        guard let collection = collection else {
+            fatalError("Test precondition failed")
+        }
+        let firstCriterion: String? = "1"
+        let secondCriterion: String = "2"
+        let thirdCriterion: String? = nil
+        let result = collection.filter("stringCol = %@ OR stringCol = %@ OR stringCol = %@",
+                                       firstCriterion as Any, secondCriterion as Any, thirdCriterion as Any)
```
bdash: Is the explicit cast to `Any` necessary? If so, why?

austinzheng: It's only necessary to get the Swift compiler to not emit an unhelpful warning.

austinzheng:
Something about Swift's unboxing of optionals in `Any` is broken only on Xcode 8.0. I'll have to figure out a workaround.

austinzheng:
There's some sort of Swift bug in the version of Swift that Xcode 8.0 ships with. Trying to figure out a workaround...

CHANGELOG.md
```diff
@@ -11,6 +11,9 @@ x.x.x Release notes (yyyy-MM-dd)
   types.
 * Remove `RLMPropertyTypeArray` in favor of a separate bool `array` property on
   `RLMProperty`/`Property`.
+* Realm Swift APIs that filter or look up the index of an object based on a
```
bdash: Is this really a breaking change rather than a bug fix? From #4197 it sounds like passing `none` through doesn't do anything useful at present, while it will after this change.

austinzheng: I think you could probably argue that this is a bugfix, since all the things this enables would have been considered incorrect code before. Would you like it retargeted to `master`?

bdash: I don't feel strongly about getting this into a 2.x release, so feel free to leave this on `master-3.0`. My question was mainly about whether this change belongs in the "Breaking changes" section of the change log.

PR #5264, opened 2017-08-24
Update objectstore

tgoyne:
The primitive arrays branch had some breaking changes, so this backports the necessary changes from the primitive arrays branch here.

PR #5168, opened 2017-07-25
Refactor `RLMSyncPermissionResults` to refine `RLMResults`

austinzheng:
~Work in progress~

Change `RLMSyncPermissionResults` to be a private subclass of `RLMResults<RLMSyncPermissionValue *>`, and change `SyncPermissionResults` to simply be a typealias for `Results<SyncPermissionValue>`.

This reduces our API surface, adds a bunch of missing functionality, and probably fixes some latent bugs at the same time.

TODO:
- [x] Implement the rest of RLMSyncPermissionResults
- [x] Swift-specific affordances?
- [x] Migration affordances?
- [x] More tests
- [x] Compatibility with older versions of Swift
- [x] ~Filter out "no rights at all" permissions~
- [x] Object store changes merged in (https://github.com/realm/realm-object-store/pull/507)
- [x] Prerequisite Cocoa changes (`tg/relax-collection-constraints`) merged into `master-3.0`
- [x] Retarget to `master-3.0`

Fixes #5075 
Fixes #5076

PR #5266, opened 2017-08-25
Fix an issue causing permissions tests to fail

austinzheng:
I overlooked the fact that #5264 caused the object server tests to break. The culprit was `RLMSyncPermissionResults`'s misuse of a protocol. (#5168 will get rid of this entire issue in Realm 3 by refactoring how results work.)

### DIFF
--- a/Realm/RLMSyncPermissionResults.mm
+++ b/Realm/RLMSyncPermissionResults.mm
@@ -149,6 +149,10 @@ using namespace realm;
     }
 }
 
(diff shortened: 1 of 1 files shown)


austinzheng:
Saw https://github.com/realm/ci/issues/145 again. Restarting CI.

austinzheng:
All further failures are Jenkins issues or CI getting stuck and not updating the webhook.

Issue #4189, opened 2016-10-06
Prevent auxiliary realm files from being backed up


jpsim:
We should be setting `NSURLIsExcludedFromBackupKey` for auxiliary Realm files to prevent them from being backed up as user-generated data.

PR #5270, opened 2017-08-28
Set `NSURLIsExcludedFromBackupKey` for auxiliary files to prevent them from being backed up

kishikawakatsumi:
Fixes https://github.com/realm/realm-cocoa/issues/4189

Set `NSURLIsExcludedFromBackupKey` for `*.realm.management ` `*.realm.lock` and `*.realm.note` to prevent them from being backed up.

cc @bdash @tgoyne

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@ x.x.x Release notes (yyyy-MM-dd)
 
 ### Enhancements
 
-* None.
+* Auxiliary files are excluded from backup by default.
 
 ### Bugfixes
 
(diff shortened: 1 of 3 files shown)


tgoyne:
Needs a changelog entry.

Does this have any meaningful performance impact on opening Realms that already exist? Some of the filesystem operations are sometimes surprisingly slow.

kishikawakatsumi:
@tgoyne I added changelog entry. I could not find any performance difference when opening an existing/new Realm file. I confirmed it with the device of iPhone 6 Plus, 10.3.3.

PR #5273, opened 2017-08-29
Rename generic type parameter on collections

### LINKED ISSUE #3991: Improve naming of generic parameters
We ought to use more descriptive generic parameter names. Our current naming leads to confusing error messages in pretty common cases.

``` swift
let x = results.sum(ofProperty: "foo")
```

For example, the above line of code will fail to compile with error "Generic parameter 'U' could not be inferred." What is `U`? Maybe it would be better named `Result` or `PropertyType`.

austinzheng:
The generic type parameter `T` on collection-like Realm types has been renamed `Element`, to better match the Swift stdlib's naming conventions. As such, the associated type `Element` on `RealmCollection` has been renamed `ElementType`.

I didn't rename the type parameters `S: Sequence` on some of the methods, ~or `U` on the aggregate methods (max, min, etc)~ since what they mean is clear in context and giving them a full name just made the type signature look more confusing. (_ed_: Renamed `U` to `T`, since there are no longer `T` parameters.)

I also renamed the param on `RealmOptional`.

Fixes #3991.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,11 @@ x.x.x Release notes (yyyy-MM-dd)
   using a `RLMResults`. This results collection supports all normal collection
   operations except for setting values using KVO (since `RLMSyncPermission`s are
   immutable) and the property aggregation operations.
+* `RealmCollection`'s associated type `Element` has been renamed `ElementType`.
+* Realm Swift collection types (`List`, `Results`, `AnyRealmCollection`, and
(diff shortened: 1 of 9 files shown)


bdash:
> or `U` on the aggregate methods (max, min, etc)

I suspect this only used `U` because `T` was taken on the enclosing class. `T` would be more conventional in this context, no?

austinzheng:
Yes, that is correct. `U` should become `T`.

PR #5274, opened 2017-08-29
Add support for arrays of more types

tgoyne:
~~This is currently targeting a branch with https://github.com/realm/realm-cocoa/pull/5267 and latest `master` merged into `master-3.0` for the sake of showing something sensible for review.~~

All functionality of RLMArray and List should now be supported for all of the Realm-supported data types, except for queries. Currently trying to query primitive lists just silently ignores the predicate passed in. See RLMTestObjects.h for examples of how the models are declared in obj-c (as with RLMObject arrays, generic type parameters can be used but are insufficient). In Swift the expected normal things work (e.g. `List<Int?>` etc.).

The bulk of the obj-c changes are adjusting how validation is done. Notably `unbox()` no longer validates its input and instead we validate all values before passing them to the object store. This is needed both for providing sensible error messages without threading even more awful state tracking into RLMAccessorContext, and to avoid duplicating all of the logic for unmanaged objects. The other obj-c changes are just adding support in the KVC and aggregate functions.

Swift changes were simpler; all that was left to do was make more types conform to RealmCollectionValue and make realm collections conditionally define aggregate functions for Lists and Results of aggregatable types.

The tests are unfortunately something of a mess. For obj-c I had the idea of generating tests on each array type from a template using a simple python script. This did not turn out as well as I hoped, and unsurprisingly the python script turned into a terrible pile of hacks and the template file is awful. For Swift I attempted to take the same approach as the C++ objectstore tests, but ran into a lot of issues with the current limitations of Swift generics (most notably the lack of conditional conformances). Both of these probably need to be rethought, but I think that can wait; the tests do at least serve to test the code for now.


austinzheng:
Nothing stands out to me as being problematic, it looks good. I'll review it again tomorrow more carefully.

bdash:
There's a linker error which suggests we'll need a core release including the fix from realm/realm-core#2823:
```
Undefined symbols for architecture i386:
  "unsigned long realm::Table::find_first<realm::BinaryData>(unsigned long, realm::BinaryData) const", referenced from:
      unsigned long realm::List::find<realm::BinaryData>(realm::BinaryData const&) const in libRealm.a(list.o)
      unsigned long realm::Results::index_of<realm::BinaryData>(realm::BinaryData const&) in libRealm.a(results.o)
ld: symbol(s) not found for architecture i386
```

Realm/RLMArray.h
```diff
@@ -67,9 +65,21 @@ NS_ASSUME_NONNULL_BEGIN
...
 /**
- The class name (i.e. type) of the objects contained in the array.
+ The type of the objects in the array.
+ */
+@property (nonatomic, readonly, assign) RLMPropertyType type;
+
+/**
+ Indicates whether the objects in the collection can be `nil`.
+ */
+@property (nonatomic, readonly) bool optional;
```
bdash: This should be `BOOL`, and should likely also have `getter=isOptional` to follow Cocoa's naming conventions.

Realm/RLMCollection.h
```diff
@@ -40,9 +41,21 @@ NS_ASSUME_NONNULL_BEGIN
...
- The class name (i.e. type) of the `RLMObject`s contained in the collection.
+ The type of the objects in the collection.
  */
-@property (nonatomic, readonly, copy) NSString *objectClassName;
+@property (nonatomic, readonly, assign) RLMPropertyType type;
+
+/**
+ Indicates whether the objects in the collection can be `nil`.
+ */
+@property (nonatomic, readonly) bool optional;
```
bdash: Ditto for `BOOL` / `getter=isOptional`.

Realm/RLMManagedArray.mm
```diff
@@ -353,34 +383,63 @@ - (id)valueForKey:(NSString *)key {
...
-    size_t column = _objectInfo->tableColumn(property);
-    auto value = translateErrors([&] { return _backingList.average(column); }, @"averageOfProperty");
+    size_t column = [self columnForProperty:property];
+    auto value = translateErrors(self, [&] { return _backingList.average(column); }, @"averageOfProperty");
     return value ? @(*value) : nil;
 }
 
 - (void)deleteObjectsFromRealm {
+    if (_type != RLMPropertyTypeObject) {
+        @throw RLMException(@"Cannot delete objects from RLMArray<%@>: only RLMObjects can be deleted.", RLMTypeToString(_type));
```
bdash: Clearing the array seems close enough in terms of semantics to what happens with arrays containing objects that it may be simpler to do that rather than throwing here?

Realm/RLMManagedArray.mm
```diff
@@ -401,16 +460,19 @@ - (RLMResults *)objectsWithPredicate:(NSPredicate *)predicate {
...
-    auto indexInTable = query.find();
-    if (indexInTable == realm::not_found) {
-        return NSNotFound;
+    realm::Query query;
+    if (_objectInfo) {
+        query = RLMPredicateToQuery(predicate, _objectInfo->rlmObjectSchema,
+                                    _realm.schema, _realm.group);
+    }
+    else {
+        // FIXME
```
bdash: Should we just thrown an exception for now?

Realm/RLMResults.h
```diff
@@ -64,6 +62,11 @@ typedef NS_ENUM(int32_t, RLMPropertyType);
 @property (nonatomic, readonly, assign) RLMPropertyType type;
 
 /**
+ Indicates whether the objects in the collection can be `nil`.
+ */
+@property (nonatomic, readonly) bool optional;
```
bdash: Ditto about `BOOL` / `getter=isOptional`.

Realm/RLMResults.mm
```diff
@@ -328,6 +336,7 @@ - (RLMResults *)objectsWithPredicate:(NSPredicate *)predicate {
         if (_results.get_mode() == Results::Mode::Empty) {
             return self;
         }
+        // FIXME: primitive array queries
```
bdash: Should we throw for now rather than silently giving incorrect results?

Realm/RLMManagedArray.mm
tgoyne: The public API that this is called by (`-[RLMRealm deleteObjects:]`) can't support deleting collections of non-objects (what would `[realm deleteObjects:@[@0]]` do?).

bdash:
One problem the tests reveal with Xcode 9 / Swift 4 is:
```
Simultaneous accesses to 0x6040000de518, but modification requires exclusive access.
Previous access (a modification) started at RealmSwift Tests`implicit closure #1 in PrimitiveListTests.testReplaceRange() + 209 (0x120c8fa21).
Current access (a read) started at:
0    libswiftCore.dylib                 0x00000001219418e0 swift_beginAccess + 605
1    RealmSwift Tests                   0x0000000120c77a40 PrimitiveListTestsBase.array.getter + 68
2    RealmSwift Tests                   0x0000000120c8e0c0 PrimitiveListTests.testReplaceRange() + 1114
3    RealmSwift Tests                   0x0000000120c8ff20 @objc PrimitiveListTests.testReplaceRange() + 36
```

This error occurs during the evaluation of:
```swift
array.replaceSubrange(0..<0, with: [values[0]])
```

I think Swift's new exclusive access checking isn't safe against Objective-C / C++ exceptions being thrown, so the earlier:
```swift
assertThrows(array.replaceSubrange(0..<1, with: []),
             reason: "Index 0 is out of bounds")
```
leaves the access checking state in an inconsistent state?

bdash:
All of the remaining CI failures are due to flakiness.

bdash:
The work remaining here appears to be:
 - [ ] Change log entry
 - [ ] Another review pass from all involved to determine if there's any other remaining work

Configuration/RealmSwift/RealmSwift.xcconfig
```diff
@@ -10,3 +10,5 @@ DEFINES_MODULE = YES;
 INFOPLIST_FILE = RealmSwift/RealmSwift-Info.plist;
 PRODUCT_NAME = RealmSwift;
 PRODUCT_BUNDLE_IDENTIFIER = io.realm.RealmSwit;
+
+OTHER_SWIFT_FLAGS = -D BUILDING_REALM_SWIFT;
```
tgoyne: Doesn't this need to be defined in the podspec as well?

If that's difficult then inverting the condition would avoid that.

bdash: I assume it would. I'm quit surprised that the CocoaPods jobs succeeded without it!

bdash: It succeeded because the Swift compiler only emits a warning about the redundant `import`. I went ahead and inverted the condition in e1b6c8d anyway.

bdash:
The failing Swift test has already been fixed on `master-3.0`.

tgoyne:
Obj-C and Swift tests now pass for me locally on both macOS and on my phone, so I think any remaining failures are just CI flakiness.

Issue #4760, opened 2017-03-22
List should have an overload of replaceSubrange(_:with:) that accepts its indices


bdash:
With an array you can replace its contents by doing:
```swift
a.replaceSubrange(a.indices, with: [1, 2, 3])
```

With `List<T>` you get an error because `List.Indices` is `DefaultRandomAccessIndices<List<T>>`, for which there's no overload. This should really work.

I took a brief stab at making it work but ran into a mysterious problem.  [`RangeReplaceableCollection`](https://developer.apple.com/reference/swift/rangereplaceablecollection) provides some overloads that take a `CountableRange<Int>`, so I tried changing `List.Indices` to be an alias for `CountableRange<Int>`, but for some reason the Swift compiler still won't use that overload. It appears to only see the overloads that take a `Range<Int>`.

bdash:
Note that this isn't really about the specific overload mentioned in the tile, but about the wider question as to why some of `RangeReplaceableCollection`'s overloads are not available on `List`. It may also apply to our other collection types.

austinzheng:
As part of #5276, I wrote a bunch of unit tests that, for pre-Swift 4.0 builds, just test the built-in functionality. One of these tests looks like:

```
func testReplaceOpenSubrange() {
        let subrange: Range<Int> = 2..<5
        let newElements = [10, 11, 12, 13, 14, 15, 16]
        list.replaceSubrange(subrange, with: makeSwiftIntObjects(from: newElements))
        array.replaceSubrange(subrange, with: newElements)
        compare(array: array, with: list)
}
```
This test compiles and passes successfully for all Swift versions from 3.0 to 3.2 (Xcode 8.0 and onward).

austinzheng:
Fixed by #5276. I talked with @bdash and we decided that we can handle any other methods if users ask for them.

PR #5276, opened 2017-08-30
Remove `List` conformance to `RangeReplaceableCollection`

### LINKED ISSUE #3691: Should List<T> keep its initializer that takes a sequence?
Due to `List<T>`'s conformance to `RangeReplaceableCollectionType` it picks up an initializer that takes a sequence:

``` swift
func init<S : SequenceType where S.Generator.Element == Generator.Element>(_: S)
```

This allows constructs of questionable value such as `List<Person>([personA, personB])`, and may set users up for failure by encouraging them to attempt to assign to `List<T>` properties.

We could:
1. Remove conformance to `RangeReplaceableCollectionType` while providing implementations of the remaining methods it declares.
2. Override the initializer to abort with a useful error.
3. Propose removing the initializer from the protocol on swift-evolution since it's not clear how it fits the described purpose of the protocol.
4. Do nothing.

Thoughts?

### LINKED ISSUE #5150: `List` should conform to `MutableCollection` rather than `RangeReplaceableCollection`?
`List` doesn't conform exactly to `RangeReplaceableCollection`'s semantics, and `MutableCollection` didn't exist when we first decided `List`'s conformance. Revisit this for 3.0.

austinzheng:
List is now a `MutableCollection`; replacements for default methods provided by RRC have been added

Not sure if this is the best approach; open to suggestions or other feedback.

Fixes #5150 
Fixes #3691

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,9 @@ x.x.x Release notes (yyyy-MM-dd)
   using a `RLMResults`. This results collection supports all normal collection
   operations except for setting values using KVO (since `RLMSyncPermission`s are
   immutable) and the property aggregation operations.
+* `List` now conforms to `MutableCollection` instead of `RangeReplaceableCollection`.
+  The empty collection initializer has been removed. Default implementations of
+  range replaceable collection methods that make sense for `List` have been added.
 
 ### Enhancements
 
--- a/RealmSwift/List.swift
+++ b/RealmSwift/List.swift
@@ -451,27 +451,13 @@ public final class List<T: RealmCollectionValue>: ListBase {
     }
 }
 
-extension List: RealmCollection, RangeReplaceableCollection {
+extension List: RealmCollection, MutableCollection {
     // MARK: Sequence Support
 
     /// Returns a `RLMIterator` that yields successive elements in the `List`.
     public func makeIterator() -> RLMIterator<T> {
         return RLMIterator(collection: _rlmArray)
     }
-
-    // MARK: RangeReplaceableCollection Support
-
-#if swift(>=3.1)
-    // These should not be necessary, but Swift 3.1's compiler fails to infer the `SubSequence`,
-    // and the standard library neglects to provide the default implementation of `subscript`
-    /// :nodoc:
-    public typealias SubSequence = RangeReplaceableRandomAccessSlice<List>
-
-    /// :nodoc:
-    public subscript(slice: Range<Int>) -> SubSequence {
-        return SubSequence(base: self, bounds: slice)
-    }
-#endif
 
     /**
      Replace the given `subRange` of elements with `newElements`.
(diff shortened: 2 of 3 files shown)


austinzheng:
Older versions of Swift are choking on this change and Xcode 8.0's Swift compiler is unusably terrible. Moving this back to 'in progress'.

austinzheng:
I talked to @bdash earlier and he suggested making this change only for projects using Swift 4 or later.

RealmSwift/List.swift
```diff
@@ -508,14 +494,168 @@ extension List: RealmCollection, RangeReplaceableCollection {
...
+            return
+        }
+        _rlmArray.removeObject(at: 0)
+    }
+
+    /**
+     Removes the specified number of objects from the beginning of the list. The
+     objects are not removed from the Realm that manages them.
+
+     This is a no-op if the list is already empty.
```
tgoyne: This comment is wrong.

RealmSwift/List.swift
```diff
@@ -508,14 +494,168 @@ extension List: RealmCollection, RangeReplaceableCollection {
...
+
+     - warning: This method may only be called during a write transaction.
+     */
+    public func removeFirst(_ number: Int) {
+        let count = Int(_rlmArray.count)
+        guard number <= count else {
+            throwRealmException("It is not possible to remove more objects from a list than it already contains.")
```
tgoyne: Error message should include the actual size and number of elements the user asked to have removed.

RealmSwift/List.swift
tgoyne: The current `removeLast()` implementation is a no-op on empty lists because that's what `NSArray` (and thus RLMArray) does, but maybe all of these should be aligned with RRC.

RealmSwift/List.swift
```diff
@@ -508,14 +494,168 @@ extension List: RealmCollection, RangeReplaceableCollection {
...
+    }
+
+    /**
+     Inserts the items in the given collection into the list at the given position.
+
+     - warning: This method may only be called during a write transaction.
+     */
+    public func insert<C: Collection>(contentsOf newElements: C, at i: Int)
+        where C.Iterator.Element == Element {
+            for item in newElements.reversed() {
```
tgoyne: `for item in newElements { insert(item, at: i); i += 1 }` is asymptotically faster, as inserting them in reverse order at `i` requires shifting all of the already-inserted values each time.

RealmSwift/Tests/ListTests.swift
```diff
@@ -643,3 +643,103 @@ class ListRetrievedTests: ListTests {
         return array
     }
 }
+
+#if swift(>=4.0)
```
tgoyne: Any reason not to run these tests on older versions of Swift? Even if they're not actually exercising our code, they should work and would serve as a sanity check.

austinzheng: Yeah, running them for all versions makes sense.

CHANGELOG.md
```diff
@@ -20,6 +20,15 @@ x.x.x Release notes (yyyy-MM-dd)
   `LinkingObjects` have had their generic type parameter changed from `T` to
   `Element`).
 * `RealmOptional`'s generic type parameter has been changed from `T` to `Value`.
+* `List` now conforms to `MutableCollection` instead of `RangeReplaceableCollection`.
```
austinzheng: Not sure how this happened. I'll remove the duplicate entry.

bdash:
I wonder if this happens to address #4760 as a side-effect?

austinzheng:
I believe it does, since there is one test (the only test still guarded by `#if`) that only works with the new code I wrote, and not with `RRC`.

RealmSwift/List.swift
```diff
@@ -509,14 +484,165 @@ extension List: RealmCollection, RangeReplaceableCollection {
     /// :nodoc:
     public func _observe(_ block: @escaping (RealmCollectionChange<AnyRealmCollection<Element>>) -> Void) ->
         NotificationToken {
-        let anyCollection = AnyRealmCollection(self)
-        return _rlmArray.addNotificationBlock { _, change, error in
-            block(RealmCollectionChange.fromObjc(value: anyCollection, change: change, error: error))
+            let anyCollection = AnyRealmCollection(self)
```
bdash: The indentation here looks a bit odd. The function body appears to be indented one level too far.

RealmSwift/List.swift
```diff
@@ -509,14 +484,165 @@ extension List: RealmCollection, RangeReplaceableCollection {
...
+    /**
+     Removes the specified number of objects from the beginning of the list. The
+     objects are not removed from the Realm that manages them.
+
+     - warning: This method may only be called during a write transaction.
+     */
+    public func removeFirst(_ number: Int = 1) {
+        let count = Int(_rlmArray.count)
+        guard number <= count else {
+            throwRealmException("It is not possible to remove more objects (\(number)) from a list"
```
bdash: Any idea why we throw an exception in cases like this rather than just using `fatalError`? The Swift standard library appears to use `fatalError` for these cases, and it's not like a Swift developer can catch an exception thrown via Obj-C like this.

RealmSwift/List.swift
```diff
@@ -509,14 +484,165 @@ extension List: RealmCollection, RangeReplaceableCollection {
...
+
+    /**
+     Inserts the items in the given collection into the list at the given position.
+
+     - warning: This method may only be called during a write transaction.
+     */
+    public func insert<C: Collection>(contentsOf newElements: C, at i: Int)
+        where C.Iterator.Element == Element {
+            var i = i
+            for item in newElements {
```
bdash: Why not:
```swift
for (i, item) in newElements.enumerated() {
    insert(item, at: i)
}
```

The indentation in this function appears to be one level too deep.

RealmSwift/List.swift
tgoyne: We use `throwRealmException` for fatal errors so that we can test them.

bdash: Hah, that makes a heck of a lot of sense!

RealmSwift/List.swift
austinzheng: You're right, I've forgotten how to program in Swift.

As for the indentation, Xcode starts the body of the function +1 indent level in from the indent level of the line with the brace, but there's no reason the `where` clause should be on its own line anyways, so I'll fix that.

austinzheng: Actually, I believe your code will start `i` at the start index of `newElements`, whereas `i` (as passed into the method) can actually be any arbitrary starting index.

bdash: Hmm, yeah. You could change the loop variable to `j` and use `i + j` for the `at:`. Dunno if it's worth it though.

austinzheng:
Same simulator shenanigans, same flaky tvOS notification tests.

PR #5283, opened 2017-09-04
Exploring a partial sync API

bigfish24:
I haven't been able to get too far with an actual implementation of the partial sync API, but started exploring how it could hook into the Cocoa API. Currently I see the API broken into three pieces:

1. SyncConfiguration support to configure the Realm to be a partially synced Realm - implemented as just an additional boolean
2. Registration of a remote query - I explored how this could hook into the existing query and notification system, however, I am starting to doubt this approach (details below)
3. De-registration and querying of registered remote queries - explored this on the SyncManager

Thoughts on registration:

My initial exploration on this PR was to tap into the existing Cocoa functionality where you can perform a local query and add a notification block which would be the registration of a remote query:
```objc
RLMNotificationToken *token = [[Person objectsWhere:@"age > 5"] addNotificationBlock:^(RLMResults<Person *> *results, RLMCollectionChange *changes, NSError *error) {
    // Handle results
}];
```
```swift
let token = Realm().objects(Person.self).filter("age > 5").addNotificationBlock { [weak self] (changes: RealmCollectionChange) in
    // Handle results
}
```
However, while this is nice in that it reuses existing patterns, it has the following problems:

1. Without `addNotificationBlock` the local query returns nothing and its not obvious why adding a notification block would register a remote query
2. The `token` is confusing in this sense because it controls the lifetime of the callback but not the lifetime of the remote query registration
3. This reusing of existing APIs won't work with Java and .Net because they don't even have a string-based query API.

Creating this PR to track a discussion.

### DIFF
--- a/Realm.xcodeproj/project.pbxproj
+++ b/Realm.xcodeproj/project.pbxproj
@@ -455,6 +455,10 @@
 		E8F992BE1F1401C100F634B5 /* RLMObjectBase_Private.h in Headers */ = {isa = PBXBuildFile; fileRef = E8F992BD1F1401C100F634B5 /* RLMObjectBase_Private.h */; settings = {ATTRIBUTES = (Private, ); }; };
 		E8F992BF1F1401C100F634B5 /* RLMObjectBase_Private.h in Headers */ = {isa = PBXBuildFile; fileRef = E8F992BD1F1401C100F634B5 /* RLMObjectBase_Private.h */; settings = {ATTRIBUTES = (Private, ); }; };
 		E8FD2E381D93345100569F10 /* keychain_helper.cpp in Sources */ = {isa = PBXBuildFile; fileRef = 1A64CA891D8763B400BC0F9B /* keychain_helper.cpp */; };
+		F98AC6A51F576F6500E811A5 /* RLMResults+RLMResults_Sync.h in Headers */ = {isa = PBXBuildFile; fileRef = F98AC6A31F576F6500E811A5 /* RLMResults+RLMResults_Sync.h */; };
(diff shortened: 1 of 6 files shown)


bigfish24:
Javascript already uses the string-based API so:
```js
let token = realm.objects('Person').filtered("age > 5").addListener((people, changes) => {
    // Handle results
})
```
For Java and .Net they will need to add an API to pass a string for the query. For .Net this seems query straight forward with the syntax:
```C#
realm.All<Person>().Where(age > 5).SubscribeForNotifications ((people, changes, error) => 
{
    // Handle results
});
```
For Java it doesn't look great. Their model for adding notifications doesn't match the other 3 bindings, where you register the change listener on the Realm itself `realm.addChangeListener(realmListener);`. Instead, it seems we should move towards unifying the functionality with the other SDKs. For example:
```java
private final OrderedRealmCollectionChangeListener<RealmResults<Person>> changeListener = new OrderedRealmCollectionChangeListener<>() {
    @Override
    public void onChange(RealmResults<Person> results, OrderedCollectionChangeSet changeSet) {
        // Handle results
    }
}
// Add another variety of `where` which accepts a class and a string query
realm.where(Person.class, "age > 5").addChangeListener(changeListener);
```

tgoyne:
The fact that our queries (can) happen to involve strings is somewhat of a red herring. Creating a server query out of a RLMResults would actually have to be done in the same way as creating a server query from Java's Query objects (i.e. building a new string from scratch out of the core Query objects). We don't have a string-based query interface in the sense that SQL does where the string is the actual interface to the query engine.

austinzheng:
I think @mbalex99 had some thoughts about potential APIs as well.

bdash:
Please ensure any API discussion that's not Cocoa-specific is had in the product issue referenced above for sake of visibility to the other bindings.

PR #5285, opened 2017-09-06
Add type-safe query with Swift 4 smart KeyPath

kishikawakatsumi:
Add a new query method that can be written using the Swift 4 smart KeyPath. This makes it possible to make a compile error before executing runtime errors due to typo or type mismatches.

- Add `Results.filter (_ :)` method that can be written like `\Person.name ==" John"` using Swift 4 KeyPath
It prevents typo and type mismatch.

```
realm.objects(Person.self).filter { \Person.name == "John" }
realm.objects(Dog.self).filter { \Dog.age < 2 }
realm.objects(Dog.self).filter { \Dog.name < 2 } // Compile error
realm.objects(Dog.self).filter { \Person.age < 2 } // Compile error

realm.objects(Dog.self).filter { \Dog.age < 2 && \Dog.owner.name == "John" }
realm.objects(Dog.self).filter { \Dog.age < 2 && \Person.name == "John" } // Compile error
```

- Add operator overloads to achieve notation like `\ Person.name ==" John "`.
  The operators receive `KeyPath` on the left side hand

NOTE: Currently it can not be used for `RealmOptional <T>` properties.
Because we cannot retrieve a string representation from non-ObjC KeyPaths.

- Add `Results.sorted(byKeyPath:ascending:)` method that receives Swift 4 KeyPath.

CC @tgoyne @bdash @austinzheng

### DIFF
--- a/RealmSwift/Results.swift
+++ b/RealmSwift/Results.swift
@@ -422,6 +422,229 @@ extension Results: AssistedObjectiveCBridgeable {
     }
 }
 
+#if swift(>=4)
+// MARK: Typesafe Queries
+public protocol RealmPropertyType {
+    var _object: AnyObject { get }
+}
+
+public protocol RealmEquatablePropertyType: RealmPropertyType {}
+public protocol RealmComparablePropertyType: RealmEquatablePropertyType {}
+
+extension Results {
+    public func filter<P: Predicate>(_ expression: @autoclosure () -> P) -> Results<T> where P.ObjectType == T {
+        return filter(expression().predicate)
+    }
+
+    public func sorted<RealmObject: Object, RealmProperty: RealmPropertyType>(byKeyPath keyPath: KeyPath<RealmObject, RealmProperty>, ascending: Bool = true) -> Results<T> {
+        return sorted(by: [SortDescriptor(keyPath: keyPath._kvcKeyPathString!, ascending: ascending)])
+    }
+
+    public func sorted<RealmObject: Object, RealmProperty: RealmPropertyType>(byKeyPath keyPath: KeyPath<RealmObject, RealmProperty?>, ascending: Bool = true) -> Results<T> {
+        return sorted(by: [SortDescriptor(keyPath: keyPath._kvcKeyPathString!, ascending: ascending)])
+    }
+}
+
+extension Bool: RealmEquatablePropertyType {
+    public var _object: AnyObject { return NSNumber(value: self) }
+}
+extension Int: RealmComparablePropertyType {
+    public var _object: AnyObject { return NSNumber(value: self) }
+}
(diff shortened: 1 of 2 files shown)


RealmSwift/Results.swift
```diff
@@ -422,6 +422,229 @@ extension Results: AssistedObjectiveCBridgeable {
...
+public protocol RealmEquatablePropertyType: RealmPropertyType {}
+public protocol RealmComparablePropertyType: RealmEquatablePropertyType {}
+
+extension Results {
+    public func filter<P: Predicate>(_ expression: @autoclosure () -> P) -> Results<T> where P.ObjectType == T {
+        return filter(expression().predicate)
+    }
+
+    public func sorted<RealmObject: Object, RealmProperty: RealmPropertyType>(byKeyPath keyPath: KeyPath<RealmObject, RealmProperty>, ascending: Bool = true) -> Results<T> {
+        return sorted(by: [SortDescriptor(keyPath: keyPath._kvcKeyPathString!, ascending: ascending)])
```
austinzheng: Is `_kvcKeyPathString` a private Swift standard library API? We may not be able to ship this PR unless it becomes part of the public API.

kishikawakatsumi: Yes, `_kvcKeyPathString`  is undocumented yet. https://github.com/apple/swift/blob/master/stdlib/public/core/KeyPath.swift#L99-L105

Apple engineers realized the needs. I asked it at the WWDC lab. I made a feature request https://bugs.swift.org/browse/SR-5220. 

Probably would it be worth to provide as a separate microframework rather than the Realm API? Like RxRealm https://github.com/RxSwiftCommunity/RxRealm

bdash: Yes, I think keeping this separate until the limitations and use of private API are addressed is the right approach.

kishikawakatsumi:
@bdash @austinzheng I created a prototype of the framework. https://github.com/kishikawakatsumi/RealmSafeQuery

Should this repository be in the Realm organization? Is there another good name? Any other feedbacks are welcome 🙏

bdash:
I think it's fine to live on your GitHub account for now. My only suggestion for naming would be to change "Safe" for "TypeSafe" to make it more explicit where the safety is.

bdash:
@kishikawakatsumi, is there any value in keeping this PR around at this point?

kishikawakatsumi:
@bdash This PR can be closed. RealmTypeSafeQuery released as a separated framework.

iDevid:
Actually can it be done trough NSExpression?
```
extension KeyPath where Root: NSObject {
    var string: String {
        return NSExpression(forKeyPath: self).keyPath
    }
}
```
It works only for @objc dynamic properties, and NSObject

PR #5290, opened 2017-09-08
Remove Swift language version symlinks from the examples

bdash:
Since all of the examples share the same code across Swift versions, the symlinks don't buy us anything. Now package_examples.rb knows which Swift versions are supported, and creates a copy of each example of each Swift version before updating its framework search paths.

This is intended to make it slightly easier to package the Swift 3.2 and 4.0 examples.

The changes in this PR consist of:
1. Rename the `swift-3.0` examples to `swift`.
2. Remove the `swift-x.y` symlinks.
3. Update `package_examples.rb`.

Only the changes to `package_examples.rb` are really worth reviewing.

The deletions shown in this PR, with exception of the symlink removals, are due to some of the examples having been copied for Swift 3.1 rather than being symlinked. There were no relevant code changes between the Swift 3.0 and 3.1 copies though.


austinzheng:
The `ios-swift Debug 8.3.3` job is failing on an error `xcodebuild: error: 'examples/ios/swift-3.1/RealmExamples.xcworkspace' does not exist.`. Not sure if that's why the other jobs are failing.

bdash:
Yeah, that's somewhat expected. I put this PR up primarily to see what fallout there would be from it.

PR #5291, opened 2017-09-08
Add Xcode 9 to CI's test matrix

bdash:
For now I've just had it copy the Xcode 8.3.3 configuration. We can likely remove some of the duplication after confirming that things work with Xcode 9.


bdash:
None of the CI failures appear to be related to what this PR is doing.

PR #5292, opened 2017-09-09
Build and package for Swift 3.2 and 4.0.

bdash:
This currently builds only Swift 4.0 and symlinks Swift 3.2 to it as https://swift.org/migration-guide-swift4/ suggests that a module built with Swift 4.0 should be compatible with Swift 3.2. We'll need to double-check this.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -1210,7 +1210,7 @@ EOM
 
     "package-ios-swift")
         cd tightdb_objc
-        for version in 8.0 8.1 8.2 8.3.3; do
+        for version in 8.0 8.1 8.2 8.3.3 9.0; do
             REALM_XCODE_VERSION=$version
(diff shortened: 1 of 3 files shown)


bdash:
This also fixes an issue with the releasability pipeline that I introduced in an earlier PR.

bdash:
Frameworks built using Swift 4.0 do appear to be usable with Swift 3.2. I took the packaged iOS frameworks from a run of the releasability pipeline on this job (https://ci.realm.io/job/Package%20iOS%20Swift/1325/), and put them into a trivial app that was set to use Swift 3.2. I was able to build and run the app successfully. After changing the Swift language version to 4.0 I was able to rebuild and run the app successfully once more.

PR #5299, opened 2017-09-11
Make changes for testing against ROS 2

### LINKED ISSUE #5296: Update object server tests to work with ROS v2.0
ROS v2.0 has different packaging and command-line interface than ROS v1.0. We need to update our test harness to accommodate the differences in order to continue to test against current versions of sync.

Work that still remains to be done:
- [ ] Enable `testReadAccess` test. _Fixed by https://github.com/realm/ros/pull/475_
- [x] Enable `testWriteAccessViaUsername` test. ~_Fix forthcoming in next ROS per https://github.com/realm/ros/issues/426_~ Filed https://github.com/realm/ros/issues/508.
- [x] Enable `testDeletingPermission` test. This test passes in isolation but fails when run in the middle of the entire suite. It's possible that https://github.com/realm/ros/issues/437 is the cause.
- [ ] Enable `testDeleteRealmUponPermissionDenied` test. _Fixed by https://github.com/realm/ros/pull/475_ 
- [x] All other permissions tests need to be re-enabled and ported over.
- [x] Fix how admin users are created; re-enable tests that depend on admin users (`testRetrieveUserInfo`) - _#5341_
- [x] Fix and re-enable the password change tests (`testUserChangePassword`, `testOtherUserChangePassword`) -- _#5341_
- [x] Fix and re-enable the test that opens a Realm with the admin token (`testOpenRealmWithAdminToken`) -- _https://github.com/realm/realm-cocoa/pull/5332_
- [x] Fix and re-enable the test that logs in a user based on the admin token (`testAdminTokenAuthentication`) -- _https://github.com/realm/realm-cocoa/pull/5332_
- [x] Fix the sync error handler test (`testSyncErrorHandlerErrorDomain`) -- _#5333_

austinzheng:
Changes:
- Workaround for an issue where Swift 4's type inference is broken
- ROS integration tests now download ROS 2 using NPM
- ROS test server state is now stored in a single directory: `test-ros-instance`
- Integration tests now die immediately if ROS did not start up successfully
- Integration tests now detect ROS failure to start up because of port in use issues and abort if so

Fixes #5296 

Current status: everything in build.sh is working. Tests can be started and some of them run successfully.

To do:
- [x] Figure out why multiprocess tests aren't working
- [x] Check permissions tests; if ROS 2 doesn't support their functionality disable them
- [x] New build of core/sync that doesn't cause bitcode errors when building for watchOS
- [ ] (Stretch goal) don't re-download ROS if the correct version already exists; just clear server state

Things that require a fix on ROS:
- [ ] Permissions
- [ ] Document error codes
- [ ] Fix "get user info" API to return all the things the ROS 1 API did
- [ ] Make change password API take the correct arguments
- [ ] Allow creation of admin privilege users


Realm/ObjectServerTests/RLMSyncTestCase.mm
```diff
@@ -379,27 +379,45 @@ - (void)lazilyInitializeObjectServer {
...
+    __block BOOL inUseError = NO;
     [RLMSyncTestCase runResetObjectServer:YES];
     NSTask *task = [[NSTask alloc] init];
-    task.currentDirectoryPath = [[RLMSyncTestCase rootRealmCocoaURL] path];
+    NSString *currentDirPath = [[[RLMSyncTestCase rootRealmCocoaURL]
+                                 URLByAppendingPathComponent:@"test-ros-instance/"] path];
+    task.currentDirectoryPath = currentDirPath;
     task.launchPath = @"/bin/sh";
-    task.arguments = @[@"build.sh", @"start-object-server"];
+    task.arguments = @[@"-c", @"/usr/local/bin/node ./ros/bin/ros start"];
```
bdash: Is it necessary to invoke `sh` here? I'd think invoking `/usr/local/bin/node ros/bin/ros start` would be sufficient.

RealmSwift/RealmCollection.swift
```diff
@@ -138,6 +138,7 @@ public protocol RealmCollectionValue {
 #if swift(>=3.2)
 /// :nodoc:
 public protocol RealmCollectionBase: RandomAccessCollection, LazyCollectionProtocol, CustomStringConvertible, ThreadConfined where Element: RealmCollectionValue {
+    associatedtype ElementType: RealmCollectionValue
```
bdash: What problem is this fixing? Is there a test that catches it? Can we split it into its own PR since it seems unrelated to this issue?

build.sh
```diff
@@ -441,15 +433,6 @@ case "$COMMAND" in
...
         sleep 1
-        package="${source_root}/sync"
-        for file in "$package"/realm-object-server-*; do
-            if [ -d "$file" ]; then
-                package="$file"
-                break
-            fi
-        done
-        rm -rf "$package/object-server/root_dir/"
-        rm -rf "$package/object-server/temp_dir/"
```
bdash: Don't we still need to nuke the server's on-disk state?

dependencies.list
```diff
@@ -1,4 +1,4 @@
 VERSION=3.0.0-beta.3
 REALM_CORE_VERSION=3.0.0-rc5
 REALM_SYNC_VERSION=2.0.0-rc15
-REALM_OBJECT_SERVER_VERSION=2.0.0-rc2
+REALM_OBJECT_SERVER_VERSION=2.0.0-alpha.25
```
bdash: Doesn't ROS v2.0.0-alpha.25 require that we use sync v2.0.0-rc18?

build.sh
```diff
@@ -302,20 +304,16 @@ fi
...
-    /usr/bin/curl -L -O "https://static.realm.io/downloads/object-server/$archive_name"
-    rm -rf sync
-    mkdir sync
-    tar xf $archive_name -C sync
-    rm $archive_name
-    cp Configuration/object-server-config.yml sync/object-server/configuration.yml
-    touch "sync/object-server/do_not_open_browser"
+    rm -rf ./test-ros-instance
+    mkdir ./test-ros-instance
+    mkdir ./test-ros-instance/ros
```
bdash: The `./` are unnecessary here.

The two `mkdir` invocations can become `mkdir -p test-ros-instance/ros`.

build.sh
```diff
@@ -302,20 +304,16 @@ fi
...
+    npm install --prefix ./test-ros-instance/ros -g realm-object-server@${REALM_OBJECT_SERVER_VERSION} -S
```
bdash: Is `-S` necessary?

RealmSwift/RealmCollection.swift
tgoyne: The whole point of RealmCollectionBase was to avoid having this associatedtype, as the compiler warned that it was redundant. If that's no longer the case then RCB should be removed entirely.

build.sh
austinzheng: I think doing so would be redundant, since we call this method before any tests are run but immediately after the latest object server is downloaded and installed, which involves deleting any existing server state.

bdash: Isn't it invoked by the first object server test case that runs? If you run the tests twice in a row you'd expect this to clear the state between the two runs, but it no longer does. The download / installation isn't repeated the second time you run the tests.

dependencies.list
austinzheng: Perhaps. I can't seem to find the correct branch in the ROS repo. I'll check again when I get home.

build.sh
austinzheng: When you say twice in a row, do you mean running `sh build.sh verify-osx-object-server`, waiting until it completes, then running it again? If so, `download-object-server` will be invoked by the first test that runs each time.

That being said, I think we should clear the state anyways, since it's surprising that a method called `reset-object-server` doesn't clear the state.

bdash: I'm talking about running tests directly within Xcode.

austinzheng: Oh, okay. In that case I should probably add some affordances to running the tests without using build.sh, because it's never worked quite right (e.g. you have to manually download the ROS).

build.sh
austinzheng: I got the invocation for ROS from Kraen, but I can check with him and figure out if `-S` makes sense for our use case.

build.sh
bdash: Having to run `sh build.sh download-object-server` once before running tests in Xcode isn't that much of a burden. There are more important things to worry about :-)

RealmSwift/RealmCollection.swift
austinzheng: Yes. I will investigate this further and break out any needed changes into a separate PR.

dependencies.list
austinzheng: rc18 doesn't seem to be published; the build scripts can't find it.

bdash: I think they may have fixed the file extension of the sync archive back to being .xz like it should have been. Try tweaking the extension here: https://github.com/realm/realm-cocoa/blob/master-3.0/build.sh#L329

Realm/ObjectServerTests/SwiftPermissionsAPITests.swift
```diff
@@ -204,18 +204,18 @@ class SwiftPermissionsAPITests: SwiftSyncTestCase {
             XCTAssertNil(error)
             ex2.fulfill()
         }
-        waitForExpectations(timeout: 2.0, handler: nil)
+        waitForExpectations(timeout: 2000.0, handler: nil)
```
bdash: Why is this change necessary?

austinzheng: I accidentally left it in. The next commit I push will change it back to 2 seconds.

bdash:
The watchOS failures look like so:
> `error: Invalid bitcode version (Producer: '802.0.42.0_0' Reader: '800.0.38_0')`

This is because the most recent core and sync releases appear to have been built with Xcode 8.3.3 rather than Xcode 8.2 like they need to be in order to generate bitcode that's compatible with Xcode 8.2 and earlier. See realm/realm-core#2856 / realm/realm-sync#1648.

austinzheng:
It doesn't look like 35 minutes is enough for the `ios-static` tests to complete, since they were killed by CI in the middle of doing work.

bdash:
I don't think they usually take anywhere near that long. ~15 minutes is more typical. I'm guessing it may have just been a CI hiccup.

PR #5304, opened 2017-09-12
Fix Swift 3.2/4.0 compilation failure

austinzheng:
Fix a compilation failure when building Realm Swift for Swift 3.2/4.0.

I think this is the least bad option. Other things I tried include:

- Getting rid of `RealmCollectionBase`, since >=3.2 doesn't require it. Unfortunately, because of the way Swift's `#if` works (i.e. not like the CPP), and because the syntax for associated types changed between 3.1 and 3.2, the entire protocol would have to be declared twice, once for 3.1 and once for >=3.2.
- Making Swift 3.1 use `Element` instead of `ElementType` in all the declarations in the protocol and classes that conform to it. Unfortunately bugs in Swift 3.1's compiler make it fail to associate the generic types on the conforming types with the associated type with the same name (`Element`, from the parent `Collection` requirement), requiring them to have different names and a typealias to link them.

Ideas are welcome.

### DIFF
--- a/RealmSwift/RealmCollection.swift
+++ b/RealmSwift/RealmCollection.swift
@@ -136,6 +136,9 @@ public protocol RealmCollectionValue {
 }
 
 #if swift(>=3.2)
+// FIXME: When we drop support for Swift 3.1, change ElementType to Element
+// throughout the project (this is a non-breaking change). We use ElementType
+// only because of limitations in Swift 3.1's compiler.
 /// :nodoc:
 public protocol RealmCollectionBase: RandomAccessCollection, LazyCollectionProtocol, CustomStringConvertible, ThreadConfined where Element: RealmCollectionValue {
     typealias ElementType = Element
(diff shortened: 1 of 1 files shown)


austinzheng:
The build was failing before because the Swift >=3.2 branch of the source doesn't define any type or type parameter named `ElementType`. This build defines that type (by making it a typealias of `Element`) so that Realm Swift compiles. Does that help?

tgoyne:
Isn't the problem just that 54bc3f9d0 missed a spot when renaming `Element` to `ElementType`? The protocol definition should be `public protocol RealmCollectionBase: RandomAccessCollection, LazyCollectionProtocol, CustomStringConvertible, ThreadConfined where ElementType: RealmCollectionValue`

austinzheng:
@tgoyne yeah, that's the correct way to fix this.

austinzheng:
Actually, it turns out the `where` clause is constraining an existing (inherited) associated type `Element`, and the typealias is required after all :|. (I hate the protocol `where` syntax.)

tgoyne:
Okay, I think I understand this now. We want to just be using `Element` everywhere, but that hits bugs(?) in Swift 3.1 that we're sidestepping by using `ElementType` on the protocol, which then requires a typealias back to `Element`. If that's all correct then this seems fine other than that it needs a comment explaining that so that we can clean it up when we drop Xcode 8 support.

austinzheng:
Yes, precisely. I'll add the comment.

PR #5306, opened 2017-09-13
Fix the Cocoa releasability CI job

bdash:
Introducing Xcode 9 / Swift 3.2 / Swift 4.0 support into our release packaging caused issues with CI. We appear to be hitting timeouts due to building all of the Swift versions for iOS taking over 40 minutes. In order to address this I've split the various Swift build steps so they can run in parallel. I've also taken this opportunity to restructure the CI job using a Jenkins pipeline using Groovy.

Remaining work:
- [x] Installation examples

### DIFF
--- a/.swiftlint.yml
+++ b/.swiftlint.yml
@@ -24,3 +24,4 @@ disabled_rules:
   - line_length
   - vertical_whitespace
   - syntactic_sugar
+  - block_based_kvo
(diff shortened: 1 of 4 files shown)


bdash:
This is now ready for review.

bdash:
The assumption is that whatever commit is being built has already been tested via the CI jobs run on our PRs. For the most part the tests that are run are ones that test release packaging. There's a few exceptions designed to prevent obviously broken builds from being released (the macOS and iOS static tests).

PR #5315, opened 2017-09-15
Offline client reset

bdash:
Expose the incompatible synced Realm exception that's raised when opening a synced Realm that was created using sync 1.x, and provide a means for accessing the data within the old Realm file.

Remaining work:
- [x] Swift support
- [x] Improved testing
- [x] Change log entry


bdash:
This is now ready for review.

austinzheng:
It looks fine, and since the tests pass it must work. But isn't there an object store flag that needs to be set on the configuration to open a Realm file with sync history in non-sync mode?

tgoyne:
`configuration.readOnly` is that flag (opening in immutable mode doesn't look at the history in any way).

tgoyne:
I guess `schemaVersionAtURL:` also needs to use the new translation function. Hopefully no one is using it on synced Realms (and I'm not sure there's any reason to use it at all now that we have built-in support for deleting realms when a migration is needed) but probably should be safe? Looks fine otherwise.

bdash:
`+schemaVersionAtURL:` can't meaningfully work with synced Realms since it doesn't take a sync configuration, and so would always open a synced Realm with an incorrect history type. That means it can never hit this offline client reset error. That said, it's currently doing an awful job of translating any exceptions raised into Cocoa errors, so there's value in updating it.

PR #5326, opened 2017-09-20
Add support for partial sync

austinzheng:
TODO:
- [x] Swift support
- [x] Rebase this onto something a little more sensible

### DIFF
--- a/Realm/RLMRealm+Sync.h
+++ b/Realm/RLMRealm+Sync.h
@@ -29,7 +29,7 @@ typedef void(^RLMPartialSyncFetchCallback)(RLMResults * _Nullable results, NSErr
 
 NS_ASSUME_NONNULL_BEGIN
 
-/// :nodoc:
+///
 @interface RLMRealm (Sync)
 
 /**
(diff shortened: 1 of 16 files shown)


Realm/RLMRealm+Sync.mm
```diff
@@ -0,0 +1,53 @@
...
+#import "results.hpp"
+#import "sync/partial_sync.hpp"
+#import "shared_realm.hpp"
+
+using namespace realm;
+
+@implementation RLMRealm (Sync)
+
+- (void)fetchResultsForQuery:(NSString *)query objectType:(Class)type callback:(RLMPartialSyncFetchCallback)callback {
+    NSString *className = NSStringFromClass(type);
```
austinzheng: Needs to check if the Realm is actually a partially synced Realm, and throw/return early if not.

bdash: I'd suggest pushing that checking down to object store if possible.

Realm/RLMRealm+Sync.h
```diff
@@ -0,0 +1,49 @@
...
+@class RLMResults;
+
+/**
+ A callback used to vend the results of a partial sync fetch.
+ */
+typedef void(^RLMPartialSyncFetchCallback)(RLMResults * _Nullable results, NSError * _Nullable error);
+
+NS_ASSUME_NONNULL_BEGIN
+
+///
```
austinzheng: Should this be blank, or a `:nodoc:`?

bdash: Blank. `:nodoc:` hides everything in the category from the documentation, which isn't what we want.

bdash:
CI has been having issues this evening, and I _think_ that's all the object server build issues are (they're timing out before the tests even run).

PR #5332, opened 2017-09-22
Update ROS testing dependency, reenable tests

austinzheng:
Changes:
- Updated ROS to alpha-36
- Fixed admin token parsing and re-enabled admin token tests
- Suppressed some spurious warnings for the object store test target
- Don't double-quote paths containing `~` in build.sh

I stole some of @bdash's fixes from #5326.


Realm/ObjectServerTests/RLMObjectServerTests.mm
```diff
@@ -454,7 +452,10 @@ - (void)disabled_testRetrieveUserInfo {
 /// The login queue argument should be respected.
 - (void)testLoginQueueForSuccessfulLogin {
     // Make global queue
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Wunguarded-availability"
```
bdash: Why not switch to using `DISPATCH_QUEUE_PRIORITY_DEFAULT` instead of a constant that isn't available on some versions of macOS that this test target builds for?

austinzheng: I think they're all gated by the same availability check. Let me try it...

bdash: I don't think so… the QoS stuff is newer than dispatch itself.

austinzheng: The `QOS` constants are new, the `DISPATCH_QUEUE_PRIORITY` ones aren't.

build.sh
```diff
@@ -439,9 +439,9 @@ case "$COMMAND" in
         ;;
 
     "reset-ros-client-state")
-        rm -rf "~/Library/Application Support/xctest"
-        rm -rf "~/Library/Application Support/io.realm.TestHost"
-        rm -rf "~/Library/Application Support/xctest-child"
+        rm -rf ~/Library/Application Support/xctest
```
tgoyne: This deletes the files `~/Library/Application` and `Support/xctest`, and -f makes it not complain about that neither of these exist. The space needs to be escaped.

build.sh
```diff
@@ -315,7 +315,7 @@ fi
 
 kill_object_server() {
 # Based on build.sh conventions we always run ROS from a path ending in 'ros/bin/ros'.
-    pid=$(ps ax | grep "ros/bin/ros start$" | sed -e 's/^[[:space:]]*//' | awk '{ print $1 }')
+    pid=$(ps ax | grep "[r]os/bin/ros start" | sed -e 's/^[[:space:]]*//' | awk '{ print $1 }')
```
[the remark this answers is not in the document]
bdash: It prevents the `grep` from matching itself in the `ps` output.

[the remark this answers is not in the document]
tgoyne: Since we don't need to support non-BSD platforms this can just be `pkill -f ros/bin/ros start`.

build.sh
austinzheng: Thank you for catching this!

austinzheng:
I did `sh build.sh build` locally and checked the Realm, Realm Swift, and object server test suites in Xcode.

PR #5341, opened 2017-09-26
Reenable more tests

austinzheng:
Changes:
- Fixed the way the change password request is made
- Updated ROS dependency to pull in bugfixes
- All non-permissions ROS integration tests are now enabled
- Changed the way tests create admin users
- Added a way for tests to be easily run without starting up their own ROS

Blocked on a ROS release that includes https://github.com/realm/ros/pull/384 (so, 39 or later)

### DIFF
--- a/Realm/RLMSyncUser.mm
+++ b/Realm/RLMSyncUser.mm
@@ -317,7 +317,8 @@ PermissionChangeCallback RLMWrapPermissionStatusCallback(RLMPermissionStatusBloc
                                      server:self.authenticationServer
                                        JSON:@{kRLMSyncTokenKey: self._refreshToken,
                                               kRLMSyncUserIDKey: userID,
-                                              kRLMSyncPasswordKey: newPassword}
+                                              kRLMSyncDataKey: @{ kRLMSyncNewPasswordKey: newPassword }
+                                              }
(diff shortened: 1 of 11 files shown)


austinzheng:
Cancelling CI because there's no point running it before everything is ready.

austinzheng:
These changes work with alpha-39 and this PR can be reviewed.

Realm/RLMSyncUser.mm
```diff
@@ -317,7 +317,8 @@ - (void)changePassword:(NSString *)newPassword forUserID:(NSString *)userID comp
                                      server:self.authenticationServer
                                        JSON:@{kRLMSyncTokenKey: self._refreshToken,
                                               kRLMSyncUserIDKey: userID,
-                                              kRLMSyncPasswordKey: newPassword}
+                                              kRLMSyncDataKey: @{ @"new_password": newPassword }
```
bdash: Should `"new_password"` be pulled out into a constant like all of the other keys?

austinzheng:
Before this can be merged the diagnostic commits **MUST BE REVERTED**.

austinzheng:
Object server tests passed. Looks like we'll have to renew the device certs on Monday.

PR #5352, opened 2017-09-28
Fix for Swift 4.0 String slicing subscript warning

fawadsuhail:
Using Swift 4's string subscript method to get substring out to remove warning on Xcode 9.

xcode 9 warning


bdash:
The change as-is will breaking using Realm Swift with previous versions of Xcode. This'll need to be done in a manner compatible with older versions of Swift.

fawadsuhail:
@bdash Right you are! Added a ```#if swift``` version check

austinzheng:
The `ios-static` failure is just CI flakiness. The object server tests are well on their way to failing; this is an unrelated issue I'm attempting to troubleshoot right now and shouldn't block this PR.

PR #5361, opened 2017-09-30
Update to sync v2.0.0-rc26, for compatibility with ROS v2.0.0-alpha.42 and newer

bdash:
This is necessary because rc26 bumped the protocol version. For that reason I've not worried about attempting to enable any further tests as part of this change.


bdash:
Object server tests all passed. A number of the CI failures are due to device code signing certificates being expired, which I'm working on fixing. Other failures are due to apparent networking issues.

bdash:
I'm not convinced they did. It could quite easily have just been a weird CI issue that decided to go away.

austinzheng:
I've set https://github.com/realm/realm-cocoa/pull/5347 to rebuild at the commit it was last on. If it succeeds this time, then it's definitely CI flakiness.

austinzheng:
It was CI flakiness. Sigh.

PR #5362, opened 2017-10-02
Reenable most tests

austinzheng:
Changes:
- Reenabled all permissions-related tests that aren't still broken
- Rewrote a number of tests to fix problems with their semantics
- Removed `RLMSyncPermissionSortPropertyUserID`, since all permissions reported to a user have the same user ID
- Upgraded to ROS alpha-44
- Pulled in object store changes
- Tests should fail in a slightly more graceful manner now

Note that a handful of tests (prefixed with `disabled_`) are still disabled. All disabled tests have a `FIXME` comment explaining what's wrong.


Realm/RLMSyncPermissionResults.mm
```diff
@@ -64,8 +63,6 @@ bool keypath_is_valid(NSString *keypath)
 
 /// Sort by the Realm Object Server path to the Realm to which the permission applies.
 RLMSyncPermissionSortProperty const RLMSyncPermissionSortPropertyPath       = @"path";
-/// Sort by the identity of the user to whom the permission applies.
-RLMSyncPermissionSortProperty const RLMSyncPermissionSortPropertyUserID     = @"userId";
```
bdash: Was this also removed from the header? I don't see a removal when searching for `RLMSyncPermissionSortPropertyUserID` in this diff.

Realm/ObjectServerTests/RLMPermissionsAPITests.m
```diff
@@ -871,56 +831,79 @@ - (void)disabled_testSortingPermissionsOnDate {
...
-    __attribute__((objc_precise_lifetime)) RLMRealm *realm = [self openRealmForURL:url user:self.userA];
     NSString *uB = self.userB.identity;
+    // Open a Realm
+    {
+        NSURL *url = CUSTOM_REALM_URL(@"r1");
+        __attribute__((objc_precise_lifetime)) RLMRealm *realm = [self openRealmForURL:url user:self.userA];
+
+        // Give user B read permission for the Realm.
+        XCTestExpectation *ex = [self expectationWithDescription:@"Setting r1 permission for user B should work."];
+        [self.userA applyPermission:[[RLMSyncPermission alloc] initWithRealmPath:[makeTildeSubstitutedURL(url, self.userA) path]
```
tgoyne: I think this is `APPLY_PERMISSION_WITH_MESSAGE`?

Realm/ObjectServerTests/RLMSyncTestCase.mm
```diff
@@ -97,6 +97,16 @@ + (RLMSyncManager *)managerForCurrentTest {
     return s_managerForTest;
 }
 
+- (void)invokeTest {
```
tgoyne: What does this change do?

Realm/RLMSyncPermissionResults.mm
```diff
@@ -115,10 +112,8 @@ - (NSUInteger)indexOfObject:(RLMSyncPermission *)object {
     if ([path rangeOfString:@"~"].location != NSNotFound) {
         path = [path stringByReplacingOccurrencesOfString:@"~" withString:object.identity];
     }
-    // Build the predicate. (Don't match the actual permission flags; path and identity identify a permission.)
-    NSPredicate *p = [NSPredicate predicateWithFormat:@"%K = %@ AND %K = %@",
-                      RLMSyncPermissionSortPropertyPath, path,
-                      RLMSyncPermissionSortPropertyUserID, object.identity];
+    // Build the predicate. Check based on the path.
+    NSPredicate *p = [NSPredicate predicateWithFormat:@"%K = %@", RLMSyncPermissionSortPropertyPath, path];
     return [self indexOfObjectWithPredicate:p];
```
tgoyne: There isn't really any advantage to building the predicate first over just doing `return [self indexOfObjectWhere:@"%K = %@", RLMSyncPermissionSortPropertyPath, path];`

RealmSwift/Sync.swift
```diff
@@ -419,6 +419,26 @@ public typealias SyncPermission = RLMSyncPermission
  */
 public typealias SyncAccessLevel = RLMSyncAccessLevel
 
+/// :nodoc:
+public final class _SyncPermissionChange: Object {
```
bdash: [SwiftLint complains about this name](https://ci.realm.io/job/objc_pr/5771/configuration=Release,target=swiftlint,xcode_version=9.0/consoleFull#-709143373594bc32c4f0ccc7b2ba4d264):
> RealmSwift/Sync.swift:423:14: error: Type Name Violation: Type name should only contain alphanumeric characters: '_SyncPermissionChange' (type_name)

Realm/RLMSyncPermissionResults.mm
austinzheng: You are right; I missed it. Turns out I was only searching for occurrences of that string inside the tests group.

Realm/ObjectServerTests/RLMPermissionsAPITests.m
austinzheng: Yes it is. Thank you for catching that.

Realm/ObjectServerTests/RLMSyncTestCase.mm
austinzheng: Stack Overflow claims that it allows a test to stop running as soon as an XCTest assert fails, without aborting the rest of the test suite. I will further verify this; it seems to work for the Objective-C tests but not for the Swift ones (which inherit from this test case class as well).

bdash:
Most of the CI failures we're seeing are due to https://github.com/realm/ci/issues/211. The object server tests appears to be passing though, which is the main thing.

Realm/ObjectServerTests/RLMSyncTestCase.mm
bdash: I assume this is primarily beneficial when working on tests locally? If so, it might be worth dropping it from this PR until you can figure out why it's not working as you expect, and getting the remaining changes merged.

austinzheng:
Yeah, I'm going to remove that code (which doesn't seem like it works at all, upon further investigation).

austinzheng:
All failures are CI related.

PR #5363, opened 2017-10-03
Replace permission offer and response objects with APIs

austinzheng:
Changes:
- Removed `RLMSyncPermissionOffer` and `RLMSyncPermissionOfferResponse` models (and Swift equivalents)
- Removed a bunch of helper methods and types associated with those models
- Removed APIs allowing the management Realm to be directly opened for a given user
- Added two APIs on `RLMSyncUser`; one for creating offers and one for accepting them
- Rewrote object server integration tests to use the new API.

Requires https://github.com/realm/realm-object-store/pull/559


austinzheng:
macOS build timed out. A tvOS unit test unrelated to this change failed.

Realm/RLMSyncUser.mm
```diff
@@ -383,6 +368,65 @@ - (void)revokePermission:(RLMSyncPermission *)permission callback:(RLMPermission
...
+        callback(nil, make_permission_error_change(@"A permission offer cannot be accepted by an invalid user."));
+        return;
+    }
+    auto cb = [callback](util::Optional<std::string> raw_url, std::exception_ptr ptr) {
+        if (ptr) {
+            NSError *error = translateSyncExceptionPtrToError(std::move(ptr), RLMPermissionActionTypeAcceptOffer);
+            REALM_ASSERT(error);
+            callback(nil, error);
+        } else {
+            REALM_ASSERT_DEBUG(raw_url != nullptr);
```
austinzheng: Get rid of `!= nullptr`; `raw_url` is an optional.

Realm/RLMSyncPermission.mm
```diff
@@ -17,44 +17,13 @@
 ////////////////////////////////////////////////////////////////////////////
 
 #import "RLMSyncPermission_Private.hpp"
+#import "RLMSyncUtil_Private.hpp"
```
austinzheng: Remove this.

Realm/RLMSyncUser.mm
```diff
@@ -383,6 +368,65 @@ - (void)revokePermission:(RLMSyncPermission *)permission callback:(RLMPermission
...
+        callback(nil, make_permission_error_change(@"A permission offer cannot be accepted by an invalid user."));
+        return;
+    }
+    auto cb = [callback](util::Optional<std::string> raw_url, std::exception_ptr ptr) {
+        if (ptr) {
+            NSError *error = translateSyncExceptionPtrToError(std::move(ptr), RLMPermissionActionTypeAcceptOffer);
+            REALM_ASSERT(error);
+            callback(nil, error);
+        } else {
+            REALM_ASSERT_DEBUG(bool(raw_url));
```
bdash: The `bool` is unnecessary here. A conversion to bool happens within the `REALM_ASSERT` macros.

Realm/RLMSyncUser.mm
```diff
@@ -383,6 +368,65 @@ - (void)revokePermission:(RLMSyncPermission *)permission callback:(RLMPermission
...
+            NSString *urlAsString = [NSString stringWithUTF8String:raw_url->c_str()];
+            NSURLComponents *mutableURL = [NSURLComponents componentsWithString:urlAsString];
+            if (mutableURL) {
```
bdash: You can combine the declaration and initialization of `mutableURL` with the `if` test.

Realm/RLMSyncUser.mm
```diff
@@ -383,6 +368,65 @@ - (void)revokePermission:(RLMSyncPermission *)permission callback:(RLMPermission
...
+        callback(nil, make_permission_error_change(@"A permission offer cannot be created using an invalid user."));
+        return;
+    }
+    auto cb = [callback](util::Optional<std::string> token, std::exception_ptr ptr) {
+        if (ptr) {
+            NSError *error = translateSyncExceptionPtrToError(std::move(ptr), RLMPermissionActionTypeOffer);
+            REALM_ASSERT(error);
+            callback(nil, error);
+        } else {
+            REALM_ASSERT_DEBUG(token != nullptr);
```
bdash: `token` is an `Optional<std::string>` so comparing against `nullptr` seems odd. This can just be `REALM_ASSERT_DEBUG(token)`.

Realm/RLMSyncUser.mm
```diff
@@ -383,6 +368,65 @@ - (void)revokePermission:(RLMSyncPermission *)permission callback:(RLMPermission
...
+            callback([NSString stringWithUTF8String:token->c_str()], nil);
```
bdash: `@(token->c_str())`

Realm/RLMSyncUser.mm
```diff
@@ -383,6 +368,65 @@ - (void)revokePermission:(RLMSyncPermission *)permission callback:(RLMPermission
...
+        return;
+    }
+    auto cb = [callback](util::Optional<std::string> raw_url, std::exception_ptr ptr) {
+        if (ptr) {
+            NSError *error = translateSyncExceptionPtrToError(std::move(ptr), RLMPermissionActionTypeAcceptOffer);
+            REALM_ASSERT(error);
+            callback(nil, error);
+        } else {
+            REALM_ASSERT_DEBUG(bool(raw_url));
+            NSString *urlAsString = [NSString stringWithUTF8String:raw_url->c_str()];
```
bdash: `@(raw_url->c_str())`, which is probably concise enough you can do away with the extra local variable.

Realm/RLMSyncUtil_Private.hpp
```diff
@@ -35,6 +40,22 @@ std::shared_ptr<realm::SyncSession> sync_session_for_realm(RLMRealm *realm);
 
 CocoaSyncUserContext& context_for(const std::shared_ptr<realm::SyncUser>& user);
 
+#pragma mark - Access level conversion
+
+realm::AccessLevel accessLevelForObjcAccessLevel(RLMSyncAccessLevel level);
```
bdash: This should probably be `accessLevelForObjCAccessLevel`.

Realm/RLMSyncUser.h
```diff
@@ -269,13 +276,47 @@ NS_SWIFT_UNAVAILABLE("Use the full version of this API.");
 - (void)revokePermission:(RLMSyncPermission *)permission callback:(RLMPermissionStatusBlock)callback;
 
 /**
- Returns an instance of the Management Realm owned by the user.
+ Create a permission offer for a Realm.
+
+ A permission offer is used to grant a specific permission to a Realm you are allowed to manage
```
tgoyne: I found it hard to parse this sentence. Maybe something like "A permission offer is used to grant access to a Realm you manage to another user."

"You" is also inaccurate here, but it is more concise than "the current user" and people will probably understand what it means?

Realm/RLMSyncUser.h
```diff
@@ -269,13 +276,47 @@ NS_SWIFT_UNAVAILABLE("Use the full version of this API.");
...
+ */
+- (void)createOfferForRealmAtURL:(NSURL *)url
+                     accessLevel:(RLMSyncAccessLevel)accessLevel
+                      expiration:(nullable NSDate *)expirationDate
+                        callback:(RLMPermissionOfferStatusBlock)callback NS_REFINED_FOR_SWIFT;
+
+/**
+ Accept a permission offer.
+
+ Pass in a token representing a permission offer. The operation will take place asynchronously.
```
tgoyne: "The operation will take place asynchronously." appears twice here.

Realm/RLMSyncUtil_Private.hpp
austinzheng: Good catch. Thank you.

Realm/RLMSyncUser.h
austinzheng: Thanks for the suggestion; it is a significant improvement. Also, I was a little worried that using "you" might be too informal or imprecise, and given your feedback I think it's probably worthwhile to state the actor more accurately.

PR #5365, opened 2017-10-03
Update sync dependency to rc27

austinzheng:
It doesn't seem object store or the binding require changes to accommodate rc27. I'll let CI run to see if it finds any problems.


austinzheng:
All failing jobs are failing because of CI reasons.

PR #5374, opened 2017-10-05
Synchronize reading `s_defaultConfiguration` in `rawDefaultConfiguration` method

kishikawakatsumi:
`rawDefaultConfiguration` method returns the copied configuration object. It is to prevent becoming a freed pointer when `rawDefaultConfiguration` is set in a different thread.

If you execute code like the following, you can see that it crashes.

```
dispatch_async(dispatch_get_global_queue(QOS_CLASS_DEFAULT, 0), ^{
    for (int i = 0; i < 10000; i++) {
        [NSThread sleepForTimeInterval:0.0001];
    }
});
dispatch_async(dispatch_get_global_queue(QOS_CLASS_DEFAULT, 0), ^{
    for (int i = 0; i < 10000; i++) {
        [RLMRealmConfiguration setDefaultConfiguration:[RLMRealmConfiguration new]];
        [NSThread sleepForTimeInterval:0.0001];
    }
});
```

This probably happens by replacing `s_defaultConfiguration` with` setDefaultConfiguration:`, but other thread has already `s_defaultConfiguration`, it becomes a freed pointer.

CC @tgoyne @bdash

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,17 +8,8 @@ x.x.x Release notes (yyyy-MM-dd)
 ### Bugfixes
 * `-[RLMResults<RLMSyncPermission *> indexOfObject:]` now properly accounts for access
   level.
-### API Breaking Changes
-
-* None.
-
-### Enhancements
-
-* None.
-
-### Bugfixes
-
-* None.
+* Fix a race condition that could lead to a crash accessing to the freed configuration object
+  if a default configuration was set from a different thread.
 
 3.0.0-rc.1 Release notes (2017-10-03)
 =============================================================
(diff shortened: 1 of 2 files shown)


tgoyne:
`rawDefaultConfiguration` exists specifically to avoid doing this copy, as that has a very negative impact on the speed of `[RLMRealm defaultRealm]` when the `RLMRealm` is cached.

Fortunately I think the assignment to a local variable is sufficient to fix the race here and the actual copy is unnecessary, as the problem is just the unsynchronized read of `s_defaultConfiguration`.

kishikawakatsumi:
You're completely right, thanks.

PR #5375, opened 2017-10-05
Make SwiftLint happy

tgoyne:


tgoyne:
These are mostly new rules in 0.23 which just came out a few days ago.

PR #5376, opened 2017-10-05
Update ROS dependency and fix some tests

austinzheng:
Changes:
- Binding now uses ROS rc-1
- Removed the lowercasing of usernames when setting permissions by username
- Changed the way URLs are generated when permission offers are accepted

Notes:
Unfortunately, this doesn't change the number of test cases that are re-enabled. The "change permissions by username" test now passes in isolation, but it fails when run with the rest of the suite (like the "delete permission" test).


Realm/RLMSyncUser.mm
```diff
@@ -400,26 +400,24 @@ - (void)acceptOfferForToken:(NSString *)token
...
         return;
     }
-    auto cb = [callback](util::Optional<std::string> raw_url, std::exception_ptr ptr) {
+    NSURLComponents *baseURL = [NSURLComponents componentsWithString:[self.authenticationServer absoluteString]];
+    if ([baseURL.scheme isEqualToString:@"http"]) {
+        baseURL.scheme = @"realm";
+    } else if ([baseURL.scheme isEqualToString:@"https"]) {
+        baseURL.scheme = @"realms";
+    } else {
+        @throw RLMException(@"A ROS server URL must begin with http:// or https://.");
```
tgoyne: This error message should include the URL that the user passed in.

Realm/RLMSyncUser.mm
```diff
@@ -400,26 +400,24 @@ - (void)acceptOfferForToken:(NSString *)token
         callback(nil, make_permission_error_change(@"A permission offer cannot be accepted by an invalid user."));
         return;
     }
-    auto cb = [callback](util::Optional<std::string> raw_url, std::exception_ptr ptr) {
+    NSURLComponents *baseURL = [NSURLComponents componentsWithString:[self.authenticationServer absoluteString]];
```
bdash: Why use `+componentsWithString:` rather than `+componentsWithURL:resolvingAgainstBaseURL:`?

Realm/RLMSyncUser.mm
bdash: Shouldn't we be validating the protocol when creating the user rather than at this later point? Can we even reach this point if the auth server URL wasn't HTTP / HTTPS?

We should use "Realm Object Server" rather than ROS in error messages, and we should distinguish between authentication and Realm URLs (http(s) vs realm(s)).

austinzheng: Yeah, you're right. If you are a normal user, you by definition have to log in successfully first to even have a user object. If you are an admin token user we can do the validation there.

Realm/RLMSyncUser.mm
```diff
@@ -529,6 +526,11 @@ + (void)_performLoginForDirectAccessTokenCredentials:(RLMSyncCredentials *)crede
     NSString *identity = credentials.userInfo[kRLMSyncIdentityKey];
     std::shared_ptr<SyncUser> sync_user;
     if (serverURL) {
+        NSString *scheme = serverURL.scheme;
+        if (![scheme isEqualToString:@"http"] && ![scheme isEqualToString:@"https"]) {
+            @throw RLMException(@"The Realm Object Server authentication URL for this user, \"%@\", is improperly"
+                                @" configured. It must begin with http:// or https://.", serverURL);
```
bdash: I feel like "invalid" makes more sense in this error than "improperly configured".

PR #5377, opened 2017-10-05
Reenable more tests

austinzheng:
Changes:
- Reenabled several tests disabled because of ROS issues
- Fixed admin token test failure when running without the build.sh version of ROS

Requires a ROS release with https://github.com/realm/ros/pull/475.


Realm/ObjectServerTests/RLMSyncTestCase.mm
```diff
@@ -260,10 +260,14 @@ - (RLMSyncUser *)createAdminUserForURL:(NSURL *)url username:(NSString *)usernam
 }
 
 + (NSString *)retrieveAdminToken {
+#if PROVIDING_OWN_ROS
+    return @"eyJhcHBfaWQiOiJpby5yZWFsbS5hdXRoIiwiaWRlbnRpdHkiOiJfX2FkbWluIiwiYWNjZXNzIjpbImRvd25sb2FkIiwidXBsb2FkIiwibWFuYWdlIl0sInNhbHQiOiI0YTM0NmVlNyJ9:WmB4swR8pQoWfLn1z06z9icasvIBxtLAMcsmt9AfC2bBMWN/4jx/MyuxZVnaassjikvMLp/yfFc4OBPhrpJsm0rjaAtTODaq242qQCaY/eFy5SOd5BH+nJ9WPsW6ByN/ZpccqlaY6ihboEdxCZZRTlIUwy92oH9ljeUzDW1JuURjPnYPmGdAUDnGp+aB9ZHtC/SrKva2J8IGrI2npPZxq0EmN9cbjvo/NQQlBwYqmTX3K0Fa3UmvvxGAuySVpx2ynGUrNsD6LqNkMlO4rkPme49BKJsG40y69itExj0gx45FGYVCd+USba1um4kLusEcePMxR6/Q2kR504M+C68PWA==";
```
tgoyne: This has to be manually set to the token from your ROS instance, right? It'd be good to have all the things that need to be touched for that in one place. Could have this come from a define set in the same place as `PROVIDING_OWN_ROS` (or maybe even combine the two).

Realm/ObjectServerTests/RLMPermissionsAPITests.m
```diff
@@ -43,7 +43,7 @@
 
 #define CHECK_PERMISSION_COUNT(ma_results, ma_count) CHECK_PERMISSION_COUNT_PREDICATE(ma_results, ma_count, ==)
 
-#define CHECK_PERMISSION_PRESENT(ma_results, ma_permission) {                                                          \
+#define CHECK_PERMISSION_PRESENT(ma_results, ma_permission) @autoreleasepool {                                                          \
```
tgoyne: Which thing is this trying to avoid having escape from the block? Now that I'm looking at it again, the whole thing is a bit suspicious: the notification block should be using the RLMResults passed to it rather than holding onto a weak reference (which would happen to eliminate the autorelease from resolving the weak reference), and it should be calling `-invalidate` on the token rather than using `objc_precise_lifetime` to better match what actual users would be doing.

austinzheng: I was originally troubleshooting a failed assertion somewhere deep inside XCTest code that was logged to console, but did not cause the offending test to crash or fail. The issues you brought up are problematic indeed; I'm thinking about just replacing all these macros with functions that take line number/file name arguments. (I added the weak reference originally because otherwise the compiler emitted a warning about a potential retain cycle, something that doesn't seem to happen in similar non-macro code.)

Realm/ObjectServerTests/RLMSyncTestCase.mm
austinzheng: That's a good idea.

Realm/ObjectServerTests/RLMPermissionsAPITests.m
```diff
@@ -57,13 +57,13 @@
             }                                                                                                          \
         }                                                                                                              \
     }];                                                                                                                \
-    [self waitForExpectationsWithTimeout:10.0 handler:nil];                                                            \
+    [self waitForExpectations:@[ex] timeout:20.0];                                                                     \
```
bdash: Some extra whitespace crept in here at the end of line.

Realm/ObjectServerTests/RLMSyncTestCase.mm
```diff
@@ -261,7 +266,12 @@ - (RLMSyncUser *)createAdminUserForURL:(NSURL *)url username:(NSString *)usernam
 
 + (NSString *)retrieveAdminToken {
 #if PROVIDING_OWN_ROS
-    return @"eyJhcHBfaWQiOiJpby5yZWFsbS5hdXRoIiwiaWRlbnRpdHkiOiJfX2FkbWluIiwiYWNjZXNzIjpbImRvd25sb2FkIiwidXBsb2FkIiwibWFuYWdlIl0sInNhbHQiOiI0YTM0NmVlNyJ9:WmB4swR8pQoWfLn1z06z9icasvIBxtLAMcsmt9AfC2bBMWN/4jx/MyuxZVnaassjikvMLp/yfFc4OBPhrpJsm0rjaAtTODaq242qQCaY/eFy5SOd5BH+nJ9WPsW6ByN/ZpccqlaY6ihboEdxCZZRTlIUwy92oH9ljeUzDW1JuURjPnYPmGdAUDnGp+aB9ZHtC/SrKva2J8IGrI2npPZxq0EmN9cbjvo/NQQlBwYqmTX3K0Fa3UmvvxGAuySVpx2ynGUrNsD6LqNkMlO4rkPme49BKJsG40y69itExj0gx45FGYVCd+USba1um4kLusEcePMxR6/Q2kR504M+C68PWA==";
+#ifdef OWN_ROS_ADMIN_TOKEN
+    NSAssert(NO, @"Cannot run admin token related tests unless you define OWN_ROS_ADMIN_TOKEN.");
```
austinzheng: The if and else branches need to be swapped.

Realm/ObjectServerTests/RLMSyncTestCase.mm
```diff
@@ -260,10 +265,19 @@ - (RLMSyncUser *)createAdminUserForURL:(NSURL *)url username:(NSString *)usernam
 }
 
 + (NSString *)retrieveAdminToken {
+#if PROVIDING_OWN_ROS
+#ifdef OWN_ROS_ADMIN_TOKEN
```
bdash: Surely this should be `#ifndef`?

PR #5379, opened 2017-10-06
Replace ROS test macros with normal methods

austinzheng:
The original point of the macros was so that, when tests failed, we'd have useful line numbers to diagnose the problem. The macros have caused more problems than they solved, though, and it's time to replace them.

### DIFF
--- a/Realm/ObjectServerTests/RLMPermissionsAPITests.m
+++ b/Realm/ObjectServerTests/RLMPermissionsAPITests.m
@@ -22,88 +22,6 @@
 
 #import "RLMTestUtils.h"
 
-#define CHECK_PERMISSION_COUNT_PREDICATE(ma_results, ma_count, ma_op) {                                                \
-    XCTestExpectation *ex = [self expectationWithDescription:@"Checking permission count"];                            \
-    __weak typeof(ma_results) weakResults = ma_results;                                                                \
-    __attribute__((objc_precise_lifetime))id token = [ma_results addNotificationBlock:^(__unused id r,                 \
-                                                                                        __unused id c,                 \
-                                                                                        NSError *err) {                \
-        XCTAssertNil(err);                                                                                             \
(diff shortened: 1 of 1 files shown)


Realm/ObjectServerTests/RLMPermissionsAPITests.m
```diff
@@ -166,6 +84,96 @@ - (void)tearDown {
     [super tearDown];
 }
 
+#pragma mark - Permission validation methods
+
+// This macro is only used for the validation methods below.
+#define RECORD_FAILURE(ma_msg) [self recordFailureWithDescription:ma_msg inFile:file atLine:line expected:YES]
+
+#define CHECK_PERMISSION_PRESENT(ma_results, ma_permission) \
+[self checkPresenceOfPermission:ma_permission inResults:ma_results line:__LINE__ file:@(__FILE__)]
```
tgoyne: These should be indented.

Realm/ObjectServerTests/RLMPermissionsAPITests.m
```diff
@@ -166,6 +84,96 @@ - (void)tearDown {
...
+                             file:(NSString *)file {
+    XCTestExpectation *ex = [self expectationWithDescription:@"Checking presence of permission..."];
+    RLMNotificationToken *token = [results addNotificationBlock:^(RLMResults *r, __unused id c, NSError *err) {
+        if (err) {
+            RECORD_FAILURE(@"Failed to retrieve permissions.");
+            [ex fulfill];
+            return;
+        }
+        for (NSUInteger i = 0; i < r.count; i++) {
+            if ([[r objectAtIndex:i] isEqual:permission]) {
```
tgoyne: Shouldn't `indexOfObject:` work on permissions now?

austinzheng:
Failing jobs are all failing because of CI issues. The 8.3.3 object server test failure is something unrelated that I've never seen before, so I think it's very likely just a transient error.

PR #5385, opened 2017-10-09
Fixed an issue that crash when enumerating after clearing data during migration.

### LINKED ISSUE #5368: Crash: OutOfBoundsIndexException after RLMMigration.delete() inside migration
## Goals

Delete some objects inside migration
## Expected Results

Migration should finish successfully

## Actual Results

It crashes around [RLMMigration deleteObjectsMarkedFprDeletion]
screen shot 2017-10-04 at 4 14 10 pm

## Steps to Reproduce

Delete some objects of a class during migration (in my case, delete 5 out of 45). After the migration is finished, it triggered a crash as depicted from above
## Code Sample

I am sorry, I cannot give you the project which I am working on has the issue. However, I also come up a sample which can also crash when executing `enumerateObjects` after `deleteDataForClassName`

```swift
import Realm

let isNewVersion = false

class Foo: RLMObject {
  @objc dynamic var id: String = ""
  
  class func primarykey() -> String {
    return #keyPath(id)
  }
}

@UIApplicationMain
class AppDelegate: UIResponder, UIApplicationDelegate {

  var window: UIWindow?


  func application(_ application: UIApplication, didFinishLaunchingWithOptions launchOptions: [UIApplicationLaunchOptionsKey: Any]?) -> Bool {

    let config = RLMRealmConfiguration.default()
    config.schemaVersion = isNewVersion ? 2 : 1
    config.migrationBlock = { (migration, oldSchemaVersion) in
      if isNewVersion {
        if oldSchemaVersion < 2 {
          migration.deleteData(forClassName: String(describing: Foo.self))
          
          migration.enumerateObjects(String(describing: Foo.self)) { _, _ in
            print("one enumerated after deletion\n")
          }
        }
      }
    }
    
    RLMRealmConfiguration.setDefault(config)
    let realm = RLMRealm.default()
    
    if !isNewVersion {
      try? realm.transaction {
        let foo1 = Foo()
        foo1.id = "1"
        let foo2 = Foo()
        foo2.id = "2"
        let foo3 = Foo()
        foo3.id = "3"
        let foo4 = Foo()
        foo4.id = "4"
        
        realm.add(foo1)
        realm.add(foo2)
        realm.add(foo3)
        realm.add(foo4)
      }
    }
    
    return true
  }
}

```
For first run, set `isNewVersion` to false. For second run, set `isNewVersion` to true, it should crash

Realm framework version: 2.10.0

Realm Object Server version: -

Xcode version: Xcode 9.0

iOS/OSX version: iOS 11 / macOS 10.12.6

Dependency manager + version: Carthage 0.25.0

kishikawakatsumi:
Fixes #5368

Defer deletion during migration and actually delete the object when migration finished.
As for `- [RLMMigration deleteObject:]` it was already doing that. But `- [RLMMigration deleteDataForClassName:]` doesn't. This PR fixed as well.

See https://github.com/realm/realm-cocoa/pull/4864

CC @bdash @austinzheng

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ x.x.x Release notes (yyyy-MM-dd)
   level.
 * Fix a race condition that could lead to a crash accessing to the freed configuration object
   if a default configuration was set from a different thread.
+* Fixed an issue that crash when enumerating after clearing data during migration.
 
 3.0.0-rc.1 Release notes (2017-10-03)
 =============================================================
--- a/Realm/RLMMigration.mm
+++ b/Realm/RLMMigration.mm
@@ -57,6 +57,7 @@ using namespace realm;
 @implementation RLMMigration {
     realm::Schema *_schema;
     NSMutableDictionary *deletedObjectIndices;
+    NSMutableSet *deletedClasses;
 }
 
 - (instancetype)initWithRealm:(RLMRealm *)realm oldRealm:(RLMRealm *)oldRealm schema:(realm::Schema &)schema {
@@ -67,6 +68,7 @@ using namespace realm;
         _schema = &schema;
         object_setClass(_oldRealm, RLMMigrationRealm.class);
         deletedObjectIndices = [NSMutableDictionary dictionary];
+        deletedClasses = [NSMutableSet set];
     }
     return self;
 }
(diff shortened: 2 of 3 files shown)


austinzheng:
Looks good! There are a few code formatting issues SwiftLint found.

PR #5393, opened 2017-10-12
Add back in support for sorting permissions on user ID

austinzheng:
ROS decided that permissions should live in both the giver and receiver's permissions Realms, so it makes sense to allow sorting by user ID again.

Also fix how permissions results' `indexForObject:` method works to properly take into account permission level.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,14 @@
+3.0.0-rc.2 Release notes (2017-??-??)
+=============================================================
+
+### Enhancements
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -1,3 +1,14 @@
+3.0.0-rc.2 Release notes (2017-??-??)
```
bdash: This should read:
```
x.x.x Release notes (yyyy-MM-dd)
```

It gets updated with the version + date during the release process.

austinzheng: Oh, that's nifty!

PR #5395, opened 2017-10-14
Update to the latest versions of core, sync and ROS

bdash:
Fixes https://github.com/realm/realm-cocoa/issues/5296.


austinzheng:
I think somehow the version of my recent PR that made it into master didn't disable `testReadAccess` and `testDeleteRealmUponPermissionDenied`. If CI fails on those tests they probably need to be turned off until I can figure out exactly why they fail when ROS is run from `NSTask`.

bdash:
The object server tests passed locally for me when I made this PR.

austinzheng:
Fixes https://github.com/realm/realm-cocoa/issues/5296.

PR #5396, opened 2017-10-14
Update object store to pull in fixes

austinzheng:
TODO:
- [x] Retarget `master` once #5395 is in.

### DIFF
--- a/Realm/Tests/AsyncTests.mm
+++ b/Realm/Tests/AsyncTests.mm
@@ -721,11 +721,11 @@
         }
     }];
 
-    // Triggers one call on each block
+    // Nested call is deferred.
     [self waitForNotification:RLMRealmDidChangeNotification realm:results.realm block:^{
         [self createObject:0];
     }];
-    XCTAssertEqual(calls, 2);
+    XCTAssertEqual(calls, 1);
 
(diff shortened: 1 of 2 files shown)


austinzheng:
Looks like a number of tests need to be updated in response to https://github.com/realm/realm-object-store/pull/573.

Realm/Tests/AsyncTests.mm
```diff
@@ -721,11 +721,11 @@ - (void)testAddingBlockFromWithinNotificationBlock {
         }
     }];
 
-    // Triggers one call on each block
+    // Nested call is deferred.
     [self waitForNotification:RLMRealmDidChangeNotification realm:results.realm block:^{
         [self createObject:0];
     }];
-    XCTAssertEqual(calls, 2);
+    XCTAssertEqual(calls, 1);
```
tgoyne: I think `XCTAssertEqual(calls, 1); [results.realm refresh]; XCTAssertEqual(calls, 2);` will work to test this behavior.

austinzheng: Yes, that would better retain the original coverage of the test.

austinzheng:
All failures are CI related.

PR #5421, opened 2017-10-23
Fix regression where Swift APIs didn't work correctly with `type(of:)`

### LINKED ISSUE #5414: Missing primary key exception for object type
When creating a Realm object as following:

```swift
realm.create(type(of: object), value: object, update: true)
```

an exception is raised:
```
*** Terminating app due to uncaught exception 'RLMException', reason: ''RLM:Unmanaged Site' does not have a primary key and can not be updated'
```

This appears to be due to `typeName` resolving to "RLM:Unmanaged Site"
```
(lldb) po typeName
"RLM:Unmanaged Site"
```
Which fails lookup in `schema` for identifying `primaryKeyProperty`. 

When adding a breakpoint within `[RLMObjectBase className]`  
The output is as follows:
```
(lldb) po [self className]
Site

(lldb) po className
RLM:Unmanaged Site

(lldb) po [self sharedSchema]
Site {
	id {
		type = string;
		objectClassName = (null);
		linkOriginPropertyName = (null);
		indexed = YES;
		isPrimary = YES;
		array = NO;
		optional = YES;
	}
```

```
ProductName:	Mac OS X
ProductVersion:	10.12.6
BuildVersion:	16G29

Xcode 9.0
Build version 9A235

/usr/local/bin/pod
1.4.0.beta.1
Realm (3.0.0)
RealmSwift (3.0.0)

/bin/bash

/usr/bin/git
git version 2.13.5 (Apple Git-94)
```

austinzheng:
Changes:
- ~Swift APIs that take an object type as a parameter can now properly handle receiving the type name of a Realm private subclass (e.g. a value returned from `type(of:)` called on an object instance)~ Fix regression.

Fixes #5414.

### DIFF
--- a/Realm/RLMObjectBase.mm
+++ b/Realm/RLMObjectBase.mm
@@ -209,6 +209,12 @@ id RLMCreateManagedAccessor(Class cls, __unsafe_unretained RLMRealm *realm, RLMC
     if ([RLMSwiftSupport isSwiftClassName:className]) {
         className = [RLMSwiftSupport demangleClassName:className];
     }
+    if ([className hasPrefix:@"RLM:"] ) {
+        // Getting an object type at runtime may return the managed or unmanaged
+        // subclass type name instead. If so, demangle it.
+        NSArray<NSString *> *parts = [className componentsSeparatedByString:@" "];
+        return [parts lastObject];
+    }
     return className;
 }
 
--- a/Realm/RLMObjectBase.mm
+++ b/Realm/RLMObjectBase.mm
@@ -209,12 +209,6 @@ id RLMCreateManagedAccessor(Class cls, __unsafe_unretained RLMRealm *realm, RLMC
     if ([RLMSwiftSupport isSwiftClassName:className]) {
         className = [RLMSwiftSupport demangleClassName:className];
     }
-    if ([className hasPrefix:@"RLM:"] ) {
-        // Getting an object type at runtime may return the managed or unmanaged
-        // subclass type name instead. If so, demangle it.
-        NSArray<NSString *> *parts = [className componentsSeparatedByString:@" "];
-        return [parts lastObject];
-    }
     return className;
 }
 
--- a/RealmSwift/Object.swift
+++ b/RealmSwift/Object.swift
@@ -133,14 +133,16 @@ open class Object: RLMObjectBase, ThreadConfined, RealmCollectionValue {
 
     /**
      WARNING: This is an internal helper method not intended for public use.
+     It should never be overriden and is not considered part of the public API.
      :nodoc:
      */
-    public override final class func className() -> String {
+    open override class func className() -> String {
         return super.className()
     }
 
     /**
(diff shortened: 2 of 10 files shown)


tgoyne:
`className` is a method we define, and the main point of it is that it returns the name of the user-defined class rather than the names of the generated subclasses. If it's not doing that, that's the bug that needs to be fixed.

Realm/RLMObjectBase.mm
```diff
@@ -209,6 +209,12 @@ + (NSString *)className {
     if ([RLMSwiftSupport isSwiftClassName:className]) {
         className = [RLMSwiftSupport demangleClassName:className];
     }
+    if ([className hasPrefix:@"RLM:"] ) {
```
tgoyne: This shouldn't ever actually be hit. During [schema discovery](https://github.com/realm/realm-cocoa/blob/master/Realm/RLMSchema.mm#L132) we [override the className method on each subclass](https://github.com/realm/realm-cocoa/blob/master/Realm/RLMAccessor.mm#L491). That should always happen before any of the subclasses are created, but if this change does something then... I'm not sure. Maybe the unmanaged generated subclass is getting created before schema discovery, and then `RLMReplaceClassNameMethod()` is getting called on it? Maybe `RLMReplaceClassNameMethod()` just plain doesn't work and that somehow doesn't break anything besides this?

austinzheng: This is good to know. I’ll investigate further; I know that this change is affecting something so there most likely is a problem lying deeper.

austinzheng: I think the cause is the `public final` declarations I added to the definition of `className()` in `Object` for 3.0.0. I'll verify and make a fix tomorrow.

austinzheng:
This is indeed a regression, caused by a change that I made for 3.0.0. At least we'll get some unit tests out of this PR.

austinzheng:
It looks like the 8.3.3 tests are failing after this change. i'll download that version of Xcode tomorrow and investigate further.

RealmSwift/Object.swift
```diff
@@ -133,14 +133,16 @@ open class Object: RLMObjectBase, ThreadConfined, RealmCollectionValue {
 
     /**
      WARNING: This is an internal helper method not intended for public use.
+     It should never be overriden and is not considered part of the public API.
      :nodoc:
      */
-    public override final class func className() -> String {
+    open override class func className() -> String {
```
tgoyne: Is there any reason for this override to still exist at all? I think it was just here to mark it as final, so if that doesn't actually work it can probably just be deleted.

austinzheng: Removing the override fixes the issue with the tests not working properly with 8.3.3. I think the pre-3.0.0 code was the useless instance method, so like you said the right thing to do was to remove the method definition altogether.

Chandlerdea:
Any idea when this will be included in the next release?

tgoyne:
We'll cut another release some time this week (probably tomorrow).

PR #5428, opened 2017-10-25
Update object store and sync

tgoyne:
I don't think there's any user-relevant changes from sync 2.1 to mention in the changelog.


austinzheng:
I'm a little concerned that upgrading sync seems to have broken the object server tests. I'm looking into it right now.

tgoyne:
Go for it.

austinzheng:
It's ready to merge whenever, assuming my changes are acceptable.

tgoyne:
Change looks fine, but going to wait for it to run on CI in case there's anything goofy in that env.

PR #5440, opened 2017-10-30
Fix a crash when dynamically getting a linking objects

### LINKED ISSUE #5436: RLMDynamicGetByName fails due to _rlmArray unrecognized selector
## Actual Results
```
2017-10-30 13:38:01.841851+1100 CleverSparky[25595:844477] -[_TtGC10RealmSwift14LinkingObjectsC12CleverSparky11Maintenance_ _rlmArray]: unrecognized selector sent to instance 0x600000686720
```

## Steps to Reproduce

Create a model which has a `LinkingObjects` property.

```swift
class Fitting: Object {
    let maintenance = LinkingObjects(fromType: Maintenance.self, property: "fitting")
```

Then add the linking object to a Realm:

```swift
let realm = try! Realm(configuration: Realm.Configuration(inMemoryIdentifier: Constants.RealmIdentifiers.InMemoryRealm))
realm.beginWrite()
realm.add(maintenance, update: true)
try! realm.commitWrite()
```

Now attempt to reference the linking objects by subscript:
```swift
let maintenance = fitting["maintenance"]
print("maintenance: \(maintenance)")
```

```
ProductName:	Mac OS X
ProductVersion:	10.12.6
BuildVersion:	16G29

Xcode 9.0
Build version 9A235

/usr/local/bin/pod
1.4.0.beta.1
Realm (3.0.1)
RealmSwift (3.0.1)

/bin/bash

/usr/local/bin/carthage
0.15.2
(not in use here)

/usr/bin/git
git version 2.13.5 (Apple Git-94)
```

austinzheng:
See the connected issue for more information. This PR makes it so that retrieving a linking objects property using the dynamic getter in Swift produces a sensible result instead of crashing.

I'm not completely happy with this solution because it relies on the Swift part of the binding to perform a conditional cast (and because it returns a `Results` rather than a `LinkingObjects`), but it was the least invasive solution I could come up with.

Fixes #5436.

### DIFF
--- a/RealmSwift/Tests/ObjectAccessorTests.swift
+++ b/RealmSwift/Tests/ObjectAccessorTests.swift
@@ -346,13 +346,12 @@ class ObjectAccessorTests: TestCase {
             XCTFail("Got an unexpected nil for fido[\"owners\"]")
             return
         }
-        XCTAssertTrue(owners is Results<Object>)
+        XCTAssertTrue(owners is LinkingObjects<SwiftOwnerObject>)
         // Make sure the results actually functions.
-        guard let firstOwner = (owners as? Results<Object>)?.first else {
+        guard let firstOwner = (owners as? LinkingObjects<SwiftOwnerObject>)?.first else {
             XCTFail("Was not able to get first owner")
             return
         }
-        let ownerName = firstOwner["name"] as? String
-        XCTAssertEqual(ownerName, "JP")
+        XCTAssertEqual(firstOwner.name, "JP")
     }
 }
--- a/RealmSwift/Tests/ObjectAccessorTests.swift
+++ b/RealmSwift/Tests/ObjectAccessorTests.swift
@@ -329,4 +329,30 @@ class ObjectAccessorTests: TestCase {
         object.optObjectCol = nil
         XCTAssertNil(object.optObjectCol)
     }
+
+    func testLinkingObjectsDynamicGet() {
+        let fido = SwiftDogObject()
+        let owner = SwiftOwnerObject()
+        owner.dog = fido
+        owner.name = "JP"
+        let realm = try! Realm()
+        try! realm.write {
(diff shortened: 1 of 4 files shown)


austinzheng:
I should test this with an `Object` property as well. I suspect that something similar might happen in that case.

tgoyne:
I think just changing the inner check to `if (prop.type != RLMPropertyTypeLinkingObjects && !list._rlmArray) {` would work without any of the other changes, and would make it return a LinkingObjects object. However, I'm not sure when the `list._rlmArray = RLMDynamicGet(obj, prop);` bit is actually needed, so maybe it'd require some similar logic? The fact that LinkingObjects properties don't exist in pure-dynamic mode might also mean it doesn't.

RealmSwift/Tests/ObjectAccessorTests.swift
```diff
@@ -329,4 +329,30 @@ class ObjectAccessorTests: TestCase {
...
     }
+
+    func testLinkingObjectsDynamicGet() {
+        let fido = SwiftDogObject()
+        let owner = SwiftOwnerObject()
+        owner.dog = fido
+        owner.name = "JP"
+        let realm = try! Realm()
+        try! realm.write {
+            realm.add([fido, owner])
```
markst: @austinzheng is it necessary to add both dog and owner to realm? in my example I had only adding the 'child' / linking object.

austinzheng: It's not, and it doesn't produce a different outcome.

austinzheng:
Changing the inner check is sufficient to return a functional LinkingObjects, without any of the Swift wrapping stuff I wrote.

markst:
Just given this a whirl with my current setup & crash still seems to be occurring.

markst:
Resolves crash for me 👍

PR #5449, opened 2017-11-02
Translate exceptions thrown when initializing RLMSyncManager to RLMExceptions

tgoyne:
Currently if an error occurs (such as the user denying access to the keychain) the program just crashes when the exception bubbles out of dispatch_once() with no indication as to what went wrong. Rethrowing the error as a NSException and using std::call_once() (which allows propagating exceptions) makes it so that there is at least an error message printed.


austinzheng:
All the failures look like various CI issues.

PR #5467, opened 2017-11-08
Support for distinct RLMResults

gcox:
This was started in PR #5182, but abandoned by the PR's contributor.
Related Issue: #1103 

- [x] Add ObjC support
- [x] Add ObjC tests
- [x] Add Swift support
- [x] Add Swift tests - ~~**need some help**~~

I am not very familiar with Swift, and I'm not sure exactly where I should be putting the Swift related tests for this addition.

### DIFF
--- a/Realm/RLMResults.mm
+++ b/Realm/RLMResults.mm
@@ -369,10 +369,6 @@ static inline void RLMResultsValidateInWriteTransaction(__unsafe_unretained RLMR
         @throw RLMException(@"Must specify at least one keypath.");
     }
     for (NSString *keyPath in keyPaths) {
-        if ([[keyPath stringByTrimmingCharactersInSet:[NSCharacterSet whitespaceAndNewlineCharacterSet]]
-             isEqualToString:@""]) {
-            @throw RLMException(@"A valid keypath is required");
-        }
         if ([keyPath containsString:@"@"]) {
             @throw RLMException(@"Cannot distinct on keypath '%@': KVC collection operators are not supported.", keyPath);
         }
--- a/Realm/RLMResults.mm
+++ b/Realm/RLMResults.mm
@@ -376,14 +376,12 @@ static inline void RLMResultsValidateInWriteTransaction(__unsafe_unretained RLMR
             return self;
         }
         
-        __block std::vector<std::string> kp;
+        std::vector<std::string> keyPathsVector;
         for (NSString *keyPath in keyPaths) {
-            kp.push_back(std::string([keyPath UTF8String]));
+            keyPathsVector.push_back(keyPath.UTF8String);
         }
         
-        return [RLMResults
-                resultsWithObjectInfo:*_info
-                results:_results.distinct(kp)];
+        return [RLMResults resultsWithObjectInfo:*_info results:_results.distinct(keyPathsVector)];
     });
 }
 
--- a/Realm/RLMResults.mm
+++ b/Realm/RLMResults.mm
@@ -364,6 +364,41 @@ static inline void RLMResultsValidateInWriteTransaction(__unsafe_unretained RLMR
     });
 }
 
+- (RLMResults *)distinctResultsUsingKeyPaths:(NSArray<NSString *> *)keyPaths {
+    if (keyPaths.count == 0) {
+        @throw RLMException(@"Must specify at least one keypath.");
+    }
+    for (NSString *keyPath in keyPaths) {
+        if (keyPath == nil
+            || [[keyPath stringByTrimmingCharactersInSet:[NSCharacterSet whitespaceAndNewlineCharacterSet]]
+                isEqualToString:@""]) {
+                @throw RLMException(@"A valid keypath is required");
+            }
+        if ([keyPath containsString:@"@"]) {
+            @throw RLMException(@"Cannot distinct on keypath '%@': KVC collection operators are not supported.", keyPath);
+        }
+    }
+    
+    return translateRLMResultsErrors([&] {
+        if (_results.get_mode() == Results::Mode::Empty) {
+            return self;
+        }
+        
+        const auto& table = _results.get_tableview().get_parent();
+        __block std::vector<std::vector<size_t>> column_indicies;
+        for (NSString *keyPath in keyPaths) {
+            std::string keyPath_name = std::string([keyPath UTF8String]);
+            size_t col_idx = table.get_descriptor()->get_column_index(keyPath_name);
+            REALM_ASSERT(col_idx != size_t(-1));
+            column_indicies.push_back({ col_idx });
+        }
+        
+        return [RLMResults
+                resultsWithObjectInfo:*_info
+                results:_results.distinct({ table, std::move(column_indicies) })];
+    });
+}
+
 - (id)objectAtIndexedSubscript:(NSUInteger)index {
     return [self objectAtIndex:index];
 }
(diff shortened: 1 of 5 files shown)


gcox:
It would seem that the Swift tests need to go in the `RealmCollectionTypeTests.swift` file in the `ResultsTests` class, but I'm not confident about that given how different the swift tests are from the obj-c tests. I'm lost without the `DogObject` class :)

bdash:
I'd suggest modeling tests for this on `testSortWithProperty` / `testSortWithDescriptor` in RealmCollectionTypeTests.swift.

Realm/RLMResults.mm
```diff
@@ -364,6 +364,41 @@ - (RLMResults *)sortedResultsUsingDescriptors:(NSArray<RLMSortDescriptor *> *)pr
     });
 }
 
+- (RLMResults *)distinctResultsUsingKeyPaths:(NSArray<NSString *> *)keyPaths {
+    if (keyPaths.count == 0) {
+        @throw RLMException(@"Must specify at least one keypath.");
+    }
+    for (NSString *keyPath in keyPaths) {
+        if (keyPath == nil
+            || [[keyPath stringByTrimmingCharactersInSet:[NSCharacterSet whitespaceAndNewlineCharacterSet]]
```
bdash: I'm not quite sure what this is doing. Whitespace isn't valid in key paths.

gcox: Right, that's why the check is there. If someone passes a nonsense empty key path for some reason, it will throw early. If you don't feel it's necessary, I'm happy to remove it though.

bdash: Key paths like that will already be rejected when attempting to look their components up as property names. It doesn't strike me as useful to add extra checking for the unusual case of passing an empty or white-space only string. There are plenty of other invalid key paths that we're not explicitly checking for here.

gcox:
I believe everything, so far, has been addressed with the exception of the Swift tests. I'll try to knock those out tomorrow.

RealmSwift/Results.swift
```diff
@@ -254,6 +254,16 @@ public final class Results<Element: RealmCollectionValue>: NSObject, NSFastEnume
...
     }
+    
+    /**
+     Returns a `Results` containing distinct objects based on the specified key paths
+     
+     - parameter keyPaths:  The key paths used produce distinct results
+     */
+    public func distinct<S: Sequence>(by keyPaths: S) -> Results<Element>
+        where S.Iterator.Element == String {
+            return Results<Element>(rlmResults.distinctResults(usingKeyPaths: keyPaths.map { $0 }))
```
bdash: Can `keyPaths.map { $0 }` just be `Array(keyPaths)`?

Realm/RLMResults.mm
```diff
@@ -364,6 +364,32 @@ - (RLMResults *)sortedResultsUsingDescriptors:(NSArray<RLMSortDescriptor *> *)pr
     });
 }
 
+- (RLMResults *)distinctResultsUsingKeyPaths:(NSArray<NSString *> *)keyPaths {
+    if (keyPaths.count == 0) {
+        @throw RLMException(@"Must specify at least one keypath.");
```
bdash: Is there any reason to throw in this case given that we don't when sorting?

Realm/RLMResults.mm
```diff
@@ -364,6 +364,32 @@ - (RLMResults *)sortedResultsUsingDescriptors:(NSArray<RLMSortDescriptor *> *)pr
...
+        }
+        
+        __block std::vector<std::string> kp;
+        for (NSString *keyPath in keyPaths) {
+            kp.push_back(std::string([keyPath UTF8String]));
+        }
+        
+        return [RLMResults
+                resultsWithObjectInfo:*_info
+                results:_results.distinct(kp)];
```
bdash: This could happily fit on one line without being excessively long, which would be preferable to the somewhat odd wrapping that's going on here.

Realm/RLMResults.mm
```diff
@@ -364,6 +364,32 @@ - (RLMResults *)sortedResultsUsingDescriptors:(NSArray<RLMSortDescriptor *> *)pr
...
+            @throw RLMException(@"Cannot distinct on keypath '%@': KVC collection operators are not supported.", keyPath);
+        }
+    }
+    
+    return translateRLMResultsErrors([&] {
+        if (_results.get_mode() == Results::Mode::Empty) {
+            return self;
+        }
+        
+        __block std::vector<std::string> kp;
```
bdash: `__block` is unnecessary here.

I don't like using `kp` for a name here, but I don't have any other great ideas for how to otherwise distinguish this from `keyPaths` 🤷‍♂️

Realm/RLMResults.mm
```diff
@@ -364,6 +364,32 @@ - (RLMResults *)sortedResultsUsingDescriptors:(NSArray<RLMSortDescriptor *> *)pr
...
+        for (NSString *keyPath in keyPaths) {
+            kp.push_back(std::string([keyPath UTF8String]));
```
bdash: The explicit conversion to `std::string` shouldn't be necessary here. `kp_push_back(keyPath.UTF8String)` should compile just fine.

Realm/RLMResults.mm
gcox: Nope

Realm/RLMResults.mm
gcox: `keyPathVector`? Verbose, but clear.

bdash: Yeah, I'd be fine with that.

gcox: Went with `keyPathsVector`...plural `keyPaths` may be redundant given that it is a vector...

RealmSwift/Results.swift
gcox: Indeed

gcox:
Cleanup, round two, done 👍

gcox:
Are the Jenkins failures anything I can address? I don't have permission to view the details of the failures.

tgoyne:
The cocoapods tests are failing because that job doesn't work for branches not in the main repo. The ios-static failure is just CI flakiness.

The SwiftLint failures are:

RealmSwift/Results.swift:257: warning: Trailing Whitespace Violation: Lines should not have trailing whitespace. (trailing_whitespace)
RealmSwift/Tests/RealmCollectionTypeTests.swift:698:40: warning: Trailing Semicolon Violation: Lines should not have trailing semicolons. (trailing_semicolon)
[a lot more lines in this file]
RealmSwift/Tests/RealmCollectionTypeTests.swift:697: warning: Trailing Whitespace Violation: Lines should not have trailing whitespace. (trailing_whitespace)
[a lot more lines in this file]
RealmSwift/Tests/RealmCollectionTypeTests.swift:732:31: warning: Colon Violation: Colons should be next to the identifier when specifying a type and next to the key in dictionary literals. (colon)
[a lot more lines in this file]

gcox:
@tgoyne Thanks!

gcox:
@bdash This isn't waiting for anything from me, is it? Just making sure I'm not holding it up. I wish Github's review system had an official way to indicate "I think I'm done with the requested changes."

grafele:
I‘m waiting for this feature a long time. What‘s holding this merge request back?

tgoyne:
Nothing other than that I guess we were all leaving it for someone else to merge since we're used to having the person who made the PR merge it.

gcox:
Oh, was I actually allowed to merge it after the changes were approved? If so, sorry for the hold up, just assumed I didn't have permission.

tgoyne:
No, pretty sure you couldn't have; I just forgot about the whole outside contributor thing.

bdash:
One thing that was missing here was a change log entry. I'll add one myself later this week if no-one beats me to it.

gcox:
#1103 needs to be closed too...maybe you do that after release

PR #5484, opened 2017-11-22
Update RLMCollection.h

mohuilin:
new xcode version build failed


bdash:
Including RLMCollection.h from itself doesn’t make sense. What problem are you trying to solve?

bdash:
Closing due to the lack of response.

PR #5493, opened 2017-11-29
adds a list of the supported types List can store

icanzilb:
The Realm docs on List do not explain what types it can store (https://realm.io/docs/swift/latest/#collections) and the 3.0 API docs aren't up to date either.

This PR adds the list of supported types to the API docs (type list taken from the 3.0 release notes)

### DIFF
--- a/RealmSwift/List.swift
+++ b/RealmSwift/List.swift
@@ -42,7 +42,9 @@ public class ListBase: RLMListBase {
 /**
  `List` is the container type in Realm used to define to-many relationships.
 
- Like Swift's `Array`, `List` is a generic type that is parameterized on the type of `Object` it stores.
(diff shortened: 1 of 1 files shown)


bdash:
The long form docs _do_ mention it, under the heading Many-to-many where lists are described.

icanzilb:
@bdash They do indeed thanks for pointing that out. Maybe we do need a bit of structural change in these docs - I mean - I couldn't find it myself. Plus `List` isn't only the way to handle relationships (as it's currently presented in the docs) in Realm now - they can also hold values. 

Sidenote: The structure of the docs in that place is strange, there's a "Collections" section that mostly just redirects to the API docs with one sub-section called "Copying objects between Realms", which doesn't seemingly have to do with Collections 🤔

PR #5502, opened 2017-12-04
Upgrade dependencies

austinzheng:
Changes:
- Upgraded object store to latest
- Upgraded ROS and sync dependencies to 2.1.0 and 2.1.8, respectively

Requires: https://github.com/realm/realm-object-store/pull/609 (then the object store pointer needs to be updated).


austinzheng:
All failures are CI issues, it seems.

PR #5521, opened 2017-12-14
removes obsolete definition of 'SWIFT_WHOLE_MODULE_OPTIMIZATION'

geoffmacd:
This fixes a warning generated by the Pods project when building with cocoa pods that warns that the SWIFT_WHOLE_MODULE_OPTIMIZATION is obsolete like so:

image 2017-12-13 at 4 33 41 pm


tgoyne:
I believe this is only obsolete in Xcode 9, and we still support building with 8.

geoffmacd:
@tgoyne makes sense

PR #5559, opened 2018-01-09
Update to latest core/sync/object store

tgoyne:


bdash:
It looks like there are linker errors due to duplicate symbols related to base64. I think this was addressed in sync v2.2.8.

PR #5560, opened 2018-01-10
Anonymous and Nickname Authentication

mbalex99:
Adds Support for Anonymous and Nickname Authentication for Sync

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,7 @@ x.x.x Release notes (yyyy-mm-dd)
 * Realm Object Server errors not explicitly recognized by the client are now reported to the application
   regardless.
 * Add support for JSON Web Token as a sync credential source.
-* Add support for Nickname Auth  as a sync credential source.
+* Add support for Nickname and Anonymous Auth as a sync credential source.
 
 3.0.2 Release notes (2017-11-08)
 =============================================================
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ x.x.x Release notes (yyyy-mm-dd)
 * Realm Object Server errors not explicitly recognized by the client are now reported to the application
   regardless.
 * Add support for JSON Web Token as a sync credential source.
+* Add support for Nickname Auth  as a sync credential source.
 
 3.0.2 Release notes (2017-11-08)
 =============================================================
--- a/Realm/RLMSyncCredentials.h
+++ b/Realm/RLMSyncCredentials.h
@@ -98,7 +98,7 @@ extern RLMIdentityProvider const RLMIdentityProviderNickname;
 /**
  Construct and return anonymous credentials
  */
-+ (instancetype)credentialsWithAnonymous;
++ (instancetype)anonymousCredentials;
     
 /**
  Construct and return credentials from a nickname
--- a/Realm/RLMSyncCredentials.h
+++ b/Realm/RLMSyncCredentials.h
@@ -103,7 +103,7 @@ extern RLMIdentityProvider const RLMIdentityProviderNickname;
 /**
  Construct and return credentials from a nickname
  */
-+ (instancetype)credentialsWithNickname:(NSString *)nickname isAdmin:(BOOL) isAdmin;
++ (instancetype)credentialsWithNickname:(NSString *)nickname isAdmin:(BOOL)isAdmin;
 
 /**
  Construct and return special credentials representing a token that can
--- a/Realm/RLMSyncCredentials.h
+++ b/Realm/RLMSyncCredentials.h
@@ -96,7 +96,7 @@ extern RLMIdentityProvider const RLMIdentityProviderNickname;
 + (instancetype)credentialsWithJWT:(NSString *)token;
 
 /**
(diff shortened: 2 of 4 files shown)


Realm/RLMSyncCredentials.m
```diff
@@ -69,6 +71,16 @@ + (instancetype)credentialsWithUsername:(NSString *)username
...
     return [[self alloc] initWithCustomToken:token provider:RLMIdentityProviderJWT userInfo:nil];
 }
+    
++ (instancetype)credentialsWithAnonymous {
+    return [[self alloc] initWithCustomToken:@"" provider:RLMIdentityProviderAnonymous userInfo:nil];
+}
+    
++ (instancetype)credentialsWithNickname:(NSString *)nickname isAdmin:(BOOL)isAdmin {
+    return [[self alloc] initWithCustomToken:nickname
+                                    provider:RLMIdentityProviderDebug
```
nirinchev: Should this be Nickname?

mbalex99:
Thanks fixed it!

CHANGELOG.md
```diff
@@ -11,6 +11,7 @@ x.x.x Release notes (yyyy-mm-dd)
 * Realm Object Server errors not explicitly recognized by the client are now reported to the application
   regardless.
 * Add support for JSON Web Token as a sync credential source.
+* Add support for Nickname Auth  as a sync credential source.
```
bdash: This should presumably mention anonymous credentials as well. There's also an extra space in the middle of this sentence.

Realm/RLMSyncCredentials.h
```diff
@@ -89,6 +95,16 @@ extern RLMIdentityProvider const RLMIdentityProviderJWT;
  */
 + (instancetype)credentialsWithJWT:(NSString *)token;
 
+/**
+ Construct and return anonymous credentials
+ */
++ (instancetype)credentialsWithAnonymous;
```
bdash: `credentialsWithAnonymous` doesn't make much sense. The `WithFoo` naming is used to label an argument. `anonymousCredentials` would be a better choice.

Realm/RLMSyncCredentials.h
```diff
@@ -89,6 +95,16 @@ extern RLMIdentityProvider const RLMIdentityProviderJWT;
...
 
+/**
+ Construct and return anonymous credentials
+ */
++ (instancetype)anonymousCredentials;
+    
+/**
+ Construct and return credentials from a nickname
+ */
++ (instancetype)credentialsWithNickname:(NSString *)nickname isAdmin:(BOOL) isAdmin;
```
bdash: There shouldn't be any whitespace between `(BOOL)` and the parameter name.

RealmSwift/Sync.swift
```diff
@@ -306,6 +306,16 @@ public struct SyncCredentials {
     public static func jwt(_ token: Token) -> SyncCredentials {
         return SyncCredentials(RLMSyncCredentials(jwt: token))
     }
+    
+    /// Initialize new credentials using a nickname.
+    public static func nickname(_ nickname: Token, _ isAdmin: Bool = false) -> SyncCredentials {
```
bdash: Why did you opt to omit the argument label for the `isAdmin` argument? It makes it unclear what the boolean argument at the call site means. I think it should be included unless there's a strong reason not to.

Shouldn't the nickname argument be of type `String`? I realize that `Token` is a type alias for `String` so it's sorta pedantic, but aren't we expecting the user to pass an arbitrary string of their choosing here rather than some form of token?

RealmSwift/Sync.swift
```diff
@@ -306,6 +306,16 @@ public struct SyncCredentials {
...
+        return SyncCredentials(RLMSyncCredentials(nickname:nickname, isAdmin: isAdmin))
```
bdash: The linter is complaining about the space that is missing after the `:` here.

RealmSwift/Sync.swift
```diff
@@ -306,6 +306,16 @@ public struct SyncCredentials {
     public static func jwt(_ token: Token) -> SyncCredentials {
         return SyncCredentials(RLMSyncCredentials(jwt: token))
     }
+
```
bdash: The linter is complaining about the trailing whitespace on this line.

RealmSwift/Sync.swift
```diff
@@ -306,6 +306,16 @@ public struct SyncCredentials {
     public static func jwt(_ token: Token) -> SyncCredentials {
         return SyncCredentials(RLMSyncCredentials(jwt: token))
     }
+    
+    /// Initialize new credentials using a nickname.
+    public static func nickname(_ nickname: Token, _ isAdmin: Bool = false) -> SyncCredentials {
+        return SyncCredentials(RLMSyncCredentials(nickname:nickname, isAdmin: isAdmin))
+    }
+
```
bdash: The linter is complaining about the trailing whitespace on this line.

RealmSwift/Sync.swift
mbalex99: No you're right. It should be String, non-nil.

PR #5564, opened 2018-01-11
Add runloop checks to the permissions API

tgoyne:
Attempting to fetch or modify permissions from a background thread would just silently fail to ever call the callback. This makes them throw an exception like RLMResults/etc. do.

### DIFF
--- a/Realm/ObjectServerTests/RLMPermissionsAPITests.m
+++ b/Realm/ObjectServerTests/RLMPermissionsAPITests.m
@@ -984,6 +984,14 @@ static NSURL *makeTildeSubstitutedURL(NSURL *url, RLMSyncUser *user) {
     CHECK_PERMISSION_ABSENT(results, p);
 }
 
+- (void)testRetrievingPermissionsChecksThreadHasRunLoop {
+    [self dispatchAsyncAndWait:^{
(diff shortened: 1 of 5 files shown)


Realm/ObjectServerTests/RLMPermissionsAPITests.m
```diff
@@ -984,6 +984,14 @@ - (void)testSettingUnownedRealmPermission {
     CHECK_PERMISSION_ABSENT(results, p);
 }
 
+- (void)testRetrievingPermissionsChecksThreadHasRunLoop {
+    [self dispatchAsyncAndWait:^{
+        RLMAssertThrowsWithReason([self.userA retrievePermissionsWithCallback:^(__unused RLMResults *r, __unused NSError *e) {
+            XCTFail(@"callback should ot have been invoked");
```
bdash: “ot”

Realm/RLMSyncUser.mm
```diff
@@ -338,7 +339,14 @@ - (void)retrieveInfoForUser:(NSString *)providerUserIdentity
 
 #pragma mark - Permissions API
 
+static void verifyInRunLoop() {
+    if (!RLMIsInRunLoop()) {
+        @throw RLMException(@"Can only access or modify permissions from a thread which has a run loop (by default, only the main thread)`.");
```
bdash: There looks to be some extraneous punctuation at the end of the message.

PR #5566, opened 2018-01-11
Strip trailing slashes from the auth server URL

### LINKED ISSUE #5565: `Sync.retrievePermissions(callback:)` fails without error message if login url contains trailing `/`
@dmead28 commented on [Thu Jan 11 2018](https://github.com/realm/realm-core/issues/2977)

### 1. Goals

`Sync.retrievePermissions(callback:)` does not work when the login url has a trailing `/`, such as: `URL(string: "http://localhost:9080/")`. The callback is never called. The goal is to either:

    A) validate the URL and pass an error if it's invalid
    B) Strip the trailing `/`

### 2. Expected results

    A) `Sync.retrievePermissions(callback:)` returns an error if the url is not formatted properly
    B) `Sync.retrievePermissions(callback:)` succeeds whether the format is: "http://localhost:9080/" or "http://localhost:9080"

### 3. Actual results

`Sync.retrievePermissions(callback:)` fails and completion is never called.

### 4. Steps to reproduce

1) Log in using a url with a trailing `/`, such as `"http://localhost:9080/"`
2) Make a call to `Sync.retrievePermissions(callback:)`
3) The call will not return

### 5. Code sample that highlights the issue (full project that we can compile ourselves are ideal)
```
        let syncCredentials = SyncCredentials.usernamePassword(username: "test", password: "abc123", register: false)

        guard let loginUrl = URL(string: "http://localhost:9080/") else {
            print("Could not construct loginUrl")
            return
        }

        SyncUser.logIn(with: syncCredentials, server: loginUrl) { (newUserResult, error) in
            guard let newUser = newUserResult else {
                print("Error getting new user. Error: \(error?.localizedDescription ?? "")")
                return
            }

            // Permissions on main explicitly (suggested in docs)
            DispatchQueue.main.async { 
                newUser.retrievePermissions(callback: { (permissionResults, error) in
                    // THIS IS NEVER CALLED
                    guard let permissions = permissionResults else {
                        print("Could not get permissions. Error: \(error?.localizedDescription ?? "")")
                        return
                    }

                    print("Realms I have permission to")
                    permissions.forEach({ (permission) in
                        print("  - \(permission)")
                    })
                })
            }
```

Also, see test project at: https://github.com/dmead28/RosTest

### 6. **Version of Realm
Realm (3.0.2)
RealmSwift (3.0.2)

tgoyne:
Fixes #5565.

Not sure that this is a better solution than just throwing an exception. This is a bit more friendly, but has the potential for confusion by differing from other bindings. Doing this at the object store level would fix that but is pretty awkward due to that it's usually just passing a const ref onto something else, and it's harder to be sure that every place that would need this behavior was actually modified.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ x.x.x Release notes (yyyy-mm-dd)
 * Realm Object Server errors not explicitly recognized by the client are now reported to the application
   regardless.
 * Add support for JSON Web Token as a sync credential source.
+* Automatically strip trailing slashes from auth server URLs rather than behaving oddly.
 
 ### Bugfixes
 
--- a/Realm/ObjectServerTests/RLMPermissionsAPITests.m
+++ b/Realm/ObjectServerTests/RLMPermissionsAPITests.m
@@ -992,6 +992,23 @@ static NSURL *makeTildeSubstitutedURL(NSURL *url, RLMSyncUser *user) {
     }];
 }
 
+- (void)testRetrievePermissionsWithTrailingSlash {
+    // Have A open a Realm and grant a permission to B.
+    NSURL *url = REALM_URL();
+    NSString *tildeSubstitutedPath = [makeTildeSubstitutedURL(url, self.userA) path];
+    __attribute__((objc_precise_lifetime)) RLMRealm *r = [self openRealmForURL:url user:self.userA];
+    id p1 = [[RLMSyncPermission alloc] initWithRealmPath:tildeSubstitutedPath
+                                                identity:self.userB.identity
+                                             accessLevel:RLMSyncAccessLevelRead];
+    APPLY_PERMISSION_WITH_MESSAGE(p1, self.userA, @"Setting r permission for user B should work.");
+
+    // Add a trailing slash to the auth URL and try to fetch this new permission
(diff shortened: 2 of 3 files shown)


bdash:
One problem with this approach is that the auth server URL is stored in the sync metadata Realm and is used as a key when looking up users. This change would mean that users created before this change would not be found when looking them up with the same auth server URL in a version that contains this change.

tgoyne:
I think those users wouldn't have worked correctly anyway? I guess I need to look more into what things do and don't work correctly with trailing slashes.

bdash:
I think they'd have worked correctly for everything but the object store's permissions APIs, which appear to be the only place that constructs URLs derived from the auth server URL.

bdash:
And if we're going to require that `SyncUser`'s always work with auth servers without the trailing /, I'd suggest that doing it at the object store level is probably a more correct way to do things anyway. We could then perform a migration to strip the trailing / from any persisted metadata, removing the concern I noted above.

nirinchev:
As a point of reference, the .NET binding doesn't strip slashes but uses the built-in `Uri` class to construct URLs, so we've accepted both formats of the url (with or without trailing slash) but because the `.ToString` implementation of `Uri` ensures that the url ends with a slash, we've always sent OS urls with a trailing slash.

PR #5567, opened 2018-01-12
Improved partial sync API

bdash:
This is a WIP that I'm publishing for sake of visibility and early feedback.

The Swift API is close to @tgoyne's proposal. The Objective-C API is similar, but uses native KVO for notifying users of changes to the partial sync state. KVO's usability isn't great from Objective-C, so we may want to do something different here. It was just the easiest thing to throw together.

* Should unsubscribing live on the Realm, or the Results? It seems like it needs to live on the Realm in order to allow unsubscribing based only on the name. It's a bit unsymmetrical though, since subscribing lives on Results.
* How should errors raised by partial sync, if any, be exposed?
   With the Swift API they could potentially be passed to the callback as the payload of an enum value. That's not possible with the current approach in Objective-C.


tgoyne:
`realm.subscribe(results)` seems like a reasonable way to resolve the asymmetry with unsubscribe. Implementation-wise it is an operation on the Realm, and exposing it as such in the API may make the side effects (i.e. it refreshing the realm since it needs to perform a write) less weird?

bdash:
Alternatively, we could avoid the need to refresh the Realm by performing the write transaction on a separate copy of the Realm (for instance, on a background thread, which may make it possible to subscribe without blocking on another thread that is actively writing).

dhmspector:
In trying to get this set up for use in the `my-first-app` iOS TODO tutorial I ran into the following:

If I just specified the RealmSwift pod using
```
pod 'RealmSwift', :git => 'https://github.com/realm/realm-cocoa', :branch => 'mar/partial-sync', :submodules => true
```
The pod installs and builds without error, however, Results objects did not have a subscription method.  As per @bdash's suggestion I forced cocoapods to build both Realm and RealmSwift with 

```
  pod 'RealmSwift', :git => 'https://github.com/realm/realm-cocoa', :branch => 'mar/partial-sync', :submodules => true
  pod 'Realm', :git => 'https://github.com/realm/realm-cocoa', :branch => 'mar/partial-sync', :submodules => true
```

This time the compilation of the Realm pod fails with a preprocessor error: `'partial_sync.hpp' file not found` in RLMResults.mm

dhmspector:
As per a use in the forums, when trying to compile the RelmSwift module after a successful compile of the Realm module the, an error occurs here:

```
<module-includes>:1:9: note: in file included from <module-includes>:1:
#import "Headers/Realm.h"
        ^
/Users/spector/Library/Developer/Xcode/DerivedData/ToDo-dbvoejisiwkcjwarfsyauvuolqfb/Build/Products/Debug-iphoneos/Realm/Realm.framework/Headers/Realm.h:38:9: error: 'Realm/RLMSyncSubscription.h' file not found
#import <Realm/RLMSyncSubscription.h>
        ^
/Users/spector/Work/Realm/my-first-realm-app/ios/PartialSync/Pods/RealmSwift/RealmSwift/Aliases.swift:20:8: error: could not build Objective-C module 'Realm'
import Realm
       ^
```

The code is included from the partial sync branch as with the previous report:
```
  pod 'RealmSwift', :git => 'https://github.com/realm/realm-cocoa', :branch => 'mar/partial-sync', :submodules => true
  pod 'Realm', :git => 'https://github.com/realm/realm-cocoa', :branch => 'mar/partial-sync', :submodules => true
```

cc @bdash

bdash:
@dhmspector, I think that'll be fixed by 879c1a3.

Realm/RLMSyncSubscription.h
```diff
@@ -0,0 +1,133 @@
...
+
+/**
+ The error associated with this subscription, if any.
+
+ Will be non-nil only when `state` is `RLMSyncSubscriptionStateError`.
+ */
+@property (nonatomic, readonly, nullable) NSError *error;
+
+/**
+ Remove this subscription.
```
tgoyne: This could use a description of what happens when a subscription is removed. A rough take:

Removing a subscription will delete all objects from the local Realm which were matched only by that subscription and not any remaining subscriptions. This happens asynchronously following contact with the server, and has no immediate impact on the contents of the local Realm.

Realm/RLMSyncSubscription.mm
```diff
@@ -0,0 +1,108 @@
...
+#import "RLMRealm_Private.hpp"
+#import "RLMResults_Private.hpp"
+#import "RLMUtil.hpp"
+
+#import "sync/partial_sync.hpp"
+
+using namespace realm;
+
+@interface RLMSyncSubscription ()
+- (instancetype)initWithName:(NSString *)name results:(realm::Results&)results realm:(RLMRealm *)realm;
```
tgoyne: Some redundant realm::s

Realm/RLMSyncSubscription.mm
```diff
@@ -0,0 +1,108 @@
...
+@end
+
+@implementation RLMSyncSubscription {
+    partial_sync::SubscriptionNotificationToken _token;
+    util::Optional<partial_sync::Subscription> _subscription;
+    RLMRealm *_realm;
+}
+
+- (instancetype)initWithName:(NSString *)name results:(realm::Results&)results realm:(RLMRealm *)realm
+{
```
tgoyne: obj-c code doesn't use newlines before {

PR #5593, opened 2018-01-26
Add support for overriding table and column names

tgoyne:
Get _realmObjectName back into a working state and fix the things that never actually worked with it (such as links and arrays), and add _realmColumnNames to allow overriding the column names used.

This is needed for object-level permissions, which involves exposing some Realm objects in both obj-c and Swift. This is functionality we need to have in the public API eventually (for the same reasons as we need it internally), but for now it's internal-only to reduce the potential complications and to punt on things like public documentation and sensible error-handling for invalid schemas using it.

Currently sorting and distincting on mapped property names doesn't really work (you have to use the column name), which probably won't be an issue for the permissions types. All other operations (including querying) should work correctly.

### DIFF
--- a/Realm/RLMObjectSchema.mm
+++ b/Realm/RLMObjectSchema.mm
@@ -29,6 +29,7 @@
 #import "RLMUtil.hpp"
 
 #import "object_store.hpp"
+#import "object_schema.hpp"
 
 using namespace realm;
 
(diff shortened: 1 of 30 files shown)


Realm/RLMObjectSchema.mm
```diff
@@ -29,6 +29,7 @@
 #import "RLMUtil.hpp"
 
 #import "object_store.hpp"
+#import "object_schema.hpp"
```
bdash: May as well add these in sorted order while we're adding a new include.

PR #5609, opened 2018-02-09
Declare dependency on Security.framework

jgavris:
`keychain_helper.cpp` depends on `Security.framework`. Users may want to build Realm with Clang modules disabled `-fno-modules` because they are using `ccache` which doesn't support `-fmodules`. When modules are disabled, auto-linking of framework modules is also disabled, so this dependency needs to be expressed explicitly.

Resolves https://github.com/realm/realm-cocoa/issues/5645

### DIFF
--- a/Realm.podspec
+++ b/Realm.podspec
@@ -67,6 +67,7 @@ Pod::Spec.new do |s|
                               'Realm/ObjectStore/src/util/*.cpp',
                               'Realm/ObjectStore/src/util/apple/*.cpp'
 
+  s.frameworks              = 'Security'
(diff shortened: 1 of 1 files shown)


jgavris:
Did I break something? I don't have read permissions on Jenkins.

jgavris:
@bdash could you recommend a reviewer for this?

jgavris:
👋

bdash:
My apologies for the delay in getting to this. Thanks for the fix.

jgavris:
Thanks!

PR #5611, opened 2018-02-09
Add class definitions and wrappers for object-level permissions

tgoyne:
Still very work in progress.


tgoyne:
I've now split off the integration tests (which don't work yet) to a separate branch and finished documenting the initial minimal API, so I guess this is now reviewable. Once the object store branch is in a mergeable state this too should be mergeable without breaking anything unrelated.

bdash:
It looks like the object server tests on CI are hitting the problems with permissions tests that I mentioned earlier on Slack. In #5567 I put an `#if 0` around the tests for now for sake of keeping tests green until we can work around those problems.

RealmSwift/Sync.swift
```diff
@@ -666,3 +666,378 @@ extension Results where Element == SyncPermission {
...
+ */
+public class Permission: Object {
+    /// The Role which this Permission applies to. All users within the Role are
+    /// granted the permissions specified by the fields below any
+    /// objects/classes/realms which use this Permission.
+    ///
+    /// This property cannot be modified once set.
+    @objc dynamic public var role: PermissionRole?
+
+    @objc dynamic public var canRead = false
```
bdash: `sh build.sh verify-docs` is complaining about the lack of documentation on these properties.

RealmSwift/Sync.swift
```diff
@@ -666,3 +666,378 @@ extension Results where Element == SyncPermission {
...
+
+ A role named "everyone" is automatically created in new Realms, and all new users which
+ connect to the Realm are automatically added to it. Any other roles you wish to use are
+ managed as normal Realm objects.
+ */
+public class PermissionRole: Object {
+    @objc dynamic public var name = ""
+    let users = List<PermissionUser>()
+
+    @objc override public class func _realmObjectName() -> String {
```
bdash: `sh build.sh verify-docs` is complaining about the lack of documentation on these methods, too. I guess a `/// :nodoc:` would suffice to silence it.

Realm/RLMSyncPermission.h
```diff
@@ -16,7 +16,250 @@
...
+@property (nonatomic) RLMArray<RLMPermission *><RLMPermission> *permissions;
+
++ (instancetype)objectInRealm:(RLMRealm *)realm;
+@end
+
+/**
+ An object which describes class-wide permissions.
+
+ An instance of this object is automatically created in the Realm for class in your schema,
+ and should not be created manually. Call +[RLMClassPermission objectInRealm:forClassNamed:]`
```
bdash: The \` appears to be unbalanced here and on the next line.

Realm/RLMSyncPermission.h
```diff
@@ -16,7 +16,250 @@
...
+ `RLMObjectPrivileges` for details about what each of the properties mean when applied to
+ that type.
+ */
+@interface RLMPermission : RLMObject
+/// The Role which this Permission applies to. All users within the Role are
+/// granted the permissions specified by the fields below any
+/// objects/classes/realms which use this Permission.
+@property (nonatomic) RLMPermissionRole *role;
+
+@property (nonatomic) bool canRead;
```
bdash: Should we be using `BOOL` rather than `bool` since this is Objective-C? Is `bool` even available in Objective-C by default without including `stdbool.h`?… I guess it must be, since this all appears to compile.

Realm/RLMSyncPermission.h
```diff
@@ -16,7 +16,250 @@
...
+ current User is a member of, obtained by calling `-[RLMRealm privilegesForRealm]` on
+ the Realm.
+
+ By default, all operations are permitted, and each privilege field indicates an operation
+ which may be forbidden.
+ */
+struct RLMRealmPrivileges {
+    /// If `false`, the current User is not permitted to see the Realm at all. This can
+    /// happen only if the Realm was created locally and has not yet been synchronized.
+    bool read : 1;
```
bdash: Ditto about `BOOL` vs `bool`.

Realm/RLMSyncPermission.h
```diff
@@ -16,7 +16,250 @@
...
+    /// governed by `setPermissions`.
+    bool update : 1;
+
+    /// If `false`, deleting this object is not permitted. Write transactions which delete
+    /// the object can be performed locally, but the server will restore it.
+    ///
+    /// It is possible to have `update` but not `delete` privileges, or vice versa. For
+    /// objects with primary keys, `delete` but not `update` is ill-advised, as an object
+    /// can be updated by deleting and recreating it.
+    bool del : 1;
```
bdash: I assume this is `del` to remain compatible with Objective-C++ where `delete` is a keyword? Something like `remove` or even `delete_` might be clearer?

Realm/RLMSyncPermission.h
tgoyne: Foundation.h pulls in stdbool.h even on platforms where BOOL isn't a typedef for bool.

PR #5637, opened 2018-02-27
Add some integration tests and helpers for object-level permissions

tgoyne:
Still pretty far from complete coverage of the functionality, but these tests pass and are better than nothing.


Realm/ObjectServerTests/RLMPermissionsAPITests.m
```diff
@@ -19,7 +19,6 @@
 #import <XCTest/XCTest.h>
 
 // FIXME: Many permission tests appears to fail with the ROS 3.0.0 alpha releases.
```
bdash: Is this still true? If not, we should remove the FIXME as well.

tgoyne: Yes, most of the tests for the old permissions API fail. The files are now not compiled at all.

RealmSwift/Sync.swift
```diff
@@ -881,7 +881,7 @@ public class PermissionRole: Object {
     /// The name of the Role
     @objc dynamic public var name = ""
     /// The users which belong to the role
-    let users = List<PermissionUser>()
+    public let users = List<PermissionUser>()
```
bdash: There's a similar oversight on `PermissionUser.roles`.

Issue #5653, opened 2018-03-06
Default Sync Realm API


bigfish24:
As part of the upcoming release of partial sync, Realm Platform Self-Hosted and Cloud will offer the concept of a default synced Realm. This is designed to simplify how developers start working with synced data. Specifically it will fix or improve the following:

1. Simplify how developers think about their data model. Previously developers needed to split their data into multiple Realms to control data access. Now with partial sync and the new permission system this is not necessary.
2. Mirror the ease of use of Realm Database which offers a similar default Realm convenience API
3. Limit the code needed to open a synced Realm - now the user simply needs to login with just the auth URL instead of needing to worry about Realm URLs and paths
4. The default synced Realm will include the master User list of the application allowing developers to incorporate relationships to users simplifying how to think about user metadata.

API Proposal:

```
/*
  Open the default synced Realm
  - Has a dedicated API so that we can have a default configuration using:
    - SyncUser.current
    - Default Realm URL based off SyncUser auth URL (supplied within the refresh token)
    - Hardcoded path
    - Setup the Realm with a sync history type

  - If opened before there is a SyncUser.current one of two things can happen depending on the implementation:
    - The simple implementation is to throw if there is no SyncUser.current
    - The advanced implementation is to open the Realm without a sync session allowing offline access
*/

// Java
Realm realm = Realm.getDefaultSyncInstance();

// Swift
let realm = try! Realm.defaultSync()

// Objective-C
RLMRealm *realm = [RLMRealm defaultSyncRealm];

// JS
// Currently lacks a default Realm, so we should add
var realm = new Realm.defaultSync()

// .Net
var realm = Realm.GetDefeaultSyncInstance();

/*
  Login Example
*/

let credentials = SyncCredentials.nickname(nickname: "Adam", isAdmin: false)
let serverURL = NSURL(“http://testapp.us1.cloud.realm.io”)
SyncUser.logIn(with: credentials,
               server: serverURL) { user, error in
    if let user = user {
        // can now open the default synchronized Realm with this user
        let realm = try! Realm.defaultSync()
    } else if let error = error {
        // handle error
    }
}

```

tgoyne:
I think our ideal API would look something like:

    // 1. Open the Realm in offline mode, and begin syncing with the first user to log in
    Realm.Configuration.defaultConfiguration = SyncConfiguration.automatic()
    let realm = try! Realm()
    SyncUser.logIn(with: credentials, server: serverURL) { ... }

[...]

PR #5658, opened 2018-03-07
Implement SyncConfiguration.automatic and friends

bdash:
This implements the items 2 and 3 of the proposal at https://github.com/realm/realm-cocoa/issues/5653#issuecomment-370973732.

One thing that feels a little weird about this is having a method on `SyncConfiguration` that returns a `Realm.Configuration`. The method looks like a factory method, but returns a different type.

There are no tests yet as I think it needs a newer ROS to ensure the Realm ends up with permissions that allow the user to modify it.


cmelchior:
Any reason the wording is `automaticConfiguration` instead of `defaultConfiguration`?
Maybe it is just me, but the first is a bit unusual while using `default` feels more normal ?

bdash:
The only remaining work I see here is adding more tests, which I'll get to after I find some lunch. @tgoyne, would you mind reviewing the changes so far when you have a chance?

bdash:
Tests and change log entry added.

PR #5667, opened 2018-03-12
Fix warnings seen when building with Xcode 9.3

bdash:


tgoyne:
I'm not particularly a fan of riddling self-> all over the place rather than just disabling the warning.

bdash:
It was only a small number of places, so it seemed easier to just go with it than worry about disabling the warning.

bdash:
The CI failures are due to a Jenkins upgrade that appears to have broken a plug-in our PR job relies on. Since this change could easily introduce problems for older Xcode versions, I'd like to see CI succeed before we merge this.

bdash:
I disabled the warning that otherwise required inserting `self->`.

PR #5674, opened 2018-03-15
Create the Carthage archives as part of the releasability pipeline, rather than during the release process

bdash:
This makes it easy to run the `carthage build` steps in parallel on different nodes, and should decrease how long it takes for releases to be made.


bdash:
The increases the runtime of the releasability pipeline by < 5 minutes, while knocking nearly 20 minutes off the release job. Seems like a great tradeoff.

PR #5677, opened 2018-03-15
Add some permissions convenience methods

bdash:
These helped to simplify the permissions code in the object permissions demo app a little. For instance:
```diff
-    let realmPermissions = realm.objects(RealmPermission.self).first!
-    let everyonePermission = realmPermissions.permissions.findOrCreate(forRoleNamed: "everyone")
+    let everyonePermission = realm.permissions.findOrCreate(forRoleNamed: "everyone")
```

And:
```diff
-        let classPermissions = realm.object(ofType: ClassPermission.self, forPrimaryKey: cls.className())!
-        let everyonePermission = classPermissions.permissions.findOrCreate(forRoleNamed: "everyone")
+        let everyonePermission = realm.permissions(forType: cls).findOrCreate(forRoleNamed: "everyone")
```

One thing I opted to do was document that these methods are only applicable to partially-synced Realms, since otherwise they'd have to return an optional, making them slightly less pleasant to use. I'm not sure if that's the best choice.

### DIFF
--- a/RealmSwift/Sync.swift
+++ b/RealmSwift/Sync.swift
@@ -1274,6 +1274,39 @@ extension Realm {
     public func getPrivileges(forClassNamed className: String) -> ClassPrivileges {
         return ClassPrivileges(rawValue: RLMGetComputedPermissions(rlmRealm, className))
     }
+
+    /**
+    Returns the class-wide permissions for the given class.
+
+     - parameter cls: An Object subclass to get the permissions for.
+     - returns: The class-wide permissions for the given class.
+     - requires: This must only be called on a partially-synced Realm.
+    */
+    public func permissions<T: Object>(forType cls: T.Type) -> List<Permission> {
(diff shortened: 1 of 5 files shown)


RealmSwift/Sync.swift
```diff
@@ -1274,6 +1274,41 @@ extension Realm {
     public func getPrivileges(forClassNamed className: String) -> ClassPrivileges {
         return ClassPrivileges(rawValue: RLMGetComputedPermissions(rlmRealm, className))
     }
+
+    /**
+    Returns the class-wide permissions for the given class.
+
+    This will return nil for non-partial-sync Realms.
```
tgoyne: This doesn't seem to be true?

bdash: Oops, I apparently only removed that from one of the two functions when I changed my mind on this point.

Issue #5685, opened 2018-03-21
hardcoded values


greggjaskiewicz:
Hi, 
This has recently caused me some grief. I am using realm inside my framework. The framework is written in ObjectiveC. Because of the nature of the language, I have to add a prefix to all RLM keywards. 
This - I hope - is the only place where magic numbers are used to represent length of a name prefix. 
I'd suggest type += strlen(NamePrefix); instead of hardcoding a number please. 

https://github.com/realm/realm-cocoa/blob/fe3bd185c286227fdf334def32af9cea1fd7e8ca/Realm/RLMProperty.mm#L143

PR #5686, opened 2018-03-21
Don't hardcode string lenghts.

greggjaskiewicz:
Fixes potential issues when classes get renamed. See issue #5685


bmunkholm:
Is this still relevant to you @greggjaskiewicz? Not sure how this would be sufficient? Wouldn't you still need to change the actual prefix?

greggjaskiewicz:
> Is this still relevant to you @greggjaskiewicz? Not sure how this would be sufficient? Wouldn't you still need to change the actual prefix?

I don't work for the company who was using this code anymore, but I'd image that this is still relevant as a potential issue.

PR #5691, opened 2018-03-22
Exclude classes that are not subclasses of `Object` from `customSchema`.

kishikawakatsumi:
The following code crashes the following exception:

>  Could not cast value of type 'RLMPermission' to 'RealmSwift.Object'.

```swift
let realm = try! Realm()
print(realm.configuration.objectTypes)
```

Because `objectTypes` contains classes derived from Objective-C Realm as default schama if `objectTypes` is not set.

### DIFF
--- a/RealmSwift/RealmConfiguration.swift
+++ b/RealmSwift/RealmConfiguration.swift
@@ -195,7 +195,7 @@ extension Realm {
                 self.customSchema = newValue.map { RLMSchema(objectClasses: $0) }
             }
             get {
-                return self.customSchema.map { $0.objectSchema.map { $0.objectClass as! Object.Type } }
+                return self.customSchema.map { $0.objectSchema.flatMap { $0.objectClass as? Object.Type } }
             }
         }
 


tgoyne:
Thanks!

RealmSwift/RealmConfiguration.swift
```diff
@@ -195,7 +195,7 @@ extension Realm {
                 self.customSchema = newValue.map { RLMSchema(objectClasses: $0) }
             }
             get {
-                return self.customSchema.map { $0.objectSchema.map { $0.objectClass as! Object.Type } }
+                return self.customSchema.map { $0.objectSchema.flatMap { $0.objectClass as? Object.Type } }
```
bdash: This introduces a compiler warning when building with Swift 4 in Xcode 9.3:
```
realm-cocoa/RealmSwift/RealmConfiguration.swift:198:64: warning: 'flatMap' is deprecated: Please use compactMap(_:) for the case where closure returns an optional value
                return self.customSchema.map { $0.objectSchema.flatMap { $0.objectClass as? Object.Type } }
                                                               ^
realm-cocoa/RealmSwift/RealmConfiguration.swift:198:64: note: use 'compactMap(_:)' instead
                return self.customSchema.map { $0.objectSchema.flatMap { $0.objectClass as? Object.Type } }
                                                               ^~~~~~~
                                                               compactMap
```

Issue #3099, opened 2016-01-13
objectForPrimaryKey


pjebs:
I am using Realm-ObjC in a mixed project.
When my obj-c model is bridged to swift, the `objectForPrimaryKey` is not visible. It is visible in Obj C world.

There are other functions that don't bridge over to swift but the provided `RLMSupport.swift` makes them visible.
Can you please add it to `RLMSupport.swift` Helper because  `objectForPrimaryKey` is very useful.

kishikawakatsumi:
@pjebs Objective-C factory methods get mapped as convenience initializers in Swift automatically. So `objectForPrimaryKey:` method is converted to `RLMObject(forPrimaryKey:)`. 

https://developer.apple.com/library/ios/documentation/Swift/Conceptual/BuildingCocoaApps/InteractingWithObjective-CAPIs.html

If `Book` model class is defined like the following:

```
@interface Book : RLMObject
@property NSString *identifier;
@end
```

You can use `Book(forPrimaryKey: "...")` initializer instead `objectForPrimaryKey:` method.

PR #5700, opened 2018-03-26
Expose the factory methods for query objects by primary key.

SandyChapman:
This ticket seems to suggest this was possible at one point, but possibly is not longer working for Swift 4.0:
https://github.com/realm/realm-cocoa/issues/3099

For the sake of this commit msg, consider our RLMObject subclass:

```
@objcMembers
class Order: RLMObject {
    dynamic var orderId: String = ""
    override class func primaryKey() -> String {
        return "orderId"
    }
}
```

Currently, the only way to retrieve an object by primary key on an RLMObject subclass in Swift (4.0), is to do this:

```
let order = Order.objects(in: realm, where: "orderId = %@", "someOrderId").firstObject()
```

This is because the RLMObject factory method `+ objectInRealm:forPrimaryKey` is only available on the base RLMObject class as a convenience initializer. In other words, we get this if we try to use the converted init on the type Order:

```
Cannot invoke initializer for type 'Order' with an argument list of type '(in: RLMRealm, forPrimaryKey: String)'
```

In order to be able to call this factory method in Swift, we need to keep it as a factory method and not convert it to a init. To do this, we add the NS_SWIFT_NAME macro which then allows us to call the method as expected.

```
let order = Order.object(in: realm, forPrimaryKey: order.orderId) // Yay! It works!
```

### DIFF
--- a/Realm/RLMObject.h
+++ b/Realm/RLMObject.h
@@ -374,7 +374,7 @@ NS_ASSUME_NONNULL_BEGIN
  @return    An object of this object type, or `nil` if an object with the given primary key does not exist.
  @see       `-primaryKey`
  */
-+ (nullable instancetype)objectForPrimaryKey:(nullable id)primaryKey;
++ (nullable instancetype)objectForPrimaryKey:(nullable id)primaryKey NS_SWIFT_NAME(object(forPrimaryKey:));
 
 
 #pragma mark - Querying Specific Realms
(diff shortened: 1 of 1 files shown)


astigsen:
Any reason that you cannot use `Realm.object(ofType:forPrimaryKey:)`?

API ref: https://realm.io/docs/swift/3.3.0/api/Classes/Realm.html#/s:FC10RealmSwift5Realm6objectu0_RxCS_6ObjectrFT6ofTypeMx13forPrimaryKeyq__GSqx_

SandyChapman:
@astigsen : Note that this is using the Obj-C Realm APIs within Swift. I'm not using RealmSwift in this project.

tgoyne:
This appears to have been broken ever since Swift 3.0.

PR #5702, opened 2018-03-26
Add a minimal test for RLMObject.object(forPrimaryKey:)

tgoyne:


Realm/Tests/Swift/SwiftObjectInterfaceTests.swift
```diff
@@ -313,6 +313,16 @@ class SwiftObjectInterfaceTests: RLMTestCase {
         try! realm.commitWriteTransaction()
     }
 
+    func testObjectForPrimaryKey() {
+        let realm = RLMRealm.default()
+        realm.beginWriteTransaction()
```
nirinchev: Should that transaction be committed or rolled back (or does it happen in a non-obvious way)?

fealebenpae: Good point - when the realm goes out of scope and is destroyed that should get rid of the transaction, and closing the transaction in one way or another isn't really the point of this test, but it might not hurt to explicitly close it like in the test above.

CHANGELOG.md
```diff
@@ -1,3 +1,21 @@
...
+* None.
+
+### Enhancements
+
+* Expose `RLMObject.object(forPrimaryKey:)` as a factory method for Swift so
+  that it is callable with recent versions of Swift.
+
+### Bugfixes
+
+* Exclude the RLMObject-derived Permissions classes from the types repored by
```
fealebenpae: I think this entry is here by mistake?

Realm/Tests/Swift/SwiftObjectInterfaceTests.swift
tgoyne: Yeah, committing or cancelling the transaction is semantically unnecessary, but we log a warning if the Realm goes out of scope while in a transaction so we end the transaction in tests just to avoid the log spam.

CHANGELOG.md
tgoyne: That was added by #5691 which was missing a changelog entry.

PR #5705, opened 2018-03-27
Cancel pending waits when fetching a refresh token fails

### LINKED ISSUE #5635: `Realm.asyncOpen` callback not called
## Goals
Open the realm with `Realm.asyncOpen`

## Expected Results
The callback is called - either with success or an error

## Actual Results
The callback is never called. Also, no log messages.

## Steps to Reproduce
Logged in via `SyncUser.logIn(with: credentials, server: authUrl)`. With the returned user, and the other config data (the sync server url, added below, and an array with my object types), create a `Realm.Configuration` and then with that call `Realm.asyncOpen`.

This started happening after I switched to Realm Cloud beta - with the previous local ROS it was working correctly. I, too, just upgraded to Realm 3.1.1 (was 3.0.2 before) to see if the problem was my version being outdated, but the same problem still happens.

When I say "switched to Realm Cloud beta" I mean only that I updated the auth and sync urls.
The urls I have now are:
Sync url: realms://myapp.us1.cloud.realm.io
Auth url: https://myapp.us1.cloud.realm.io

```
ProductName:	Mac OS X
ProductVersion:	10.12.6
BuildVersion:	16G1036

Xcode 9.2
Build version 9C40b

/usr/local/bin/pod
1.3.1
Realm (3.1.1)
RealmSwift (3.1.1)
RealmSwift (~> 3.1.1)

/bin/bash

/usr/local/bin/carthage
0.25.0
(not in use here)

/usr/bin/git
git version 2.14.3 (Apple Git-98)
```

tgoyne:
If we get a non-transient error and we aren't going to retry fetching the access token later, switch the session back to Inactive and notify any pending `asyncOpen` calls that they were cancelled rather than just hanging forever.

Fixes #5635.

### DIFF
--- a/Realm/ObjectServerTests/RLMObjectServerTests.mm
+++ b/Realm/ObjectServerTests/RLMObjectServerTests.mm
@@ -1451,6 +1451,22 @@
     }
 }
 
+- (void)testDownloadCancelsOnAuthError {
+    RLMSyncUser *user = [self logInUserForCredentials:[RLMObjectServerTests basicCredentialsWithName:NSStringFromSelector(_cmd)
+                                                                                            register:self.isParent]
+                                               server:[RLMObjectServerTests authServerURL]];
+    auto c = [RLMRealmConfiguration defaultConfiguration];
+    c.syncConfiguration = [[RLMSyncConfiguration alloc] initWithUser:user realmURL:[NSURL URLWithString:@"realm://localhost:9080/invalid"]];
+    auto ex = [self expectationWithDescription:@"async open"];
+    [RLMRealm asyncOpenWithConfiguration:c callbackQueue:dispatch_get_main_queue()
+                                callback:^(RLMRealm *realm, NSError *error) {
+                                    XCTAssertNil(realm);
+                                    XCTAssertNotNil(error);
+                                    [ex fulfill];
+    }];
+    [self waitForExpectationsWithTimeout:2.0 handler:nil];
+}
+
 #pragma mark - Validation
 
 - (void)testCompactOnLaunchCannotBeSet {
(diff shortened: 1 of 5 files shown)


Realm/ObjectServerTests/RLMObjectServerTests.mm
```diff
@@ -1451,6 +1451,22 @@ - (void)testDownloadWhileOpeningRealm {
     }
 }
 
+- (void)testDownloadCancelsOnAuthError {
+    RLMSyncUser *user = [self logInUserForCredentials:[RLMObjectServerTests basicCredentialsWithName:NSStringFromSelector(_cmd)
+                                                                                            register:self.isParent]
+                                               server:[RLMObjectServerTests authServerURL]];
+    auto c = [RLMRealmConfiguration defaultConfiguration];
+    c.syncConfiguration = [[RLMSyncConfiguration alloc] initWithUser:user realmURL:[NSURL URLWithString:@"realm://localhost:9080/invalid"]];
```
nirinchev: I'm likely missing something, but why is that path invalid?

fealebenpae: The user shouldn't have permissions to open a realm at that path.

nirinchev: Ah... right, I was trying to relate the test to the other changes, but I guess they are not directly related.

tgoyne: A permission denied error hits the new code added in RLMSyncSessionRefreshHandle.mm. The new error code used to be relevant to this, but it turns out that with ROS 3.0 leaving out the realm path gives you the default realm so it's not actually an error.

PR #5712, opened 2018-04-02
Convert `flatMap` → `compactMap` for Swift 4.1+ users

irace:
Convert `flatMap` → `compactMap` for Swift 4.1+ users.


bmunkholm:
Has been fixed through another PR.

PR #5721, opened 2018-04-07
Correction to key generation in Encryption Example

osianSmith:
This is a little change but when reviewing the Encryption Example, I found that current key method was just producing NSData of just 64 0's as its key. This was also discovered on a test I did of this method (not committed) 

Although that this is a key and that the rest of the code looks like its correct, I am concerned with the non random key generation. This is because I feel for developers new to encrypting their realms they might have looked at this code and think this is correct and that it generates a random key. There is also a risk that a developer running late on a deadline may just copy this code into production code. 

I feel that this tutorial does not need to explain what is going on here that a non-unique key is being generated, and the importance of why you want a random unique key. This just means that if someone was to get the hands on on the .realm file it would be very easy to decrypt - a brute force would find this issue immediately.

This commit just adds a function that I have used in other projects to generate a unique key and a few comments to suggest to me what it is doing. 

Thanks 
Osian

### DIFF
--- a/examples/ios/swift/Encryption/ViewController.swift
+++ b/examples/ios/swift/Encryption/ViewController.swift
@@ -107,9 +107,8 @@ class ViewController: UIViewController {
         }
 
         // No pre-existing key from this application, so generate a new one
-        let keyData = NSMutableData(length: 64)!
-        let result = SecRandomCopyBytes(kSecRandomDefault, 64, keyData.mutableBytes.bindMemory(to: UInt8.self, capacity: 64))
-        assert(result == 0, "Failed to get random bytes")
+        let keyData = generateNewKey()
+		
 
         // Store the key in the keychain
         query = [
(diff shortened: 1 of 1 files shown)


tgoyne:
The existing code successfully generates a random key for me when I run it on an iPhone 6S running iOS 11.

osianSmith:
During my testing it was not actually unique - if you ran a couple of times you would see while yes it would create a value. if you were to ask it again it would then then just return the same number in the key as 64's 0

tgoyne:
If you are not clearing the keychain between runs (which merely deleting the app does not do), then it will return the same value every time, because that's the entire point.

tgoyne:
If you are able to construct a consistent repro case for SecRandomCopyBytes() failing to work on a non-jailbroken device then you should report that to Apple as that would be a critical OS-level security flaw.

bmunkholm:
Closing due to lack of repro.

PR #5740, opened 2018-04-23
Add password reset wrappers to RLMSyncUser

tgoyne:
I started writing tests for this and it turned out that proper e2e tests will be a bit involved (we'll need to set up a dummy email handler which itself requires switching away from `ros start`), so currently they don't really test much of anything.


dependencies.list
```diff
@@ -1,4 +1,4 @@
 VERSION=3.4.0
 REALM_CORE_VERSION=5.4.2
 REALM_SYNC_VERSION=3.0.1
```
bmunkholm: Perhaps upgrade to Core 5.5.0 and Sync 3.1.0

Realm/RLMSyncUser.h
```diff
@@ -219,6 +219,88 @@ NS_SWIFT_UNAVAILABLE("Use the full version of this API.");
...
+
+ If `email` is an email address which is associated with a user account that was
+ registered using the "password" authentication service, the server will send an
+ email to that address with a password reset token. No error is reported if the
+ email address is invalid or not associated with an account.
+
+ @param serverURL  The authentication server URL for the user.
+ @param email      The email address to send the email to.
+ @param completion A block which will be called when the request completes or
+                   failes. The callback will be invoked on a background queue
```
bmunkholm: fails

Realm/RLMSyncUser.h
```diff
@@ -219,6 +219,88 @@ NS_SWIFT_UNAVAILABLE("Use the full version of this API.");
...
+
+ @warning Changing a user's password using an authentication server that doesn't
+          use HTTPS is a major security flaw, and should only be done while
+          testing.
+
+ @param serverURL   The authentication server URL for the user.
+ @param token       The one-time use token from the URL.
+ @param newPassword The user's new password.
+ @param completion  A block which will be called when the request completes or
+                    failes. The callback will be invoked on a background queue
```
bmunkholm: fails

PR #5758, opened 2018-05-07
 Allow setting the shouldCompactOnLaunch block for sync realms

tgoyne:


tgoyne:
We don't really have any binding-level tests that try to verify that query serialization/deserialization works; just some basic tests to verify that the actual binding code works.

ironage:
`= nil` is also tested in core, so it should be fine. Just wanted to mention it in case it might be easy to add.

PR #5795, opened 2018-06-01
Make Query-based sync the default mode  

cmelchior:
#5795 This PR makes query-based the default mode for synced Realms

TODO
- [x] Fix changelog
- [ ] Fix unit tests
- [x] Cleanup

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,25 @@ x.x.x Release notes (yyyy-MM-dd)
 
 * None.
 
+3.7.0 Release notes (YYYY-MM-DD)
+=============================================================
+
+The feature known as Partial Sync has been renamed to Query-based Synchronization. This
+has impacted a number of API's. See below for the details.
+
+###  Deprecated
+
+* `-[RLMSyncConfiguration automaticConfiguration] has been deprecated in favor of `-[RLMSyncUser defaultConfiguration]. 
+* `-[RLMSyncConfiguration automaticConfigurationForUser] has been deprecated in favor of `-[RLMSyncUser defaultConfiguration].
+* `-[RLMSyncConfiguration isPartial] has been deprecated in favor of `-[RLMSyncConfiguration fullSynchronization]`.
+
+
+### Enhancements
+
+* Added `-[RLMSyncUser createConfiguration:url]`. Query-based sync is the default sync mode for this configuration.
+* Added `-[RLMSyncUser defaultConfiguration]`. Query-based sync is the default sync mode for this configuration.
+
+
 3.6.0 Release notes (2018-05-29)
 =============================================================
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,19 +27,19 @@ has impacted a number of API's. See below for the details.
 
 ###  Deprecated
 
-* `+[RLMSyncConfiguration initWithUser] has been deprecated in favor of `-[RLMSyncUser configurationWithUrl:url]. 
+* `+[RLMSyncConfiguration initWithUser] has been deprecated in favor of `-[RLMSyncUser configurationWithURL:url].
 * `+[RLMSyncConfiguration automaticConfiguration] has been deprecated in favor of `-[RLMSyncUser configuration]. 
 * `+[RLMSyncConfiguration automaticConfigurationForUser] has been deprecated in favor of `-[RLMSyncUser configuration].
 * `-[RLMSyncConfiguration isPartial] has been deprecated in favor of `-[RLMSyncConfiguration fullSynchronization]`.
 
 ### Enhancements
 
-* Added `-[RLMSyncUser configurationWithUrl:url]`. Query-based sync is the default sync mode for this configuration.
+* Added `-[RLMSyncUser configurationWithURL:url]`. Query-based sync is the default sync mode for this configuration.
 * Added `-[RLMSyncUser configuration]`. Query-based sync is the default sync mode for this configuration.
 
 * Add `-[RLMRealm syncSession]` and  `Realm.syncSession` to obtain the session
   used for a synchronized Realm.
-* Added `-[RLMSyncUser configurationWithUrl:url]`. Query-based sync is the default sync mode for this configuration.
+* Added `-[RLMSyncUser configurationWithURL:url]`. Query-based sync is the default sync mode for this configuration.
 * Added `-[RLMSyncUser configuration]`. Query-based sync is the default sync mode for this configuration.
 
 ### Bugfixes
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,10 +15,15 @@ x.x.x Release notes (yyyy-MM-dd)
 * None.
 
 3.7.0 Release notes (YYYY-MM-DD)
+x.x.x Release notes (yyyy-MM-dd)
 =============================================================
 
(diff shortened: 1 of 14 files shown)


CHANGELOG.md
```diff
@@ -1,3 +1,21 @@
+3.7.0 Release notes (YYYY-MM-DD)
```
tgoyne: This should be part of the existing changelog section below.

Realm/ObjectServerTests/RLMSyncTestCase.mm
```diff
@@ -344,13 +344,10 @@ - (RLMRealm *)immediatelyOpenRealmForURL:(NSURL *)url
...
-    auto syncConfig = [[RLMSyncConfiguration alloc] initWithUser:user realmURL:url];
-    syncConfig.stopPolicy = stopPolicy;
-    syncConfig.isPartial = [self isPartial];
-
-    auto c = [RLMRealmConfiguration defaultConfiguration];
-    c.encryptionKey = encryptionKey;
-    c.syncConfiguration = syncConfig;
+    auto c = [user createConfiguration:url];
+    c.encryptionKey = encryptionKey;  
+    c.syncConfiguration.stopPolicy = stopPolicy;
```
tgoyne: This doesn't work; the `syncConfiguration` property does not return an object which mutates the RLMRealmConfiguration it's obtained from.

Realm/RLMSyncConfiguration.h
```diff
@@ -75,21 +85,21 @@ NS_ASSUME_NONNULL_BEGIN
                 contain the wildcard marker `~`, which will automatically be filled in with
                 the user identity by the Realm Object Server.
  */
-- (instancetype)initWithUser:(RLMSyncUser *)user realmURL:(NSURL *)url;
+- (instancetype)initWithUser:(RLMSyncUser *)user realmURL:(NSURL *)url __attribute__((deprecated("Use 'RLMSyncUser.createConfiguration' instead")));
```
tgoyne: "Use '-[RLMSyncUser createConfiguration:]' instead."

Realm/RLMSyncUser.h
```diff
@@ -145,6 +153,17 @@ NS_ASSUME_NONNULL_BEGIN
...
 
+/**
+ Create a sync configuration instance.
+ 
+ @param url     The unresolved absolute URL to the Realm on the Realm Object Server, e.g.
+ `realm://example.org/~/path/to/realm`. "Unresolved" means the path should
+ contain the wildcard marker `~`, which will automatically be filled in with
+ the user identity by the Realm Object Server.
+ */
+-(RLMRealmConfiguration *)createConfiguration:(NSURL *)url
```
tgoyne: The Cocoa name for this is `configurationWithURL:`.

I was expecting this to return a RLMSyncConfiguration as `automaticConfiguration` did. Returning a RLMRealmConfiguration makes modifying the other RLMSyncConfiguration properties very awkward.

Realm/RLMSyncUser.h
```diff
@@ -111,6 +112,13 @@ NS_ASSUME_NONNULL_BEGIN
  */
 @property (nonatomic, readonly) RLMSyncUserState state;
 
+/**
+ Returns the default configuration for the user. The default configuration points to the default
+ query-based Realm on the server the user authenticated against.
+*/
+@property (nonatomic, readonly) RLMRealmConfiguration *defaultConfiguration NS_REFINED_FOR_SWIFT;
```
tgoyne: This should just be a method named `configuration`.

RealmSwift/Sync.swift
```diff
@@ -482,6 +504,28 @@ extension SyncUser {
...
+     
+     `enableSSLValidation` is true by default. It can be disabled for debugging
+     purposes.
+     
+     - warning: The URL must be absolute (e.g. `realms://example.com/~/foo`), and cannot end with
+     `.realm`, `.realm.lock` or `.realm.management`.
+     
+     - warning: NEVER disable SSL validation for a system running in production.
+     */
+    public func createConfiguration(realmURL: URL, fullSynchronization: Bool = false, enableSSLValidation: Bool = true, urlPrefix: String? = nil) -> Realm.Configuration {
```
tgoyne: This should just be named `configuration` and have URL be optional rather than a separate `defaultConfiguration` function.

Returning a `SyncConfiguration` struct works better with how configurations are currently constructed.

Realm/RLMSyncUser.h
cmelchior: Ok, I wasn't 100% how this was used. I'll make it return a `RLMSyncConfiguration` instead.

Realm/RLMSyncUser.h
cmelchior: Shouldn't it just be deleted if we rename the `createConfiguration` to `configuration` ?

tgoyne: In obj-c we'd need both `configuration` and `configurationWithURL:` because there's no default parameters, but in Swift they can be a single function.

cmelchior: One concern I have about this is that `automatic()` currently returns `Realm.Configuration`. If we want to still have a sensible default is it then common in Swift to have a property named `configuration` that is default and a method named `configuration()` that can be used to customize it?

RealmSwift/Sync.swift
tgoyne: Huh, I thought `automatic()` returned a `SyncConfiguration`, but on second thought of course it doesn't. Given that, I guess it's correct for this to return a `Realm.Configuration` as well.

Realm/RLMSyncUser.h
tgoyne: Objc-c:
```
- (RLMRealmConfiguration *)configuration;
- (RLMRealmConfiguration *)configurationWithURL:(NSURL *)realmURL;
```

The developer would write `user.configuration` or `[user configuration]` to get the default Realm, or `[user configurationWithURL:realmURL]` to set the URL.

Swift:
```
public func configuration(realmURL: URL? = nil, fullSynchronization: Bool = false, enableSSLValidation: Bool = true, urlPrefix: String? = nil) -> Realm.Configuration
```

The developer would call `user.configuration()` for the default realm, or `user.configuration(realmURL: url)` to customize it. The property would just let them drop the `()`, so there's no real point.

cmelchior:
Ready for review. There are still some failing tests, some looks entirely unrelated (osx debug), others I'm not sure about (osx-object-server).

In particular, I'm uncertain about https://github.com/realm/realm-cocoa/pull/5795/files#diff-316d20300d435ba8e37f48d511faf8b8R347 and https://github.com/realm/realm-cocoa/pull/5795/files#diff-c64398ac8b77c37143815b54e83e9427R638

If this is not the correct way to do it. I'm not certain what the correct approach would be.

tgoyne:
Setting properties on the syncConfiguration object of a pre-existing RLMRealmConfiguration has to be done as:

```
RLMRealmConfiguration *config = ...;
RLMSyncConfiguration *syncConfig = config.syncConfiguration;
syncConfig.property = newValue;
config.syncConfiguration = syncConfig;
```

This is because the RLMSyncConfiguration object is just a plain data class, and the logic around actually setting the objectstore config fields is done in the setSyncConfiguration setter.

For the tests you can probably just use the four-argument version of the factory method.

PR #5806, opened 2018-06-08
Compare ascending properly in SortDescriptor comparison operator

artem-alexandrov:
It's a very small PR with a simple fix of a wrong comparison method. 
It was comparing left hand side with itself, now it compares correctly.

I've also added a unit test to cover this method.


artem-alexandrov:
Anyone?

tgoyne:
Thanks!

PR #5815, opened 2018-06-12
Update to sync 3.5.3

tgoyne:


tgoyne:
CocoaPods failures appear to be just a bug in the old version of cocoapods we currently have installed on the CI machines. It's unclear why it's just now failing, but it's seemingly unrelated to the changes here.

PR #5828, opened 2018-06-24
adding support for Swift Package Manager

heyzooi:
Environment:

```
ProductName:	Mac OS X
ProductVersion:	10.13.5
BuildVersion:	17F77

Xcode 9.4.1
Build version 9F2000

/usr/local/bin/pod
1.5.3
(not in use here)

/bin/bash
GNU bash, version 3.2.57(1)-release (x86_64-apple-darwin17)

/usr/local/bin/carthage
0.29.0
(not in use here)

/usr/bin/git
git version 2.15.2 (Apple Git-101.1)
```

Build Parameters:

```
swift build -Xcxx -IRealm/ObjectStore/src -Xcxx -Isync-3.5.6/include -Xcxx -DREALM_ENABLE_SYNC=1
```


heyzooi:
`Realm.h` was duplicated because it needs to add the header `RLMVersion.h`

tgoyne:
This doesn't work when building from a clean source directory as it relies on files downloaded and generated by some of the build steps in the Xcode project, and the changes to the source files would break some of our other pre-existing packages.

heyzooi:
hi @tgoyne,

which files are downloaded? as far I know the downloaded files are from the ObjectStore submodule, so you need to clone the repo and clone the submodules as well

which files are generated? as far I know the generated files are `RLMPlatform.h` and `RLMVersion.h` right? Those could be generated by the Swift Package Manager as well, but we would need to set `--disable-sandbox` during the `swift build` call, so the files can be generated, just as a Xcode build step.

which changes are breaking existing packages? are you talking about the replace from `#import <Realm/HeaderFile.h>` to `#import "HeaderFile.h"`? if so, how could we make this work for all packages?

I believe add support for Swift Package Manager could be an important step for Realm.
Let's make this work!

bmunkholm:
We would really like to support it, but it's still not on the top of our priority list, unfortunately. This PR still have a way to go and if others are able to finish it that would, of course, be very welcomed.

heyzooi:
hi @bmunkholm, thanks for your response. Could you please be more specific in what is missing when you said the PR still have a way to go?

bmunkholm:
@heyzooi - as @tgoyne mentions there are issues with the current approach that need to be resolved. Unfortunately, @tgoyne doesn't have the time to address those questions just now. Just wanted to let you know that.

AFcgi:
How does it look like with this PR?

Luccifer:
so?

ukushu:
.package(url: "https://github.com/realm/realm-cocoa.git", from: "4.3.0" )

PR #5900, opened 2018-08-30
Add support for SSL certificate pinning

tgoyne:
Currently only the simple case of pinning to a specific certificate file is supported and not more complex things such as an arbitrary callback because the sync library does not implement those in the SecureTransport backend.

Fixes https://github.com/realm/realm-cocoa-private/issues/402.


tgoyne:
This isn't mergeable yet as the sync side of things doesn't work, but this PR should be done and is now reviewable.

Realm/ObjectServerTests/RLMObjectServerTests.mm
```diff
@@ -1805,4 +1805,175 @@ - (void)testPartialSync {
...
+    RLMSyncManager.sharedManager.pinnedCertificatePaths = @{@"localhost": certificateURL(@"not-localhost.cer")};
+
+    RLMSyncUser *user = [self logInUserForCredentials:[RLMSyncTestCase basicCredentialsWithName:NSStringFromSelector(_cmd)
+                                                                                       register:YES]
+                                               server:[RLMSyncTestCase authServerURL]];
+    [self verifyOpenFails:[user configurationWithURL:[NSURL URLWithString:@"realms://localhost:9443/~/default"]]];
+}
+
+- (void)DISABLE_testOpenSecureRealmWithMissingCertFile {
+    // FIXME: this currently crashes inside the sync library
```
cmelchior: Is there a sync issue tracking this?

Realm/RLMSyncConfiguration.h
```diff
@@ -39,6 +39,20 @@ NS_ASSUME_NONNULL_BEGIN
...
+ if the file contains a CA cert) will be accepted when initiating a connection
+ to a server. This prevents certain certain kinds of man-in-the-middle (MITM)
+ attacks, and can also be used to trust a self-signed certificate which would
+ otherwise be untrusted.
+
+ On macOS, the file may be in any of the formats supported by SecItemImport(),
+ including PEM and .cer (see SecExternalFormat for a complete list of possible
+ formats). On iOS and other platforms, only DER .cer files are supported.
+ */
+@property (nonatomic, nullable) NSURL *pinnedCertificateURL;
```
cmelchior: Is there a reason you can set this property? Since any certificate validation should be enabled for both auth calls and sync, it seems a bit odd you can set it on both the `SyncManager` and the `SyncConfiguration` as it just opens up for mismatching config settings. Shouldn't the configuration just inherit it from the `SyncManager` and then make this readonly?

RealmSwift/Sync.swift
```diff
@@ -176,6 +176,37 @@ public typealias SyncLogLevel = RLMSyncLogLevel
...
+/**
+ * How the SSL certificate for the Realm Object Server should be validated.
+ *
+ * By default, when connecting to the Realm Object Server over HTTPS, Realm will
+ * validate the server's HTTPS certificate using the system trust store and root
+ * certificates. For additional protection against man-in-the-middle (MITM)
+ * attacks and similar vulnerabilities, you can pin a certificate or public key,
+ * and reject all others, even if they are signed by a trusted CA.
+ */
+public enum ServerValidationPolicy {
```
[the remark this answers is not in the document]
tgoyne: This controls how the client validates that the server is who it claims to be. `.none` doesn't disable SSL; it just disables validation of the certificate (which is a pre-existing option).

Realm/RLMSyncConfiguration.h
tgoyne: This is automatically set to the one used for auth when getting a configuration from a user, but it's possible to have the sync server and auth servers on different domains, which requires different pinned certificates. `RLMSyncConfiguration` should probably be entirely immutable like the Swift version is, but for now it has mutable fields.

CHANGELOG.md
```diff
@@ -7,7 +7,8 @@ x.x.x Release notes (yyyy-MM-dd)
 
 ### Enhancements
 
-* None.
+* Add support for pinning SSL certificates used for https and realms
```
bmunkholm: Are docs updated to reflect this? (or a PR pending)

RealmSwift/Sync.swift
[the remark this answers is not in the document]
cmelchior: Good point, I was thinking about this backwards.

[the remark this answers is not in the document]
tgoyne: Probably worth rewording the docs a bit to make sure that's clear, since I guess it doesn't outright say it's controlling how the client validates the server.

CHANGELOG.md
tgoyne: Docs PR: https://github.com/realm/realm.io/pull/4569

PR #5907, opened 2018-09-06
Add support for limiting the number of objects in a sync subscription

tgoyne:
Proper support for limit on Results probably requires breaking changes on the Swift side (we already support slicing a `Results` using the standard library functionality, but with somewhat different semantics and different types involved), so this just punts on the issue and only implements applying a limit specifically to a sync subscription.

I also fleshed out the documentation for the existing functionality a bit while I was looking at it.


CHANGELOG.md
```diff
@@ -1,3 +1,19 @@
+x.x.x Release notes (yyyy-MM-dd)
+=============================================================
+
+### Breaking Changes
+
+* None.
+
+### Enhancements
+
+* Add support for limiting the number of objects matched by a query-based sync
```
bmunkholm: What happens if you run a server with an older ROS?

Realm/RLMSyncSubscription.h
```diff
@@ -115,6 +115,14 @@ typedef NS_ENUM(NSInteger, RLMSyncSubscriptionState) {
 /**
  Subscribe to the query represented by this `RLMResults`.
 
+ Subscribing to a query asks the server to synchronize all objects to the
+ client which match the query, along with all objects which are reachable
+ from those objects via links. This happens asynchronously, and the local
+ client Realm may not immediately have all objects which match the query.
+ Observer the `state` property of the returned subscription object to be
```
bmunkholm: Observer -> Observing ?

Realm/RLMSyncSubscription.h
```diff
@@ -126,13 +134,60 @@ typedef NS_ENUM(NSInteger, RLMSyncSubscriptionState) {
...
```
bmunkholm: Observe

Realm/RLMSyncSubscription.h
```diff
@@ -126,13 +134,60 @@ typedef NS_ENUM(NSInteger, RLMSyncSubscriptionState) {
...
+- (RLMSyncSubscription *)subscribeWithName:(nullable NSString *)subscriptionName;
+
+/**
+ Subscribe to a subset of the query represented by this `RLMResults`.
+
+ Subscribing to a query asks the server to synchronize all objects to the
+ client which match the query, along with all objects which are reachable
+ from those objects via links. This happens asynchronously, and the local
+ client Realm may not immediately have all objects which match the query.
+ Observer the `state` property of the returned subscription object to be
```
bmunkholm: Observe

Realm/RLMSyncSubscription.h
```diff
@@ -126,13 +134,60 @@ typedef NS_ENUM(NSInteger, RLMSyncSubscriptionState) {
...
+ an object referring to the existing sync subscription. This means that
+ performing the same subscription twice followed by removing it once will
+ result in no subscription existing.
+
+ The number of top-level matches may optionally be limited. This limit
+ respects the sort and distinct order of the query being subscribed to,
+ if any. Please note that the limit does not count or apply to objects
+ which are added indirectly due to being linked to by the objects in the
+ subscription. If the limit is larger than the number of objects which
+ match the query, all objects will be included.
```
bmunkholm: Perhaps note that there is currently no "offset" functionality.

CHANGELOG.md
cmelchior: Query-based sync will throw an error saying LIMIT isn't supported.

RealmSwift/Sync.swift
```diff
@@ -905,7 +905,13 @@ public class SyncSubscription<Type: RealmCollectionValue> {
 
     /// Observe the subscription for state changes.
     ///
-    /// When the state of the subscription changes, `block` will be invoked and passed the new state.
+    /// When the state of the subscription changes, `block` will be invoked and
+    /// passed the new state.
+    ///
+    /// The token returns from this function does not hold a strong reference to
```
bmunkholm: returned

RealmSwift/Sync.swift
```diff
@@ -935,19 +941,38 @@ public class SyncSubscription<Type: RealmCollectionValue> {
 extension Results {
     /// Subscribe to the query represented by this `Results`
     ///
-    /// The subscription will not have an explicit name.
+    /// Subscribing to a query asks the server to synchronize all objects to the
+    /// client which match the query, along with all objects which are reachable
+    /// from those objects via links. This happens asynchronously, and the local
+    /// client Realm may not immediately have all objects which match the query.
+    /// Observer the `state` property of the returned subscription object to be
```
bmunkholm: Observe

bmunkholm:
Looks good. Do you have a PR for the docs as well?
We need to mention the dependency on ROS in the docs and at the limit functions.

CHANGELOG.md
bmunkholm: That should then be documented in the methods.

nirinchev: I would rather add a compatibility section in the changelog than document stuff in the API docs.

tgoyne: I think it makes sense to mention it in the API docs for these specific methods because we aren't bumping the minimum version of ROS supported; you just can't use these specific functions with older ones.

Realm/RLMSyncSubscription.h
tgoyne: While we tend to think about offset and limit together, there's nothing in the API that makes it look like you could have offsets and it just isn't implemented yet.

Issue #5791, opened 2018-05-30
Distinct(by:) returns wrong values


1int:
Here's my data:
<img width="952" alt="screen shot 2018-05-30 at 19 27 07" src="https://user-images.githubusercontent.com/1285300/40733919-8d68babc-643f-11e8-8afe-ae26e2e50e44.png">

There are 5 object, which have 3 different 'Estimates'.

Here's my query:
<code> self.inviati = self.realm.objects(QuoteRequest.self).filter("user.id=%s", Auth.userId!).sorted(byKeyPath: "timestamp", ascending: false).distinct(by: ["estimate.id"])</code>

I get 3 models returned, all of them have the same estimate.id

<img width="302" alt="screen shot 2018-05-30 at 19 33 57" src="https://user-images.githubusercontent.com/1285300/40734223-77386566-6440-11e8-80e6-8aaed65ae8f9.png">

bmunkholm:
Thanks for reporting @1int - we will check it.
---
Easy started task for anyone wanting to help. Start with creating a testcase that reproduce this.

qubblr:
https://github.com/realm/realm-cocoa/pull/5936 fixes this.

PR #5936, opened 2018-10-01
Create testcase that reproduces wrong distinction.

qubblr:
https://github.com/realm/realm-cocoa/issues/5791


tgoyne:
Thanks! With this repro case I was able to track down the problem and made a PR to fix it: https://github.com/realm/realm-core/pull/3102

PR #5952, opened 2018-10-09
Add some missing validation to property getters and setters

tgoyne:
6c02999 shifted around where the checks for valid threads and objects were done, and a few turned out to be wrong. This resulted in either crashing within core rather than throwing an exception, or throwing an untranslated C++ exception rather than the appropriate obj-c exception.

RLMArray and RLMResults already have tests similar to the ones added here.


tgoyne:
No, I noticed it myself while writing some other code.

PR #5953, opened 2018-10-12
Fix the json arguments of SyncUser.requestEmailConfirmation

nirinchev:
The `provider_id` should have been outside the `data` argument, which had caused ROS to silently ignore the request (as it was trying too hard to ensure that an attacker cannot figure out if an email is registered with the server).

Fixes https://github.com/realm/realm-object-server-private/issues/1283.

### DIFF
--- a/Realm/RLMSyncUser.mm
+++ b/Realm/RLMSyncUser.mm
@@ -371,7 +371,7 @@ PermissionChangeCallback RLMWrapPermissionStatusCallback(RLMPermissionStatusBloc
                                     userEmail:(NSString *)email
                                    completion:(RLMPasswordChangeStatusBlock)completion {
     [RLMSyncUpdateAccountEndpoint sendRequestToServer:serverURL
-                                                 JSON:@{@"provider_id": email, @"data": @{                                                                   @"action": @"request_email_confirmation"}}
+                                                 JSON:@{@"provider_id": email, @"data": @{@"action": @"request_email_confirmation"}}
                                               options:[[RLMSyncManager sharedManager] networkRequestOptions]
                                            completion:completion];
 }
--- a/Realm/RLMSyncUser.mm
+++ b/Realm/RLMSyncUser.mm
@@ -371,8 +371,7 @@ PermissionChangeCallback RLMWrapPermissionStatusCallback(RLMPermissionStatusBloc
                                     userEmail:(NSString *)email
                                    completion:(RLMPasswordChangeStatusBlock)completion {
     [RLMSyncUpdateAccountEndpoint sendRequestToServer:serverURL
-                                                 JSON:@{@"data": @{@"provider_id": email,
-                                                                   @"action": @"request_email_confirmation"}}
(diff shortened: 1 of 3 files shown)


tgoyne:
There's a [test for this which passes](https://github.com/realm/realm-cocoa/blob/master/Realm/ObjectServerTests/RLMObjectServerTests.mm#L478). Is the test wrong, or is this just updating things to work with a newer ROS version that broke existing clients? If it's the latter, the changelog needs to mention that version.

Realm/RLMSyncUser.mm
```diff
@@ -371,8 +371,7 @@ + (void)requestEmailConfirmationForAuthServer:(NSURL *)serverURL
                                     userEmail:(NSString *)email
                                    completion:(RLMPasswordChangeStatusBlock)completion {
     [RLMSyncUpdateAccountEndpoint sendRequestToServer:serverURL
-                                                 JSON:@{@"data": @{@"provider_id": email,
-                                                                   @"action": @"request_email_confirmation"}}
+                                                 JSON:@{@"provider_id": email, @"data": @{                                                                   @"action": @"request_email_confirmation"}}
```
tgoyne: excessive whitespace

nirinchev:
The ROS API has never changed, so there must be something wrong with the test. Unfortunately, I'm completely unfamiliar with the test setup in Cocoa, so it's very hard for me to understand what the test does - essentially, calling this API should generate a token and send it as an email at the ROS level - it's not clear to me if the Cocoa test framework has made any modifications to the ROS codebase to allow intercepting this token and passing it back to the test framework.

tgoyne:
The test server script passes ROS a [custom email handler](https://github.com/realm/realm-cocoa/blob/master/Realm/ObjectServerTests/test-ros-server.js#L32) which just write the token to a file. The test calls `requestEmailConfirmation`, reads the token from the file, and then passes it to `confirmEmail`. This all succeeds, which indicates that the current request made to ROS results in ROS invoking the email handler with a valid token. There's another test which calls `requestEmailConfirmation` with a user that doesn't exist and verifies that the email handler is not invoked.

nirinchev:
Ah... I think I know what' happening! ROS automatically "sends" an email confirmation email when the user registers, so I believe this is what the test is seeing and mistakenly believes that this was generated due to calling that method, when in reality it was there all along. I can modify the test slightly to assert that a token was generated immediately after logging the user in, then that a different one is generated after calling that method.

nirinchev:
@tgoyne I confirmed that this is the case. I also confirmed that the modified test fails without this fix - it fails on this line: https://github.com/realm/realm-cocoa/pull/5953/files#diff-c64398ac8b77c37143815b54e83e9427R497, because calling `[self emailForAddress:userName]` immediately after creating the user removes the token generated by ROS automatically.

nirinchev:
The failing tests seem to be unrelated to the changes in this PR, so I'll go ahead and merge it 🤞

PR #5958, opened 2018-10-18
Update to latest objectstore

### LINKED ISSUE #5791: Distinct(by:) returns wrong values

tgoyne:
This reimplements the deprecated partial-sync API on top of the non-deprecated one as the implementation for the deprecated one was removed. It just passes an empty RLMResults to the callback as that part was already broken and no one's complained.

Fixes #5936. Fixes #5791.


Realm/RLMRealm+Sync.mm
```diff
@@ -19,37 +19,64 @@
...
-    partial_sync::register_query(_realm, className.UTF8String, query.UTF8String, std::move(cb));
+        switch (state->subscription.state()) {
+            case partial_sync::SubscriptionState::Invalidated:
+            case partial_sync::SubscriptionState::Pending:
+            case partial_sync::SubscriptionState::Creating:
+                return;
+
+            case partial_sync::SubscriptionState::Error:
+                try {
+                    rethrow_exception(state->subscription.error());
```
ironage: is this just a fancy way to pass the stack trace to the callback?

PR #6007, opened 2018-11-16
Add support for setting the user agent for sync requests

tgoyne:
I couldn't figure out how to write a meaningful test for this as the only actual result is a change in ROS's log output, but I did manually verify that those were updated as expected. An example of the log output with this:

2018-11-16T19:10:36.464Z - info: [sync] Sync Connection[327]: Session[1]: Client info: (path='/testClassRead/__partial/3deca6d376853ed7c14837d89e3194ee/c57d830d4f57760fa30b7ba2d8c1caa2256230af', user='3deca6d376853ed7c14837d89e3194ee', from=[127.0.0.1]:51573, protocol=25) RealmSync/3.13.3 (macOS Darwin 16.7.0 Darwin Kernel Version 16.7.0: Wed Oct 10 20:06:00 PDT 2018; root:xnu-3789.73.24~1/RELEASE_X86_64 x86_64)  Realm Swift/3.11.2 -[SwiftPermissionsAPITests testClassRead]

(`-[SwiftPermissionsAPITests testClassRead]` is the user-set application identifier, which I set to the test name for testing purposes)


kspangsege:
Looks like this does not follow the intended syntax: https://github.com/realm/realm-sync/blob/770e814e7f136950bec9648a0d5c45c1034e9c75/src/realm/sync/client.hpp#L89-L113.

Not critical, just mentioning it.

tgoyne:
How does it differ?

kspangsege:
> How does it differ?

Either leave it alone, or read the syntax rules, or tell me if they are not intelligible.

cmelchior:
```
[SwiftPermissionsAPITests testClassRead]
```

Should have been `(SwiftPermissionsAPITests testClassRead)` or `([SwiftPermissionsAPITests testClassRead])` according to syntax suggestion in the docs.

Realm/RLMSyncManager.mm
```diff
@@ -111,6 +115,7 @@ - (instancetype)initWithCustomRootDirectory:(NSURL *)rootDirectory {
         auto mode = should_encrypt ? SyncManager::MetadataMode::Encryption : SyncManager::MetadataMode::NoEncryption;
         rootDirectory = rootDirectory ?: [NSURL fileURLWithPath:RLMDefaultDirectoryForBundleIdentifier(nil)];
         SyncManager::shared().configure(rootDirectory.path.UTF8String, mode, "", none, true);
```
cmelchior: Minor, but the ObjectStore API is designed so the `""` here should have been the binding part since it never changes. Object Store then combines the binding and application parts.

Instead, you construct binding+application part in the `userAgent` method.

Realm/RLMSyncManager.mm
```diff
@@ -123,6 +128,17 @@ - (NSString *)appID {
...
 
+- (void)setUserAgent:(NSString *)userAgent {
+    bool isSwift = !!NSClassFromString(@"RealmSwiftObjectUtil");
+    auto fullUserAgent = [[NSMutableString alloc] initWithFormat:@"Realm %@/%@",
+                          isSwift ? @"Swift" : @"Objective C", REALM_COCOA_VERSION];
+    if (userAgent.length > 0) {
+        [fullUserAgent appendFormat:@" %@", userAgent];
+    }
+    SyncManager::shared().set_user_agent(RLMStringDataWithNSString(fullUserAgent));
+    _userAgent = userAgent;
```
cmelchior: Is this used anywhere?

tgoyne: The property getter returns it.

tgoyne:
> Should have been (SwiftPermissionsAPITests testClassRead) or ([SwiftPermissionsAPITests testClassRead]) according to syntax suggestion in the docs.

`-[SwiftPermissionsAPITests testClassRead]` is the name of the application, not additional details.

regalcinemas:
Cansel  please

Realm/RLMSyncManager.mm
cmelchior: The reason for splitting them is that we want to send the binding part irrespectively of the user setting the application part.

PR #6029, opened 2018-12-03
Add subscription management functions

### LINKED ISSUE #6010: Better subscription management

tgoyne:
This adds `-[RLMRealm subscriptions]` which returns a `RLMResults` of subscriptions and `-[RLMRealm subscriptionWithName:]` to look up a specific subscription, along with Swift wrappers for these. Rather than having a special thing using LIKE syntax as JS has, the `RLMResults` is simply filterable like any other collection.

The implementation of this is made awkward by that we already have a RLMSyncSubscription type that's not an RLMObject and doesn't behave like an RLMObject. Fortunately, obj-c is sufficiently dynamic that we can sometimes return the existing objectstore `Subscription`-backed type and sometimes return a new `RLMObjectBase`-backed type and have them behave (hopefully) identically.

Closes #6010.

### DIFF
--- a/Realm/RLMSyncSubscription.mm
+++ b/Realm/RLMSyncSubscription.mm
@@ -36,16 +36,68 @@ using namespace realm;
 @end
 
 @implementation RLMSyncSubscription {
-    @public
-    NSString *_name;
+    partial_sync::SubscriptionNotificationToken _token;
+    util::Optional<partial_sync::Subscription> _subscription;
+    RLMRealm *_realm;
 }
 
-- (instancetype)initPrivate {
-    return (self = [super init]);
+- (instancetype)initWithName:(NSString *)name results:(Results const&)results realm:(RLMRealm *)realm {
+    if (!(self = [super init]))
+        return nil;
+
+    _name = [name copy];
+    _realm = realm;
+    try {
+        _subscription = partial_sync::subscribe(results, name ? util::make_optional<std::string>(name.UTF8String) : util::none);
+    }
+    catch (std::exception const& e) {
+        @throw RLMException(e);
+    }
+    self.state = (RLMSyncSubscriptionState)_subscription->state();
+    __weak auto weakSelf = self;
+    _token = _subscription->add_notification_callback([weakSelf] {
+        auto self = weakSelf;
+        if (!self)
+            return;
+
+        // Retrieve the current error and status. Update our properties only if the values have changed,
+        // since clients use KVO to observe these properties.
+
+        if (auto error = self->_subscription->error()) {
+            try {
+                std::rethrow_exception(error);
+            } catch (...) {
+                NSError *nsError;
+                RLMRealmTranslateException(&nsError);
+                if (!self.error || ![self.error isEqual:nsError])
+                    self.error = nsError;
+            }
+        }
+        else if (self.error) {
+            self.error = nil;
+        }
+
+        auto status = (RLMSyncSubscriptionState)self->_subscription->state();
+        if (status != self.state) {
+            if (status == RLMSyncSubscriptionStateCreating) {
+                // If a subscription is deleted without going through this
+                // object's unsubscribe() method the subscription will transition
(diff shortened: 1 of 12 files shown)


Realm/RLMSyncSubscription.mm
```diff
@@ -18,38 +18,144 @@
...
+
+@interface RLMSyncSubscriptionObject : RLMObjectBase
+@end
+@implementation RLMSyncSubscriptionObject {
+    util::Optional<NotificationToken> _token;
+    Object _obj;
+}
+
+- (NSString *)name {
+    return _row.is_attached() ? RLMStringDataToNSString(_row.get_string(_row.get_column_index("name"))) : nil;
```
cmelchior: Are managed RealmObjects normally allowed to be accessed when the Realm is closed? In Java this would throw an exception.

Realm/RLMSyncSubscription.mm
```diff
@@ -18,38 +18,144 @@
...
+    objectType = objectType.substr(0, objectType.size() - strlen("_matches"));
+    return [NSString stringWithFormat:@"RLMSyncSubscription {\n\tname = %@\n\tobjectType = %@\n\tquery = %@\n\tstatus = %@\n\terror = %@\n}",
+            self.name, RLMStringDataToNSString(objectType),
+            RLMStringDataToNSString(_row.get_string(_row.get_column_index("query"))),
+            @(self.state), self.error];
+}
+
+- (void)unsubscribe {
+    if (_row) {
+        partial_sync::unsubscribe(Object(_realm->_realm, *_info->objectSchema, _row));
```
cmelchior: In Java we implemented this as a normal managed operation, i.e. it requires a write transaction and it basically just an alias for deleting the object. I'm not sure I can what semantics make the most sense for the Cocoa binding though.

Realm/RLMSyncSubscription.mm
```diff
@@ -83,18 +200,89 @@ - (void)unsubscribe {
...
+@end
 
+@implementation RLMSubscriptionResults
++ (instancetype)resultsWithRealm:(RLMRealm *)realm {
+    auto table = ObjectStore::table_for_object_type(realm->_realm->read_group(), partial_sync::result_sets_type_name);
+    if (!table) {
+        @throw RLMException(@"-[RLMRealm subscriptions] can only be called on a Realm using query-based sync");
+    }
+    // The server automatically adds a few subscriptions for the permissions
+    // types which we want to hide. They're just an implementation detail and
```
cmelchior: In Java we actually just exposes them. Mostly because these methods might also be used for debugging and not showing all subscriptions can also make it hard to reason about the system, especially if the API's are used by us internally. I do agree that if a user by accident deletes them they will probably be very surprised, but IMO the chance is pretty slim and will probably be detected pretty quick.

Realm/RLMSyncSubscription.mm
```diff
@@ -18,38 +18,144 @@
...
+    @public
+    NSString *_name;
+}
+
+- (instancetype)initPrivate {
+    return (self = [super init]);
+}
+
+- (void)unsubscribe {
+    __builtin_unreachable();
```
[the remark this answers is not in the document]
tgoyne: This is just an abstract class that's never instantiated. There isn't actually any reason for it to be, though (initially RLMSyncSubscriptionObject was a subclass of it, but that didn't work at all) and I think I can just re-merged RLMSyncSubscription and RLMNewSyncSubscription.

Realm/RLMSyncSubscription.mm
tgoyne: No, this class is specifically trying to act like the pre-existing RLMSyncSubscription type, which isn't a managed object.

Realm/RLMSyncSubscription.mm
tgoyne: This is something that's already part of the released API and not a new thing. I don't think the semantics we chose are great, but it's not bad enough to justify a breaking change to them.

PR #6041, opened 2018-12-18
Add Realm module map generation

matt:
**Goal**
Allow RealmSwift to be added as a static library via Cocoapods.

**Workarounds**
1. Add `use_modular_headers!` to the Podfile.
2. Add `pod 'Realm', :modular_headers => true` to the Podfile.

**Additional Reading**
[CocoaPods 1.5.0 - Swift Static Libraries :: *Modular Headers*](http://blog.cocoapods.org/CocoaPods-1.5.0/#modular-headers)

**Issue**
https://github.com/realm/realm-cocoa/issues/6042


tgoyne:
This looks like just a CocoaPods bug (the Realm pod defines a module using a custom modulemap and the validator just fails to check for that), but explicitly setting DEFINES_MODULE should be harmless as a workaround. This currently fails to build due to a syntax error, though.

matt:
Added the missing comma. I originally added the new key at the end, but realized the others were alphabetized.

PR #6076, opened 2019-02-04
Fix warnings when building with Xcode 10.2

tgoyne:


joshbuhler:
Is there a target date for when this merge might be added to Realm? Just wondering when we'll have support for Xcode 10.2 available.

PR #6095, opened 2019-03-06
Update to core 3.14.0 and sync 3.15.2

tgoyne:
The changes which require test updates should both be non-breaking; the notification one is specifically related to a change in the behavior of something not exposed in the public API, and the other is because something that used to be an error no longer is. The change in RLMRealm.mm is just to make it so that the test actually can create a subclass of RLMRealm.


CHANGELOG.md
```diff
@@ -1,11 +1,19 @@
 x.y.z Release notes (yyyy-MM-dd)
 =============================================================
 ### Enhancements
-* None.
+* Reduce memory usage when committing write transactions.
+* Improve performance of compacting encrypted Realm files.
+* Throw an exception when opening a Realm file with an invalid top ref rather
```
bmunkholm: don't think users know what a "top ref" is.

CHANGELOG.md
```diff
@@ -1,11 +1,19 @@
...
+  than crashing with an assertion failure.
 
 ### Fixed
 * Fix a memory leak whenever Realm makes a HTTP(s) request to the Realm Object
   Server (since 3.8.0, Issue [#6058](https://github.com/realm/realm-cocoa/issues/6058)).
+* Fix an assertion failure when creating an object after creating an object
```
bmunkholm: In JS we made this description: 
> Fixed an assertion failure when using a synchronized Realm and an object was created after another object was created with an int primary key of null. ([#3227](https://github.com/realm/realm-core/pull/3227))

CHANGELOG.md
```diff
@@ -1,11 +1,19 @@
...
+  with a null int primary key in the same write transaction.
+  ([#3227](https://github.com/realm/realm-core/pull/3227)).
+* Properly clean up sync sessions when the last Realm object using the session
```
bmunkholm: How could a user experience this? 
Issue or PR #?

CHANGELOG.md
bmunkholm: This is an `InvalidDatabase` C++ exception now. How is that propagated to end users? Could they catch it and do something meaningful like deleting the database and continue? This has been heavily requested in some issue i fI recall correctly?

CHANGELOG.md
tgoyne: No idea how a user would hit it or even notice. I discovered it was broken while writing tests for the audit functionality, so there's no issue and the PR has nothing to do with it.

CHANGELOG.md
tgoyne: It surfaces as an untyped exception without a unique error code. Without further SDK-level work there isn't a good way to handle just this exception.

CHANGELOG.md
bmunkholm: Then it should be in the ### internal section

PR #6103, opened 2019-03-18
fix a typo in the api usage.

### LINKED ISSUE #6102: Typo in Results.swift
`let results = realm.objects(Dog.self)`

results -> dogs

see 
https://github.com/realm/realm-cocoa/blob/master/RealmSwift/Results.swift#L340

P3PPP:
fix #6102


bmunkholm:
Thanks @P3PPP ! Instead of having you go through the hassle of signing CLA etc, I just fixed it in master. But thanks for fixing!

PR #6106, opened 2019-03-21
Use new async open functionality

tgoyne:


raudabaugh:
Any update on this PR?

We're using this successfully in v3.15.0-beta.2, but holding off on the v3.15.0 release until we get confirmation that it's been merged.

raudabaugh:
@ericjordanmossman said this PR might be delayed due to the stability issues we've had with using the new asyncOpen. Though I think for us, it happens so infrequently that we'd rather keep using it, and we don't want to fall too far behind the improvements in stable updates to the client. 

What do you think about including this in the next stable release not as default behavior, but as an opt-in flag (`useExperimentalAsyncOpen` or something like that)?

tgoyne:
We've decided to just go ahead and release this, so it'll be in the next stable version.

PR #6110, opened 2019-03-25
Update Sync.swift docs with deprecation warning.

bmunkholm:


tgoyne:
`- warning` is correct, and changing it to `@warning` will make the docs not render correct inside Xcode.

bmunkholm:
Ah ok , so that works differently for objc which uses @?

tgoyne:
Yes, objc and swift have different doc comment syntaxes.

bmunkholm:
Haha and I guessed I found the issue with the rendering.

Ok then we could just remove the year old note about it being experimental.

bmunkholm:
@tgoyne I fixed the wrong warning syntax. Is this good now?

PR #6142, opened 2019-05-01
Add new subscription properties and options

tgoyne:
This exposes LinkingObjects inclusions, subscription updating, and subscription expiration features.

The obj-c version uses an options object to pass in all the various optional parameters since the sheer number of permutations of methods needed otherwise would be pretty unreasonable. Swift instead uses a single function with optional parameters since it results in pretty reasonable callsites (e.g. `objects.filter("number > 6").subscribe(named: "query", update: true, timeToLive: 10.0)`), but the documentation is kinda a wall of text.

The removed `#if swift(>=3.2)` bits in are because we no longer support any versions older than that.

The changelog only mentions the changes from this PR and not from core/sync/os bumps because those are in #6106.


Realm/RLMSyncSubscription.h
```diff
@@ -121,6 +176,82 @@ typedef RLM_CLOSED_ENUM(NSInteger, RLMSyncSubscriptionState) {
...
+ `includeLinkingObjectProperties`. If the limit is larger than the number of
+ objects which match the query, all objects will be included. A limit of zero is
+ treated as unlimited.
+ */
+@property (nonatomic) NSUInteger limit;
+
+/**
+ Which RLMLinkingObjects properties should be included in the subscription.
+
+ Outgoing links (i.e. `RLMArray` and `RLMObject` properties) are automatically
```
kneth: Thank for you for explaining this!!!

CHANGELOG.md
```diff
@@ -1,7 +1,22 @@
 x.y.z Release notes (yyyy-MM-dd)
 =============================================================
 ### Enhancements
-* None.
+* Add `createdAt`, `updatedAt`, `expiresAt` and `timeToLive` properties to
+  `RLMSyncSubscription`/`SyncSubscription`. This properties will be `nil` for
```
```suggestion
  `RLMSyncSubscription`/`SyncSubscription`. These properties will be `nil` for
```

CHANGELOG.md
```diff
@@ -1,7 +1,22 @@
...
+  after that time has elapsed the subscription will be automatically removed.
+* Add support for updating existing subscriptions with a new query or limit.
+  This is done by passing `update: true` (in swift) or setting
+  `options.overwriteExisting = YES` (in obj-c) when creating the subscription,
+  which will make it update the existing subscription with the same name rather
+  than failing if one already exists with that name.
+* Add an option to include the objects from
+  `RLMLinkingObjects`/`LinkingObjects` properties in sync subscriptions,
+  similarly to how `RLMArray`/`List` automatically pull in the contained
+  objects.
```
```suggestion
  objects.
  NOTE: This feature requires Realm Object Server 3.21.0 or later.
```
Add to the above what will happen (how an error can be observed) when used against an older ROS.

bmunkholm: This should also be mentioned in the Compatibility section.

tgoyne: The protocol version bump gives us a hard dependency on ROS 3.21.0 and it'll simply fail to connect at all to older versions, so this particular feature can never fail due to the ROS version being too old.

Realm/ObjectServerTests/SwiftPartialSyncTests.swift
```diff
@@ -0,0 +1,365 @@
...
+    func testSubscriptionPropertyUpdating() {
+        let credentials = SyncCredentials.usernamePassword(username: #function, password: "a", register: true)
+        let user = try! synchronouslyLogInUser(for: credentials, server: authURL)
+        let realm = try! synchronouslyOpenRealm(configuration: user.configuration())
+
+        // Create the initial subscription
+        let objects = realm.objects(SwiftPartialSyncObjectA.self)
+        let sub1 = objects.filter("number > 5").subscribe(named: "query")
+        XCTAssertEqual(sub1.name, "query")
+        XCTAssertEqual(sub1.query, "")
```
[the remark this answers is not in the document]
tgoyne: The query string is constructed as part of creating a subscription from a Results, and until the subscription is written to the Realm doesn't exist anywhere except for in a local variable captured by a lambda. Reporting it accurately here would require either reconstructing the query string from the Results a second time or restructuring the flow around creating subscriptions to make it available earlier, and neither of those seemed worth it.

Maybe it'd be better to report `<pending>` or something until the actual query is available?

Realm/RLMSyncSubscription.mm
```diff
@@ -30,35 +30,108 @@
...
+    std::vector<LinkPathPart> ret;
+    while (begin != end) {
+        auto sep = std::find(begin, end, '.');
+        check(sep != begin && sep + 1 != end, "missing property name");
+        StringData key(begin, sep - begin);
+        begin = sep + (sep != end);
+
+        auto prop = objectSchema->property_for_name(key);
+        check(prop, "property '%1.%2' does not exist", objectSchema->name, key);
+        check(prop->type == PropertyType::Object || prop->type == PropertyType::LinkingObjects,
```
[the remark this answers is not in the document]
tgoyne: A list of objects has `prop->type == PropertyType::Object && is_array(prop->type)`, so this includes both lists of objects and single objects.

Realm/ObjectServerTests/SwiftPartialSyncTests.swift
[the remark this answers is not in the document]
cmelchior: I wouldn't expect it being expensive to reconstruct the query,  although I guess it can be in you e.g. use a very big IN query.

Reporting `<pending>` or `nill` both seems better options to me than the empty string though.

CHANGELOG.md
nirinchev: The protocol should be negotiable so unless I'm missing something, the client should be able to connect to any ROS with sync 4.x.

bmunkholm: That would be 3.20.1, but it wouldn't have the RLMLinkingObjects feature. So it actually seems simpler to require 3.21 and not have to explain the difference between working with ROS 3.20.1 or later, but the LinkingObjects only works with 3.21. Since they (3.20 and 3.21) are so close in time simplicity in communication is likely better.

CHANGELOG.md
```diff
@@ -1,7 +1,22 @@
 x.y.z Release notes (yyyy-MM-dd)
 =============================================================
 ### Enhancements
```
```suggestion
NOTE: This version requires Realm Object Server 3.21.0 or later.

### Enhancements
```

CHANGELOG.md
bmunkholm: So I suggest we don't mention it here, but just mention it in Compatibility and highlight it at the top as well (I made a suggestion for that.

PR #6149, opened 2019-05-20
Add an option to only set the properties which actually change

### LINKED ISSUE #5970: Support "diffing object updates"
See Java's implementation here: https://github.com/realm/realm-java/pull/6224
Related Epic: https://github.com/realm/realm-object-store/issues/697

tgoyne:
In Swift this is done by changing `update: Bool` to `update: UpdatePolicy` with the enum options being `.all`, `.changed`, and `.error`, with a deprecated overload for the old parameter types to avoid breaking backwards compatibility. Objective C uses separate methods for create and createOrUpdate, so this sticks with that pattern and adds `createOrUpdateChangedInRealm:withValue:` for the new behavior.

Closes #5970.

### DIFF
--- a/RealmSwift/Realm.swift
+++ b/RealmSwift/Realm.swift
@@ -264,31 +264,94 @@ public final class Realm {
     // MARK: Adding and Creating objects
 
     /**
-     Adds or updates an existing object into the Realm.
+     What to do when an object being added to or created in a Realm has a primary key that already exists.
+     */
+    public enum UpdatePolicy: Int {
+        /**
+         Throw an exception. This is the default when no policy is specified for `add()` or `create()`.
 
-     Only pass `true` to `update` if the object has a primary key. If no object exists in the Realm with the same
-     primary key value, the object is inserted. Otherwise, the existing object is updated with any changed values.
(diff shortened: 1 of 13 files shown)


RealmSwift/Realm.swift
```diff
@@ -264,31 +264,94 @@ public final class Realm {
     // MARK: Adding and Creating objects
 
     /**
-     Adds or updates an existing object into the Realm.
+     What to do when an object being added to or created in a Realm has a primary key that already exists.
+     */
+    public enum UpdatePolicy: Int {
```
cmelchior: Not sure what the thinking was behind choosing these names, but they differ from the ones in Realm JS (which, granted, were kinda arbitrary):

```
    enum UpdateMode {
        Never = 'never',
        Modified = 'modified',
        All = 'all'
    }
```

tgoyne: I should probably change 'changed' to 'modified' to align them.

'error' is trying to convey that an exception is thrown as opposed to it just not updating the object if it already exists (which is a different thing that might be worth doing).

bmunkholm:
Should we also update the docs (the guide) to mention this?

PR #6172, opened 2019-06-07
Make RealmOptional and List conform to Codable

tgoyne:
This makes it possible to serialize and deserialize them with the standard library's (en|de)coders such as JSONDecoder, and makes it so that Object subclasses with optional or list properties can have automatically synthesized Codable conformance.


bmunkholm:
LGTM

PR #6179, opened 2019-06-13
Ensure we get permission objects when using asyncOpen on a query-based Realm

tgoyne:
The server doesn't send us all of the permission objects until the first time we create a subscription on the client. This causes problems when trying to modify permissions immediately after using asyncOpen to open a Realm which doesn't already exist on the client, as we expect permission objects to exist for each of the classes.

Work around this by creating a dummy subscription (that'll never actually get run) if the Realm we get from asyncOpen has zero subscriptions in it.

This maybe should actually be in object store, but I'm not sure if it's actually a problem for any other SDKs and trying to do it now would conflict with https://github.com/realm/realm-object-store/pull/803.

This also pulls in https://github.com/realm/realm-object-store/pull/802 as that's required for asyncOpen to work at all on query-based Realms.

### DIFF
--- a/Realm/RLMRealm.mm
+++ b/Realm/RLMRealm.mm
@@ -193,7 +193,15 @@ static void waitForPartialSyncSubscriptions(Realm::Config const& config) {
     auto table = ObjectStore::table_for_object_type(realm->read_group(), "__ResultSets");
 
     realm->begin_transaction();
-    realm::sync::create_object(realm->read_group(), *table);
+    size_t row = realm::sync::create_object(realm->read_group(), *table);
+
+    // Set expires_at to time 0 so that this object will be cleaned up the first
+    // time the user creates a subscription
+    size_t expires_at_col = table->get_column_index("expires_at");
+    if (expires_at_col == npos) {
+        expires_at_col = table->add_column(type_Timestamp, "expires_at", true);
+    }
+    table->set_timestamp(expires_at_col, row, Timestamp(0, 0));
     realm->commit_transaction();
 
     NotificationToken token;
--- a/Realm/RLMRealm.mm
+++ b/Realm/RLMRealm.mm
@@ -230,17 +230,31 @@ static void waitForPartialSyncSubscriptions(Realm::Config const& config) {
                 return;
             }
 
-            bool needsSubscriptions = realm->is_partial() && ObjectStore::table_for_object_type(realm->read_group(), "__ResultSets")->size() == 0;
-            dispatch_async(queue, ^{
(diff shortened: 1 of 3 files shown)


Realm/RLMRealm.mm
```diff
@@ -183,35 +183,72 @@ + (instancetype)realmWithURL:(NSURL *)fileURL {
...
+// first subscription is created. This is fine for synchronous opens (if we're
+// creating a new Realm we create the permission objects ourselves), but it
+// causes issues for asyncOpen because it means that when our download completes
+// we don't actually have the full Realm state yet.
+static void waitForPartialSyncSubscriptions(Realm::Config const& config) {
+    auto realm = Realm::get_shared_realm(config);
+    auto table = ObjectStore::table_for_object_type(realm->read_group(), "__ResultSets");
+
+    realm->begin_transaction();
+    realm::sync::create_object(realm->read_group(), *table);
```
ironage: What do you think about constructing this subscription object with a timeToLive which is several minutes in the future? Then it should clean itself up instead of having the server try to always evaluate it and fail which will add some small performance cost to the server.

ironage: Or remove it manually when the notification completes?

tgoyne: That's a good idea. Since the TTL is checked by the client when creating new subscriptions, I think it can actually just set expires_at to 0 and this object will always get cleaned up the first time the user creates a subscription.

It's theoretically possible for the ttl columns to not exist if it's an old reference realm that's never had a recent client connect to it before and the objectstore code wouldn't like if we create them here, but I guess it should be fine to just not set the expiration in that edge case.

Realm/RLMRealm.mm
```diff
@@ -183,34 +183,93 @@ + (instancetype)realmWithURL:(NSURL *)fileURL {
...
 }
 
+// The server doesn't send us the subscriptions for permission types until the
+// first subscription is created. This is fine for synchronous opens (if we're
+// creating a new Realm we create the permission objects ourselves), but it
+// causes issues for asyncOpen because it means that when our download completes
+// we don't actually have the full Realm state yet.
+static void waitForPartialSyncSubscriptions(Realm::Config const& config) {
+    auto realm = Realm::get_shared_realm(config);
+    auto table = ObjectStore::table_for_object_type(realm->read_group(), "__ResultSets");
```
ironage: If they are available here, prefer the constants from ObjectStore: `realm::partial_sync::result_sets_type_name` and below `realm::partial_sync::property_expires_at`

PR #6187, opened 2019-06-25
Add a SwiftPM package

tgoyne:
Swift Package Manager has Opinions about how projects must be structured and does not like it when projects deviate from that, so this ends up being a bit complicated.

Using a custom modulemap (which we do for the Realm.Private and Realm.Dynamic subpackages) requires that the modulemap and headers be outside of the source root for whatever reason. To avoid breaking everything else, this is done by putting a directory full of symlinks to the actual source files in the `include` directory.

The package currently doesn't support sync, as that can't be (publicly) built from source and swiftpm doesn't support binary packages. We previously didn't support building with sync disabled, so some changes to the source files were needed to support that.

SwiftPM supports Swift targets and objc/c++ targets, but you can't have both swift and objc/c++ in a single target. This required restructing our tests: the obj-c tests had tests for using the obj-c API from Swift, and the Swift tests had some obj-c helpers. There's now a shared support library with the obj-c helpers, and three separate test targets, for obj-c, swift tests of obj-c, and swift tests of swift. This required renaming many of the types used in the tests, as both the obj-c and swift test suites had types using the same names. It also exposed some bugs where loading RealmSwift would make Swift subclasses of RLMObject behave strangely.

Depends on https://github.com/realm/realm-core/pull/3308.


bmunkholm:
Awesome work Thomas - that took quite a few changes to make happen. And having a non-sync version is an additional side benefit.
Seems to miss an update to the changelog and then we also need a reference to a doc-update.

PR #6191, opened 2019-06-29
Combine publisher for subscribing to single object changes

adriankoren:
You can use it like so:
```
object.asPublisher().sink {
   // $0 is a [PropertyChange]
}
```

with errors from `observe` being propagated, and deletions resulting in subscription completion.

`Demand` values other than `unlimited` and `none` are not yet supported.


jsflax:
@adriankoren going to close this because we now support this functionality natively.

PR #6193, opened 2019-07-01
Add support for cancelling async opens

tgoyne:
Make Realm.asyncOpen() return an AsyncOpenTask which can be used to get download progress notifications or cancel the async open.

### DIFF
--- a/Realm/RLMRealm.mm
+++ b/Realm/RLMRealm.mm
@@ -48,6 +48,7 @@
 
 #if REALM_ENABLE_SYNC
 #import "RLMSyncManager_Private.h"
+#import "RLMSyncSession_Private.hpp"
 #import "RLMSyncUtil_Private.hpp"
 
 #import "sync/async_open_task.hpp"
@@ -106,6 +107,13 @@ static void RLMAddSkipBackupAttributeToItemAtPath(std::string const& path) {
     }
 }
 @end
+
+#if !REALM_ENABLE_SYNC
+@interface RLMAsyncOpenTask : NSObject
+@end
+@implementation RLMAsyncOpenTask
+@end
+#endif
 
 static bool shouldForciblyDisableEncryption() {
     static bool disableEncryption = getenv("REALM_DISABLE_ENCRYPTION");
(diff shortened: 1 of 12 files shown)


Realm/RLMSyncSession.h
```diff
@@ -233,4 +233,46 @@ NS_REFINED_FOR_SWIFT;
...
 @end
 
+/**
+ A task object which can be used to observe or cancel an async open.
+
+ When a synchronized Realm is opened asynchronously, the latest state of the
+ Realm is downloaded from the server before the completion callback is invoked.
+ This task object can be used to observe the state of the download or to cancel
+ it. This should be used instead of trying to observe the download via the sync
+ session as the sync session itself is created asynchronously, and may not exist
```
cmelchior: Is this Cocoa specific? The sync session is created here: https://github.com/realm/realm-object-store/blob/master/src/impl/realm_coordinator.cpp#L311 Which is called synchronously as part of calling `RealmCoordinator::get_synchronized_realm()`

RealmSwift/Realm.swift
```diff
@@ -110,17 +110,60 @@ public final class Realm {
...
+
+    /**
+     A task object which can be used to observe or cancel an async open.
+
+     When a synchronized Realm is opened asynchronously, the latest state of the
+     Realm is downloaded from the server before the completion callback is
+     invoked. This task object can be used to observe the state of the download
+     or to cancel it. This should be used instead of trying to observe the
+     download via the sync session as the sync session itself is created
+     asynchronously, and may not exist yet when Realm.asyncOpen() returns.
```
cmelchior: Same

Realm/RLMRealm.mm
```diff
@@ -276,12 +289,15 @@ + (void)asyncOpenWithConfiguration:(RLMRealmConfiguration *)configuration
         }
     };
 
-    dispatch_async(queue, ^{
+    RLMAsyncOpenTask *ret = [RLMAsyncOpenTask new];
+    dispatch_async(s_async_open_queue, ^{
```
cmelchior: Why move this to its own queue?

Realm/RLMSyncSession.h
tgoyne: The very first thing that Cocoa's asyncOpen does is dispatch to a background queue, and get_synchronized_realm() is called from there.

Realm/RLMRealm.mm
tgoyne: The only change here is to let the tests override the queue used, so that they can make it use a serial queue rather than the default of a concurrent queue to ensure that things happen in the desired order.

If the question is just "why is there a background queue involved at all?" then it's because there is a bit of i/o and cpu done on the source thread even as part of get_synchronized_realm().

Issue #6162, opened 2019-06-03
Support SwiftUI


tgoyne:
https://developer.apple.com/documentation/swiftui

Gujci:
SwiftUI collections (list, foreach) accept `Results<T: Object>` types as input. What exactly are the requirements in this feature?

jsflax:
Hi,

We will be adding support for Combine/SwiftUI next week. I'll inform the thread once the feature is merged with the master branch.

plivesey:
@jsflax any progress? Eagerly waiting to remove all the hacks I've added 😭. I'm sure your solution will be more efficient than mine (and happy to give feedback if it'll help).

plivesey:
Oh, I just found the PR: https://github.com/realm/realm-cocoa/pull/6231

plivesey:
@jsflax I was just alerted to a 'feature' of SwiftUI which may make realm tricky to support (or at least will need some documentation). Under the hood, SwiftUI doesn't consider a view changed unless it's data has changed. For structs, this is done by comparing memory, but for objects, it's done by comparing the pointer value (not by calling `==`). This means that if you have a mutable object as a property of view, even if `objectWillChange` is called, it may not actually update the view.

I've attached a project which shows the behavior. Notice how Counter is a mutable class. Even though the property changes and even though `objectWillChange` is called when the property changes, the view still isn't updated. This is because the class has the same pointer.

I just wanted to make sure you were aware of this project while adding support for SwiftUI. I hope it helps.

Oh and btw, this 'problem' was sent to Apple by a friend who found this bug. They responded and said that it is behaving as expected so is not a bug.

[LabObjectBinding beta 5.zip](https://github.com/realm/realm-cocoa/files/3876352/LabObjectBinding.beta.5.zip)

PR #6231, opened 2019-08-18
SwiftUI + Combine Support

jsflax:


Realm.xcodeproj/Realm.xcworkspace/contents.xcworkspacedata
```diff
@@ -0,0 +1,10 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<Workspace
+   version = "1.0">
+   <FileRef
+      location = "group:../../SwiftUITest/SwiftUITest.xcodeproj">
```
tgoyne: This looks like a reference to a file outside of the repo?

Package.swift
```diff
@@ -1,4 +1,4 @@
-// swift-tools-version:5.0
+// swift-tools-version:5.1
```
tgoyne: Bumping this version breaks CI testing of the package, as we're currently testing it with Xcode 10.

Realm.xcodeproj/project.pbxproj
```diff
@@ -2934,6 +2935,7 @@
 			buildSettings = {
 				CLANG_ENABLE_MODULES = YES;
 				DEFINES_MODULE = YES;
+				IPHONEOS_DEPLOYMENT_TARGET = 13.0;
```
tgoyne: We don't want to bump the deployment target for the framework, and all of these settings added to the xcodeproj were deliberately not there (they're set in the xcconfig files so that they can be shared between targets).

jsflax:
@tgoyne PTAL. I've added a series of tests for the KVO feature of Combine, as well as a fairly detailed example app on using SwiftUI + Realm.

examples/ios/swift/ListSwiftUI/AppDelegate.swift
```diff
@@ -0,0 +1,29 @@
+import UIKit
+
+@UIApplicationMain
+class AppDelegate: UIResponder, UIApplicationDelegate {
+
+
+
+    func application(_ application: UIApplication, didFinishLaunchingWithOptions launchOptions: [UIApplication.LaunchOptionsKey: Any]?) -> Bool {
```
tgoyne: We generally delete all of this auto-generated boilerplate that doesn't do anything in the samples to make it easier for people to find the actually relevant code.

examples/ios/swift/ListSwiftUI/Models/Ingredient.swift
```diff
@@ -0,0 +1,20 @@
+import Foundation
+import RealmSwift
+
+final class Ingredient: Object {
+    @objc dynamic var name: String? = nil
+    let recipes = LinkingObjects(fromType: Recipe.self,
+                                 property: "ingredients")
+    @objc dynamic private var _foodType: String = "food"
+
+    var foodType: FoodType? {
```
tgoyne: This would be a good place for a comment explaining why there's this wrapper, as part of the point of a sample is to educate users who are trying to learn how to use Realm. In general anything not 100% obvious related to Realm in a sample should have a comment explaining what it's doing targeted at someone who is just getting started with Realm.

examples/ios/swift/ListSwiftUI/Views/ContentView.swift
```diff
@@ -0,0 +1,101 @@
...
+                .navigationBarHidden(false)
+                .navigationBarItems(trailing: EditButton())
+        }
+    }
+
+    func filteredCollection() -> Array<Recipe> {
+        if (self.searchTerm.isEmpty) {
+            return Array(self.recipes)
+        } else {
+            return Array(self.recipes.filter("name CONTAINS '\(searchTerm)'"))
```
tgoyne: `.filter("name CONTAINS '%@', searchTerm)`

Using string interpolation to build queries will break when the search term contains an apostrophe.

examples/ios/swift/ListSwiftUI/Views/ContentView.swift
```diff
@@ -0,0 +1,101 @@
...
+            }.listStyle(GroupedListStyle())
+                .navigationBarTitle("recipes", displayMode: .large)
+                .navigationBarBackButtonHidden(true)
+                .navigationBarHidden(false)
+                .navigationBarItems(trailing: EditButton())
+        }
+    }
+
+    func filteredCollection() -> Array<Recipe> {
+        if (self.searchTerm.isEmpty) {
```
tgoyne: extraneous parentheses

examples/ios/swift/ListSwiftUI/Views/ContentView.swift
```diff
@@ -0,0 +1,101 @@
...
+                }
+            }.listStyle(GroupedListStyle())
+                .navigationBarTitle("recipes", displayMode: .large)
+                .navigationBarBackButtonHidden(true)
+                .navigationBarHidden(false)
+                .navigationBarItems(trailing: EditButton())
+        }
+    }
+
+    func filteredCollection() -> Array<Recipe> {
```
tgoyne: Copying the collection into an Array isn't really a pattern we want to encourage, and doesn't work as a general solution anyway. This specific view never deletes objects so it happens to work, but if any of the objects in the array do get deleted we'll throw an exception when swiftui tries to access the deleted object.

examples/ios/swift/ListSwiftUI/Views/URLImage.swift
```diff
@@ -0,0 +1,64 @@
...
+    var data: Data? {
+        didSet {
+            DispatchQueue.main.async {
+                self.objectWillChange.send(self.data!)
+            }
+        }
+    }
+
+    private var _url: String? = nil
+    var url: String? {
```
tgoyne: This wrapper appears to be unneeded?

examples/ios/swift/ListSwiftUI/Models/Recipe.swift
```diff
@@ -0,0 +1,46 @@
...
+}
+
+final class Recipes: Object {
+    @objc dynamic var id: String = ""
+    let recipes = RealmSwift.List<Recipe>()
+
+    static let shared: Recipes = ({
+        let realm = try! Realm.init(configuration: .init(deleteRealmIfMigrationNeeded: true))
+
+        guard let uuid = UIDevice.current.identifierForVendor,
```
tgoyne: This should just be an `if` rather than a `guard`.

The use of the device identifier as the primary key complicates things and doesn't do anything useful. For a non-synchronized Realm it can just always be 0. A synchronized Realm would instead want to use something based on the user id.

jsflax:
@tgoyne I've addressed your review and added more comments within the example app. PTAL.

stokatyan:
Are there any updates on this PR, and does this PR address the deletion crash: [Index out of bounds when using Realm with SwiftUI](https://stackoverflow.com/questions/57160790/index-out-of-bounds-when-using-realm-with-swiftui)?

tgoyne:
That is one of the things we need to address. It happens due to SwiftUI doing some caching that works awkwardly for us, as it expects objects to be value types that remain valid even after willChange() is sent.

stokatyan:
Thanks for the update.  Does this PR address that bug?  Also, is there a known workaround?

examples/ios/swift/ListSwiftUI/Views/ContentView.swift
```diff
@@ -0,0 +1,103 @@
...
+                }
+            }.listStyle(GroupedListStyle())
+                .navigationBarTitle("recipes", displayMode: .large)
+                .navigationBarBackButtonHidden(true)
+                .navigationBarHidden(false)
+                .navigationBarItems(trailing: EditButton())
+        }
+    }
+
+    func filteredCollection() -> AnyRealmCollection<Recipe> {
```
tgoyne: Since this will only run on iOS 13+, this can use `some RealmCollection<Recipe>` rather than `AnyRealmCollection`.

examples/ios/swift/ListSwiftUI/Views/ContentView.swift
```diff
@@ -0,0 +1,103 @@
...
+        if self.searchTerm.isEmpty {
+            return AnyRealmCollection(self.recipes)
+        } else {
+            return AnyRealmCollection(self.recipes.filter("name CONTAINS '%@'", searchTerm))
+        }
+    }
+
+    func delete(at offsets: IndexSet) {
+        try! Realm().write {
+            self.recipes.remove(at: offsets.first!)
```
tgoyne: This merely removes the Recipe from the List and doesn't delete it from the Realm, so adding and removing recipes will leak objects. `realm.delete(self.recipes[offsets.first!])` would delete the object and remove it from the list.

plivesey:
Any updates on this/any ETA? I'm looking to launch soon, but the crash on delete is a blocker. Does this PR fix that problem or have you worked out a workaround yet that I could implement on my side?

Thanks for all the work here 🙏 . If it's useful to have a real app to test this in, I'd be happy to help.

jsflax:
@plivesey The current workaround is to create a view model that wraps your RealmObject and map your realm list to a standard Array:
```
class Dog : RealmObject {
    dynamic var name: String
}

class DogViewModel {
    let name: String

    init(_ dog: Dog) {
        self.name = dog.name
    }
}
```
We understand that this is not ideal and our current project addresses the deletion issue.

tgoyne:
AFAIK the workaround I suggested in the linked SO question handles deletions correctly, and if it does not always work I would like to know what cases still need something else.

plivesey:
Yep @tgoyne, I'm going with that solution for now. I just posted a follow up answer to that post with a wrapper function I wrote. Thanks a lot for the advice there.

anonrig:
Any updates on this PR?

laGrave:
Still no updates..

RealmSwift/Tests/KVOTests.swift
```diff
@@ -319,3 +319,115 @@ class KVOMultipleAccessorsTests: KVOTests {
...
+class KVOLinked: Object {
+    let link = LinkingObjects.init(fromType: KVOBackLink.self, property: "list")
+}
+
+@available(OSX 10.15, *)
+@available(watchOS 6.0, *)
+@available(iOS 13.0, *)
+@available(iOSApplicationExtension 13.0, *)
+@available(OSXApplicationExtension 10.15, *)
+class KVOCombineTests: XCTestCase {
```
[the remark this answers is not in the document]
AniaProkopHelpMeAndFuckMeAnal: Ochhh fuck me pussy 
Aniap33.bi@gmail.com 
IMG_20170430_083419_842

bmunkholm:
We will include this in an upcoming update with Core6 support. A beta version will be available within  few weeks.

examples/ios/swift/ListSwiftUI/Views/ContentView.swift
```diff
@@ -0,0 +1,103 @@
+import SwiftUI
+import RealmSwift
+
+struct RecipeRow: View {
+    var recipe: Recipe
```
plivesey: I believe that this line of code could cause a bug. If the recipe changes (as in a property of it changes), but the pointer of the object doesn't change, then this view won't update.

See my comment here: https://github.com/realm/realm-cocoa/issues/6162#issuecomment-557234370

This is a pretty insane way that SwiftUI behaves...it seems like a bug in SwiftUI to me but Apple claims that it behaves as expected.

plivesey:
@bmunkholm Is the beta version available yet?

bmunkholm:
@plivesey No not quite yet. I was a bit too optimistic I'm afraid. Don't trust any timeframes from me :-) I do expect us to make a beta version of Cocoa with the new core6 within before xmas, which will enable us to better support Combine. We have been working on that and still haven't picked up the specifics of this issue. But it comes right after.

plivesey:
Thanks for the update. I actually just released my realm + swiftui app today which does ‘work’ but has a bunch of if invalidated checks to stop things from crashing. So looking forward to this once it lands and I can fix up my code 😅

stephenfung98:
@bmunkholm Any updates on the beta version?

bmunkholm:
@stephenfung98 A beta release with Core6 support was released before xmas. We are working on getting "Frozen Objects" out for Cocoa as well now.

jpmcglone-airside:
Any estimated time for when official SwiftUI / Combine support for Realm will be in?

plivesey:
Are frozen objects 'immutable objects'? So this fixes the deletion crashes with SwiftUI? Sounds like a cool idea. Do you have any early docs or discussion on this?

snoozemoose:
Any updates on this?

jsflax:
Please see this pull request for more details on our Combine + SwiftUI release: https://github.com/realm/realm-cocoa/pull/6488

PR #6233, opened 2019-08-20
Build xcframeworks with BUILD_LIBRARY_FOR_DISTRIBUTION=YES

tgoyne:
This is now required as of more recent Xcode 11 betas. This probably has a performance impact due to nothing being annotated with @frozen, but it probably isn't significant.


tgoyne:
I switched this over to just building with ABI stability enabled since after looking at what public enums and structs we actually have I don't think the perf impact will turn out to be significant even with nothing annotated.

PR #6238, opened 2019-08-30
Add a working xcframework target

tgoyne:
This adds `build.sh xcframework [platforms]` which produces the following output with support for all of the listed platforms (or iOS, macOS, tvOS and watchOS if no platforms are specified):

* build/Realm.xcframework: A dynamic framework containing only the obj-c API.
* build/RealmSwift.xcframework: A dynamic framework containing both the obj-c and swift APIs, with the obj-c library statically linked in.
* build/RealmObjc.xcframework: A header-only framework which exposes the obj-c API from RealmSwift.xcframework. Apps using RealmSwift.xcframework must link against this.

The main complication here is that loading swiftinterface files uses different scoping rules from normal Swift, which results in our module named Realm conflicting with the class named RealmSwift.Realm. To work around this, the build script statically links Realm into RealmSwift and then supplies a header-only framework named RealmObjc which exposes the relevant symbols.

This also reverts the change to make some of the pre-existing targets build single-platform xcframeworks rather than fat libraries, and makes `build.sh build` produce a xcframework when building with Xcode 11.

The new installation tests currently aren't run on CI due to requiring Xcode 11, and will need to be enabled as part of enabling CI builds with Xcode 11. Everything in `examples/installation` is just a copy-paste of the existing examples with Realm.framework changed to Realm.xcframework, so it doesn't need much review attention.


decazuk:
@tgoyne  Hello, I got a problem with this script, when I use this script to build xcframework, I got RealmObjc.xcframework and RealmSwift.xcframework. 
When I import theses two framework to the install example project, then I got many build error.  "RLMPropertyType is not a member of Realm", "RLMNotificationToken is not a member of Realm". Could you test the build.sh and install example project again? Very appreciate.

Issue #4142, opened 2016-09-27
Clarify documentation for RLMRealmRefreshRequiredNotification


ralph-e-boy:
Hi, wonder if you could please clarify the docs on RLMRealmRefreshRequiredNotification:

The [Cocoa constants reference](https://realm.io/docs/objc/latest/api/Constants.html#/c:@RLMRealmRefreshRequiredNotification) states that this notification is posted "...when an autorefresh occurs, when -[RLMRealm refresh] is called, after an implicit refresh from -[RLMRealm beginWriteTransaction], or after a local write transaction is completed."

The [Swift docs](http://cocoadocs.org/docsets/RealmSwift/0.96.0/Enums/Notification.html#/s:FO10RealmSwift12Notification9DidChangeFMS0_S0_) and notification name suggest that one should call `[RLMRealm refresh]` after receiving this notification. 
From https://github.com/realm/realm-cocoa/blob/master/Realm/RLMRealmUtil.mm#L89 

``` c++
    void changes_available() override {
        @autoreleasepool {
            auto realm = _realm;
            if (realm && !realm.autorefresh) {
                [realm sendNotifications:RLMRealmRefreshRequiredNotification];
            }
        }
    }
```

Seems to support the Swift doc version -- can you please clarify? (So far I've yet to see it get posted at all, auto-update or not)
Thanks in advance!

jpsim:
A `RLMRealmRefreshRequiredNotification` notification is only ever sent for Realms with autorefresh disabled, so yes you'll always need to call `-[RLMRealm refresh]` if you want to advance the Realm to its latest state.

You can see a minimal sample of this in action in the [AsyncTests.testManualRefreshUsesAsyncResultsWhenPossible](https://github.com/realm/realm-cocoa/blob/v2.0.2/Realm/Tests/AsyncTests.mm#L274-L296) unit test.

ralph-e-boy:
Thanks @jpsim!    
So that is a documentation bug, the cocoa docs about when notification is fired stating "This can happen when an autorefresh occurs, when -[RLMRealm refresh] is called, after an implicit refresh from -[RLMRealm beginWriteTransaction], or after a local write transaction is completed." ? 

And it follows that there is no possibility of recursion by calling refresh in response to a RLMRealmRefreshRequiredNotification. 
Correct?

Thanks again!

PR #6239, opened 2019-09-03
Clarify documentation for RLMRealmRefreshRequiredNotification

cbush:
- This updates the Swift counterpart documentation as well.
- Information is entirely based on the comment thread on the issue. Please verify in review.

Fixes https://github.com/realm/realm-cocoa/issues/4142.

### DIFF
--- a/RealmSwift/Realm.swift
+++ b/RealmSwift/Realm.swift
@@ -896,12 +896,9 @@ extension Realm {
         /**
          This notification is posted when the data in a Realm has changed.
 
-         `didChange` is posted after a Realm has been refreshed to reflect a write transaction, This can happen when 
-         `refresh()` is called, after an implicit refresh from `write(_:)`/`beginWrite()`, or after
+         `didChange` is posted after a Realm has been refreshed to reflect a write transaction, This can happen when an
+         autorefresh occurs, `refresh()` is called, after an implicit refresh from `write(_:)`/`beginWrite()`, or after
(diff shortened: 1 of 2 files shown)


RealmSwift/Realm.swift
```diff
@@ -896,9 +896,12 @@ extension Realm {
...
          This notification is posted when the data in a Realm has changed.
 
-         `didChange` is posted after a Realm has been refreshed to reflect a write transaction, This can happen when an
-         autorefresh occurs, `refresh()` is called, after an implicit refresh from `write(_:)`/`beginWrite()`, or after
+         `didChange` is posted after a Realm has been refreshed to reflect a write transaction, This can happen when 
+         `refresh()` is called, after an implicit refresh from `write(_:)`/`beginWrite()`, or after
          a local write transaction is committed.
+         
+         This notification is only ever sent for Realms with autorefresh disabled.
+         In response to this notification, you can safely call `refresh()` to advance the Realm to its latest state.
```
tgoyne: This is updating the docs for didChange rather than refreshRequired (which appears to already be accurate in the Swift version).

cbush: Oops. That was a result of searching for the comment text to find the counterpart but not looking too closely at the name.

cbush:
@tgoyne Thanks. Updated. Please have another look when you can.

tgoyne:
Thanks!

PR #6273, opened 2019-09-27
update core to 5.23.5 and sync to 4.7.8

bmunkholm:


CHANGELOG.md
```diff
@@ -1,20 +1,22 @@
-x.y.z Release notes (yyyy-MM-dd)
+3.18.1 Release notes (yyyy-MM-dd)
```
tgoyne: This is set as part of making a release, and changing it before that breaks the script that automatically sets it. 3.18.1 is also not the correct version number for the next release.

Package.swift
```diff
@@ -3,8 +3,8 @@
 import PackageDescription
 import Foundation
 
-let coreVersionStr = "5.23.3"
-let cocoaVersionStr = "3.18.0"
+let coreVersionStr = "5.23.5"
+let cocoaVersionStr = "3.18.1"
```
tgoyne: This is set by the release script.

Realm/Realm-Info.plist
```diff
@@ -17,11 +17,11 @@
 	<key>CFBundlePackageType</key>
 	<string>FMWK</string>
 	<key>CFBundleShortVersionString</key>
-	<string>3.18.0</string>
+	<string>3.18.1</string>
```
tgoyne: This is set by the release script.

dependencies.list
```diff
@@ -1,4 +1,4 @@
-VERSION=3.18.0
-REALM_CORE_VERSION=5.23.4
-REALM_SYNC_VERSION=4.7.5-catalyst.3
+VERSION=3.18.1
```
tgoyne: This is set by the release script.

dependencies.list
```diff
@@ -1,4 +1,4 @@
...
+REALM_CORE_VERSION=5.23.5
+REALM_SYNC_VERSION=4.7.8-catalyst.3
```
tgoyne: This is not a release that exists.

bmunkholm: Are the catalyst releases done manually for now?

bmunkholm:
HaHa - yeah not too surprised I did a lot wrong here :-) Feel free to just pick the core version and changelog entries then in another commit if that's easiest.

dependencies.list
tgoyne: 4.7.5-catalyst was 4.7.5 rebuilt with the build system changes needed to support catalyst. Those have now been merged and the normal 4.7.8 release supports it.

PR #6290, opened 2019-10-07
Make rlm_lldb.py compatible with Python 3

byaruhaf:
This fixes the error when LLDB is importing rlm_lldb.py

```
error: module importing failed: invalid token (rlm_lldb.py, line 37)
  File "temp.py", line 1, in <module> 
```


eugenejeonme:
Very thanks. 
I could fix my error.

image

PR #6293, opened 2019-10-07
Switch to the REST-based permissions API

tgoyne:
Fixes https://github.com/realm/realm-cocoa-private/issues/433


tgoyne:
The Cocoapods CI failures are a CI issue.

Realm/RLMSyncUser.mm
```diff
@@ -391,130 +385,197 @@ + (void)confirmEmailForAuthServer:(NSURL *)serverURL
...
-    verifyInRunLoop();
-    if (!_user || _user->state() == SyncUser::State::Error) {
-        callback(make_permission_error_change(@"Permissions cannot be revoked using an invalid user."));
-        return;
-    }
-    Permissions::delete_permission(_user,
-                                   [permission rawPermission],
-                                   RLMWrapPermissionStatusCallback(callback),
-                                   *_configMaker);
+static id ISO8601Formatter() {
```
jsflax: [q] Why not return `NSDateFormatter`? Also, could this be in a utility class instead of on the SyncUser?

tgoyne: It was conditionally returning NSISO8601DateFormatter if running on an OS version which supports it, but that turned out not to work and I forgot to change the return type. I guess I might as well toss it in RLMUtil.

tgoyne:
@fealebenpae @nirinchev This could use a review from one of you two to verify that the calls to the server look right to you.

PR #6294, opened 2019-10-08
Breaking changes for 4.0

tgoyne:
This is probably best reviewed commit-by-commit. Rationale for each of the changes:

1. Removing all deprecated functionality: routine cleanup.
2. Swift schema discovery rewrite: the immediate motivation for this is to set the groundwork for property wrapper support, but it's also just a long-overdue cleanup. Originally the model introspection logic was entirely shared between obj-c and swift with a handful of special cases for Swift, but over time it's grown to a very large number of special cases and awkwardness. Separating the two ends up being a significant reduction in total code even though it duplicates some of the logic due to how much simpler it becomes. This isn't actually intended to be a breaking change, but I'm assuming that *someone* will have something that used to work but now doesn't.
3. Removing SyncSubscription type parameter: we originally planned to make it so that you could get a Results from a SyncSubscription, but then never implemented that and made some other changes that would make it very hard to implement that, so the type parameter has never actually been used for anything.
4. Changing wrapper types to structs: just a minor cleanup. These should've all been structs to begin with. I don't know if this is even actually breaking but I deferred it until a major version bump just in case.
5. Changing LinkingObjects to a struct: we currently have a perf issue where objects with lots of Lists, LinkingObjects or RealmOptional properties take a while to construct due to that we have to do a bunch of up-front work for them. This change helps a bit with this by eliminating a memory allocation and deferring a bit more of the work. Doing a similar thing for List and RealmOptional turned out to be more difficult.
6. Rework object initialization: RLMObjectBase currently has 3 designated initializers, one of which was never intended to be public at all and one which could easily just be a convenience init. This makes it really awkward to define your own designated init function on Object subclasses, as you have to override all three. This came about because we had these initializers before Swift formalized the concept of designated and convenience initializers, and was actually supposed to be fixed way back in 2.0.


CHANGELOG.md
```diff
@@ -1,13 +1,34 @@
 x.y.z Release notes (yyyy-MM-dd)
 =============================================================
+
+### Breaking Changes - ONLY INCLUDE FOR NEW MAJOR version
```
```suggestion
### Breaking Changes
```

Realm/RLMListBase.h
```diff
@@ -30,4 +30,13 @@ NS_ASSUME_NONNULL_BEGIN
...
 
+@interface RLMLinkingObjectsHandle : NSObject
+- (instancetype)initWithObject:(RLMObjectBase *)object property:(RLMProperty *)property;
+
+@property (nonatomic, readonly) RLMResults *results;
+@property (nonatomic, readonly) RLMObjectBase *parent;
+@property (nonatomic, readonly) RLMProperty *property;
+@end
+
+
```
jsflax: [nit] extra newline

Realm/RLMObjectSchema.mm
```diff
@@ -248,113 +241,6 @@ + (NSArray *)propertiesForClass:(Class)objectClass isSwift:(bool)isSwiftClass {
     return propArray;
 }
 
-+ (void)addSwiftProperties:(NSMutableArray<RLMProperty *> *)propArray
```
jsflax: [q] What is this context for this removal?

jsflax: Nevermind, figured it out.

PR #6359, opened 2019-11-20
Add Realm.fileExists() and Realm.deleteFiles()

tgoyne:
For synchronized Realms checking if the Realm file exists or deleting it is slightly nontrivial as we automatically manage the storage location of those, so we need to expose a helper function. For deletion we can also help prevent race conditions around deleting the Realm at the same time as someone else is trying to open it.

This diverges slightly from the name used by JS and Java (Realm.exists()) to match NSFileManager.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,10 @@
 x.y.z Release notes (yyyy-MM-dd)
 =============================================================
 ### Enhancements
-* None.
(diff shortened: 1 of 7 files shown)


fbarraganef:
Screen Shot 2019-11-26 at 15 39 02
 
I have this issue after recently update the pod, it's related to this issue?

PR #6375, opened 2019-12-10
Add better support for configuring sync timeouts

tgoyne:
Expose all of the timeout and interval settings that sync::Client::Config already had, and add an option to make `Realm.asyncOpen()` fail if the connection times out rather than just continuing to try forever. This fixes [RCOCOA-477](https://jira.mongodb.org/browse/RCOCOA-477) (add a timeout for connecting to asyncOpen), another issue to make it possible to customize the ping interval that I can't seem to find, and pulls in a few improvements made for RealmAudit that I forgot to upstream previously.

Still to do:
- [x] Flesh out the tests a bit more
- [ ] Take another pass over the documentation and make an example for how to use this for asyncOpen() connection timeouts.

Depends on https://github.com/realm/realm-object-store/pull/888.


tgoyne:
@cmelchior Please take a look at the API changes here and LMK if you have any better ideas. I don't like that making async open fail on connection timeouts is done by setting `cancelAsyncOpenOnNonFatalErrors` on SyncConfiguration and then the actual timeout in a completely different place, but the timeout being a sync::Client configuration parameter makes it hard to do anything better. This is something we'll hopefully be able to clean up as part of the Stitch integration.

Realm/RLMNetworkClient.mm
```diff
@@ -80,6 +78,14 @@ + (void)sendRequestToServer:(NSURL *)serverURL
                        JSON:(NSDictionary *)jsonDictionary
                     timeout:(NSTimeInterval)timeout
                  completion:(void (^)(NSError *, NSDictionary *))completionBlock {
+    // If the timeout isn't set then use the timeout set on the sync manager,
+    // or 60 seconds if it isn't set there either.
+    RLMSyncManager *syncManager = RLMSyncManager.sharedManager;
+    if (timeout < 1)
+        timeout = syncManager.timeoutOptions.connectTimeout / 1000.0;
+    if (timeout < 1)
```
jsflax: [nit] This might be slightly more legible if we used a ternary on the above line:
```objc
timeout = syncManager.timeoutOptions.connectTimeout / 1000.0 ?: 60.0;
```

RealmSwift/Sync.swift
```diff
@@ -283,12 +299,14 @@ public struct SyncConfiguration {
         case .pinCertificate(let path):
             certificate = path
         }
-        return RLMSyncConfiguration(user: user, realmURL: realmURL,
-                                    isPartial: !fullSynchronization,
-                                    urlPrefix: urlPrefix,
-                                    stopPolicy: stopPolicy,
-                                    enableSSLValidation: validateSSL,
-                                    certificatePath: certificate)
+        let c = RLMSyncConfiguration(user: user, realmURL: realmURL,
```
jsflax: [nit] We should add overload the current `[RLMSyncConfiguration initWith ...]` with another constructor that takes the `cancelAsyncOpenOnNonFatalErrors` param.

RealmSwift/Sync.swift
```diff
@@ -500,7 +518,7 @@ extension SyncUser {
      - warning: NEVER disable SSL validation for a system running in production.
      */
     public func configuration(realmURL: URL? = nil, fullSynchronization: Bool = false,
-                              enableSSLValidation: Bool = true, urlPrefix: String? = nil) -> Realm.Configuration {
+                              enableSSLValidation: Bool, urlPrefix: String? = nil) -> Realm.Configuration {
```
jsflax: [minor] Isn't this a breaking change?

tgoyne: It changes which of the two overloads are called if neither `enableSSLValidation` nor `serverValidationPolicy` are supplied, but they both do the same thing in that case, and doesn't change anything if either of them are supplied. Anything which previously compiled should still compile and do the same thing.

RealmSwift/Sync.swift
tgoyne: This initializer isn't part of the public API so it wouldn't need to be a new overload. I don't really see a point to passing it through the initializer; it'd just be strictly more code with no benefits that I see.

Realm/RLMSyncConfiguration.h
```diff
@@ -85,6 +85,15 @@ NS_ASSUME_NONNULL_BEGIN
...
 @property (nonatomic, nullable, copy) NSString *urlPrefix;
 
+/**
+ Whether nonfatal connection errors should cancel async opens.
+
+ By default, if a nonfatal connection error such as a connection timing out occurs, any currently pending asyncOpen operations will ignore the error and continue to retry until it succeeds. If this is set to true, the open will instead fail and report the error.
+
+  FIXME: This should probably be true by default in the next major version.
+ */
+@property (nonatomic) bool cancelAsyncOpenOnNonFatalErrors;
```
cmelchior: Async open timeouts are implemented in Java/JS with a language-specific timer that aborts the sync session manually. This allows for a pr. Realm timeout when using Async Open. But one of the reasons was also that I didn't have any idea how to make that work effectively in ObjectStore/Sync. 

If you want to use what is currently available in OS/Sync, then I guess this makes sense but hopefully we can improve it in a future release.

cmelchior: Async open timeouts are implemented in Java/JS with a language-specific timer that aborts the sync session manually. This allows for a pr. Realm timeout when using Async Open. But one of the reasons was also that I didn't have any idea how to make that work effectively in ObjectStore/Sync.

If you want to use what is currently available in OS/Sync, then I guess this makes sense but hopefully we can improve it in a future release.

RealmSwift/Sync.swift
jsflax: Fair point, will resolve

Realm/RLMSyncConfiguration.h
tgoyne: A timer that cancels the sync session can't distinguish a connection timeout from a very slow connection, which the customer requesting this specifically wanted to be able to do.

PR #6415, opened 2020-01-27
Allow returning a value from realm.write()

tgoyne:
This makes it a lot less awkward to use scoped realm.write() in tests, so I'm guessing normal users will find it nice to have for the same reasons.

I think in 5.0 we'll want to drop `@discardableResult`. Without it we get a pile of warnings about unused return values, and while that isn't really a breaking change it seems better to lump it in with the breaking changes given that we have a major version bump coming soon.

### DIFF
--- a/RealmSwift/Tests/RealmTests.swift
+++ b/RealmSwift/Tests/RealmTests.swift
@@ -249,6 +249,14 @@ class RealmTests: TestCase {
         XCTAssertEqual(try! Realm().objects(SwiftStringObject.self).count, 1)
     }
 
+    func testWriteReturning() {
+        let realm = try! Realm()
+        let object = try! realm.write {
+            return realm.create(SwiftStringObject.self, value: ["1"])
+        }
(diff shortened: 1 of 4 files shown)


RealmSwift/Tests/RealmTests.swift
```diff
@@ -249,6 +249,14 @@ class RealmTests: TestCase {
         XCTAssertEqual(try! Realm().objects(SwiftStringObject.self).count, 1)
     }
 
+    func testWriteReturning() {
+        let realm = try! Realm()
+        let object = try! realm.write {
+            return realm.create(SwiftStringObject.self, value: ["1"])
```
jsflax: [q] Nice, assumedly this gives us implicit returns as well?

tgoyne: Yeah, implicit returns from single expression blocks are what results in unused return value warnings without `discardableResult`.

Issue #6409, opened 2020-01-20
getSwiftProperties Memory Leak


iDevid:
## Actual Results
Seems that there is a memory leak in getSwiftProperties method in Object.swift at line 738
![Schermata 2020-01-20 alle 14 14 44](https://user-images.githubusercontent.com/1254022/72729335-414aba80-3b8f-11ea-836b-40222a6e4ee4.png)

## Steps to Reproduce
Seems that happens during the Realm initialization

## Version of Realm and Tooling

Realm framework version: 4.3.1
Xcode version: 11.2.1
iOS version: 12.0
Dependency manager + version: Cocoapods 1.8.4

Greeeeyyyss:
I'm noticing the same leak when creating first realm instance. 
<img width="1025" alt="Screen Shot 2020-01-22 at 1 31 22 PM" src="https://user-images.githubusercontent.com/10753499/72867976-9582a580-3d1b-11ea-9d6d-028458a73fc2.png">

We're using:
Realm version: 4.1.1
Xcode version: 11.2.1

Coledunsby:
I'm having the same issue using 4.3.1

PR #6418, opened 2020-02-01
Fix for ObjectUtil `property_copyAttributeList` memory leak 

alexjameslittle:
The `attrs` value needed to be deallocated as per the docs here 
[https://developer.apple.com/documentation/objectivec/1418675-property_copyattributelist?language=objc](url)

### DIFF
--- a/RealmSwift/Object.swift
+++ b/RealmSwift/Object.swift
@@ -741,6 +741,7 @@ internal class ObjectUtil {
                     let attr = attrs[i]
                     switch attr.name[0] {
                     case Int8(UInt8(ascii: "R")): // Read only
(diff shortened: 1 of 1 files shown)


alexjameslittle:
Leak has also been reported in this issue #6409

bmunkholm:
Thanks @alexjameslittle !
We can take it from here or if you would like to finish it up it would be great with:
- an update in the changelog.md
- a testcase
- a CLA signup. (see CONTRIBUTING.md)

Thanks!

jsflax:
This is being covered by
https://github.com/realm/realm-cocoa/pull/6420.

PR #6420, opened 2020-02-03
RCOCOA-532 - property_copyAttributeList produces memory leak

### LINKED ISSUE #6409: getSwiftProperties Memory Leak

leemaguire:
Fixes #6409

The issue was that property_copyAttributeList requires you to manually free allocated memory. To solve this we are using free() and this is called in a defer block.

To test this fix I made a test app which starts Realm when a button is tapped. The current master branch would produce a memory leak when the button was tapped. The code in this branch now stops the memory leak from happening. I used Instruments to inspect.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,8 +5,7 @@ x.y.z Release notes (yyyy-MM-dd)
   the block passed to it returns. Returning `Void` from the block is still allowed.
 
 ### Fixed
-* <How to hit and notice issue? what was the impact?> ([#????](https://github.com/realm/realm-js/issues/????), since v?.?.?)
-* None.
+* Fix a memory leak attributed to `property_copyAttributeList` the first time a Realm is opened when using Realm Swift. (#6409, since 4.0.0).
 
 <!-- ### Breaking Changes - ONLY INCLUDE FOR NEW MAJOR version -->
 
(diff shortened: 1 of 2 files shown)


bmunkholm:
Needs an update in the changelog.md file.

tgoyne:
Since this is a user-created public issue, we'll want to close it via github rather than via jira. This means that instead of having the internal ticket number in the title, it should have "Fixes #6409" in the body of the PR so that github's linking works.

leemaguire:
@tgoyne PTAL

CHANGELOG.md
```diff
@@ -20,6 +20,20 @@ x.y.z Release notes (yyyy-MM-dd)
 Upgraded realm-core from ? to ?
 Upgraded realm-sync from ? to ?
 
+4.3.2 Release notes (2020-02-04)
```
tgoyne: The changelog entry should go in the above section, not add a new one. The flow for the changelog is that each PR adds entries to the x.y.z section for the next release, and then as part of the release process that is replaced with the actual version number and release date. After the release is complete, a new x.y.z section is added to the top of the changelog.

CHANGELOG.md
```diff
@@ -20,6 +20,20 @@ x.y.z Release notes (yyyy-MM-dd)
...
+=============================================================
+### Enhancements
+* Fix for memory leak caused by `property_copyAttributeList`
```
tgoyne: There's no need to list this in both enhancements and bug fixes.

CHANGELOG.md
```diff
@@ -20,6 +20,20 @@ x.y.z Release notes (yyyy-MM-dd)
...
+
+### Fixed
+* property_copyAttributeList was not being released after being assigned to a local variable.
```
tgoyne: The changelog is targeted at developers using Realm, and should focus on what symptoms they'd see rather than the exact change made to fix it. For example, here I'd write something like "Fix a memory leak attributed to `property_copyAttributeList` the first time a Realm is opened when using Realm Swift. ([#6409](https://github.com/realm/realm-cocoa/issues/6409), since 4.0.0)." The main thing a person reading the changelog wants to know is if upgrading to the new version will fix a problem they're hitting; if they want to know more about the specific details of the change then the link to the issue helps them find that.

leemaguire: @tgoyne noted and updated with your feedback.

PR #6427, opened 2020-02-13
Add support for frozen objects

tgoyne:
I think this exactly matches the API and semantics of the Java implementation outside of spelling differences, so see the [PR there](https://github.com/realm/realm-java/pull/6590) for information on design decisions.

### DIFF
--- a/Realm/RLMArray.h
+++ b/Realm/RLMArray.h
@@ -91,6 +91,15 @@ NS_ASSUME_NONNULL_BEGIN
  */
 @property (nonatomic, readonly, getter = isInvalidated) BOOL invalidated;
 
+/**
+ Indicates if the array is frozen.
+
+ Frozen arrays are immutable and can be accessed from any thread. Frozen arrays
+ are created by calling `-freeze` on a managed live array. Unmanaged arrays are
+ never frozen.
+ */
+@property (nonatomic, readonly, getter = isFrozen) BOOL frozen;
+
 #pragma mark - Accessing Objects from an Array
 
 /**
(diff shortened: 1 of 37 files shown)


winstondu:
@tgoyne ,

How does freezing affect one-to-many relationships?

For example, if I have:
```
typealias ID = String

class User: Object {
    @objc dynamic var ID: ID!
    let friends = List<User>()

    // For Realm
        return "ID"
    }

    convenience required init(ID: ID) {
        self.init()
        self.ID = ID
    }
}
```

What does the following return?
```
// user is of type User, and let's assume user's friend list is not empty
user.freeze().friends.first?.isFrozen
```

tgoyne:
All objects obtained from a frozen object are themselves frozen as well.

winstondu:
> All objects obtained from a frozen object are themselves frozen as well.

Yup, I just experimented and that seems to be the case.

@tgoyne , Now, for efficiency, are the objects obtained from a frozen object lazy generated, or is it the case that when I freeze even a single object, I could be making an expensive copy of the reachable object graph from the object I froze?

I hope it is the former.

tgoyne:
Everything is lazy. Frozen objects work the same as regular objects and read data from the file as needed on a per-property basis, with the only difference being that they never update the version of the data which they read. The fact that it's lazy is sort of an implementation detail (there shouldn't be any observable effects from it being lazy other than that freezing objects doesn't take a long time and a bunch of memory), but also isn't in that it's the main reason to use the feature rather than manually copying objects out of the Realm.

winstondu:
So in other words, if I do the following:

```
let frozenUser = user.freeze()
```

I could get different values for `frozenUser.friends.count` depending on when I query it (due to the fact that the db may have changed in between queries of the friend count)

tgoyne:
The entire object graph is frozen and you will not see different values after writes are made.

winstondu:
> The entire object graph is frozen and you will not see different values after writes are made.

So I will always see the same value for `frozenUser.friends.count`, regardless of how I make writes to the actual db afterwards?

tgoyne:
Yes.

PR #6437, opened 2020-02-27
Migrate Core and Sync to 10.0.0-alpha.1

jsflax:
This PR largely removes vestigial code tied to Sync Subscriptions, removes login from SyncUser (to be replaced by the new App::login), and removes references to full or partial sync.

There are stubs in place where unboxing for Decimal128s, ObjectIds, and Embedded Objects will occur.

Tests related to authentication and sync have temporarily been decommissioned or removed.

NOTE: This will not pass on the CI. `alpha.2` needs to be released for sync, and any test involving authentication will not pass until authentication is migrated. This does, however, leave everything in a compilable state, and unblocks work related to the new Realm Applications.


Realm/RLMRealm.mm
```diff
@@ -281,21 +242,7 @@ + (RLMAsyncOpenTask *)asyncOpenWithConfiguration:(RLMRealmConfiguration *)config
...
-                    @autoreleasepool {
-                        waitForPartialSyncSubscriptions(realm->config());
-                        complete();
-                    }
-                });
-            }
-            else {
-                complete();
-            }
+            complete();
```
tgoyne: The `auto realm = Realm::get_shared_realm(std::move(ref));` line above should also go away and this block should just be inlined since it's no longer called from multiple places.

PR #6439, opened 2020-02-28
Add RealmApp structure; Add basic login

jsflax:
Changes:
- Add RLMApp, the Realm App structure.
- Add RLMAppConfiguration, the Realm App config structure.
- Add CocoaNetworkTransport and the RLMNetwork Transporting protocol. This hooks into the object store so that the object store can take advantage of the platform level network layer.
- Add loginWithCredentials, the functionality to log in to a user.
- Add stubs for multiple user semantics.
- Remove vestigial functionality from RLMSyncUser.

Important notes!
- Many tests have been disabled or rendered moot. 
- This is not well tested. The smoke test for login will pass, but multiple user semantics will fail.
- Stubs have been added for pieces missing related to the new data types.
- Swift glue has not been added.

This pull request grew far too much in size. Hoping to be able to push this through and then iterate over it.


Realm/ObjectServerTests/RLMObjectServerTests.mm
```diff
@@ -110,8 +137,8 @@ - (void)testInvalidPasswordAuthentication {
 
 /// A non-existsing user should not be able to log in and a corresponding error should be generated.
 - (void)testNonExistingUsernameAuthentication {
-    RLMSyncCredentials *credentials = [RLMSyncTestCase basicCredentialsWithName:NSStringFromSelector(_cmd)
-                                                                       register:NO];
+    RLMAppCredentials *credentials = [RLMAppCredentials credentialsWithUsername:@"INVALID_USERNAME"
+    password:@"INVALID_PASSWORD"];
```
tgoyne: There's a bunch of places here where subsequent lines are indented incorrect or not at all.

Realm/ObjectServerTests/RLMSyncTestCase.mm
```diff
@@ -67,7 +67,7 @@ @interface RLMSyncTestCase ()
 @property (nonatomic) NSTask *task;
 @end
 
-@interface RLMSyncCredentials ()
+@interface RLMAppCredentials ()
```
tgoyne: This (and createAdminUserForURL) presumably aren't used any more since the debug admin provider isn't a thing?

Realm/RLMApp.h
```diff
@@ -0,0 +1,97 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+////////////////////////////////////////////////////////////////////////////
+
+#import "RLMNetworkTransporting.h"
+
+#ifndef RLMApp_h
```
tgoyne: We don't use header guards for obj-c code since `#import` takes care of that.

Realm/RLMApp.h
```diff
@@ -0,0 +1,97 @@
...
+typedef void(^RLMUserCompletionBlock)(RLMSyncUser * _Nullable, NSError * _Nullable);
+
+#pragma mark RLMAppConfiguration
+
+/// Properties representing the configuration of a client
+/// that communicate with a particular Realm application.
+@interface RLMAppConfiguration : NSObject
+
+/// A custom base URL to request against.
+@property (nonatomic, strong) NSString* _Nullable baseURL;
```
tgoyne: Nullable properties should be marked with `nullable` in the property attributes list rather than annotating the pointer.

Realm/RLMApp.h
```diff
@@ -0,0 +1,97 @@
...
+/// A custom app name.
+@property (nonatomic, strong) NSString* _Nullable localAppName;
+
+/// A custom app version.
+@property (nonatomic, strong) NSString* _Nullable localAppVersion;
+
+/// The default timeout for network requests.
+@property (nonatomic, assign) NSUInteger defaultRequestTimeoutMS;
+
+-(instancetype _Nonnull) initWithBaseURL:(NSString* _Nullable) baseURL
```
tgoyne: Each header file should normally be wrapped in `NS_ASSUME_NONNULL_BEGIN`/`NS_ASSUME_NONNULL_END` so that things are non-null by default rather than marking individual things as nonnull.

Space between - and (, no space between ) and argument name.

Realm/RLMApp.h
```diff
@@ -0,0 +1,97 @@
...
+/**
+ Get an application with a given appId and configuration.
+
+ @param appId The unique identifier of your Realm app.
+ @param configuration A configuration object to configure this client.
+ */
++(_Nonnull instancetype) app:(NSString * _Nonnull) appId
+               configuration:(RLMAppConfiguration * _Nullable)configuration;
+
+- (NSDictionary * _Nonnull)allUsers;
```
tgoyne: This dictionary should have a generic type indicating what's in it.

Realm/RLMApp.mm
```diff
@@ -0,0 +1,178 @@
...
+        self.localAppVersion = localAppVersion;
+        self.defaultRequestTimeoutMS = defaultRequestTimeoutMS;
+        return self;
+    }
+    return nil;
+}
+@end
+
+/// A map of ObjectStore HTTP methods to RLMHTTPMethods
+std::map<app::HttpMethod, RLMHTTPMethod> HttpMethod_toRLMHTTPMethod = {
```
tgoyne: Rather than building a lookup table and translating at runtime, we normally just give the two enums the same values and cast between them, with `static_assert()`s verifying that the values actually are the same. See RLMPropertyType/realm::PropertyType for an example of this (RLMConstants.h and RLMProperty.mm have the relevant bits).

Realm/RLMApp.mm
```diff
@@ -0,0 +1,178 @@
...
+    {app::HttpMethod::put, PUT},
+    {app::HttpMethod::patch, PATCH},
+    {app::HttpMethod::del, DELETE}
+};
+
+/// Internal transport struct to bridge RLMNetworkingTransporting to the GenericNetworkTransport.
+struct CocoaNetworkTransport : public realm::app::GenericNetworkTransport {
+    static id <RLMNetworkTransporting> transport;
+
+    void send_request_to_server(const app::Request request,
```
tgoyne: The request and response should be passed by const reference rather than copying them.

Realm/RLMApp.mm
```diff
@@ -0,0 +1,178 @@
...
+            headers[@(header.first.data())] = @(header.second.data());
+        }
+        rlmRequest.headers = headers;
+        rlmRequest.method = HttpMethod_toRLMHTTPMethod[request.method];
+        rlmRequest.timeoutMS = request.timeout_ms;
+
+        // Send the request through to the Cocoa level transport
+        [transport sendRequestToServer:rlmRequest completion:^(RLMResponse * _Nonnull response) {
+            std::map<std::string, std::string> bridgingHeaders;
+            for (id key in response.headers) {
```
tgoyne:
```
[response.headers enumerateKeysAndObjectsUsingBlock:^(NSString *key, NSString *value, BOOL *) {
    bridgingHeaders[key.UTF8String] = value.UTF8String;
}];
```

Realm/RLMApp.mm
```diff
@@ -0,0 +1,178 @@
...
+                bridgingHeaders[[key cStringUsingEncoding:NSUTF8StringEncoding]] = [response.headers[key] cStringUsingEncoding:NSUTF8StringEncoding];
+            }
+
+            // Convert the RLMResponse to an app:Response and pass downstream to
+            // the object store
+            completion(app::Response {
```
tgoyne: no space before { for object initialization

Realm/RLMApp.mm
```diff
@@ -0,0 +1,178 @@
...
+                .body = [response.body cStringUsingEncoding:NSUTF8StringEncoding],
```
tgoyne: response.body.UTF8String

Realm/RLMApp.mm
```diff
@@ -0,0 +1,178 @@
...
+std::map<app::HttpMethod, RLMHTTPMethod> HttpMethod_toRLMHTTPMethod = {
+    {app::HttpMethod::get, GET},
+    {app::HttpMethod::post, POST},
+    {app::HttpMethod::put, PUT},
+    {app::HttpMethod::patch, PATCH},
+    {app::HttpMethod::del, DELETE}
+};
+
+/// Internal transport struct to bridge RLMNetworkingTransporting to the GenericNetworkTransport.
+struct CocoaNetworkTransport : public realm::app::GenericNetworkTransport {
```
tgoyne: This should be inside an anonymous namespace as it's a local type used only in this file.

Realm/RLMApp.mm
```diff
@@ -0,0 +1,178 @@
...
+        }];
+    }
+};
+
+id <RLMNetworkTransporting> CocoaNetworkTransport::transport = [RLMNetworkTransport new];
+
+@implementation RLMApp : NSObject
+
++(NSMutableDictionary<NSString *, RLMApp*>*)apps {
+    static NSMutableDictionary<NSString *, RLMApp*>* _apps = nil;
```
tgoyne:
```
static NSMutableDictionary<NSString *, RLMApp *> *apps = [NSMutableDictionary new];
returns apps;
```

Realm/RLMApp.mm
```diff
@@ -0,0 +1,178 @@
...
+            });
+        }];
+    }
+};
+
+id <RLMNetworkTransporting> CocoaNetworkTransport::transport = [RLMNetworkTransport new];
+
+@implementation RLMApp : NSObject
+
++(NSMutableDictionary<NSString *, RLMApp*>*)apps {
```
tgoyne: This probably should be a free function and not a method on RLMApp? It doesn't need anything from RLMApp and shouldn't be exposed as part of RLMApp's interface.

Realm/RLMApp.mm
```diff
@@ -0,0 +1,178 @@
...
+    return _apps;
+}
+
+-(instancetype) initWithAppId:(NSString *) appId configuration:(RLMAppConfiguration *)configuration {
+    if (self = [super init]) {
+        app::App::Config boundConfiguration = {
+            .app_id = [appId cStringUsingEncoding:NSUTF8StringEncoding]
+        };
+        boundConfiguration.transport_generator = []{
+            return std::unique_ptr<app::GenericNetworkTransport>(new CocoaNetworkTransport);
```
tgoyne: `[]() -> std::unique_ptr<GenericNetworkTransport> { return std::make_unique<CocoaNetworkTransport>(); };`

Realm/RLMApp.mm
```diff
@@ -0,0 +1,178 @@
...
+        };
+        if (configuration) {
+            if (configuration.baseURL) {
+                boundConfiguration.base_url = util::Optional<std::string>([configuration.baseURL cStringUsingEncoding:NSUTF8StringEncoding]);
+            }
+            if (configuration.transport) {
+                CocoaNetworkTransport::transport = configuration.transport;
```
tgoyne: This doesn't look like it needs to be a global variable? `configuration.transport` could just be captured by the `transport_generator` lambda and then stored as a normal member of `CocoaNetworkTransport`.

Realm/RLMApp.mm
```diff
@@ -0,0 +1,178 @@
...
+            }
+            boundConfiguration.local_app_name = std::string([configuration.localAppName cStringUsingEncoding:NSUTF8StringEncoding]);
+            boundConfiguration.local_app_version = std::string([configuration.localAppVersion cStringUsingEncoding:NSUTF8StringEncoding]);
+            boundConfiguration.default_request_timeout_ms = (uint64_t)configuration.defaultRequestTimeoutMS;
+        }
+        self->_app = std::make_shared<app::App>(boundConfiguration);
+        [[RLMApp apps] setValue:self forKey:appId];
```
tgoyne: `[RLMApp apps][appId] = self`

Realm/RLMApp.mm
```diff
@@ -0,0 +1,178 @@
...
+    return [[NSError alloc] initWithDomain:@(appError.error_code.category().name())
+                                      code:appError.error_code.value()
+                                  userInfo:@{
+        @(appError.error_code.category().name()) : @(appError.error_code.message().data())
+    }];
+}
+
+-(void) loginWithCredential:(RLMAppCredentials *)credentials
+          completionHandler:(RLMUserCompletionBlock)completionHandler {
+    self->_app->login_with_credentials(*credentials.appCredentials, ^(std::shared_ptr<SyncUser> user, util::Optional<app::AppError> error) {
```
tgoyne: We don't normally self-qualify ivar access as the leading underscore already indicates it's an ivar.

Realm/RLMAppCredentials.h
```diff
@@ -0,0 +1,89 @@
+////////////////////////////////////////////////////////////////////////////
+//
+// Copyright 2016 Realm Inc.
```
tgoyne: 2020

Realm/RLMAppCredentials_Private.hpp
```diff
@@ -0,0 +1,31 @@
...
+
+#ifndef RLMAppCredentials_Private_hpp
+#define RLMAppCredentials_Private_hpp
+
+#import "RLMAppCredentials.h"
+#import "sync/app_credentials.hpp"
+
+@interface RLMAppCredentials()
+
+@property std::shared_ptr<realm::app::AppCredentials> appCredentials;
```
tgoyne: There doesn't appear to be any reason for this to be a shared_ptr. It should also be marked as `nonatomic`.

Realm/RLMApp_Private.hpp
```diff
@@ -0,0 +1,33 @@
...
+////////////////////////////////////////////////////////////////////////////
+
+#ifndef RLMApp_Private_h
+#define RLMApp_Private_h
+
+#include "RLMApp.h"
+#import "sync/app.hpp"
+
+@interface RLMApp() {
+    std::shared_ptr<realm::app::App> _app;
```
tgoyne: This doesn't appear to need to be exposed?

Realm/RLMNetworkTransporting.h
```diff
@@ -0,0 +1,97 @@
...
+NS_ASSUME_NONNULL_BEGIN
+
+@interface RLMNetworkRequestOptions : NSObject
+@property (nonatomic, copy, nullable) NSString *authorizationHeaderName;
+@property (nonatomic, copy, nullable) NSDictionary<NSString *, NSString *> *customHeaders;
+@property (nullable, nonatomic, copy) NSDictionary<NSString *, NSURL *> *pinnedCertificatePaths;
+@end
+
+typedef enum RLMHTTPMethod {
+    GET, POST, PUT, PATCH, DELETE
```
tgoyne: These are very generic names to put into the global namespace and are basically guaranteed to cause conflicts. Enums should also be declared with `RLM_CLOSED_ENUM()` (or `NS_ENUM` for open enums).

Realm/RLMNetworkTransporting.h
```diff
@@ -0,0 +1,97 @@
...
+} RLMHTTPMethod;
+
+/// An HTTP request that can be made to an arbitrary server.
+@interface RLMRequest : NSObject
+
+/// The HTTP method of this request.
+@property (nonatomic, assign) RLMHTTPMethod method;
+
+/// The URL to which this request will be made.
+@property (nonatomic, strong) NSString* url;
```
tgoyne: Space before * rather than after in obj-c code (this is different from the pure c++ repos because we were trying to look like "normal" obj-c code and not obj-c code written by c++ people).

Realm/RLMNetworkTransporting.h
```diff
@@ -0,0 +1,97 @@
...
+
+/// The number of milliseconds that the underlying transport should spend on an
+/// HTTP round trip before failing with an error.
+@property (nonatomic, assign) NSUInteger timeoutMS;
```
tgoyne: `NSTimeInterval timeout` would be more idiomatic here.

Realm/RLMNetworkTransporting.h
```diff
@@ -0,0 +1,97 @@
...
+/// The body of the HTTP response.
+@property (nonatomic, strong) NSString* body;
+
+@end
+
+typedef void(^RLMNetworkTransportCompletionBlock)(RLMResponse *);
+
+/// Transporting protocol for foreign interfaces. Allows for custom
+/// request/response handling.
+@protocol RLMNetworkTransporting <NSObject>
```
tgoyne: This is a strange name for a protocol. It should just be `RLMNetworkTransport`.

Realm/RLMNetworkTransporting.mm
```diff
@@ -0,0 +1,248 @@
+////////////////////////////////////////////////////////////////////////////
+//
+// Copyright 2016 Realm Inc.
```
tgoyne: 2020

Realm/RLMNetworkTransporting.mm
```diff
@@ -0,0 +1,248 @@
...
+typedef void(^RLMServerURLSessionCompletionBlock)(NSData *, NSURLResponse *, NSError *);
+
+#pragma mark RLMSessionDelegate
+
+@interface RLMSessionDelegate <NSURLSessionDelegate> : NSObject
++ (instancetype)delegateWithCertificatePaths:(NSDictionary *)paths
+                                  completion:(RLMNetworkTransportCompletionBlock)completion;
+@end
+
+NSString * const RLMHTTPMethod_toString[] = {
```
tgoyne: This isn't a naming scheme we use anywhere.

Realm/RLMSyncUser.h
```diff
@@ -83,29 +83,26 @@ NS_ASSUME_NONNULL_BEGIN
 @interface RLMSyncUser : NSObject
 
 /**
- A dictionary of all valid, logged-in user identities corresponding to their user objects.
+ The unique Realm Object Server user ID string identifying this user.
```
tgoyne: These references to ROS seem out of date?

Realm/RLMSyncUser.mm
```diff
@@ -159,6 +143,20 @@ - (void)logOut {
     _user->log_out();
     context_for(_user).invalidate_all_handles();
 }
+#import "RLMAccessor.h"
```
tgoyne: Header imports should be at the top of the file.

Realm/RLMSyncUser.mm
```diff
@@ -159,6 +143,20 @@ - (void)logOut {
...
+
+- (NSDictionary*)customData {
+    if (!_user || !_user->custom_data()) {
+        return @{};
+    }
+
+    NSMutableDictionary *dictionary = [NSMutableDictionary new];
+    for (auto value : *_user->custom_data()) {
```
tgoyne: This is copying each of the values. `auto& value` to avoid that.

Realm/RLMUtil.mm
```diff
@@ -390,6 +391,25 @@ BOOL RLMIsRunningInPlayground() {
     return [[NSBundle mainBundle].bundleIdentifier hasPrefix:@"com.apple.dt.playground."];
 }
 
+std::map<size_t, id(^)(realm::util::Any const&)> cppToObjC_Mappings = {
```
tgoyne: This is a very strange way to switch on the type which is verbose, slow, and not guaranteed to work (both types could legally have the same hash code).

The one place that is using this should probably actually be using `realm::Mixed`. `Any` is only appropriate when the value actually could be literally any type, not when it could be one of a fixed set of supported types.

Realm/RLMApp_Private.hpp
tgoyne: Also doesn't appear to need to be a shared_ptr.

Realm/RLMNetworkTransporting.h
```diff
@@ -0,0 +1,97 @@
...
+
+@end
+
+/// The contents of an HTTP response.
+@interface RLMResponse : NSObject
+
+/// The status code of the HTTP response.
+@property (nonatomic, assign) NSInteger httpStatusCode;
+
+/// A custom status code provided by the language binding.
```
bmunkholm:
```suggestion
/// A custom status code provided by the SDK.
```

Realm/RLMUtil.mm
jsflax: In retrospect, this will not work any way. It seems we should not expose `customData` until we are able to handle xjson– none of these types will map properly until then. Going to remove this for now.

Realm/RLMSyncUser.h
jsflax: Going to remove all ROS references in a future PR.

jsflax:
@tgoyne think I addressed everything. PTAL.

Realm.xcodeproj/project.pbxproj
```diff
@@ -2849,6 +2861,7 @@
 				DEFINES_MODULE = YES;
 				LD_RUNPATH_SEARCH_PATHS = "$(inherited) @executable_path/Frameworks @loader_path/Frameworks";
 				SWIFT_OPTIMIZATION_LEVEL = "-Onone";
+				SYSTEM_HEADER_SEARCH_PATHS = "\"$(SRCROOT)/Realm/ObjectStore/external/json\"";
```
tgoyne: This is set in the xcconfig file and shouldn't also be in the project.

Realm.xcodeproj/project.pbxproj
```diff
@@ -2918,6 +2932,7 @@
 		C04B4BD81B55C47600FAE58E /* Debug */ = {
 			isa = XCBuildConfiguration;
 			buildSettings = {
+				IPHONEOS_DEPLOYMENT_TARGET = 8.0;
```
tgoyne: Incorrect merge?

Realm/RLMApp.h
```diff
@@ -0,0 +1,96 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+////////////////////////////////////////////////////////////////////////////
+
+#import "RLMNetworkTransport.h"
```
tgoyne: `@protocol RLMNetworkTransport`` suffices here and we don't need to drag in the header.

Realm/RLMApp.mm
```diff
@@ -0,0 +1,171 @@
...
+    realm::app::App _app;
+}
+@end
+
+@implementation RLMApp : NSObject
+
+- (instancetype)initWithAppId:(NSString *)appId configuration:(RLMAppConfiguration *)configuration {
+    if (self = [super init]) {
+        app::App::Config boundConfiguration = {
+            .app_id = [appId cStringUsingEncoding:NSUTF8StringEncoding]
```
tgoyne: All of these should be `.UTF8String`

Realm/RLMApp.mm
```diff
@@ -0,0 +1,171 @@
...
+}
+
+- (RLMSyncUser *)currentUser {
+    return [[RLMSyncManager sharedManager] _currentUser];
+}
+
+/**
+ Convert an object store AppError to an NSError.
+ */
+static NSError* RLMAppErrorToNSError(const app::AppError& appError) {
```
tgoyne: `RLM` prefix is for exported symbols so this shouldn't have it.

Realm/RLMAppCredentials_Private.hpp
```diff
@@ -0,0 +1,31 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+////////////////////////////////////////////////////////////////////////////
+
+#ifndef RLMAppCredentials_Private_hpp
```
tgoyne: No header guards in obj-c(++) code.

Realm/RLMApp_Private.hpp
```diff
@@ -0,0 +1,27 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+////////////////////////////////////////////////////////////////////////////
+
+#ifndef RLMApp_Private_h
+#define RLMApp_Private_h
+
+#include "RLMApp.h"
```
tgoyne: `#import`

Realm/RLMNetworkTransport.h
```diff
@@ -0,0 +1,101 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+////////////////////////////////////////////////////////////////////////////
+
+#import <Foundation/Foundation.h>
+
+#import "RLMSyncUtil_Private.h"
```
tgoyne: Public headers can't import private ones. The private ones aren't included in the packaged framework at all.

Realm.xcodeproj/project.pbxproj
```diff
@@ -246,6 +236,18 @@
...
+		49B2998A2409338200B7319C /* RLMApp.mm in Sources */ = {isa = PBXBuildFile; fileRef = 49B299862409338200B7319C /* RLMApp.mm */; };
+		49B2998B2409338200B7319C /* RLMApp.h in Headers */ = {isa = PBXBuildFile; fileRef = 49B299872409338200B7319C /* RLMApp.h */; settings = {ATTRIBUTES = (Public, ); }; };
+		49B299902409351600B7319C /* app.cpp in Sources */ = {isa = PBXBuildFile; fileRef = 49B2998C2409351500B7319C /* app.cpp */; };
+		49B299912409351600B7319C /* app.hpp in Headers */ = {isa = PBXBuildFile; fileRef = 49B2998D2409351500B7319C /* app.hpp */; settings = {ATTRIBUTES = (Public, ); }; };
+		49B299922409351600B7319C /* app_credentials.cpp in Sources */ = {isa = PBXBuildFile; fileRef = 49B2998E2409351600B7319C /* app_credentials.cpp */; };
+		49B299932409351600B7319C /* app_credentials.hpp in Headers */ = {isa = PBXBuildFile; fileRef = 49B2998F2409351600B7319C /* app_credentials.hpp */; settings = {ATTRIBUTES = (Public, ); }; };
+		49B299962409351B00B7319C /* generic_network_transport.hpp in Headers */ = {isa = PBXBuildFile; fileRef = 49B299942409351B00B7319C /* generic_network_transport.hpp */; settings = {ATTRIBUTES = (Public, ); }; };
```
tgoyne: These headers are in the wrong section, which indicates that they're probably incorrectly part of the Realm target when they shouldn't be.

Realm.xcodeproj/project.pbxproj
```diff
@@ -246,6 +236,18 @@
 		3FEB38401E70AC8800F22712 /* ObjectCreationTests.mm in Sources */ = {isa = PBXBuildFile; fileRef = 3FEB383C1E70AC6900F22712 /* ObjectCreationTests.mm */; };
 		3FEC4A3F1BBB18D400F009C3 /* SwiftSchemaTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 3FEC4A3D1BBB188B00F009C3 /* SwiftSchemaTests.swift */; };
 		3FF3FFAF1F0D6D6400B84599 /* KVOTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 5D660FFF1BE98D880021E04F /* KVOTests.swift */; };
+		49B299882409338200B7319C /* RLMAppCredentials.mm in Sources */ = {isa = PBXBuildFile; fileRef = 49B299842409338200B7319C /* RLMAppCredentials.mm */; };
+		49B299892409338200B7319C /* RLMAppCredentials.h in Headers */ = {isa = PBXBuildFile; fileRef = 49B299852409338200B7319C /* RLMAppCredentials.h */; settings = {ATTRIBUTES = (Public, ); }; };
+		49B2998A2409338200B7319C /* RLMApp.mm in Sources */ = {isa = PBXBuildFile; fileRef = 49B299862409338200B7319C /* RLMApp.mm */; };
```
tgoyne: The newly added source files need to be added to the Realm iOS Static target in addition to the Realm target.

jsflax:
@tgoyne 🤞

Configuration/Base.xcconfig
```diff
@@ -52,7 +52,7 @@ OTHER_SWIFT_FLAGS = $(REALM_CATALYST_FLAGS);
 
 OTHER_CPLUSPLUSFLAGS = $(inherited) -isystem core/include -isystem Realm/ObjectStore/external/json;
 HEADER_SEARCH_PATHS = $(inherited) Realm/ObjectStore/src;
-
+SYSTEM_HEADER_SEARCH_PATHS = $(SRCROOT)/Realm/ObjectStore/external/json;
```
tgoyne: This is already being set on the line above and doesn't need to be added.

PR #6445, opened 2020-03-03
Remove the sync permissions and privileges code

tgoyne:
This conflicts a little with #6439 but since it's almost entirely just deleting things it should be easy to resolve, and deleting all this stuff is required to update the objectstore version.


Package.swift
```diff
@@ -3,7 +3,7 @@
 import PackageDescription
 import Foundation
 
-let coreVersionStr = "6.0.3"
+let coreVersionStr = "10.0.0-alpha.1"
 let cocoaVersionStr = "5.0.0-beta.3"
```
jsflax: [q] Should we up this version as well?

tgoyne: That's done as part of making a release.

PR #6450, opened 2020-03-09
Add support for object id and decimal128 property types

tgoyne:
This depends on https://github.com/realm/realm-core/pull/3654 and so won't build on CI.

This adds support for ObjectID and Decimal128 as property types, but doesn't try to implement an actual API for those types beyond what the tests needed. This means, for example, that you can't actually do math with a Decimal128. Fleshing out those types so that they're more than an opaque wrapper for the underlying types and properly documenting them will come in a future PR, as this one is bloated enough already.

PrimitiveArrayPropertyTests.m is a generated file (produced by PrimitiveArrayPropertyTests.m.tpl and tests.py) so no need to look at it when reviewing.


Realm/RLMDecimal128.h
```diff
@@ -0,0 +1,40 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+////////////////////////////////////////////////////////////////////////////
+
+#import <Foundation/Foundation.h>
+
+NS_ASSUME_NONNULL_BEGIN
+
+@interface RLMDecimal128 : NSObject
```
jsflax: [docs] This class and its methods should have documentation.

Realm/RLMDecimal128.mm
```diff
@@ -0,0 +1,146 @@
...
+    if ((self = [self init])) {
+        _value = RLMObjcToDecimal128(value);
+    }
+    return self;
+}
+
+- (instancetype)initWithNumber:(NSNumber *)number {
+    if ((self = [self init])) {
+        auto type = number.objCType[0];
+        // FIXME: NSDecimalNumber
```
jsflax: [q] Why can't this be done currently?

Realm/RLMObjectId.mm
```diff
@@ -0,0 +1,86 @@
...
+        }
+    }
+    return self;
+}
+
+- (instancetype)initWithString:(NSString *)string error:(NSError **)error {
+    if ((self = [self init])) {
+        auto str = string.UTF8String;
+        auto len = strlen(str);
+        if (len != 24 || !std::all_of(str, str + len, [](unsigned char c) { return std::isxdigit(c); })) {
```
jsflax: [q] Shouldn't this check be done in core?

Realm/RLMQueryUtil.mm
```diff
@@ -877,6 +878,23 @@ void process_or_group(Query &query, id array, Func&& func) {
...
 }
 
+template <>
+Decimal128 convert<Decimal128>(id value) {
+    return RLMObjcToDecimal128(value);
+}
+
+template <>
+ObjectId convert<ObjectId>(id value) {
+    // FIXME: extract this somewhere and share with object creation
```
jsflax: Right, shouldn't this also account for number types?

Realm/RLMRealm.mm
```diff
@@ -905,6 +905,7 @@ - (BOOL)isFrozen {
 }
 
 - (RLMRealm *)freeze {
+    [self verifyThread];
```
jsflax: [👍 ]

Realm/RLMSyncSessionRefreshHandle.mm
```diff
@@ -270,12 +273,11 @@ - (void)_timerFired {
...
-// FIXME: if this should be handled in the OS or at this layer.
-//    [RLMSyncAuthEndpoint sendRequestToServer:self.authServerURL
-//                                        JSON:json
-//                                     timeout:60.0
-//                                  completion:handler];
+    [RLMSyncAuthEndpoint sendRequestToServer:self.authServerURL
+                                        JSON:json
+                                     timeout:60.0
+                                  completion:handler];
+#endif
```
jsflax: I have investigated this, there are several tasks that need to be completed first before this is fixed.

Realm/Tests/ObjectTests.m
```diff
@@ -325,11 +334,13 @@ - (void)testDataTypes {
     AllTypesObject *o = [[AllTypesObject alloc] initWithValue:row1];
     o.floatCol = NAN;
     o.doubleCol = NAN;
+//    o.decimalCol = [RLMDecimal128 decimalWithNumber:[NSDecimalNumber notANumber]];
```
jsflax: If this isn't working, can we leave a FIXME?

Realm/Tests/tests.py
```diff
@@ -1,58 +1,69 @@
 import os, re
```
jsflax: [q] What is this file?

RealmSwift/Util.swift
```diff
@@ -173,6 +173,20 @@ extension Optional: CustomObjectiveCBridgeable {
...
+    static func bridging(objCValue: Any) -> Decimal128 {
+        if let decimal = objCValue as? NSDecimalNumber {
+            return Decimal128(nsDecimal: decimal)
+        }
+        if let number = objCValue as? NSNumber {
+            return Decimal128(number: number)
+        }
+        return objCValue as! Self
+    }
+    var objCValue: Any {
```
jsflax: [style] missing newline

Realm/RLMUtil.mm
```diff
@@ -188,6 +190,12 @@ BOOL RLMValidateValue(__unsafe_unretained id const value,
             RLMObjectBase *objBase = RLMDynamicCast<RLMObjectBase>(value);
             return objBase && [objBase->_objectSchema.className isEqualToString:objectClassName];
         }
+        case RLMPropertyTypeObjectId:
+            return [value isKindOfClass:[RLMObjectId class]];
+        case RLMPropertyTypeDecimal128:
+            return [value isKindOfClass:[NSNumber class]]
+                || [value isKindOfClass:[RLMDecimal128 class]]
+                || ([value isKindOfClass:[NSString class]] && realm::Decimal128::is_valid_str([value UTF8String]));
```
jsflax: I can't build the branch because of the `is_valid_str` method. Where is this?

tgoyne: The core PR linked in the description.

Realm/RLMAccessor.mm
```diff
@@ -252,6 +286,13 @@ id makeNumberGetter(NSUInteger index, bool boxed, bool optional) {
     }
     return makeGetter<Type>(index);
 }
+template<typename Type>
+id makeWrapperGetter(NSUInteger index, bool optional) {
```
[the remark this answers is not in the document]
tgoyne: Prior to this, there's 5 categories of properties which we supported:
1. Primitive numbers, which don't box at all for non-optional. Optionals are wrapped in util::Optional and converted to NSNumber
2. Types like StringData which are handled identically for optional and non-optional due to being able to store null internally.
3. Lists
4. Object links
5. LinkingObjects

ObjectId doesn't fall into any of these categories: unlike primitive numbers it needs to be boxed for non-optional, and unlike the other non-primitive atomic types it can't represent null and needs to be wrapped in util::Optional for that.

Realm/RLMObjectId.mm
tgoyne: Core currently just asserts that it's valid, but yeah I should add a non-asserting check to ObjectId.

Realm/RLMSyncSessionRefreshHandle.mm
tgoyne: This change is just shuffling around how much is commented out to eliminate the unused variable warnings.

Realm/Tests/tests.py
tgoyne: This file generates PrimitiveArrayPropertyTests.m.

Realm/RLMQueryUtil.mm
tgoyne: What conversion would make sense for number to objectid?

jsflax: Going to resolve this– can't remember the context of why I put this here.

PR #6456, opened 2020-03-24
Remove Xcode 10.0 and 10.1 and add Xcode 11.4

tgoyne:
Xcode 10.1 was the last version to support Swift 3.4, so this also means we can drop Swift 3 support.


doncl:
Thanks, @tgoyne !   It's all working now, building, I'm building our app in Xcode 11.4 and deploying to sim and device, very nice!

PR #6472, opened 2020-04-14
Remove ROS; Add baas

jsflax:
Pulls in mongo, go, stitch (baas), and all of its dependencies.

This will setup at runtime and clean itself up at exit. There should be no remaining state. Tests do not share state either as they operate on separate baas apps.

Tests should be passing now.

TODO: Once our CI is provisioned properly, we should run checks in the script to determine if mongo/go/baas already exist, so as to not set them up again.


Realm/ObjectServerTests/run_baas.sh
```diff
@@ -0,0 +1,140 @@
...
+    chmod +x assisted_agg
+
+    mkdir -p stitch/etc/dylib
+    cd stitch/etc/dylib
+    curl -s "$server_stitch_lib_url" | tar xvfz - --strip-components=1
+    cd -
+
+    echo "building transpiler"
+    cd stitch/etc/transpiler
+    curl -O "https://nodejs.org/dist/v8.11.2/node-v8.11.2-darwin-x64.tar.gz"
```
tgoyne, line 74: Node's already installed on the CI machines (and if 8 specifically is needed then it's available via nvm).

Realm/ObjectServerTests/run_baas.sh
```diff
@@ -0,0 +1,140 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+######################################
+
+mongodb_url=http://fastdl.mongodb.org/osx/mongodb-osx-ssl-x86_64-4.0.2.tgz;
+transpiler_target=node8-macos;
+server_stitch_lib_url="https://s3.amazonaws.com/mciuploads/mongodb-mongo-master/stitch-support/osx-1010/80f9a13324fc36b2deb400e5a185968f6fa8f64a/stitch-support-4.1.7-319-g80f9a13324.tgz";
```
tgoyne, line 23: Pulling all of the various version numbers that appear in the script below up to here would make updating things easier.

Realm/ObjectServerTests/run_baas.sh
```diff
@@ -0,0 +1,140 @@
...
+    ./bin/mongo --nodb --eval 'assert.soon(function(x){try{var d = new Mongo("localhost:26000"); return true}catch(e){return false}}, "timed out connecting")'
+    ./bin/mongo --port 26000 --eval 'rs.initiate()'
+    echo "mongod is up."
+    popd
+}
+
+function setup_stitch() {
+    OG_DIR=`pwd`
+    echo "cloning stitch"
+    git clone git@github.com:10gen/stitch
```
tgoyne, line 51: We'll want to check out a specific version of stitch rather than just using the latest commit on the default branch. We also don't really want to be cloning, building, and then deleting stitch every time we run the tests; instead it should clone it to a semi-persistent location if it's not already there, or make sure that the correct version is checked out if there is already one there. This could also be done as a build step similar to how the download core build step works rather than at runtime.

Realm/ObjectServerTests/RLMSyncTestCase.mm
```diff
@@ -122,56 +118,72 @@ + (instancetype)sharedServer {
...
 
-- (void)launch {
-    if (_task) {
-        return;
+- (void)cleanUp {
+    NSString *pidfile = [[@(__FILE__) stringByDeletingLastPathComponent] stringByAppendingString:@"/build/mongodb-osx-x86_64-4.0.2/pid.txt"];
+    if ([[NSFileManager defaultManager] fileExistsAtPath:pidfile]) {
+        NSString *content = [NSString stringWithContentsOfFile:pidfile encoding:NSUTF8StringEncoding error:nil];
+        // Clean up any old state from the server
+        [[NSTask launchedTaskWithLaunchPath:@"/usr/bin/pkill"
```
tgoyne, line 147: pkill can take a pidfile directly, so this can just be `[[NSTask launchedTaskWithLaunchPath:@"/usr/bin/pkill" arguments:@[@"-F", pidfile]] waitUntilExit]`.

Realm/ObjectServerTests/RLMSyncTestCase.mm
```diff
@@ -233,7 +245,7 @@ - (void)downloadObjectServer {
                        @"--no-save",
                        @"--no-package-lock",
                        @"install",
-                       [@"realm-object-server@" stringByAppendingString:desiredVersion]
+                       @"mongodb-stitch"
```
tgoyne, line 248: The `desiredVersion` part here is still relevant; we don't want to just install whatever version happens to have the `latest` tag at the moment.

Realm/ObjectServerTests/SwiftObjectServerTests.swift
```diff
@@ -31,522 +31,522 @@ class Person: Object {
...
-        _ = try! Realm(configuration: config)
-        self.waitForExpectations(timeout: 4.0)
-    }
-
-    private func realmURLForFile(_ fileName: String) -> URL {
-        let testDir = RLMRealmPathForFile("realm-object-server")
-        let directory = URL(fileURLWithPath: testDir, isDirectory: true)
-        return directory.appendingPathComponent(fileName, isDirectory: false)
-    }
+//
```
tgoyne, line 34: Disabling code by commenting out each line separately makes the diffs a mess. Doing it via `#if false` makes it so that when the code is reenabled later it's actually possible to see what's changing to make it work again.

jsflax:
@tgoyne ptal

Realm/ObjectServerTests/RLMSyncTestCase.mm
```diff
@@ -233,7 +233,7 @@ - (void)downloadObjectServer {
                        @"--no-save",
                        @"--no-package-lock",
                        @"install",
-                       [@"realm-object-server@" stringByAppendingString:desiredVersion]
+                       [@"mongodb-stitch" stringByAppendingString:@"@3.11.0"]
```
tgoyne, line 236: This should be `desiredVersion` (which is read from dependencies.list), not a hardcoded version.

Realm/ObjectServerTests/setup_baas.sh
```diff
@@ -0,0 +1,167 @@
...
+    popd
+}
+
+function setup_stitch() {
+    set -e
+    OG_DIR=`pwd`
+    echo "setting up stitch"
+    if [ ! -d stitch ]
+    then
+    echo "cloning stitch"
```
tgoyne, line 76: The lack of indentation of blocks in this script make the control flow very hard to follow.

PR #6478, opened 2020-04-20
Add better support for using dispatch queues

tgoyne:
This has two main pieces: Realms which are confined to a dispatch queue rather than a thread, and notifications which are delivered to a scheduler other than the current one (currently only implemented for dispatch queues, but could support other things).

Queue-confined threads replace the check for if we're on the correct thread for a check if we're on the correct queue, using dispatch_queue_set_specific()/dispatch_queue_get_specific() to set some queue-local state to perform the checking. Notifications are done via a dispatch_async() to the queue rather than a signal in the target runloop. Because we can't automatically detect which queue we're running on (dispatch_queue_get_current() is inherently broken and doesn't really work), the queue to bind to has to be passed in when the Realm is opened.

A potentially surprising implication of this is that it's possible to have both a thread-confined and a queue-confined Realm instance usable at the same time, and they potentially could be at different versions.  This hopefully will not be too confusing for users? The main queue is special-cased to actually give the main thread's Realm instance, which hopefully will eliminate the most common place where this would arise (and makes it so that registering for notifications delivered to the main thread from a background thread works sensibly).

Notifications delivered to a queue are done by adding an optional queue parameter to observe(). Implementation-wise this is just a dispatch_async() to the queue with the notification registered there, but with some extra logic to be able to return a token synchronously.

I changed asyncOpen() to produce a queue-confined Realm mostly because it seemed weird for it not to be. I'm not sure if it's actually a functionally useful change.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,13 +1,23 @@
 x.y.z Release notes (yyyy-MM-dd)
 =============================================================
 ### Enhancements
-* None.
+* Add support for queue-confined Realms. Rather than being bound to a specific
+  thread, queue-confined Realms are bound to a dispatch queue and can be used
+  within blocks dispatched to that queue regardless of what thread they happen
+  to run on. In addition, change notifications will be delivered to that queue
+  rather than the thread's run loop.
 
 ### Fixed
 * <How to hit and notice issue? what was the impact?> ([#????](https://github.com/realm/realm-js/issues/????), since v?.?.?)
 * None.
 
-<!-- ### Breaking Changes - ONLY INCLUDE FOR NEW MAJOR version -->
+### Breaking Changes
+
+* The Realm instance passed in the callback to asyncOpen() is now confined to
+  the callback queue passed to asyncOpen() rather than the thread which the
+  callback happens to be called on. This means that the Realm instance may be
+  stored and reused in further blocks dispatched to that queue, but the queue
+  must now be a serial queue.
 
 ### Compatibility
 * File format: Generates Realms with format v9 (Reads and upgrades all previous formats)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@ x.y.z Release notes (yyyy-MM-dd)
   within blocks dispatched to that queue regardless of what thread they happen
   to run on. In addition, change notifications will be delivered to that queue
   rather than the thread's run loop.
+* Add an option to deliver object and collection notifications to a specific
+  queue rather than the current thread.
 
 ### Fixed
 * <How to hit and notice issue? what was the impact?> ([#????](https://github.com/realm/realm-js/issues/????), since v?.?.?)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,12 +2,12 @@ x.y.z Release notes (yyyy-MM-dd)
 =============================================================
 ### Enhancements
 * Add support for queue-confined Realms. Rather than being bound to a specific
-  thread, queue-confined Realms are bound to a dispatch queue and can be used
-  within blocks dispatched to that queue regardless of what thread they happen
-  to run on. In addition, change notifications will be delivered to that queue
-  rather than the thread's run loop.
+  thread, queue-confined Realms are bound to a serial dispatch queue and can be
+  used within blocks dispatched to that queue regardless of what thread they
+  happen to run on. In addition, change notifications will be delivered to that
+  queue rather than the thread's run loop. ([PR #6478](https://github.com/realm/realm-cocoa/pull/6478)).
 * Add an option to deliver object and collection notifications to a specific
-  queue rather than the current thread.
+  serial queue rather than the current thread. ([PR #6478](https://github.com/realm/realm-cocoa/pull/6478)).
 
 ### Fixed
 * <How to hit and notice issue? what was the impact?> ([#????](https://github.com/realm/realm-js/issues/????), since v?.?.?)
--- a/Realm/RLMArray.h
+++ b/Realm/RLMArray.h
@@ -370,9 +370,49 @@ NS_ASSUME_NONNULL_BEGIN
  @param block The block to be called each time the array changes.
  @return A token which must be held for as long as you want updates to be delivered.
  */
-- (RLMNotificationToken *)addNotificationBlock:(void (^)(RLMArray<RLMObjectType> *__nullable array,
-                                                         RLMCollectionChange *__nullable changes,
-                                                         NSError *__nullable error))block __attribute__((warn_unused_result));
+- (RLMNotificationToken *)addNotificationBlock:(void (^)(RLMArray<RLMObjectType> *_Nullable array,
+                                                         RLMCollectionChange *_Nullable changes,
+                                                         NSError *_Nullable error))block
+__attribute__((warn_unused_result));
+
+/**
+ Registers a block to be called each time the array changes.
+
+ The block will be asynchronously called with the initial array, and then
+ called again after each write transaction which changes any of the objects in
+ the array, which objects are in the results, or the order of the objects in the
+ array.
+
+ The `changes` parameter will be `nil` the first time the block is called.
+ For each call after that, it will contain information about
+ which rows in the array were added, removed or modified. If a write transaction
+ did not modify any objects in the array, the block is not called at all.
+ See the `RLMCollectionChange` documentation for information on how the changes
+ are reported and an example of updating a `UITableView`.
+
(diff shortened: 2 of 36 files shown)


CHANGELOG.md
```diff
@@ -1,13 +1,25 @@
 x.y.z Release notes (yyyy-MM-dd)
 =============================================================
 ### Enhancements
-* None.
+* Add support for queue-confined Realms. Rather than being bound to a specific
+  thread, queue-confined Realms are bound to a dispatch queue and can be used
```
bmunkholm, line 5:
```suggestion
  thread, queue-confined Realms are bound to a serial dispatch queue and can be used
```

Realm/RLMArray.h
```diff
@@ -370,9 +370,49 @@ NS_ASSUME_NONNULL_BEGIN
...
+
+ Notifications are delivered on the given queue. If the queue is blocked and
+ notifications can't be delivered instantly, multiple notifications may be
+ coalesced into a single notification.
+
+ You must retain the returned token for as long as you want updates to continue
+ to be sent to the block. To stop receiving updates, call `-invalidate` on the token.
+
+ @warning This method cannot be called when the containing Realm is read-only or frozen.
+ @warning The queue must be a serial queue.
```
bmunkholm, line 405: Can we check is the queue is serial?

tgoyne: The queue type is checked in the object store code and an exception is thrown if it's concurrent.

bmunkholm: Same goes for other places where we take the queue parameter. Perhaps naming the queue parameter `serialQueue` would make it more obvious?

Realm/RLMRealm.mm
```diff
@@ -229,29 +241,25 @@ static void waitForPartialSyncSubscriptions(Realm::Config const& config) {
...
-    Results results(realm, table);
-    CFRunLoopRef runLoop = CFRunLoopGetCurrent();
-    CFRunLoopPerformBlock(runLoop, kCFRunLoopDefaultMode, [&]() mutable {
-        token = results.add_notification_callback([&](CollectionChangeSet const&, std::exception_ptr) mutable {
-            if (table->size() > 1) {
-                token = {};
-                CFRunLoopStop(runLoop);
-            }
-        });
+    struct State {
```
jsflax, line 244: [q] Assumedly this struct exists to maintain a reference to the `NotificationToken` during the lifecycle of the results notification callback?

Realm/RLMRealm.mm
```diff
@@ -229,29 +241,25 @@ static void waitForPartialSyncSubscriptions(Realm::Config const& config) {
...
+        if (table->size() > 1) {
+            state->token = {};
+            completion();
+        }
     });
-    CFRunLoopRun();
 #else
+    static_cast<void>(queue);
     static_cast<void>(config);
+    static_cast<void>(completion);
```
jsflax, line 259: [q] What is the point of enabling this method if sync is not enabled?

Realm/RLMRealm.mm
```diff
@@ -415,16 +423,38 @@ REALM_NOINLINE static void translateSharedGroupOpenException(NSError **error) {
...
+                                 queue:(dispatch_queue_t)queue
+                                 error:(NSError **)error {
     bool dynamic = configuration.dynamic;
     bool cache = configuration.cache;
     bool readOnly = configuration.readOnly;
 
+    // The main thread and main queue share a cache key of 1 so that they give
+    // the same instance. Other Realms are keyed on either the thread or the queue.
+    // Note that despite being a void* the cache key is not actually a pointer;
+    // this is just an artifact of NSMapTable's strange API.
```
jsflax, line 439: 👍

Realm/RLMRealm.mm
```diff
@@ -516,8 +557,7 @@ + (instancetype)realmWithConfiguration:(RLMRealmConfiguration *)configuration er
         RLMRealmCreateAccessors(realm.schema);
 
         if (!readOnly) {
-            // initializing the schema started a read transaction, so end it
-            [realm invalidate];
+            REALM_ASSERT(!realm->_realm->is_in_read_transaction());
```
jsflax, line 560: [q] Is this unrelated to the rest of the PR? Are adding an assertion instead of allowing this?

Realm/RLMResults.h
```diff
@@ -256,9 +256,54 @@ NS_ASSUME_NONNULL_BEGIN
...
+ @warning The queue must be a serial queue.
+
+ @param block The block to be called whenever a change occurs.
+ @param queue The serial queue to deliver notifications to.
+ @return A token which must be held for as long as you want updates to be delivered.
+ */
+- (RLMNotificationToken *)addNotificationBlock:(void (^)(RLMResults<RLMObjectType> *_Nullable results,
+                                                         RLMCollectionChange *_Nullable change,
+                                                         NSError *_Nullable error))block
+                                receiveOnQueue:(nullable dispatch_queue_t)queue
```
jsflax, line 305: This naming is inconsistent elsewhere. Seems like `realmWithConfiguration` should have its queue infix be `confineOnQueue`.

Realm/RLMCollection.mm
```diff
@@ -404,20 +376,79 @@ - (NSString *)description {
...
 
+template<typename RLMCollection>
+RLMNotificationToken *RLMAddNotificationBlock(RLMCollection *collection,
+                                              void (^block)(id, RLMCollectionChange *, NSError *),
+                                              dispatch_queue_t queue) {
+    RLMRealm *realm = collection.realm;
+    if (!realm) {
+        @throw RLMException(@"Linking objects notifications are only supported on managed objects.");
+    }
+    auto skip = std::is_same_v<RLMCollection, RLMResults> ? std::make_shared<bool>(true) : nullptr;
```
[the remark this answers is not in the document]
tgoyne: `std::shared_ptr<bool>` captured by a lambda is basically equivalent to a `__block bool` captured by a block.

Of course, you'll note that it isn't actually being captured by a lambda and after the refactoring to make CollectionCallbackWrapper a separate type I think this needs some adjustments to make more sense.

Realm/RLMListBase.h
```diff
@@ -33,11 +33,10 @@ NS_ASSUME_NONNULL_BEGIN
 
 @interface RLMLinkingObjectsHandle : NSObject
 - (instancetype)initWithObject:(RLMObjectBase *)object property:(RLMProperty *)property;
+- (instancetype)initWithLinkingObjects:(RLMResults *)linkingObjects;
```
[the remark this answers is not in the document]
tgoyne: RLMLinkingObjects is just a subclass of RLMResults used for property declarations in obj-c and none of the Swift LinkingObjects stuff uses it. The problem is that RLMLinkingObject's generic type is constrained to RLMObject, while LinkingObjects is constrained to Object (which is a sibling of RLMObject).

Realm/RLMRealm.mm
tgoyne: The function which calls this function exists in non-sync builds, so the callsite would need to be ifdefed out too if no definition was provided.

Realm/RLMRealm.mm
tgoyne: Yes. The previous version of this code blocked the calling thread until it completed and so could stick everything on the stack, but this version doesn't so it needs to store everything needed for the callback on the heap.

Realm/RLMRealm.mm
tgoyne: I tried to make it so that Realms could be opened for dispatch queues other than the current one, and part of that was making it so that update_schema() ends the read transaction it starts rather than this code. Supporting that turned out to be harder than it was worth, but I think it's still worth leaving in the change to update_schema()'s behavior. There's no user-visible difference here.

The reason we don't want newly created Realm instances to have a read transaction is to make code like `while (true) dispatch_async(queue, ^{ [RLMRealm.defaultRealm transactionWithBlock:^{ ... }]);` less bad. Starting a read transaction as part of the RLMRealm initialization means that all the threads waiting for the write lock are pinning a Realm version and the file size will grow rapidly, while with this behavior there's no pinning and it'll stay fairly constant.

Realm/RLMResults.h
tgoyne: It looks like the Foundation types which do similar things just use the label `queue:` so I guess I'll switch this to that. At that point the only outlier will be `observe(on:)` in Swift, and that's the normal label for that in Swift.

koneksa-sunny:
just want to say @tgoyne this is fantastic work! I've been waiting for this improvement for a while

PR #6488, opened 2020-04-27
Add support for Combine/SwiftUI

tgoyne:
This is currently missing API documentation, which I plan to write after finishing up the first draft of the release post.

The public API here is:

function | description
------------ | -------------
`RealmCollection.objectWillChange` | A publisher which publishes Void each time the collection changes.
`RealmCollection.publisher` | A publisher which publishes the collection each time it changes.
`RealmCollection.changesetPublisher` | A publisher which publishes a RealmCollectionChangeset each time the collection changes.
`Object.objectWillChange` | A publisher which publishes Void each time the collection changes.
`publisher(Object)` | A publisher which publishes the object each time it changes.
`changesetPublisher(Object)` | A publisher which publishes an ObjectChange<T> each time the object changes.
`Publisher.freeze()` | A transform operator which is conceptually equivalent to `.map { $0.freeze() }` but also support object and collection changesets.
`Publisher.threadSafeReference()` | A transform operator which allows calling `receive(on:)` on a publishers which outputs thread-confined objects.

`publisher(Object)` and `changesetPublisher(Object)` free functions rather than properties due to a shortcoming of Swift's type system: it's impossible to have a function/property which returns `Foo<Self>` when subclassing is involved. Defining `var publisher: Publishers.Value<Self>` on `Object` results in subclasses actually having `var publisher: Publishers.Value<Object>` and not the subclass type.

Most of the code here ends up being related to making `subscribe(on:)` and `receive(on:)` work properly.

The SwiftUI support mostly comes in the form of a small sample app which shows that things work without anything explicit. The main SwiftUI-specific thing is the `objectWillChange` property (and supporting `WillChange` publisher) to support `@ObservedObject`. There is also a `ObjectKeyIdentifable` protocol which users can declare their object type as conforming to in order to get `Identifiable` conformance. Since this defines a property named `id` I don't think it's something we want to do automatically on `Object`.


tgoyne:
The API docs are now complete and could use a look over them.

jsflax:
I'll take a look today.

RealmSwift/Combine.swift
```diff
@@ -74,17 +160,86 @@ extension Publisher {
...
+    /// objects is unneccesary but is allowed.
+    ///
+    /// - returns: A publisher that supports `receive(on:)` for thread-confined objects.
     public func threadSafeReference() -> Publishers.MakeThreadSafe<Self> {
         Publishers.MakeThreadSafe(self)
     }
 }
 
 @available(OSX 10.15, watchOS 6.0, iOS 13.0, iOSApplicationExtension 13.0, OSXApplicationExtension 10.15, *)
 extension Publisher {
```
jsflax, line 194: It might be nice to add a comment to this extension and the above one, giving some context on what functionality is being added and why.

Cay-Zhang:
Is there a publisher for observing any change made to a `Realm`? Currently, I have to call `realm.observe` and invoke the view update myself in my SwiftUI app.

PR #6543, opened 2020-05-29
Add MongoClient functionality

leemaguire:
This PR adds the RemoteMongoClient functionality that exists in Object Store. The Mongo client can be accessed from RLMApp / RealmApp and the user will be given a RLMMongoClient to then create a RLMMongoDatabase which in turn produces a RLMMongoCollection which can be used to query a remote mongodb instance.


RealmSwift/MongoClient.swift
```diff
@@ -0,0 +1,497 @@
...
+    /// - Parameters:
+    ///   - filter: A `Document` that should match the query.
+    ///   - completion: The result of the attempt to delete a document.
+    public func findOneAndDelete(_ filter: Document,
+                                 _ completion: @escaping MongoDeleteBlock) {
+        self.__findOneAndDelete(toRLMBSON(filter),
+                                completion: completion)
+    }
+
+    private func toRLMBSON(_ document: Document) -> RLMBSON {
```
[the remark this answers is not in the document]
leemaguire: Removed, error passed back to callers completion handler

Realm/RLMUpdateResult.mm
```diff
@@ -0,0 +1,35 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+////////////////////////////////////////////////////////////////////////////
+
+#import <Realm/RLMUpdateResult_Private.hpp>
```
jsflax, line 19: weird import

Realm/RLMMongoDatabase_Private.hpp
```diff
@@ -0,0 +1,37 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+////////////////////////////////////////////////////////////////////////////
+
+#import <Realm/RLMMongoClient.h>
```
jsflax, line 19: weird import

Realm/RLMMongoCollection_Private.hpp
```diff
@@ -0,0 +1,39 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+////////////////////////////////////////////////////////////////////////////
+
+#import <Realm/RLMMongoClient.h>
```
jsflax, line 19: weird import

Realm/RLMUpdateResult_Private.hpp
```diff
@@ -0,0 +1,30 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+////////////////////////////////////////////////////////////////////////////
+
+#import <Realm/RLMUpdateResult.h>
```
jsflax, line 19: weird import

Realm/RLMUsernamePasswordProviderClient.h
```diff
@@ -96,12 +98,12 @@ NS_ASSUME_NONNULL_BEGIN
...
  @param email  The email address of the user.
  @param password The desired new password.
- @param args A pre-serialized list of arguments. Must be a JSON array.
+ @param args A pre-serialized list of arguments passed in as a BSON array.
  @param completionHandler A callback to be invoked once the call is complete.
 */
 - (void)callResetPasswordFunction:(NSString *)email
                          password:(NSString *)password
-                             args:(NSString *)args
+                             args:(NSArray<NSDictionary<NSString *, id<RLMBSON>> *> *)args
```
jsflax, line 106: This should not be an `NSArray` of documents. In this case, it should match callFunction: `NSArray<id<RLMBSON>> *`

RealmSwift/MongoClient.swift
```diff
@@ -0,0 +1,572 @@
...
+            return value.documentValue
+        }
+        set {
+            if let value = newValue {
+                __sort = ObjectiveCSupport.convert(object: AnyBSON(value))
+            }
+        }
+    }
+    
+    public convenience init(_ limit: uint64?, _ projectedBSON: Document?, _ sortBSON: Document?) {
```
jsflax, line 88: Awkward naming, and uint64 should not be used. Should be an Int, `projection`, and `sort`

RealmSwift/MongoClient.swift
```diff
@@ -0,0 +1,572 @@
...
+        set {
+            if let value = newValue {
+                __sort = ObjectiveCSupport.convert(object: AnyBSON(value))
+            }
+        }
+    }
+    
+    public convenience init(_ projection: Document?,
+                            _ sort: Document?,
+                            _ upsert: Bool=false,
```
jsflax, line 135: probably going to cause a SwiftLint error

RealmSwift/MongoClient.swift
jsflax: Also missing doc string

RealmSwift/MongoClient.swift
```diff
@@ -0,0 +1,572 @@
...
+            return value.documentValue
+        }
+        set {
+            if let value = newValue {
+                __sort = ObjectiveCSupport.convert(object: AnyBSON(value))
+            }
+        }
+    }
+    
+    public convenience init(_ projection: Document?,
```
jsflax, line 133: missing doc string

RealmSwift/MongoClient.swift
```diff
@@ -0,0 +1,572 @@
...
+        }
+    }
+    
+    /// Counts the number of documents in this collection matching the provided filter.
+    /// - Parameters:
+    ///   - filter: A `Document` as bson that should match the query.
+    ///   - limit: The max amount of documents to count
+    ///   - completion: Returns the count of the documents that matched the filter.
+    public func count(filter: Document,
+                      limit: uint64,
```
jsflax, line 297: Weird type

RealmSwift/RealmApp.swift
```diff
@@ -119,4 +119,8 @@ public extension RealmApp {
     var functions: Functions {
         return Functions(app: self)
     }
+    
+    func mongoClient(_ serviceName: String) -> MongoClient {
```
jsflax, line 123: missing doc string

RealmSwift/MongoClient.swift
```diff
@@ -84,12 +84,17 @@ extension FindOptions {
...
+    ///   - limit: The maximum number of documents to return. Specifying 0 will return all documents.
+    ///   - projected: Limits the fields to return for all matching documents.
+    ///   - sort: The order in which to return matching documents.
+    public convenience init(_ limit: Int?, _ projected: Document?, _ sort: Document?) {
         self.init()
-        self.limit = UInt(limit ?? 0)
-        self.projected = projectedBSON
-        self.sort = sortBSON
+        self.limit = limit ?? 0
+        self.projected = projected
```
jsflax, line 96: This is still the wrong name, should be projection

RealmSwift/MongoClient.swift
```diff
@@ -129,16 +134,26 @@ extension FindOneAndModifyOptions {
...
     }
-    
+
+    /// Options to use when executing a `findOneAndUpdate`, `findOneAndReplace`,
+    /// or `findOneAndDelete` command on a `RLMMongoCollection`
+    /// - Parameters:
+    ///   - projection: Limits the fields to return for all matching documents.
+    ///   - sort: The order in which to return matching documents.
+    ///   - upsert: Whether or not to perform an upsert, default is false
+    ///   (only available for find_one_and_replace and find_one_and_update)
```
jsflax, line 144: This comment is still wrong

RealmSwift/MongoClient.swift
```diff
@@ -178,28 +193,28 @@ public typealias MongoDeleteBlock = RLMMongoDeleteBlock
...
     /// Encodes the provided values to BSON and inserts them. If any values are missing identifiers,
     /// they will be generated.
     /// - Parameters:
     ///   - documents: The `Document` values in a bson array to insert.
     ///   - completion: The result of the insert, returns an array inserted document ids in order.
     public func insertMany(_ documents: [Document], _ completion: @escaping MongoInsertManyBlock) {
         do {
-            let convertedDocuments = try documents.map { (document: Document) -> [String : RLMBSON] in
-                guard let rlmBSON = ObjectiveCSupport.convert(object: AnyBSON(document)) as? [String : RLMBSON] else {
+            let convertedDocuments = try documents.map { (document: Document) -> [String: RLMBSON] in
```
jsflax, line 216: Is this actually needed? Shouldn't you just be able to convert the whole array?

RealmSwift/MongoClient.swift
```diff
@@ -263,21 +277,21 @@ extension MongoCollection {
     ///   - completion: The resulting bson or error if one occurs
     public func findOneDocument(filter: Document,
                                 _ completion: @escaping MongoFindOneBlock) {
-        guard let rlmBSON = ObjectiveCSupport.convert(object: AnyBSON(filter)) as? [String : RLMBSON] else {
+        guard let rlmBSON = ObjectiveCSupport.convert(object: AnyBSON(filter)) as? [String: RLMBSON] else {
```
jsflax, line 280: These guards should really never fail– any reason we are throwing this error?

RealmSwift/MongoClient.swift
leemaguire: Changed to: `        let bson = ObjectiveCSupport.convert(object: .array(documents.map {.document($0)}))
`

RealmSwift/MongoClient.swift
leemaguire: Error is no longer needed with new bson conversion method

Realm.xcodeproj/project.pbxproj
```diff
@@ -460,6 +460,21 @@
...
+		CF56B0A9248A8ADB0001251D /* RLMMongoClient_Private.hpp in Headers */ = {isa = PBXBuildFile; fileRef = CF76F7D324816AAA00890DD2 /* RLMMongoClient_Private.hpp */; settings = {ATTRIBUTES = (Private, ); }; };
+		CF56B0AA248A8AE00001251D /* RLMFindOptions_Private.hpp in Headers */ = {isa = PBXBuildFile; fileRef = CF76F7D424816AAA00890DD2 /* RLMFindOptions_Private.hpp */; settings = {ATTRIBUTES = (Private, ); }; };
+		CF56B0AB248A8AE00001251D /* RLMFindOptions_Private.hpp in Headers */ = {isa = PBXBuildFile; fileRef = CF76F7D424816AAA00890DD2 /* RLMFindOptions_Private.hpp */; settings = {ATTRIBUTES = (Private, ); }; };
+		CF56B0AC248A8AE40001251D /* RLMFindOneAndModifyOptions_Private.hpp in Headers */ = {isa = PBXBuildFile; fileRef = CF76F7D824816AAB00890DD2 /* RLMFindOneAndModifyOptions_Private.hpp */; settings = {ATTRIBUTES = (Private, ); }; };
+		CF56B0AD248A8AE40001251D /* RLMFindOneAndModifyOptions_Private.hpp in Headers */ = {isa = PBXBuildFile; fileRef = CF76F7D824816AAB00890DD2 /* RLMFindOneAndModifyOptions_Private.hpp */; settings = {ATTRIBUTES = (Private, ); }; };
```
tgoyne, line 478: Private headers should not be part of the target. Membership in the target determines whether the header is included in the built framework and that should only have public headers.

Realm/RLMUpdateResult.mm
tgoyne: This is failing because it needs to be `#import "RLMUpdateResult_Private.hpp"` as it isn't a public header.

Realm/RLMMongoDatabase.mm
```diff
@@ -0,0 +1,50 @@
...
+#import "RLMApp_Private.hpp"
+#import "RLMMongoDatabase_Private.hpp"
+#import "RLMMongoCollection_Private.hpp"
+#import "sync/app.hpp"
+#import "sync/remote_mongo_client.hpp"
+#import "sync/remote_mongo_database.hpp"
+
+#import <realm/util/optional.hpp>
+
+@implementation RLMMongoDatabase
```
tgoyne, line 30: These ten line classes really don't need separate implementation files for each of them, and putting them in separate files is bad for compile and link times.

Realm/ObjectServerTests/RLMObjectServerTests.mm
```diff
@@ -81,9 +81,9 @@ - (void)testAnonymousAuthentication {
         syncUser = user;
         [expectation fulfill];
     }];
-
+
```
tgoyne, line 84: Adding trailing whitespace to a few places.

Realm/RLMFindOptions.mm
```diff
@@ -0,0 +1,100 @@
...
+};
+@end
+
+@implementation RLMFindOptions
+
+- (instancetype)initWithLimit:(NSInteger)limit
+                   projection:(id<RLMBSON> _Nullable)projection
+                         sort:(id<RLMBSON> _Nullable)sort {
+    if (self = [super init]) {
+        [self setProjection:projection];
```
tgoyne, line 34: Using property setter syntax (i.e. `self.projection = projection`) is more idiomatic.

Realm/RLMFindOptions.mm
```diff
@@ -0,0 +1,100 @@
...
+    if (sort) {
+        auto bson = realm::bson::BsonDocument(RLMConvertRLMBSONToBson(sort));
+        _options.sort_bson = realm::util::Optional<realm::bson::BsonDocument>(bson);
+    } else {
+        _options.sort_bson = realm::util::none;
+    }
+}
+
+- (NSInteger)limit {
+    if (_options.limit) {
```
tgoyne, line 89: `return static_cast<NSInteger>(_options.limit.value_or(0));`

Casting to int is incorrect as NSInteger can be 64-bit.

Realm/RLMMongoClient.mm
```diff
@@ -0,0 +1,45 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+////////////////////////////////////////////////////////////////////////////
+
+#import "RLMMongoClient.h"
+#import "RLMMongoClient_Private.hpp"
```
tgoyne, line 20: Generally we only import one of the public or the private header since the private one is guaranteed to import the public one.

Realm/RLMMongoCollection.h
```diff
@@ -0,0 +1,264 @@
...
+/**
+* The `RLMMongoCollection` represents a MongoDB collection.
+*
+* You can get an instance from a `RLMMongoDatabase`.
+*
+* Create, read, update, and delete methods are available.
+*
+* Operations against the Realm Cloud server are performed asynchronously.
+*
+* - Note:
```
tgoyne, line 36: This isn't the correct syntax for this in obj-c  and it'll end up an unreadable mess in both the alt-click window in Xcode and in the API docs.

Realm/RLMMongoCollection.mm
```diff
@@ -0,0 +1,307 @@
...
+
+- (void)insertManyDocuments:(NSArray<NSDictionary<NSString *, id<RLMBSON>> *> *)documents
+                 completion:(RLMMongoInsertManyBlock)completion {
+    [self collection:self.name].insert_many(static_cast<realm::bson::BsonArray>(RLMConvertRLMBSONToBson(documents)),
+                                            [completion](std::vector<realm::ObjectId> insertedIds,
+                                                         realm::util::Optional<realm::app::AppError> error) {
+        if (error) {
+            return completion(nil, RLMAppErrorToNSError(*error));
+        }
+        NSMutableArray *insertedArr = [[NSMutableArray alloc]  init];
```
tgoyne, line 112: `[[NSMutableArray alloc] initWithCapacity:insertedIds.size()]` is an easy minor optimization.

Realm/RLMMongoCollection.mm
```diff
@@ -0,0 +1,307 @@
...
+             limit:(NSInteger)limit
+        completion:(RLMMongoCountBlock)completion {
+    [self collection:self.name].count(static_cast<realm::bson::BsonDocument>(RLMConvertRLMBSONToBson(document)),
+                                      limit,
+                                      [completion](uint64_t count,
+                                                   realm::util::Optional<realm::app::AppError> error) {
+        if (error) {
+            return completion(0, RLMAppErrorToNSError(*error));
+        }
+        completion((int)count, nil);
```
tgoyne, line 142: `int` is incorrect in all of these as there's platforms where it's 32 bits and NSInteger is 64.

Realm/RLMMongoCollection.mm
```diff
@@ -0,0 +1,307 @@
...
+                                                    [completion](realm::util::Optional<realm::bson::BsonDocument> document,
+                                                                 realm::util::Optional<realm::app::AppError> error) {
+        if (error) {
+            return completion(nil, RLMAppErrorToNSError(*error));
+        }
+
+        if (document) {
+            return completion((NSDictionary *)RLMConvertBsonToRLMBSON(*document), nil);
+        }
+        // no docs where found
```
tgoyne, line 238: were

Realm/RLMMongoCollection.mm
```diff
@@ -0,0 +1,307 @@
...
+    [self collection:self.name].find_one_and_update(static_cast<realm::bson::BsonDocument>(RLMConvertRLMBSONToBson(filterDocument)), static_cast<realm::bson::BsonDocument>(RLMConvertRLMBSONToBson(updateDocument)),
+                                                    [options _findOneAndModifyOptions],
+                                                    [completion](realm::util::Optional<realm::bson::BsonDocument> document,
+                                                                 realm::util::Optional<realm::app::AppError> error) {
+        if (error) {
+            return completion(nil, RLMAppErrorToNSError(*error));
+        }
+
+        if (document) {
+            return completion((NSDictionary *)RLMConvertBsonToRLMBSON(*document), nil);
```
tgoyne, line 236: I notice that we do the equivalent of `foo ? RLMConvertBsonToRLMBSON(*foo) : nil` in an awful lot of places, which suggests that a RLMConvertBsonToRLMBSON overload which takes an optional directly would eliminate a lot of repetition.

RealmSwift/MongoClient.swift
```diff
@@ -0,0 +1,506 @@
...
+    /// Limits the fields to return for all matching documents.
+    public var projection: Document? {
+        get {
+            guard let value = ObjectiveCSupport.convert(object: __projection) else {
+                return nil
+            }
+            return value.documentValue
+        }
+        set {
+            if let value = newValue {
```
tgoyne, line 67: This doesn't let you change the value from non-nil to nil. Something like `__projection = newValue.map(AnyBSON.init).map(ObjectiveCSupport.convert)` should work to handle that correctly as a one-liner.

RealmSwift/MongoClient.swift
```diff
@@ -0,0 +1,506 @@
...
+
+/// Options to use when executing a `find` command on a `MongoCollection`.
+public typealias FindOptions = RLMFindOptions
+
+extension FindOptions {
+
+    /// Limits the fields to return for all matching documents.
+    public var projection: Document? {
+        get {
+            guard let value = ObjectiveCSupport.convert(object: __projection) else {
```
tgoyne, line 61: `return ObjectiveCSupport.convert(object: __projection)?.documentValue`

Realm/RLMMongoCollection.mm
leemaguire: This suggestion would be good for a future improvements PR, @jsflax

jsflax: It's a fairly simple addition if you want to add it:

```objc
id<RLMBSON> RLMConvertBsonToRLMBSON(realm::util::Optional<Bson> b) {
    return b ? RLMConvertBsonToRLMBSON(*b) : nil;
}
```

or

```objc
id<RLMBSON> RLMConvertBsonToRLMBSON(realm::util::Optional<Bson> b) {
    return RLMConvertBsonToRLMBSON(b.value_or(Bson()));
}
```

tgoyne:
The Cocoapods build is failing because any new public headers need to be added to the list in the podspec.

PR #6558, opened 2020-06-08
10.0.0-beta.1

jsflax:


Realm.xcodeproj/project.pbxproj
```diff
@@ -3207,7 +3207,9 @@
 			buildSettings = {
 				CLANG_ENABLE_MODULES = YES;
 				DEFINES_MODULE = YES;
+				IPHONEOS_DEPLOYMENT_TARGET = 8.0;
```
tgoyne, line 3210: These are set in Configuration/Base.xcconfig, not the project file.

tgoyne: They also need to be changed back in the podspec.

PR #6570, opened 2020-06-10
Fix a few bugs with frozen objects in obj-c

### LINKED ISSUE #6568: Freeze objects: isFrozen and linkingObjects don't work
## Goals
Generate a frozen copy of an RLMObject subclass

## Expected Results
The generated frozen object should return true. The frozen object should be able to access its linking objects.

## Actual Results
The generated frozen object returns false to `isFrozen`. Accessing its linkingObjects generates an EXC_BAD_ACCESS crash

## Code Sample
```
func testImmutableCopies() throws {
        let realm = RLMRealm.default()
        let person = Person()
        person.name = "Jennifer"
        person.age = "77"
        try realm.transaction {
            realm.add(person)
        }
        let newPerson  = person.freeze()
        XCTAssertTrue(newPerson.isFrozen) //This line fails
}
```

```
    func testImmutableLinkingObjecta() throws {
        let realm = RLMRealm.default()
        
        let person = Person()
        person.name = "Jennifer"
        person.age = "77"
        
        let dog = Dog()
        dog.name = "Max"
        dog.owner = person
        try realm.transaction {
            realm.add(person)
            realm.add(dog)
        }
        let newPerson  = person.freeze()
        XCTAssertNotNil(newPerson.dog) // This line crashes
}
```

```
@objcMembers public class Person: RLMObject {
    override init() {
        super.init()
    }
    public override class func primaryKey() -> String? {
        return "name"
    }
    
    @objc dynamic var dog: RLMLinkingObjects<Dog>?
    @objc dynamic var name: String?
    @objc dynamic var date: Date?
    @objc dynamic var age: String?

    @objc public override class func linkingObjectsProperties() -> [String : RLMPropertyDescriptor] {
        return ["dog": RLMPropertyDescriptor.init(with: Dog.self, propertyName: "owner")]
    }
    @objc dynamic var _dog: Dog? {
        return self.dog?.firstObject() as? Dog
    }
}
```

```
@objcMembers open class Dog: RLMObject {
    @objc dynamic var name: NSString?
    @objc dynamic var owner: Person?

    override init() {
        super.init()
    }
}

```
Realm framework version: ? 5.0.2

Realm Object Server version: ? N/A

Xcode version: 11.3.1

iOS/OSX version: Simulator iOS 13

Dependency manager + version: N/A

tgoyne:
Linking objects on frozen objects were only tested in Swift and not obj-c, and so unsurprisingly the obj-c version had a dumb bug. -[RLMObject isFrozen] was tested, but the test was wrong. The other change here is just a bit of error checking to give a better error when the user does something that can't work.

Fixes #6568.

### DIFF
--- a/Realm/RLMRealm.mm
+++ b/Realm/RLMRealm.mm
@@ -1017,6 +1017,7 @@ REALM_NOINLINE static void translateSharedGroupOpenException(NSError **error) {
     try {
         RLMRealm *realm = [[RLMRealm alloc] initPrivate];
         realm->_realm = _realm->freeze();
+        realm->_realm->set_schema_subset(_realm->schema());
         realm->_realm->read_group();
         realm->_dynamic = _dynamic;
         realm->_schema = _schema;
--- a/Realm/Tests/LinkingObjectsTests.mm
+++ b/Realm/Tests/LinkingObjectsTests.mm
@@ -87,6 +87,39 @@
                               @"Linking objects notifications are only supported on managed objects.");
 }
 
+- (void)testLinkingObjectsOnFrozenObject {
+    NSArray *(^asArray)(id) = ^(id arrayLike) {
+        return [arrayLike valueForKeyPath:@"self"];
+    };
+
+    RLMRealm *realm = [self realmWithTestPath];
+    [realm beginWriteTransaction];
+    PersonObject *hannah = [PersonObject createInRealm:realm withValue:@[@"Hannah", @0]];
+    PersonObject *mark   = [PersonObject createInRealm:realm withValue:@[@"Mark",  @30, @[hannah]]];
+    [realm commitWriteTransaction];
+
+    PersonObject *frozenHannah = hannah.freeze;
+    PersonObject *frozenMark = mark.freeze;
+    XCTAssertEqualObjects(asArray(frozenHannah.parents), (@[frozenMark]));
+
+    [realm beginWriteTransaction];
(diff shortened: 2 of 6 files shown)


Realm/Tests/LinkingObjectsTests.mm
```diff
@@ -87,6 +87,39 @@ - (void)testLinkingObjectsOnUnmanagedObject {
                               @"Linking objects notifications are only supported on managed objects.");
 }
 
+- (void)testLinkingObjectsOnFrozenObject {
+    NSArray *(^asArray)(id) = ^(id arrayLike) {
```
jsflax, line 91: Nice.

Realm/Tests/ObjectTests.m
```diff
@@ -1354,7 +1354,7 @@ - (void)testIsFrozen {
     IntObject *frozen = [managed freeze];
     XCTAssertFalse(standalone.isFrozen);
     XCTAssertFalse(managed.isFrozen);
-    XCTAssertFalse(frozen.isFrozen);
```
jsflax, line 1357: Was this a holdover to get tests passing?

Realm/RLMRealm.mm
```diff
@@ -1017,6 +1017,7 @@ - (RLMRealm *)frozenCopy {
     try {
         RLMRealm *realm = [[RLMRealm alloc] initPrivate];
         realm->_realm = _realm->freeze();
+        realm->_realm->set_schema_subset(_realm->schema());
```
jsflax, line 1020: Why is this needed?

PR #6579, opened 2020-06-16
Allow opening full-sync Realms in read-only mode

tgoyne:
This is something that object store has supported for a while but we never exposed.

This has a few effects:

1. The schema won't be initialized locally, which means that read-only Realms no longer have to use asyncOpen the first time they're opened to avoid making a write that'll be rejected by the server.
2. Having extra types in the local schema that aren't present on the server will work fine. This works around the problem with partial sync permissions types ending up in the schema for full-sync Realms, and just generally removes the need to always explicitly list object types for a read-only Realm.
3. Incompatible schema changes such as adding a property to a type that isn't present on the server is more likely to produce a useful error immediately rather than a confusing one via the sync session error handler.
4. Better error reporting in general for accidental misuse.

### DIFF
--- a/Realm/ObjectServerTests/RLMObjectServerTests.mm
+++ b/Realm/ObjectServerTests/RLMObjectServerTests.mm
@@ -21,12 +21,15 @@
 #import "RLMSyncSessionRefreshHandle+ObjectServerTests.h"
 #import "RLMSyncUser+ObjectServerTests.h"
 
+#import "RLMObjectSchema_Private.hpp"
 #import "RLMRealm+Sync.h"
 #import "RLMRealmConfiguration_Private.h"
 #import "RLMRealmUtil.hpp"
 #import "RLMRealm_Dynamic.h"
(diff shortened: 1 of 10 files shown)


Realm/ObjectServerTests/RLMObjectServerTests.mm
```diff
@@ -2351,4 +2354,98 @@ - (void)testModifyCustomHeadersAfterOpeningRealm {
     [self waitForExpectations:@[ex3, ex4] timeout:4.0];
 }
 
+#pragma mark - Read Only
+
+- (RLMSyncUser *)userForTest:(SEL)sel {
```
jsflax, line 2359: Doesn't this method already exist elsewhere?

tgoyne: You'd think so, right? I couldn't find it anywhere though and the existing tests just all duplicate this logic. I skipped updating all the existing tests to use this to simplify merging it to v10.

PR #6589, opened 2020-06-22
Add Apple authentication helper

leemaguire:
This PR focuses on creating an extension for 'Sign in with Apple' functionality. The user will use `RLMApp` or `App` to set the `ASAuthorizationControllerDelegate` by passing in the `ASAuthorizationController` via `setASAuthorizationControllerDelegateWithController:` in `RLMApp`. The user can then subscribe to the result of the authentication using `RLMASLoginDelegate` or `ASLoginDelegate`


RealmSwift/App.swift
```diff
@@ -81,3 +81,29 @@ public typealias Credentials = RLMCredentials
...
+public typealias ASLoginDelegate = RLMASLoginDelegate
+
+@available(OSX 10.15, watchOS 6.0, iOS 13.0, iOSApplicationExtension 13.0, OSXApplicationExtension 10.15, tvOS 13.0, *)
+extension App {
+    /**
+     Sets the ASAuthorizationControllerDelegate to be handled by `App`
+     - Parameter controller: The ASAuthorizationController in which you want `App` to consume its delegate.
+
+     Usage:
+     ```let app = App(id: "my-app-id")
```
tgoyne, line 96: \``` needs to be on its own line for jazzy to render this correctly.

Realm/RLMApp.h
```diff
@@ -175,3 +175,37 @@ to obtain a reference to an RLMApp.
 @end
 
 NS_ASSUME_NONNULL_END
+
+#pragma mark - Sign In With Apple Extension
```
jsflax, line 179: All of this could be in a separate file. See conventions with regards to categories in objc. Something like `RLMApp+ ASAuthorizationControllerDelegate.h`.

RealmSwift/App.swift
```diff
@@ -81,3 +81,29 @@ public typealias Credentials = RLMCredentials
...
+     let appleIDProvider = ASAuthorizationAppleIDProvider()
+     let request = appleIDProvider.createRequest()
+     request.requestedScopes = [.fullName, .email]
+
+     let authorizationController = ASAuthorizationController(authorizationRequests: [request])
+     app.setASAuthorizationControllerDelegate(controller: authorizationController)
+     authorizationController.presentationContextProvider = self
+     authorizationController.performRequests()```
+    */
+    public func setASAuthorizationControllerDelegate(controller: ASAuthorizationController) {
```
jsflax, line 106: Seems this would be more idiomatic as:
`public func setASAuthorizationController(_ controller: ASAuthorizationController)`

examples/ios/swift/AppleAuthentication/ContentView.swift
```diff
@@ -0,0 +1,96 @@
...
+
+import SwiftUI
+import AuthenticationServices
+import RealmSwift
+
+/// Your MongoDB Realm App ID
+let appId = "your-app-id"
+
+struct ContentView: View {
+
```
jsflax, line 28: extra newline

examples/ios/swift/AppleAuthentication/ContentView.swift
```diff
@@ -0,0 +1,96 @@
...
+            SignInWithAppleView(accessToken: $accessToken, error: $error)
+                .frame(width: 200, height: 50, alignment: .center)
+            Text(self.accessToken)
+            Text(self.error)
+        }
+    }
+}
+
+class SignInCoordinator: ASLoginDelegate {
+
```
jsflax, line 43: extra newline

examples/ios/swift/AppleAuthentication/ContentView.swift
```diff
@@ -0,0 +1,96 @@
...
+        parent.error = error.localizedDescription
+    }
+
+    func authenticationDidComplete(with user: User) {
+        parent.accessToken = user.accessToken ?? "Could not get access token"
+    }
+}
+
+struct SignInWithAppleView: UIViewRepresentable {
+
```
jsflax, line 73: extra newline

examples/ios/swift/AppleAuthentication/ContentView.swift
```diff
@@ -0,0 +1,96 @@
...
+    var parent: SignInWithAppleView
+    var app: App
+
+    init(parent: SignInWithAppleView) {
+        self.parent = parent
+        app = App(id: appId)
+        app.authorisationDelegate = self
+    }
+
+    @objc func didTapLogin() {
```
jsflax, line 53: Is `@objc` needed here?

leemaguire: Yes because it uses `#selector`

Realm/RLMApp.h
leemaguire: Resolved offline

RealmSwift/App.swift
leemaguire: I think it's important to inform the user that the purpose of this call is to set the delegate, and that no other manipulation is performed on the controller.

jsflax: I see. Then the preposition should probably be `For` instead of `With`.

PR #6593, opened 2020-06-22
Drop support for Xcode 10

tgoyne:
Xcode 12 beta starting means it's time to prune the set of supported Xcode versions. Xcode 11.3 is the final version to support macOS 10.14, so it's now the new minimum version supported.

Also switch to building as Swift 5 by default. This doesn't really change much (the release packages are built with specific versions of Swift) but it's probably better for local development.


tgoyne:
Release pipeline test run with this: https://ci.realm.io/blue/organizations/jenkins/cocoa-pipeline/detail/cocoa-pipeline/1102/pipeline

PR #6596, opened 2020-06-22
Fix building with Xcode 12b1

tgoyne:
Currently only the SPM build actually *works*; anything using a prebuilt core binary crashes whenever we call sort() and we don't build for watchOS at all.


cnstoll:
I'm also seeing a few runtime issues in Realm running in the iOS 14 simulator. 

Screen Shot 2020-06-22 at 8 10 20 PM
Screen Shot 2020-06-22 at 8 10 29 PM

tgoyne:
@cnstoll [I made a separate PR to fix that](https://github.com/realm/realm-object-store/pull/1048). It was probably harmless other than being UB.

tgoyne:
Everything but watchOS is now working. watchOS is just waiting on a new sync release.

Realm/RLMRealm+Sync.h
```diff
@@ -18,7 +18,7 @@
 
 #import <Foundation/Foundation.h>
 
-#import "RLMRealm.h"
+#import <Realm/RLMRealm.h>
```
jsflax, line 21: Is this angled brace notation required now?

tgoyne: Headers which are part of the framework umbrella header have to be imported via the framework path (and everything else needs to not be). This is a hard requirement for spm, and for non-spm cases not doing it was broken in ways that probably didn't matter (if someone had a file named RLMRealm.h in their project we'd import that rather than the framework's RLMRealm.h). Xcode now warns about bad imports in framework headers by default.

PR #6612, opened 2020-06-29
Add admin sdk, clean up test process

jsflax:
This PR effectively sandboxes Stitch (the backend server), and drops some of the external dependencies. It notably replaces the JS admin SDK with a Swift version, and replaces the ruby based proxy server with a Swift version. The Mongo and Stitch processes will also be managed directly from the test suite as opposed to being spawned externally. This will enable quicker set up and tear down, and a simpler cleaning path.

Stitch will also be precompiled and ran as a binary.

This also gives us Big Sur compatibility (as far as I can tell, they've significantly beefed up the permissions model).


Realm/ObjectServerTests/RLMSyncTestCase.mm
```diff
@@ -675,21 +469,21 @@ - (void)resetSyncManager {
         XCTestExpectation *ex = [self expectationWithDescription:@"Wait for logout"];
         [exs addObject:ex];
         [user logOutWithCompletion:^(NSError *) {
+            // Sessions are removed from the user asynchronously after a logout.
```
tgoyne, line 472: I don't think moving this code here works. It's adding expectations to the array after the call for waitForExpectations and then depending on that call waiting on the new ones, and if that does actually work it's only by coincidence.

Realm/ObjectServerTests/RealmServer.swift
```diff
@@ -0,0 +1,780 @@
...
+            guard let session = adminSession else {
+                XCTFail("Could not initiate admin session: \(loginError?.localizedDescription ?? "unknown cause")")
+                return
+            }
+            self.session = session
+        }
+    }
+
+    /// Lazy teardown for exit only.
+    private lazy var tearDown: () = {
```
tgoyne, line 366: Why is this a lazy var and not just a function?

Realm/ObjectServerTests/RealmServer.swift
```diff
@@ -0,0 +1,780 @@
...
+        serverProcess.terminate()
+        
+        // step down the replica set
+        let rsStepDownProcess = Process()
+        rsStepDownProcess.launchPath = mongoUrl.appendingPathComponent("bin").appendingPathComponent("mongo").absoluteString
+        rsStepDownProcess.arguments = [
+            "admin",
+            "--port", "26000",
+            "--eval", "'db.adminCommand({replSetStepDown: 0, secondaryCatchUpPeriodSecs: 0, force: true})'"]
+        rsStepDownProcess.launch()
```
tgoyne, line 376: launch() is deprecated and you're supposed to use run() now. These also could be `Process.run(launchPath: ..., arguments: ...).waitUntilExit()` for the ones where we're not capturing stdout.

Realm/ObjectServerTests/RealmServer.swift
```diff
@@ -0,0 +1,780 @@
...
+            "--configFile",
+            bundle.path(forResource: "test_config", ofType: "json")!
+        ]
+
+        let pipe = Pipe()
+        pipe.fileHandleForReading.readabilityHandler = { file in
+            // prettify server output
+            let available = String(data: file.availableData, encoding: .utf8)?.split(separator: "\t")
+            print(available!.map { part -> String in
+                if part.contains("INFO") {
```
tgoyne, line 458: Isn't this replacing the entire log line with just the emoji?

Realm/ObjectServerTests/RealmServer.swift
```diff
@@ -0,0 +1,780 @@
...
+                ]
+            ]
+        ])
+
+        guard serviceCreationError == nil,
+              let serviceId = (serviceResponse as? [String: Any])?["_id"] as? String else {
+            throw serviceCreationError ?? URLError(.badServerResponse)
+        }
+        
+        let dogRule: [String : Any] = [
```
tgoyne, line 580: Obviously you don't need to do this now, but I think it'd be pretty easy to create these programmatically from an ObjectSchema and that'd simplify adding new types in the future.

Realm/ObjectServerTests/SwiftSyncTestCase.swift
```diff
@@ -74,7 +63,7 @@ class SwiftSyncTestCase: RLMSyncTestCase {
     }
 
     func synchronouslyLogInUser(for credentials: Credentials,
-                                file: StaticString = #file,
+                                file: StaticString = #filePath,
```
tgoyne, line 66: This breaks Xcode 11. The Swift change was poorly thought out and there's no reasonable way to support both 11 and 12 without warnings, so they're thinking of changing things in a future Xcode beta.

Realm/ObjectServerTests/RLMObjectServerTests.mm
```diff
@@ -1508,19 +1515,17 @@ - (void)testAsyncOpenProgressNotifications {
 - (void)testAsyncOpenConnectionTimeout {
     [self resetSyncManager];
 
-    NSTask *task = [[NSTask alloc] init];
-    NSString *directory = [@(__FILE__) stringByDeletingLastPathComponent];
-    task.currentDirectoryPath = directory;
-    task.launchPath = @"/usr/bin/ruby";
-    task.arguments = @[[directory stringByAppendingPathComponent:@"run_baas.rb"], @"start_proxy", @"5678", @"2"];
-    [task launch];
+    TimeoutProxyServer *proxy = [[TimeoutProxyServer alloc] initWithPort:5678];
```
tgoyne, line 1518: I think this type needs a `stop` method. Currently this is relying on the whims of ARC to ensure that it's deallocated at the right time, but it could legally be deallocated on the line after the call to startAndReturnError.

Realm/ObjectServerTests/RealmServer.swift
jsflax: To act as a `dispatch_once` replacement.

Realm/ObjectServerTests/RealmServer.swift
```diff
@@ -0,0 +1,786 @@
...
+                                         groupId: groupId,
+                                         url: url.appendingPathComponent(id))
+                }
+            }
+
+            private func request(httpMethod: String, data: Any? = nil, completionHandler: @escaping (Any?, Error?) -> ()) {
+                var request = URLRequest(url: url)
+                request.httpMethod = httpMethod
+                request.allHTTPHeaderFields = [
+                    "Authorization": "Bearer \(String(describing: accessToken))",
```
leemaguire, line 115: Why not `"Bearer \(accessToken)"`?

jsflax: This was likely nullable before and is an artefact of that. Will change.

Realm/ObjectServerTests/RealmServer.swift
jsflax: No, stitch logs are separated by tabs, and the INFO/ERROR/DEBUG/etc. tags are standalone parts. So this will only replace the tag.

PR #6651, opened 2020-07-16
Upgrade core and object store

tgoyne:


tgoyne:
Failing due to https://github.com/realm/realm-core/pull/3824

PR #6663, opened 2020-07-21
Add getter/setter for associated object in category

leemaguire:
This PR addresses a crash that will happen when trying to access `authorizationDelegate`. As we are storing a property within a category we need the getter/setter for associated objects


tgoyne:
Since we control the base class we can just explicitly declare an ivar for the property in the class rather than using associated objects.

leemaguire:
@tgoyne I've removed the associated object and replaced with an ivar. PTAL

PR #6667, opened 2020-07-21
Change Cocoapods to build with xcframework

leemaguire:
This PR allows Cocoapods to build Realm with an XCFramework rather than the fat libs for each individual platform. This adds support for Catalyst and Watch arm64. This implementation currently needs a fix-me that should be addressed in the the next release of pods (1.10) as using an XCFramework containing static libs inside a Cocoa framework currently does not setup and link correctly. To get around this the linker flags & library search paths are set manually. The end user does not need to do anything new to use this new implementation.


tgoyne:
I don't think it's worth trying to hack around CocoaPods 1.9 rather than just waiting for 1.10 to be released. The goal of using a xcframework is to simplify the build setup and support catalyst, and this just makes it more complicated and doesn't support both catalyst and normal macos.

leemaguire:
It's good to point that this PR will require changes in the way xcframeworks are built, here is an error from Cocoapods 1.10 beta 

```
 Unable to install vendored xcframework `realm-sync` for Pod `Realm-iOS` because it contains static libraries
with differing binary names: realm-maccatalyst, realm-iphone-device, realm-macosx, realm-watch-device, realm-watch-simulator, realm-appletv-simulator, realm-appletv-device, and realm-iphone-simulator.
```

AndrewKochulab:
Why you closed this PR? Something not working as expected?
Because only one library in my project can't be ported to the Mac OS using Mac Catalyst and this is Realm... I need this.

leemaguire:
@AndrewKochulab https://github.com/realm/realm-cocoa/pull/6710 we are still on it :)

AndrewKochulab:
> @AndrewKochulab #6710 we are still on it :)

Because I was already afraid that you would refuse support of Catalyst. Thank you a lot!

PR #6673, opened 2020-07-27
Switch to using the core/sync xcframework

tgoyne:
This is mostly a trivial change that simplifies the build system a bit. There's only two complicated bits:

Xcode doesn't seem to support configuration-specific target membership, so we can't just have realm-sync.xcframework be part of the Release configuration and realm-sync-dbg.xcframework be part of the Debug configuration. Instead, realm-sync.xcframework is always used, but the path to it is overridden to actually be pointing at the debug framework for Debug builds. This seems to work fine and just looks a bit odd in the UI.

Xcode doesn't want to link xcframeworks into static frameworks, but we can just tell Xcode that the static framework is a regular framework and set the MACH_O_TYPE to staticlib and then it seems to work fine and do the right thing.

Xcode requires that any xcframework dependencies exist when the build is started, which means that downloading it as a build step happens too late. To work around this, there's now a skeleton xcframework for core stored in the repo which contains the Info.plist and directory structure. The download core step now just copies the files into that skeleton. I also fiddled around with the logic to handle swapping between core versions a bit better; it should no longer redownload if the new version is already present locally.

The hacks used to build for catalyst on 10.14 didn't work when using xcframeworks so I just got rid of them.


tgoyne:
And apparently downloading a xcframework as a build step just doesn't work at all because it has to be present when the build system is built.

PR #6702, opened 2020-08-14
reference implementation

pavel-ship-it:
This is sample implementation oа Combine promise.

I have some questions.
A Promise needs an unwrapped value for a Result. Неre you can see the unwrapping.
The callback, potentially, may have a (nil, nil) return. For that matter I added the last 'else' block.
Is there any "Generic"/Unknown error for that matter? Should I use that approach to complete the task or we are more into exceptions? 
Is it what's expected from this task?


jsflax:
This should not be on master. Please point it to v10.

RealmSwift/Sync.swift
```diff
@@ -1616,3 +1616,32 @@ extension List where Element == Permission {
         return RLMPermissionForRole(_rlmArray, role) as! Permission
     }
 }
+
+#if canImport(Combine)
+import Combine
+
+@available(OSX 10.15, watchOS 6.0, iOS 13.0, iOSApplicationExtension 13.0, OSXApplicationExtension 10.15, tvOS 13.0, macCatalyst 13.0, macCatalystApplicationExtension 13.0, *)
```
leemaguire, line 1623: is the `@available` statement required if the whole thing is wrapped in `#if canImport(Combine)`?

jsflax: Yes, it's all required– the macro is "compiled" differently than the extension is.

jsflax (commented):
Generally speaking, it's fine to only assign one or two of us for a PR (especially a draft).

I would also recommend having a bit more substance to the draft before requesting a review. This one is specifically also tough to judge until it is seen branching off v10.

pavel-ship-it:
Thanks @jsflax. I need this only to ask some questions. I'll remove it.

PR #6703, opened 2020-08-14
Change Configuration.objectTypes property type

ericjordanmossman:
This changes the property type of Realm.Configuration.objectTypes from
```
public var objectTypes: [Object.Type]?
```
to 
```
public var objectTypes: [ObjectBase.Type]?
```
where `public typealias ObjectBase = RLMObjectBase`.

This would allow developers to use embedded objects with a custom defined schema in a Realm.Configuration. Since `EmbeddedObject` inherits from `RLMObjectBase` and not `RealmSwift.Object` passing an embedded object into a `objectTypes` is invalid:
```
Cannot assign value of type '[RLMObjectBase.Type]' to type '[Object.Type]?'
```
Developer defined classes that inherit from `Embedded Object` instead of `Object` will be never registered directly to the Realm, ie. the class won't be in the realm schema.
```
Unlike normal top-level objects,
embedded objects cannot be directly created in or added to a Realm. Instead,
they can only be created as part of a parent object, or by assigning an
unmanaged object to a parent object's property.
```
But leaving out a custom class in the `config.objTypes` will throw this error if the embedded parent class is included:
```
'child.memberOf' links to class 'Parent', which is missing from the list of classes managed by the Realm'
```

[036e857](https://github.com/realm/realm-cocoa/pull/6703/commits/036e8571ce71f66d5d31c1d28808122507e8b120) addresses a change to SwiftLint rules.


ericjordanmossman:
These are the remaining failures, with a description of why I'm moving this to review.
**Jenkins: watchos Release 12.0:**
```
Aug 20 10:38:18 ** BUILD FAILED **
Aug 20 10:38:18 
Aug 20 10:38:18 
Aug 20 10:38:18 The following build commands failed:
Aug 20 10:38:18 	CompileC /Users/realm/workspace/objc_pr/configuration/Release/target/watchos/xcode_version/12.0/build/DerivedData/Realm/Build/Intermediates.noindex/Realm.build/Release-watchsimulator/Realm.build/Objects-normal/x86_64/weak_realm_notifier.o /Users/realm/workspace/objc_pr/configuration/Release/target/watchos/xcode_version/12.0/Realm/ObjectStore/src/impl/weak_realm_notifier.cpp normal x86_64 c++ com.apple.compilers.llvm.clang.1_0.compiler
Aug 20 10:38:18 (1 failure)
```
This is an issue we had with Xcode 12-beta3. The issue is solved in master but not v10, which this PR branches from.
**Jenkins: xcframework Release 12.0:** Similar to above.
**Jenkins: osx-object-server Release 11.3 
and Jenkins: osx-object-server Release 12.0**
`testDownloadToCustomPath()` does in fact fail consistently, in the CI and locally. However this [happens in the base v10](https://github.com/realm/realm-cocoa/issues/6719).
`testClientReset()` also fails consistently locally and in the CI builds. This tests also fails from an update to date v10.

Given that the object server tests fail even without my changes and my changes theoretically wouldn't affect those paths, I'm going to move this PR into review.

bmunkholm:
It also needs an update in CHANGELOG.md

pavel-ship-it (approved):

jsflax (approved):

drinkius:
This broke the standard database cleanup we used to have: 

```
try realm.configuration.objectTypes?
.filter { type in AppDelegate.keepInCache.contains { $0 != type } == false }
.forEach { type in try realm.write { realm.delete(realm.objects(type.self)) } }
```

tgoyne:
You just need a `.compactMap { $0 as Object.self }` in there.

drinkius:
Tried this previously, unfortunately get an error:

`Instance method 'self()' is not a member type of 'Object' (aka 'RealmSwiftObject')`

wilmarvh:
@drinkius Did you ever figure out how to work around the error?

drinkius:
@wilmarvh just sitting on Realm v5.5.2

PR #6704, opened 2020-08-15
Add generic key path support to aggregate operations

sereivoanyong:


tgoyne:
Huh, I'd somehow missed that `NSExpression(forKeyPath:)` exists. We didn't want to rely on `_kvcKeyPathString` continuing to exist but this should be fine.

Do you mind adding some basic tests for this? Since it's just a simple wrapper there's not much to test, but we do like to at least have something verifying that code calling it compiles and superficially works. Adding a check for a single property type to each of testMin()/testMax()/testSum()/testAverage() in RealmCollectionTypeTests would be enough.

PR #6708, opened 2020-08-17
Upgrade to core 6.0.18

tgoyne:


tgoyne:
The SwiftLint failure is unrelated to this and is due to a new check added.

Issue #6671, opened 2020-07-25
iOS 14 + Xcode 12 (Beta 3, Beta 4, Beta 5 & Beta 6): When Realm is stored in a shared app group container, backgrounding the app triggers: Message from debugger: Terminated due to signal 9


kunalsood:
## Goals
Have my app continue to run normally, when backgrounded, with Realm stored in a shared app group container. 

## Expected Results
Same as "Goals" [above].

## Actual Results
When, I run the app on an actual device, and swipe up while my app is in the foreground to send it to the background, the app quits with the following message in Xcode:-

`Message from debugger: Terminated due to signal 9`

Additionally: This only happens if I'm holding a reference to a live RLMResults or a notification token (those are the two things I have tested this with so far).

Interestingly, when I run the project with the default realm configurations (i.e. Realm is stored in the documents directory), this issue disappears.

It might be worth mentioning: It appears that the crash report produced on device for this issue when debugger isn't attached shows: `Termination Reason: Namespace RUNNINGBOARD, Code 0xdead10cc`, which would indicate that the application is being terminated by the OS because it held on to a file-lock/database-lock during suspension. (Could be un-related, though there is nothing other than Realm in my project that would hold on to a file-lock/database-lock ever).

## Steps for others to Reproduce
Build and run the [sample project](https://github.com/kunalsood/KSRealmTerminationSample) using Xcode 12 (Beta 3, Beta 4,  Beta 5, or Beta 6) on an actual device running iOS 14 (Beta 3, Beta 4,  Beta 5, or Beta 6).
(I have tested this on an iPhone & an iPad)

## Code Sample
Here is a (quick & dirty) sample project: https://github.com/kunalsood/KSRealmTerminationSample

Most of the relevant code in the project is in: [ViewController.m](https://github.com/kunalsood/KSRealmTerminationSample/blob/main/KSRealmTerminationSample/ViewController.m) file & one realm model object file [KSSampleObject.h](https://github.com/kunalsood/KSRealmTerminationSample/blob/main/KSRealmTerminationSample/KSSampleObject.h).

## Version of Realm and Tooling

Realm framework version: 5.3.2 (pre-built dynamic framework)
**UPDATE**: Sample project has now been updated to include Realm using SPM.

Realm Object Server version: N/A

Xcode version: 12.0 (Beta 3, Beta 4, Beta 5 & Beta 6)

iOS/OSX version: iOS 14 (Beta 3, Beta 4, Beta 5 & Beta 6)

[...]

PR #6722, opened 2020-08-26
Upgrade to core 6.0.23

tgoyne:


gongzhang:
@tgoyne `coreVersionStr` in `Swift.package` also needs to be updated in this PR, right? 😊

aalenliang:
When will this PR be merged?

tgoyne:
When it is passing the tests.

marciogranzotto:
Was this supposed to fix https://github.com/realm/realm-cocoa/issues/6671? Pulling the latest commit still gives me the same error. Also if I don't use my Realm DB in the app, I don't have the issue.

I'm running with cocopods like so:
```ruby
pod 'RealmSwift', :git => 'https://github.com/realm/realm-cocoa.git', :commit => 'ee357b28e6fec9fc25fa65a5e2afff063825a69c', :submodules => true
```

tgoyne:
This branch currently doesn't build at all when installing via cocoapods.

marciogranzotto:
Hm... cocoapods didn't throw any errors, but it's possible that it only falls back to the release version or something...
```bash
$ pod install --clean-install                                                                                                  [16:29:32]
Analyzing dependencies
Pre-downloading: `RealmSwift` from `https://github.com/realm/realm-cocoa.git`, commit `ee357b28e6fec9fc25fa65a5e2afff063825a69c`, submodules `true`
Downloading dependencies
Installing RealmSwift 5.3.5
Generating Pods project
Integrating client project
Pod installation complete! There are 21 dependencies from the Podfile and 56 total pods installed.
```

thodang888:
I tested my app on iOS 12 with release v5.3.6 and my app freezes. This issue doesn't occur if I don't use v5.3.6. Do anyone have the same problem with me?

SquaredTiki:
We've been seeing this too, raised #6749.

thodang888:
> We've been seeing this too, raised #6749.

With my situation, I reverted the Realm database to the last version that I used and recommended users should use iOS 14 beta 7.

PR #6723, opened 2020-08-27
Update dependancies and add tests for null partition value

leemaguire:
This PR adds some basic tests for opening a Realm on a `nil` partition key value. The Object Store commit has also been updated to the latest v10. References to backend services have also been removed from `dependancies.list` as the values are no longer read from there.


jsflax (changes requested):
This is still missing a few implementation details.

- Updating the method signature in objc to denote that the `id<RLMBSON>` can be nullable.
- Updating and/or figuring out what to do with the signature of `User.configuration(partitionValue:)` in Swift. A test case for swift would have likely caught this.

tgoyne:
The stitch version in dependencies.list is used [by the Jenkins setup to check out the appropriate version](https://ci.realm.io/job/cocoa-pr-trigger/configure). setup_baas.rb should read the version from there as well.

jsflax (approved):

PR #6742, opened 2020-09-08
Polish ObjectId & Decimal128

leemaguire:
Requires https://github.com/realm/realm-core/pull/3896 for completion


Realm.xcodeproj/project.pbxproj
```diff
@@ -3459,6 +3481,7 @@
 			isa = XCBuildConfiguration;
 			baseConfigurationReference = 5DD755E01BE05C19002800DA /* Tests.xcconfig */;
 			buildSettings = {
+				ALWAYS_EMBED_SWIFT_STANDARD_LIBRARIES = YES;
```
jsflax, line 3484: This seems out of place? @tgoyne

RealmSwift/ObjectId.swift
```diff
@@ -81,3 +93,46 @@ extension ObjectId: Encodable {
...
 }
+
+extension ObjectId: Comparable {
+    /// Returns a Boolean value indicating whether the value of the first
+    /// argument is less than that of the second argument.
+    ///
+    /// - Parameters:
+    ///   - lhs: An ObjectId value to compare.
+    ///   - rhs: Another ObjectId value to compare.
+    public static func < (lhs: ObjectId, rhs: ObjectId) -> Bool {
```
jsflax, line 104: Extra space after operators

Realm/RLMDecimal128.h
```diff
@@ -67,6 +73,49 @@ NS_ASSUME_NONNULL_BEGIN
 
 /// Gets if this Decimal128 represents a NaN value.
 @property (nonatomic, readonly) BOOL isNaN;
+
+/// The magnitude of this RLMDecimal128.
+@property (nonatomic, readonly) RLMDecimal128 *magnitude NS_REFINED_FOR_SWIFT;
+
+/// Replaces this RLMDecimal128 value with its additive inverse.
+- (void)negate NS_REFINED_FOR_SWIFT;
```
jsflax, line 81: Seems to me that `negate` is the only one here that should _not_ be `NS_REFINED_FOR_SWIFT` while the rest should be.

jsflax: I see now that you were just mirroring the implementation of other swift Number structs, but I still find this a bit odd.

Realm.xcodeproj/project.pbxproj
```diff
@@ -3338,6 +3358,7 @@
 					Realm/ObjectStore/external/json,
 				);
 				SWIFT_OPTIMIZATION_LEVEL = "-Onone";
+				SWIFT_VERSION = 5.0;
```
tgoyne, line 3361: This should not be set in the project file (it's set in the xcconfig files).

Realm.xcodeproj/project.pbxproj
tgoyne: There shouldn't be any reason that this is needed.

RealmSwift/Decimal128.swift
```diff
@@ -83,10 +83,16 @@ public final class Decimal128: RLMDecimal128, Decodable {
             throw DecodingError.dataCorruptedError(in: container, debugDescription: "Cannot convert value to Decimal128")
         }
     }
-}
 
-extension Decimal128 {
+    /// The mininum value for Decimal128
+    public static var min: Self {
+        minimumDecimalNumber as! Self
```
jsflax, line 89: Seems like minimum and maximum decimal number should both be refined for swift here.

RealmSwift/Decimal128.swift
```diff
@@ -122,3 +128,191 @@ extension Decimal128: ExpressibleByStringLiteral {
...
+    /// A type that represents the distance between two values.
+    public typealias Stride = Decimal128
+
+    /// Returns the distance from this Decimal128 to the given value, expressed as a
+    /// stride.
+    ///
+    /// - Parameter other: The Decimal128 value to calculate the distance to.
+    /// - Returns: The distance from this value to `other`.
+    public func distance(to other: Decimal128) -> Stride {
+        other-self
```
jsflax, line 292: Space? Same for the below

Realm/RLMDecimal128.h
```diff
@@ -67,6 +73,49 @@ NS_ASSUME_NONNULL_BEGIN
 
 /// Gets if this Decimal128 represents a NaN value.
 @property (nonatomic, readonly) BOOL isNaN;
+
+/// The magnitude of this RLMDecimal128.
+@property (nonatomic, readonly) RLMDecimal128 *magnitude NS_REFINED_FOR_SWIFT;
+
+/// Replaces this RLMDecimal128 value with its additive inverse.
+- (void)negate;
```
tgoyne, line 81: I don't think we want any mutating functions? NSDecimalNumber doesn't have them and they won't work how users expect them to (i.e. `[foo.decimalProp additionAssignmentWith:num]` won't actually mutate the managed object).

jsflax (approved):

Realm/RLMDecimal128.mm
```diff
@@ -139,4 +139,82 @@ - (NSString *)stringValue {
...
+}
+
+- (RLMDecimal128 *)decimalNumberByAdding:(RLMDecimal128 *)decimalNumber {
+    auto rhs = RLMObjcToDecimal128(decimalNumber);
+    return [[RLMDecimal128 alloc] initWithDecimal128:_value+rhs];
+}
+
+- (void)additionAssignmentWith:(RLMDecimal128 *)decimalNumber {
+    auto rhs = RLMObjcToDecimal128(decimalNumber);
+    _value+=rhs;
```
tgoyne, line 167: Spaces around operators for all of these.

Realm/Tests/Decimal128Tests.m
```diff
@@ -0,0 +1,178 @@
...
+#import <Realm/RLMDecimal128.h>
+
+@interface Decimal128Tests : RLMTestCase
+@end
+
+@implementation Decimal128Tests
+
+#pragma mark - Initialization
+
+- (void)testDecimal128Initialization {
```
tgoyne, line 29: The tests should check some values which cannot be represented as doubles.

Realm/RLMDecimal128.h
leemaguire: They are there more-so for the swift implementation. It should be the case that the managed object is mutated

```
- (void)additionAssignmentWith:(RLMDecimal128 *)decimalNumber {
    auto rhs = RLMObjcToDecimal128(decimalNumber);
    _value+=rhs;
}
```

tgoyne: Modifying the `realm::Decimal128` ivar doesn't write to the Realm (and can't, as it doesn't have any association to the Realm it was read from).

leemaguire: I've removed the assignment operators but it removes the `SignedNumeric` conformance for `Decimal128`

jsflax (approved):
Losing conformance to Strideable and SignedNumeric isn't great, but if we choose to add them back in, at least it's an additive change.

PR #6746, opened 2020-09-09
Add support for using a sync xframework for spm

tgoyne:
Using a prebuilt framework rather than building core from source lets us support sync. Setting the REALM_BUILD_FROM_SOURCE environment variable will restore the old behavior of building from source (but not supporting sync) as it's useful for debugging and sanitizer testing.

Most of the changes here are related to updating to Xcode 12 rather than the actual functional changes. SPM now emits a warning if there's a file in your source directory that's not explicitly included or excluded, so we now need a giant list of exclusions. This giant list results in the Project expression taking too long to type check, so it has to be split into a bunch of smaller expressions with explicit types to help the type checker.

### DIFF
--- a/.jenkins.yml
+++ b/.jenkins.yml
@@ -587,6 +587,10 @@ exclude:
     target: swiftpm
     configuration: Debug
 
+  - xcode_version: 11.3
+    target: swiftpm
+    configuration: Release
+
   - xcode_version: 11.4.1
     target: swiftpm
     configuration: Debug
@@ -610,10 +614,6 @@ exclude:
   - xcode_version: 11.6
     target: swiftpm
     configuration: Release
-
-  - xcode_version: 12.0
-    target: swiftpm
-    configuration: Debug
 
   - xcode_version: 11.3
     target: swiftpm-address
--- a/Package.swift
+++ b/Package.swift
@@ -1,14 +1,18 @@
-// swift-tools-version:5.0
+// swift-tools-version:5.3
 
 import PackageDescription
 import Foundation
 
 let coreVersionStr = "6.0.26"
-let cocoaVersionStr = "5.4.1"
+let syncVersionStr = "5.0.23"
+let cocoaVersionStr = "5.3.5"
+
+let buildFromSource = ProcessInfo.processInfo.environment["REALM_BUILD_FROM_SOURCE"] != nil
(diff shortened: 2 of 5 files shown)


tgoyne:
This isn't mergeable until Xcode 12 is released but it can be reviewed now. No really hurry, though.

.jenkins.yml
```diff
@@ -611,10 +615,6 @@ exclude:
     target: swiftpm
     configuration: Release
 
-  - xcode_version: 12.0
```
jsflax, line 614: Could you explain why this is being removed but 11.3 added?

build.sh
```diff
@@ -345,7 +345,7 @@ download_common() {
         version=$REALM_SYNC_VERSION
         url="${REALM_BASE_URL}/sync/realm-sync${kind}-${version}.tar.xz"
     else
-        echo "Unknown dowload_type: $download_type"
+        echo "Unknown download_type: $download_type"
```
jsflax, line 348: 🙈

jsflax (approved):
With this, seems like we should do a merge to v10 soon.

.jenkins.yml
tgoyne: This is the excludes section, so it's disabling building on 11 and enabling building on 12.

leemaguire (approved):

Package.swift
```diff
@@ -1,14 +1,18 @@
-// swift-tools-version:5.0
+// swift-tools-version:5.3
 
 import PackageDescription
 import Foundation
 
 let coreVersionStr = "6.0.26"
-let cocoaVersionStr = "5.4.1"
+let syncVersionStr = "5.0.23"
+let cocoaVersionStr = "5.3.5"
```
leemaguire, line 8: Cocoa version seems incorrect?

jsflax:
We've been testing swiftPM against this branch and it seems to be unable to pull in RealmSwift.

tgoyne:
On iOS this somehow works when building for the simulator but not when targeting a device.

tgoyne:
I've filed FB8706513 because it turns out that archiving an app which depends on a static xcframework via spm just doesn't work at all.

jsflax (commented):
This probably needs an update after merging

PR #6755, opened 2020-09-11
Add multiple realm apps support

jsflax:
Notably desingletonizes the SyncManager and leaves the RLMApp caching to the ObjectStore.

Unfortunately, the integration I added to test that this works... isn't actually working: `testAddObjectsMultipleApps`. The realms in the parent process turn up empty. So I could use a second set of eyes here @tgoyne


tgoyne:
There's two problems with the test:
1. `clientDataRoot` deletes the directory before returning the url so calling it twice does bad things.
2. `appId1` and `appId2` are not populated in the child process and `self.appIds[0]` is passed directly to the RLMApp initializer. This causes problems because the local variables are also used at the partition values below. Setting the local variables and pulling the RLMApp initialization outside of the `if` also eliminates some duplicated code.

tgoyne:
The client reset tests are failing because relying on the object store caching for App doesn't work due to RLMApp having it's own state. Specifically, `_ = App(id: "a"); _ = App(id: "a")` will create two `RLMSyncManager`s, and each `RLMSyncManager` has its own `errorHandler` (and a few other ivars not immediately relevant).

jsflax:
This now fixes some of the functionality in `watch` that wasn't working appropriately:
- Calls `isOpen` on event subscribers at the appropriate time
- Appropriately applies and uses delegate queues when an event fires
- The event buffer was not properly being fed or read from, hence why `watch` tests took so long 
- Fixes a race condition in tests that did not ensure that the watch stream was open before inserting documents/checking for events

This also fixes a bug that inappropriately placed `nil` into our `RLMBSONTypeDocument`, which is not allowed since you cannot set `nil` to collections. It now uses `[NSNull null]` instead.

I have also attempted to make `clientDataRoot` not overly delete-y.

@tgoyne PTAL

Realm/ObjectServerTests/RLMObjectServerTests.mm
```diff
@@ -2002,64 +1998,65 @@ - (void)testMongoFind {
...
-    [collection countWhere:@{@"name" : @"fido"}
-                     limit:1
-                completion:^(NSInteger count, NSError * error) {
-        XCTAssertEqual(count, 1);
-        XCTAssertNil(error);
-        [countExpectation2 fulfill];
-    }];
-    [self waitForExpectationsWithTimeout:60.0 handler:nil];
-}
+/// FIXME: Re-enable once we understand why the server is not setup correctly
```
tgoyne, line 2001: Temporarily disabling tests should be done by renaming the test rather than commenting it out so that it still gets compiled and doesn't bitrot if we rename things.

When things should be commented out temporarily it should be done with block comments rather than line comments so that it doesn't make the diff and history a mess.

Realm/ObjectServerTests/RLMBSONTests.mm
```diff
@@ -33,7 +33,7 @@ @implementation RLMBSONTestCase
 - (void)testNilRoundTrip {
     auto bson = Bson();
     id<RLMBSON> rlm = RLMConvertBsonToRLMBSON(bson);
-    XCTAssertEqual(rlm, nil);
+    XCTAssertTrue([rlm isEqual:[NSNull null]]);
```
tgoyne, line 36:
```suggestion
    XCTAssertEqual(rlm, [NSNull null]);
```

Realm/ObjectServerTests/RLMBSONTests.mm
```diff
@@ -122,7 +123,9 @@ - (void)testDocumentRoundTrip {
     auto bson = RLMConvertRLMBSONToBson(document);
     
     auto bsonDocument = static_cast<BsonDocument>(bson);
-    
+
+    XCTAssertTrue([document[@"nil"] isEqual:[NSNull null]]);
```
tgoyne, line 127:
```suggestion
    XCTAssertEqual(document[@"nil"], [NSNull null]);
```

Realm/ObjectServerTests/RLMObjectServerTests.mm
```diff
@@ -2393,10 +2390,11 @@ - (void)performWatchTest:(nullable dispatch_queue_t)delegateQueue {
                                                   expectation:expectation];
 
     __block RLMChangeStream *changeStream = [collection watchWithDelegate:testUtility delegateQueue:delegateQueue];
-
     dispatch_async(dispatch_get_global_queue(DISPATCH_QUEUE_PRIORITY_DEFAULT, 0), ^{
+        dispatch_semaphore_wait(testUtility.isOpenSemaphore, DISPATCH_TIME_FOREVER);
         for (int i = 0; i < 3; i++) {
             [collection insertOneDocument:@{@"name": @"fido"} completion:^(RLMObjectId * objectId, NSError * error) {
+                NSLog(@"!!!!!!!!trigger number %d", i);
```
tgoyne, line 2397:
```suggestion
```

Realm/RLMMongoCollection.mm
```diff
@@ -32,42 +32,52 @@ @implementation RLMChangeStream {
...
     __weak NSURLSession *_session;
+    dispatch_queue_t _queue;
 }
-- (instancetype)initWithChangeEventSubscriber:(id<RLMChangeEventDelegate>)subscriber {
+
+- (instancetype)initWithChangeEventSubscriber:(id<RLMChangeEventDelegate>)subscriber
+                                delegateQueue:(nullable dispatch_queue_t)queue {
     if (self = [super init]) {
         _subscriber = subscriber;
+        _queue = queue;
```
tgoyne, line 42:
```suggestion
        _queue = queue ?: dispatch_get_main_queue();
```

Better to just ensure the queue is valid here than check it each time.

Realm/RLMNetworkTransport.mm
```diff
@@ -198,6 +199,13 @@ + (instancetype)delegateWithEventSubscriber:(RLMEventSubscriber *)subscriber {
 - (void)URLSession:(__unused NSURLSession *)session
           dataTask:(__unused NSURLSessionDataTask *)dataTask
     didReceiveData:(NSData *)data {
+    {
+        std::lock_guard<std::mutex> lock(_eventMutex);
```
tgoyne, line 203: NSURLSessionDelegate methods are invoked on a serial queue so there shouldn't be any need for a mutex here.

jsflax: Ah, good to know, will remove 👍

tgoyne (approved):

PR #6758, opened 2020-09-14
Change API acronym to uppercase

leemaguire:
Remove any lowercase acronym of API

### DIFF
--- a/Realm/RLMAPIKeyAuth.h
+++ b/Realm/RLMAPIKeyAuth.h
@@ -40,7 +40,7 @@ typedef void(^RLMUserAPIKeysBlock)(NSArray<RLMUserAPIKey *> *  _Nullable, NSErro
   @param name The name of the API key to be created.
   @param completion A callback to be invoked once the call is complete.
 */
-- (void)createApiKeyWithName:(NSString *)name
+- (void)createAPIKeyWithName:(NSString *)name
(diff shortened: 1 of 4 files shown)


Realm/RLMAPIKeyAuth.h
```diff
@@ -40,7 +40,7 @@ typedef void(^RLMUserAPIKeysBlock)(NSArray<RLMUserAPIKey *> *  _Nullable, NSErro
   @param name The name of the API key to be created.
   @param completion A callback to be invoked once the call is complete.
 */
-- (void)createApiKeyWithName:(NSString *)name
+- (void)createAPIKeyWithName:(NSString *)name
                   completion:(RLMOptionalUserAPIKeyBlock)completion NS_SWIFT_NAME(createApiKey(named:completion:));
```
pavel-ship-it, line 44: Should we change Swift name too?

leemaguire: true

jsflax: Im not sure these are extended in Swift until Pavel's Combine PR.

pavel-ship-it: Yes, this is available in both callback and future forms

jsflax (approved):
Make sure tests are compiling before merging 😅 . As far as passing, I have a number of fixes in my branch that will cover us.

PR #6768, opened 2020-09-16
Fix Combine extensions visibility

pavel-ship-it:
This fix should hide extensions related to Combine feature and expose it inside Sync scope only.


jsflax (dismissed):
The code looks fine, but there are many SwiftLint failures and one docs failure. Please fix those and then I'll take another look.

leemaguire (approved):

jsflax (approved):

PR #6777, opened 2020-09-17
Fixing SwiftPM file to support iOS 9

diniska:
iOS 9 support is promised by the documentation (https://realm.io/docs/swift/latest/#installation) as well as specified in the cocoapods file (https://github.com/realm/realm-cocoa/blob/master/RealmSwift.podspec).

I think it is a good idea to align the SwiftPM file with those.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -32,7 +32,7 @@ let package = Package(
     name: "Realm",
     platforms: [
         .macOS(.v10_10),
-        .iOS(.v11),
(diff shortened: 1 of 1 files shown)


tgoyne:
We can't support iOS 9 when installing via spm because spm does not give us a way to pass the `-fno-aligned-new` compiler flag, which is required when building c++14 for 32-bit iOS.

PR #6793, opened 2020-09-22
Add Xcode 12.2 to the PR job matrix

tgoyne:


tgoyne:
Xcode 12.2 seems to be pretty broken currently. Deleting a file from within a simulator just hangs indefinitely.

PR #6794, opened 2020-09-22
Rename Realm.Publishers to RealmPublishers

pavel-ship-it:
Remove Realm.Publishers to avoid confusion with Combine.Publishers


jsflax:
The issue with doing this "as is" is that our publishers are too generically named and will likely lead to confusion with users.

leemaguire (approved):

PR #6796, opened 2020-09-22
Add iOS 9 support for SPM

goncharik:
Please review correct flags setup. 

Unfortunately there are no build flags separation for different archs in SPM for now so `-fno-aligned-new` is used for all architectures. 

Archive of Realm package has passed successfully.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -32,7 +32,7 @@ let package = Package(
     name: "Realm",
     platforms: [
         .macOS(.v10_10),
-        .iOS(.v11),
+        .iOS(.v9),
         .tvOS(.v9),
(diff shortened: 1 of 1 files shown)


tgoyne:
Using `unsafeFlags()` makes it so that the package cannot be imported by other packages:

> As some build flags can be exploited for unsupported or malicious behavior, the use of unsafe flags makes the products containing this target ineligible for use by other packages.

In Xcode this gives the error: 

> The package product 'Realm' cannot be used as a dependency of this target because it uses unsafe build flags.

goncharik:
Ok. Got it. Let's close this one. Thanks for response

PR #6816, opened 2020-09-29
Save NotificationToken from Combine call

leemaguire:
Adds the ability to capture a NotificationToken when using a Combine publisher that observes a Realm Object or Collection. The user will call `saveToken(on:at:)` directly after invoking the publisher to use the feature.


RealmSwift/Combine.swift
```diff
@@ -324,7 +324,7 @@ extension Realm {
     /// A publisher that emits Void each time the object changes.
     ///
     /// Despite the name, this actually emits *after* the collection has changed.
-    public var objectWillChange: Publishers.RealmWillChange {
+    public var objectWillChange: Publishers.RealmWillChange { // Maybe use `Never`? Would using Never confuse users?
```
jsflax, line 327: Stray comment

RealmSwift/Combine.swift
```diff
@@ -437,14 +439,14 @@ extension AnyRealmCollection: RealmSubscribable {
 /// A subscription which wraps a Realm notification.
 @available(OSX 10.15, watchOS 6.0, iOS 13.0, iOSApplicationExtension 13.0, OSXApplicationExtension 10.15, tvOS 13.0, *)
 public struct ObservationSubscription: Subscription {
-    private var token: NotificationToken
-    internal init(token: NotificationToken) {
-        self.token = token
+    public var cancellable: NotificationToken
```
jsflax, line 442: This requires a docstring now.

RealmSwift/Combine.swift
```diff
@@ -685,37 +852,46 @@ public enum Publishers {
     /// should always be the first operation in the pipeline.
     ///
     /// Create this publisher using the `objectChangeset()` function.
-    public struct ObjectChangeset<T: Object>: Publisher {
+    public struct ObjectChangeset<O: Object>: Publisher {
```
jsflax, line 855: Probably could've kept this `T`.

RealmSwift/Combine.swift
leemaguire: changed back to private, no need

RealmSwift/Combine.swift
leemaguire: It can but below it is `ObjectChangesetWithToken<O: Object, T>`. Maybe for consistency it should be `O`?

jsflax: Maybe– I'm not sure there is mandated naming convention here. Generally speaking, `T` for type, `U` for the next type (and so on).  `K`/`Key` for key and `V`/`Value`. But if `O` gives you consistency, then I wouldn't be terribly pedantic. As you were.

RealmSwift/Combine.swift
```diff
@@ -503,14 +544,57 @@ public enum Publishers {
         /// This publisher emits Void.
         public typealias Output = Void
 
+        private let collection: Collection
+
+        internal init(_ collection: Collection) {
+            self.collection = collection
+        }
+
+        /// Allows you to capture the `NotificationToken` produced by observing a Realm Collection.
```
tgoyne, line 553: This needs a bit more description of what it's doing and why you'd want to use it. It should mention notification skipping, how it makes that possible, and when the token is actually written to the target object/keypath.

RealmSwift/Combine.swift
```diff
@@ -503,14 +544,57 @@ public enum Publishers {
...
+        /// - Parameters:
+        ///   - object: The object which owns the `NotificationToken`.
+        ///   - keyPath: The KeyPath of the `NotificationToken`.
```
tgoyne, line 556: The object which the notification token is written to, and the keypath which is it written to.

RealmSwift/Combine.swift
```diff
@@ -503,14 +544,57 @@ public enum Publishers {
...
+        /// - Returns: A `WillChangeWithToken` Publisher.
+        public func saveToken<T>(on object: T, for keyPath: WritableKeyPath<T, NotificationToken?>) -> WillChangeWithToken<Collection, T> {
```
tgoyne, line 558: `for:` doesn't make sense here. Maybe `at:`?

CHANGELOG.md
```diff
@@ -1,3 +1,25 @@
+x.y.z Release notes (yyyy-MM-dd)
+=============================================================
+### Enhancements
+* Add ability to save a NotificationToken when using Combine for observing an Object or Collection by using `saveToken(on:for:)` before `sink` in a chain.
```
tgoyne, line 4: Using this doesn't require that `.sink` appears in the chain and the position is quite a bit more constrained than that.

RealmSwift/Combine.swift
```diff
@@ -486,10 +486,12 @@ public enum Publishers {
             self.realm = realm
         }
 
-        /// Allows you to capture the `NotificationToken` produced by observing a Realm.
+        /// Captures the `NotificationToken` produced by observing a Realm Collection. This allows you to do notification skipping
```
tgoyne, line 489: Paragraph break after the first sentence, as when the docstrings are rendered the first line is used as the short description and then everything after that is formatted a bit differently.

tgoyne:
Needs changelog entries for the bug fix pulled in via the object store upgrade.

tgoyne (approved):

PR #6817, opened 2020-09-29
Add a temporary solution for arm64 build issues with Cocoapods

leemaguire:
This should help users facing arm64 related build issues when integrating with Cocoapods (by both podfile & podspec dependency). The temporary solution is to explicitly exclude arm64 as an architecture to build on.


Realm.podspec
```diff
@@ -92,6 +92,7 @@ Pod::Spec.new do |s|
                                 'OTHER_CPLUSPLUSFLAGS' => '-isystem "${PODS_ROOT}/Realm/include/core" -fvisibility-inlines-hidden',
                                 'OTHER_CPLUSPLUSFLAGS[arch=armv7]' => '-isystem "${PODS_ROOT}/Realm/include/core" -fvisibility-inlines-hidden -fno-aligned-new',
                                 'USER_HEADER_SEARCH_PATHS' => '"${PODS_ROOT}/Realm/include" "${PODS_ROOT}/Realm/include/Realm"',
+                                'EXCLUDED_ARCHS[sdk=*]' => 'arm64'
```
tgoyne, line 95: This'll exclude it when building for device too. We need to only exclude it for simulator builds.

tgoyne (approved):

leemaguire:
Closing as adding excluded archs causes issues with Xcode 11 and would require users to use the old build system.

PR #6825, opened 2020-10-01
Change some methods on RLMApp, RLMUser to properties

leemaguire:
Changes some methods on RLMApp & RLMUser to properties where it sees fit.

### DIFF
--- a/Realm/RLMUser.h
+++ b/Realm/RLMUser.h
@@ -69,7 +69,7 @@ NS_ASSUME_NONNULL_BEGIN
 @property (nonatomic, readonly) NSString *identifier NS_SWIFT_NAME(id);
 
 /// Returns an array of identities currently linked to a user.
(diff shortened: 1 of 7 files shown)


Realm/RLMUser.h
```diff
@@ -66,29 +66,26 @@ NS_ASSUME_NONNULL_BEGIN
...
+@property (nonatomic, readonly) NSString *refreshToken;
 
 /**
  The user's refresh token used to access the Realm Application.
 
  This is required to make HTTP requests to MongoDB Realm's REST API
  for functionality not exposed natively. It should be treated as sensitive data.
  */
-@property (nullable, nonatomic, readonly) NSString *accessToken;
+@property (nonatomic, readonly) NSString *accessToken;
```
jsflax, line 88: Access token and refresh token actually can be nil, due to logged in/logged out state.

tgoyne (commented):
Needs changelog entries

tgoyne (approved):

PR #6833, opened 2020-10-05
Add onOpen(:) to watch Combine call

leemaguire:
This PR adds the ability to listen for when a Watch Change Stream is opened when using Combine. Use `onOpen(event:)` directly after opening a `WatchPublisher`, this will allow a callback to be invoked once the change stream is opened.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,24 @@
+x.y.z Release notes (yyyy-MM-dd)
+=============================================================
+### Enhancements
+* Add ability to listen for when a Watch Change Stream is opened when using Combine. Use `onOpen(event:)` directly after opening a `WatchPublisher`, this will allow a callback to be invoked once the change stream is opened.
+
+### Fixed
+* <How to hit and notice issue? what was the impact?> ([#????](https://github.com/realm/realm-cocoa/issues/????), since v?.?.?)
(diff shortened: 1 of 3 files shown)


jsflax (approved):
Good stuff. Shame testMultipleRegisterDevice is still failing 😩

CHANGELOG.md
```diff
@@ -1,3 +1,24 @@
+x.y.z Release notes (yyyy-MM-dd)
+=============================================================
+### Enhancements
+* Add ability to listen for when a Watch Change Stream is opened when using Combine. Use `onOpen(event:)` directly after opening a `WatchPublisher`, this will allow a callback to be invoked once the change stream is opened.
```
tgoyne, line 4:
```suggestion
* Add the ability to listen for when a Watch Change Stream is opened when using Combine. Use `onOpen(event:)` directly after opening a `WatchPublisher` to register a callback to be invoked once the change stream is opened.
```

The second sentence is a comma splice.

tgoyne (approved):
I think this probably isn't the correct way to implement this concept with Combine, but the obvious most Combiney way (send an enum with .open or .changeEvent(bson)) seems like it'd be awkward in practice and if there's a better way to do it I don't know what it is.

PR #6841, opened 2020-10-06
Switch to building against the monorepo

tgoyne:
~~Currently has three sets of failing tests:~~

1. ~~Some Decimal128 functionality hasn't been merged from core v10 yet~~
2. ~~The indexed value has changed on primary keys. Need to look into if this is a correct change, and if so update tests.~~
3. ~~Queries which compare a link property to an unmanaged object relied on a hack which has been broken by core changes and needs a proper solution.~~

Now passes all tests


fealebenpae:
@tgoyne is there anything I can help with, or is Cocoa all set to start building against the monorepo instead of binary artifacts?

tgoyne:
This part is done, but the other two build systems still need to be updated. I believe Lee is currently working on xcframework packaging for monorepo and no one has touched cocoapods yet.

tgoyne:
This needs a new core release before it's mergeable but otherwise is ready to be reviewed.  The changes to source files are mostly just boring mechanical changes to update the include paths but the query builder changes and build system stuff could use a look-over.

Realm/RLMUtil.mm
```diff
@@ -426,7 +426,7 @@ id RLMMixedToObjc(realm::Mixed const& mixed) {
             return [[RLMObjectId alloc] initWithValue:mixed.get<realm::ObjectId>()];
         case realm::type_Link:
         case realm::type_LinkList:
-        case realm::type_OldMixed:
+        // case realm::type_OldMixed:
```
jsflax, line 429: Can we just remove this?

jsflax (approved):

leemaguire (approved):
LGTM, the CI seems very red though.

tgoyne:
The 10.3.0 core release still doesn't work (it's missing a header) so everything but SPM will fail with it.

tgoyne:
Remaining CI failures are now just the pre-existing cocoapods catalyst issue and the `testBasicWithoutNotifying` test.

PR #6844, opened 2020-10-07
Throw an exception for Object children without @objc properties.

pavel-ship-it:


CHANGELOG.md
```diff
@@ -2,6 +2,7 @@ x.y.z Release notes (yyyy-MM-dd)
 =============================================================
 ### Enhancements
 * Add the ability to listen for when a Watch Change Stream is opened when using Combine. Use `onOpen(event:)` directly after opening a `WatchPublisher` to register a callback to be invoked once the change stream is opened.
+* Throw an exception for Object children without @objc properties.
```
leemaguire, line 5:
```suggestion
* Throw an exception for Objects that have none of its properties marked with @objc.
```

Realm/RLMSchema.mm
```diff
@@ -95,6 +95,9 @@ @implementation RLMSchema {
         s_sharedSchema.objectSchemaByName[schema.className] = schema;
     }
 
+    if ([cls shouldIncludeInDefaultSchema] && schema.isSwiftClass && schema.properties.count == 0) {
+        @throw RLMException(@"No properties was defined for '%@'. Did you remember to add '@obj' in your model?", schema.className);
```
leemaguire, line 99:
```suggestion
        @throw RLMException(@"No properties are defined for '%@'. Did you remember to mark them with '@obj' in your model?", schema.className);
```

leemaguire (changes requested):
This needs a test

pavel-ship-it:
There is a problem @leemaguire - it will throw an exception before any test will start it's cycle.

tgoyne:
You can look at SchemaTests.mm to see examples of how we test schema errors. The basic idea is that they inherit from a different base class which won't be included in the default schema.

Realm/RLMSchema.mm
```diff
@@ -95,6 +95,9 @@ @implementation RLMSchema {
         s_sharedSchema.objectSchemaByName[schema.className] = schema;
     }
 
+    if ([cls shouldIncludeInDefaultSchema] && schema.isSwiftClass && schema.properties.count == 0) {
```
tgoyne, line 98: This isn't the right place for this check. It should go with the existing validation logic in RLMObjectSchema.mm.

Realm/RLMObjectSchema.mm
```diff
@@ -173,6 +173,9 @@ + (instancetype)schemaForObjectClass:(Class)objectClass {
         }
     }
 
+    if ([objectClass shouldIncludeInDefaultSchema] && schema.isSwiftClass && schema.properties.count == 0) {
+        @throw RLMException(@"No properties are defined for '%@'. Did you remember to mark them with '@obj' in your model?", schema.className);
```
tgoyne, line 177: Missing the 'c' in `@objc`

leemaguire (approved):

tgoyne (approved):

Issue #5665, opened 2018-03-12
Bug: Comparing two List<String> objects fails


tomaskraina:
## Goals
Compare 2 List<String> objects with the same items.

## Expected Results
The objects should be equal.

## Actual Results
The objects are not equal.

## Steps to Reproduce
1. Create 2 List<String> objects.
2. Append the same Array<String> to them
3. Compare them

## Code Sample
```
class RealmExampleTests: XCTestCase {
    
    func testRealmListOfStringsComparison() {
    
        let strings = ["a", "b", "c"]
        let list1 = List<String>()
        list1.append(objectsIn: strings)
        
        let list2 = List<String>()
        list2.append(objectsIn: strings)

        // 1. ok
        XCTAssertTrue(list1 !== list2, "instances should not be identical")

        // 2. fails
        XCTAssertTrue(list1 == list2, "instances should be equal by `==` operator")

        // 3. fails
        XCTAssertTrue(list1.isEqual(list2), "instances should be equal by `isEqual` method")

        // 4. ok
        XCTAssertTrue(Array(list1) == Array(list2), "instances converted to Swift.Array should be equal")
    }
    
}
```

## Version of Realm and Tooling

Realm framework version: 3.1.1
Realm Object Server version: N/A
Xcode version: 9.2
iOS/OSX version: 10.12.6
Dependency manager + version: CocoaPods 1.3.1

```
ProductName:	Mac OS X
ProductVersion:	10.12.6
BuildVersion:	16G1114

/Applications/Xcode-9.2.app/Contents/Developer
Xcode 9.2
Build version 9C40b

/Users/tomkraina/.rbenv/shims/pod
1.3.1
Realm (3.1.1)
RealmSwift (3.1.1)

/bin/bash
GNU bash, version 3.2.57(1)-release (x86_64-apple-darwin16)

/usr/local/bin/carthage
0.18.1
(not in use here)

/usr/bin/git
git version 2.14.3 (Apple Git-98)
```

PR #6845, opened 2020-10-07
Fix comparing unmanaged lists

leemaguire:
When comparing two unmanaged `List<>`/`RLMArray` objects, the backing array was not used for the comparison. Therefor doing any sort of comparison on `List<>` or `RLMArray` would always give the incorrect result.

Fixes: https://github.com/realm/realm-cocoa/issues/5665

### DIFF
--- a/Realm/RLMArray.mm
+++ b/Realm/RLMArray.mm
@@ -514,6 +514,13 @@ static bool canAggregate(RLMPropertyType type, bool allowDate) {
     return [_backingArray objectsAtIndexes:indexes];
 }
 
+- (BOOL)isEqual:(id)object {
+    if (![object isKindOfClass:[RLMArray class]]) {
+        return NO;
+    }
+    return [_backingArray isEqual:((RLMArray *)object)->_backingArray];
+}
+
 - (void)addObserver:(NSObject *)observer forKeyPath:(NSString *)keyPath
             options:(NSKeyValueObservingOptions)options context:(void *)context {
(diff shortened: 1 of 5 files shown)


Realm/RLMArray.mm
```diff
@@ -514,6 +514,13 @@ - (NSArray *)objectsAtIndexes:(NSIndexSet *)indexes {
     return [_backingArray objectsAtIndexes:indexes];
 }
 
+- (BOOL)isEqual:(id)object {
+    if (![object isKindOfClass:RLMArray.class]) {
+        return NO;
+    }
+    return [_backingArray isEqual:((RLMArray *)object)->_backingArray];
+}
```
tgoyne, line 522: This'll return false for `[[RLMArray new] isEqual:[RLMArray new]]` because `_backingArray` is nil. I think the following is correct?

```suggestion
    if (auto array = RLMDynamicCast<RLMArray>(object)) {
        return !array.realm
            && ((_backingArray.count == 0 && array->_backingArray.count == 0)
                || [_backingArray isEqual:array->_backingArray]);
    }
    return NO;
```

The !array.realm check is needed as otherwise an empty unmanaged array will compare equal to all managed arrays. Checking that the count of both is 0 rather than checking for `nil` is because a nil backing array and an empty backing array are equal here.

leemaguire: yep, good spot. Updated.

tgoyne (approved):

PR #6855, opened 2020-10-13
Add support for swift library evolution mode

tgoyne:
Define RealmSwift.Object in obj-c rather than swift so that it doesn't get built in evolution mode, as that result in subclasses of it lazily generating obj-c metadata and breaking our schema discovery.


jsflax (approved):
Nice

PR #6856, opened 2020-10-13
Prepare final v10 changelog

tgoyne:


CHANGELOG.md
```diff
@@ -1,3 +1,106 @@
...
+  `-[RLMMongoCollection insertOneDocument:completion:]`, `-[RLMMongoCollection
+  insertManyDocuments:completion:]`, `-[RLMMongoCollection
+  updateOneDocument:completion:]`, `-[RLMMongoCollection
+  updateManyDocuments:completion:]`, `-[RLMMongoCollection
+  deleteOneDocument:completion:]`, and `-[RLMMongoCollection
+  deleteManyDocuments:completion:]`. If you are already familiar with MongoDB
+  drivers, it is important to understand that the remote MongoCollection only
+  provides access to the operations available in MongoDB Realm.
+* Change `[RLMUser
+  configurationWithPartitionValue:]`/`User.configuration(with:)` to accept
```
nirinchev, line 74: This should probably be reworded for the combined changelog as `[RLMUser configurationWithPartitionValue:]` didn't exist pre-v10.

CHANGELOG.md
```diff
@@ -1,3 +1,106 @@
+10.0.0 Release notes (yyyy-MM-dd)
+=============================================================
+
+This release is functionally identical to v10.0.0-rc.2.
+
+NOTE: This version upgrades the Realm file format to version 20. Realm files
```
nirinchev, line 6: We should not be using the version of the file format (was agreed some time ago that this is not something we want exposed publicly). Instead, we should be using more generic terms like "The internal layout of the Realm file has changed. Realm files will be automatically upgraded when opened..."

CHANGELOG.md
jsflax: That's a fair critique, this should be edited to be in the context of adding, not changing.

jsflax (commented):
We should add some more detail about the switch from everything being contained in `SyncUser` to `App`. Also the fact that app exists now at all.

PR #6858, opened 2020-10-14
Update core and object store and merge master

tgoyne:
This turned out to be rather involved as there were several breaking changes in OS that we hadn't pulled in yet.


Realm/RLMMongoDatabase_Private.hpp
```diff
@@ -17,18 +17,17 @@
...
+@class RLMUser;
 
 @interface RLMMongoDatabase ()
 
-@property (nonatomic, strong) RLMApp *app;
+@property (nonatomic, strong) RLMUser *user;
 @property (nonatomic, strong) NSString *serviceName;
 
-- (instancetype)initWithApp:(RLMApp *)app
+- (instancetype)initWithUser:(RLMUser *)user
```
leemaguire, line 30: indentation looks slightly off

leemaguire (approved):

Realm/ObjectServerTests/RLMObjectServerTests.mm
```diff
@@ -515,7 +513,7 @@ - (void)testInvalidPasswordAuthentication {
 
     XCTestExpectation *expectation = [self expectationWithDescription:@"login should fail"];
 
-    [self.app loginWithCredential:credentials completion:^(RLMUser * user, NSError * error) {
+    [self.app loginWithCredential:credentials completion:^(RLMUser * user, NSError *error) {
```
jsflax, line 516: There are a bunch of extra spaces in this file before `*`, but they were already there

Realm/RLMMongoCollection.mm
```diff
@@ -93,31 +95,41 @@ - (void)close {
 
 @implementation RLMMongoCollection
 
-- (instancetype)initWithApp:(RLMApp *)app
-                serviceName:(NSString *)serviceName
-               databaseName:(NSString *)databaseName
-             collectionName:(NSString *)collectionName {
+static realm::bson::BsonDocument toBsonDocument(id<RLMBSON> bson) {
```
jsflax, line 98: Should this be prefixed by `RLM`?

RealmSwift/RealmConfiguration.swift
```diff
@@ -200,7 +200,11 @@ extension Realm {
         /// The classes managed by the Realm.
         public var objectTypes: [ObjectBase.Type]? {
             get {
-                return self.customSchema.map { $0.objectSchema.compactMap { $0.objectClass as? Object.Type } }
+                return self.customSchema.map {
+                    $0.objectSchema.compactMap {
+                        $0.objectClass as? Object.Type ?? $0.objectClass as? EmbeddedObject.Type
```
jsflax, line 205: Can we just leave a comment here saying what this is doing? It's fairly obvious when reading it with context, but it certainly looks strange.

Realm/RLMMongoCollection.mm
tgoyne: It's static, and only exported functions should be prefixed.

PR #6864, opened 2020-10-15
Add return Result<Value, Error> for paired returns

pavel-ship-it:
Each completion with return of (value, error) should have a completion with the return of Result type


RealmSwift/MongoClient.swift
```diff
@@ -155,6 +155,21 @@ public typealias MongoUpdateBlock = (UpdateResult?, Error?) -> Void
 /// Block which returns the deleted Document on a successful delete operation, or an error should one occur.
 public typealias MongoDeleteBlock = RLMMongoDeleteBlock
 
+/// Block which returns Result.success(DocumentId) on a successful insert or Result.failure(error)
+public typealias MongoInsertResultBlock = (Result<AnyBSON?, Error>) -> Void
```
jsflax, line 159: Some of these result types should not be nullable.

tgoyne:
I think it'd be better to replace the existing functions rather than add the Result version as an additional overload. Having both is mostly just confusing for people getting started with the API.

jsflax (commented):
Agreed with Thomas.

RealmSwift/MongoClient.swift
tgoyne: I think AnyBSON? is actually correct. BSON null is translated to `nil` and I think null is a legal document id?

CHANGELOG.md
```diff
@@ -2,6 +2,7 @@ x.y.z Release notes (yyyy-MM-dd)
 =============================================================
 ### Enhancements
 * Add the ability to listen for when a Watch Change Stream is opened when using Combine. Use `onOpen(event:)` directly after opening a `WatchPublisher` to register a callback to be invoked once the change stream is opened.
+* For paired return from completion handlers add return Result<Value, Error>
```
jsflax, line 5: This is a breaking change.

It also should start with an imperative:
```suggestion
* Change paired return types from Swift completion handlers to return `Result<Value, Error>`
```

RealmSwift/MongoClient.swift
jsflax: Hmm... you are correct. I still find that to be a strange use case.

pavel-ship-it: There is a possible nullable successful result in some cases. I'll check it another time as it is not for all collection methods.

tgoyne (approved):

jsflax (approved):

Issue #6801, opened 2020-09-23
MaximumNumberOfActiveVersions debug feedback


RikvanderLinde:
## Goals

I want to keep the realm file on disk small, and to stop all bug related to it (multiple versions, unable to compress on launch)
## Expected Results

I would like to query the Realm instance for number of active versions, to keep track of possible problem area's

## Actual Results

You can set maximumNumberOfActiveVersions to a number, run your app and see if it triggers or not

I would like to request an option for me to live debug (or observe) how many active versions I have at a current point in time, making it easier to locate and repair relevant problems.
Currently I have resorted to checking the realm size on disk after every write, resulting in random increases in size that are nerver reclaimed until the app compresses on launch. So, if the app runs for too long, the size becomes too big, cant load anymore and bricks the app.

There have been multiple issues that I have been tracking about these problems, and while offering Ideas on what could cause the problem, they dont provide an active means of problem tracking, other then debugging by triggering MaximumNumberOfActiveVersions.

The cost of the call would not be relevant, as this would be a debug only check, letting you track the points where you are "version leaking"

leemaguire:
We've considered this request and added it to our backlog to be implemented.

PR #6869, opened 2020-10-15
Add numberOfActiveVersions property to RLMRealm

leemaguire:
This adds a `numberOfActiveVersions ` property onto `RLMRealm`. Its use is for helping users debug where a Realm instance might be getting pinned.

Addresses: https://github.com/realm/realm-cocoa/issues/6801

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,7 @@
 x.y.z Release notes (yyyy-MM-dd)
 =============================================================
 ### Enhancements
-* Add ability to view the active number of a versions of a Realm by using `RLMRealm.numberOfActiveVersions`. The purpose of this is to
+* Add `RLMRealm.numberOfActiveVersions` which returns the number of versions of a Realm that's currently in use. The purpose of this is to
   help remedy a situation where a Realm is growing in size and the user can not figure out where in their codebase the Realm may be getting pinned.
 
 ### Fixed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,26 @@
+x.y.z Release notes (yyyy-MM-dd)
+=============================================================
+### Enhancements
+* Add ability to view the active number of a versions of a Realm by using `RLMRealm.numberOfActiveVersions`. The purpose of this is to
+  help remedy a situation where a Realm is growing in size and the user can not figure out where in their codebase the Realm may be getting pinned.
+
+### Fixed
+* <How to hit and notice issue? what was the impact?> ([#????](https://github.com/realm/realm-cocoa/issues/????), since v?.?.?)
+* None.
+
+<!-- ### Breaking Changes - ONLY INCLUDE FOR NEW MAJOR version -->
+
+### Compatibility
+* File format: Generates Realms with format v11 (Reads and upgrades all previous formats)
+* Realm Object Server: 3.21.0 or later.
+* Realm Studio: 5.0.0 or later.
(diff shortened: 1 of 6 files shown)


pavel-ship-it (approved):

Realm/RLMRealm.h
```diff
@@ -815,6 +815,16 @@ NS_REFINED_FOR_SWIFT;
  */
 - (struct RLMClassPrivileges)privilegesForClassNamed:(NSString *)className;
 
+#pragma mark - Utilities
+
+/**
+ Returns the current active number of versions of a Realm instance.
```
tgoyne, line 821: I don't think users will have any clue what this means and it isn't really accurate. "instance" in particular makes no sense here; we usually use that to refer to RLMRealm objects and that's not applicable to this.

leemaguire: would simply saying ' Returns the number of active versions of this Realm.' suffice?

CHANGELOG.md
```diff
@@ -1,3 +1,26 @@
+x.y.z Release notes (yyyy-MM-dd)
+=============================================================
+### Enhancements
+* Add ability to view the active number of a versions of a Realm by using `RLMRealm.numberOfActiveVersions`. The purpose of this is to
```
bmunkholm, line 4:
```suggestion
* Add `RLMRealm.numberOfActiveVersions` which returns the number of versions of a Realm that's currently in use. The purpose of this is to
```

Realm/RLMRealm.h
```diff
@@ -815,6 +815,16 @@ NS_REFINED_FOR_SWIFT;
...
 - (struct RLMClassPrivileges)privilegesForClassNamed:(NSString *)className;
 
+#pragma mark - Utilities
+
+/**
+ Returns the number of active versions of this Realm.
+
+ There will always be at least 2 active versions of the Realm
+ because the previous version won't be cleaned up until the next commit.
+*/
```
tgoyne, line 825: Maybe something like `@see RLMRealmConfiguration.maximumNumberOfActiveVersions` since that tries to describe what active versions mean. I'm not sure that that's the correct syntax.

saltyskip:
Any updates here, would have to have this merged in for logging purposes

leemaguire:
@saltyskip we've decided to shelve this PR for the moment in favour of creating a larger debugging suite down the line.

saltyskip:
@leemaguire do you have any ETAs on the debugging suite? The version increases is by far the biggest problem we are experiencing using realm and with the absence of any debugging tools its making it very difficult to work with. 

If it is still very far out, can we consider merging this in as a temporary measure?

leemaguire:
@saltyskip We do not have any ETA at the moment, nor can we merge this as a temporary solution. A suggestion would be to fork Realm-Cocoa and add these changes in yourself if they will help you debug.

saltyskip:
Thanks for the response @leemaguire 

The main thing is that realm threading/pinned versioning issues have been an issue dating back many years now and this was the first time that we ever had a method of detecting a possible issue at runtime and having a more graceful method to recover. It's very similar to the memoryWarning API that is often used by iOS developers to take the appropriate steps needed when their app starts reaching a memory threshold.

Ideally the code is perfect and the developer never introduces realms getting accidentally pinned, but in practice there can always be some production issue where this could happen. Our only option at this time is to crash the app because we have no warning that pinned realms are starting to grow or surpass appropriate thresholds. 

I think there would be many developers who would benefit from the ability to be able to detect issues at runtime in production applications and I hope you will consider again to include this.

PR #6870, opened 2020-10-16
Py/result type callbacks

pavel-ship-it:
Remove old login and linkUser
Update asyncOpen

### DIFF
--- a/Realm/ObjectServerTests/SwiftObjectServerTests.swift
+++ b/Realm/ObjectServerTests/SwiftObjectServerTests.swift
@@ -505,8 +505,9 @@ class SwiftObjectServerTests: SwiftSyncTestCase {
             let ex = expectation(description: "async open")
             let config = user.configuration(partitionValue: self.appId)
             Realm.asyncOpen(configuration: config) { result in
-                if case .failure = result {
-                    XCTFail("No realm on async open")
+                guard case .failure = result else {
+                    XCTFail("No error on cancelled async open")
+                    return ex.fulfill()
                 }
                 ex.fulfill()
             }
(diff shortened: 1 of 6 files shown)


Realm/ObjectServerTests/SwiftObjectServerTests.swift
```diff
@@ -501,11 +504,14 @@ class SwiftObjectServerTests: SwiftSyncTestCase {
 
             let ex = expectation(description: "async open")
             let config = user.configuration(partitionValue: self.appId)
-            Realm.asyncOpen(configuration: config) { _, error in
-                XCTAssertNotNil(error)
+            Realm.asyncOpen(configuration: config) { result in
+                switch result {
+                case .failure: XCTFail("No realm on async open")
+                default: break
+                }
```
tgoyne, line 511:
```suggestion
                if case .failure = result {
                    XCTFail("No realm on async open")
                }
```

RealmSwift/Realm.swift
```diff
@@ -131,10 +131,16 @@ import Realm.Private
     @discardableResult
     public static func asyncOpen(configuration: Realm.Configuration = .defaultConfiguration,
                                  callbackQueue: DispatchQueue = .main,
-                                 callback: @escaping (Realm?, Swift.Error?) -> Void) -> AsyncOpenTask {
-        return AsyncOpenTask(rlmTask: RLMRealm.asyncOpen(with: configuration.rlmConfiguration, callbackQueue: callbackQueue) { rlmRealm, error in
-            callback(rlmRealm.flatMap(Realm.init), error)
-        })
+                                 callback: @escaping (Result<Realm, Swift.Error>) -> Void) -> AsyncOpenTask {
+        return AsyncOpenTask(rlmTask: RLMRealm.asyncOpen(with: configuration.rlmConfiguration,
```
tgoyne, line 135: I think switching this away from trailing closure syntax is a net reduction in readability due to that it results in the body of the callback being indented so much.

pavel-ship-it: agree

tgoyne (approved):

tgoyne (approved):

PR #6886, opened 2020-10-23
Add RLMSupport.swift to RealmSwift

ericjordanmossman:
This PR aims to enable developers to interact with RLMRealm, RLMObject subclasses and the v10 RealmSwift API easily.
Previous support for the Obj-c API in Swift came from downloading and adding [RLMSupport.swift](https://realm.io/docs/objc/5.5.0/#rlmsupport-swift) file to a project's `Realm` target. 

RealmSupport.swift needed to be updated to account for all v10 changes to the Obj-c api. But using RLMSupport.swift to wrap v10 objc functions through the previous pattern is unsustainable because developers would need to bridge cast swift types to cocoa types (very unintuitive), or the RLMSupport.swift file would require forked code from `RealmSwift` ([unsustainable](https://github.com/realm/realm-cocoa/pull/6882/)).

The simpler solution is to include RLMSupport.swift in the `RealmSwift` target. For a mixed obj-c/swift project, the developer then 

- continues to define their models in obj-c as `RLMObject` subclasses and accesses them via `RLMRealm` rather than `Realm`. 

- Otherwise uses RealmSwift APIs in Swift. This is possible because the `typealias` nature of most `RealmSwift` classes.


ericjordanmossman:
**[Jenkins: catalyst Release 12.0](https://ci.realm.io/job/objc_pr/configuration=Release,target=catalyst,xcode_version=12.0/7249/consoleFull)** : Fails due to an "unexpected exit, crash, or test timeout in ResultsTests/testQueryMemoryUsage". Passes in previous builds and locally.
```
Oct 25 21:23:32 Failing tests:
Oct 25 21:23:32 	Tests:
Oct 25 21:23:32 		-[ResultsTests testQueryMemoryUsage]
Oct 25 21:23:32 
Oct 25 21:23:32 ** TEST FAILED **
```

**[Jenkins: osx Debug 11.3](https://ci.realm.io/job/objc_pr/configuration=Debug,target=osx,xcode_version=11.3/7249/consoleFull)** : Succeeded but couldn't connect to CI after tests(?)
```
Oct 25 21:08:45 Executed 1303 tests, with 0 failures (0 unexpected) in 61.834 (62.517) seconds
Oct 25 21:08:46 ▸ Test Succeeded
...
FATAL: Remote call on JNLP4-connect connection from 10.61.12.35/10.61.12.35:58470 failed
```
**[Jenkins: swiftpm Release 12.0](https://ci.realm.io/job/objc_pr/configuration=Release,target=swiftpm,xcode_version=12.0/7249/)** : Succeeded but couldn't connect to CI after tests(?)
```
Oct 25 21:26:14 Executed 3008 tests, with 0 failures (0 unexpected) in 33.940 (34.288) seconds
...
FATAL: Remote call on JNLP4-connect connection from 10.61.12.35/10.61.12.35:58470 failed
```

**[Jenkins: tvos-swift Release 12.0](https://ci.realm.io/job/objc_pr/configuration=Release,target=tvos-swift,xcode_version=12.0/7249/)** : Succeeded but couldn't connect to CI after tests(?)
```
Oct 25 21:18:58 Executed 1704 tests, with 0 failures (0 unexpected) in 8.810 (9.505) seconds
...
FATAL: Remote call on JNLP4-connect connection from 10.61.12.35/10.61.12.35:58470 failed
```

ericjordanmossman:
If this approach to objc/swift projects is merged, we'd have to update documentation too.

Realm/Swift/RLMSupport.swift
```diff
@@ -19,32 +19,55 @@
...
                               _ args: CVarArg...) -> RLMResults<RLMObject> {
         return objects(in: realm, with: NSPredicate(format: predicateFormat, arguments: getVaList(args))) as! RLMResults<RLMObject>
     }
 }
 
-public struct RLMIterator<T>: IteratorProtocol {
+// Sequence conformance for RLMArray and RLMResults is provided by RLMCollection's
+// `makeIterator()` implementation.
+extension RLMArray: Sequence {}
+extension RLMResults: Sequence {}
```
jsflax, line 65: These extensions should be public.

Realm/Swift/RLMSupport.swift
```diff
@@ -19,32 +19,55 @@
 import Realm
 
 extension RLMRealm {
+    /**
+     Swift wrapper for `+ [RLMRealm schemaVersionAtURL:`.
```
jsflax, line 23: Bad formatting.

Realm/Swift/RLMSupport.swift
```diff
@@ -19,32 +19,55 @@
...
+     */
     @nonobjc public class func schemaVersion(at url: URL, usingEncryptionKey key: Data? = nil) throws -> UInt64 {
         var error: NSError?
         let version = __schemaVersion(at: url, encryptionKey: key, error: &error)
         guard version != RLMNotVersioned else { throw error! }
         return version
     }
 
+    /**
+     Swift wrapper for ` - [RLMRealm resolveThreadSafeReference:]`.
```
jsflax, line 34: Weird formatting.

Realm/Swift/RLMSupport.swift
ericjordanmossman: Which is better?
```
    /**
     Returns the schema version for a Realm at a given local URL.
     
     - see: `RLMRealm.schemaVersionAtURL`
     */
```
or 
```
    /**
     Returns the schema version for a Realm at a given local URL.
     
     - see: `- [RLMRealm schemaVersionAtURL:encryptionKey:error:]`
     */
```

ericjordanmossman: We have a couple cases where we reference swift methods in comments, but not Objc. We only use `- see` for RLM classes. So I'm kinda guessing.

Realm/Swift/RLMSupport.swift
ericjordanmossman: Extensions that conform to protocols can't be public:
Screen Shot 2020-11-06 at 1 17 35 PM

The developer only accesses `Sequence` functionality through the `public func makeIterator()`. So something like this is still possible:
```
let rlmRealm = try! RLMRealm.default()
        try! rlmRealm.transaction {
            for i in 1...3 {
                let person = Person(value: ["age": "\(i)")
                rlmRealm.add(person)
            }
        }

        let rlmResults = Person.allObjects()
        var iterator = rlmResults.makeIterator()
        while let person = iterator.next() {
            // do something
        }
```

ericjordanmossman:
@tgoyne 
This is ready for another look. I'm going to make a pull request in realm.io to update the RLMSupport section in parallel.

tgoyne (approved):

ericjordanmossman:
Related docs changes: https://github.com/realm/realm.io/pull/4751

szotp:
@ericjordanmossman @tgoyne I know I'm digging this out, but it was not added to [RealmSwift.podspec](https://github.com/realm/realm-swift/blob/master/RealmSwift.podspec).

dianaafanador3:
Hi @szotp do you have a mixed obj-c/Swift Project?, Any reason to not migrate the implementation completely to Swift?,
What it is the specific use case so you need `RLMSupport.swift`?

szotp:
@dianaafanador3 yes, we do have a mixed project, and we will for the foreseeable future (due to legacy code).

Note that I was able to fix it. I added RLMSupport.swift by manually to my project, even though docs state that it should not be necessary:
> Prior to 10.2.0 RealmSwift doesn’t include this file by default. If using a version of Realm prior to this, we recommend you compile the [Swift/RLMSupport.swift](https://github.com/realm/realm-cocoa/blob/master/Realm/Swift/RLMSupport.swift) file (which is also available in our [release zip](https://static.realm.io/downloads/objc/realm-objc-10.5.1.zip)).

It's a minor issue because a workaround exists, but just wanted to let you know.

PR #6894, opened 2020-10-27
Improved metrics

ericjordanmossman:
Includes parallel upload to MongoDB Realm webhook.

### DIFF
--- a/Realm/RLMAnalytics.mm
+++ b/Realm/RLMAnalytics.mm
@@ -242,14 +242,16 @@ void RLMSendAnalytics() {
     if (getenv("REALM_DISABLE_ANALYTICS") || !RLMIsDebuggerAttached() || RLMIsRunningInPlayground()) {
         return;
     }
-
-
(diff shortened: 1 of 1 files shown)


ericjordanmossman:
@tgoyne @jsflax 
I'm not sure what to make of these two latest failures. Both happen because of timeouts. They succeeded in [previous re-builds](https://ci.realm.io/job/objc_pr/7253/) and locally. And in `osx-object-server Release 12.0` I think the equals assertion happens because the delete was supposed to happen in the wait.

```
Oct 28 09:19:22     ✗ testMultipleRealmsDeleteObjects, Asynchronous wait failed: Exceeded timeout of 20 seconds, with unfulfilled expectations: "Wait for download completion".
Oct 28 09:19:42     ✗ testMultipleRealmsDeleteObjects, ((0) equal to ([self runChildAndWait])) failed: ("0") is not equal to ("1") - Tests in child process failed
```

ericjordanmossman:
Is a changelog entry needed for this?

tgoyne (approved):

tgoyne:
We set `REALM_DISABLE_ANALYTICS` when running CI since we obviously don't care about our own builds, so none of the code changed is actually run on CI and can't be responsible for any failures. We generally don't mention analytics things in the changelog since it shouldn't be relevant to the users.

PR #6926, opened 2020-11-11
Add testing for Mac Catalyst using Cocoapods

leemaguire:
This PR adds some tests to the build script which allows us to verify the installation of Realm via Cocoapods on a Mac Catalyst target. A new folder called `catalyst` has been added under the `examples/installation` directory which holds the Swift & Obj-c test projects.

The Jenkinsfile explicitly declares the builds because if we are to just tack on `"catalyst"` to:

```
for (def platform in ["osx", "ios", "watchos"]) {
      def platformName = platformNames[platform]
      for (def test in ["dynamic", "cocoapods", "carthage"]) {
        parallelBuilds["Installation - ${platformName} Swift ${test}"] = installationTest(platform, test, 'swift')
      }
    }
```

It would mean that this PR would also need to add test projects for both `dynamic` and `carthage`. Which of course can be done but I'll leave that open to discussion.


jsflax (approved):

jsflax:
Some of the licenses need updated dates, otherwise LGTM.

examples/installation/build.sh
```diff
@@ -135,14 +140,18 @@ xctest() {
...
+
+    if [[ $PLATFORM != osx ]] && [[ $PLATFORM != catalyst ]]; then
         [[ $PLATFORM == 'ios' ]] && SDK=iphoneos || SDK=$PLATFORM
-        if [ -d "$workspace" ]; then
-            [[ $LANG == 'swift' ]] && scheme=(-scheme RealmSwift) || scheme=(-scheme Realm)
-        else
-            scheme=()
-        fi
         xcodebuild "${project[@]}" "${scheme[@]}" -sdk "$SDK" build "${code_signing_flags[@]}"
+    elif [[ $PLATFORM == catalyst ]]; then
```
tgoyne, line 152: Moving this above the other case so that it's "if catalyst, else if not osx" rather than "if not osx and not catalyst, else if catalyst" makes the conditionals a little less weird.

tgoyne (approved):

PR #6927, opened 2020-11-11
Update LICENSE

bmunkholm:
With Sync open source we can remove part of the license.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,8 +1,7 @@
 TABLE OF CONTENTS
 
 1. Apache License version 2.0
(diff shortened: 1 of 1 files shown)


tgoyne:
We haven't switched over to using the open source version of sync yet and the binary library we're currently using has stuff not present in the open source release, so it seems like the binary license is still applicable.

tgoyne:
Oh, this is pointed at the monorepo branch, nevermind.

tgoyne (approved):

PR #6931, opened 2020-11-12
Not allow set 'deleteRealmIfMigrationNeeded' when sync is enabled.

pavel-ship-it:
Ending session with error: failed to validate upload changesets: clients cannot upload destructive schema changes, received "EraseColumn" instruction (ProtocolErrorCode=212)

This is a bug, the client should not be emitting this instruction, There was a bug in Object Store semi-recently that caused it to be emitted if the SDK passed configuration to the Object Store in a particular way (realm-js had this problem), The problematic config option that cause this in realm-js was deleteRealmIfMigrationNeeded

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@ x.y.z Release notes (yyyy-MM-dd)
 * None.
 
 ### Fixed
-* Not allow set 'deleteRealmIfMigrationNeeded' when sync is enabled.
+* Disallow `RLMRealmConfiguration.deleteRealmIfMigrationNeeded`/`Realm.Config.deleteRealmIfMigrationNeeded` when sync is enabled.
 
 <!-- ### Breaking Changes - ONLY INCLUDE FOR NEW MAJOR version -->
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,23 @@
+x.y.z Release notes (yyyy-MM-dd)
+=============================================================
+### Enhancements
+* None.
+
+### Fixed
+* Not allow set 'deleteRealmIfMigrationNeeded' when sync is enabled.
+
+<!-- ### Breaking Changes - ONLY INCLUDE FOR NEW MAJOR version -->
+
+### Compatibility
+* Realm Studio: 10.0.0 or later.
+* APIs are backwards compatible with all previous releases in the 10.x.y series.
+* Carthage release for Swift is built with Xcode 12.1.
+* CocoaPods: 1.10 or later.
+
+### Internal
+* Upgraded realm-core from ? to ?
+* Upgraded realm-sync from ? to ?
+
 10.1.2 Release notes (2020-11-06)
 =============================================================
 
--- a/Realm/ObjectServerTests/RLMObjectServerTests.mm
+++ b/Realm/ObjectServerTests/RLMObjectServerTests.mm
@@ -2580,4 +2580,12 @@ static const NSInteger NUMBER_OF_BIG_OBJECTS = 2;
     [self waitForExpectations:@[expectation] timeout:60.0];
 }
 
+- (void)testSyncConfigShouldNotMigrate {
+    RLMUser *user = [self userForTest:_cmd];
+    NSString *realmId = self.appId;
+    RLMRealm *realm = [self openRealmForPartitionValue:realmId user:user];
+
+    XCTAssertThrows([realm.configuration setDeleteRealmIfMigrationNeeded:YES], @"Not allow 'deleteRealmIfMigrationNeeded' when sync is enabled.");
+}
+
 @end
--- a/Realm/ObjectServerTests/RLMObjectServerTests.mm
+++ b/Realm/ObjectServerTests/RLMObjectServerTests.mm
@@ -2586,6 +2586,9 @@ static const NSInteger NUMBER_OF_BIG_OBJECTS = 2;
     RLMRealm *realm = [self openRealmForPartitionValue:realmId user:user];
 
     XCTAssertThrows([realm.configuration setDeleteRealmIfMigrationNeeded:YES], @"Not allow 'deleteRealmIfMigrationNeeded' when sync is enabled.");
+    
+    RLMRealmConfiguration *localRealmConfiguration = [RLMRealmConfiguration defaultConfiguration];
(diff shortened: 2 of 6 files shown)


Realm/RLMRealmConfiguration.mm
```diff
@@ -233,6 +233,8 @@ - (void)setDeleteRealmIfMigrationNeeded:(BOOL)deleteRealmIfMigrationNeeded {
     if (deleteRealmIfMigrationNeeded) {
         if (self.readOnly) {
             @throw RLMException(@"Cannot set `deleteRealmIfMigrationNeeded` when `readOnly` is set.");
+        } else if (isSync(_config) && !_config.sync_config->partition_value.empty()) {
+            @throw RLMException(@"Cannot set 'deleteRealmIfMigrationNeeded' when sync is enabled ('sync.partitionValue' is set).");
```
jsflax, line 237:
```suggestion
            @throw RLMException(@"Cannot set 'deleteRealmIfMigrationNeeded' to true when sync is enabled ('syncConfig' is set).");
```

Realm/ObjectServerTests/RLMObjectServerTests.mm
```diff
@@ -2580,4 +2580,12 @@ - (void)performMultipleWatchStreamsTest:(nullable dispatch_queue_t)delegateQueue
     [self waitForExpectations:@[expectation] timeout:60.0];
 }
 
+- (void)testSyncConfigShouldNotMigrate {
+    RLMUser *user = [self userForTest:_cmd];
+    NSString *realmId = self.appId;
+    RLMRealm *realm = [self openRealmForPartitionValue:realmId user:user];
+
+    XCTAssertThrows([realm.configuration setDeleteRealmIfMigrationNeeded:YES], @"Not allow 'deleteRealmIfMigrationNeeded' when sync is enabled.");
```
jsflax, line 2588: This test should also test that valid configurations still work, even though it is tested elsewhere. Should also write on for swift (why not?).

CHANGELOG.md
```diff
@@ -1,3 +1,23 @@
+x.y.z Release notes (yyyy-MM-dd)
+=============================================================
+### Enhancements
+* None.
+
+### Fixed
+* Not allow set 'deleteRealmIfMigrationNeeded' when sync is enabled.
```
jsflax, line 7:
```suggestion
* Disallow `RLMRealmConfiguration.deleteRealmIfMigrationNeeded`/`Realm.Config.deleteRealmIfMigrationNeeded` when sync is enabled.
```

Realm/RLMRealmConfiguration.mm
```diff
@@ -233,6 +233,8 @@ - (void)setDeleteRealmIfMigrationNeeded:(BOOL)deleteRealmIfMigrationNeeded {
     if (deleteRealmIfMigrationNeeded) {
         if (self.readOnly) {
             @throw RLMException(@"Cannot set `deleteRealmIfMigrationNeeded` when `readOnly` is set.");
+        } else if (isSync(_config) && !_config.sync_config->partition_value.empty()) {
```
leemaguire, line 236: Is there any way Sync can be enabled with an empty partition value?

leemaguire: Also won't this need some `#if REALM_ENABLE_SYNC` guards here?

Realm/ObjectServerTests/RLMObjectServerTests.mm
pavel-ship-it: Ok

Realm/RLMRealmConfiguration.mm
pavel-ship-it: Fixed, both

Realm/RLMRealmConfiguration.mm
```diff
@@ -234,6 +234,11 @@ - (void)setDeleteRealmIfMigrationNeeded:(BOOL)deleteRealmIfMigrationNeeded {
         if (self.readOnly) {
             @throw RLMException(@"Cannot set `deleteRealmIfMigrationNeeded` when `readOnly` is set.");
         }
+#if REALM_ENABLE_SYNC
```
leemaguire, line 237: You don't need those here

RealmSwift/RealmConfiguration.swift
```diff
@@ -184,7 +184,19 @@ extension Realm {
...
          - note: Setting this property to `true` doesn't disable file format migrations.
          */
-        public var deleteRealmIfMigrationNeeded: Bool = false
+        public var deleteRealmIfMigrationNeeded: Bool {
+            get {
+                return _deleteRealmIfMigrationNeeded
+            }
+            set(newValue) {
+                if newValue && syncConfiguration != nil {
+                    throwRealmException("Cannot set 'deleteRealmIfMigrationNeeded' when sync is enabled ('syncConfig' is set).")
```
leemaguire, line 193: A throw on a computed property isn't correct, also anything that should should declare `throws`. RLMRealmConfiguration will perform the throw.

Realm/ObjectServerTests/SwiftObjectServerTests.swift
```diff
@@ -3311,4 +3311,17 @@ class CombineObjectServerTests: SwiftSyncTestCase {
             .store(in: &cancellable)
         wait(for: [dergisterDeviceExpectation], timeout: 4.0)
     }
+
+    func testShouldNotDeleteOnMigrationWithSync_2() {
```
leemaguire, line 3315: What does _2 mean?

leemaguire (changes requested):

RealmSwift/RealmConfiguration.swift
tgoyne: throwRealmException throws an obj-c exception and not a swift error, so `throws` isn't applicable.

Realm/ObjectServerTests/SwiftObjectServerTests.swift
pavel-ship-it: shouldn't be there

Realm/RLMRealmConfiguration.mm
pavel-ship-it: Yes, it's guarded in isSync

jsflax (approved):
Good stuff 👍

leemaguire (approved):

PR #6933, opened 2020-11-13
Add decode of $oid to ObjectID

demzpaul:
Added to the ObjectID decoder so that $oid wrapped ids are deserialized.

`{"_id": "507f1f77bcf86cd799439011"}`

And now:

`{"_id": {"$oid":"507f1f77bcf86cd799439011"}}`

both decode to valid ObjectID objects.


jsflax:
This is going to be worked on in the near future, with full xJSON support.

PR #6939, opened 2020-11-14
Make Results, Objects, Lists, and Properties more ergonomic with SwiftUI

jsflax:
Currently, `Results` cannot properly bind to SwiftUI views. Views will not re-render when underlying updates to the table occur.

Since `Results` is a `struct`, normally, you would be able to use `@State` to bind to the view, but since the underlying value of `Results` does not actually change, the view will not automagically respond to updates to the underlying table. `@State` variables must be mutated _from the context of the view_ as well, so having a custom property wrapper that attempts to mutate State through composition is a flawed approach (it doesn't work at all).

So, because of the way Realm emits updates, all Realm classes/structs bound to the view must be classes, and must be `ObservableObjects`. Since `Results` Is a `struct`, we need a wrapper class. To avoid "getting in the way," this is all handled by property wrappers. The final result (which works) is as follows:

```swift
import RealmSwift
import SwiftUI

class Person : Object, Identifiable {
    @objc dynamic var name = names.randomElement()!
}

struct ResultsView: View {
    @Environment(\.realm) var realm: Realm
    @BindResults(Person.self) var results

    var body: some View {
        VStack {
            Button(action: {
                try! realm.write { realm.add(Person()) }
            }, label: {
                Text("Add")
            })
            List {
                ForEach(results) { person in
                    Text(person.name)
                }
            }
        }
    }
}

@main
struct ContentView: SwiftUI.App {
    var realm = try! Realm()

    var view: some View {
        ResultsView().environment(\.realm, realm)
    }

    var body: some Scene {
        WindowGroup {
            view
        }
    }
}
```
 
The Realm Environment variable is also a nice added "feature" that will help users pass realms between views (and help *us* access the realms they are using).

I'm going to put this in draft state at first, as there are discussions to be had around this. Remaining open question:
- Should we make `BindResults` more generic (`RealmBind` or something) to allow for other collection types? This would allow us to also wrap the functionality we have been chatting about (auto freeze/thaw, auto write).

cc @ianpward @astigsen @cbush


astigsen:
> Should we make BindResults more generic (RealmBind or something) to allow for other collection types?

We should at a minimum make sure that it works on `RealmCollection`, so it can be used for both `Results` and `List`.

jsflax:
The latest commit enables this syntax:
```swift
class Dog: Object, Identifiable {
    // note, this will be awkward for Sync users who want to conform to identifiable.
    @objc dynamic var id = ObjectId.generate()
    @objc dynamic var name = randomAdjectives.randomElement()!
}

class Person: Object, Identifiable {
    @objc dynamic var name = randomNouns.randomElement()!
    var dogs = RealmSwift.List<Dog>()
}

struct PersonView: View {
    @BindObject var person: Person

    var body: some View {
        VStack {
            TextField("name", text: $person.name)
                .font(Font.largeTitle.bold()).padding()
            List {
                ForEach($person.dogs, id: \.id) { dog in
                    TextField("dog name", text: dog.name)
                }
                .onDelete(perform: $person.dogs.remove)
                .onMove(perform: $person.dogs.move)
            }
        }
        .navigationBarItems(trailing: Button("Add Dog") {
            $person.dogs.append(Dog())
        })
    }
}

struct ResultsView: View {
    @Environment(\.realm) var realm: Realm
    @BindResults(Person.self) var results

    var body: some View {
        NavigationView {
            List {
                ForEach(results) { person in
                    NavigationLink(destination: PersonView(person: person)) {
                        Text(person.name)
                    }
                }
            }
            .navigationBarTitle("People", displayMode: .large)
            .navigationBarItems(trailing: Button("Add") {
                try! realm.write { realm.add(Person()) }
            })
        }
    }
}

@main
struct ContentView: SwiftUI.App {
    var realm = try! Realm()

    var view: some View {
        ResultsView().environment(\.realm, realm)
    }

    var body: some Scene {
        WindowGroup {
            view
        }
    }
}
```

This is a significant improvement over what we currently have (see the Quick Start to compare). It implicitly writes, freezes, thaws, and abstracts all the frustrating boilerplate users have to do right now. I'm not crazy keen on the `$` syntax, but it's what SwiftUI users have to use anyway when it comes to `Binding` (a SwiftUI struct), so it feels native.

View bound Realm types assume the intention to write, e.g., `$person.dogs.append(Dog())` is automagically wrapped in a write, and `TextField("dog name", text: dog.name)` will also automagically write on edit.

There's still some work to be done: view bound Realms, BoundResults should pass back BoundObjects (not regular Objects), ideally generalising our propertyWrapper to be a single, generic type (e.g., `RealmBinding`), and of course cleanup. But before I continue, I'd like to make sure this is the direction we want to go in. To be frank, given the experimenting we've all been doing, I don't see a cleaner option.

jsflax:
I've updated the PR to have a single `RealmBind` type and added an example.

RealmSwift/List.swift
```diff
@@ -526,7 +538,7 @@ extension List: RealmCollection {
 }
 
 // MARK: - MutableCollection conformance, range replaceable collection emulation
-extension List: MutableCollection {
+extension List: MutableCollection, RangeReplaceableCollection {
```
tgoyne, line 541: List deliberately doesn't conform to RangeReplaceableCollection because not all of the operations it requires make sense for List.

RealmSwift/SwiftUI.swift
```diff
@@ -0,0 +1,395 @@
...
+                self.value[keyPath: member].freeze()
+            }, set: {
+                self.value[keyPath: member] = $0
+            })
+        }
+    }
+}
+
+@available(OSX 10.15, watchOS 6.0, iOS 13.0, iOSApplicationExtension 13.0, OSXApplicationExtension 10.15, tvOS 13.0, *)
+final class BoundResults<Element: RealmCollectionValue>: ResultsBase, Bound {
```
tgoyne, line 96: It seems like BoundResults could be replaced entirely with an extension on RLMResults.

RealmSwift/SwiftUI.swift
```diff
@@ -0,0 +1,395 @@
...
+                self.value[keyPath: member]
+            }, set: { newValue in
+                try! self.realm!.write {
+                    self.value[keyPath: member] = newValue
+                }
+            })
+        }
+    }
+
+    // if property is a list, we want to freeze it before fetching it
```
tgoyne, line 165: This isn't specific to List; it applies to all realm collections.

RealmSwift/SwiftUI.swift
```diff
@@ -0,0 +1,395 @@
...
+                self.wrappedValue[keyPath: member] = $0
+            })
+        }
+    }
+}
+
+
+// Allows conformance with SwiftUI's `ForEach`
+@available(iOS 13.0, macOS 10.15, tvOS 13.0, watchOS 6.0, *)
+extension Binding: Identifiable where Value: RealmCollectionValue, Value: Identifiable {
```
tgoyne, line 208: Declaring conditional conformances on other people's types is a no-no. There can only be one conformance for a type to a protocol, so this will break if another library (or the app) tries to do the same thing.

jsflax: For `Identifiable`, I take your point, as that may be common (though it will make getting the `\.id` key path more awkward for the end user).

But for the List binding, can we simply say that's a cost of using Realm with SwiftUI? A user wanting to conform `Binding` to `RandomAccessCollection`, `Sequence`, `Collection`, `BidirectionalCollection` seems like it would be highly unlikely?

I have to think about this. There might be an alternative path here (though I fear this was a fundamental part of my implementation).

jsflax: Suppose we may not need to extend Binding at all...

ianpward:
Is this PR moving forward or is it covered in some other branch?

jsflax:
@ianpward it will move forward on this branch. I am busy with a few higher priority tasks at the moment.

jsflax:
Closing for #7045

PR #6945, opened 2020-11-17
Add support for running the object server tests from the swift package

tgoyne:
The package is pretty awkward due to that spm doesn't support mixed-language targets, so the single target in xcode has to be turned into four separate targets (two for the support things which are used in both obj-c and swift, and two for the actual tests). Rather than going up to three dummy targets for the xcode project, I got rid of the RealmTestSupport dummy target and switched to using canImport(). All of the things now marked as public are because they're defined in a separate target from where they're used with spm.

Importing a Swift framework from obj-c doesn't work correctly with SPM, so I just manually declared the interfaces we need in obj-c.

SPM runs all of the test suites in a single process, which results in our sync tests seeing the object types for the non-sync tests. This fails schema validation, so the sync tests now need to all explicitly list the object types they use rather than relying on automatic discovery.

Automatically setting up BaaS from SPM turned out to be very difficult and I wanted to preserve the option to run the tests without the sync tests, so I decided to require manually calling setup-baas if you want to run the sync tests.


Realm/ObjectServerTests/RLMSyncTestCase.mm
```diff
@@ -419,6 +423,23 @@ - (void)manuallySetRefreshTokenForUser:(RLMUser *)user value:(NSString *)tokenVa
 
 #pragma mark - XCUnitTest Lifecycle
 
++ (XCTestSuite *)defaultTestSuite {
+    if ([RealmServer haveServer]) {
+        return [super defaultTestSuite];
+
```
leemaguire, line 429: [minor] extra newline

Realm/ObjectServerTests/SwiftSyncTestCase.swift
```diff
@@ -90,9 +117,9 @@ class SwiftSyncTestCase: RLMSyncTestCase {
         return theUser
     }
 
-    func synchronouslyLogOutUser(_ user: User,
-                                 file: StaticString = #file,
-                                 line: UInt = #line) throws {
+    public func synchronouslyLogOutUser(_ user: User,
```
leemaguire, line 120: Should the 'synchronously' prefix be removed to make it consistent with `openRealm`?

leemaguire (approved):
Couple of minor comments

jsflax (approved):
🎉

Issue #6960, opened 2020-11-26
build.sh xcframework failing to link for Catalyst


ppamorim:
## Goals
I would like to compile the library into an `xcframework` containing the libraries for `ios-swift` and `catalyst-swift`.

## Expected Results
`build.sh` should create a `xcframework` with the libraries described above.

## Actual Results

Crash when compiling using Xcode 12:

<img width="700" alt="Screenshot 2020-11-26 at 14 09 21" src="https://user-images.githubusercontent.com/5521682/100360616-0c9a6180-2ff1-11eb-9781-2d99f2217115.png">

## Steps for others to Reproduce

Clone the project, sync the code (submodule) and run `./build.sh xcframework`.

## Code Sample

Not applicable.

## Version of Realm and Tooling

Realm framework version: 10.1.4

Realm Object Server version: ?

Xcode version: 12.2

iOS/OSX version: ?

Dependency manager + version: Submodules (all in sync)

### Additional question:

The documentation in the `build.sh` file says that I can call `build.sh xcframework [plats]` but I am not sure how to use it. Analysing the code I can't find any line that reads the arguments from `[plats]`, is that right? I would like to create the `xcframework` only for `ios` and `catalyst`.

leemaguire:
@ppamorim you should use `./build.sh xcframework ios catalyst` to build for just iOS and Catalyst.

ppamorim:
I tried to compile the library twice, one framework for `iOS` and another for `Catalyst`:

<img width="762" alt="Screenshot 2020-11-26 at 14 51 53" src="https://user-images.githubusercontent.com/5521682/100365131-fa232680-2ff6-11eb-88cd-4d44ef20d65e.png">

But I am getting this error when compiling the Catalyst project:

`.../ios/Realm.framework/Headers/RLMPlatform.h:32:2: Attempting to use Realm's iOS framework in a non-iOS target.`

<img width="461" alt="Screenshot 2020-11-26 at 14 50 33" src="https://user-images.githubusercontent.com/5521682/100365078-e4adfc80-2ff6-11eb-9a6d-8ef70357284a.png">

ppamorim:
@leemaguire I am trying to compile it again with your command.

leemaguire:
I do see an issue here related to how Catalyst uses macOS style versioned frameworks

ppamorim:
@leemaguire Same issue with your command: 

[...]

PR #6962, opened 2020-11-27
Add placeholder.cpp to compile sources

leemaguire:
placeholder.cpp was not part of the compile sources for the targets Realm & Realm iOS static. This meant that when building an XCFramework from build.sh the final steps of assembling the XCFramework would fail as doing `ar -t tmp.a | grep -v placeholder | tr '\n' '\0' | xargs -0 ar -d tmp.a >/dev/null 2>&1` would produce nothing, therefor there would be nothing to link against. This PR simply adds placeholder.cpp back into the compile sources.

Addresses: https://github.com/realm/realm-cocoa/issues/6960


tgoyne:
The placeholder.o that we're trying to keep here is coming from core since it's a workaround for a cmake quirk.

ironage (approved):
I'm not very familiar with the build steps at this level but the changes seem fine to me. Do note that core is removing placeholder.cpp beginning from the monorepo releases. I wasn't aware that it was used at this level, so hopefully there is another way around this issue here going forward.

tgoyne:
This logic is all going away with the monorepo stuff anyway.

jsflax:
@tgoyne @leemaguire should we close this then?

PR #6964, opened 2020-12-01
Switch to packaging releases as XCFrameworks

tgoyne:
Turns out that `-allow-internal-distribution` exists and lets us package a Swift framework as an xcframework without using library evolution mode, which eliminates the need to do a bunch of awkward things to work around that. Switching to XCFrameworks lets us properly support all platforms with the prebuilt frameworks and simplifies the setup process (just drag the xcframework into your project and you're done, rather than having to pick the correct platform and add the strip frameworks step).

Most of the changes here are just updating the examples to use the new packaging. The Swift playground example isn't tested and so turned out to have bitrotted a bit. I deleted the RubyMotion example entirely because it's been years since anyone actually gave any indication that they were using RubyMotion and we never really got it working well to begin with.

https://ci.realm.io/blue/organizations/jenkins/cocoa-pipeline/detail/cocoa-pipeline/1293/pipeline is an almost-successful run of cocoa-pipeline using this; the catalyst tests fail due to https://github.com/CocoaPods/CocoaPods/pull/10224.


leemaguire (approved):
Nice cleanup!

jsflax (approved):

PR #6969, opened 2020-12-02
Add MutableSet<> / RLMSet datatype

leemaguire:
Adds implementation for `MutableSet<>` / `RLMSet` datatype. These are based on the respective `NSSet` and `Set`API's.

This PR currently relies on https://github.com/realm/realm-core/pull/4150 to be merged first.


Package.swift
```diff
@@ -62,7 +62,8 @@ let package = Package(
             targets: ["Realm", "RealmSwift"]),
     ],
     dependencies: [
-        .package(url: "https://github.com/realm/realm-core", .exact(Version(coreVersionStr)!))
+        //.package(url: "https://github.com/realm/realm-core", .branch("lm/os-set"))
+        .package(path: "../realm-core")
```
pavel-ship-it, line 66: is it belongs to the local changes only?

leemaguire: Yep, I need a PR merged in core

PR #6971, opened 2020-12-03
Actually enable encryption in the encryption-enabled tests

tgoyne:
Turns out that the encryption-enabled CI job hasn't actually been enabling encryption for the last few years because defaultConfiguration returns a copy, so mutating it doesn't do anything.

### DIFF
--- a/Realm/Tests/RealmConfigurationTests.mm
+++ b/Realm/Tests/RealmConfigurationTests.mm
@@ -96,7 +96,7 @@
     RLMRealmConfiguration *defaultConfiguration = [RLMRealmConfiguration defaultConfiguration];
     XCTAssertEqualObjects(defaultConfiguration.fileURL, RLMDefaultRealmURL());
     XCTAssertNil(defaultConfiguration.inMemoryIdentifier);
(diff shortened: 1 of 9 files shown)


Realm/Tests/RealmConfigurationTests.mm
```diff
@@ -96,7 +96,7 @@ - (void)testDefaultConfiguration {
     RLMRealmConfiguration *defaultConfiguration = [RLMRealmConfiguration defaultConfiguration];
     XCTAssertEqualObjects(defaultConfiguration.fileURL, RLMDefaultRealmURL());
     XCTAssertNil(defaultConfiguration.inMemoryIdentifier);
-    XCTAssertNil(defaultConfiguration.encryptionKey);
+    XCTAssertEqual(!!defaultConfiguration.encryptionKey, self.encryptTests);
```
jsflax, line 99: Double unary `!`?

tgoyne: To convert the pointer value to a boolean, since it should be non-nil only if encryptTests is true.

jsflax (approved):

PR #6975, opened 2020-12-03
"Thaw" frozen objects

ericjordanmossman:
This PR adds `.thaw()`. Developers can use this to get a live version of a previously frozen object. This is helpful for passing a collection into a SwiftUI view that will mutate values. ex)

```
var body: some View {
        List {
            ForEach(items.freeze()) { frozenItem in
                ItemRow(item: frozenItem.thaw())
            }
        }
    }
```

Thaw works by taking the configuration of the collection's frozen realm, opens a live realm, then resolves a new reference.


ericjordanmossman:
Moved out of draft.

Some [swiftpm](https://ci.realm.io/job/objc_pr/configuration=Release,target=swiftpm,xcode_version=11.3/7349/consoleFull) `CombinePublisherTests<Results>` are failing with one more call than expected. ex)
```
Dec 07 07:18:23   testBasicWithoutNotifying, XCTAssertEqual failed: ("2") is not equal to ("1")
```
**Could the PR be causing this?** Issue with CI? They succeed locally.

**[Jenkins: cocoapods-catalyst Release 11.3](https://ci.realm.io/job/objc_pr/configuration=Release,target=cocoapods-catalyst,xcode_version=11.3/7349/consoleFull)**
This is the first I've seen this. **Appears CI related(?)**
```
Dec 07 07:16:24 xcodebuild: error: Failed to build workspace CocoaPodsExample with scheme CocoaPodsExample.
Dec 07 07:16:24 	Reason: Cannot test target “CocoaPodsExampleTests” on “My Mac”: My Mac’s macOS 10.15.5 doesn’t match CocoaPodsExampleTests’s macOS 11.0 deployment target.
```

**[Jenkins: cocoapods-catalyst Release 12.2](https://ci.realm.io/job/objc_pr/configuration=Release,target=cocoapods-catalyst,xcode_version=12.2/7349/consoleFull)**
Claims build succeeds: `Dec 07 07:16:24 ** BUILD SUCCEEDED **`
Target fails like above:
```
Dec 07 07:16:24 xcodebuild: error: Failed to build workspace CocoaPodsExample with scheme CocoaPodsExample.
Dec 07 07:16:24 	Reason: Cannot test target “CocoaPodsExampleTests” on “My Mac”: My Mac’s macOS 10.15.5 doesn’t match CocoaPodsExampleTests’s macOS 11.0 deployment target.
```

**[Jenkins: osx Release 12.1](https://ci.realm.io/job/objc_pr/configuration=Release,target=osx,xcode_version=12.1/7349/consoleFull)**
```
0x7fa7227e64c0 (My Mac | x86_64h)> (10.15.5 (19F101))
Dec 07 06:48:19 objc[17627]: Memory corruption in class list.
Dec 07 06:48:19 build.sh: line 112: 17627 Abort trap: 6           xcodebuild -IDECustomDerivedDataLocation=build/DerivedData -scheme Realm -configuration Release test SWIFT_VERSION=5.3 GCC_GENERATE_DEBUGGING_SYMBOLS=NO -allowProvisioningUpdates -workspace examples/osx/objc/RealmExamples.xcworkspace
```
**I need help understanding this one.** Other 12.1 release build and local builds succeed.

Realm/RLMResults.h
```diff
@@ -395,6 +395,18 @@ __attribute__((warn_unused_result));
  */
 - (instancetype)freeze;
 
+/**
+ Returns a live, mutable version of this frozen collection.
+
+ This method resolves a reference to a live copy of the same frozen collection.
+ If a called on a non-frozen collection, will return itself.
```
jsflax, line 402: Weird grammar

Realm/RLMClassInfo.mm
```diff
@@ -76,10 +76,9 @@
     return linkTargetType(&property - &objectSchema->persisted_properties[0]);
 }
 
-RLMClassInfo &RLMClassInfo::freeze(__unsafe_unretained RLMRealm *const frozenRealm) {
-    REALM_ASSERT(frozenRealm.frozen);
+RLMClassInfo &RLMClassInfo::resolve(__unsafe_unretained RLMRealm *const realm) {
     // FIXME
```
jsflax, line 80: Can this FIXME be removed now?

Realm/RLMArray.h
```diff
@@ -492,6 +492,18 @@ __attribute__((warn_unused_result));
  */
 - (instancetype)freeze;
 
+/**
+ Returns a live, mutable version of this frozen collection.
+
+ This method resolves a reference to a live copy of the same frozen collection.
+ If a called on a non-frozen collection, will return itself.
```
jsflax, line 499: Weird grammar

Realm/RLMManagedArray.mm
```diff
@@ -505,14 +506,31 @@ - (instancetype)freeze {
...
+        return self;
+    }
+
+    RLMRealmConfiguration *rlmConfig = _realm.configuration;
+    RLMRealm *liveRealm = [RLMRealm realmWithConfiguration:rlmConfig error:nil];
+
+    auto tsr = realm::ThreadSafeReference(_backingList);
+    auto& parentInfo = _ownerInfo->resolve(liveRealm);
+    return translateRLMResultsErrors([&] {
+        return [[self.class alloc] initWithList: tsr.resolve<realm::List>(liveRealm->_realm)
```
jsflax, line 528: Formatting (no space)

RealmSwift/RealmCollection.swift
```diff
@@ -501,6 +501,18 @@ public protocol RealmCollection: RealmCollectionBase, _RealmCollectionEnumerator
      `Realm.Configuration.maximumNumberOfActiveVersions` for more information.
     */
     func freeze() -> Self
+
+    /**
+     Returns a live, mutable version of this frozen collection.
+
+     This method resolves a reference to a live copy of the same frozen collection.
+     If a called on a non-frozen collection, will return itself.
```
jsflax, line 509: Weird grammar

RealmSwift/Tests/RealmCollectionTypeTests.swift
```diff
@@ -522,6 +522,13 @@ class RealmCollectionTypeTests: TestCase {
         XCTAssertTrue(collection.freeze().isFrozen)
     }
 
+    func testThaw() {
+        let frozen = collection.freeze()
+        XCTAssertTrue(frozen.isFrozen)
+        let thawed = frozen.thaw()
+        XCTAssertFalse(thawed.isFrozen)
```
jsflax, line 529: This test should assert that that frozen collection cannot be mutating and that the thawed collection can.

jsflax (commented):
Missing a bunch of tests (needs test for objc, each collection ideally).

Also, this is missing thaw for Objects, which is the titular feature.

Realm/RLMArray.h
tgoyne: I think we probably just want to call non-frozen collections "live collections"?

The comma is weird too.

Realm/RLMArray.h
```diff
@@ -492,6 +492,18 @@ __attribute__((warn_unused_result));
...
+
+ @warning Holding onto a thawed collection for an extended period while performing write
```
tgoyne, line 501: Incorrect copy/paste.

Realm/RLMManagedArray.mm
```diff
@@ -505,14 +506,31 @@ - (instancetype)freeze {
...
 
+- (instancetype)thaw {
+    if (!self.frozen) {
+        return self;
+    }
+
+    RLMRealmConfiguration *rlmConfig = _realm.configuration;
+    RLMRealm *liveRealm = [RLMRealm realmWithConfiguration:rlmConfig error:nil];
+
+    auto tsr = realm::ThreadSafeReference(_backingList);
```
tgoyne, line 525: This doesn't need to use realm::ThreadSafeReference; it can just call `_backingList.freeze(liveRealm->_realm)` to do the same thing (and then at that point freeze and thaw can be extracted to a single function which takes the target Realm as a parameter).

Realm/RLMArray.h
ericjordanmossman: Removed in later commit.
What makes it incorrect, though? Is it because the thread safe references is no longer a liability once resolved?

tgoyne: A thawed collection is just a normal live collection and this warning is about frozen collections.

ericjordanmossman: Got it, thanks. I misunderstood the nature of what made frozen collections dangerous to hold onto, then.

Realm/RLMObjectBase.mm
```diff
@@ -488,6 +488,18 @@ id RLMObjectFreeze(RLMObjectBase *obj) {
     return frozen;
 }
 
+id RLMObjectThaw(RLMObjectBase *obj) {
+    RLMVerifyAttached(obj);
```
ericjordanmossman, line 492: There is no check for whether this is object is managed because it couldn't be unmanaged if it were previously frozen. If it were not frozen, then the method returns.

Realm/Tests/ObjectTests.m
```diff
@@ -1469,4 +1472,18 @@ - (void)testFreezeInsideWriteTransaction {
...
+- (void)testThaw {
+    IntObject *frozen = [managedObject() freeze];
+    XCTAssertTrue([frozen isFrozen]);
+
+    IntObject *live = [frozen thaw];
+    XCTAssertFalse([live isFrozen]);
+
+    RLMRealm *liveRealm = live.realm;
+    [liveRealm beginWriteTransaction];
+    XCTAssert(live.intCol = 1);
```
tgoyne, line 1484: Putting XCTAssert here doesn't really make any sense.

Realm/Tests/ObjectTests.m
```diff
@@ -1469,4 +1472,18 @@ - (void)testFreezeInsideWriteTransaction {
...
+    [liveRealm commitWriteTransaction];
+    XCTAssert(live.intCol != frozen.intCol);
```
tgoyne, line 1486: Use XCTAssertEqual so that it prints the values if it fails.

Realm/Tests/RealmTests.mm
```diff
@@ -1858,10 +1858,40 @@ - (void)testIsFrozen {
...
+    XCTAssertTrue(frozenObj.frozen);
+
+    RLMRealm *thawedRealm = [realm thaw];
+    XCTAssertFalse(thawedRealm.frozen);
+    IntObject *thawedObj = [[IntObject objectsInRealm:thawedRealm where:@"intCol = 1"] firstObject];
+
+    [realm beginWriteTransaction];
+    thawedObj.intCol = 2;
+    [realm commitWriteTransaction];
+    XCTAssertFalse(thawedObj.intCol == frozenObj.intCol);
```
tgoyne, line 1889: XCTAssertNotEqual

RealmSwift/Tests/ObjectTests.swift
```diff
@@ -1002,4 +1000,24 @@ class ObjectTests: TestCase {
...
+        let frozenObj = obj.freeze()
+        XCTAssertTrue(frozenObj.isFrozen)
+        assertThrows(try! frozenObj.realm!.write { frozenObj.realm!.delete(frozenObj) }, reason: "Can't perform transactions on a frozen Realm")
+
+        let liveObj = frozenObj.thaw()
+        XCTAssertFalse(liveObj.isFrozen)
+        XCTAssertEqual(liveObj.boolCol, frozenObj.boolCol)
+
+        try! liveObj.realm!.write({ liveObj.boolCol = false })
+        XCTAssert(liveObj.boolCol != frozenObj.boolCol)
```
tgoyne, line 1021: XCTAssertEqual

RealmSwift/Tests/RealmCollectionTypeTests.swift
```diff
@@ -522,6 +522,22 @@ class RealmCollectionTypeTests: TestCase {
         XCTAssertTrue(collection.freeze().isFrozen)
     }
 
+    func testThaw() {
+        let frozen = collection.freeze()
+        XCTAssertTrue(frozen.isFrozen)
+
+        let frozenRealm = frozen.realm!
+        assertThrows(try! frozenRealm.write { frozenRealm.delete(frozen) }, reason: "Can't perform transactions on a frozen Realm")
```
tgoyne, line 530: The write shouldn't have a body as it's not supposed to be reachable.

tgoyne (commented):
Some other things to test:

1. Thaw on a different thread than it was frozen on.
2. Thaw on a thread which already has a Realm should use that realm instance
3. Thaw on a thread which doesn't already have a Realm instance
4. Thaw at an older version than the object/list/results was created at.
5. That at a version where the object or list's parent object has been deleted

Realm/Tests/RealmTests.mm
```diff
@@ -1907,6 +1909,68 @@ - (void)testReadAfterInvalidateFrozen {
...
+    IntObject *nilObj = [[IntObject objectsInRealm:thawedRealm where:@"intCol = 1"] firstObject];
+    XCTAssertNil(nilObj);
+}
+
+- (void)testThawDifferentThread {
+    RLMRealm *frozenRealm = [[RLMRealm defaultRealm] freeze];
+    XCTAssertTrue(frozenRealm.frozen);
+    
+    // Thaw on a thread which already has a Realm should use existing reference.
+    [self dispatchAsyncAndWait:^{
```
ericjordanmossman, line 1943: @tgoyne 

> Thaw on a thread which already has a Realm should use that realm instance
> Thaw on a thread which doesn't already have a Realm instance

Do these tests demonstrate what you mean? ie. thaw isn't creating a bunch of references when one already exists?
Screen Shot 2020-12-17 at 6 23 28 PM

RealmSwift/Tests/RealmCollectionTypeTests.swift
```diff
@@ -522,6 +522,66 @@ class RealmCollectionTypeTests: TestCase {
...
+        try! realm.write({ realm.delete(collection) })
+        XCTAssertNotEqual(frozen.count, collection.count, "Frozen collections should not change")
+        
+        let live = frozen.thaw()
+        XCTAssertTrue(live.isEmpty, "Thawed collection should reflect transactions since the original reference was frozen")
+        XCTAssertFalse(frozen.isEmpty)
+        XCTAssertEqual(live.count, self.collection.count)
+    }
+    
+    func testThawDeletedParent() {
```
ericjordanmossman, line 571: @tgoyne 
>That at a version where the object or list's parent object has been deleted

A couple questions on this:
This is technically using the Object.thaw inside of the Collection tests. But the collection tests are already well suited to running through every collection type. Should this be in Object tests?

When thawing an object whose parent link was deleted creates an invalid object. What _should_ a developer see if they hit this case? Nil? An exception?

Screen Shot 2020-12-17 at 6 16 26 PM

tgoyne: thaw() should probably return nil if the object no longer exists.

RealmSwift/Tests/RealmCollectionTypeTests.swift
```diff
@@ -522,6 +522,66 @@ class RealmCollectionTypeTests: TestCase {
...
+            XCTAssertFalse(live.isFrozen)
+
+            let liveRealm = live.realm!
+            try! liveRealm.write { liveRealm.delete(live) }
+            XCTAssertTrue(live.isEmpty)
+            XCTAssertFalse(frozen.isEmpty)
+        }
+    }
+    
+    func testThawPreviousVersion() {
```
tgoyne, line 556: This is testing a thing that should be tested, but wasn't quite what I was referring to. An interesting scenario to verify is

1. Open realm on main thread and create object.
2. On background thread mutate object then freeze it
3. Thaw the frozen object on the main thread and verify that it has the value from before the write (as the main thread hasn't refreshed yet)

A similar scenario would be instead creating the object initially on the background thread so that when it's thawed on the main thread it won't exist. The most likely to be problematic case for that is a Results which filters a List which doesn't exist yet.

RealmSwift/Tests/RealmCollectionTypeTests.swift
ericjordanmossman: This would mean that thaw has to be tried, right?

jsflax: Why wouldn't this be true for frozen as well?

tgoyne: freeze() always returns an object at the same version as the source object, and you can't call it on an object that doesn't exist. thaw() may give you an object at a different version if there's already a Realm open on the current thread.

tgoyne (commented):
Looks reasonably complete and tested now. Just needs a changelog entry and a core release.

ericjordanmossman:
> Looks reasonably complete and tested now. Just needs a changelog entry and a core release.

So could this be merged before it passes CI with this: https://github.com/realm/realm-core/pull/4254?

jsflax:
No, core release needs to come first.

jsflax (changes requested):
One noticeably missing thing is that `thaw` should be a part of the `ThreadConfined` protocol.

jsflax:
Also I find it slightly awkward that Objects and EmebeddedObjects can return nil but collections cannot.

ericjordanmossman:
@tgoyne I was trying to add the method `public func thaw<T: ThreadConfined>(_ obj: T) -> T?` in Realm.swift with [this](https://github.com/realm/realm-cocoa/pull/6975/files#diff-ba5328248b5679c4493ce8df152d03e275316dd65aa63dc75641f75531a8f49fR846). 
But this would completely bypass any other Objc thaw methods. So I started experimenting with some routing through [RLM](https://github.com/realm/realm-cocoa/pull/6975/files#diff-9c30eeb472947a4fd4e2a3499a123161201f6331b7b7479721fd43f38b4501f8R66)/[ThreadSafeReference](https://github.com/realm/realm-cocoa/pull/6975/files#diff-9c30eeb472947a4fd4e2a3499a123161201f6331b7b7479721fd43f38b4501f8R66).

However in this warning is a pretty big issue. And it became clear I don't really understand how I can successfully get ThreadConfined return value. What is a better direction to go in?
Screen Shot 2021-01-15 at 2 12 32 PM

jsflax (approved):
Nice.

ericjordanmossman:
@jsflax To answer your question - the required core change is slated for [next release](https://github.com/realm/realm-core/blob/master/CHANGELOG.md#fixed).

jsflax:
@ericjordanmossman it appears this has been released.

ericjordanmossman:
**What is the correct approach here?** @tgoyne @jsflax 
I retried the same code from  job [7483](https://ci.realm.io/job/objc_pr/7483/) (first pic) because swiftpm 12.1 and 12.3 had not completed. [7503](https://ci.realm.io/job/objc_pr/7503/) (second) swiftpm 12.1 and 12.3 passed, but 11.3 and 11.7 flaked.
Screen Shot 2021-01-22 at 2 31 01 PM
Screen Shot 2021-01-22 at 2 31 48 PM


- The swiftlint failure is due to the playground file formatting. 
- Cocoapods failure I attributed to ongoing cocoapods bug. 
- swiftpm-thread fails at `asyncOpenTimeout`. 
I'm unaware of a way to rerun just a single part of the matrix.

ericjordanmossman:
@tgoyne 
I believe I've resolved all concerns you raised. I'm interpreting the inconsistent CI failures as not caused by this PR.
I've just merged the latest master. Unless you have any objection, I'll merge this branch into master at 2pm eastern to unblock the SwiftUI improvements.

RealmSwift/List.swift
```diff
@@ -432,6 +432,10 @@ public final class List<Element: RealmCollectionValue>: ListBase {
         return List(objc: _rlmArray.freeze())
     }
 
+    public func thaw() -> List? {
```
tgoyne, line 435: Missing doc comment

CHANGELOG.md
```diff
@@ -1,6 +1,12 @@
 x.y.z Release notes (yyyy-MM-dd)
 =============================================================
 ### Enhancements
+* Add support for "thawing" objects. `Realm`, `Results`, `List` and `Object` now
+have `thaw()` methods which return a live copy of the frozen object. This enables app behvaior where a frozen object can be made live again in order to mutate values. For example, first freezing an object passed into UI view, then thawing the object in the view to update values.
+
+### Fixed
+* <How to hit and notice issue? what was the impact?> ([#????](https://github.com/realm/realm-cocoa/issues/????), since v?.?.?)
+* None.
```
tgoyne, line 9: Changelog merge was incorrect.

RealmSwift/List.swift
ericjordanmossman: The doc comment for List is pulled from [AnyRealmCollection](https://github.com/realm/realm-cocoa/pull/6975/files#diff-53ffa4c09fd61e70eef1403c51e39baa5ac2145873029bd5c9916f302a970852R1099). This is symmetric with freeze.
Screen Shot 2021-01-26 at 12 12 18 PM

In an editor, appears like so:
Screen Shot 2021-01-26 at 12 11 43 PM
Screen Shot 2021-01-26 at 12 11 50 PM

CHANGELOG.md
ericjordanmossman: Removed empty ###fixed and incorrect internal body.

ericjordanmossman:
FYI I cancelled the triggered build after the changelog commit. [7520](https://ci.realm.io/job/objc_pr/7520/) was still running and almost finished. I'm not sure if 7520 is reflected in the github UI, but it's the latest CI build with the most recent master.

tgoyne (approved):

PR #6992, opened 2020-12-12
Update RealmSwift.podspec: specify a Swift version '5.0'

duyhungtnn:
# What
- This PR to fix cocaopods error when integrate with new swift project.

# Error 
[!] Unable to determine Swift version for the following pods:

- `RealmSwift` does not specify a Swift version and none of the targets (`dep`) integrating it have the `SWIFT_VERSION` attribute set. Please contact the author or set the `SWIFT_VERSION` attribute in at least one of the targets that integrate this pod.

# Step to reproduce 

- Create new swift project 
- Add realmcocoa to podfile 
- Run pod install
- We will get the error.

# Changes 

- [x] Update RealmSwift.podspec: specify a Swift version '5.0'

### DIFF
--- a/RealmSwift.podspec
+++ b/RealmSwift.podspec
@@ -17,7 +17,7 @@ Pod::Spec.new do |s|
   has_versioned_docs          = !(version =~ /alpha|beta|rc/)
   s.documentation_url         = "https://realm.io/docs/swift/#{has_versioned_docs ? s.version : 'latest'}"
   s.license                   = { :type => 'Apache 2.0', :file => 'LICENSE' }
-
+  s.swift_version    = '5.0'
   s.dependency 'Realm', "= #{s.version}"
   s.source_files = 'RealmSwift/*.swift'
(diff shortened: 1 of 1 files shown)


tgoyne:
We do not set a swift version because we support building as whatever version the app consuming us is using. This error only occurs if the target depending on RealmSwift also doesn't have a swift version set.

duyhungtnn:
> We do not set a swift version because we support building as whatever version the app consuming us is using. This error only occurs if the target depending on RealmSwift also doesn't have a swift version set.

Thank you. 

# For any dev get this error and go here from google search :)  , here is how to bypass this error : 

In podfile for new project , need set `target` like this and run `pod install` 
`
# Uncomment this line to define a global platform for your project
platform :ios, '9.0'

target 'realm_new_app' do
    use_frameworks!
    pod 'RealmSwift'
end
`

if no target is set , this error will occur

PR #7005, opened 2020-12-16
Refactor parts of RLMQueryUtil to reduce template instantiations

tgoyne:
The main change here is to transform queries which have a constant value on the LHS and a keypath on the RHS into one with the keypath on the RHS when possible. This eliminates a large number of redundant template instantiations, reducing the library size a bit and improving the build time. It also happens to make some queries which previous threw an exception work.

C++17 lets us replace some of the template metaprogramming with `if constexpr` making things a bit simpler.

Exception throwing has been shuffled around to reduce the size of the exception handler tables, and a few things have been refactored to cut down on unneeded duplication due to tmeplating.

In total this bumps about 100 kB off the macOS release binary and is a small net reduction in the amount of non-test code.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@ x.y.z Release notes (yyyy-MM-dd)
 =============================================================
 ### Enhancements
 * Add Xcode 12.3 binary to release package.
+* Add support for queries which have nil on the left side and a keypath on the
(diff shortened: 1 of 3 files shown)


tgoyne:
Test failures are unrelated; https://github.com/realm/realm-core/pull/4234 fixed the one new one.

jsflax (approved):

PR #7038, opened 2021-01-11
Refactor Response model to allow for separate Error field or model

pavel-ship-it:
Refactor Response model to allow for separate Error field or model


jsflax:
@pavel-ship-it can you also make the description the title of the PR? We need to keep our formatting of pull requests unified.

jsflax:
@pavel-ship-it should we close this considering the alternate route we took?

PR #7047, opened 2021-01-15
Remove weak declaration from user property of Functions struct

### LINKED ISSUE #6959: Functions API does not work with SwiftUI+Combine
When using the Functions API in the SDK the call to the function on the backend fails to reach the server. Looking into this it seems that the `weak User` property on the `Functions` struct is nil by the time the call is to be performed.

DominicFrei:
Fixes #6959


leemaguire:
Why is still in draft? Has it been tested that the removal of weak user won't result in a memory leak?

DominicFrei:
@leemaguire I had it in draft because I was still talking with @jsflax about the failing checks that I was not sure about. Problem solved, hence the assign. I did forget I had it still in draft though. Thanks for pointing out.
It's all tested and works fine on my side.

RealmSwift/Sync.swift
```diff
@@ -268,7 +268,7 @@ import Combine
 /// This handler is executed on a non-main global `DispatchQueue`.
 @dynamicMemberLookup
 @frozen public struct Functions {
-    weak var user: User?
+    var user: User?
```
jsflax, line 271: This should now be `private let user: User`. No reason for the user to be nullable if it's not weak.

jsflax:
I would also change the name of this PR to "Remove weak declaration from user property of Functions struct"

RealmSwift/Sync.swift
DominicFrei: Done.

DominicFrei:
> I would also change the name of this PR to "Remove weak declaration from user property of Functions struct"

@jsflax  Done.

@leemaguire I profiled and tested for leaks with an example app. Looks good on my side.

leemaguire (approved):
A changelog entry with details of the fix is needed.

DominicFrei:
> A changelog entry with details of the fix is needed.

Done.

CHANGELOG.md
```diff
@@ -22,6 +22,8 @@
 * Fast-enumerating a List after deleting the parent object would crash with an
   assertion failure rather than a more appropriate exception.
   ([Core #4114](https://github.com/realm/realm-core/issues/4114), since v5.0.0).
+* Fix an issue with the user in Functions which can be nil if no strong reference
```
leemaguire, line 25:
```suggestion
* Fix an issue where calling a MongoDB Realm Function would never be performed as the reference to the weak `User` was lost.
```

CHANGELOG.md
```diff
@@ -22,6 +22,8 @@
 * Fast-enumerating a List after deleting the parent object would crash with an
   assertion failure rather than a more appropriate exception.
   ([Core #4114](https://github.com/realm/realm-core/issues/4114), since v5.0.0).
+* Fix an issue where calling a MongoDB Realm Function would never be performed as the reference to the weak `User` was lost.
+  is held due to it being weak.
```
leemaguire, line 27: extra text here

leemaguire (approved):

jsflax (approved):

PR #7050, opened 2021-01-16
Change setup_baas.rb to Swift; Fix testAsyncOpenTimeout

jsflax:


Realm/ObjectServerTests/RealmServer.swift
```diff
@@ -49,6 +55,282 @@ extension URLSession {
...
+        process.environment = environment
+        createProcesses(commandString, process: &process, processes: &processes)
+
+        processes.forEach {
+            $0.launch()
+        }
+        processes.forEach {
+            $0.waitUntilExit()
+        }
+        if let _ = (processes.last!.standardOutput as? Pipe)?.fileHandleForReading.readDataToEndOfFile() {
```
leemaguire, line 166: Is there a reason this is written twice, and has an unused block?

jsflax: The unused block is some leftover cruft, will remove

Realm/ObjectServerTests/RealmServer.swift
```diff
@@ -49,6 +55,282 @@ extension URLSession {
...
+
+        let stitchWorktree = goRoot
+            .appendingPathComponent("src")
+            .appendingPathComponent("github.com")
+            .appendingPathComponent("10gen")
+            .appendingPathComponent("stitch")
+
+
+        if !FileManager.default.fileExists(atPath: stitchDir.appendingPathComponent(".git").absoluteString) {
+
```
leemaguire, line 233: extra newline

pavel-ship-it (approved):
LGTM
less ruby for best

Realm/ObjectServerTests/RealmServer.swift
```diff
@@ -736,7 +1059,19 @@ public class RealmServer: NSObject {
...
+        ]
+
+        let rules = app.services[serviceId].rules
+
+        rules.post(on: group, dogRule, failOnError)
+        rules.post(on: group, personRule, failOnError)
+        rules.post(on: group, hugeSyncObjectRule, failOnError)
+        // When running ObjcObjectServerTests,
+        // we do not want to pull in swift schema reqs
+        #if SWIFT_PACKAGE && REALM_HAVE_COMBINE
```
leemaguire, line 1071: Wont using these macros mean this class will cause issues when running tests from the xcproj?

leemaguire (approved):
LGTM, one comment

Realm/ObjectServerTests/RealmServer.swift
```diff
@@ -20,6 +20,12 @@ import Foundation
 import RealmSwift
 import XCTest
 
+private extension URL {
```
tgoyne, line 23: This looks like an incorrect workaround for the source URL being constructed incorrectly.

Realm/ObjectServerTests/RealmServer.swift
```diff
@@ -285,42 +375,84 @@ public class RealmServer: NSObject {
...
-    /// Log level for the server and mongo processes.
-    public var logLevel = LogLevel.none
-
-    /// Process that runs the local mongo server. Should be terminated on exit.
-    private let mongoProcess = Process()
-    /// Process that runs the local backend server. Should be terminated on exit.
-    private let serverProcess = Process()
+        do {
+            return try JSONDecoder().decode(Dependencies.self,
+                                            from: JSONEncoder().encode(dependenciesString.components(separatedBy: "\n").dropLast().reduce(into: [String: String]()) {
```
tgoyne, line 401: This line really needs to be split into something more reasonable.

Realm/ObjectServerTests/RealmServer.swift
```diff
@@ -341,11 +474,187 @@ public class RealmServer: NSObject {
...
+            shutdownProcess.waitUntilExit()
+            mongoProcess.terminate()
+        }
+    }
+
+    private func setupStitch() throws {
+        print("Setting up stitch")
+
+        let goRoot = buildDir.appendingPathComponent("go")
+        if try fileManager.contentsOfDirectory(atPath: binDir.absoluteString).count == 7 {
```
tgoyne, line 514: This is a really brittle way to do this.

Realm/ObjectServerTests/SwiftObjectServerTests.swift
```diff
@@ -27,24 +27,10 @@ import RealmSyncTestSupport
...
-        let fakeDataSize = 1000000
-        return SwiftHugeSyncObject(value: ["data": Data(repeating: 16, count: fakeDataSize)])
-    }
-}
-
 extension User {
     func configuration(testName: String) -> Realm.Configuration {
         var config = self.configuration(partitionValue: testName)
-        config.objectTypes = [SwiftPerson.self, SwiftHugeSyncObject.self]
+        config.objectTypes = [SwiftPerson.self, Person.self, Dog.self, HugeSyncObject.self, SwiftHugeSyncObject.self]
```
tgoyne, line 33: These tests shouldn't need the obj-c types.

Realm/ObjectServerTests/RealmServer.swift
```diff
@@ -341,11 +474,187 @@ public class RealmServer: NSObject {
...
+            print("server binary built")
+        }
+
+        subprocess.popen("/bin/cp stitch/etc/configs/test_config.json ./")
+    }
+
+    private func buildServer() throws {
+        try fileManager.createDirectory(at: tempDir, withIntermediateDirectories: true)
+        if !fileManager.fileExists(atPath: buildDir.absoluteString) {
+            try! FileManager.default.createDirectory(at: URL.init(fileURLWithPath: buildDir.absoluteString,
```
tgoyne, line 635: Redundant .init

RealmSwift/Tests/CombineTests.swift
```diff
@@ -842,14 +842,16 @@ private class CombineCollectionPublisherTests<Collection: RealmCollection>: Comb
...
         cancellable = collection
             .changesetPublisher
             .saveToken(on: self, at: \.notificationToken)
             .sink { _ in
                 calls += 1
+                ex.fulfill()
             }
         XCTAssertNotNil(notificationToken)
-
+        waitForExpectations(timeout: 2.0, handler: nil)
```
tgoyne, line 854: This wait shouldn't be needed and if it does fix the test then it's just working around a bug which the test was exposing.

jsflax: The test always passes for me locally, this was to test on the CI.

Why isn't this wait needed though? Isn't the initial `sink` value still called asynchronously?

tgoyne: I think maybe rather than this wait being unnecessary, there's actually two different things that should be tested. Adding a notifier and then immediately skipping it before receiving the initial notification is something that should work. This is what the test was previously testing on CI, but for that it needs a wait *after* the writes because it's possible for all of them to happen before initial is ever delivered. Adding the wait here validates that it successfully skips after receiving the initial notification, which of course it also something that should be tested.

So basically the test is correct, but there should be another copy of this test added with the wait after the loop but before the assert on call count.

Realm/ObjectServerTests/RLMSyncTestCase.mm
```diff
@@ -451,6 +452,7 @@ - (void)setUp {
     [NSFileManager.defaultManager removeItemAtURL:self.clientDataRoot error:nil];
     [NSFileManager.defaultManager createDirectoryAtURL:self.clientDataRoot
                            withIntermediateDirectories:YES attributes:nil error:nil];
+//    [[RealmServer shared] dropTestData];
```
tgoyne, line 455: commented out code

Realm/ObjectServerTests/RealmServer.swift
```diff
@@ -684,6 +989,34 @@ public class RealmServer: NSObject {
...
 
+        let swiftHugeSyncObjectRule: [String: Any] = [
+            "database": "test_data",
+            "collection": "SwiftHugeSyncObject",
+            "roles": [[
+                "name": "default",
+                        "apply_when": [:],
+                "insert": true,
+                "delete": true,
+                        "additional_fields": [:]
```
tgoyne, line 1000: indentation is weird

build.sh
```diff
@@ -1106,8 +1101,7 @@ case "$COMMAND" in
 
             if [[ "$target" = *"server"* ]] || [[ "$target" = "swiftpm"* ]]; then
                 source "$(brew --prefix nvm)/nvm.sh" --no-use
-                nvm install 8.11.2
-                sh build.sh setup-baas
+                nvm install 14.15.4
```
tgoyne, line 1104: Grab node version from dependencies.list?

bmunkholm:
@jsflax Is this still relevant?

jsflax:
We could probably still use this but it was never necessary.

PR #7051, opened 2021-01-16
Add MutableSet<> / RLMSet datatype

leemaguire:
Adds implementation for MutableSet<> / RLMSet datatype. These are based on the respective NSSet and SetAPI's.

I'm pointing to a specific core version in the package.swift as `realm::SchemaMode::Additive` has been renamed to `realm::SchemaMode::AdditiveDiscovered` and it needs a core release with the xcframework


Realm/RLMAccessor.mm
```diff
@@ -170,17 +179,31 @@ void setValue(__unsafe_unretained RLMObjectBase *const obj, ColKey key,
...
               __unsafe_unretained id<NSFastEnumeration> const value) {
     auto prop = obj->_info->propertyForTableColumn(key);
-    RLMValidateValueForProperty(value, obj->_info->rlmObjectSchema, prop, true);
-
-    realm::List list(obj->_realm->_realm, obj->_row, key);
-    RLMClassInfo *info = obj->_info;
-    if (list.get_type() == realm::PropertyType::Object) {
-        info = &obj->_info->linkTargetType(prop.index);
+    if (prop.array) {
+        RLMValidateValueForProperty(value, obj->_info->rlmObjectSchema, prop, true);
```
tgoyne, line 183: Looks like about half these lines could be pulled out of the if.

Realm/RLMAccessor.mm
```diff
@@ -477,14 +514,27 @@ id unmanagedSetter(RLMProperty *prop, const char *) {
...
 
-        // make copy when setting (as is the case for all other variants)
-        RLMArray *ar;
-        if (prop.type == RLMPropertyTypeObject)
-            ar = [[RLMArray alloc] initWithObjectClassName:prop.objectClassName];
-        else
-            ar = [[RLMArray alloc] initWithObjectType:prop.type optional:prop.optional];
-        [ar addObjects:values];
-        superSet(obj, propName, ar);
+        if (prop.array) {
```
tgoyne, line 517: This also could pull the addObjects and superSet() calls out of the if.

Realm/RLMAccessor.mm
```diff
@@ -731,6 +783,18 @@ id RLMDynamicGetByName(__unsafe_unretained RLMObjectBase *const obj,
...
 
+id RLMAccessorContext::box(realm::object_store::Set&& s) {
+    REALM_ASSERT(_parentObjectInfo);
+    REALM_ASSERT(currentProperty);
+    return [[RLMManagedSet alloc] initWithSet:std::move(s)
+                                   parentInfo:_parentObjectInfo
+                                     property:currentProperty];
+}
+
+id RLMAccessorContext::box(realm::object_store::Dictionary&& d) {
```
tgoyne, line 794:
```suggestion
id RLMAccessorContext::box(realm::object_store::Dictionary&&) {
```

Realm/RLMAccessor.mm
```diff
@@ -975,6 +1035,13 @@ static auto to_optional(__unsafe_unretained id const value, Fn&& fn) {
     return [v respondsToSelector:@selector(isBackedByList:)] && [v isBackedByList:list];
 }
 
+bool RLMAccessorContext::is_same_set(realm::object_store::Set const& set, __unsafe_unretained id const v) const noexcept {
+    return [v respondsToSelector:@selector(isBackedBySet:)] && [v isBackedBySet:set];
+}
+
+bool RLMAccessorContext::is_same_dictionary(realm::object_store::Dictionary const& dictionary, __unsafe_unretained id const v) const noexcept {
```
tgoyne, line 1042:
```suggestion
bool RLMAccessorContext::is_same_dictionary(realm::object_store::Dictionary const&, __unsafe_unretained id const) const noexcept {
```

Realm/RLMArray.mm
```diff
@@ -27,6 +27,7 @@
 #import "RLMSwiftSupport.h"
 #import "RLMThreadSafeReference_Private.hpp"
 #import "RLMUtil.hpp"
+#import "RLMConstants.h"
```
tgoyne, line 30: Header imports should be sorted (probably applies to other files as well but not going to mark them).

Realm/RLMManagedSet.mm
```diff
@@ -0,0 +1,564 @@
...
+    if (!set.realm) {
+        @throw RLMException(@"Right hand side value must be a managed Set.");
+    }
+
+    if (_type != set.type) {
+        @throw RLMException(@"Set must match type of \"self\" '%@'", RLMTypeToString(_type));
+    }
+
+    return static_cast<RLMManagedSet *>(set);
+}
```
tgoyne, line 298:
```suggestion
- (RLMManagedSet *)managedObjectFrom:(RLMSet *)set {
    auto managedSet = RLMDynamicCast<RLMSet>(set);
    if (!managedSet) {
        @throw RLMException(@"Right hand side value must be a managed Set.");
    }
    if (_type != managedSet->_type || _objectInfo != managedSet->_objectInfo) {
        @throw RLMException(@"Set must match type of \"self\" '%@'", RLMTypeToString(_type));
    }
    return managedSet;
}
```

seems like a more correct implementation of this function, but not allowing an unmanaged set on the RHS is odd here.

Realm/RLMManagedSet.mm
```diff
@@ -0,0 +1,564 @@
...
+- (BOOL)intersectsSet:(RLMSet<id> *)set {
+    RLMManagedSet *rhs = [self managedObjectFrom:set];
+    return _backingSet.intersects(rhs->_backingSet);
+}
+
+- (BOOL)containsObject:(id)obj {
+    RLMSetValidateMatchingObjectType(self, obj);
+    RLMAccessorContext context(*_objectInfo);
+    auto r = _backingSet.find(context, obj);
+    return (r >= 0) && (r != realm::npos);
```
tgoyne, line 314: Find returns a size_t, which can't be less than zero.

Realm/RLMObject.h
```diff
@@ -709,4 +709,17 @@ typedef void (^RLMObjectChangeBlock)(BOOL deleted,
...
+/**
+ Properties on `RLMObject`s of type `RLMSet` must have an associated type. A type is associated
+ with an `RLMSet` property by defining a protocol for the object type that the array should contain.
+ To define the protocol for an object, you can use the macro RLM_SET_TYPE:
+
+     RLM_SET_TYPE(ObjectType)
+     ...
+     @property RLMSet<ObjectType *><ObjectType> *setOfObjectTypes;
+  */
+#define RLM_SET_TYPE(RLM_OBJECT_SUBCLASS)\
```
tgoyne, line 721: We don't really need a separate macro for this. Maybe it should just be RLM_COLLECTION_TYPE with RLM_ARRAY_TYPE marked as deprecated?

Realm/RLMObjectBase.mm
```diff
@@ -208,6 +210,16 @@ - (void)setValue:(id)value forUndefinedKey:(NSString *)key {
                                                              RLMSchema.partialPrivateSharedSchema)];
             }
         }
+        else if (property.set) {
```
tgoyne, line 213: This could be shared between array and set pretty easily. Perhaps rename both _rlmArray and _rlmSet to _rlmCollection so that the same code works on both?

Realm/RLMObservation.mm
```diff
@@ -236,6 +244,18 @@ auto reverse(Container const& c) {
         return value;
     }
 
+    if (lastProp.set) {
```
tgoyne, line 247: This is the exact same code as the array case and could just be combined with that.

Realm/RLMUtil.mm
```diff
@@ -136,19 +151,40 @@ bool RLMIsSwiftObjectClass(Class cls) {
...
+            }
+            return YES;
+        }
+        if (!value || value == NSNull.null) {
+            return YES;
+        }
+        return NO;
+    }
+
+    if (collectionType == RLMCollectionTypeSet) {
```
tgoyne, line 180: There doesn't appear to be any reason to distinguish arrays and sets here?

tgoyne (commented):
I think all of the newly added .gitkeep files are incorrect and are marking directories that no longer exist.

I took a first pass of skimming over this and it looks broadly correct. There's... a lot to review though.

Realm/RLMObject.h
leemaguire: I've deprecated `RLM_ARRAY_TYPE ` and updated usage of it across the examples and sdk source.

Realm/RLMObjectBase.mm
leemaguire: renaming both to `_rlmCollection ` resulted in a compiler error. I managed to reduce the code with 
```
id collection = property.array ?
                [object_getIvar(self, ivar) _rlmArray] : [object_getIvar(self, ivar) _rlmSet];
```

Realm/RLMUtil.mm
leemaguire: I've also removed `RLMCollectionTypeSet ` as this was the only use case for it

Realm/RLMManagedSet.mm
leemaguire: `managedObjectFrom` is used for methods that mutate and compare against the managed object, so I thought it was right to make a helper that could perform the cast.

jsflax:
Should pull master into new_data_types, as this PR has become pretty difficult to read with all the new changes

pavel-ship-it (approved):
LGTM. Great job

Realm/Realm.modulemap
```diff
@@ -8,8 +8,10 @@ framework module Realm {
         header "RLMAccessor.h"
         header "RLMApp_Private.h"
         header "RLMArray_Private.h"
+        header "RLMSet_Private.h"
         header "RLMCollection_Private.h"
         header "RLMListBase.h"
+        header "RLMSetBase.h"
```
jsflax, line 14: alphabetical order

Realm/Realm.h
```diff
@@ -19,6 +19,7 @@
 #import <Foundation/Foundation.h>
 
 #import <Realm/RLMArray.h>
+#import <Realm/RLMSet.h>
```
jsflax, line 22: alphabetical order, but I'm not going to mark every one

Realm/Tests/KVOTests.mm
```diff
@@ -389,11 +433,11 @@ - (void)testRegisterForUnknownProperty {
 - (void)testRemoveObserver {
     // iOS 14.2 beta 2 has stopped throwing exceptions when a KVO observer is removed that does not exist
     // FIXME: revisit this once 14.2 is out to see if this was an intended change
-#if REALM_PLATFORM_IOS
-    if (@available(iOS 14.2, *)) {
-        return;
-    }
-#endif
+    #if REALM_PLATFORM_IOS
+        if (@available(iOS 14.2, *)) {
```
jsflax, line 437: no indent for the actual conditional branches

Realm/Tests/KVOTests.mm
```diff
@@ -1182,6 +1303,72 @@ - (void)testPrimitiveArrayDiffs {
...
+    AssertSetChanged();
+    [mutator2 addObject:obj2.obj];
+    AssertSetChanged();
+    // This will emit 2 KVO notifications instead of one. It seems that
+    // 2 cascade notifications are being created at the core level.
+    // table::for_each_backlink_column looks suspicious as there is a
+    // fixme comment stating it should be optimised to not itterate through
+    // all backlinks.
+    // The expected count here will be 1, as the first notification will be popped
+//    [mutator setSet:mutator2];
```
jsflax, line 1331: should this still be commented ut or should it still work?

jsflax (commented):
First pass done, good so far. Could you pull in the thaw changes as well and conform to ThreadConfined thereafter?

Package.swift
```diff
@@ -62,7 +62,7 @@ let package = Package(
             targets: ["Realm", "RealmSwift"]),
     ],
     dependencies: [
-        .package(url: "https://github.com/realm/realm-core", .exact(Version(coreVersionStr)!))
+        .package(url: "https://github.com/realm/realm-core", .revision("b197995779c300f5af321e839fb35d49688bc575"))
```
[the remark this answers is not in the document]
leemaguire: can't point to master until the next core release, due to renaming of schema modes.

Realm/RLMAccessor.mm
```diff
@@ -160,29 +162,64 @@ void setValue(__unsafe_unretained RLMObjectBase *const obj, ColKey key,
...
 
-    realm::List list(obj->_realm->_realm, obj->_row, key);
-    RLMClassInfo *info = obj->_info;
-    if (list.get_type() == realm::PropertyType::Object) {
+template <typename Collection>
+void assign_value(Collection&& collection,
+                  RLMClassInfo *info,
+                  __unsafe_unretained RLMObjectBase *const obj,
+                  __unsafe_unretained RLMProperty *const prop,
+                  __unsafe_unretained id<NSFastEnumeration> const value) {
```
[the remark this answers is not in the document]
leemaguire: The RHS can be an array literal when setting an RLMArray. So RLMCollection doesn't really make sense.

Realm/RLMAccessor.mm
```diff
@@ -160,29 +162,64 @@ void setValue(__unsafe_unretained RLMObjectBase *const obj, ColKey key,
...
+        collection.assign(ctx, value, realm::CreatePolicy::ForceCreate);
     });
 }
 
+void setValue(__unsafe_unretained RLMObjectBase *const obj, ColKey key,
+              __unsafe_unretained id<NSFastEnumeration> const value) {
+    auto prop = obj->_info->propertyForTableColumn(key);
+    RLMValidateValueForProperty(value, obj->_info->rlmObjectSchema, prop, true);
+
+    if (prop.array) {
```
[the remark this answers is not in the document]
leemaguire: The compiler will complain about incompatible operand types

Realm/RLMAccessor.mm
```diff
@@ -467,24 +518,33 @@ id unmanagedGetter(RLMProperty *prop, const char *) {
...
-        // make copy when setting (as is the case for all other variants)
-        RLMArray *ar;
-        if (prop.type == RLMPropertyTypeObject)
-            ar = [[RLMArray alloc] initWithObjectClassName:prop.objectClassName];
-        else
-            ar = [[RLMArray alloc] initWithObjectType:prop.type optional:prop.optional];
-        [ar addObjects:values];
-        superSet(obj, propName, ar);
+        id collection;
+        if (prop.array) {
```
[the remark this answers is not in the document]
leemaguire: IMO that would hurt the readability. Unless something like this was done but it probably is another little thing that will increase compile time.

```
id collection = [prop](Class cls, bool object) {
            return object ? [[cls alloc] initWithObjectClassName:prop.objectClassName] :
                            [[cls alloc] initWithObjectType:prop.type optional:prop.optional];
        };
```

Realm/RLMAccessor.mm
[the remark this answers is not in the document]
jsflax: List and Set should probably just share this functionality in core, in which case you wouldn't need to template the method and could just use a base type

Realm/RLMAccessor.mm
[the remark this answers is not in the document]
jsflax:
```objc
        Class cls = prop.array ? RLMArray.class : RLMSet.class;
        // make copy when setting (as is the case for all other variants)
        if (prop.type == RLMPropertyTypeObject)
            collection = [[cls alloc] initWithObjectClassName:prop.objectClassName];
        else
            collection = [[cls alloc] initWithObjectType:prop.type optional:prop.optional];
```
If you still think it's not readable (I think it's fine), I'd consider that, since there is so much repeated code between the types, that maybe RLMCollection should own some of that.

Realm/RLMAccessor.mm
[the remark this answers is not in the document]
jsflax: You could make a function that returns `auto` for this:
```objc
auto CollectionFromObject(__unsafe_unretained RLMObjectBase *const obj, ColKey key) {
    auto prop = obj->_info->propertyForTableColumn(key);
    if (prop.array) {
        return realm::List(obj->_realm->_realm, obj->_row, key);
    } else {
        realm::object_store::Set(obj->_realm->_realm,
                                 obj->_row, key);
    }
}
```
but I think if this is the only place it would be used, then it's not worth it.

Realm/RLMAccessor.mm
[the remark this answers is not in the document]
jsflax: I would consider doing this for a number of the introduced conditionals. There are many `if (prop.array)`s with identical code between them, bar the instantiating type.

leemaguire:
@jsflax @tgoyne can you PTAL?

CHANGELOG.md
```diff
@@ -1,6 +1,8 @@
 x.y.z Release notes (yyyy-MM-dd)
 =============================================================
 ### Enhancements
+* Add `RLMSet`/`MutableSet<>` datatype. This is a Set collection type used for storing distinct values in a collection.
+
```
jsflax, line 5: Extra newline and the `None.` is incorrect below here.

Package.swift
```diff
@@ -73,13 +73,16 @@ let package = Package(
...
                 "Realm/RLMAnalytics.mm",
                 "Realm/RLMArray.mm",
+                "Realm/RLMSet.mm",
                 "Realm/RLMClassInfo.mm",
                 "Realm/RLMCollection.mm",
                 "Realm/RLMConstants.m",
                 "Realm/RLMDecimal128.mm",
                 "Realm/RLMEmbeddedObject.mm",
                 "Realm/RLMListBase.mm",
+                "Realm/RLMSetBase.mm",
```
jsflax, line 83: These still need to be ordered alphabetically.

Realm/RLMAccessor.mm
```diff
@@ -467,8 +491,8 @@ id unmanagedGetter(RLMProperty *prop, const char *) {
 }
 
 id unmanagedSetter(RLMProperty *prop, const char *) {
-    // Only RLMArray needs special handling for the unmanaged setter
-    if (!prop.array) {
+    // Only RLMArray & RLMSet needs special handling for the unmanaged setter
```
jsflax, line 494:
```suggestion
    // Only RLMArray & RLMSet need special handling for the unmanaged setter
```

Realm/RLMManagedSet.mm
```diff
@@ -0,0 +1,572 @@
...
+//
+- (RLMRealm *)realm {
+    return _realm;
+}
+
+- (NSUInteger)count {
+    return translateErrors([&] { return _backingSet.size(); });
+}
+
+- (NSArray<id> *)allObjects {
```
jsflax, line 202: Shouldn't this be an NSSet?

Realm/RLMObservation.hpp
```diff
@@ -21,6 +21,7 @@
 #import <realm/obj.hpp>
 #import <realm/object-store/binding_context.hpp>
 #import <realm/table.hpp>
+#import <Realm/RLMConstants.h>
```
jsflax, line 24: Why is this required?

Realm/RLMProperty.mm
```diff
@@ -19,11 +19,12 @@
 #import "RLMProperty_Private.hpp"
 
 #import "RLMArray_Private.hpp"
-#import "RLMListBase.h"
+//#import "RLMListBase.h"
```
jsflax, line 22: leftover comment

RealmSwift/MutableSet.swift
```diff
@@ -0,0 +1,548 @@
...
+    public var endIndex: Int { return count }
+
+    public func index(after i: Int) -> Int { return i + 1 }
+    public func index(before i: Int) -> Int { return i - 1 }
+
+    /// :nodoc:
+    // swiftlint:disable:next identifier_name
+    public func _observe(_ queue: DispatchQueue?,
+                         _ block: @escaping (RealmCollectionChange<AnyRealmCollection<Element>>) -> Void)
+        -> NotificationToken {
```
jsflax, line 459: no newline

Realm/RLMManagedSet.mm
leemaguire: This is mirroring the NSSet API which returns an NSArray of all items in the set.

RealmSwift/MutableSet.swift
leemaguire: `List` does the same here, I presume its to preserve some style relating the the amount of max characters per line.

CHANGELOG.md
```diff
@@ -10,6 +12,7 @@ x.y.z Release notes (yyyy-MM-dd)
 * Add Xcode 12.4 binaries to the release package.
 
 ### Fixed
+* <How to hit and notice issue? what was the impact?> ([#????](https://github.com/realm/realm-cocoa/issues/????), since v?.?.?)
```
tgoyne, line 15: bad merge

CHANGELOG.md
```diff
@@ -129,6 +132,7 @@ x.y.z Release notes (yyyy-MM-dd)
 * CocoaPods: 1.10 or later.
 
 ### Internal
+* Upgraded realm-core from v10.0.1 to v10.2.0
```
tgoyne, line 132: bad merge

Realm/RLMObservation.hpp
leemaguire: removed.

Realm/RLMObservation.mm
```diff
@@ -521,8 +529,12 @@ void RLMWillChange(std::vector<realm::BindingContext::ObserverState> const& obse
         NSMutableIndexSet *indexes = [NSMutableIndexSet new];
         for (auto const& o : observed) {
             forEach(o, [&](realm::ColKey colKey, auto const& change, RLMObservationInfo *info) {
-                info->willChange(info->columnName(colKey),
-                                 convert(change.kind), convert(change.indices, indexes));
+                if (colKey.is_set()) {
+                    info->willChange(info->columnName(colKey));
+                } else {
```
tgoyne, line 534: Newline before else in everything but swift code.

Realm/RLMObservation.mm
```diff
@@ -521,8 +529,12 @@ void RLMWillChange(std::vector<realm::BindingContext::ObserverState> const& obse
         NSMutableIndexSet *indexes = [NSMutableIndexSet new];
         for (auto const& o : observed) {
             forEach(o, [&](realm::ColKey colKey, auto const& change, RLMObservationInfo *info) {
-                info->willChange(info->columnName(colKey),
-                                 convert(change.kind), convert(change.indices, indexes));
+                if (colKey.is_set()) {
```
tgoyne, line 532: Is this conditional actually needed? It's just ensuring that the second parameter is `nil` even if `change.indices` isn't empty for a set, but if we aren't going to be using `change.indices` for sets then we shouldn't be populating it in the first place.

Realm/RLMObservation.mm
```diff
@@ -538,7 +550,11 @@ void RLMDidChange(std::vector<realm::BindingContext::ObserverState> const& obser
         NSMutableIndexSet *indexes = [NSMutableIndexSet new];
         for (auto const& o : reverse(observed)) {
             forEach(o, [&](realm::ColKey col, auto const& change, RLMObservationInfo *info) {
-                info->didChange(info->columnName(col), convert(change.kind), convert(change.indices, indexes));
+                if (col.is_set()) {
```
tgoyne, line 553: Similarly this shouldn't be needed.

Realm/RLMUtil.mm
```diff
@@ -108,11 +110,24 @@ static inline bool numberIsDouble(__unsafe_unretained NSNumber *const obj) {
...
+}
+
 static inline bool checkArrayType(__unsafe_unretained RLMArray *const array,
-                                  RLMPropertyType type, bool optional,
+                                  RLMPropertyType type,
+                                  bool optional,
                                   __unsafe_unretained NSString *const objectClassName) {
     return array.type == type && array.optional == optional
-        && (type != RLMPropertyTypeObject || [array.objectClassName isEqualToString:objectClassName]);
+    && (type != RLMPropertyTypeObject || [array.objectClassName isEqualToString:objectClassName]);
```
tgoyne, line 122: The previous indentation was correct.

Realm/RLMUtil.mm
```diff
@@ -108,11 +110,24 @@ static inline bool numberIsDouble(__unsafe_unretained NSNumber *const obj) {
...
+}
+
+static inline bool checkSetType(__unsafe_unretained RLMSet *const set,
```
tgoyne, line 125: This appears to just be the exact same function as checkArrayType?

Realm/RLMUtil.mm
```diff
@@ -136,14 +151,19 @@ bool RLMIsSwiftObjectClass(Class cls) {
...
                       __unsafe_unretained NSString *const objectClassName) {
     if (optional && !RLMCoerceToNil(value)) {
         return YES;
     }
-    if (array) {
+
+    if (collection) {
         if (auto rlmArray = asRLMArray(value)) {
             return checkArrayType(rlmArray, type, optional, objectClassName);
+        } else if (auto rlmSet = asRLMSet(value)) {
```
tgoyne, line 165: Newline before else.

Realm/Swift/RLMSupport.swift
```diff
@@ -101,7 +102,10 @@ extension RLMCollection {
      Returns the index of the first object in the collection matching the predicate.
      */
     public func indexOfObject(where predicateFormat: String, _ args: CVarArg...) -> UInt {
-        return indexOfObject(with: NSPredicate(format: predicateFormat, arguments: getVaList(args)))
+        guard let index = indexOfObject?(with: NSPredicate(format: predicateFormat, arguments: getVaList(args))) else {
```
tgoyne, line 105: Is this valid syntax?

Realm/RLMCollection.mm
```diff
@@ -211,6 +234,17 @@ NSUInteger RLMFastEnumerate(NSFastEnumerationState *state, NSUInteger len, id<RL
                 [array addObject:list];
             }
             return array;
+        } else if (prop && prop.set && prop.swiftIvar) {
```
tgoyne, line 237: The duplication here could be eliminated by renaming `initWithList:` and `initWithSet:` to something like `initWithBackingCollection:`, and then extracting this code to a function templated on the type. For just list/set this is kinda questionably worth doing, but since we know that we're adding a third container type it's worth doing some refactoring for that now.

tgoyne (commented):
RLMSetBase doesn't appear to be needed. Instead of duplicating all of that code for List, Set and Dictionary, you could rename RLMListBase to RLMSwiftCollectionBase and change all the references to `RLMArray` to `id<RLMCollection>`. `ListBase` would then define a `_rlmArray` property which performs the downcast to `RLMArray` (which means `List` shouldn't need any changes). The only functional difference between RLMSetBase and RLMListBase is that RLMSetBase calls RLMEnsureSetObservationInfo instead of RLMEnsureArrayObservationInfo, but those could also be combined into a single function.

Realm/RLMObservation.mm
leemaguire: The reason why this conditional is here is because the `forEach` iterates over the changes in `ObserverState` and passes back the `ColKey` allowing us to differentiate between collection types. I couldn't see any other way to avoid passing the populated indices without this.

Realm/Swift/RLMSupport.swift
leemaguire: Because `indexOfObject` has been marked optional in `RLMCollection` it requires to be unwrapped. The `?` in the method signature does look funky though.

Realm/RLMCollection.mm
leemaguire: I've condensed this down to 
```
        if (prop && prop.collection && prop.swiftIvar) {
            get_collection_type(prop, [&](auto t) {
                for (size_t i = 0; i < count; ++i) {
                    id<RLMCollectionBase> base = [[cls alloc] init];
                    base._rlmCollection = RLMManagedCollectionFromCollection<std::decay_t<decltype(*t)>>(info, collection.get(i), prop);
                    [array addObject:base];
                }
            });
```

the managed collection is constructed with
```
template<typename Collection>
id RLMManagedCollectionFromCollection(RLMClassInfo& info, realm::Obj&& obj, RLMProperty *prop) {
    if constexpr(std::is_same_v<Collection, realm::object_store::Set>) {
        return [[RLMManagedSet alloc] initWithBackingCollection:Collection(info.realm->_realm,
                                                                           obj,
                                                                           info.tableColumn(prop))
                                                     parentInfo:&info
                                                       property:prop];
    }
    else if constexpr(std::is_same_v<Collection, realm::List>) {
        return [[RLMManagedArray alloc] initWithBackingCollection:Collection(info.realm->_realm,
                                                                             obj,
                                                                             info.tableColumn(prop))
                                                       parentInfo:&info
                                                         property:prop];
    }
    else {
        REALM_UNREACHABLE();
    }
}
```
Is there a way to not require the use of `constexpr` in this case? The compiler gives out that it can't differentiate between `realm::List` and `realm::object_store::Set` when using an ambiguous `Class`

Realm/RLMObservation.mm
tgoyne: The code which is populating ObserverState should be skipping sets rather than going through the work of calculating the change indices and then just discarding them.

Realm/Swift/RLMSupport.swift
tgoyne: Ah, the method itself is optional. I don't think I've ever seen that before. I guess it's only possible via obj-c protocols so it won't show up much.

Realm/RLMCollection.mm
tgoyne:
```
template<typename ObjcCollection, typename Collection>
id RLMManagedCollectionFromCollection(RLMClassInfo& info, realm::Obj&& obj, RLMProperty *prop) {
    return [[ObjcCollection alloc] initWithBackingCollection:Collection(info.realm->_realm,
                                                                        obj,
                                                                        info.tableColumn(prop))
                                                  parentInfo:&info
                                                    property:prop];
}
```
should work. The key is that it needs to be passed as a c++ template parameter rather than as a type-erased Class.

Realm/RLMObservation.mm
leemaguire: I'll need to change this from the core side. Should this PR be blocked until that is done, or could it be placed in with Dictionary when that PR is available?

Realm/RLMCollection.mm
leemaguire: nice, I've updated it

Realm/RLMSet.mm
```diff
@@ -0,0 +1,515 @@
...
+
+// The compiler complains about the method's argument type not matching due to
+// it not having the generic type attached, but it doesn't seem to be possible
+// to actually include the generic type
+// http://www.openradar.me/radar?id=6135653276319744
+#pragma clang diagnostic ignored "-Wmismatched-parameter-types"
+- (RLMNotificationToken *)addNotificationBlock:(void (^)(RLMSet *, RLMCollectionChange *, NSError *))block {
+    return [self addNotificationBlock:block queue:nil];
+}
+- (RLMNotificationToken *)addNotificationBlock:(void (^)(RLMArray *, RLMCollectionChange *, NSError *))block
```
pavel-ship-it, line 475: RLMSet

leemaguire:
@tgoyne @jsflax can you PTAL

Realm/RLMObservation.mm
leemaguire: https://github.com/realm/realm-core/pull/4422/files

Realm/Tests/KVOTests.mm
```diff
@@ -300,6 +326,12 @@ bool empty() const {
     } \
 } while (false)
 
+#define AssertSetChanged(s) do { \
+    (r).refresh(); \
+    r.pop_front(); \
```
tgoyne, line 331:
```suggestion
#define AssertSetChanged() do { \
    AssertNotification(r); \
```

This currently is just asserting that there's at most one notification rather than exactly one.

tgoyne: Making this change reveals that many of the tests are incorrect.

leemaguire: Changed, I see some issues were due to expecting changes for objects with no KVO recorder setup

jsflax (approved):
Nice 🎉

leemaguire:
I have merged in the latest master. I will need to add support for primitive queries but will add that in an additional PR after this has been merged. @tgoyne could you PTAL?

Realm.xcodeproj/project.pbxproj
```diff
@@ -2969,6 +2969,7 @@
 			isa = XCBuildConfiguration;
 			baseConfigurationReference = 5DD755E31BE05EA1002800DA /* Tests iOS static.xcconfig */;
 			buildSettings = {
+				OTHER_LDFLAGS = "-ObjC";
```
tgoyne, line 2972: This should be in the xcconfig file.

Realm/RLMQueryUtil.mm
```diff
@@ -768,8 +770,8 @@ void validate_and_extract_between_range(id value, RLMProperty *prop, id *from, i
         // For arrays this effectively checks if there are any objects in the
         // array, while for links it's just always constant true or false
         // (for != and = respectively).
-        if (column.property().collection) {
-            add_bool_constraint(RLMPropertyTypeObject, operatorType, column.resolve<Link>(), null());
+        if (!column.link_map().only_unary_links()) {
```
tgoyne: I don't think this change is correct. `("array.link = %@", unmanagedObject)` should be FalseExpression.

Realm/RLMObservation.mm
leemaguire: I merged in the above, but it will need a core release so the xcproj can consume the changes

Realm/RLMQueryUtil.mm
leemaguire: This now uses what is in master, is master not correct? Just confused as it says this comment is Outdated.

tgoyne: Oh, that's a change from merging in master and I got confused by looking at the changes since last review. I think that's my bug then...

tgoyne (approved):
Looks good to go once we have a core release.

PR #7056, opened 2021-01-18
Implement queries on lists of primitives

### LINKED ISSUE #5334: Support queries involving arrays of primitives
We recently added support for arrays of primitives (#1120), but omitted support for queries on these properties due to time constraints. Queries involving these properties should be supported.

tgoyne:
Also incidentally adds support for a few other things along the way. Binary columns used to not support a lot of queries which were supported on everything else, but they now do and removing the special cases made it a lot easier to support `List<Data>`.

Depends on some unreleased functionality in core master, so can't be merged until a release is made there and all non-spm tests will fail.

Fixes #5334.


Realm/RLMQueryUtil.mm
```diff
@@ -192,10 +226,20 @@ bool operator()(Mixed v1, Mixed v2) const
...
+bool isNSNull(id value) {
+    return !value || value == NSNull.null;
+}
+
+template<typename T>
+bool isNSNull(T) {
+    return false;
+}
+
+
```
jsflax, line 242: extra newline

Realm/RLMQueryUtil.mm
```diff
@@ -446,41 +498,42 @@ void add_numeric_constraint(RLMPropertyType datatype,
...
+                                                   T value);
+
     template <typename R>
-    void add_constraint(RLMPropertyType type,
-                        NSPredicateOperatorType operatorType,
+    void add_constraint(NSPredicateOperatorType operatorType,
                         NSComparisonPredicateOptions predicateOptions,
                         ColumnReference const& column, R const& rhs);
-    template <typename... T>
+    template <template<typename> typename W, typename T>
```
jsflax, line 522: Can we give these template types better names? I can see how `W` is being used in the method body, but it seems like a fairly arbitrary designation that could use a bit more self documenting.

Realm/RLMQueryUtil.mm
```diff
@@ -835,9 +855,15 @@ void process_or_group(Query &query, id array, Func&& func) {
...
-    return RLMStringDataWithNSString(value);
+    return isNSNull(value) ? StringData() : RLMStringDataWithNSString(value);
 }
 
+template <>
+Binary convert<Binary>(id value) {
+    return isNSNull(value) ? BinaryData() : RLMBinaryDataForNSData(value);
+}
+
+
```
jsflax, line 866: extra newline

Realm/RLMQueryUtil.mm
```diff
@@ -869,77 +895,96 @@ auto value_of_type(const ColumnReference& column) {
...
         case RLMPropertyTypeLinkingObjects:
-            add_link_constraint(operatorType, values...);
+            convert_null(value, [&](auto&& value) {
+                add_link_constraint(operatorType, column.resolve<Link>(), value);
+            });
             break;
-        default:
+        case RLMPropertyTypeAny:
             throwException(@"Unsupported predicate value type",
                            @"Object type %@ not supported", RLMTypeToString(type));
```
jsflax, line 971:
```suggestion
                           @"Object type '%@' not supported", RLMTypeToString(type));
```

Realm/RLMQueryUtil.mm
tgoyne: W was initially Wrapper but that name wasn't very accurate and I failed to come up with a good name for it.

tgoyne:
Depends on https://github.com/realm/realm-core/pull/4417.

tgoyne:
This hit a new assertion failure in one of the CI runs which https://github.com/realm/realm-core/pull/4428 fixes. Since it's a core 10.5.0 bug and not something new in 10.5.1 it's not blocking this. The other CI failures are pre-existing issues and network failures.

PR #7057, opened 2021-01-20
Add unit tests for non-bridged numeric property values

### LINKED ISSUE #4011: Add test for `Object(value:)` initializer with non-bridged numeric property values
Just like https://github.com/realm/realm-cocoa/issues/4007 required a fix to explicitly bridge the primary key before calling into Objective-C code, we need to do a similar thing for `Object(value:)` and `Realm.create(value:)`.

This is a little tricky though because, when the value is passed from Swift to Objective-C, it is a `Dictionary` or an `Array` or maybe even some other KVC-compliant object. We need a way to wall into Swift's bridging functionality from Objective-C. It won't be pretty. The alternative is to eagerly convert it into a bridged collection, recursively, which is less than ideal.

DominicFrei:
Fixes #4011.

Summary:
- unit tests for `Object(value:)`
- unit tests for `Realm.create()`
- add `SwiftInt8Object`, `SwiftInt16Object`, `SwiftInt32Object` and `SwiftInt64Object` on top of the existing `SwiftIntObject`

### DIFF
--- a/RealmSwift/Sync.swift
+++ b/RealmSwift/Sync.swift
@@ -268,7 +268,7 @@ import Combine
 /// This handler is executed on a non-main global `DispatchQueue`.
 @dynamicMemberLookup
 @frozen public struct Functions {
-    
+
     private let user: User
 
     fileprivate init(user: User) {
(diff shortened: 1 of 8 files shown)


jsflax:
All pull requests should have meaningful names in the imperative mood. In this case, it's close, but needs the imperative and should be: `Add unit tests for non-bridged numeric property values`.

jsflax (approved):

tgoyne (approved):

PR #7061, opened 2021-01-21
Propagate error code to object store

pavel-ship-it:
This will fix several known issues (timeout, no internet connection) by providing the error code to the object store where the error could be recognised.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ x.y.z Release notes (yyyy-MM-dd)
 * Fixed an issue where creating an object after file format upgrade may fail
   with assertion "Assertion failed: lo() <= std::numeric_limits<uint32_t>::max()"
  ([#4295](https://github.com/realm/realm-core/issues/4295), since v5.0.0)
+ * Fix - `RLMResponse` will have a non nil `customStatusCode` in case of error. ([#4188](https://github.com/realm/realm-core/issues/4188))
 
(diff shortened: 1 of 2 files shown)


Realm/RLMNetworkTransport.mm
```diff
@@ -156,7 +156,8 @@ - (void)URLSession:(__unused NSURLSession *)session
     response.httpStatusCode = httpResponse.statusCode;
 
     if (error) {
-        response.body = [error localizedDescription];
+        response.body = error.localizedDescription;
+        response.customStatusCode = error.code;
```
leemaguire, line 160: Would it make sense to have a test that ensures the `customStatusCode` gets mapped to the expected code?

pavel-ship-it: Main thing here is to expose a code and any error will have one.

CHANGELOG.md
```diff
@@ -43,6 +42,7 @@ x.y.z Release notes (yyyy-MM-dd)
   assertion failure rather than a more appropriate exception.
   ([Core #4114](https://github.com/realm/realm-core/issues/4114), since v5.0.0).
 * Fix an issue where calling a MongoDB Realm Function would never be performed as the reference to the weak `User` was lost.
+* Fix error code propagation to object store.
```
tgoyne, line 45: This doesn't mean anything to the person reading it. The changelog is for developers using the SDK and needs to explain what the change is to them.

Also it's in the wrong section.

leemaguire (approved):

Issue #6654, opened 2020-07-17
Insert document via MongoClient, the value of AnyBSON(Date) will be transferred to Timestamp @ v10.0.0-beta.2


ananfang:
## Goals

Using the app.mongoClient to insert the document, the value of AnyBSON(Date) will be transferred to Timestamp in MongoDB, and even the timestamp is wrong.

## Expected Results

It should be recorded as the Date type in the MongoDB.

## Actual Results

```
let app = RealmApp(id: "realm-app-id")
let collection = app.mongoClient("mongodb-atlas").database(withName: "some-database").collection(withName: "some-collection")
let document: Document = [
  "updatedAt": AnyBSON(Date(timeIntervalSince1970: 0))
]
collection.insertOne(document) { objectId, error in
  // Do something after inserting
}
```
The field "updatedAt" will be the type of Timestamp, and the value will be always different.
(e.g. 6850268160453509127, 6850268302187429898 and etc., extremely big and obviously not timestamp number)

## Steps to Reproduce

Check **Actual Results**

## Code Sample

Check **Actual Results**

## Version of Realm and Tooling

Realm framework version: v10.0.0-beta.2

Realm Object Server version: MongoDB Realm

Xcode version: 11.5

iOS/OSX version: 13.5 (simulator)

Dependency manager + version: CocoaPods 1.9.3

homathko:
Same issue. Passing Foundation.Date object into AnyBSON.datetime( ) stores the document in Atlas with a timestamp value, which cannot be read by the realm client -> "Object was deleted or invalidated". I've spent about 5 hours now trying to sort this out and can't find any documentation. This is the closest hit yet and it was posted almost 5 months ago...?

leemaguire:
@ananfang this is the expected behaviour, as the BSON speciation states:

```
BSON Date is a 64-bit integer that represents the number of milliseconds since the Unix epoch (Jan 1, 1970). This results in a representable date range of about 290 million years into the past and future.
```
https://docs.mongodb.com/manual/reference/bson-types/#date

leemaguire:
@homathko can you share your MongoDB Realm schema & Swift object with us so we can investigate?

homathko:
I wish I could!  This was awhile ago and is out of my scope now - I am having trouble recalling any details. It was working with cloud functions and custom user data, so this probably isn't the place for further troubleshooting anyhow..

ananfang:
[...]

PR #7062, opened 2021-01-21
Use realm::Timestamp instead of MongoTimestamp for RLMBSONTypeDatetime

leemaguire:
As shown in https://github.com/realm/realm-cocoa/issues/6654 when you insert a `Date` into a collection with `AnyBSON.datetime(...)` the value is inserted as a `Timestamp` and not `Date`. Synced Realm objects rely on the `Date` field and not `Timestamp` so inserting/updating a record via `mongoClient` breaks and invalidates the modified objects.


Realm/RLMBSON.mm
```diff
@@ -324,7 +324,7 @@ Bson RLMConvertRLMBSONToBson(id<RLMBSON> b) {
         case RLMBSONTypeTimestamp:
             return RLMTimestampForNSDate((NSDate *)b);
         case RLMBSONTypeDatetime:
-            return MongoTimestamp(((NSDate *)b).timeIntervalSince1970, 0);
+            return realm::Timestamp(((NSDate *)b).timeIntervalSince1970, 0);
```
jsflax, line 327: Something feels a bit off about this– the reason MongoTimestamp was used here is because mongo and realm hold timestamps differently, and they were only compatible with lossiness. 

Can we at least get an integration test demonstrating this? I'm surprised this doesn't break more.

jsflax (approved):

PR #7064, opened 2021-01-22
Allow enumerating objects of types which are no longer present in the schema

### LINKED ISSUE #6734: Unable to migrate objects that are deleted from schema
## Goals
I want to enumerate objects that are deleted from schema in migration block so that I could move information from those objects to separate place.

## Expected Results
RLMMigration.enumerateObjects block to be called with valid oldObject and newObject is null.

## Actual Results
RLMMigration.enumerateObjects block is not called at all for object types that are deleted from schema.

## Probable cause
Update to latest object store in Realm 5.0 seemed to change condition when migration block is called. Previously block was called if queried object type was available only in old schema but current implementation seems to require that object type exists in both old and new schema.

Link to the commit that might have caused the issue:
[https://github.com/realm/realm-cocoa/commit/982c0780aa307e3410264f77983e3115c20c795c](https://github.com/realm/realm-cocoa/commit/982c0780aa307e3410264f77983e3115c20c795c#diff-b5dc054208ebfe86b2516e480a62d4c5R97)

Realm framework version: 5.3.6
Realm Object Server version: -
Xcode version: 11.6
iOS version: 13.6
Dependency manager + version: -

DominicFrei:
Fixes #6734.

The same way we enumerate over freshly added objects during a migration this PR allows for the enumeration of objects that will be deleted during this migration to make sure data can be saved or transferred to other objects if necessary.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,7 @@
 x.y.z Release notes (yyyy-MM-dd)
 =============================================================
 ### Enhancements
-* None.
+* Allow enumerating objects within the migration they get deleted in.
 
 ### Fixed
 * <How to hit and notice issue? what was the impact?> ([#????](https://github.com/realm/realm-cocoa/issues/????), since v?.?.?)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,13 +1,14 @@
 x.y.z Release notes (yyyy-MM-dd)
 =============================================================
 ### Enhancements
-* Allow enumerating objects within the migration they get deleted in.
+* None. 
 
 ### Fixed
 * <How to hit and notice issue? what was the impact?> ([#????](https://github.com/realm/realm-cocoa/issues/????), since v?.?.?)
 * Fixed an issue where creating an object after file format upgrade may fail
   with assertion "Assertion failed: lo() <= std::numeric_limits<uint32_t>::max()"
  ([#4295](https://github.com/realm/realm-core/issues/4295), since v5.0.0)
+* Allows enumerating objects of types which are no longer present in the schema.
 
 <!-- ### Breaking Changes - ONLY INCLUDE FOR NEW MAJOR version -->
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,7 @@
 x.y.z Release notes (yyyy-MM-dd)
 =============================================================
 ### Enhancements
-* None. 
+* None.
 
 ### Fixed
 * Inserting a date into a synced collection via `AnyBSON.datetime(...)` would be of type `Timestamp` and not `Date`. 
@@ -10,7 +10,7 @@ x.y.z Release notes (yyyy-MM-dd)
   with assertion "Assertion failed: lo() <= std::numeric_limits<uint32_t>::max()"
  ([#4295](https://github.com/realm/realm-core/issues/4295), since v5.0.0)
 * Allows enumerating objects of types which are no longer present in the schema.
- * Fix - `RLMResponse` will have a non nil `customStatusCode` in case of error. ([#4188](https://github.com/realm/realm-core/issues/4188))
+* Add `RLMResponse.customStatusCode`. This fixes timeout exceptions that were occuring with a poor connection. ([#4188](https://github.com/realm/realm-core/issues/4188))
 
 <!-- ### Breaking Changes - ONLY INCLUDE FOR NEW MAJOR version -->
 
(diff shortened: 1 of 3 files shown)


Realm/Tests/MigrationTests.mm
```diff
@@ -1133,21 +1133,29 @@ - (void)testEnumerateObjectsAfterDeleteDataForRemovedType {
...
         }];
         [migration deleteDataForClassName:IntObject.className];
+    };
+    XCTAssertTrue([RLMRealm performMigrationForConfiguration:config1 error:nil]);
+    
+    RLMRealmConfiguration *config2 = self.config;
+    config2.objectClasses = @[StringObject.class];
+    config2.schemaVersion = 2;
+    config2.migrationBlock = ^(RLMMigration *migration, uint64_t) {
         [migration enumerateObjects:IntObject.className block:^(RLMObject *, RLMObject *) {
```
jsflax, line 1154: Should add an assertion inside the migration block to ensure that it's executed

DominicFrei: There is an `XCTFail(@"should not have enumerated any objects");` in the next line. This block is supposed to **not** be executed since `IntObject` was already deleted one version earlier.

jsflax: Yes, but the migration block is meant to be called. My comment should have been one line higher

DominicFrei: Oh, true! Good point. Adding a check there too. 👍

jsflax (approved):

CHANGELOG.md
```diff
@@ -1,7 +1,7 @@
 x.y.z Release notes (yyyy-MM-dd)
 =============================================================
 ### Enhancements
-* None.
+* Allow enumerating objects within the migration they get deleted in.
```
tgoyne, line 4: This is a bug fix and not an enhancement. It's also not an accurate description of the change; it allows enumerating objects of types which are no longer present in the schema, which is unrelated to objects being deleted.

Realm/Tests/MigrationTests.mm
```diff
@@ -1133,21 +1133,35 @@ - (void)testEnumerateObjectsAfterDeleteDataForRemovedType {
         [IntObject createInRealm:realm withValue:@[@2]];
     }];
 
-    RLMRealmConfiguration *config = self.config;
-    config.objectClasses = @[StringObject.class];
-    config.schemaVersion = 1;
-    config.migrationBlock = ^(RLMMigration *migration, uint64_t) {
+    RLMRealmConfiguration *config1 = self.config;
```
tgoyne, line 1136: This test passes without the newly added code, and is unrelated to the reported bug. The thing which didn't work is the following:

```
- (void)testEnumerateObjectTypeRemovedFromSchema {
    [self createTestRealmWithClasses:@[IntObject.class] block:^(RLMRealm *realm) {
        [IntObject createInRealm:realm withValue:@[@1]];
        [IntObject createInRealm:realm withValue:@[@2]];
    }];

    RLMRealmConfiguration *config = self.config;
    config.objectClasses = @[StringObject.class];
    config.schemaVersion = 1;
    __block int enumerateCalls = 0;
    config.migrationBlock = ^(RLMMigration *migration, uint64_t) {
        [migration enumerateObjects:IntObject.className block:^(RLMObject *oldObject, RLMObject *newObject) {
            XCTAssertNotNil(oldObject);
            XCTAssertNil(newObject);
            ++enumerateCalls;
            XCTAssertGreaterThan([oldObject[@"intCol"] intValue], 0);
        }];
    };
    XCTAssertTrue([RLMRealm performMigrationForConfiguration:config error:nil]);
    XCTAssertEqual(enumerateCalls, 2);
}
```

DominicFrei:
The adjustments in the change log were made in with Jason together.
We also talked about the test and I want to further explain why i removed the other test with Thomas' suggestion.
The old test did test if the enumerate block does not called again after `deleteData` was called. That call was not executed if `oldObjects` or `newObjects` is empty.
Since this is not being tested anymore before the migration block is called for `oldObjects` the test was no longer applicable and was therefore replaced with the new test instead of using them both.

tgoyne (approved):

CHANGELOG.md
```diff
@@ -9,7 +9,8 @@ x.y.z Release notes (yyyy-MM-dd)
 * Fixed an issue where creating an object after file format upgrade may fail
   with assertion "Assertion failed: lo() <= std::numeric_limits<uint32_t>::max()"
  ([#4295](https://github.com/realm/realm-core/issues/4295), since v5.0.0)
- * Fix - `RLMResponse` will have a non nil `customStatusCode` in case of error. ([#4188](https://github.com/realm/realm-core/issues/4188))
+* Allows enumerating objects of types which are no longer present in the schema.
```
jsflax, line 12:
```suggestion
* Allow enumerating objects of types which are no longer present in the schema.
```

jsflax (approved):
One suggestion, otherwise LGTM

PR #7066, opened 2021-01-25
Fix SPM test compilation for iOS

tgoyne:
`isEqualTo:` is a dumb thing from the AppleScript API that doesn't exist on iOS and shouldn't be used, and the sync tests need platform guards since we can't make the files macOS-specific.


jsflax (approved):
Nice, this makes life a bit easier

PR #7073, opened 2021-01-27
Change RealmSwift Analytics

ericjordanmossman:
RLMAnalytics will check for `RealmSwiftDecimal128` when determining if RealmSwift is in use.
"SwiftVersion" is removed. Instead "Clang Version" and "Clang Major Version" are reported. These are [tied to Xcode](https://gist.github.com/yamaya/2924292) version.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -246,7 +246,6 @@ build_docs() {
     local objc="--objc"
 
     if [[ "$language" == "swift" ]]; then
-        sh build.sh set-swift-version
         xcodebuild_arguments="-scheme,RealmSwift"
         module="RealmSwift"
         objc=""
@@ -425,21 +424,6 @@ case "$COMMAND" in
 
     "setup-baas")
         ruby Realm/ObjectServerTests/setup_baas.rb
-        exit 0
-        ;;
-
-    ######################################
-    # Swift versioning
-    ######################################
-    "set-swift-version")
-        version=${2:-$REALM_SWIFT_VERSION}
-
-        SWIFT_VERSION_FILE="RealmSwift/SwiftVersion.swift"
-        CONTENTS="let swiftLanguageVersion = \"$version\""
-        if [ ! -f "$SWIFT_VERSION_FILE" ] || ! grep -q "$CONTENTS" "$SWIFT_VERSION_FILE"; then
-            echo "$CONTENTS" > "$SWIFT_VERSION_FILE"
-        fi
-
         exit 0
         ;;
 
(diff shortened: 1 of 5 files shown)


build.sh
```diff
@@ -1057,21 +1041,17 @@ case "$COMMAND" in
     # CocoaPods
     ######################################
     "cocoapods-setup")
-        if [[ "$2" != "swift" ]]; then
-          if [ ! -f core/version.txt ]; then
-            sh build.sh download-core
-          fi
+        if [ ! -f core/version.txt ]; then
```
tgoyne, line 1044: We don't want to do any of this for the RealmSwift pod, but now that there's *no* setup for it that should be done by removing the prepare_command from RealmSwift.podspec.

ericjordanmossman:
[Jenkins: swiftlint 12.4](https://ci.realm.io/job/objc_pr/target=swiftlint,xcode_version=12.4/7528/console) - Plaground/Contents.swift
[Jenkins: swiftpm-thread 12.4](https://ci.realm.io/job/objc_pr/target=swiftpm-thread,xcode_version=12.4/7528/consoleFull) - testAsyncOpenTimeout
<details>
Jan 28 13:34:32 SwiftObjectServerTests.SwiftObjectServerTests
Jan 28 13:34:32   testAsyncOpenTimeout, failed - Did not fail: success(RealmSwift.Realm(rlmRealm: <RLMRealm: 0x7b1c001fe900>))
Jan 28 13:34:32   /Users/realm/workspace/objc_pr/target/swiftpm-thread/xcode_version/12.4/Realm/ObjectServerTests/SwiftObjectServerTests.swift:603
Jan 28 13:34:32   ```
Jan 28 13:34:32                 guard case .failure(let error) = result else {
Jan 28 13:34:32                     XCTFail("Did not fail: \(result)")
Jan 28 13:34:32                     return
Jan 28 13:34:32   ```
</details>

[Jenkins: tvos 11.3](https://ci.realm.io/job/objc_pr/target=tvos,xcode_version=11.3/7528/console) Didn't run correctly. Can't tell if it's connection problems or build problem introduced by build.sh. @tgoyne is there a way to run the CI again for only a single portion of the matrix instead of each module?

tgoyne:
No, only the full matrix can be rerun. Both of those are pre-existing issues and the analytics code isn't even run on CI, so they are unrelated to the changes here.

build.sh
ericjordanmossman: If I read this right: resolved by 1195e5a.

tgoyne (approved):

PR #7077, opened 2021-01-29
Extend the Migration examples project

### LINKED ISSUE #7069: Update migration examples to better reflect possible pitfalls (Swift)
## Goals
The [current example for migrations](https://github.com/realm/realm-cocoa/blob/master/examples/ios/swift/Migration/AppDelegate.swift) should showcase as many pitfalls and problems as possible and their solutions.

## Expected Results
A better explanation how to work around issues like the following:
- [changing the 'embeddedness' of an object](https://github.com/realm/realm-cocoa/issues/7060)
- [adding properties and changing them in a later migration](https://github.com/realm/realm-cocoa/issues/6263)
- [renaming tables](https://github.com/realm/realm-cocoa/issues/2491)
- [enumerating objects with deleted properties](https://github.com/realm/realm-cocoa/issues/6734)

DominicFrei:
Fixes #7069.

This PR enhances our migration examples. I also tried to achieve a state that is self explanatory (aside from a brief read me on how to create new versions) therefore I will not explain to much here.
Open questions will be an indicator that the example needs to be further re-worked.

I will apply that example to the Objective-C project when everything is reviewed so that the changes do not have to be applied twice.


examples/ios/swift/Migration/README
```diff
@@ -0,0 +1,29 @@
...
+You can just build and run the project since this is the default state (see
+`shouldCreateExamples` in `AppDelegate`).
+There is nothing much so see in terms of unit tests, console log or in the app itself.
+If there is no error the migration was successful.
+The files to look at are located in the `Examples` folder. Every file contains an extract
+of everything necessary for this version (schema version, objects, migration).
+
+How to create new examples:
+- Set `shouldCreateExamples` in `AppDelegate` to true.
+- Duplicate the most recent `Example_vx`.
```
jsflax, line 16: What does this mean?

examples/ios/swift/Migration/README
```diff
@@ -0,0 +1,29 @@
+The Migration project shows several examples of migrations and migration blocks.
+
+Goal is to provide a more indept view of certain problems and pitfalls users might
+face when migrations get more complex and the number of versions increases.
+
+How to use the example:
+You can just build and run the project since this is the default state (see
```
jsflax, line 7: Should avoid words like `just` in instructions like this.

examples/ios/swift/Migration/URLExtension.swift
```diff
@@ -0,0 +1,30 @@
...
+import Foundation
+import RealmSwift
+
+extension URL {
+    init(for version: RealmVersion, usingTemplate: Bool) {
+        let defaultURL = Realm.Configuration.defaultConfiguration.fileURL!
+        let defaultParentURL = defaultURL.deletingLastPathComponent()
+        let fileName = "default-v\(version.rawValue)"
+        let destinationUrl = defaultParentURL.appendingPathComponent(fileName + ".realm")
+        print("destinationUrl: \(destinationUrl)")
```
jsflax, line 18: Should avoid print statements in examples unless absolutely necessary.

examples/ios/swift/Migration/URLExtension.swift
```diff
@@ -0,0 +1,30 @@
...
+//  Migration
+//
+//  Created by Dominic Frei on 28/01/2021.
+//  Copyright © 2021 Realm. All rights reserved.
+//
+
+import Foundation
+import RealmSwift
+
+extension URL {
```
jsflax, line 12: I would avoid extensions like this as it reduces readability, but if you can't find a better way to do this, I would move it into the AppDelegate file.

jsflax (commented):
Deleting the realm files from git but keeping them included in the xcodeproj prevents the project from being built as well.

examples/ios/swift/Migration/AppDelegate.swift
```diff
@@ -59,85 +26,23 @@ extension UIApplication {
...
-            print("Migrated objects in the Realm migrated from v1: \(realmv1.objects(Person.self))")
-            let realmv2 = try! Realm(configuration: realmv2Configuration)
-            print("Migrated objects in the Realm migrated from v2: \(realmv2.objects(Person.self))")
+        
+        // Choose to either create example files or execute them.
+        let shouldCreateExamples = false
+        if shouldCreateExamples {
+            RealmTemplate.create()
+        } else {
+            MigrationExample.execute()
```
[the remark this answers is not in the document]
DominicFrei: I've moved both into one file, removed static and renamed one function to tackle `static methods on arbitrary structs in different files`.

examples/ios/swift/Migration/README
DominicFrei: Would this be clearer?
```
Duplicate the most recent `Example_v5.swift` file
```

examples/ios/swift/Migration/URLExtension.swift
DominicFrei: I made another suggestion and moved it to `extension RealmVersion`.

jsflax:
There are still a number of unresolved comments.

jsflax (changes requested):
There are still a number of unresolved comments.

examples/ios/swift/Migration/Examples/Example_v0.swift
```diff
@@ -0,0 +1,51 @@
...
+////////////////////////////////////////////////////////////////////////////
+
+// swiftlint:disable orphaned_doc_comment
+// swiftlint:disable comment_spacing
+// swiftlint:disable mark
+
+import Foundation
+import RealmSwift
+
+//// MARK: - Schema
```
leemaguire, line 26: Why is all of this commented out?

leemaguire: What if the Realm API changes? we won't be to catch any compiler errors

examples/ios/swift/Migration/README
```diff
@@ -0,0 +1,29 @@
...
+- Comment out the previous one. Update the current one:
+  - Add a new version to the `RealmVersion` enum.
+  - Update the `Object` definitions according to your new example.
+  - Update the migration block.
+  - Create some examples for that specific schema version (they will be used in later schema version).
+- Run the app.
+- The realm file (path printed to console) created by this run must then be added to project.
+
+Important notes:
+- Code in all `Example_vx` files except the most recent one needs to be commented out to avoid
```
leemaguire, line 26: We shouldn't do anything with commenting out code, its not stable and there are more efficient ways to do it, i.e preprocessor flags

leemaguire (changes requested):

examples/ios/swift/Migration/AppDelegate.swift
```diff
@@ -17,39 +17,6 @@
...
-    @objc dynamic var fullName = ""
-    @objc dynamic var age = 0
-    let pets = List<Pet>() // Add pets field
-}
-
-func bundleURL(_ name: String) -> URL? {
-    return Bundle.main.url(forResource: name, withExtension: "realm")
-}
 
 #if !swift(>=4.2)
```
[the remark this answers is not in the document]
DominicFrei: I don't know. Do we support Swift < 4.2 still?

examples/ios/swift/Migration/AppDelegate.swift
```diff
@@ -59,85 +26,23 @@ extension UIApplication {
...
-            try! Realm.performMigration(for: realmv1Configuration)
-
-            // print out all migrated objects in the migrated realms
-            let realmv1 = try! Realm(configuration: realmv1Configuration)
-            print("Migrated objects in the Realm migrated from v1: \(realmv1.objects(Person.self))")
-            let realmv2 = try! Realm(configuration: realmv2Configuration)
-            print("Migrated objects in the Realm migrated from v2: \(realmv2.objects(Person.self))")
+        
+        // Choose to either create example files or execute them.
+        let shouldCreateExamples = false
```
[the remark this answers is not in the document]
DominicFrei: I thought about the target idea too but that would create a lot of new targets over time when we add more versions. I do think it would be easier for the user though, yes. `xcconfig` makes it more clean on the other side. Mentioning that in the README might be enough.

@jsflax Will we keep the config idea then?

examples/ios/swift/Migration/Examples/Example_v0.swift
DominicFrei: This will be fixed as soon as we have decided on config vs. target.

examples/ios/swift/Migration/AppDelegate.swift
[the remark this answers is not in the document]
jsflax: Not sure multiple targets make the user's experience simpler though. Config flags make it pretty clear what's going on and easy enough to test

examples/ios/swift/Migration/Examples/Example_v5.swift
```diff
@@ -0,0 +1,158 @@
...
+// limitations under the License.
+//
+////////////////////////////////////////////////////////////////////////////
+
+import Foundation
+import RealmSwift
+
+// MARK: - Schema
+
+enum RealmVersion: Int, CaseIterable {
```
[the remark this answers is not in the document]
DominicFrei: I removed it completely and implemented the global `Int` idea.

examples/ios/swift/Migration/AppDelegate.swift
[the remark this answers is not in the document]
jsflax: Yeah, I think a config file with comments around the guards explaining the "what's and why's" is the cleanest solution here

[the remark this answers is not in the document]
DominicFrei: @leemaguire addresses a valid point there though. API changes and compile errors won't get caught. Neither by my solution nor by the pre-compiler flags. Having multiple targets (all included in the `Migration` schema) would solve this quite nicely. @jsflax 

In both cases we would have to explain this in the `README` or some other appropriate location.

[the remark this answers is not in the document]
jsflax: The examples compilation should be tested on the CI, in which case we would pass the flags to the compiler. Being able to compile this locally isn't a great test to see if something has broken.

But yes, the explanation and instructions need to be in the README.

examples/ios/swift/Migration/README
DominicFrei: Fixed with the `xcconfig` file.

examples/ios/swift/Migration/AppDelegate.swift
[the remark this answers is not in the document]
DominicFrei: @jsflax How to decide this?

DominicFrei:
@jsflax @leemaguire  I marked everything as `Resolved` that I worked on. Hope that was the correct workflow. We still have to decide about the Swift 4.2 question though.

examples/ios/swift/Migration/AppDelegate.swift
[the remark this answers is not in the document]
jsflax: We support the minimum version of Swift that our minimum supported Xcode version supports (and we support 1 major version beneath the current Xcode version): https://developer.apple.com/support/xcode/.  So, yes.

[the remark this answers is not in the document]
jsflax: However, I think the better question is, are we still using `UIApplicationLaunchOptionsKey`?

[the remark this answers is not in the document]
DominicFrei: `UIApplicationLaunchOptionsKey` was substituted by `UIApplication.LaunchOptionsKey` in Swift 4.2. So `UIApplication.LaunchOptionsKey` will probably not work is someone is using Swift 4.0 on Xcode 11.0, which is still supported according to your commend above.
I don't understand how the `typealias` solves this though since `UIApplicationLaunchOptionsKey` is not nested inside `UIApplication` anyway but a global key.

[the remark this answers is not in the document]
jsflax: But does the example still need it given your changes?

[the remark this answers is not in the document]
DominicFrei: @jsflax  It doesn't have anything to do with the examples. It's defined in all `AppDelegate` files in all examples since it's part of

```
func application(_ application: UIApplication, didFinishLaunchingWithOptions launchOptions: [UIApplication.LaunchOptionsKey: Any]? = nil) -> Bool {
```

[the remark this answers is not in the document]
DominicFrei: So the answer was already given by you by saying we still support Swift <4.2. :)
Until Xcode 13 is has to stay there then.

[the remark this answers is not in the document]
jsflax: @DominicFrei has tested this locally and it no longer seems to be used and can be removed.

examples/ios/swift/Migration/Migration.xcconfig
```diff
@@ -0,0 +1,26 @@
...
+////////////////////////////////////////////////////////////////////////////
+
+//OTHER_SWIFT_FLAGS = $(inherited) -DCREATE_EXAMPLES;
+
+//OTHER_SWIFT_FLAGS = $(inherited) -DSCHEMA_VERSION_0;
+//OTHER_SWIFT_FLAGS = $(inherited) -DSCHEMA_VERSION_1;
+//OTHER_SWIFT_FLAGS = $(inherited) -DSCHEMA_VERSION_2;
+//OTHER_SWIFT_FLAGS = $(inherited) -DSCHEMA_VERSION_3;
+//OTHER_SWIFT_FLAGS = $(inherited) -DSCHEMA_VERSION_4;
+OTHER_SWIFT_FLAGS = $(inherited) -DSCHEMA_VERSION_5;
```
leemaguire, line 26: Why is this starting on the highest version?

examples/ios/swift/Migration/README.md
```diff
@@ -0,0 +1,28 @@
+# Migration Examples
+
+The Migration project shows several examples of migrations and migration blocks.
+
+The purpose of this example is to provide a more in depth view of certain problems and pitfalls users might face when migrations get more complex and the number of versions increases.
+
+## How to use the example
+
+You can build and run the project since this is the default state (see `CREATE_EXAMPLES` pre-compiler flag in `AppDelegate`).
+There is nothing much so see in terms of unit tests, console log or in the app itself. If there is no error the migration was successful.
```
leemaguire, line 10: You should `assert` objects to see if they contain the correct data. The 'no news is good news' approach here is flaky.

examples/ios/swift/Migration/README.md
```diff
@@ -0,0 +1,28 @@
...
+The files to look at are located in the `Examples` folder. Every file contains an extract of everything necessary for this version (schema version, objects, migration).
+
+How to create new examples:
+1. Uncomment out the `-DCREATE_EXAMPLES` line in `Migration.xcconfig`.
+2. Duplicate `OTHER_SWIFT_FLAGS = $(inherited) -DSCHEMA_VERSION_5;` and increase the version by 1.
```
leemaguire, line 15: This needs more direction. Are you telling the user to go to the 'Build Settings' for the Xcode project and add this flag (`-DSCHEMA_VERSION_5;`) to `Other Swift Flags`?

examples/ios/swift/Migration/README.md
```diff
@@ -0,0 +1,28 @@
...
+3. Duplicate the most recent `Example_v`x`.swift` file.
+4. Comment out the previous one. Update the current one:
+  * Increase the `schemaVersion`.
+  * Update the `Object` definitions according to your new example.
+  * Update the migration block.
+  * Create some examples for that specific schema version (they will be used in later schema version).
```
leemaguire, line 21: All of this looks to be in the specified Example_vX.swift file?

examples/ios/swift/Migration/README.md
```diff
@@ -0,0 +1,28 @@
...
+## How to use the example
+
+You can build and run the project since this is the default state (see `CREATE_EXAMPLES` pre-compiler flag in `AppDelegate`).
+There is nothing much so see in terms of unit tests, console log or in the app itself. If there is no error the migration was successful.
+The files to look at are located in the `Examples` folder. Every file contains an extract of everything necessary for this version (schema version, objects, migration).
+
+How to create new examples:
+1. Uncomment out the `-DCREATE_EXAMPLES` line in `Migration.xcconfig`.
+2. Duplicate `OTHER_SWIFT_FLAGS = $(inherited) -DSCHEMA_VERSION_5;` and increase the version by 1.
+3. Duplicate the most recent `Example_v`x`.swift` file.
```
leemaguire, line 16: Is this step required even with the XCConfig flags?

examples/ios/swift/Migration/README.md
leemaguire: also version 5 looks to be the highest version, should they not be starting on version 0?

leemaguire (changes requested):

examples/ios/swift/Migration/Examples/Example_v4.swift
```diff
@@ -0,0 +1,165 @@
...
+@objc enum PetType: Int, RealmEnum {
+    case unspecified
+    case dog
+    case chicken
+    case cow
+}
+
+class Pet: Object {
+    @objc dynamic var name = ""
+    let type = RealmOptional<PetType>()
```
jsflax, line 40: Why is PetType optional?

examples/ios/swift/Migration/Examples/Example_v4.swift
```diff
@@ -0,0 +1,165 @@
...
+import RealmSwift
+
+// MARK: - Schema
+
+let schemaVersion = 4
+
+// Changes from previous version:
+// Add an `Address` to the `Person`.
+
+@objc enum PetType: Int, RealmEnum {
```
jsflax, line 31: `PetType` should be an inner enum of `Pet` and just be called `Type`

examples/ios/swift/Migration/MigrationExample.swift
```diff
@@ -0,0 +1,61 @@
...
+        let url = realmUrl(for: schemaVersion, usingTemplate: false)
+        let configuration = Realm.Configuration(fileURL: url, schemaVersion: UInt64(schemaVersion))
+        let realm = try! Realm(configuration: configuration)
+
+        try! realm.write {
+            exampleData(realm)
+        }
+
+        // Uncomment the following line to print the location of the newly created Realm.
+//        print("Realm created at: \(String(describing: configuration.fileURL!)).")
```
jsflax, line 34: It's unclear to me why this is here– if for some reason the user does need the location of the Realm, this comment should explain why they might, and printing it definitely doesn't seem like the right way to inform the user of this information

examples/ios/swift/Migration/MigrationExample.swift
```diff
@@ -0,0 +1,61 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+////////////////////////////////////////////////////////////////////////////
+
+import Foundation
+import RealmSwift
+
+struct MigrationExample {
```
jsflax, line 22: Why not have these methods just be in the AppDelegate class? Why make the user have to jump to another file?

examples/ios/swift/Migration/README.md
jsflax, line 21: Indentation looks off for these

examples/ios/swift/Migration/README.md
```diff
@@ -0,0 +1,28 @@
...
+
+The Migration project shows several examples of migrations and migration blocks.
+
+The purpose of this example is to provide a more in depth view of certain problems and pitfalls users might face when migrations get more complex and the number of versions increases.
+
+## How to use the example
+
+You can build and run the project since this is the default state (see `CREATE_EXAMPLES` pre-compiler flag in `AppDelegate`).
+There is nothing much so see in terms of unit tests, console log or in the app itself. If there is no error the migration was successful.
+The files to look at are located in the `Examples` folder. Every file contains an extract of everything necessary for this version (schema version, objects, migration).
```
jsflax, line 11: should add a new line somewhere in here so that the line isn't excessively long

jsflax (commented):
There should also be an adjustment to the compilation test that runs on RealmExamples.xcworkspace (from `build.sh`) to account for the multiple configuration flags.

examples/ios/swift/Migration/Migration.xcconfig
DominicFrei: `-DSCHEMA_VERSION_5` means this is the version we are migrating to using all versions lower than that.

If `-DCREATE_EXAMPLES` is set this is the version that will be created.

What do you think about `TARGET_SCHEMA_VERSION`? Would that clarify what we are doing here?

examples/ios/swift/Migration/MigrationExample.swift
DominicFrei: My idea behind this was a separation of concerns:
The `AppDelegate` manages the app lifecycle by implementing the `AppDelegate` `protocol` functions.
`MigrationExample` manages the actual migration example.

Would you prefer to move this to the `AppDelegate`?

examples/ios/swift/Migration/MigrationExample.swift
DominicFrei: My initial idea was to make it easier to find the created realm example files. Removing this might make things easier to understand for the user though. 👍

examples/ios/swift/Migration/Examples/Example_v4.swift
```diff
@@ -0,0 +1,167 @@
...
+
+// MARK: - Schema
+
+let schemaVersion = 4
+
+// Changes from previous version:
+// Add an `Address` to the `Person`.
+
+class Pet: Object {
+
```
jsflax, line 32: extra newline

examples/ios/swift/Migration/MigrationExample.swift
jsflax: Problem is, `MigrationExample` isn't a "thing"– having it as a struct is a completely arbitrary choice for the purpose of the example, and we should avoid anything being "for the purpose of the example" as much as possible.

If you're hesitant to put these in the `AppDelegate` itself, they can just be private global functions in the app delegate file. It's not like they have any state.

examples/ios/swift/Migration/README.md
DominicFrei: I updated the readme file. Does this answer the questions? If not, I need to further adjust the readme file. My goal would be to answer those questions right there.

examples/ios/swift/Migration/README.md
DominicFrei: I updated the readme file. Does this answer the questions? If not, I need to further adjust the readme file. My goal would be to answer those questions right there.

examples/ios/swift/Migration/README.md
DominicFrei: Correct. I tried to express that by the sub bullet points. Should I clarify further?

leemaguire (approved):
Good job!

examples/ios/swift/Migration/README.md
```diff
@@ -0,0 +1,13 @@
+# Migration Examples
+
+The Migration project shows several examples of migrations and migration blocks.
+
+The purpose of this example is to provide a more in depth view of certain problems and pitfalls users might face when 
+migrations get more complex and the number of versions increases.
+
+## How to use the example
+
+You can build and run the project since this is the default state (the `Migration.xcconfig` is only used to create new examples).
```
jsflax, line 10: This isn't really true? The `Migration.xcconfig` file is so that the user has the option to step through the migration in each "phase" of their schema.

jsflax (commented):
The different versions still need to be tested on the CI. The `build.sh` file in the root directory contains the compilation test. This target will need to be specially with the varying flags.

examples/ios/swift/Migration/README.md
DominicFrei: I thought the newest version would be all a user would use since it migrated all prior version to the most recent version anyway. 0->5, 1->5, etc. (https://github.com/realm/realm-cocoa/blob/df/extend-migration-examples/examples/ios/swift/Migration/AppDelegate.swift#L53)
We do have multiple phases of `migrationBlock` though. Would a user want to step through earlier versions or just use that as a reference to understand how the 5 migration steps in v5 are done?

jsflax: They may want to see 0->1, 0->2, 2->3, 3->4, or any combination. The idea is to cover multiple use cases, for users of the SDK that want to understand basic migrations to complex migrations to partial migrations. Probably not something every user will use, but saying that the config file is only used to create new examples is incorrect

examples/ios/swift/Migration/AppDelegate.swift
```diff
@@ -67,77 +29,50 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
...
+        try! realm.write {
+            exampleData(realm)
+        }
+    }
 
-            // migrate realms at realmv1Path manually, realmv2Path is migrated automatically on access
-            try! Realm.performMigration(for: realmv1Configuration)
+    // Any version before the current versions will be migrated to check if all version combinations work.
+    private func performMigration() {
+        for realmVersion in 0..<schemaVersion {
```
jsflax, line 53: should avoid calling this `realmVersion`, since most people think of realm version as the version of the SDK, not the data base. Would consider `oldSchemaVersion`

examples/ios/swift/Migration/AppDelegate.swift
```diff
@@ -67,77 +29,50 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
...
+        }
+        if usingTemplate {
+            let bundleUrl = Bundle.main.url(forResource: fileName, withExtension: "realm")!
+            try! FileManager.default.copyItem(at: bundleUrl, to: destinationUrl)
         }
 
-        return true
+        return destinationUrl
     }
+
```
jsflax, line 77: extra new line

DominicFrei:
> The different versions still need to be tested on the CI. The `build.sh` file in the root directory contains the compilation test. This target will need to be specially with the varying flags.

@jsflax Added the versions to the build script. This also helped me spot another missing flag in the v5 file (the one for v0 was no in there). 👍

build.sh
```diff
@@ -952,8 +952,11 @@ case "$COMMAND" in
         fi
 
         examples="Simple TableView Migration Backlink GroupedTableView Encryption"
+        versions="0 1 2 3 4 5"
         for example in $examples; do
-            xc "-workspace $workspace -scheme $example -configuration $CONFIGURATION -sdk iphonesimulator build ARCHS=x86_64 ${CODESIGN_PARAMS}"
+            for version in $versions; do
```
jsflax, line 957: Should add a check here to ensure `$example` is `"Migration"` so that we don't build the other examples 6 times when it's not necessary. Something like:
```bash
if [ "$example" = "Migration" ];
    # The migration example needs to be built for each schema version
    # to ensure each compiles 
    for version in $versions; do
        xc "-workspace $workspace -scheme $example -configuration $CONFIGURATION -sdk iphonesimulator build ARCHS=x86_64 ${CODESIGN_PARAMS} OTHER_SWIFT_FLAGS='\$(OTHER_SWIFT_FLAGS) -DSCHEMA_VERSION_$version'"
    done
else
    xc "-workspace $workspace -scheme $example -configuration $CONFIGURATION -sdk iphonesimulator build ARCHS=x86_64 ${CODESIGN_PARAMS}"
fi
```

jsflax (approved):
LGTM– nice!

PR #7084, opened 2021-02-03
Add RLMDictionary / Map Datatype

pavel-ship-it:
This pull request implements `RLMDictionary` / `Map<>`.

This is a key-value pair collection type similar to `NSDictionary` / `Dictionary`.

Values can be set via subscripts like so:
```
    let myMap: Map<String, MyObject> = ...
    myMap["aKey"] = MyObject()
    let foo = myMap["aKey"]
```

Currently only `String` key types are supported for `RLMDictionary` / `Map`. You can store the same types of objects that you would in a `RLMArray` / `List` or `RLMSet` / `MutableSet` in a `RLMDictionary` / `Map`.

`RLMDictionary` / `Map` can be queried, filtered and sorted, returning only the contained objects in a `RLMResults`.


leemaguire:
this should point to the new_data_types branch

pavel-ship-it:
> this should point to the new_data_types branch

It's intentionally pointed to set as new_data_types brings huge diff.

Realm/Realm.h
```diff
@@ -20,6 +20,7 @@
 
 #import <Realm/RLMArray.h>
 #import <Realm/RLMSet.h>
+#import <Realm/RLMDictionary.h>
```
leemaguire, line 23: This needs a type alias in `include/Realm`

leemaguire (commented):
Missing from RLMDictionary:

`(void)enumerateKeysAndObjectsUsingBlock:`
`(nullable ObjectType)objectForKeyedSubscript`
`@property (readonly, copy) NSArray<ObjectType> *allValues;`
`@property (readonly, copy) NSArray<KeyType> *allKeys;`
`- (NSEnumerator<ObjectType> *)objectEnumerator;`
`- (void)removeAllObjects;`
`(void)removeObjectsForKeys:(NSArray<KeyType> *)keyArray`
`(void)setDictionary:(NSDictionary<KeyType, ObjectType> *)otherDictionary;`

Realm/RLMDictionary.h
```diff
@@ -0,0 +1,217 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+////////////////////////////////////////////////////////////////////////////
+
+#ifndef RLMDictionary_h
+#define RLMDictionary_h
```
leemaguire, line 20: We don't need the header guards

Realm/RLMDictionary.h
```diff
@@ -0,0 +1,217 @@
...
+ to be sent to the block. To stop receiving updates, call `-invalidate` on the token.
+
+ @warning This method cannot be called during a write transaction, or when the
+          containing Realm is read-only.
+ @warning This method may only be called on a non-frozen managed set.
+
+ @param block The block to be called each time the set changes.
+ @return A token which must be held for as long as you want updates to be delivered.
+ */
+- (RLMNotificationToken *)addNotificationBlock:(void (^)(RLMDictionary<RLMObjectType> *_Nullable dictionary,
```
leemaguire, line 120: The order of methods in the header should match similarly to `RLMArray`. Methods should also be grouped under `#pragma` marks

Realm/RLMDictionary.mm
```diff
@@ -0,0 +1,133 @@
...
+- (NSUInteger)count {
+    return 0;
+}
+
+// The compiler complains about the method's argument type not matching due to
+// it not having the generic type attached, but it doesn't seem to be possible
+// to actually include the generic type
+// http://www.openradar.me/radar?id=6135653276319744
+#pragma clang diagnostic ignored "-Wmismatched-parameter-types"
+- (nonnull RLMNotificationToken *)addNotificationBlock:(nonnull void (^)(RLMDictionary *, RLMCollectionChange *, NSError *))block {
```
leemaguire, line 37: Order of methods should match that of `RLMArray`

Realm/RLMDictionary_Private.hpp
```diff
@@ -0,0 +1,37 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+////////////////////////////////////////////////////////////////////////////
+
+#ifndef RLMDictionary_Private_hpp
+#define RLMDictionary_Private_hpp
```
leemaguire, line 20: Header guards not required

leemaguire (changes requested):

Realm/RLMDictionary.h
```diff
@@ -37,32 +68,63 @@ NS_ASSUME_NONNULL_BEGIN
  */
 @property (nonatomic, readonly, getter = isFrozen) BOOL frozen;
 
+#pragma mark - Accessing Objects from an Array
```
leemaguire, line 71:
```suggestion
#pragma mark - Accessing Objects from an Dictionary
```

RealmSwift/Map.swift
```diff
@@ -0,0 +1,332 @@
...
+
+import Foundation
+import Realm
+import Realm.Private
+
+/**
+ * Key-value collection. Where the key is a string and value is one of the available Realm types.
+ * We use Map to don't intefere with the native Swift's Dictionary type.
+ */
+public final class Map<Element: RealmCollectionValue>: RLMDictionaryBase {
```
leemaguire, line 17: `public subscript(key: Key) -> Value?` is missing. This subscript will need to handle `get` and `set`

RealmSwift/Map.swift
```diff
@@ -0,0 +1,332 @@
...
+        fatalError("index(of:) is not available on Map")
+    }
+
+    /// :nodoc:
+    public func index(matching predicate: NSPredicate) -> Int? {
+        fatalError("index(matching:) is not available on Map")
+    }
+
+    /**
+     - warning: Ordering is not guaranteed on a Map. `first` is implemented for
```
leemaguire, line 318: I don't think this comment makes sense

RealmSwift/Map.swift
leemaguire: `public var count: Int { return Int(_rlmCollection.count) }` is missing

RealmSwift/RealmCollection.swift
```diff
@@ -196,6 +202,23 @@ private func setType<T>(_ type: T.Type) -> RLMSet<AnyObject> {
...
+    case is Bool.Type:       return RLMDictionary(objectType: .bool, optional: true)
+    case is Float.Type:      return RLMDictionary(objectType: .float, optional: true)
+    case is Double.Type:     return RLMDictionary(objectType: .double, optional: true)
+    case is String.Type:     return RLMDictionary(objectType: .string, optional: true)
+    case is Data.Type:       return RLMDictionary(objectType: .data, optional: true)
+    case is Date.Type:       return RLMDictionary(objectType: .date, optional: true)
+    case is Decimal128.Type: return RLMDictionary(objectType: .decimal128, optional: true)
+    case is ObjectId.Type:   return RLMDictionary(objectType: .objectId, optional: true)
+    case is UUID.Type:       return RLMDictionary(objectType: .UUID, optional: true)
+    default: fatalError("Unsupported type for MutableSet: \(type)?")
```
leemaguire, line 218:
```suggestion
    default: fatalError("Unsupported type for Map: \(type)?")
```

Realm/RLMDictionary.h
```diff
@@ -0,0 +1,280 @@
...
+ coalesced into a single notification. This can include the notification
+ with the initial results. For example, the following code performs a write
+ transaction immediately after adding the notification block, so there is no
+ opportunity for the initial notification to be delivered first. As a
+ result, the initial notification will reflect the state of the Realm after
+ the write transaction.
+
+     Person *person = [[Person allObjectsInRealm:realm] firstObject];
+     NSLog(@"person.dogs.count: %zu", person.dogs.count); // => 0
+     self.token = [person.dogs addNotificationBlock(RLMSet<Dog *> *dogs,
```
leemaguire, line 179:
```suggestion
     self.token = [person.dogs addNotificationBlock(RLMDictionary<Dog *> *dogs,
```

RealmSwift/Map.swift
```diff
@@ -0,0 +1,332 @@
...
+//
+
+import Foundation
+import Realm
+import Realm.Private
+
+/**
+ * Key-value collection. Where the key is a string and value is one of the available Realm types.
+ * We use Map to don't intefere with the native Swift's Dictionary type.
+ */
```
leemaguire, line 16: /**
 `Map` is a key-value storage container used to store supported Realm types.

 `Map` is a generic type that is parameterized on the type it stores. This can be either an `Object`
 subclass or one of the following types: `Bool`, `Int`, `Int8`, `Int16`, `Int32`, `Int64`, `Float`, `Double`,
 `String`, `Data`, `Date`, `Decimal128`, and `ObjectId` (and their optional versions)

`Map` only supports `String` as a key.

 Unlike Swift's native collections, `Map` is a reference types, and are only immutable if the Realm that manages them
 is opened as read-only.

 A `Map` can be filtered and sorted with the same predicates as `Results<Element>`.

 Properties of `Map` type defined on `Object` subclasses must be declared as `let` and cannot be `dynamic`.
 */

RealmSwift/Map.swift
```diff
@@ -0,0 +1,332 @@
...
+public final class Map<Element: RealmCollectionValue>: RLMDictionaryBase {
+
+    // MARK: Properties
+
+    /// The Realm which manages the dictionary, or `nil` if the dictionary is unmanaged.
```
leemaguire, line 21:
```suggestion
    /// The Realm which manages the map, or `nil` if the map is unmanaged.
```

RealmSwift/Map.swift
```diff
@@ -0,0 +1,332 @@
...
+    public func sorted<S>(by sortDescriptors: S) -> Results<Element> where S : Sequence, S.Element == SortDescriptor {
+        fatalError("Not implemented in Map")
+    }
+
+    // MARK: Mutation
+
+    /**
+     Updates the value stored in the dictionary for the given key, or adds a new key-value pair if the key does not exist.
+     */
+    func updateValue(_ value: Element, forKey key: String) {
```
leemaguire, line 123: parameters are missing from the doc

RealmSwift/Map.swift
```diff
@@ -0,0 +1,332 @@
...
+     */
+    public var first: Element? {
+        return self[0]
+    }
+
+    /**
+     - warning: Ordering is not guaranteed on a Map. `last` is implemented for
+                convenience should not be relied on.
+     */
+    public var last: Element? {
```
leemaguire, line 329: you can remove this property

leemaguire (changes requested):

RealmSwift/Map.swift
pavel-ship-it: thanks, fixed

RealmSwift/Map.swift
pavel-ship-it: nice, thanks

RealmSwift/RealmCollection.swift
```diff
@@ -144,6 +144,8 @@ public protocol RealmCollectionValue: Hashable {
     /// :nodoc:
     static func _rlmCollection() -> RLMSet<AnyObject>
     /// :nodoc:
+    static func _rlmCollection() -> RLMDictionary<AnyObject>
```
leemaguire, line 147: pull in latest Set changes, me renaming these to `_rlmCollection` was a mistake

Realm/RLMDictionary.h
```diff
@@ -0,0 +1,286 @@
...
+//
+////////////////////////////////////////////////////////////////////////////
+
+#import <Realm/RLMCollection.h>
+
+NS_ASSUME_NONNULL_BEGIN
+
+/**
+ * Key-value collection. Where the key is a string and value is one of the available Realm types.
+ */
```
leemaguire, line 25: Needs more descriptive doc

RealmSwift/Map.swift
```diff
@@ -0,0 +1,418 @@
...
+    public var count: Int { return Int(_rlmCollection.count) }
+
+    // MARK: KVC
+
+    /**
+     Returns an `Array` containing the results of invoking `valueForKey(_:)` using `key` on each of the collection's
+     objects.
+     */
+    @nonobjc public func value(forKey key: String) -> [AnyObject] {
+        return (rlmDictionary.value(forKeyPath: key)! as! NSSet).allObjects as [AnyObject]
```
leemaguire, line 78: `NSSet `?

RealmSwift/Map.swift
```diff
@@ -0,0 +1,418 @@
...
+        return Results<Element>(_rlmCollection.objects(with: predicate))
+    }
+    
+    /**
+     Returns a Boolean value indicating whether the Map contains the
+     given object.
+
+     - parameter object: The element to find in the Map.
+     */
+    public func contains(_ object: Element) -> Bool {
```
leemaguire, line 120: Set specific

RealmSwift/Map.swift
```diff
@@ -0,0 +1,418 @@
...
+/**
+ Map is a key-value storage container used to store supported Realm types.
+ 
+ Map is a generic type that is parameterized on the type it stores. This can be either an Object
+ subclass or one of the following types: Bool, Int, Int8, Int16, Int32, Int64, Float, Double,
+ String, Data, Date, Decimal128, and ObjectId (and their optional versions)
+ 
+ Map only supports String as a key.
+ 
+ Unlike Swift's native collections, Map is a reference types, and are only immutable if the Realm that manages them
```
leemaguire, line 32:
```suggestion
 Unlike Swift's native collections, Map is a reference type, and are only immutable if the Realm that manages them
```

Realm/RLMAccessor.mm
```diff
@@ -500,7 +524,17 @@ id unmanagedSetter(RLMProperty *prop, const char *) {
...
+        if (prop.array) {
+            cls = [RLMArray class];
+        } else if (prop.set) {
+            cls = [RLMSet class];
+        } else if (prop.dictionary) {
+            cls = [RLMDictionary class];
+        } else {
+            @throw RLMException(@"Invalid collection '%@' for class '%@'.",
+                                propName, obj->_objectSchema.className);
+        }
```
tgoyne, line 537: This shows up in multiple spots and would be worth extracting to a function.

Realm/RLMCollection.mm
```diff
@@ -124,9 +125,19 @@ - (NSUInteger)countByEnumeratingWithState:(NSFastEnumerationState *)state
 
     @autoreleasepool {
         RLMAccessorContext ctx(*_info);
-        for (NSUInteger index = state->state; index < count && batchCount < len; ++index) {
-            _strongBuffer[batchCount] = _results->get(ctx, index);
-            batchCount++;
+
+        if ([_collection isKindOfClass:[RLMDictionary class]]) {
+            NSMutableSet<NSString *> *dictKeys = [NSMutableSet new];
```
tgoyne, line 130: Unused?

Realm/RLMDictionary.h
```diff
@@ -0,0 +1,382 @@
...
+
+/**
+ Adds an array of distinct objects to the set.
+
+ @warning This method may only be called during a write transaction.
+
+ @param objects      `NSDictionary` which contains objects of the
+                    same class as the dictionary.
+ */
+- (void)addObjects:(NSDictionary *)objects;
```
tgoyne, line 138: This is `addEntriesFromDictionary:`.

Realm/RLMManagedDictionary.mm
```diff
@@ -0,0 +1,546 @@
...
+        return keys;
+    });
+}
+
+- (NSArray *)allValues {
+    return translateErrors<RLMManagedDictionary>([&] {
+        NSMutableArray *values = [NSMutableArray array];
+        auto valueResult = _backingCollection.get_values();
+        for (size_t i=0; i<valueResult.size(); i++) {
+            RLMAccessorContext c(*_objectInfo);
```
tgoyne, line 221: Context should be constructed outside of the loop.

Realm/RLMProperty.mm
```diff
@@ -219,12 +223,22 @@ - (BOOL)setTypeFromRawType:(NSString *)rawType {
...
+    static const char dictionaryPrefix[] = "@\"RLMDictionary<";
+    static const int dictionaryPrefixLen = sizeof(dictionaryPrefix) - 1;
+
     static const char numberPrefix[] = "@\"NSNumber<";
     static const int numberPrefixLen = sizeof(numberPrefix) - 1;
 
     static const char linkingObjectsPrefix[] = "@\"RLMLinkingObjects";
     static const int linkingObjectsPrefixLen = sizeof(linkingObjectsPrefix) - 1;
 
+    auto isCollection = [&]() {
```
tgoyne, line 235: This is a misleading name for a function which mutates ivars. It also appears to be more code than just not having it?

Realm/RLMManagedDictionary.mm
```diff
@@ -0,0 +1,546 @@
...
+}
+
+- (void)removeObjectForKey:(id<RLMDictionaryKey>)key {
+    changeDictionary(self, ^{
+        RLMAccessorContext context(*_objectInfo);
+        _backingCollection.erase(keyFromRLMDictionaryKey(key, context));
+    });
+}
+
+inline realm::StringData keyFromRLMDictionaryKey(id<RLMDictionaryKey> key, RLMAccessorContext &context) {
```
tgoyne, line 337: This function doesn't really make any sense for string-only keys and also doesn't fit with the future design for other key types.

Realm/Swift/RLMSupport.swift
```diff
@@ -91,7 +104,7 @@ extension RLMCollection {
      Returns a `RLMCollectionIterator` that yields successive elements in the collection.
      This enables support for sequence-style enumeration of `RLMObject` subclasses in Swift.
      */
-    public func makeIterator() -> RLMCollectionIterator<RLMObject> {
+    public func makeIterator() -> RLMCollectionIterator<AnyObject> {
```
tgoyne, line 107: This is a breaking change.

Realm/TestUtils/include/RLMTestObjects.h
```diff
@@ -21,9 +21,11 @@
 #if __has_extension(objc_generics)
 #define RLM_GENERIC_ARRAY(CLASS) RLMArray<CLASS *><CLASS>
 #define RLM_GENERIC_SET(CLASS) RLMSet<CLASS *><CLASS>
+#define RLM_GENERIC_DICTIONARY(KEY, VALUE) RLMDictionary<KEY, VALUE>
```
tgoyne, line 24:
```suggestion
#define RLM_GENERIC_DICTIONARY(KEY, VALUE) RLMDictionary<KEY *, VALUE *><KEY, VALUE>
```

The purpose of these macros is to define the properties with generic types without repeating the names.

Realm/RLMDictionary.h
pavel-ship-it: it is used in RLMAccessor's unmanagedSetter(RLMProperty *prop, const char *). addEntriesFromDictionary is defined on it's own.

Realm/RLMProperty.mm
pavel-ship-it: true

Realm/TestUtils/include/RLMTestObjects.h
pavel-ship-it: @tgoyne , can we talk about it as I see it works fine?

RealmSwift/MutableSet.swift
```diff
@@ -34,6 +34,7 @@ import Realm.Private
 
  Properties of `MutableSet` type defined on `Object` subclasses must be declared as `let` and cannot be `dynamic`.
  */
+//public final class MutableSet<Element/*: RealmCollectionValue*/>: RLMSwiftCollectionBase {
```
pavel-ship-it, line 37: remove comment

Realm/RLMManagedDictionary.mm
```diff
@@ -0,0 +1,665 @@
...
+                                                     results:_backingCollection.as_results()];
+            return [results valueForKeyPath:keyPath];
+        });
+    }
+    return [super valueForKeyPath:keyPath];
+}
+
+- (id)valueForKey:(NSString *)key {
+    if ([key hasPrefix:@"@"]) {
+        if ([key isEqualToString:[NSString stringWithFormat:@"@%@", RLMInvalidatedKey]]) {
```
tgoyne, line 441: This is an awfully roundabout way of writing @"invalidated"

Realm/RLMManagedDictionary.mm
```diff
@@ -0,0 +1,665 @@
...
+            return @(!_backingCollection.is_valid());
+        }
+        return [super managedValueForKey:[key substringFromIndex:1]];
```
tgoyne, line 444: Why is this calling NSObject's valueForKey on the thing after the @?

RealmSwift/Map.swift
```diff
@@ -0,0 +1,610 @@
...
+    public subscript(key: Key) -> Value? {
+        get {
+            if rlmDictionary.type == .object {
+                let obj = rlmDictionary[objcKey(from: key)]
+                // A Map can keep the key of an object that has been deleted by the Realm.
+                // If the object is deleted it will be stored as NSNull.null so we want to
+                // return that as `nil`.
+                if obj is NSNull {
+                    return nil
+                }
```
tgoyne, line 121: This is incorrectly conflating "value for this key is null" and "dictionary does not have an entry for this key".  There shouldn't be any need to special-case object here at all; the only thing different about object from other types is that object-valued Maps must be optional, and the handling of option values of any type should be the same.

RealmSwift/RealmCollection.swift
```diff
@@ -182,16 +215,12 @@ extension Optional: RealmCollectionValue where Wrapped: RealmCollectionValue {
 }
 
 /// :nodoc:
-public protocol RealmCollectionBase: RandomAccessCollection, LazyCollectionProtocol, CustomStringConvertible, ThreadConfined where Element: RealmCollectionValue {
-    // This typealias was needed with Swift 3.1. It no longer is, but remains
-    // just in case someone was depending on it
-    typealias ElementType = Element
```
tgoyne, line 188: Removing this is a breaking change

RealmSwift/Tests/SwiftTestObjects.swift
```diff
@@ -78,6 +78,7 @@ class SwiftObject: Object {
 
     let arrayCol = List<SwiftBoolObject>()
     let setCol = MutableSet<SwiftBoolObject>()
+    let mapCol = Map<String, SwiftBoolObject>()
```
tgoyne, line 81: This isn't a valid property declaration and should be throwing an exception complaining that object values need to be optional.

Realm/RLMDictionary.h
```diff
@@ -0,0 +1,416 @@
...
+////////////////////////////////////////////////////////////////////////////
+
+#import <Realm/RLMCollection.h>
+
+NS_ASSUME_NONNULL_BEGIN
+
+@class RLMObject, RLMResults<RLMObjectType>, RLMDictionaryChange;
+
+/// A protocol defining a key type for RLMDictionary.
+@protocol RLMDictionaryKey <NSCopying>
```
tgoyne, line 26: This isn't really how obj-c protocols work and doing this just makes the obj-c API a lot more obnoxious to call from swift without providing any additional safety.

Realm/RLMManagedDictionary.mm
leemaguire: I was trying to copy the behaviour of NSDictionary here. 

`If key does not start with “@”, invokes objectForKey:. If key does start with “@”, strips the “@” and invokes [super valueForKey:] with the rest of the key.`. 

So without the `@` prefix the key is assumed to be a value in the collection, and it might be possible someone wants to use `invalidated` as a key for their value.

RealmSwift/RealmCollection.swift
leemaguire: Should we just deprecate it then?

Realm/RLMDictionary.h
leemaguire: Are you talking about the protocol as a whole or the `NSCopying` part?

Realm/RLMManagedDictionary.mm
leemaguire: I was trying to copy the behaviour of NSDictionary here.

If key does not start with “@”, invokes objectForKey:. If key does start with “@”, strips the “@” and invokes [super valueForKey:] with the rest of the key..

So without the @ prefix the key is assumed to be a value in the collection, and it might be possible someone wants to use invalidated as a key for their value.

Realm/RLMManagedDictionary.mm
leemaguire: I want to forward the kvc call to the class that won't invoke pulling an object from the collection. The only way I seen fit to do that was to use the super of RLMDictionary.

RealmSwift/Object.swift
```diff
@@ -262,6 +262,29 @@ extension Object: RealmCollectionValue {
...
+     components that integrate with Realm. If you are simply building an app on Realm, it is
+     recommended to use instance variables or cast the values returned from key-value coding.
+
+     - parameter propertyName: The name of the property.
+
+     - returns: A map of  `String` with `DynamicObject`s.
+
+     :nodoc:
+     */
+    public func dynamicMap<Key: _MapKey>(_ propertyName: String) -> Map<Key, DynamicObject> {
```
jsflax, line 280: I'm not entirely sure what the point of this method is? Could you explain what the use case would be for this to be in the public API?

RealmSwift/Tests/MapTests.swift
```diff
@@ -0,0 +1,858 @@
...
+
+fileprivate extension Map {
+    func addTestObjects(from dictionary: [Key: Value]) where Key: Hashable {
+        dictionary.forEach { (k, v) in
+            self[k] = v
+        }
+    }
+}
+
+class MapTests<M: RealmKeyedCollection, EM: RealmKeyedCollection>: TestCase where M.Key == String,
```
jsflax, line 30: What is the point of this being generic with all of the constraints baked in like this?

RealmSwift/Tests/SwiftTestObjects.swift
leemaguire: Does this apply to embedded objects too?

RealmSwift/Tests/MapTests.swift
leemaguire: So I could test against a normal `Map<String, SwiftStringObject?>` collection and a `AnyMap()` collection

Realm/RLMDictionary.mm
```diff
@@ -0,0 +1,559 @@
...
+- (NSUInteger)indexOfObject:(id)value {
+    return [_backingCollection.allValues indexOfObject:value];
+}
+
+#pragma clang diagnostic pop // unused parameter warning
+
+#pragma mark - Thread Confined Protocol Conformance
+
+- (realm::ThreadSafeReference)makeThreadSafeReference {
+    REALM_TERMINATE("Unexpected handover of unmanaged `RLMDictionary`");
```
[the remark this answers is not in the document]
leemaguire: This is the same message as RLMArray and RLMSet. I think its self explanatory but maybe changing 'Unexpected' to 'Unsupported' might work better?

Realm/RLMDictionary.h
tgoyne: The whole protocol.

Realm/RLMManagedDictionary.mm
tgoyne: But why are you trying to forward the KVC call to the class? That's not a normal thing for obj-c collection types to do and it's not obvious why it'd be useful. The logic for @invalidated is just so that you can use KVO to observe that property.

Realm/RLMDictionary.h
tgoyne: The reason why maps of objects have to be optional doesn't seem to apply to embedded objects, so it'd make sense to me to allow non-optional embedded objects but I assume that'll require some objectstore changes.

Realm/RLMDictionary.mm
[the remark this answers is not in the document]
tgoyne: This is REALM_TERMINATE and not an exception because it's not possible to hit this code path unless you've bypassed the public API. There's a (hopefully) better error message that tries to explain what's wrong at the public API boundary.

Realm/RLMManagedDictionary.mm
```diff
@@ -0,0 +1,665 @@
...
+                                                        parentInfo:parentInfo
+                                                          property:parentInfo->rlmObjectSchema[metadata.key]];
+}
+
+static RLMNotificationToken *RLMAddNotificationBlock(RLMManagedDictionary *collection,
+                                                     void (^block)(id, RLMDictionaryChange *, NSError *),
+                                                     dispatch_queue_t queue) {
+    RLMRealm *realm = collection.realm;
+    if (!realm) {
+        @throw RLMException(@"Linking objects notifications are only supported on managed objects.");
```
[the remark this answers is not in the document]
leemaguire: This message is borrowed from RLMCollection. So I presume this message should be standard across all collections.

Realm/RLMUtil.mm
```diff
@@ -437,10 +465,12 @@ id RLMMixedToObjc(realm::Mixed const& mixed,
         case realm::type_ObjectId:
             return [[RLMObjectId alloc] initWithValue:mixed.get<realm::ObjectId>()];
         case realm::type_TypedLink:
-            return RLMObjectFromObjLink(realm, mixed.get<realm::ObjLink>(), parentIsSwiftObject);
+            return RLMObjectFromObjLink(realm, mixed.get<realm::ObjLink>(), classInfo->isSwiftClass());
         case realm::type_Link:
-        case realm::type_LinkList:
-            REALM_UNREACHABLE();
+        case realm::type_LinkList: {
```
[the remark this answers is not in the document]
leemaguire: No, you caught a bug.

Realm/RLMAccessor.mm
```diff
@@ -497,14 +532,26 @@ id unmanagedSetter(RLMProperty *prop, const char *) {
...
 
-        Class cls = prop.array ? [RLMArray class] : [RLMSet class];
+        Class cls = RLMCollectionClassForProperty(prop, false);
         id collection;
             // make copy when setting (as is the case for all other variants)
-        if (prop.type == RLMPropertyTypeObject)
-            collection = [[cls alloc] initWithObjectClassName:prop.objectClassName];
+        if (prop.type == RLMPropertyTypeObject) {
+            if (prop.dictionary)
+                collection = [[cls alloc] initWithObjectClassName:prop.objectClassName keyType:prop.dictionaryKeyType];
```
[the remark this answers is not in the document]
leemaguire: is that not just shifting one code smell from one place to another?

Realm/RLMCollection_Private.hpp
```diff
@@ -75,7 +83,16 @@ NSUInteger RLMFastEnumerate(NSFastEnumerationState *state, NSUInteger len, id<RL
 - (instancetype)initWithChanges:(realm::CollectionChangeSet)indices;
 @end
 
+@interface RLMCancellationToken : RLMNotificationToken {
```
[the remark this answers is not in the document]
leemaguire: I shifted it here because RLMManagedDictionary.mm references it

Realm/RLMCollection.mm
```diff
@@ -75,6 +78,40 @@ - (instancetype)initWithBackingCollection:(realm::object_store::Collection const
...
+            _snapshot = backingDictionary.get_keys().snapshot();
+            for (id key in dictionary.allKeys) {
+                _dictionaryKeys.push_back(key);
+            }
+        }
+        else {
+            _snapshot = backingDictionary.get_keys();
+            for (id key in dictionary.allKeys) {
+                _dictionaryKeys.push_back(key);
+            }
```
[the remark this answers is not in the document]
leemaguire: We need to keep a stable set of the keys for the enumeration, for the case that mutation happens during it. Snapshotting a Results of dictionary does nothing useful so we can't use that.

Realm/RLMDictionary.h
```diff
@@ -0,0 +1,409 @@
...
+
+/**
+ A `RLMDictionaryChange` object encapsulates information about changes to dictionaries
+ that are reported by Realm notifications.
+
+ `RLMDictionaryChange` is passed to the notification blocks registered with
+ `-addNotificationBlock` on `RLMDictionary`, and reports what keys in the
+ dictionary changed since the last time the notification block was called.
+ */
+@interface RLMDictionaryChange : NSObject
```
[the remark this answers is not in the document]
leemaguire: This is a limitation in core. Deletions are thrown back as indices and not keys. Which doesn't make much sense on the cocoa side. Dotnet has the same approach https://github.com/realm/realm-dotnet/blob/master/Realm/Realm/Linq/DictionaryChangeSet.cs

Realm/RLMProperty.mm
```diff
@@ -265,53 +280,57 @@ - (BOOL)setTypeFromRawType:(NSString *)rawType {
...
+        _objectClassName = [[NSString alloc] initWithBytes:code + prefixLen
+                                                    length:strlen(code + prefixLen) - 2 // drop trailing >"
                                                   encoding:NSUTF8StringEncoding];
 
         if ([RLMSchema classForString:_objectClassName]) {
-            _optional = false;
+            // Dictionaries require object types to be nullable. This is due to
+            // the fact that if you delete a realm object that exists in a dictionary
+            // the key should stay present but the value should be null.
+            _optional = _dictionary ? true : false;
```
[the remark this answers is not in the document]
leemaguire: 🥸

Realm/Swift/RLMSupport.swift
```diff
@@ -65,34 +65,79 @@ extension RLMObject {
...
+extension RLMArray: Sequence, _RLMCollectionIterator { }
+extension RLMDictionary: Sequence, _RLMDictionaryIterator {}
+extension RLMSet: Sequence, _RLMCollectionIterator {}
+extension RLMResults: Sequence, _RLMCollectionIterator {}
 
 /**
  This struct enables sequence-style enumeration for RLMObjects in Swift via `RLMCollection.makeIterator`
  */
-public struct RLMCollectionIterator<T>: IteratorProtocol {
+public struct RLMCollectionIterator: IteratorProtocol {
```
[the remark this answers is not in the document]
leemaguire: On master this is only specialised for RLMObject, so I was under the impression that primitives were never supported

Realm/RLMCollection.mm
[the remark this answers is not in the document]
tgoyne: Mutating a collection from inside a call to enumerateKeysAndObjectsUsingBlock is invalid and is allowed to break. We iterate over a snapshot for fast enumeration because live queries make it so that mutations the objects inside the collection can remove them from the query and indirectly mutate the collection being iterated, but that's not applicable here.

Realm/RLMAccessor.mm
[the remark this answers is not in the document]
tgoyne: It's shifting a code smell from a whole bunch of places to just two.

Realm/RLMDictionary.h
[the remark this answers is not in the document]
tgoyne: That needs to be fixed, then. Only having insertions and modifications is sufficiently useless that it's probably better to just not have change notifications at all until we can make it work.

Realm/Swift/RLMSupport.swift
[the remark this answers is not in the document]
tgoyne: Yeah, it looks like making this generic was an incomplete attempt at making iterating collections from swift give the correct RLMObject subclass and not something that ever worked.

Realm/RLMDictionary.h
```diff
@@ -0,0 +1,442 @@
...
+- (nullable NSNumber *)averageOfProperty:(NSString *)property;
+
+#pragma mark - Notifications
+
+/**
+ Registers a block to be called each time the dictionary changes.
+
+ The block will be asynchronously called with the initial dictionary, and then
+ called again after each write transaction which changes any of the key-value in
+ the dictionary or which objects are in the results.
```
tgoyne, line 294: This looks like an incorrect find/replace on the docs from the array version and isn't a grammatically correct sentence or an accurate description of the behavior.

Realm/RLMDictionary.h
```diff
@@ -0,0 +1,442 @@
...
+ coalesced into a single notification. This can include the notification
+ with the initial results. For example, the following code performs a write
+ transaction immediately after adding the notification block, so there is no
+ opportunity for the initial notification to be delivered first. As a
+ result, the initial notification will reflect the state of the Realm after
+ the write transaction.
+
+     Person *person = [[Person allObjectsInRealm:realm] firstObject];
+     NSLog(@"person.dogs.count: %zu", person.dogs.count); // => 0
+     self.token = [person.dogs addNotificationBlock(RLMDictionary<NSString *, Dog *><RLMString, Dog> *dogs,
```
tgoyne, line 317: This is missing a colon.

Realm/RLMDictionary.mm
```diff
@@ -0,0 +1,541 @@
...
+    return _backingCollection.count;
+}
+
+- (NSArray *)allKeys {
+    return _backingCollection.allKeys;
+}
+
+- (NSArray *)allValues {
+    return _backingCollection.allValues;
+}
```
tgoyne, line 157: These need a `?: @[]` at the end so that we return an empty array rather than nil if _backingCollection isn't initialized yet.

Realm/RLMDictionary.mm
```diff
@@ -0,0 +1,541 @@
...
+
+- (nullable id)objectForKey:(id)key {
+    return [_backingCollection objectForKey:key];
+}
+
+- (nullable id)objectForKeyedSubscript:(id)key {
+    return [_backingCollection objectForKey:key];
+}
```
tgoyne, line 165: These need to initialize _backingCollection if it isn't as `[@[] valueForKey:...]` has different results depending on the key.

Realm/RLMDictionary.mm
```diff
@@ -0,0 +1,541 @@
...
+        return @"<Maximum depth exceeded>";
+    }
+
+    const NSUInteger maxObjects = 100;
+    auto str = [NSMutableString stringWithFormat:@"%@<%@, %@> <%p> (\n", name,
+                RLMTypeToString([dictionary keyType]),
+                [dictionary objectClassName] ?: RLMTypeToString([dictionary type]),
+                (void *)dictionary];
+    size_t index = 0, skipped = 0;
+    for (id key in dictionary.allKeys) {
```
tgoyne, line 502: `.allKeys` is redundant here.

Realm/RLMManagedDictionary.mm
```diff
@@ -0,0 +1,688 @@
...
+            for (auto index : _changes.deletions) {
+                realm::Mixed key = dictionary.get_key(index);
+                switch (key.get_type()) {
+                    case realm::type_String:
+                        [array addObject:@(key.get_string().data())];
+                        break;
+                    default:
+                        // Don't throw so older SDK versions can handle any new key types.
+                        break;
+                }
```
tgoyne, line 120: Using RLMMixedToObjc would be both simpler and more forward-compatible.

Realm/RLMManagedDictionary.mm
```diff
@@ -0,0 +1,688 @@
...
+
+- (NSUInteger)count {
+    return translateErrors([&] {
+        return _backingCollection.size();
+    });
+}
+
+- (NSArray *)allKeys {
+    return translateErrors([&] {
+        NSMutableArray<id> *keys = [NSMutableArray array];
```
tgoyne, line 320: `[NSMutableArray arrayWithCapacity:keyResult.size()]`

Realm/RLMManagedDictionary.mm
```diff
@@ -0,0 +1,688 @@
...
+                                  objects:(__unused __unsafe_unretained id [])buffer
+                                    count:(NSUInteger)len {
+    return RLMFastEnumerate(state, len, self);
+}
+
+#pragma mark - Object Retrieval
+
+- (nullable id)objectForKey:(id)key {
+    return translateErrors([&] {
+        [self.realm verifyThread];
```
tgoyne, line 368: `try_get_any()` should be checking the thread

Realm/RLMObjectStore.mm
```diff
@@ -20,6 +20,7 @@
 
 #import "RLMAccessor.hpp"
 #import "RLMArray_Private.hpp"
+#import "RLMDictionary_Private.hpp"
```
tgoyne, line 23: Probably shouldn't need to add this if it's the only change in the file?

Realm/RLMUtil.hpp
```diff
@@ -20,6 +20,8 @@
 #import <Realm/RLMSwiftValueStorage.h>
 #import <Realm/RLMValue.h>
 
+#import "RLMClassInfo.hpp"
```
tgoyne, line 23: This could be just a forward declaration.

RealmSwift/Combine.swift
```diff
@@ -1237,7 +1369,7 @@ public enum RealmPublishers {
         }
     }
 
-    /// A publisher which emits RealmCollectionChange<T> each time the observed object is modified
+    /// A publisher which emits collection change each time the observed object is modified
```
tgoyne, line 1372: Incorrect find/replace?

RealmSwift/RealmCollection.swift
```diff
@@ -45,6 +45,39 @@ import Realm
     }
 }
 
+/// A container used for storing key-value pairs which are to be used as elements in `RLMMapIterator`.
```
tgoyne, line 48: This appears to be something which should be `:nodoc:` as it's just an implementation detail that users shouldn't know about and would be internal if Swift let us.

RealmSwift/RealmKeyedCollection.swift
```diff
@@ -0,0 +1,856 @@
...
+    /// Returns the number of key-value pairs in this map.
+    var count: Int { get }
+
+     /// A human-readable description of the objects contained in the Map.
+    var description: String { get }
+
+    // MARK: Mutation
+
+    /**
+     Updates the value stored in the dictionary for the given key, or adds a new key-value pair if the key does not exist.
```
tgoyne, line 48: Each of these should note that they require a write transaction for managed dictionaries. Things which can add values to the dictionary should note that if the value being added is an unmanaged object the object will be added to the realm.

RealmSwift/RealmKeyedCollection.swift
```diff
@@ -0,0 +1,856 @@
...
+
+     - parameter value: The object value.
+     - parameter key:   The name of the property whose value should be set on each object.
+    */
+    func setValue(_ value: Any?, forKey key: String)
+
+    // MARK: Filtering
+
+    /**
+     Returns a `Results` containing all matching key-value pairs the given predicate in the Map.
```
tgoyne, line 105: This doesn't return `Results<Key, Value>`

RealmSwift/RealmKeyedCollection.swift
```diff
@@ -0,0 +1,856 @@
...
+    /**
+     Returns the object at the given `position`.
+
+     - parameter position: The position of the element in the collection.
+     */
+    subscript(position: MapIndex) -> (key: Key, value: Value) { get }
+
+    /// Returns the position of an element in the collection.
+    /// - Parameter object: The object to find.
+    func index(of object: Value) -> MapIndex?
```
tgoyne, line 172: Do we actually need to expose these? If we do the docs need a bunch of caveats about how dictionary orders are unstable and not the insertion order.

RealmSwift/RealmKeyedCollection.swift
```diff
@@ -0,0 +1,856 @@
...
+     - warning: Only a property whose type conforms to the `AddableType` protocol can be specified.
+
+     - parameter property: The name of a property whose values should be summed.
+     */
+    func average<T: AddableType>(ofProperty property: String) -> T?
+
+    // MARK: Notifications
+
+    /**
+     Registers a block to be called each time the dictionary changes.
```
tgoyne, line 218: There's a bunch of wording changes needed here and the example isn't even using a dictionary.

RealmSwift/RealmKeyedCollection.swift
```diff
@@ -0,0 +1,856 @@
...
+
+    override func freeze() -> AnyMap<C.Key, C.Value> { return AnyMap(base.freeze()) }
+    override func thaw() -> AnyMap<C.Key, C.Value>? { return AnyMap(base.thaw()!)  }
+}
+/**
+ A type-erased `RealmKeyedCollection`.
+
+ Instances of `RealmKeyedCollection` forward operations to an opaque underlying collection having the same `Key`, `Value` type.
+ */
+public struct AnyMap<Key: _MapKey, Value: RealmCollectionValue>: RealmKeyedCollection {
```
tgoyne, line 532: We only have one RealmKeyedCollection type so there doesn't appear to be anything this would be useful for?

RealmSwift/Map.swift
```diff
@@ -0,0 +1,600 @@
...
+    static var _rlmType: RLMPropertyType { get }
+}
+extension String: _MapKey { }
+
+/**
+ Map is a key-value storage container used to store supported Realm types.
+ 
+ Map is a generic type that is parameterized on the type it stores. This can be either an Object
+ subclass or one of the following types: Bool, Int, Int8, Int16, Int32, Int64, Float, Double,
+ String, Data, Date, Decimal128, and ObjectId (and their optional versions)
```
tgoyne, line 34: This should note that their optional versions are only supported in non-synchronized Realms.

RealmSwift/Map.swift
```diff
@@ -0,0 +1,600 @@
...
+
+    // MARK: Count
+
+    /// Returns the number of key-value pairs in this map.
+    @objc public var count: Int { return Int(_rlmCollection.count) }
+
+    // MARK: Mutation
+
+    /**
+     Updates the value stored in the map for the given key, or adds a new key-value pair if the key does not exist.
```
tgoyne, line 86: I noted some things in the docs for RealmKeyedCollection that apply to this file as well.

RealmSwift/Map.swift
```diff
@@ -0,0 +1,600 @@
...
+        return _rlmCollection.isFrozen
+    }
+
+    public func freeze() -> Map {
+        return Map(objc: rlmDictionary.freeze())
+    }
+
+    public func thaw() -> Map? {
+        return Map(objc: rlmDictionary.thaw())
+    }
```
tgoyne, line 393: Missing docs

RealmSwift/Map.swift
```diff
@@ -0,0 +1,600 @@
...
+     `.initial` indicates that the initial run of the query has completed (if
+     applicable), and the collection can now be used without performing any
+     blocking work.
+     */
+    case initial(Collection)
+
+    /**
+     `.update` indicates that a write transaction has been committed which
+     either changed which objects are in the collection, and/or modified one
+     or more of the objects in the collection.
```
tgoyne, line 530: "which keys are in the collection, or the values of the objects for those keys" or something like that.

RealmSwift/Map.swift
```diff
@@ -0,0 +1,600 @@
...
+
+     All three of the change arrays are always sorted in ascending order.
```
tgoyne, line 532: I don't think this is true for keyed changes.

RealmSwift/Map.swift
```diff
@@ -0,0 +1,600 @@
...
+        return lhs.value == rhs.value
+    }
+    /// :nodoc:
+    public func hash(into hasher: inout Hasher) {
+        hasher.combine(key)
+    }
+    /// :nodoc:
+    public var key: Self.Key
+    /// :nodoc:
+    public var value: Self.Value
```
tgoyne, line 590: These don't need very interesting docs but we do want these properties to appear in the generated docs.

include/module.modulemap
```diff
@@ -11,11 +11,13 @@ module Realm {
         header "Realm/RLMApp_Private.h"
         header "Realm/RLMArray_Private.h"
         header "Realm/RLMCollection_Private.h"
+        header "Realm/RLMDictionary_Private.h"
         header "Realm/RLMObject_Private.h"
         header "Realm/RLMObjectBase_Dynamic.h"
         header "Realm/RLMObjectBase_Private.h"
         header "Realm/RLMObjectSchema_Private.h"
         header "Realm/RLMObjectStore.h"
+        header "Realm/RLMObject_Private.h"
```
tgoyne, line 20: Repeated include

RealmSwift/RealmKeyedCollection.swift
leemaguire: True and I'm all for code removal.

RealmSwift/RealmKeyedCollection.swift
leemaguire: No, I added them because they are a feature of Foundation, but I was hard strung to find real world examples of where they are actually used. I'll remove.

Realm/RLMManagedDictionary.mm
leemaguire: My rational was they it would add more overhead

Realm/RLMManagedDictionary.mm
leemaguire: It doesn't seem to throw where it should without it

Realm/RLMDictionary.h
```diff
@@ -0,0 +1,442 @@
...
+
+/**
+ Returns the minimum (lowest) value of the given property among all the values in the dictionary.
+
+     NSNumber *min = [object.dictionaryProperty minOfProperty:@"age"];
+
+ @warning You cannot use this method on `RLMObject`, `RLMDictionary`, and `NSData` properties.
+
+ @param property The property whose minimum value is desired. Only properties of
+                 types `int`, `float`, `double`, and `NSDate` are supported.
```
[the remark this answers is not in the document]
leemaguire: This doc was taken from RLMArray, does this imply RLMArray is wrong too? Also it seems like Decimal128 is the only supported one left out. I don't really get stating 'You cannot use this method on' I think it would be better just to tell people what they can use it on?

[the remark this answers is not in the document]
tgoyne: There's probably a whole pile of stale lists of types scattered around. Explicitly listing the types it works on and having it be implied that everything else doesn't work is probably the way to go. I think NSData was originally called out specifically in some places because there were some operations you'd expect to work that didn't but that was then fixed.

Realm/RLMQueryUtil.mm
```diff
@@ -1400,6 +1650,37 @@ void get_collection_type(__unsafe_unretained RLMProperty *prop, Fn&& fn) {
         case CollectionOperation::Average:
             add_collection_operation_constraint<CollectionOperation::Average>(operatorType, collectionOperation, rhs);
             break;
+        case CollectionOperation::AllKeys: {
+            // BETWEEN and IN are not supported by @allKeys as the parsing for collection
+            // operators happens before and disection of a rhs array of values.
```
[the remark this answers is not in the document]
leemaguire: basically support for `@allKeys BETWEEN @[...]` is not supported for these collection operators, due to the sequence the calls constructing the query. if the RHS is an array, we do not handle it. The RHS of array types is handled later on in the code by other supported predicates

PR #7090, opened 2021-02-03
Add Xcode 12.5 to PR matrix and fix running tests on macOS 11

tgoyne:
Xcode 12.5 has changed when the ProcessXCFramework build step runs and it's no longer possible to download the library in a build step as a result. Downloading it as a scheme pre-build action fixes this and removes the need for weird workarounds, but at the cost of duplicating the build command in each scheme and errors in pre-build steps don't get displayed anywhere in the Xcode UI.

Xcode 11's implementation of main-thread checking doesn't work on macOS 11 and just makes the program crash on startup. While it's never actually caught anything for us (since we don't have much in the way of UI code...) we'd still like to have it on when building with Xcode 12, which is a bit awkward as it's a strange Schema setting and not something that can be set on the xcodebuild command line. Switching to using test plans appears to be the least-bad way to achieve this, although it's still pretty awkward. For installation examples I just disabled main-thread checking entirely as those aren't testing anything where it could be useful.

Setting the test plan in build.sh lead to lots quoting issues that I fixed by just fixing (almost) all of the things which shellcheck complained about and switching to passing around arrays of arguments correctly. The main functional change is in the xctest function and everything else is just making shellcheck happy.

The check for lingering files between CI runs never actually worked, isn't useful as we delete and recreate simulators between each run anyway, and when made to work fails because Shortcuts.realm exists in a newly-created simulator.


jsflax (approved):
LGTM, though there are some compilation errors on 12.5 around Identifiable that I thought would have borked a few tests as well

tgoyne:
I haven't even tried to build with 12.5 locally yet and this is just to test if CI is working correctly with macOS 11.

jsflax:
Fair enough, LGTM then

jsflax:
This has changed pretty significantly since I reviewed it, so I'm re-requesting myself.

tgoyne:
I think the remaining failures are now unrelated things.

jsflax (commented):
So does using xctestplans incidentally enable both PreActions and the ability to disable the mainThreadChecker? Just trying to understand the impetus behind using them.

tgoyne:
No, pre-build actions are unrelated. Conditionally disabling the main-thread checker is the only thing they're being used for here.

jsflax (approved):
Looks good, great job solving this.

PR #7094, opened 2021-02-04
Tests for new migration from RLMObject to RLMEmbeddedObject

### LINKED ISSUE #7060: Unable to execute a migration that changes the "embeddedness" of an object
## Goals
A user has reported on the [forums](https://developer.mongodb.com/community/forums/t/migrating-to-using-the-new-embedded-object-in-mongodb-realm/11920) that they would like to be able to migrate a model from `Object` to `EmbeddedObject`. 

## Expected Results
Migration function gets executed allowing them to ensure each object only has one parent, after which the "embeddedness" of the table changes and the model is now treated as embedded.

## Actual Results
App crashes with `Cannot convert object type ‘MyObject’ to embedded because objects have multiple incoming links*.`. Based on a conversation with @jsflax and @tgoyne, it appears the reason is that the SDK eagerly changes the table to embedded before running the migration function, which then triggers a guard at Core that ensures that only one parent points to the child object.

## Steps for others to Reproduce
Don't have exact steps, but seems like:
1. Create a model as inheriting from `Object`.
2. Have several other objects link to it.
3. Bump schema version and modify your model to inherit from `EmbeddedObject`.

Realm framework version: 10.5.0

DominicFrei:
Tests on our side for the changed `embedded table` functionality introduced by https://github.com/realm/realm-core/pull/4414.


Realm/Tests/MigrationTests.mm
```diff
@@ -1398,6 +1398,390 @@ - (void)testModifyPrimaryKeyInMigration {
     }
 }
 
+- (void)testChangeEmptyTableFromTopLevelToEmbedded {
+    RLMProperty *intProperty = [[RLMProperty alloc] initWithName:@"intProperty"
```
tgoyne, line 1402: Rather than manually defining the base schema, this should just define a normal subclass and use that for the initial schema (or reuse an existing type if possible). The general pattern these tests should take is obtain schema from classes with `schemaForObjectClass:`, mutate a second copy of the schema to match the thing being tested, then call `assertMigrationRequiredForChangeFrom:to:` which initializes the Realm with the initial schema, verifies that the schema version has to be bumped to migrate to the new schema, and then verifies that the tables are in the correct state after migrating to the new schema.

The tests which actually have a migration block should use `createTestRealmWithSchema :` + `migrateTestRealmWithBlock:`

Realm/Tests/MigrationTests.mm
```diff
@@ -1398,6 +1398,390 @@ - (void)testModifyPrimaryKeyInMigration {
...
+    realmConfiguration.schemaVersion = 1;
+    __block int parentEnumerateCalls = 0;
+    __block int childEnumerateCalls = 0;
+    realmConfiguration.migrationBlock = ^(RLMMigration *migration, uint64_t) {
+        [migration enumerateObjects:@"ParentClass" block:^(RLMObject *oldObject, RLMObject *newObject) {
+            parentEnumerateCalls++;
+            XCTAssertNotNil(oldObject);
+            XCTAssertNotNil(newObject);
+            XCTAssertEqual(oldObject[@"childProperty"][@"intProperty"], newObject[@"childProperty"][@"intProperty"]);
+            XCTAssert([newObject[@"childProperty"][@"intProperty"] intValue] == 42 || [newObject[@"childProperty"][@"intProperty"] intValue] == 43);
```
tgoyne, line 1660: This should read the value to a local variable rather than duplicating all that stuff.

Realm/Tests/MigrationTests.mm
```diff
@@ -1398,6 +1398,390 @@ - (void)testModifyPrimaryKeyInMigration {
...
+        }];
+        [migration enumerateObjects:@"ChildClass" block:^(RLMObject *oldObject, RLMObject *newObject) {
+            childEnumerateCalls++;
+            XCTAssertNotNil(oldObject);
+            XCTAssertNotNil(newObject);
+            XCTAssertEqual([oldObject[@"intProperty"] intValue], [newObject[@"intProperty"] intValue]);
+            XCTAssert([newObject[@"intProperty"] intValue] == 42);
```
tgoyne, line 1667: XCTAssertEqual

PR #7098, opened 2021-02-05
Add @available attribute to ObjectKeyIdentifiable

### LINKED ISSUE #7083: Xcode 12.5 beta - Swift Compiler Error for RealmSwift module
## Goals
Build Realm on Xcode 12.5 beta

## Expected Results
To be able to build project

## Actual Results
Got a Swift Compiler Error for the RealmSwift module in Combine.swift
with the error being that 'Identifiable' is only available in iOS 13 or newer.

## Steps for others to Reproduce
Add Realm and RealmSwift modules via SPM in Xcode 12.5 beta
Build project

Realm framework version: 10.5.1

Realm Object Server version: ?

Xcode version: 12.5 beta

iOS/OSX version: macOS 11.2

Dependency manager + version: SPM (also tried with same error with Cocoapods)

jsflax:
Fixes #7083.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,9 +17,8 @@ x.y.z Release notes (yyyy-MM-dd)
 * Allow enumerating objects of types which are no longer present in the schema.
 * Add `RLMResponse.customStatusCode`. This fixes timeout exceptions that were occuring with a poor connection. ([#4188](https://github.com/realm/realm-core/issues/4188))
 * Fix - `RLMResponse` will have a non nil `customStatusCode` in case of error.
-   ([#4188](https://github.com/realm/realm-core/issues/4188))
-* Make ObjectKeyIdentifiable available for macOS 10.15, iOS 13, tvOS 13, watchOS 6 and above
-
+([#4188](https://github.com/realm/realm-core/issues/4188))
+* Limit availability of ObjectKeyIdentifiable to platforms which support Combine to match the change made in the Xcode 12.5 SDK. ([#7083](https://github.com/realm/realm-cocoa/issues/7083))
 <!-- ### Breaking Changes - ONLY INCLUDE FOR NEW MAJOR version -->
 
 ### Compatibility
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,15 +9,16 @@ x.y.z Release notes (yyyy-MM-dd)
 * Add Xcode 12.4 binaries to the release package.
 
 ### Fixed
-* Inserting a date into a synced collection via `AnyBSON.datetime(...)` would be of type `Timestamp` and not `Date`. 
+* Inserting a date into a synced collection via `AnyBSON.datetime(...)` would be of type `Timestamp` and not `Date`.
   This could break synced objects with a `Date` property  ([#6654](https://github.com/realm/realm-cocoa/issues/6654), since v10.0.0).
 * Fixed an issue where creating an object after file format upgrade may fail
   with assertion "Assertion failed: lo() <= std::numeric_limits<uint32_t>::max()"
  ([#4295](https://github.com/realm/realm-core/issues/4295), since v5.0.0)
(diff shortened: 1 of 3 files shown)


RealmSwift/Combine.swift
```diff
@@ -35,15 +35,18 @@ import Realm.Private
 ///
 /// You can also manually conform to `Identifiable` if you wish, but note that
 /// using the object's memory address does *not* work for managed objects.
+@available(macOS 10.15, iOS 13, tvOS 13, watchOS 6, *)
```
leemaguire, line 38: should these guards not match `RealmSubscribable `?

```suggestion
@available(OSX 10.15, watchOS 6.0, iOS 13.0, iOSApplicationExtension 13.0, OSXApplicationExtension 10.15, tvOS 13.0, *)
```

leemaguire (commented):
Small comment

RealmSwift/Combine.swift
jsflax: I simply forwarded the @available attribute from `Identifiable` itself:
Screenshot 2021-02-05 at 09 43 30
I'd say `macOS` is more correct than `OSX` at this point as well?

As far as the `_ApplicationExtension`s, Combine and SwiftUI don't seem to explicitly mark them.

leemaguire: True, I guess forwarding the message from Apples API is the correct thing here anyway. Marking as resolved.

leemaguire (approved):

tgoyne:
Needs a changelog entry.

CHANGELOG.md
```diff
@@ -9,15 +9,16 @@ x.y.z Release notes (yyyy-MM-dd)
...
   This could break synced objects with a `Date` property  ([#6654](https://github.com/realm/realm-cocoa/issues/6654), since v10.0.0).
 * Fixed an issue where creating an object after file format upgrade may fail
   with assertion "Assertion failed: lo() <= std::numeric_limits<uint32_t>::max()"
  ([#4295](https://github.com/realm/realm-core/issues/4295), since v5.0.0)
 * Allow enumerating objects of types which are no longer present in the schema.
 * Add `RLMResponse.customStatusCode`. This fixes timeout exceptions that were occuring with a poor connection. ([#4188](https://github.com/realm/realm-core/issues/4188))
- * Fix - `RLMResponse` will have a non nil `customStatusCode` in case of error.
+* Fix - `RLMResponse` will have a non nil `customStatusCode` in case of error.
    ([#4188](https://github.com/realm/realm-core/issues/4188))
+* Make ObjectKeyIdentifiable available for macOS 10.15, iOS 13, tvOS 13, watchOS 6 and above
```
tgoyne, line 21: This makes me think that it was previously not available on those versions, and only on older versions. I'd word this as something like "Limit availability of ObjectKeyIdentifiable to platforms which support Combine to match the change made in the Xcode 12.5 SDK." Also needs a link to the issue since we did get a bug report for this.

tgoyne (approved):

tomcondon:
How exactly is this fixed? I get the same error using the latest Xcode 12.5 beta

tomcondon:
I do as well. How are developers to trust that Realm mobile is not just an afterthought for Mongo when this gets released?

jsflax:
Which dependency manager are you using? We had an issue with our cocoa pods release.

tomcondon:
Swift Package Manager. Now that it is an integral part of Xcode, why would you use anything else?

tomcondon:
We are basically screwed going forward until this works.

jsflax:
What version of Realm are you using? Have you resolved the latest versions of the Realm package?

If so, can you show us the error?

JulianKahnert:
@jsflax the problem still exists in `v5.5.0` any chance to fix this in that version? Not sure if we can update from `v5.5.0` to `v10.x.x` at this point.

PR #7101, opened 2021-02-09
Extend the Migration examples project (#7077)

DominicFrei:
Rework of #7077 after squash error in there.


DominicFrei:
Closed in favour of #7102.

PR #7104, opened 2021-02-09
Upgrade to core 10.5.0

tgoyne:
Schema validation for sync realms now distinguishes between auto-discovered schemas and explicitly set schemas so we need a bunch of awkward logic to set that properly.

List notifiers now send a notification when the list's parent object is deleted. We currently don't have a way to report that in the public API, so we need to explicitly swallow those notifications as otherwise they'll look like `.initial` notifications.


Realm/ObjectServerTests/SwiftObjectServerTests.swift
```diff
@@ -588,7 +588,9 @@ class SwiftObjectServerTests: SwiftSyncTestCase {
             proxy.delay = 1.0
             let ex = expectation(description: "async open")
             Realm.asyncOpen(configuration: config) { result in
-                XCTAssertNotNil(try? result.get())
+                let realm = try? result.get()
+                XCTAssertNotNil(realm)
+                realm?.syncSession?.suspend()
```
jsflax, line 593: does this fix the test?

tgoyne: The commit message explains why.

jsflax: A-ha, thanks.

jsflax (approved):

PR #7138, opened 2021-02-23
Improve asynchronous writes with Realm

### LINKED ISSUE #6953: Create wrapper API around threadSafeReference for writes
[~alexander.stigsen] wrote up an example of how to use a thread safe reference to make a write on a background thread. Should we wrap this functionality and expose this as a convenience API?

[https://realm.io/docs/cookbook/swift/object-to-background/](https://realm.io/docs/cookbook/swift/object-to-background/)

jsflax:
Fixes #6953.

EDIT: Updated for Swift 5.5. See:  https://github.com/realm/realm-cocoa/pull/7138#issuecomment-841694131

~Prototype API for async writes via combine. Incidentally, doing this through Combine only seems to have revealed the cleanest way to add this feature:~

The challenges of async write are:
- Users can pass anything into a block, even if we offer a convenient and obvious API for them to be passing in their `ThreadConfined` types
- We have no way of discerning the block's capture list, so attempting to freeze/thaw via that route is untenable, bar doing gnarly memory magic that would likely break with each new swift version.

~IMO, because Combine already has a fairly unique DSL/UX, the guidelines of how to use the API will be clearer. More specifically, the way objects are passed/mapped through combine's Publishers lends itself well to an API like this.~

~The added test `testAsyncWrite` shows the functionality in action.~

cc @ianpward @astigsen


RealmSwift/Combine.swift
```diff
@@ -880,6 +909,204 @@ public enum RealmPublishers {
...
+                let r = realm(self.config.rlmConfiguration, self.scheduler)!
+                r.beginWrite()
+                do {
+                    self.block(r, thawed1)
+                    try r.commitWrite()
+                    _ = subscriber.receive(thawed1)
+                    subscriber.receive(completion: .finished)
+                } catch {
+                    subscriber.receive(completion: .failure(error))
+                }
```
astigsen, line 962: Don't we need an autoreleasepool here to prevent the chance of the realm staying around, or has that been superseded in later versions of Swift?

RealmSwift/Combine.swift
```diff
@@ -310,6 +310,35 @@ public func changesetPublisher<T: Object>(_ object: T) -> RealmPublishers.Object
     RealmPublishers.ObjectChangeset<T>(object)
 }
 
+extension Realm {
+    @available(OSX 10.15, watchOS 6.0, iOS 13.0, iOSApplicationExtension 13.0, OSXApplicationExtension 10.15, tvOS 13.0, *)
+    public func asyncWrite<S: Scheduler>(on scheduler: S = DispatchQueue.main as! S, _ block: @escaping (Realm) -> Void) -> RealmPublishers.AsyncWriteFromRealm<S> {
```
astigsen, line 315: Shouldn't `asyncWrite` default to `DispatchQueue.global`?

astigsen:
How about naming the method `writeAsync()` instead of `asyncWrite()`? Just for pedagogical reasons. It would be nice if it showed up as an option on autocompletion every time someone tries to do a `write()`.

RealmSwift/Combine.swift
```diff
@@ -310,6 +310,35 @@ public func changesetPublisher<T: Object>(_ object: T) -> RealmPublishers.Object
...
+        on scheduler: S = DispatchQueue.main as! S,
+        _ confined: Confined,
+        _ block: @escaping (Realm, Confined) -> Void) -> RealmPublishers.AsyncWriteRealmCombine<S, Confined> {
+        RealmPublishers.AsyncWriteRealmCombine(scheduler,
+                                               configuration: configuration,
+                                               confined: confined,
+                                               block)
+    }
+    @available(OSX 10.15, watchOS 6.0, iOS 13.0, iOSApplicationExtension 13.0, OSXApplicationExtension 10.15, tvOS 13.0, *)
+    public func asyncWrite<S: Scheduler, Confined: ThreadConfined, Confined2: ThreadConfined>(
```
nirinchev, line 329: Does this mean you can't pass 3 objects to `asyncWrite`?

RealmSwift/Tests/CombineTests.swift
```diff
@@ -190,6 +190,48 @@ class CombineRealmTests: CombinePublisherTestCase {
...
+            realm.add([intObject, boolObject, stringObject, listObject])
+        }
+
+        XCTAssertTrue(Thread.isMainThread)
+
+        cancellable = self.realm.asyncWrite(on: subscribeOnQueue, intObject, boolObject) { realm, intObject, boolObject in
+            XCTAssertFalse(Thread.isMainThread)
+            intObject.intCol = 42
+            boolObject.boolCol = true
+        }.combineWrite(on: receiveOnQueue, with: listObject.double, stringObject) { realm, doubleList, stringObject in
```
nirinchev, line 213: Just looking at the API (without any idea about combine), wouldn't it be desirable to have a to pass a result from `asyncWrite` to `combineWrite`? E.g. something like:

```swift
self.realm.asyncWrite(on: queue1) { realm in
    let newObject = SwiftBoolObject()
    newObject.boolCol = true
    realm.add(newObject)
    return newObject
}.combineWrite(on: queue2) { realm, boolObject in
    // boolObject here is the value returned from asyncWrite

    if (boolObject.boolCol) {
        // blah blah
    }
}
```

nirinchev (commented):
Agree with Alex about asyncWrite -> writeAsync, although I'm coming from the C# world where Async is a suffix. If Swift developers expect it to be a prefix, then probably fine to keep it as is.

RealmSwift/Combine.swift
jsflax: Because of a lack of variadic generics support in Swift, we will have to write many overloads for this method that each take n amount of Objects/ThreadConfined types. In this prototype, I simply wrote two because it was easier and I was trying to see people's appetite for this API.

RealmSwift/Tests/CombineTests.swift
jsflax: I thought about that– Thomas also mentioned to me that `combineWrite` doesn't seem to make sense to him, considering that write transactions cannot occur at the same time anyways.

It would be fairly trivial to pass objects down the chain with Combine, but I couldn't think of a good use case. Even in your use case above, it's technically pointless– whatever operation you are doing can just be done after the `realm.add`.

RealmSwift/Combine.swift
nirinchev: Gotcha, makes sense - just wanted to understand if I'm misunderstanding the Swift syntax.

RealmSwift/Tests/CombineTests.swift
nirinchev: Agree - combineWrite seems to be a bit niche. The only kind-of-sort-of valid use case I can think of is if you want to dispatch the second write to the UI thread because you need to access something UIKit-y, although would be hard pressed to find a realistic example.

jsflax:
I've updated this PR to use the Swift 5.5 toolchain. I've removed the Combine integration that I previously proposed. What we have now (in 5.5) is fairly different but significantly cleaner that was users currently have to do for concurrent programming with Realm.

Consider the example I added to `AsyncTest/main.swift`:
```swift
@available(macOS 9999, *)
@objcMembers class Vegetable: Object {
    dynamic var _id = ObjectId.generate()
    dynamic var chopped = false
​
        "_id"
    }
​
    class func chop(@RealmSendable vegetable: Vegetable) async throws {
        print("Task \(Task.current!.hashValue): Vegetable \(vegetable._id) starts chop")
        vegetable.chopped = true
        Thread.sleep(forTimeInterval: 2)
        print("Vegetable \(vegetable._id) ends chop")
        return
    }
}
​
@available(macOS 9999, *)
@objcMembers class Meat: Object {
    dynamic var _id = ObjectId.generate()
    dynamic var marinated = false
​
    class func marinateMeat(@RealmSendable meat: Meat) async throws {
        print("Task \(Task.current!.hashValue): Marinate meat \(meat._id) starts")
        meat.marinated = true
        Thread.sleep(forTimeInterval: 2)
        print("Task \(Task.current!.hashValue): Marinate meat \(meat._id) ends")
    }
}
​
@available(macOS 9999, *)
@objcMembers class Dish: Object {
    dynamic var meat: Meat?
    dynamic var veggies = List<Vegetable>()
    dynamic var cooked = false
​
    convenience init(meat: Meat, veggies: [Vegetable]) {
        self.init()
        self.meat = meat
        self.veggies.append(objectsIn: veggies)
    }
}
​
@available(macOS 9999, *)
@objcMembers class Oven: Object {
    dynamic var _id = ObjectId.generate()
    dynamic var preheated = false
    dynamic var isCooking = false
​
        "_id"
    }
​
    class func preheat(@RealmSendable oven: Oven) async throws {
        print("Task \(Task.current!.hashValue): Oven \(oven._id) preheat starts")
        oven.preheated = true
        Thread.sleep(forTimeInterval: 2)
        print("Oven \(oven._id) preheat ends")
    }
​
    func cook(@RealmSendable dish: Dish) async throws {
        isCooking = true
        print("Task \(Task.current!.hashValue): Oven \(_id) cook starts")
        dish.cooked = true
        Thread.sleep(forTimeInterval: 2)
        print("Task \(Task.current!.hashValue): Oven \(_id) cook ends")
        isCooking = false
    }
}
​
@Sendable func gatherRawVeggies() async throws -> [Vegetable] {
    let realm = AsyncRealm()
    let onion = Vegetable()
    let shallot = Vegetable()
    try await realm.write {
        realm.add(onion)
        realm.add(shallot)
    }
    return [onion, shallot]
}
​
@Sendable func chopVegetables(_ vegetables: [Vegetable]) async throws {
    for veggie in try await gatherRawVeggies() {
        try await Vegetable.chop(vegetable: veggie)
    }
}
​
@Sendable func makeDinner() async throws -> Dish {
    let realm = try Realm()
​
    let oven = Oven()
    let meat = Meat()
​
    // async write and await
    try await realm.asyncWrite {
        realm.add(meat)
        realm.add(oven)
    }
​
    let veggies = try await gatherRawVeggies()
​
    try await Meat.marinateMeat(meat: meat)
    try await Oven.preheat(oven: oven)
​
    try await chopVegetables(veggies)
​
    let dish = Dish(meat: meat, veggies: veggies)
​
    try realm.write {
        realm.add(dish)
    }
​
    try await oven.cook(dish: dish)
    // async write and forget
    detachWrite(to: oven) { oven in
        oven.preheated = false
    }
    return dish
}
​
@Sendable
func main() async throws {
    let dishes = try await withThrowingTaskGroup(of: Dish.self) { group -> [Dish] in
        var dishes: [Dish] = []
​
        // cook two dinners in parallel
        group.spawn {
            try await makeDinner()
        }
        group.spawn {
            try await makeDinner()
        }
        // collect the dinners when complete
        for try await dish in group { // (2)
            dishes.append(dish)
        }
​
        return dishes
    }
​
    print("Bon Appetit!: \(dishes)")
}
```

Note that the `@available(macOS 9999, *)` is temporary, and `@Sendable` on the methods is because they are global.

### How do the async class methods on the models work? 

They are automatically being wrapped in write transactions by the `@RealmSendable` property wrapper.

`@RealmSendable` begins a write transaction on `init` and commits it on `deinit`. This all happens under the hood so the user doesn’t have to do anything.

Having `propertyWrappers` for function parameters is a huge boon for us. This will be incredibly handy with projections as well, and lines up nicely as we ask users to place their business logic onto their models/projections.

This takes pressure off the "stray async write" use case:
```swift
// async write and forget
detachWrite(to: oven) { oven in
    oven.preheated = false
}
```
With Swift 5.5, we still cannot achieve a great syntax for this. However, encouraging users to move the business logic to `@RealmSendable` async write methods yields a clean cut way for users to asynchronously write to Realm.

PR #7143, opened 2021-02-25
Needed quotes around the simulator location

LucasVanDongen:
I have multiple versions of macOS running on my machine (Catalina, Big Sur, Big Sur Beta) and I only keep a bunch of Xcode versions on my main partition because of HD space issues. The default HD name is "Macintosh HD" which contains a space and if I target an Xcode that is on that disk the `build.sh` script fails when installing Cocoapods because it encounters an unexpected space in the path.

I patched it by generating the simulator path first and then putting quotes around the variable containing the path. After a clean it started building again, so that seems to fix the problem.

Since Realm is a dependency of a dependency (matrix) it's hard to work around the problem so I would really appreciate if you could merge it into the main branch sometime soon.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -364,8 +364,9 @@ download_common() {
         # Xcode 11 dsymutil crashes when given debugging symbols created by
         # Xcode 12. Check if this breaks, and strip them if so.
         local test_lib=core/realm-monorepo.xcframework/ios-*-simulator/librealm-monorepo.a
+        local isysroot=$(xcrun --sdk iphonesimulator --show-sdk-path)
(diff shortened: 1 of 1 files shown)


dianaafanador3:
hi @LucasVanDongen seems like this fix is not valid any more for latest released version, we have removed support for Xcode 11 and no longer to do this check. Let me know if you are still having issues, and I can help you updating the PR, or you can lose this PR if is no longer relevant.

dianaafanador3:
@LucasVanDongen closing this as this PR is no longer relevant. If you still have some issues feel free to open it again, or open a new issue.

PR #7158, opened 2021-03-03
Add support of arm64 in Carthage build

pavel-ship-it:
arm64 was disabled as Carthage was unable to build XCFrameworks.
Now it's fixed and arm64 can be enabled for M1.

Fix for [Could not build arm64 contains xcframework for iOS simulator by Carthage](https://github.com/realm/realm-swift/issues/7151)


tgoyne:
This breaks the non-xcframework build, which is currently the only thing that Carthage supports for shipping prebuilt libraries. I'm not sure if there's a way to support both, but as-is this'll break our release pipeline.

https://github.com/Carthage/Carthage/pull/3123 looks like it fixes that and might land soon so maybe we can wait for that?

egorf:
@tgoyne Hi!

Do you think it's time we could revisit this issue? The newest Carthage release supports xc frameworks.

Currently it does not seem possible to build Realm with `carthage update --use-xcframeworks`.

tgoyne:
You have to include `--no-use-binaries` to have Carthage actually build the XCFramework. [Carthage 0.38](https://github.com/Carthage/Carthage/pull/3152) will add support for prebuilt XCFrameworks.

timothycosta:
It would be great to get this merged now that Carthage 0.38.0 is out. I've installed this branch in my cartfile and it seems to be working well.

```
# Cartfile
github "realm/realm-cocoa" "py/carthage_arm64"
```

r-plus:
by the way, before release of Carthage 0.38, realm-cocoa already supported xcframework version of prebuild binary since v10.7.2.
no need additional change for release flow?

Tovkal:
This PR seems to be a duplicate of https://github.com/realm/realm-cocoa/pull/7095. Either way, I have tried the changes and they work. Hoping this can be merged soon 😄

r-plus:
Hi @tgoyne 

Do you have any blocking issue for this?

M1 + Carthage users still cannot use realm-cocoa as native on simulator if using incompatible version of Xcode/Swift between project and pre-build binary.

iamszabo:
Is there any update on this? It would be nice to use Carthage on M1

Tovkal:
Seems like like the latest release, compatible with Xcode 13, still does not work on M1s

dianaafanador3:
@pavel-ship-it After rebase seems like CI is not breaking, swiftpm-debug 12.5.1 fails because of this test `testEncryptedSyncedRealmWrongKey` and I've seen this failing in other PRs

hiragram:
@pavel-ship-it Do you have any plans to work on this PR?

gongzhang:
I upgraded to Xcode 13.2 RC yesterday, and since the Swift compiler version was updated, this resulted in Realm's prebuilt XCFramework having to be recompiled.

When I executed `carthage bootstrap --use-xcframeworks`, Carthage went through the source build process as expected. But the result of the build is missing arm64 arch, so it doesn't work properly.

As a workaround, I can clone the `realm-cocoa` repository and build the correct XCFramework with `build.sh`. But I think we need to _at least_ clarify this issue in the installation section of the documentation. After all, the current Carthage support is incomplete.

olejnjak:
Probably to solve this it would rather make sense for Realm to make binary only XCFramework distribution, that would be ABI stable...this compilation exclusion would keep old framework distribution still possible and will save a lot of trouble for XCFramework users, still I think that Realm should be distributed XCFramework-first nowadays...

And if someone needs old fat framework, he should take care of it himself using a workaround script that will take care of removing arm64 simulator like [this one](https://github.com/Carthage/Carthage/blob/master/Documentation/Xcode12Workaround.md)...

pavel-ship-it:
Hi @hiragram, @olejnjak the code available here right now is not enough to address the original problem with the Carthage builds. Even though the code here eliminates the problem for some users, it is not flexible enough to let us build all of the required configurations.
Please let me some time to work on it. I'd expect the fix in the next month. I'll keep you all posted here in case I face any delay.

Jenkinsfile.releasability
```diff
@@ -223,8 +224,8 @@ def doBuild() {
 
           ditto -ck merged/ Carthage.framework.zip
           '''
-
           archiveArtifacts artifacts: 'Carthage.framework.zip'
+
```
dianaafanador3, line 228: Extra line

dianaafanador3 (approved):
I tested this and seems to be working fine, Only a small comment and LGTM

tgoyne (approved):

hiragram:
@pavel-ship-it Thank you for your great work. When will the version including this patch be released?

pavel-ship-it:
@hiragram Thank you. It will be released later this month.

pitoneux:
This issue exists also with spm. Hope this fix address this too ?

tgoyne:
Any problem you're having with SPM is entirely unrelated to this, and you should create a new issue.

PR #7167, opened 2021-03-05
Remove class keyword from RealmEnum example

dacharyc:
When working on a documentation update to add RealmEnum to the [MongoDB Realm iOS SDK docs site](https://docs.mongodb.com/realm/sdk/ios/), I referenced this in the auto-generated SDK docs, and noticed the `class` keyword shouldn't be in the example. Hope this quick-fix PR is helpful!


jsflax (approved):
Nice! Thanks!

PR #7170, opened 2021-03-08
Mixed datatype

### LINKED ISSUE #6840: Implement Mixed Data type

ericjordanmossman:
Closes #6840

Adds a 'Mixed' datatype which can hold different value types. ex)
```
class AnyValue: Object {
    var anyCol = RealmProperty<AnyRealmValue>()
    let anyList = List<AnyRealmValue>()
}
// ...
object.anyCol.value = .int(5)
object.anyCol.value = .bool(true)
object.anyCol.value = .string("hi")
```
```
@interface MixedObject: RLMObject
@property id<RLMValue> anyCol;
@property RLMArray<RLMValue> *anyArray;
@end
// ...
object.anyCol = @YES;
object.anyCol = @12
object.anyCol = @1.145
object.anyCol = @"hi"
```
Mixed properties:
- Cannot be marked optional since null is always a permitted value.
- Can be indexed
- Cannot be a primary key

Comparatibility between mixed value types:
- Numeric: int, bool, float, double, decimal
- Byte based: string, binary
- Time based: timestamp, objectId
- The remaining types do not attempt any conversions: link, UUID

Querying:
- Equals queries match on the value and the type.
- Not equals queries will match objects with either different value or different type.
- Numeric values are converted where possible - e.g. mixed == 1 will match all of 1.0, 1, and true.
- String properties will not match numeric queries - e.g. mixed == 1 will not match “1”. mixed == “1” will not match 1, 1.0, or true.


Realm/ObjectServerTests/RLMObjectServerTests.mm
```diff
@@ -561,31 +561,72 @@ - (void)testAddObjects {
...
     if (self.isParent) {
         CHECK_COUNT(0, Person, realm);
+        CHECK_COUNT(0, AllTypesSyncObject, realm)
         RLMRunChildAndWait();
         [self waitForDownloadsForRealm:realm];
         CHECK_COUNT(4, Person, realm);
+        CHECK_COUNT(1, AllTypesSyncObject, realm);
+
+        AllTypesSyncObject *obj = [[AllTypesSyncObject allObjectsInRealm:realm] firstObject];
+        XCTAssertEqual(obj.boolCol,                           [values[@"boolCol"] boolValue]);
```
leemaguire, line 575: weird formatting

ericjordanmossman: Instead of the alignment on values, should it just be natural line length?

Realm/ObjectServerTests/RealmServer.swift
```diff
@@ -291,7 +291,7 @@ public class RealmServer: NSObject {
     @objc public static var shared = RealmServer()
 
     /// Log level for the server and mongo processes.
-    public var logLevel = LogLevel.none
+    public var logLevel = LogLevel.info
```
ericjordanmossman, line 294: I left this in by mistake.

Realm/ObjectServerTests/RLMObjectServerTests.mm
leemaguire: yes

Realm/RLMUser.mm
```diff
@@ -226,6 +226,10 @@ - (void)logOutWithCompletion:(RLMOptionalErrorBlock)completion {
     });
 }
 
+- (id)profile {
+    _user->user_profile();
```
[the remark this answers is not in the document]
ericjordanmossman: I think exposing syncUserProfile is a separate feature and can be taken out of this PR.
Need to confirm with @jsflax .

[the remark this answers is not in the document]
jsflax: Yes, this should not be here. Looks like it got pulled in from a dead branch.

Realm/RLMObjectStore.mm
```diff
@@ -170,6 +171,16 @@ void RLMAddObjectToRealm(__unsafe_unretained RLMObjectBase *const object,
     return object;
 }
 
+RLMObjectBase *RLMObjectFromObjLink(RLMRealm *realm, realm::ObjLink&& objLink) {
+    auto& info = realm->_info[objLink.get_table_key()];
+    RLMAccessorContext c{info};
+    RLMObjectBase *object = RLMCreateManagedAccessor(info.rlmObjectSchema.accessorClass, &info);
+    auto table = realm->_realm->read_group().get_table(objLink.get_table_key());
+    object->_row = table->get_object(objLink.get_obj_key());
```
tgoyne, line 179: realm.group.get_object(objLink)

Realm/RLMObjectStore.mm
```diff
@@ -170,6 +171,16 @@ void RLMAddObjectToRealm(__unsafe_unretained RLMObjectBase *const object,
     return object;
 }
 
+RLMObjectBase *RLMObjectFromObjLink(RLMRealm *realm, realm::ObjLink&& objLink) {
+    auto& info = realm->_info[objLink.get_table_key()];
+    RLMAccessorContext c{info};
```
tgoyne, line 176: unused

Realm/RLMProperty_Private.hpp
```diff
@@ -31,3 +29,11 @@ namespace realm {
 - (realm::Property)objectStoreCopy:(RLMSchema *)schema;
 
 @end
+
+static inline bool isNullable(const realm::PropertyType& t) {
+    if (t == realm::PropertyType::Mixed) {
```
tgoyne, line 34: return t != realm::PropertyType::Mixed || is_nullable(t);

Realm/RLMQueryUtil.mm
```diff
@@ -972,12 +1040,105 @@ void convert_null(T&& value, Fn&& fn) {
...
+        else if (numberIsInteger(i)) {
+            add_numeric_constraint(RLMPropertyTypeAny, operatorType, column, value_of_type<Int>(value));
+        }
+        else if ([value objCType][0] == *@encode(float)) {
+            add_numeric_constraint(RLMPropertyTypeAny, operatorType, column, value_of_type<Float>(value));
+        }
+        else if ([value objCType][0] == *@encode(double)) {
+            add_numeric_constraint(RLMPropertyTypeAny, operatorType, column, value_of_type<Double>(value));
+        }
+    }
```
tgoyne, line 1078: I don't think we want any of this. While writing literally `"mixedProp > true"` would obviously be silly, `"mixedProp > %@", someMixedValue` is perfectly reasonable and should work regardless of what the mixed value happens to be.

Realm/RLMQueryUtil.mm
```diff
@@ -972,12 +1040,105 @@ void convert_null(T&& value, Fn&& fn) {
...
+            add_bool_constraint(RLMPropertyTypeAny, operatorType, column, value);
+            break;
+        case realm::type_String:
+        case realm::type_Binary:
+            add_string_constraint(operatorType, predicateOptions, std::move(column), value);
+            break;
+        default:
+            @throw RLMException(@"Unsupported RLMValue type.");
+    }
+}
```
tgoyne, line 1122: Similarly to above, I don't think the value that a Mixed happens to have should have an effect on what operations are legal on it.

Realm/RLMUtil.mm
```diff
@@ -436,7 +389,64 @@ BOOL RLMIsRunningInPlayground() {
...
+            return realm::Mixed(RLMObjcToUUID(v));
+        case RLMPropertyTypeString:
+            return realm::Mixed([(NSString *)v cStringUsingEncoding:NSUTF8StringEncoding]);
+        case RLMPropertyTypeData:
+            return realm::Mixed(realm::BinaryData((const char*)[(NSData *)v bytes],
+                                                  (size_t)[(NSData *)v length]));
+        case RLMPropertyTypeDate:
+            return realm::Mixed(realm::Timestamp([(NSDate *)v timeIntervalSince1970], 0));
+        case RLMPropertyTypeObject: {
+            // If we are unboxing an object and it is unmanaged, we need to
```
tgoyne, line 428: This is called in places where it absolutely should not be mutating anything. This function needs to take a CreatePolicy and honor that.

Realm/RLMUtil.mm
```diff
@@ -436,7 +389,64 @@ BOOL RLMIsRunningInPlayground() {
...
+    else if ([v conformsToProtocol:@protocol(RLMValue)]) {
+        type = [v valueType];
+    }
+    else {
+        REALM_TERMINATE("Unexpected Type");
+    }
+
+    switch (type) {
+        case RLMPropertyTypeInt:
+            return realm::Mixed([(NSNumber *)v intValue]);
```
tgoyne, line 411: All the uses of `intValue` need to be `longLongValue` to avoid truncating 64-bit ints.

Realm/TestUtils/include/RLMTestObjects.h
```diff
@@ -207,6 +218,16 @@ RLM_COLLECTION_TYPE(AllTypesObject)
 @property RLMArray<RLMDecimal128> *decimalObj;
 @property RLMArray<RLMObjectId> *objectIdObj;
 @property RLMArray<RLMUUID> *uuidObj;
+@property RLMArray<RLMValue> *anyObj1;
```
tgoyne, line 221: These need actual names.

Realm/ObjectServerTests/RealmServer.swift
```diff
@@ -531,10 +531,11 @@ public class RealmServer: NSObject {
         }
 
         let app = session.apps[appId]
-        let group = DispatchGroup()
+        let group0 = DispatchGroup()
+        let group1 = DispatchGroup()
```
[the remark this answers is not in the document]
ericjordanmossman: group1 rules rely on group0 having been registered by the server.
I'm drawing a blank on what an actual name would be. Something like `baseRulesGroup`, `superRulesGroup`?

Realm/ObjectServerTests/SwiftObjectServerTests.swift
```diff
@@ -106,9 +125,25 @@ class SwiftObjectServerTests: SwiftSyncTestCase {
...
             if isParent {
                 checkCount(expected: 0, realm, SwiftPerson.self)
+                checkCount(expected: 0, realm, SwiftTypesSyncObject.self)
                 executeChild()
                 waitForDownloads(for: realm)
-                checkCount(expected: 3, realm, SwiftPerson.self)
+                checkCount(expected: 4, realm, SwiftPerson.self)
+                checkCount(expected: 1, realm, SwiftTypesSyncObject.self)
+
+                let obj = realm.objects(SwiftTypesSyncObject.self).first!
```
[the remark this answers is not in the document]
ericjordanmossman: Are you referring to the 'nilpartition' value portion? Or checking the properties in a test that's defined as "adding objects"?
If it's the first: I'll remove `SwiftTypesSyncObject`s from `testSwiftAddObjectsWithNilPartitionValue()` and it's Objc version.
If it's the second: I'll remove `SwiftTypesSyncObject` from `testSwiftAddObjects()` also and check `SwiftTypesSyncObject` have the proper values in more explicitly defined tests.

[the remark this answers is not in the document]
ericjordanmossman: or is it neither?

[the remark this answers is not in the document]
tgoyne: This test is validating that we can open the nil partition and is just syncing an object as a way to check that it's not failing silently. There's no reason to change this test when adding new data types and all of the code added to it is unrelated to the test.

Realm/RLMValueBase.mm
```diff
@@ -0,0 +1,92 @@
...
+    if (!_managed) {
+        REALM_ASSERT_DEBUG(parent->_realm);
+        _parent = parent;
+        _propertyName = property.name;
+        _managed = YES;
+    }
+
+    // If this mixed container stores an `Object` we need to manually add it
+    // to the Realm.
+    if (_backingValue && [(id)_backingValue isKindOfClass:[RLMObjectBase class]]) {
```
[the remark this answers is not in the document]
leemaguire: Basically if the `AnyRealmValue` property is on a managed object, and you store an unmanaged object in that `AnyRealmValue` property, we need to add the unmanaged object to the realm so that we can get the `ObjKey` from it later. It is an oversight that this is here, as the adding to the Realm now takes part in `RLMObjcToMixed`. The name of the enclosing method is also something I need to change because it doesn't make sense

Realm/RLMObjectStore.mm
leemaguire: this is used in `RLMMixedToObjc`

Realm/RLMProperty_Private.hpp
tgoyne: Er right that should be && and not ||.

Realm/RLMQueryUtil.mm
leemaguire: The reason why this is where is because `@NO` can represent an int. So while we can get rid of all this, doing so would not let us throw an exception if a user tried to do  `("mixedProp > %@), @NO)`

Realm/RLMQueryUtil.mm
leemaguire: We need to make sense of what the comparing value is in each case and what constraint should be assigned. For example if I pass in `@123.456` how would we differentiate that from `@"hello"`?

Realm/RLMValueBase.mm
```diff
@@ -0,0 +1,92 @@
...
+        catch (std::exception const& err) {
+            @throw RLMException(err);
+        }
+    }
+    else {
+        return _backingValue;
+    }
+}
+
+- (void)attachIfNeededWithParent:(RLMObjectBase *)parent
```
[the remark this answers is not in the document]
leemaguire: I went back and forth a few times on the name, its on my todo list. It should never be called twice

Realm/RLMUtil.mm
leemaguire: Who should be in charge of the CreatePolicy? Should it be `ValueUpdater` in Object Store?

Realm/Tests/RLMValueTests.m
```diff
@@ -0,0 +1,509 @@
...
+@interface RLMValueTests : RLMTestCase
+@end
+
+@implementation RLMValueTests
+
+#pragma mark - Initialization
+
+- (void)testIntInitialization {
+    id<RLMValue> v = @123;
+    XCTAssertEqual(v, @123);
```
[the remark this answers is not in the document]
leemaguire: The intention here is more so that the `valueType` / `rlm_valueType` gives the correct value. Changed the tests to reflect that.

Realm/RLMValue.h
```diff
@@ -0,0 +1,107 @@
...
+ `NSUUID`
+ `NSString`
+ `RLMObject
+ `RLMObjectId`
+ `RLMDecimal128`
+ */
+@protocol RLMValue
+
+/// Describes the type of property stored.
+@property (readonly) RLMPropertyType valueType NS_REFINED_FOR_SWIFT;
```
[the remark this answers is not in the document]
leemaguire: Changed to rlm_ valueType

Realm/RLMQueryUtil.mm
tgoyne: Yes, I think that should be allowed. `>` is a supported operation for Mixed properties.

Realm/RLMUtil.mm
tgoyne: unbox() needs to forward on the value it was passed. Queries should pass CreatePolicy::Skip. The accessor function should pass CreatePolicy::SetLink.

Realm/RLMArray.mm
```diff
@@ -215,7 +215,7 @@ void RLMArrayValidateMatchingObjectType(__unsafe_unretained RLMArray *const arra
         @throw RLMException(@"Invalid nil value for array of '%@'.",
                             array->_objectClassName ?: RLMTypeToString(array->_type));
     }
-    if (array->_type != RLMPropertyTypeObject) {
+    if (array->_type != RLMPropertyTypeObject && array->_type != RLMPropertyTypeAny) {
```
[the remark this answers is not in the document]
leemaguire: changed to `((array->_type == RLMPropertyTypeAny) || (array->_type != RLMPropertyTypeObject))`

Realm/RLMQueryUtil.mm
leemaguire: Also without these conditionals for specific type information queries like `@"anyCol BETWEEN %@", @[@0, @5]` don't work because numbers like `@0` and `@1` appear as `BOOL`'s

tgoyne: That shouldn't matter? The query engine appears to do the same thing for `col < Mixed(true)` and `col < Mixed(1)`.

Realm/ObjectServerTests/SwiftCollectionSyncTests.swift
```diff
@@ -0,0 +1,209 @@
...
+                    try realm.write({
+                        object[keyPath: keyPath][0] = values[1]
+                    })
+                    XCTAssertEqual(object[keyPath: keyPath][0], values[1])
+                }
+                waitForUploads(for: realm)
+                checkCount(expected: 1, realm, SwiftCollectionSyncObject.self)
+            }
+        } catch {
+            XCTFail("Got an error: \(error) (process: \(isParent ? "parent" : "child"))")
```
jsflax, line 91: Is there any advantage to not just having this method throw? Too much indentation makes this slightly harder to read.

Realm/RLMClassInfo.mm
```diff
@@ -95,6 +95,15 @@
...
 
+RLMClassInfo& RLMSchemaInfo::operator[](realm::TableKey const& key) {
+    for (auto& pair : m_objects) {
+        if (pair.second.table()->get_key() == key)
+            return pair.second;
+    }
+    @throw RLMException(@"Table Key '%@' is not managed by the Realm. ", @(key.value));
+}
+
+
```
jsflax, line 106: extra new line

Realm/RLMValue.mm
```diff
@@ -0,0 +1,134 @@
...
+////////////////////////////////////////////////////////////////////////////
+
+#import "RLMValue.h"
+#import "RLMUtil.hpp"
+
+#pragma mark NSData
+
+@implementation NSData (RLMValue)
+
+- (RLMPropertyType)rlm_valueType {
```
jsflax, line 26: This is named oddly

Realm/RLMValue.mm
```diff
@@ -0,0 +1,134 @@
...
+}
+
+@end
+
+#pragma mark NSNull
+
+@implementation NSNull (RLMValue)
+
+- (RLMPropertyType)rlm_valueType {
+    return RLMPropertyTypeAny;
```
jsflax, line 71: This doesn't seem correct

Realm/RLMValue.mm
```diff
@@ -0,0 +1,134 @@
...
+
+@implementation RLMObject (RLMValue)
+
+- (RLMPropertyType)rlm_valueType {
+    return RLMPropertyTypeObject;
+}
+
+@end
+
+#pragma mark RLMObjectBase
```
jsflax, line 116: Do we need conformance for Object and ObjectBase here?

Realm/RLMValueBase.h
```diff
@@ -0,0 +1,35 @@
...
+////////////////////////////////////////////////////////////////////////////
+
+#import <Foundation/Foundation.h>
+
+NS_ASSUME_NONNULL_BEGIN
+
+@class RLMObjectBase, RLMProperty;
+@protocol RLMValue;
+
+@interface RLMValueBase : NSObject
```
jsflax, line 26: This definitely needs comments containing an explanation of why this exists, as it's not readily clear nor is it _really_ a base class for `RLMValue`.

Realm/Tests/PrimitiveRLMValuePropertyTests.tpl.m
```diff
@@ -0,0 +1,175 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+////////////////////////////////////////////////////////////////////////////
+
+#import "RLMTestCase.h"
+#include <objc/runtime.h>
```
jsflax, line 20: What is this required for?

RealmSwift/AnyRealmValue.swift
```diff
@@ -0,0 +1,150 @@
...
+    /// - Returns: A Realm Object of the supplied type if that is what the underlying value is,
+    /// otherwise `nil` is returned.
+    public func objectValue<T: Object>(_ objectType: T.Type) -> T? {
+        guard case let .object(o) = self else {
+            return nil
+        }
+        return o as? T
+    }
+
+    internal var _rlmValue: RLMValue? {
```
jsflax, line 142: Since this is internal, it doesn't need the `_`.

RealmSwift/ObjectiveCSupport+AnyRealmValue.swift
```diff
@@ -0,0 +1,120 @@
...
+// limitations under the License.
+//
+////////////////////////////////////////////////////////////////////////////
+
+
+import Foundation
+import Realm
+
+public extension ObjectiveCSupport {
+
```
jsflax, line 24: extra new line

RealmSwift/ObjectiveCSupport+AnyRealmValue.swift
```diff
@@ -0,0 +1,120 @@
...
+            return d as NSDate
+        case let .objectId(o):
+            return o as RLMObjectId
+        case let .decimal128(o):
+            return o as RLMDecimal128
+        case let .uuid(u):
+            return u as NSUUID
+        case let .object(o):
+            return o.unsafeCastToRLMObject()
+
```
jsflax, line 49: extra newline

RealmSwift/RealmProperty.swift
```diff
@@ -0,0 +1,85 @@
...
+
+/**
+ A `RealmProperty` instance represents an polymorphic value for supported types.
+
+ To change the underlying value stored by a `RealmProperty` instance, mutate the instance's `value` property.
+
+ - Note:
+ An `RealmProperty` should not be declared as `@objc dynamic` on a Realm Object. Use `let` instead.
+ */
+public final class RealmProperty<Value: RealmPropertyType>: RLMValueBase {
```
jsflax, line 31: Is there any reason this is generic? Seems like this is only useful for `AnyRealmValue`

RealmSwift/Tests/ObjectCreationTests.swift
```diff
@@ -1426,7 +1428,7 @@ class ObjectCreationTests: TestCase {
         case .object:   return ["invalid", ["a"], ["boolCol": "a"], SwiftIntObject()]
         case .objectId: return ["invalid", 123]
         case .decimal128: return ["invalid"]
-        case .any: fatalError("not supported")
+        case .any:      return [List<String>()]
```
jsflax, line 1431: extra spaces

RealmSwift/Tests/ObjectSchemaInitializationTests.swift
```diff
@@ -22,6 +22,24 @@ import Realm.Private
 import Realm.Dynamic
 import Foundation
 
+class RealmPropObj: Object {
+    let any = RealmProperty<AnyRealmValue>()
+    let int = RealmProperty<Int?>()
+    @objc dynamic var blah = 0
+}
+
+//class LeeTest: XCTestCase {
```
jsflax, line 31: commented out code

RealmSwift/AnyRealmValue.swift
```diff
@@ -0,0 +1,150 @@
...
+        }
+        return u
+    }
+
+    /// Returns the stored value as a Realm Object of a specific type.
+    ///
+    /// - Parameter objectType: The type of the Object to return.
+    /// - Returns: A Realm Object of the supplied type if that is what the underlying value is,
+    /// otherwise `nil` is returned.
+    public func objectValue<T: Object>(_ objectType: T.Type) -> T? {
```
jsflax, line 135: Might want to consider changing the name slightly to reflect that this is a function, not a variable.

Also, based on @nirinchev points, we should also expose a method that returns the value as a `DynamicObject`:

```swift
public fun dynamicObject() -> DynamicObject?
```

RealmSwift/RealmProperty.swift
leemaguire: It will pretty much replace RealmOptional, for the short while that property wrappers aren't around. @tgoyne do you have any opinions on restricting this to just `AnyRealmValue`?

Realm/RLMValue.mm
leemaguire: as we are adding a property name to the global namespace we cannot risk conflicting with what Apple may add in the future. Adding a 3 character prefix helps us guard against that.

Realm/RLMValueBase.h
leemaguire: This is a left over from the `AnyRealmValue` class, I'll clean it up

Realm/Tests/PrimitiveRLMValuePropertyTests.tpl.m
ericjordanmossman: It isn't. I'll remove.

Realm/RLMValue.mm
leemaguire: RLMPropertyType is an enum, I think returning nil would not be the correct approach.

jsflax: I agree that would shouldn't return nil. I guess this should either return what you have here or `-1` to imply that it's an unknown type.

Realm/RLMValue.mm
leemaguire: just RLMObjectBase, I'll remove RLMObject

Realm/ObjectServerTests/SwiftCollectionSyncTests.swift
leemaguire: Resolved in https://github.com/realm/realm-cocoa/pull/7190

Realm/RLMObjectBase.mm
```diff
@@ -212,6 +213,13 @@ - (void)setValue:(id)value forUndefinedKey:(NSString *)key {
         else if (property.optional) {
             RLMSetOptional(object_getIvar(self, ivar), value);
         }
+        else if (auto valueBase = RLMDynamicCast<RLMPropertyBase>(object_getIvar(self, ivar))) {
+            if (auto v = RLMDynamicCast<RLMPropertyBase>(value)) {
```
leemaguire, line 217: can we simplify this.

Realm/ObjectServerTests/RLMSyncTestCase.mm
```diff
@@ -175,38 +175,74 @@ + (NSString *)primaryKey {
...
-
-+ (NSDictionary *)values:(int)i {
+    return @[@"boolCol", @"cBoolcol",
+             @"intCol", @"doubleCol",
+             @"stringCol", @"binaryCol",
+             @"dateCol", @"longCol",
+             @"decimalCol", @"uuidCol", @"objectIdCol"];
+}
+
++ (NSDictionary *)values:(int)idx {
```
tgoyne, line 185: The parameter here isn't an index. It's just a number with no particular meaning.

Realm/ObjectServerTests/RLMCollectionSyncTests.mm
```diff
@@ -0,0 +1,411 @@
...
+#if TARGET_OS_OSX
+
+#pragma mark RLMSet Sync Tests
+
+@interface RLMSetObjectServerTests : RLMSyncTestCase
+@end
+
+@implementation RLMSetObjectServerTests
+
+- (void)roundTripWithKeyPath:(NSString *)keyPath
```
tgoyne, line 30: Passing a `RLMSet (^)(id)` block which returns the appropriate value rather than the property name will probably make this method a lot more readable as it's slightly more typed.

Realm/ObjectServerTests/SwiftObjectServerTests.swift
```diff
@@ -2062,6 +2062,58 @@ class SwiftMongoClientTests: SwiftSyncTestCase {
     }
 }
 
+class AnyRealmValueSyncTests: SwiftSyncTestCase {
+    /// The purpose of this test is to confirm that when an Object is set on a mixed Column and an old
+    /// version of an app does not have that Realm Object / Schema, we must fail gracefully and return nil
+    /// when calling `obj!.anyCol.value.object(Object.self)`
```
tgoyne, line 2068: We probably want to be returning a dynamic object with a dynamically created ObjectSchema unless that's impossible to implement.

Realm/RLMObjectStore.h
```diff
@@ -34,6 +34,8 @@ NS_ASSUME_NONNULL_BEGIN
 
 void RLMVerifyHasPrimaryKey(Class cls);
 
+void RLMVerifyInWriteTransaction(RLMRealm *const realm);
```
tgoyne, line 37: This doesn't appear to be needed in the header?

Realm/RLMObjectStore.mm
```diff
@@ -110,9 +110,9 @@ void RLMInitializeSwiftAccessorGenerics(__unsafe_unretained RLMObjectBase *const
             id managedCollection = [[cls alloc] initWithParent:object property:prop];
             [ivar set_rlmCollection:managedCollection];
         }
-        else if (prop.optional) {
+        else if (RLMDynamicCast<RLMSwiftValueStorage>(object_getIvar(object, prop.swiftIvar)) != nil) {
```
tgoyne, line 113: `isKindOfClass:` should be used for just checking the type.

Realm/RLMUtil.mm
```diff
@@ -421,9 +417,19 @@ BOOL RLMIsRunningInPlayground() {
         case RLMPropertyTypeObject: {
             // If we are unboxing an object and it is unmanaged, we need to
             // add it to the Realm.
-            if (RLMObjectBase *objBase = RLMDynamicCast<RLMObjectBase>(v); !objBase->_realm) {
-                RLMAddObjectToRealm(objBase, realm, RLMUpdatePolicyError);
+            if (RLMObjectBase *objBase = RLMDynamicCast<RLMObjectBase>(v); !objBase->_realm
+                && createPolicy.create) {
```
tgoyne, line 421: Indenting like this makes the second line of the conditional loop like part of the if body.

Realm/RLMUtil.mm
```diff
@@ -421,9 +417,19 @@ BOOL RLMIsRunningInPlayground() {
...
+                RLMVerifyInWriteTransaction(realm);
+                // We do not want to copy the object, instead we want to promote
+                // it to a managed one.
+                createPolicy.copy = false;
```
tgoyne, line 425: Shouldn't this just forward copy on as-is? We do want to copy the object if we're inside a call to `create()`.

Realm/ObjectServerTests/RLMCollectionSyncTests.mm
```diff
@@ -0,0 +1,411 @@
...
+            CHECK_COUNT(0, RLMSetSyncObject, realm);
+            RLMRunChildAndWait();
+            [self waitForDownloadsForRealm:realm];
+            CHECK_COUNT(1, RLMSetSyncObject, realm);
+            // Run the child again to add the values
+            RLMRunChildAndWait();
+            [self waitForDownloadsForRealm:realm];
+            CHECK_COUNT(1, RLMSetSyncObject, realm);
+            RLMResults<RLMSetSyncObject *> *results
+            = [RLMSetSyncObject allObjectsInRealm:realm];
```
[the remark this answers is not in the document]
leemaguire: I'll address this in https://github.com/realm/realm-cocoa/pull/7190

Realm/ObjectServerTests/RLMCollectionSyncTests.mm
leemaguire: But the object is created within the method and is extracted from a Results collection. By passing the block isn't it constraining you to create the Set outside of the roundTrip method?

Realm/ObjectServerTests/RLMCollectionSyncTests.mm
```diff
@@ -0,0 +1,411 @@
...
+
+            XCTAssertEqual(((RLMSet *)obj[keyPath]).count, values.count);
+            XCTAssertEqual(((RLMSet *)obj[otherKeyPath]).count, otherValues.count);
+            // Run the child again to intersect the values
+            RLMRunChildAndWait();
+            [self waitForDownloadsForRealm:realm];
+            CHECK_COUNT(1, RLMSetSyncObject, realm);
+
+            if (!isObject) {
+                XCTAssertTrue([((RLMSet *)obj[keyPath]) intersectsSet:((RLMSet *)obj[otherKeyPath])]);
```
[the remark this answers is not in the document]
tgoyne: Reading properties requires casting to the type, but calling methods doesn't (and it's not even really useful as documentation because casting to an unrelated type which happens to have the same method works).

Realm/ObjectServerTests/RLMCollectionSyncTests.mm
tgoyne: No, the roundTrip method passes the object to the block. It'd be `[self roundTripWithPropertyGetter:^(RLMSetSyncObject *o) { return o.intSet; }` rather than [self roundTripWithKeyPath:@"intSet"`. More verbose at the callsite, but basically identical within the function (`getter(obj)` rather than `obj[keyPath]`) and it lets it return a `RLMSet *` rather than `id`.

Realm/RLMObjectStore.h
leemaguire: this exposes the method that allows you see if a Realm is in a transaction, the other `RLMVerifyInWriteTransaction` verifies if an attached object is in a transaction

Realm/ObjectServerTests/RLMCollectionSyncTests.mm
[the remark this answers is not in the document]
leemaguire: addressed this in #7190

Realm/ObjectServerTests/RLMCollectionSyncTests.mm
leemaguire: addressed this in #7190

Realm/RLMUtil.mm
leemaguire: true, I changed this to actually honour the create policy

Realm/ObjectServerTests/SwiftObjectServerTests.swift
leemaguire: Creating an object dynamically is fine. Getting the RLMDynamicObject to cast to a Swift DynamicObject, or even RealmSwiftObject is an issue.

Here is how I'm creating the object: https://github.com/realm/realm-cocoa/blob/1a1c0c7e7f95b7bbb2be7983ee8a8f36d7bb3f30/Realm/RLMObjectStore.mm#L183

leemaguire: The only thing I can think of to get around this and add the proper swift castings is to have a property on the realm instance itself that would indicate that the access came from a swift caller.

leemaguire: I have added https://github.com/realm/realm-cocoa/blob/c5fb724db3f3a1bc83e7ee5ba3ba14f0ca131dfc/RealmSwift/AnyRealmValue.swift#L148

leemaguire:
@tgoyne could you PTAL

Realm/RLMClassInfo.hpp
```diff
@@ -101,6 +115,15 @@ class RLMSchemaInfo {
...
     RLMClassInfo& operator[](NSString *name);
+    // Look up by table key, return none if its not present.
+    realm::util::Optional<RLMClassInfo&> operator[](realm::TableKey const& tableKey);
+
+    // Emplaces a locally derived object schema into RLMSchemaInfo. This is used
+    // when creating objects dynamically that are not registered in the Cocoa schema.
+    // Note: `RLMClassInfo` assumes ownership of `schema`.
+    void append_dynamic_object_schema(realm::ObjectSchema* schema,
+                                      RLMObjectSchema *objectSchema,
+                                      __unsafe_unretained RLMRealm *const target_realm);
```
tgoyne, line 126: `__unsafe_unretained` here doesn't do anything and we normally exclude it from declarations as it just makes things less readable (and potentially misleading).

Realm/RLMClassInfo.mm
```diff
@@ -95,6 +104,14 @@
     return *&it->second;
 }
 
+util::Optional<RLMClassInfo&> RLMSchemaInfo::operator[](realm::TableKey const& key) {
+    for (auto& pair : m_objects) {
+        if (pair.second.table()->get_key() == key)
+            return pair.second;
+    }
```
tgoyne, line 111:
```suggestion
    for (auto& pair : m_objects) {
        if (pair.second.objectSchema.tableKey == key)
            return pair.second;
    }
```

`table()` does a bit of actual work.

Realm/RLMObjectStore.mm
```diff
@@ -73,7 +74,7 @@ static inline void RLMVerifyRealmRead(__unsafe_unretained RLMRealm *const realm)
     }
 }
 
-static inline void RLMVerifyInWriteTransaction(__unsafe_unretained RLMRealm *const realm) {
+inline void RLMVerifyInWriteTransaction(__unsafe_unretained RLMRealm *const realm) {
```
tgoyne, line 77: Non-static inline in an implementation file doesn't make sense. If this is intended to be called from other files it shouldn't be inline.

Realm/RLMObjectStore.mm
```diff
@@ -170,6 +171,26 @@ void RLMAddObjectToRealm(__unsafe_unretained RLMObjectBase *const object,
...
 
+RLMObjectBase *RLMObjectFromObjLink(RLMRealm *realm, realm::ObjLink&& objLink, bool parentIsSwiftObject) {
+    if (auto tableInfo = realm->_info[objLink.get_table_key()]) {
+        return RLMCreateObjectAccessor(*tableInfo, objLink.get_obj_key().value);
+    } else {
+        // Construct the object dynamically.
+        // This code path should only be hit on first access of the object.
+        Class cls = parentIsSwiftObject ? [RealmSwiftObject class] : [RLMDynamicObject class];
+        realm::Schema const groupSchema = realm::ObjectStore::schema_from_group(realm->_realm->read_group());
+        realm::ObjectSchema *schema = new realm::ObjectSchema;
```
tgoyne, line 182:
```suggestion
        auto& group = realm->_realm->read_group();
        auto schema = std::make_unique<realm::ObjectSchema>(group, group.get_table_name(objLink.get_table_key()), objLink.get_table_key());
```

No need to construct the entire Schema just to get a single ObjectSchema, and the object should be passed around as a unique_ptr.

Realm/RLMQueryUtil.mm
```diff
@@ -1147,6 +1278,58 @@ auto collection_operation_expr(CollectionOperation operation) {
...
+                    add_numeric_constraint(type, operatorType,
+                                           collection_operation_expr<Mixed, Operation, IsLinkCollection>(collectionOperation),
+                                           Mixed(value_of_type<Decimal128>(rhs)));
+                }
+                else {
+                    throwException(@"Unsupported predicate value type",
+                                   @"Unsupported comparision value type for mixed. Value must be numeric.");
+                }
+            }
+            else if constexpr (std::is_empty_v<R>) {
```
tgoyne, line 1327: What's this a check for?

Realm/RLMUtil.mm
```diff
@@ -436,7 +383,68 @@ BOOL RLMIsRunningInPlayground() {
...
+                            realm::CreatePolicy createPolicy) {
+    if (!v || v == NSNull.null) {
+        return realm::Mixed();
+    }
+
+    RLMPropertyType type;
+    if ([v conformsToProtocol:@protocol(RLMValue)])
+        type = [v rlm_valueType];
+    else
+        REALM_TERMINATE("Unexpected Type");
```
tgoyne, line 397:
```suggestion
    REALM_ASSERT([v conformsToProtocol:@protocol(RLMValue)]);
    RLMPropertyType type = [v rlm_valueType];
```

if .. else terminate is just a longwinded way of writing an assert assuming it's supposed to be impossible to hit the else case regardless of what the user does. If the user passing an invalid value somewhere can hit this, then it needs to throw an exception instead.

Realm/RLMUtil.mm
```diff
@@ -436,7 +383,68 @@ BOOL RLMIsRunningInPlayground() {
...
+
+    switch (type) {
```
tgoyne, line 399: I think this whole switch is equivalent to the following:
```
RLMAccessorContext c{objBase->_info};
return switch_on_type(static_cast<realm::PropertyType>(type), [&](auto t) {
        return realm::Mixed(c.unbox<std::decay_t<decltype(*t)>>(v, createPolicy));
});
```

Realm/RLMValue.mm
tgoyne: I think `Any` is probably the least-bad option here. The options basically are:
1. Make the return type optional.
2. Add Null to RLMPropertyType
3. Add RLMMixedType
4. Just return Any

1 would be ideal but isn't really possible in obj-c. 2 is clearly incorrect as Null isn't a valid property type. 3 is probably the most correct but makes everything a lot more awkward. 4 is incorrect in a less problematic way and probably isn't an actual problem in practice.

RealmSwift/Optional.swift
```diff
@@ -43,14 +43,15 @@ extension Bool: RealmOptionalType {}
 
  To change the underlying value stored by a `RealmOptional` instance, mutate the instance's `value` property.
  */
-public final class RealmOptional<Value: RealmOptionalType>: RLMOptionalBase {
+@available(*, deprecated, message: "RealmOptional has been deprecated, use RealmProperty<T?> instead.")
```
tgoyne, line 46: `@available(*, deprecated, renamed: "RealmProperty")` makes Xcode have a fix-it to automatically update the name.

RealmSwift/Object.swift
```diff
@@ -421,6 +422,15 @@ public final class DynamicObject: Object {
         }
     }
 
+    public subscript(dynamicMember member: String) -> Any? {
```
tgoyne, line 425: Are there any tests using this? I didn't spot any, but I could have easily missed them with how much stuff is being added in this PR. Since it's just wrapping the existing functionality it doesn't need much.

Realm/RLMQueryUtil.mm
leemaguire: It is for checking if R is null

leemaguire: I can simplify this entire path

Realm/RLMUtil.mm
leemaguire: this would only work with realm objects as `objBase->_info` holds the class info required for RLMAccessorContext

RealmSwift/Optional.swift
leemaguire: would there be any issues with users having to manually mark the type as optional after? eg RealmOptional<Int> -> RealmProperty<Int?>`

RealmSwift/Object.swift
leemaguire: I have added `testDynamicMemberSubscript` to ObjectTests

Realm/RLMUtil.mm
tgoyne: Ah yes, we don't actually have objBase outside of the switch. Pushing that inside would give:

```
return switch_on_type(static_cast<realm::PropertyType>(type), util::overloads{[&](Obj*) {
    RLMAccessorContext c{objBase->_info};
    return realm::Mixed(c.unbox<Obj>(v, createPolicy));
}, [&](auto t) {
    RLMAccessorContext c;
    return realm::Mixed(c.unbox<std::decay_t<decltype(*t)>>(v));
});
```

Awkwardly it isn't possible to construct an accessor context inside the else case. In the property wrappers prototype I extracted [RLMStatelessAccessorContext](https://github.com/realm/realm-cocoa/blob/tg/property-wrapper/Realm/RLMAccessor.hpp#L43) which implements box/unbox on primitives specifically for this sort of thing.

RealmSwift/Optional.swift
tgoyne: I think we can still specify a message with renamed telling them that? It'd be nice if we could do it automatically, but it's probably still worth doing the fixit even if it's incomplete.

Realm/RLMUtil.mm
leemaguire: I changed things around, @tgoyne if you could take a look

Realm/RLMAccessor.mm
```diff
@@ -888,6 +890,83 @@ static auto to_optional(__unsafe_unretained id const value, Fn&& fn) {
     return to_optional(v, [&](__unsafe_unretained NSUUID *v) { return [v rlm_uuidValue]; });
 }
 
+#pragma mark - RLMStatelessAccessorContext::unbox -
```
tgoyne, line 893: This appears to be adding a second copy of the functions rather than moving them from RLMAccessorContext to RLMStatelessAccessorContext.

Realm/RLMClassInfo.mm
```diff
@@ -38,11 +38,11 @@
 : realm(realm), rlmObjectSchema(rlmObjectSchema), objectSchema(objectSchema) { }
 
 RLMClassInfo::RLMClassInfo(RLMRealm *realm, RLMObjectSchema *rlmObjectSchema,
-                           std::unique_ptr<const realm::ObjectSchema *> objectSchema)
+                           std::unique_ptr<realm::ObjectSchema> schema)
 : realm(realm),
 rlmObjectSchema(rlmObjectSchema),
-objectSchema(*objectSchema),
-dynamicObjectSchema(std::move(objectSchema)) { }
+objectSchema(std::move(&*schema)),
```
tgoyne, line 44: Newline before comma for member initializers.

The std::move() doesn't do anything here. Moving a pointer and copying it are the exact same operation.

Realm/Tests/RealmTests.mm
```diff
@@ -1002,42 +1002,43 @@ - (void)testBeginWriteTransactionFromWithinRealmChangedNotification {
...
-        [token invalidate];
-    };
-    token = [StringObject.allObjects addNotificationBlock:block];
-    [self waitForExpectationsWithTimeout:2.0 handler:nil];
-
-    createObject();
-    expectation = [self expectationWithDescription:@""];
-    [self waitForExpectationsWithTimeout:2.0 handler:nil];
-}
+// FIXME:
```
tgoyne, line 1005: Rather than commenting this out, rename the method. This makes it so that the test still gets compiled and avoids causing problems for `git blame`.

leemaguire: the reason why I commented it out is because this method stalls the tests

tgoyne: Renaming it to make it not start with `test` makes it so that the test doesn't run.

Realm/RLMAccessor.mm
leemaguire: I have moved the specialisations to RLMStatelessAccessorContext

leemaguire:
@tgoyne PTAL

Realm/RLMAccessor.hpp
```diff
@@ -40,19 +41,8 @@ struct RLMOptionalId {
...
-    // Accessor context interface
-    RLMAccessorContext(RLMAccessorContext& parent, realm::Obj const& parent_obj, realm::Property const& property);
-
-    id box(realm::object_store::Dictionary&&);
-    id box(realm::object_store::Set&&);
-    id box(realm::List&&);
-    id box(realm::Results&&);
-    id box(realm::Object&&);
-    id box(realm::Obj&&);
+// The subset of RLMAccessorContext which does not require any member variables
```
jsflax, line 44: Comments like this really need a "why". The "what" is fairly useless and the name of the struct tells me nothing.

Realm/RLMSwiftValueStorage.h
```diff
@@ -0,0 +1,50 @@
+////////////////////////////////////////////////////////////////////////////
+//
+// Copyright 2015 Realm Inc.
```
jsflax, line 3: wrong date for new file

Realm/Tests/ObjectTests.m
```diff
@@ -607,25 +607,59 @@ static void addProperty(Class cls, const char *name, const char *type, size_t si
...
+        [realm beginWriteTransaction];
+        id object = [objectClass createInRealm:realm withValue:@{@"objectCol": [[objectClass alloc] init], @"intCol": @17}];
+        RLMObjectSchema *schema = [object objectSchema];
+        XCTAssertNotNil(schema[@"objectCol"]);
+        XCTAssertNotNil(schema[@"intCol"]);
+        XCTAssert([[object objectCol] isKindOfClass:objectClass]);
+        XCTAssertEqual([object intCol], 17);
+        [realm commitWriteTransaction];
+    }
+    // Creating the class via ObjC runtime can cause issues with other tests. Specifically
```
tgoyne, line 631: This all seems very dubious? There shouldn't be any difference between classes created explicitly via the obj-c runtime functions and ones created via loading a bundle. If there is a difference, then this test appears to just be showing that our code is broken, and deleting the classes is just working around the bug that the test uncovered, since we do intend to support dynamically created RLMObject subclasses.

Realm/Tests/Swift/SwiftSchemaTests.swift
```diff
@@ -222,6 +224,11 @@ class SwiftRLMSchemaTests: RLMMultiProcessTestCase {
     }
 
     func testAccessSchemaCreatesObjectWhichAttempsInsertionsToArrayProperty() {
+        // This test depends on InitAppendsToArrayValue having a nil _objectSchema.
+        // There is a chicken and egg race problem here as InitAppendsToArrayValue may
+        // have its object schema registered before InitAppendsToArrayProperty.
```
tgoyne, line 229: This again is a test revealing a bug. `InitAppendsToArrayValue` must have a nil `_objectSchema` at the point where the getter is called because the getter needs to be called before we can construct the ObjectSchema for that type. If it's non-nil, that means we assigned a garbage value to the ivar.

Realm/RLMAccessor.hpp
```diff
@@ -103,17 +79,53 @@ class RLMAccessorContext : public RLMStatelessAccessorContext {
...
+
+    RLMOptionalId value_for_property(id dict, realm::Property const&, size_t prop_index);
+    RLMOptionalId default_value_for_property(realm::ObjectSchema const&,
+                                             realm::Property const& prop);
+
+    template<typename T>
+    T unbox(id v, realm::CreatePolicy = realm::CreatePolicy::Skip, realm::ObjKey = {}) {
+        if constexpr(std::is_same_v<T, realm::Mixed>) {
+            unbox<realm::Mixed>(v);
+        }
```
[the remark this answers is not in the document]
leemaguire: In different environments build errors were thrown if `realm::Mixed RLMStatelessAccessorContext::unbox(id)` did not exist, so this shoehorns this method to invoke the one defined in `RLMAccessorContext` and not `RLMStatelessAccessorContext`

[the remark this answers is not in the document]
tgoyne: The correct way to do that is to define an explicit instantiation of the unbox() function as is done for Obj. Currently this just calls itself and then discards the result of doing so.

Realm/Tests/ObjectTests.m
leemaguire: I located the issue to be at ```
util::Optional<RLMClassInfo&> RLMSchemaInfo::operator[](realm::TableKey const& key) {
    for (auto& pair : m_objects) {
        if (pair.second.objectSchema->table_key == key)
            return pair.second;
    }
    return util::none;
}
```
where the `table_key` is a garbage value for this runtime registered object. So I suppose guarding against that somehow is the fix

File: Realm/Tests/Swift/SwiftSchemaTests.swift (hunk 58 above)
Comment (leemaguire) [reply Realm/Tests/Swift/SwiftSchemaTests.swift:229]: The strange thing about this test is that it will pass when running the test suite in the Xcode project / swift package from Xcode. But it will fail when run from `build.sh test-swiftpm`

**Comment by tgoyne** [comment]
One thing I notice while poking at it is that the whole dynamically created schema thing is only tested in the sync tests but it is not sync-specific. Anything which doesn't require a server to test should have non-server tests.

File: Realm/Tests/Swift/SwiftSchemaTests.swift (hunk 58 above)
Comment (tgoyne) [reply Realm/Tests/Swift/SwiftSchemaTests.swift]: I don't see any failures with `test-swiftpm` or `test-swiftpm-address` when I remove the cleanup code, and I don't really see how extra classes could be producing failures there.

File: Realm/RLMClassInfo.hpp [hunk 60]
```diff
@@ -101,6 +116,15 @@ class RLMSchemaInfo {
 
     // Look up by name, throwing if it's not present
     RLMClassInfo& operator[](NSString *name);
+    // Look up by table key, return none if its not present.
+    realm::util::Optional<RLMClassInfo&> operator[](realm::TableKey const& tableKey);
+
+    // Emplaces a locally derived object schema into RLMSchemaInfo. This is used
+    // when creating objects dynamically that are not registered in the Cocoa schema.
+    // Note: `RLMClassInfo` assumes ownership of `schema`.
+    void append_dynamic_object_schema(std::unique_ptr<realm::ObjectSchema> schema,
```
tgoyne, line 125: The other functions here are camelcase.

Realm/RLMClassInfo.mm
```diff
@@ -37,6 +37,14 @@
                            const realm::ObjectSchema *objectSchema)
 : realm(realm), rlmObjectSchema(rlmObjectSchema), objectSchema(objectSchema) { }
 
+RLMClassInfo::RLMClassInfo(RLMRealm *realm, RLMObjectSchema *rlmObjectSchema,
+                           std::unique_ptr<realm::ObjectSchema> schema)
+: realm(realm),
+rlmObjectSchema(rlmObjectSchema),
```
tgoyne, line 43: Commas at the start of the line for initializer lists,

Realm/Tests/Swift/SwiftSchemaTests.swift
leemaguire: hmm I also no longer see it

leemaguire: I can get it to fail when I disable the sync tests in the swift package and then run it with the asan.

Realm/RLMObjectBase.mm
```diff
@@ -50,7 +50,12 @@ static bool isManagedAccessorClass(Class cls) {
     return strncmp(className, accessorClassPrefix, sizeof(accessorClassPrefix) - 1) == 0;
 }
 
+static bool preventInitObjectSchema;
 static bool maybeInitObjectSchemaForUnmanaged(RLMObjectBase *obj) {
+    if (preventInitObjectSchema) {
```
tgoyne, line 55: This change doesn't make sense to me. -[RLMDynamicObject sharedSchema] returns nil, so this function already does nothing for RLMDynamicObject. Maybe the Swift version is missing that override?

Realm/RLMClassInfo.mm
```diff
@@ -123,6 +123,9 @@
 
     m_objects.reserve(schema.size());
     for (RLMObjectSchema *rlmObjectSchema in rlmSchema.objectSchema) {
+        if (schema.find(rlmObjectSchema.objectName.UTF8String) == schema.end()) {
```
tgoyne, line 126: This could avoid the double lookup pretty easily.

Realm/RLMObjectBase.mm
leemaguire: removed and added the override for `sharedSchema` on `RealmSwiftObject`

CHANGELOG.md
```diff
@@ -40,6 +44,19 @@ x.y.z Release notes (yyyy-MM-dd)
...
+* Fixed an issue where creating an object after file format upgrade may fail
+  with assertion "Assertion failed: lo() <= std::numeric_limits<uint32_t>::max()"
+ ([#4295](https://github.com/realm/realm-core/issues/4295), since v5.0.0)
+
+<!-- ### Breaking Changes - ONLY INCLUDE FOR NEW MAJOR version -->
+
+### Compatibility
+* Realm Studio: 10.0.0 or later.
+* APIs are backwards compatible with all previous releases in the 10.x.y series.
+* Carthage release for Swift is built with Xcode 12.3.
```
tgoyne, line 56:
```suggestion
* Carthage release for Swift is built with Xcode 12.4.
```

Realm/ObjectServerTests/RLMCollectionSyncTests.mm
```diff
@@ -76,7 +77,9 @@ - (void)roundTripWithPropertyGetter:(RLMSet *(^)(id))propertyGetter
             RLMResults<RLMSetSyncObject *> *results
                 = [RLMSetSyncObject allObjectsInRealm:realm];
             if (RLMSetSyncObject *obj = results.firstObject) {
-                if (propertyGetter(obj).count == 0) {
+                CHECK_COUNT(1, RLMSetSyncObject, realm);
+                if (propertyGetter(obj).count == 0
+                    && otherPropertyGetter(obj).count == 0) {
```
tgoyne, line 82: Wrapping the line here makes it look like this line is part of the block following the if rather than part of the condition.

Realm/ObjectServerTests/RLMObjectServerTests.mm
```diff
@@ -779,6 +803,47 @@ - (void)testDeleteObjects {
...
+        Person *p = [Person john];
+        [o.anySet addObjects:@[p]];
+        obj.anyCol = o;
+        obj.objectCol = p;
+        [realm beginWriteTransaction];
+        [realm addObject:obj];
+        [realm commitWriteTransaction];
+        [self waitForUploadsForRealm:realm];
+        CHECK_COUNT(1, AllTypesSyncObject, realm);
+    }
```
tgoyne, line 844: This test would be a little easier to follow with a more linear flow:

```suggestion
    if (!self.isParent) {
        c.objectClasses = @[Person.self,
                            AllTypesSyncObject.self,
                            RLMSetSyncObject.self];
        RLMRealm *realm = [RLMRealm realmWithConfiguration:c error:nil];
        AllTypesSyncObject *obj = [[AllTypesSyncObject alloc] initWithValue:[AllTypesSyncObject values:0]];
        RLMSetSyncObject *o = [RLMSetSyncObject new];
        Person *p = [Person john];
        [o.anySet addObjects:@[p]];
        obj.anyCol = o;
        obj.objectCol = p;
        [realm beginWriteTransaction];
        [realm addObject:obj];
        [realm commitWriteTransaction];
        [self waitForUploadsForRealm:realm];
        CHECK_COUNT(1, AllTypesSyncObject, realm);

        return;
    }
    RLMRunChildAndWait();

    c.objectClasses = @[Person.self, AllTypesSyncObject.self];
    RLMRealm *realm = [RLMRealm realmWithConfiguration:c error:nil];
    [self waitForDownloadsForRealm:realm];
    RLMResults <AllTypesSyncObject *> *res = [AllTypesSyncObject allObjectsInRealm:realm];
    AllTypesSyncObject *o = res.firstObject;
    Person *p = o.objectCol;
    RLMSet<RLMValue> *anySet = ((RLMObject *)o.anyCol)[@"anySet"];
    XCTAssertTrue([anySet.allObjects[0][@"firstName"] isEqualToString:p.firstName]);
    [realm beginWriteTransaction];
    anySet.allObjects[0][@"firstName"] = @"Bob";
    [realm commitWriteTransaction];
    XCTAssertTrue([anySet.allObjects[0][@"firstName"] isEqualToString:p.firstName]);
    CHECK_COUNT(1, AllTypesSyncObject, realm);
```

This makes it so that the steps happen in the same order as the code even though some of it is happening in a different process.

Realm/ObjectServerTests/SwiftObjectServerTests.swift
```diff
@@ -2091,6 +2115,65 @@ class SwiftMongoClientTests: SwiftSyncTestCase {
...
 
+class AnyRealmValueSyncTests: SwiftSyncTestCase {
+    /// The purpose of this test is to confirm that when an Object is set on a mixed Column and an old
+    /// version of an app does not have that Realm Object / Schema we can still access that object via
+    /// `AnyRealmValue.dynamicSchema`.
+    func testMissingSchema() {
+        do {
+            let user = try logInUser(for: basicCredentials())
+
+            if isParent {
```
tgoyne, line 2126: Similar note as above where putting the child block first makes it a bit easier to follow.

Realm/RLMArray.mm
```diff
@@ -215,7 +215,7 @@ void RLMArrayValidateMatchingObjectType(__unsafe_unretained RLMArray *const arra
         @throw RLMException(@"Invalid nil value for array of '%@'.",
                             array->_objectClassName ?: RLMTypeToString(array->_type));
     }
-    if (array->_type != RLMPropertyTypeObject) {
+    if ((array->_type == RLMPropertyTypeAny) || (array->_type != RLMPropertyTypeObject)) {
```
tgoyne, line 218: The first half of this conditional isn't doing anything. `array->_type == RLMPropertyTypeAny` can never be true without `array->_type != RLMPropertyTypeObject` also being true.

Realm/RLMClassInfo.hpp
```diff
@@ -101,6 +116,15 @@ class RLMSchemaInfo {
 
     // Look up by name, throwing if it's not present
     RLMClassInfo& operator[](NSString *name);
+    // Look up by table key, return none if its not present.
+    realm::util::Optional<RLMClassInfo&> operator[](realm::TableKey const& tableKey);
```
tgoyne, line 120: Optional references have strange and potentially surprising semantics, so outside of generic programming contexts just using a pointer is preferable.

Realm/RLMObjectBase.mm
```diff
@@ -24,7 +24,8 @@
 #import "RLMObjectSchema_Private.hpp"
 #import "RLMObjectStore.h"
 #import "RLMObservation.hpp"
-#import "RLMOptionalBase.h"
+#import "RLMSwiftValueStorage.h"
+#import "RLMSwiftValueStorage.h"
```
tgoyne, line 28: Duplicated import

Realm/RLMObjectBase.mm
```diff
@@ -750,6 +751,9 @@ uint64_t RLMObjectBaseGetCombineId(__unsafe_unretained RLMObjectBase *const obj)
 }
 
 @implementation RealmSwiftObject
++ (RLMObjectSchema *)sharedSchema {
```
tgoyne, line 754: This is not a correct implementation for RealmSwiftObject. DynamicObject is what needs this, not Object.

Realm/RLMSet.mm
```diff
@@ -424,7 +410,7 @@ void RLMSetValidateMatchingObjectType(__unsafe_unretained RLMSet *const set,
         @throw RLMException(@"Invalid nil value for set of '%@'.",
                             set->_objectClassName ?: RLMTypeToString(set->_type));
     }
-    if (set->_type != RLMPropertyTypeObject) {
+    if ((set->_type == RLMPropertyTypeAny) || (set->_type != RLMPropertyTypeObject)) {
```
tgoyne, line 413: The first half of this isn't doing anything.

Realm/RLMUtil.hpp
```diff
@@ -17,11 +17,14 @@
...
 #import <Realm/RLMConstants.h>
-#import <Realm/RLMOptionalBase.h>
+#import <Realm/RLMSwiftValueStorage.h>
+#import <Realm/RLMValue.h>
+
 #import <objc/runtime.h>
 
 #import <realm/array.hpp>
 #import <realm/binary_data.hpp>
+#import <realm/object-store/object.hpp>
```
tgoyne, line 27: This doesn't appear to do anything which would need this?

Realm/Tests/QueryTests.m
```diff
@@ -1370,26 +1578,49 @@ - (void)testDecimalQuery {
     [realm beginWriteTransaction];
     [DecimalObject createInRealm:realm withValue:@[@"-Inf"]];
     [DecimalObject createInRealm:realm withValue:@[@"Inf"]];
+    // 123456789.123456789e12345 represents Inf
```
tgoyne, line 1581: This was actually just an incorrect test which is now fixed on master.

RealmSwift/RealmProperty.swift
```diff
@@ -0,0 +1,89 @@
...
+
+    public func encode(to encoder: Encoder) throws {
+        try self.value.encode(to: encoder)
+    }
+}
+
+/// A protocol describing types that can parameterize a `RealmPropertyType`.
+public protocol RealmPropertyType {}
+/// A protocol describing types that can be represented as optional in a `RealmProperty<>`
+//internal protocol OptionalRealmPropertyType {}
```
tgoyne, line 86: Commented code

Realm/RLMClassInfo.hpp
leemaguire: This will create issues for `RLMObjectFromObjLink` where the purpose of returning optional is so that if the table key does not exist in RLMSchemaInfo we can create the object dynamically.

Realm/RLMUtil.hpp
leemaguire: without this import the compiler complains about not being able to find `realm::CreatePolicy`

Realm/RLMObjectBase.mm
leemaguire: So, another weird issue I have found is that if the 'dynamic object from mixed' code is run (`RLMObjectFromObjLink`) before the test `testInvalidObjects()` objects with subclasses will have duplicated properties (only properties from the base class are showing up). Preventing `maybeInitObjectSchemaForUnmanaged` being called from `RLMObjectFromObjLink` solves the issue but its probably pointing to something more sinister that I have yet to find the root cause for.

leemaguire: Add this solved the failing test: https://github.com/realm/realm-cocoa/blob/53244c2cc555c5c99a304cb0c796f49076407757/Realm/RLMObjectBase.mm#L759

So now when constructing the dynamic mixed object, the accessor class will be RealmDynamicSwiftObject

Realm/RLMClassInfo.hpp
tgoyne: Returning null and returning none do the exact same thing there.

Realm/RLMObjectBase.h
```diff
@@ -95,6 +95,9 @@ NS_ASSUME_NONNULL_BEGIN
 @interface RealmSwiftObject : RLMObjectBase
 @end
 
+@interface RealmDynamicSwiftObject : RealmSwiftObject
```
tgoyne, line 98: Why does this need a new type? `DynamicObject` is already there for this.

Realm/RLMObjectBase_Private.h
```diff
@@ -26,6 +26,7 @@ NS_ASSUME_NONNULL_BEGIN
 @interface RLMObjectBase ()
 + (void)initializeLinkedObjectSchemas;
 + (bool)isEmbedded;
++ (bool)isDynamic;
```
tgoyne, line 29: The thing this is used for is already covered by `shouldIncludeInDefaultSchema`.

Realm/RLMObjectBase.mm
```diff
@@ -353,6 +353,10 @@ + (bool)isEmbedded {
     return false;
 }
 
++ (bool)isDynamic {
```
tgoyne, line 356: Leftover addition.

tgoyne (approved):

PR #7191, opened 2021-03-25
Add UUID Sync Tests

dianaafanador3:
This PR adds 
* Sync tests for objectId/UUID/String/int primary keys objects
* Sync tests for objectId/UUID/String/int partition values types .
* Support for querying using UUID as primaryKey
 `realm.object(ofType: SwiftPrimaryUUIDObject.self, forPrimaryKey: UUID(uuidString: "4ee1fa48-8b23-11eb-8dcd-0242ac130003")!)`
* Additional object test for Swift/Int/UUID as primary keys
* Additional test for indexable UUID

Updated baas commit to use latest version to 31/03/2021
realm-Core branch: v11


Realm/ObjectServerTests/RLMSyncTestCase.h
```diff
@@ -40,6 +39,7 @@ NS_ASSUME_NONNULL_BEGIN
 @property NSString *firstName;
 @property NSString *lastName;
 
+- (instancetype) initWithPrimaryKey:(RLMObjectId *)primaryKey age:(NSInteger)strCol firstName:(NSString *)intCol lastName:(NSString *)intCol;
```
tgoyne, line 42: No space between the type and the name in obj-c methods.

Realm/ObjectServerTests/RLMSyncTestCase.mm
```diff
@@ -200,9 +296,14 @@ - (RLMUser *)anonymousUser {
 }
 
 - (RLMCredentials *)basicCredentialsWithName:(NSString *)name register:(BOOL)shouldRegister {
+    return [self basicCredentialsWithName:name register:shouldRegister app:nil];
+}
+
+- (RLMCredentials *)basicCredentialsWithName:(NSString *)name register:(BOOL)shouldRegister app:(nullable RLMApp*) app {
```
tgoyne, line 302: Space before \* in obj-c

Realm/ObjectServerTests/RLMSyncTestCase.mm
```diff
@@ -200,9 +296,14 @@ - (RLMUser *)anonymousUser {
...
     if (shouldRegister) {
         XCTestExpectation *expectation = [self expectationWithDescription:@""];
-        [self.app.emailPasswordAuth registerUserWithEmail:name password:@"password" completion:^(NSError *error) {
+        RLMApp *currentApp = app != nil ? app : self.app;
```
tgoyne, line 305: `app ?: self.app`

Realm/ObjectServerTests/RLMSyncTestCase.mm
```diff
@@ -559,6 +639,55 @@ - (NSTask *)childTask {
     return [self childTaskWithAppIds:_appId ? @[_appId] : @[]];
 }
 
+- (RLMApp *)createAppForPartition:(id<RLMBSON>)partition {
```
tgoyne, line 642: This being `createAppForPartition:` confused me a bit, as I expected it to be doing something which actually required an app-per-partition, while it appears to only need an app per partition key type. I think making it `createAppForPartitionType:` instead would be a bit clearer.

Realm/ObjectServerTests/RLMSyncTestCase.mm
```diff
@@ -559,6 +639,55 @@ - (NSTask *)childTask {
...
+    syncManager.userAgent = self.name;
+
+    return _app;
+}
+
+- (NSString *)partitionBsonType:(RLMBSONType)type {
+    switch(type){
+        case RLMBSONTypeString:
+            return @"string";
+            break;
```
tgoyne, line 674: breaks here are redundant

Realm/ObjectServerTests/RealmServer.swift
```diff
@@ -518,7 +518,7 @@ public class RealmServer: NSObject {
     }
 
     /// Create a new server app
-    @objc public func createApp() throws -> AppId {
+    @objc public func createApp(forBSONType bsonType: String) throws -> AppId {
```
tgoyne, line 521: Swift naming usually doesn't use `for` here.

Realm/ObjectServerTests/RealmServer.swift
```diff
@@ -684,25 +686,110 @@ public class RealmServer: NSObject {
             "relationships": [:]
         ]
 
-        let userDataRule: [String: Any] = [
+        let uUIDPrimaryKeyObjectRule: [String: Any] = [
```
tgoyne, line 689: `uuidPrimaryKeyObjectRule`

Initialisms always have the same capitalization for the entire thing. In obj-c they're always all uppercase unless they're at the start of a local variable name. In Swift they're uppercase unless they're at the start of any non-type name.

dianaafanador3:
@tgoyne comments addressed, PTAL

Realm/ObjectServerTests/RLMObjectServerPartitionTests.mm
```diff
@@ -0,0 +1,108 @@
...
+@interface RLMObjectServerPartitionTests : RLMSyncTestCase
+@end
+
+@implementation RLMObjectServerPartitionTests
+
+- (void)roundTripForPartitionValue:(id<RLMBSON>)value  {
+    NSString *appId;
+    if (self.isParent) {
+        NSError *error;
+        appId = [RealmServer.shared createAppForBSONType: [self partitionBsonType:value]  error:&error];
```
tgoyne, line 46: No space after colon for obj-c message sends.

Realm/ObjectServerTests/RLMObjectServerPartitionTests.mm
```diff
@@ -0,0 +1,108 @@
...
+
+        if (error) {
+            XCTFail(@"Could not create app for partition value %@d", value);
```
tgoyne, line 49: You'll want to also return after this so that it doesn't try to run the actual test.

tgoyne (approved):

PR #7196, opened 2021-04-01
Auto Open Prototype

jsflax:
Info TK.


phatblat:
If I'm understanding this correctly, this property wrapper would like the following in use:

```swift
@AutoOpen(appId: "app-id", partitionValue: "PUBLIC")
var syncedRealm: Realm
```

Where does authentication come in? The current flow involves the following steps:

1. Create `App` with `appId`
2. Log in
3. Set partition value on the configuration
4. Open Realm

Looks like `AutoOpen` will take care of 3 of these, but wondering how `Credentials` should be provided.

tgoyne:
The idea is that if you don't specify a user, it'll open the Realm using `app.currentUser`, and if that's `nil` then wait for you to log in a user for that app. Changing the App's current user would reopen the Realm using that new user. More complicated scenarios like having two different logged-in users for the same App with Realms open at the same time would probably require continuing to do things manually.

phatblat:
Ah, that makes perfect sense @tgoyne. Thanks!

PR #7206, opened 2021-04-12
Upgrade to core 10.6.0

tgoyne:


Realm/RLMDecimal128.h
```diff
@@ -49,9 +49,9 @@ NS_ASSUME_NONNULL_BEGIN
 
 /// Parses the given string to a RLMDecimal128.
 ///
-/// Returns `nil` and sets `error` if the string cannot be parsed as a
-/// RLMDecimal128.
-- (nullable instancetype)initWithString:(NSString *)string error:(NSError *_Nullable*)error;
```
tgoyne, line 54: The old documentation here was incorrect; we actually only returned `nil` and set `error` for syntactically valid numbers which were outside the range of what the old parsing code supported, and we already returned nan for invalid syntax.

Realm/RLMDecimal128.mm
```diff
@@ -73,17 +73,9 @@ - (instancetype)initWithNumber:(NSNumber *)number {
...
     if ((self = [self init])) {
-        try {
-            _value = realm::Decimal128(string.UTF8String);
-        }
-        catch (std::exception const& e) {
-            if (error) {
-                *error = RLMMakeError(RLMErrorInvalidInput, e);
-            }
-            return nil;
-        }
```
tgoyne, line 86: The only thing which threw here was out-of-range errors that should have been accepted.

Realm/Tests/QueryTests.m
```diff
@@ -1307,7 +1307,7 @@ - (void)testDecimalQuery {
     [realm beginWriteTransaction];
     [DecimalObject createInRealm:realm withValue:@[@"-Inf"]];
     [DecimalObject createInRealm:realm withValue:@[@"Inf"]];
-    [DecimalObject createInRealm:realm withValue:@[@"123456789.123456789e12345"]];
```
tgoyne, line 1310: This number was overflowing and becoming negative with the old parser, while with the new parser it's correctly rounded to Inf. The test doesn't actually want an Inf value so I made the number fit within the representable range instead.

jsflax (approved):

PR #7233, opened 2021-04-30
Rework how obj-c accesses Swift-only types

tgoyne:
This is most of the changes to existing code required for property wrappers. This PR should mkae no functional changes to how anything works currently, but in addition to adding the flexibility needed for property wrappers I think it makes the existing code generally easier to follow.

The core idea here is to expand the SwiftAccessor object used for LinkingObjects properties to cover *all* interactions with Swift-defined types. Currently we have a lot of code which reads an ivar and then does different things with it based on if it's optional, an array, or a set. This has a few problems:
1. It means that all ivars must have a base class which obj-c can interact with, and therefore all non-bridged property types must be classes rather than structs. Using classes for property wrappers is infeasible performance-wise, and using classes for RealmOptional/RealmValue is a pre-existing performance problem we'd like to fix.
2. The logic for dealing with a given property type is scattered all over the place. There's several places where we read an ivar and go "if this is optional handle that, if it's an array do something else, if it's a set do a third thing". Ideally all of the set-related logic should be in one place, and adding new types shouldn't require touching every place that reads/writes properties.

Instead of doing this, the SwiftAccessor interface now abstracts that all away from obj-c, and obj-c now just calls the `initialize()` function supplied by that to do things like initialize a managed list, rather than constructing a RLMManagedArray and setting the appropriate property on the List directly. This ensures that all of the logic related to dealing with List properties from obj-c appears in a single place, and adding new Swift-only types does not require any obj-c changes.

There is also some other incidental refactoring: the changes in RLMAccessor.mm are just shuffling around some code to eliminate some duplicated logic. In the current version of the code the duplication wasn't particularly bad but PWs add a bunch of functions that Swift calls to get/set values which are one line with the refactoring and considerably more without it.

I also introduced an Impl directory in RealmSwift and moved a bunch of code which isn't part of the public API there. Originally we had no such distinction because RealmSwift didn't really have any code other than the public API, but that's no longer the case. Sticking it in a separate directory is hopefully useful for organization and also lets us just exclude that directory from jazzy rather than marking everything as nodoc.

This probably has some exciting conflicts with the dictionary branch, but it should also make things simpler in the places where they conflict...


Realm/RLMCollection.mm
```diff
@@ -198,22 +198,17 @@ NSUInteger RLMFastEnumerate(NSFastEnumerationState *state,
...
-                for (size_t i = 0; i < count; ++i) {
-                    RLMSwiftCollectionBase *base = [[cls alloc] init];
-                    base._rlmCollection =
-                        RLMManagedCollectionFromCollection<std::decay_t<decltype(*type)>,
-                                                           std::decay_t<decltype(*objcType)>>
-                                                           (info, collection.get(i), prop);
-                    [array addObject:base];
-                }
-            });
+        if (prop.collection && prop.swiftAccessor) {
```
leemaguire, line 201: nice cleanup

Realm/Realm.modulemap
```diff
@@ -9,14 +9,14 @@ framework module Realm {
         header "RLMApp_Private.h"
         header "RLMArray_Private.h"
         header "RLMCollection_Private.h"
-        header "RLMObject_Private.h"
-        header "RLMObjectBase_Dynamic.h"
         header "RLMObjectBase_Private.h"
         header "RLMObjectSchema_Private.h"
         header "RLMObjectStore.h"
+        header "RLMObjectBase_Dynamic.h"
```
leemaguire, line 15: why have these headers moved position?

RealmSwift/EmbeddedObject.swift
```diff
@@ -137,34 +137,13 @@ extension EmbeddedObject: RealmCollectionValue {
     /// Returns or sets the value of the property with the given name.
     @objc open subscript(key: String) -> Any? {
         get {
-            if realm == nil {
-                return value(forKey: key)
-            }
-            return dynamicGet(key: key)
+            dynamicGet(object: self, key: key)
```
leemaguire, line 140: should `return` precede this getter for consistency purposes?

RealmSwift/List.swift
```diff
@@ -433,6 +433,10 @@ public final class List<Element: RealmCollectionValue>: RLMSwiftCollectionBase {
         return Element._rlmArray()
     }
 
+    @objc public override class func backingCollectionType() -> AnyClass {
```
leemaguire, line 436: should this not be `internal`?

leemaguire (commented):
implementation looks good. The refactor makes things a lot more understandable. I have left some minor comments

RealmSwift/List.swift
tgoyne: Ideally it would be but it can't be since it's overriding a base class method. It should be underscore-prefixed though.

Realm/Realm.modulemap
tgoyne: I added some things, sorted the list, then removed the new things as part of reducing this branch to just the changes to existing functionality, leaving the list unchanged other than being sorted.

RealmSwift/List.swift
leemaguire: Ahh I see. Does it require any doc? I see its missing `:nodoc:`

RealmSwift/Optional.swift
```diff
@@ -43,7 +43,7 @@ extension Bool: RealmOptionalType {}
 
  To change the underlying value stored by a `RealmOptional` instance, mutate the instance's `value` property.
  */
-@available(*, deprecated, renamed: "RealmProperty", message: "RealmOptional has been deprecated, use RealmProperty<T?> instead.")
+@available(*, deprecated, renamed: "RealmProperty", message: "RealmOptional<T> has been deprecated, use RealmProperty<T?> instead.")
```
jsflax, line 46: `T?` seems wrong? If we go by other wrappers, it would be represented as `T` and can be optional or not optional.

leemaguire: with RealmOptional T was an optional type. With the introduction of RealmProperty you are able to box optional and non optional properties

jsflax: Ah, I see. It's a catchall wrapper now. This notation does make sense

tgoyne: Yeah, I'm trying to hint "you need to add a ? here" because it isn't quite just renaming the existing type.

RealmSwift/Impl/SchemaDiscovery.swift
```diff
@@ -0,0 +1,152 @@
...
+    let ignoredPropNames: Set<String>
+    let columnNames = cls._realmColumnNames()
+    if let realmObject = object as? Object {
+        indexedProperties = Set(type(of: realmObject).indexedProperties())
+        ignoredPropNames = Set(type(of: realmObject).ignoredProperties())
+    } else {
+        indexedProperties = Set()
+        ignoredPropNames = Set()
+    }
+    return Mirror(reflecting: object).children.filter { (prop: Mirror.Child) -> Bool in
```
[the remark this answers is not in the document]
tgoyne: This is just the existing schema discovery code moved to a new file with no changes other than moving it to a separate function.

RealmSwift/Impl/SchemaDiscovery.swift
```diff
@@ -0,0 +1,152 @@
...
+    }.compactMap { prop in
+        guard let label = prop.label else { return nil }
+        var rawValue = prop.value
+        if let value = rawValue as? RealmEnum {
+            rawValue = type(of: value)._rlmToRawValue(value)
+        }
+
+        guard let value = rawValue as? _RealmSchemaDiscoverable else {
+            if class_getProperty(cls, label) != nil {
+                throwRealmException("Property \(cls).\(label) is declared as \(type(of: prop.value)), which is not a supported managed Object property type. If it is not supposed to be a managed property, either add it to `ignoredProperties()` or do not declare it as `@objc dynamic`. See https://realm.io/docs/swift/latest/api/Classes/Object.html for more information.")
```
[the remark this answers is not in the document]
tgoyne: Currently how that works is there's a [new much shorter discovery function for the new stuff](https://github.com/realm/realm-cocoa/blob/tg/property-wrapper/RealmSwift/Impl/SchemaDiscovery.swift#L33) and then this function is called only if that doesn't find anything.

jsflax (approved):
There's noticeably no testing added, but considering what this reworks, as long as our current test suite is passing, that seems fine.

tgoyne:
I did some more refactoring and now RealmCollectionValue doesn't have its own type -> PropertyType mapping. While this is all theoretically covered by our existing test suite, I think I am going to add some more focused unit tests for the new code.

leemaguire (approved):

PR #7257, opened 2021-05-17
Fix sync test failures

tgoyne:
Developer mode for Mixed is currently broken and results in the schema having a null typed property, which hits an assertion failure when it's synchronized to the client. Defining a rule setting the server-side schema explicitly works around this. Manually writing rules for all of the types we use is obnoxious, so I automated creating rules for all of the sync-compatible types in the test schema.


Realm/ObjectServerTests/RealmServer.swift
```diff
@@ -51,6 +52,108 @@ extension URLSession {
...
+        }
+
+        return [
+            "bsonType": type
+        ]
+    }
+}
+
+private extension ObjectSchema {
+    func stitchRule(_ partitionKeyType: String, _ schema: Schema, id: String? = nil) -> [String: Any] {
```
jsflax, line 107: Nice.

jsflax (approved):

Issue #7234, opened 2021-04-30
Realm crashes as it refers to the SwiftUI framework when run on older iOS versions


anlaital:
## Goals

Run an app that uses RealmSwift as a dependency on an iOS 11.4 simulator.

## Expected Results

Project can be run and it does not crash.

## Actual Results

Launching the app in the simulator crashes with the following error message:

```
dyld: Library not loaded: /System/Library/Frameworks/SwiftUI.framework/SwiftUI
  Referenced from: /.../myApp.app/Frameworks/RealmSwift.framework/RealmSwift
  Reason: image not found
```

This issue seems to affect all version of iOS 11 and 12 that do not contain the SwiftUI framework. The bug was first introduced in release v10.6.0 and affects all later releases. Last working release is therefore v10.5.2.

## Steps for others to Reproduce

I suppose this can be reproduced simply by creating a new iOS application that uses the latest (10.7.4) RealmSwift as a dependency.

## Code Sample

N/A

## Version of Realm and Tooling

Realm framework version: 10.7.4

Xcode version: 12.5

iOS/OSX version: iOS 11.4 simulator

Dependency manager + version: CocoaPods 1.10.1

jongarate:
+1

kubankredit:
+1 I've added SwiftUI.framework as Optional in Link libraries phase in Xcode and app crashes anyway

aelam:
for cocoapods users
```
post_install do |installer|
  installer.pods_project.targets.each do |target|
    if target.name == 'RealmSwift'
      target.build_configurations.each do |config|
        config.build_settings['OTHER_LDFLAGS'] = '$(inherited) -weak_framework SwiftUI'
      end
    end

  end
end
```

kubankredit:
Thank you, the code above did the trick!

jongarate:
Initial tests reveal that it's possible to set `SwiftUI` as `-weak_framework` reference in the `OTHER_LDFLAGS` setting of the module that imports the SwiftPM dependency. Now it successfully compiles on both iOS 12 and iOS 14.

jsflax:
This is an interesting one. We only reference SwiftUI if it can be imported using `canImport`. Might be tied to this: https://forums.swift.org/t/weak-linking-of-frameworks-with-greater-deployment-targets/26017/28. It's a bit odd as well that no one seems to have to do this with Combine.

dianaafanador3:
[...]

PR #7268, opened 2021-05-28
Fix - Force SwiftUI weak link on Cocoapods

dianaafanador3:
Fix https://github.com/realm/realm-cocoa/issues/7234
 Fix a runtime crash which happens in some Xcode version (Xcode < 12, reported in Xcode 12.5), where SwiftUI is not weak linked by default. This fix only works for Cocoapods projects.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@ x.y.z Release notes (yyyy-MM-dd)
 * None.
 
 ### Fixed
+* Fix a runtime crash which happens in some Xcode version (Xcode < 12, reported in Xcode 12.5), where SwiftUI is not weak linked by default. This fix only works for Cocoapods projects.
+  ([#7234](https://github.com/realm/realm-cocoa/issues/7234)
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -4,6 +4,8 @@ x.y.z Release notes (yyyy-MM-dd)
 * None.
 
 ### Fixed
+* Fix a runtime crash which happens in some Xcode version (Xcode < 12, reported in Xcode 12.5), where SwiftUI is not weak linked by default. This fix only works for Cocoapods projects.
```
jsflax, line 7: Are they ways to get the crash through methods other than Cocoapods?

dianaafanador3: yep, there is a reported case of someone getting the same error using SPM, in that case the solution is to add "SwiftUI as -weak_framework" to "Other linker flags" OTHER_LDFLAGS to the project importing the module

dianaafanador3: I supposed this is an Xcode issue (that happens sometimes), and not related to which dependency manager the user is using

tgoyne (approved):

PR #7274, opened 2021-06-04
Fix running the tests via spm with Xcode 12.5

tgoyne:
Loading test bundles with a deployment target older than macOS 11 into xctest when using SPM and Xcode 12.5 causes crashes when accessing objective-c metadata, presumably because it's incorrectly expecting the modern format. This appears to be specific to xctest, so we don't want to bump the minimum deployment target of everything, and instead only the tests.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -45,6 +45,22 @@ func combineFlags() -> [SwiftSetting]? {
     return nil
 }
 
+// Xcode 12.5's xctest crashes when reading obj-c metadata if the Swift tests
+// aren't built targeting macOS 11. We still want all of the non-test code to
+// target the normal lower version, though.
+func hostMachineArch() -> String {
+    var systemInfo = utsname()
+    uname(&systemInfo)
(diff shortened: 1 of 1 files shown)


ericjordanmossman:
Interestingly [`Jenkins: swiftpm 12.5`](https://ci.realm.io/job/objc_pr/target=swiftpm,xcode_version=12.5/8127/consoleFull) still failed with code 11.
```
Jun 04 17:13:44 Test Case '-[PrimitiveArrayPropertyTests testFastEnumeration]' started.
Jun 04 17:13:44 Exited with signal code 11
```

tgoyne:
That test appears to be crashing for a different reason, as it crashes (in release builds only) even with the package's deployment target set to v11.

ericjordanmossman:
Okay I think I misunderstood --tThis change addresses `testAllMethodsCheckForInvalidation`, which now passes.
```
Jun 04 16:39:11     ✓ testAllMethodsCheckForInvalidation (0.010 seconds)
```

ericjordanmossman (approved):

tgoyne:
Merging this even though it doesn't fix all of the crashes as it does fix one of them and the other appears to have an unrelated cause.

PR #7279, opened 2021-06-07
Add support for Xcode 13 beta 1

tgoyne:
Mostly just fixing some new warnings. Xcode 13 works without the SPM changes, but we'll need to bump the tools version eventually to take advantage of newer SPM features. The changes to `ObjectiveCSupport.convert()` are because `as T?` to collapse a `T??` now produces warnings so we want to avoid introducing the second Optional, but changing the existing function is a breaking change.


tgoyne:
This is now mostly pre-existing failures. The SwiftUI test failure is a real one: on iOS 15 [the tap on the "New List" button](https://github.com/realm/realm-cocoa/blob/master/Realm/Tests/SwiftUITestHostUITests/SwiftUITestHostUITests.swift#L84) is opening the text field to rename the entry rather than navigating the desired view. I couldn't figure out how to fix this; the interaction which the recorder produces for tapping on the correct part of the row has the exact same result. This is something we'll want to fix but probably doesn't need to block this PR.

jsflax (approved):
Awesome, let's get this out there.

PR #7282, opened 2021-06-09
Fix crashes in -[PrimitiveArrayPropertyTests testFastEnumeration]

tgoyne:
The compiler appears to be miscompiling this function and attempting to unify all of the `values` arrays into a single one, but what it actually does is read from an uninitialized local from later in the function instead of the correct one.

Work around this by sticking each of the subparts of the test in a separate obj-c block to foil the optimization.

### DIFF
--- a/Realm/Tests/PrimitiveArrayPropertyTests.m
+++ b/Realm/Tests/PrimitiveArrayPropertyTests.m
@@ -58,6 +58,34 @@ static double average(NSArray *values) {
     return sum / c;
 }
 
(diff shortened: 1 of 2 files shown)


ericjordanmossman:
How did you figure this out?

tgoyne:
After trying a bunch of things I eventually discovered that deleting the body of the test after the section which was crashing made it work. From there I stepped through the disassembly in the debugger and compared the stack locations being initialized with the ones being read and found that it was reading from a stack location which was not initialized until later in the function.

jsflax (approved):
Great catch.

Issue #5662, opened 2018-03-09
Can't make Realm class fileprivate


mrh-is:
## Goals
Make a Realm class `fileprivate` (or `private`).

## Expected Results
The app doesn't crash. 😱

## Actual Results
App crashes when Realm tries to set up schemas.

Crash:

```
*** Terminating app due to uncaught exception 'RLMException', reason: 'RLMObject subclasses cannot be nested within other declarations. Please move _TtC7KidFundP33_98F658B71D628B2D9C0CBA606381CB1825Event to global scope.'
```

<details>
<summary>Stacktrace</summary>

```
0   CoreFoundation     0x000000010b7391cb __exceptionPreprocess + 171
1   libobjc.A.dylib    0x0000000111a64f41 objc_exception_throw + 48
2   Realm              0x000000010e039dde _ZL26RLMRegisterClassLocalNamesPP10objc_classm + 414
3   Realm              0x000000010e03afdf +[RLMSchema sharedSchema] + 847
4   Realm              0x000000010e01bda4 +[RLMRealm realmWithConfiguration:error:] + 5620
5   RealmSwift         0x000000010e9374b2 _T0So8RLMRealmCABSo0A13ConfigurationC13configuration_tKcfCTO + 162
6   RealmSwift         0x000000010e937c46 _T010RealmSwift0A0CA2C13ConfigurationV13configuration_tKcfc + 918
7   RealmSwift         0x000000010e93781f _T010RealmSwift0A0CA2C13ConfigurationV13configuration_tKcfC + 687
...
```

</details>

## Steps to Reproduce
Make a Realm class `fileprivate` (but still at the top level).

## Code Sample
Class looks like this:

```swift
fileprivate class Event: Object {
    ... // Normal class stuff
}
```

I can confirm that the same class works if I remove `fileprivate`.

## Version of Realm and Tooling
```
ProductName:	Mac OS X
ProductVersion:	10.13.3
BuildVersion:	17D102

/Applications/Xcode.app/Contents/Developer
Xcode 9.2
Build version 9C40b

/usr/local/bin/pod
1.3.1
Realm (2.10.0)
RealmSwift (2.10.0)
RealmSwift (~> 2.0)

/bin/bash
GNU bash, version 3.2.57(1)-release (x86_64-apple-darwin17)


(not in use here)

/usr/local/bin/git
git version 2.4.3
```

lslv1243:
any update?

PR #7291, opened 2021-06-16
Allow nested and private object subclasses in Swift

tgoyne:
This is something that's actually worked fine ever since Swift 3 added the ability to override the obj-c name for Swift types and we just needed an error message explaining what to do.

Fixes https://github.com/realm/realm-cocoa/issues/5662.

### DIFF
--- a/Realm/Tests/Swift/SwiftSchemaTests.swift
+++ b/Realm/Tests/Swift/SwiftSchemaTests.swift
@@ -132,6 +132,35 @@ class NoProps: FakeObject {
     // no @objc properties
 }
 
+@MainActor
+class RequiresObjcName: RLMObject {
+    static var enable = false
+    @MainActor
+    override class func _realmIgnoreClass() -> Bool {
+        return !enable
+    }
+}
+
+class ClassWrappingObjectSubclass {
+    class Inner: RequiresObjcName {
+        @objc dynamic var value = 0
+    }
+}
+struct StructWrappingObjectSubclass {
+    class Inner: RequiresObjcName {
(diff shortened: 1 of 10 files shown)


RealmSwift/Tests/SwiftTestObjects.swift
```diff
@@ -840,3 +840,18 @@ class EmbeddedTreeObject3: EmbeddedObject, EmbeddedTreeObject {
...
+    @objc dynamic var inner: NestedInnerClass?
+
+    @objc(NestedInnerClass)
+    class NestedInnerClass: EmbeddedObject {
+        @objc dynamic var value = 0
+    }
+}
+
+@objc(PrivateObjectSubclass)
+private class PrivateObjectSubclass: Object {
```
leemaguire, line 855: these two classes don't seem to be used anywhere?

Realm/Tests/Swift/SwiftSchemaTests.swift
```diff
@@ -246,6 +273,23 @@ class SwiftRLMSchemaTests: RLMMultiProcessTestCase {
...
     }
+
+    func testInvalidNestedClass() {
+        if isParent {
+            XCTAssertEqual(0, runChildAndWait(), "Tests in child process failed")
+            return
+        }
+
+        RequiresObjcName.enable = true
+        assertThrowsWithReasonMatching(RLMSchema.sharedSchema(for: ClassWrappingObjectSubclass.Inner.self),
```
leemaguire, line 284: Should we also test that adding `@objc(className)` produces the correct schema?

PR #7293, opened 2021-06-16
Don't create unmanaged accessor classes for types which don't need it

tgoyne:
We only need unmanaged accessor subclasses for objective-c classes which use RLMArray, as the only thing they override is the getters and setters for collection properties. Creating accessors subclasses for all types isn't hugely wasteful but it does use a bit of memory.

### DIFF
--- a/Realm/RLMObjectBase.mm
+++ b/Realm/RLMObjectBase.mm
@@ -48,15 +48,15 @@ static bool isManagedAccessorClass(Class cls) {
     return strncmp(className, accessorClassPrefix, sizeof(accessorClassPrefix) - 1) == 0;
 }
 
-static bool maybeInitObjectSchemaForUnmanaged(RLMObjectBase *obj) {
+static void maybeInitObjectSchemaForUnmanaged(RLMObjectBase *obj) {
     Class cls = obj.class;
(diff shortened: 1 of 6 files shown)


Realm/RLMObjectBase.mm
```diff
@@ -294,7 +292,7 @@ - (RLMObjectSchema *)objectSchema {
 
 - (BOOL)isInvalidated {
     // if not unmanaged and our accessor has been detached, we have been deleted
-    return self.class == _objectSchema.accessorClass && !_row.is_valid();
+    return _info && !_row.is_valid();
```
tgoyne, line 295: The previous version of this check worked, but didn't really do what it was intending to do. `_objectSchema.accessorClass` is only actually set to the managed accessor subclass on the object schemas used for managed objects, and is the user-defined class on unmanaged object schemas. This happened to work when unmanaged objects were always the unmanaged subclass, but broke with this change. `_info` was added after this code was written and now is an easier way to check if the object is managed, as it's non-nil only for managed objects (the more obvious check of `_realm` doesn't work because `_realm` is set to `nil` when an object is deleted directly).

leemaguire (approved):
nice

PR #7296, opened 2021-06-17
Fix the sync compactOnLaunch test

tgoyne:
Even with RLMSyncStopPolicyImmediately, closing a sync session is an async operation since it has to happen on the sync worker thread. This can result in the test failing due to sync not closing the Realm file fast enough. Fix this by waiting until we can acquire an exclusive lock on the Realm before we try to reopen it.


leemaguire (approved):
nice 👍🏼

Issue #7298, opened 2021-06-19
List<T> and Results<T> should support access by IndexSet


bdkjones:
## Goals
Swift's `Array` type offers an API to return all indexes in a given `IndexSet`: 

    let objects: [Any] = someArray.objects(at: someIndexSet)

Currently, Realm's collection types do not support this API. This is inconvenient because `NSTableView` extensively uses `IndexSet`. If I want to translate the selected rows in the table to an Array of objects on which I can act, I can't just pass the `IndexSet` returned by `selectedRowIndexes` to the Realm collection. Instead, I must manually loop through the indexes.

Additionally, because `IndexSet` is heavily optimized for its task, I assume the `objects(at:)` API is more performant than the naive for-loop approach.

## Expected Results

Realm collection types should support accessing objects by `IndexSet`.

## Actual Results

Realm collection types do not support `IndexSet`.

## Steps for others to Reproduce

N/A

## Code Sample

N/A

## Version of Realm and Tooling

Realm framework version: 10.7.2

Xcode version: 13.0 Beta 1

iOS/OSX version: macOS 11.4

Dependency manager + version: SPM

tgoyne:
There isn't really much room for us to do anything significantly faster; skipping some redundant checks might make it 5-10% faster but there isn't a lot of unnecessary overhead to begin with. A convenience helper would be easy to add, though.

bdkjones:
The convenience method would make interfacing with NSTableView much nicer, for sure.

PR #7307, opened 2021-06-23
Add objects(at indexes:) for collections

leemaguire:
Adds ability to select elements from a collection with a given `IndexSet`. This adds functionality to our collection API's that is already present in Foundation.

Addresses https://github.com/realm/realm-cocoa/issues/7298


Realm/RLMArray.mm
```diff
@@ -509,10 +509,14 @@ - (NSUInteger)indexOfObjectWithPredicate:(NSPredicate *)predicate {
...
 
 - (NSArray *)objectsAtIndexes:(NSIndexSet *)indexes {
-    if (!_backingCollection) {
-        _backingCollection = [NSMutableArray new];
+    try {
+        if (!_backingCollection) {
+            _backingCollection = [NSMutableArray new];
+        }
+        return [_backingCollection objectsAtIndexes:indexes];
+    } catch(...) {
```
tgoyne, line 517: Why is this swallowing errors? If I remove this and make the managed version throw on out-of-bounds indexing the only tests which fail are things which should be failing.

leemaguire: The managed type causes problems with KVO if it throws so I didn’t see a point in letting it throw anywhere at all

tgoyne: It appears to work fine to me. The only KVO test which fails is showing that it fixes a place where RLMManagedArray incorrectly differed from RLMArray and NSArray.

tgoyne: Oh, it might actually be making a currently weird thing broken in a different way rather than fixing something. Either way, the fix is probably just to validate argument bounds earlier in the process?

leemaguire: Changed to not swallow the exception

tgoyne (approved):

CHANGELOG.md
```diff
@@ -1,6 +1,9 @@
 x.y.z Release notes (yyyy-MM-dd)
 =============================================================
 ### Enhancements
+Add `List.objects(at indexes:)` in Swift and `[RLMCollection objectsAtIndexes:]` in Objective-C. This allows you to select elements in a collection with a given IndexSet ([#7298](https://github.com/realm/realm-cocoa/issues/7298).
```
jsflax, line 4: Missing an asterisk.

jsflax (approved):
LGTM minus one nit.

PR #7325, opened 2021-07-06
Add ability to get string from PartialKeyPath

leemaguire:
This PR adds the ability to retrieve the string representation for a given `PartialKeyPath<>` on a Realm `ObjectBase` type.

The API usage is as follows:
```
ObjectBase._name(for: \ModernAllTypesObject.boolCol) // "boolCol"
ObjectBase._name(for: \ModernKeyPathObject.list // "list"
ObjectBase._name(for: \ModernKeyPathObject.list[0].objectCol // "list.objectCol"
// More complex
ObjectBase._name(for: \ModernKeyPathObject.map[""]??.objectCol?.linkingObjects[0].boolCol) // "map.objectCol.linkingObjects.boolCol"
```

**Note:** The API is not intended for public use, but it exists in the public domain.

**Limitations:**
Objects that declare properties using the old `@objc dynamic` syntax will have limitations with this API. The limitations are as follows:
- Nested key path components will not produce a string if they do not terminate on a `@objc dynamic` property. We make use of `KeyPath._kvcKeyPathString` for key paths that terminate with `@objc dynamic`.
- Indexing on collection types to perform nesting i.e ` \OldObject.list[0].objectCol` is not supported even if they terminate on a `@objc dynamic` property.
- Mixing the old `@objc dynamic` syntax and `@Persisted` syntax in nested key paths is not supported.

Given the above limitations, it is recommended to only use this API with properties declared with `@Persisted`.


RealmSwift/Impl/ComplexTypes.swift
```diff
@@ -301,8 +310,8 @@ extension RawRepresentable where Self: _OptionalPersistable, RawValue: _Optional
     }
 }
 
-extension RawRepresentable where Self: CaseIterable {
-    public static func _rlmDefaultValue() -> Self {
+extension PersistableEnum {
```
leemaguire, line 313: I didn't see a reason why `extension RawRepresentable where Self: PersistableEnum` would be better than the non specialised equivalent.

Realm/RLMObjectBase_Private.h
```diff
@@ -26,6 +26,9 @@ NS_ASSUME_NONNULL_BEGIN
 @interface RLMObjectBase ()
 + (void)initializeLinkedObjectSchemas;
 + (bool)isEmbedded;
+// Key Path Strings
+- (void)prepareForRecording;
+@property (nonatomic, nullable) NSMutableArray* lastAccessedNames;
```
tgoyne, line 31: Asterisks in obj-c code are right-aligned (there's a bunch of these and I won't note them all).

Realm/RLMObjectBase.mm
```diff
@@ -410,6 +412,21 @@ - (id)objectiveCMetadata {
     return nil;
 }
 
+#pragma mark - Key Path Strings
+
+- (void)prepareForRecording {
+    for (RLMProperty *property in self.objectSchema.properties) {
+        if (property.swiftAccessor) {
+            [property.swiftAccessor observe:property on: self];
```
tgoyne, line 420: No space after colon in obj-c message sends.

RealmSwift/Impl/PropertyAccessors.swift
```diff
@@ -281,6 +286,9 @@ internal class PersistedLinkingObjectsAccessor<Element: ObjectBase>: RLMManagedP
     }
     @objc override class func observe(_ property: RLMProperty, on parent: RLMObjectBase) {
         // N/A: unmanaged LinkingObjects are always empty
+        if parent.lastAccessedNames != nil {
```
tgoyne, line 289: This needs to remove the comment above as it's no longer true.

RealmSwift/RealmCollection.swift
```diff
@@ -197,6 +197,10 @@ public protocol RealmCollectionValue: Hashable, _RealmSchemaDiscoverable {
     // Iterating over collections requires mapping NSNull to nil, but we can't
     // just do `nil as T` because of non-nullable collections
     static func _nilValue() -> Self
+    /// :nodoc:
+    // Get the zero/empty/nil value for this type. Used to supply a default
+    // when the user does not declare one in their model.
+    static func _rlmDefaultValue(_ doNotReturnNilValue: Bool) -> Self
```
tgoyne, line 203: This name is really bad and needs to be something better.

RealmSwift/MutableSet.swift
```diff
@@ -471,6 +474,14 @@ extension MutableSet: RealmCollection {
                 convenience should not be relied on.
      */
     public subscript(position: Int) -> Element {
+        if lastAccessedNames != nil {
+            let value = Element._rlmDefaultValue(true)
+            if let value = value as? ObjectBase {
+                value.lastAccessedNames = lastAccessedNames
+                value.prepareForRecording()
+            }
```
tgoyne, line 482: This logic is repeated in a whole bunch of places and should be extracted to one place. Probably just `Element._rlmKeypathRecorder()` or something?

Realm/RLMObjectBase.mm
```diff
@@ -410,6 +412,21 @@ - (id)objectiveCMetadata {
     return nil;
 }
 
+#pragma mark - Key Path Strings
+
+- (void)prepareForRecording {
```
tgoyne, line 417: This function probably could be written in Swift instead and if so should be as it's swift-specific.

RealmSwift/Impl/ComplexTypes.swift
```diff
@@ -209,8 +209,12 @@ extension RealmOptional: SchemaDiscoverable, _RealmSchemaDiscoverable where Valu
 }
 
 extension LinkingObjects: _Persistable where Element: _Persistable {
-    public static func _rlmDefaultValue() -> Self {
-        fatalError("LinkingObjects properties must set the origin property name")
+    public static func _rlmDefaultValue(_ doNotReturnNilValue: Bool) -> Self {
+        if doNotReturnNilValue {
+            return Self.init(propertyName: "", handle: nil)
```
tgoyne, line 214: .init is redundant.

RealmSwift/Object.swift
```diff
@@ -540,3 +540,45 @@ extension Object: AssistedObjectiveCBridgeable {
...
+     */
+    public static func _name<T: ObjectBase>(for keyPath: PartialKeyPath<T>) -> String {
+        if let name = keyPath._kvcKeyPathString {
+            return name
+        }
+        let traceObject = T()
+        traceObject.lastAccessedNames = NSMutableArray()
+        traceObject.prepareForRecording()
+        let obj = traceObject[keyPath: keyPath]
+        let schema = RLMObjectSchema.init(forObjectClass: T.self)
```
tgoyne, line 570: Why is this creating a new RLMObjectSchema instead of just getting it from the class?

RealmSwift/Object.swift
```diff
@@ -540,3 +540,45 @@ extension Object: AssistedObjectiveCBridgeable {
...
+        // if both getter and setter are nil then this class uses @Persisted
```
tgoyne, line 571: This isn't true; they'll all be `nil` for a legacy class where none of the properties are @objc.

RealmSwift/PersistedProperty.swift
```diff
@@ -159,10 +159,31 @@ public struct Persisted<Value: _Persistable> {
...
             return value
-        case let .unmanagedObserved(value, _):
+        case let .unmanagedObserved(value, key):
+            var name: String = ""
+            if Value._rlmType == .linkingObjects {
+                name = RLMObjectBaseObjectSchema(object)!.computedProperties[Int(key)].name
+            } else {
+                name = RLMObjectBaseObjectSchema(object)!.properties[Int(key)].name
+            }
+            if object.lastAccessedNames != nil {
```
tgoyne, line 172: Looking up the name should be inside this block rather than doing it even if we're not recording the keypath.

RealmSwift/Impl/ComplexTypes.swift
leemaguire: This is required when `Persisted.get` is called.

Realm/RLMObjectBase.mm
```diff
@@ -17,6 +17,7 @@
 ////////////////////////////////////////////////////////////////////////////
 
 #import "RLMObject_Private.hpp"
+#import "RLMObjectBase_Private.h"
```
leemaguire, line 20: this is required for `setLastAccessedNames`

RealmSwift/EmbeddedObject.swift
```diff
@@ -95,11 +95,6 @@ extension EmbeddedObject: RealmCollectionValue {
         return nil
     }
 
-    /// The object schema which lists the managed properties for the object.
-    public var objectSchema: ObjectSchema {
-        return ObjectSchema(RLMObjectBaseObjectSchema(self)!)
-    }
```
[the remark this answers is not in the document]
leemaguire: It now lives in the extension for `ObjectBase`, both `Object` and `EmbeddedObject` reference it from there now

RealmSwift/Object.swift
```diff
@@ -540,3 +535,55 @@ extension Object: AssistedObjectiveCBridgeable {
...
+     */
+    public static func _name<T: ObjectBase>(for keyPath: PartialKeyPath<T>) -> String {
+        if let name = keyPath._kvcKeyPathString {
+            return name
+        }
+        let traceObject = T()
+        traceObject.lastAccessedNames = NSMutableArray()
+        traceObject.prepareForRecording()
+        let obj = traceObject[keyPath: keyPath]
+        let isLegacy = traceObject.objectSchema.rlmObjectSchema.properties.allSatisfy { $0.isLegacy }
```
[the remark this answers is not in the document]
leemaguire: I've moved things around and added `_isLegacyProperty` to the collections

Realm/RLMSwiftCollectionBase.mm
```diff
@@ -99,6 +99,14 @@ - (BOOL)conformsToProtocol:(Protocol *)aProtocol {
     return aProtocol == @protocol(NSFastEnumeration) || [self._rlmCollection conformsToProtocol:aProtocol];
 }
 
+- (NSString *)_propertyKey {
+    return [(id)self._rlmCollection propertyName];
+}
+
+- (BOOL)_isLegacyProperty {
+    return (BOOL)(size_t)[(id)self._rlmCollection performSelector:@selector(isLegacyProperty)];
```
tgoyne, line 107: `isLegacyProperty` should just be a property declared in the private extension for each of the collections. It'd be less code than declaring separate ivars and getters and eliminates the need for performSelector.

RealmSwift/Impl/Persistable.swift
```diff
@@ -36,25 +36,47 @@ public protocol _Persistable: _RealmSchemaDiscoverable {
...
 
-// A tag protocol for persistable types which can appear inside Optional
-public protocol _OptionalPersistable: _Persistable {
+extension _RealmSchemaDiscoverable where Self: _Persistable {
+    public static func _rlmKeyPathRecorder(with lastAccessedNames: NSMutableArray) -> Self {
+        let value = Self._rlmDefaultValue(true)
+
+        if let value = value as? ObjectBase {
+            value.lastAccessedNames = lastAccessedNames
+            value.prepareForRecording()
```
tgoyne, line 66: Should return `value` here to skip trying to dynamic cast it to KeyPathStringCollection below.

RealmSwift/EmbeddedObject.swift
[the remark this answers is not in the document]
tgoyne: I think that may cause issues for the generated docs. There's a bunch of stuff redeclared in Object which don't strictly need to be because it's the only way to make jazzy put the docs on the right page.

Realm/RLMSwiftCollectionBase.mm
leemaguire: done, removed a bunch of things and moved isLegacyCollection to a property

RealmSwift/EmbeddedObject.swift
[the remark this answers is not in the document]
leemaguire: reverted the change

tgoyne (approved):

RealmSwift/Object.swift
```diff
@@ -540,3 +540,50 @@ extension Object: AssistedObjectiveCBridgeable {
...
+
+     Example:
+     ```
+     let name = ObjectBase._name(for: \Person.dogs[0].name) // "dogs.name"
+     // Note that the above KeyPath expression is only supported with properties declared
+     // with `@Persisted`.
+     let nested = ObjectBase._name(for: \Person.address.city.zip) // "address.city.zip"
+     ```
+     */
+    public static func _name<T: ObjectBase>(for keyPath: PartialKeyPath<T>) -> String {
```
jsflax, line 563: Any particular reason this is public? It's not part of any protocol and if it is just for testing, that is not a good reason.

RealmSwift/RealmCollection.swift
```diff
@@ -223,7 +239,16 @@ extension AnyRealmValue: RealmCollectionValue {
     }
 }
 
-extension Optional: RealmCollectionValue where Wrapped: RealmCollectionValue {
+extension Optional: RealmCollectionValue where Wrapped: RealmCollectionValue,
+                                               Wrapped: _DefaultConstructible {
+    /// :nodoc:
+    public static func _rlmDefaultValue(_ forceDefaultInstantiation: Bool) -> Optional<Wrapped> {
```
jsflax, line 245: For `_rlmDefaultValue` in general I would consider changing the parameter name from `forceDefaultInstantiation` to `forceDefaultConstruction` or `forceDefaultInitialization` so that we have _some_ consistency.

RealmSwift/RealmCollection.swift
```diff
@@ -1190,3 +1215,14 @@ extension LinkingObjects: ObservableCollection {
         return objc === rlmResults
     }
 }
+
+// MARK: Key Path Strings
+
+/// Tag protocol which allows a collection to produce its property name
+internal protocol KeyPathStringCollection {
```
jsflax, line 1222: This protocol needs an adjective as it makes little sense otherwise. I would call it `PropertyNameDiscoverable` or `PropertyNameConvertible`.

RealmSwift/Object.swift
tgoyne: Making it internal and testing it via @testable import means that we can't test it in release builds, and I could imagine this having ARC lifetime issues with optimizations. I guess it doesn't actually need to be on ObjectBase, though, so maybe it should be a free function?

jsflax: I acquiesce. As discussed offline, @leemaguire will make it a free function and we will keep it public.

RealmSwift/RealmCollection.swift
leemaguire: changed to PropertyNameConvertible

RealmSwift/RealmCollection.swift
leemaguire: changed to forceDefaultInitialization

jsflax (approved):
🎉

Issue #7323, opened 2021-07-05
RealmProperty cannot decode null values


miklselsoe:
After RealmOptional was deprecated, I changed all my `RealmOptional<Type>` properties to `RealmProperty<Type?>`

I have a few places where I relied on `RealmOptional` being `Codable`. With RealmProperty also being Codable, I thought it was safe to change the property definitions. Unfortunately, while `RealmOptional` was able to decode `null` values, `RealmProperty` cannot:

```
Swift.DecodingError.valueNotFound(Swift.Optional<Swift.Double>, Swift.DecodingError.Context(codingPath: [CodingKeys(stringValue: "fromKm", intValue: nil)], debugDescription: "Expected Optional<Double> but found null value instead.", underlyingError: nil))
```

Looking at the implementation for the two types, it is evident that `RealmProperty` ignores null values, while `RealmOptional` does not:

```
extension RealmProperty: Codable where Value: Codable {
    public convenience init(from decoder: Decoder) throws {
        self.init()
        // `try decoder.singleValueContainer().decode(Value?.self)` incorrectly
        // rejects null values: https://bugs.swift.org/browse/SR-7404
        let container = try decoder.singleValueContainer()
        self.value = try container.decode(Value.self)
    }

    public func encode(to encoder: Encoder) throws {
        try self.value.encode(to: encoder)
    }
}
```
https://github.com/realm/realm-cocoa/blob/v10.9.0/RealmSwift/RealmProperty.swift#L70-L76

```
extension RealmOptional: Codable where Value: Codable {
    public convenience init(from decoder: Decoder) throws {
        self.init()
        // `try decoder.singleValueContainer().decode(Value?.self)` incorrectly
        // rejects null values: https://bugs.swift.org/browse/SR-7404
        let container = try decoder.singleValueContainer()
        self.value = container.decodeNil() ? nil : try container.decode(Value.self)
    }

    public func encode(to encoder: Encoder) throws {
        try self.value.encode(to: encoder)
    }
}
```
https://github.com/realm/realm-cocoa/blob/v10.9.0/RealmSwift/Optional.swift#L78-L84

I think this is a bug. I wonder if there is a better workaround than reverting to `RealmOptional`?

## Version of Realm and Tooling
Realm framework version: 10.8.0

Xcode version: 12.5.1

iOS/OSX version: 14.6

Dependency manager + version: Carthage 0.38.0

leemaguire:
Hi @miklselsoe 

[...]

PR #7328, opened 2021-07-07
Fix for RealmProperty<T?> decoding null json values

leemaguire:
RealmProperty could not properly decode `null` json values as type system would expect the `nil` representation of the underlying `Value` type

Fixes https://github.com/realm/realm-cocoa/issues/7323


RealmSwift/RealmProperty.swift
```diff
@@ -81,7 +79,17 @@ extension RealmProperty: Codable where Value: Codable {
 }
 
 /// A protocol describing types that can parameterize a `RealmPropertyType`.
-public protocol RealmPropertyType {}
+public protocol RealmPropertyType {
+    static func nilValue() -> Self
```
tgoyne, line 83: Can this be combined with the existing `_nilValue()`?

leemaguire: I have moved `_nilValue` to `_RealmSchemaDiscoverable`

tgoyne (approved):

PR #7329, opened 2021-07-07
AsyncOpen & AutoOpen property Wrappers

dianaafanador3:
This PR adds AsyncOpen & AutoOpen Property wrappers

API declaration and use for both will be as follow

**@AsyncOpen**

`@AsyncOpen(appId: "appId", partitionValue: <partition>, timeout: 4000) var asyncOpen`

```
struct ContentView: View {
         @AsyncOpen(appId: "app_id", partitionValue: <partition_value>, timeout: 4000) var asyncOpen

        var body: some View {
           switch asyncOpen {
            case .notOpen:
               ProgressView()
            case .open(let realm):
                ListView()
                   .environment(\.realm, realm)
           case .error(_):
                ErrorView()
           case .progress(let progress):
               ProgressView(progress)
            }
         }
     }
```

**@AutoOpen**
`@AutoOpen(appId: "appId", partitionValue: <partition>, timeout: 4000) var autoOpen`

```
struct ContentView: View {
         @AutoOpen(appId: "app_id", partitionValue: <partition_value>, timeout: 4000) var autoOpen

        var body: some View {
           switch autoOpen {
            case .notOpen:
               ProgressView()
            case .open(let realm):
                ListView()
                   .environment(\.realm, realm)
           case .error(_):
                ErrorView()
           case .progress(let progress):
               ProgressView(progress)
            }
         }
     }
```

This PR includes 
- [x]  Implementation for AsyncOpen & AutoOpen/documentation
- [x]  Server tests for both property wrappers
- [x]  UI Tests for both property wrappers
- [x]  Example for realm.asyncOpen() and AsyncOpen/AutoOpen 
- [x] Solve CI Issues


RealmSwift/Optional.swift
```diff
@@ -74,13 +74,12 @@ extension RealmOptional: Equatable where Value: Equatable {
 }
 
 @available(*, deprecated, message: "RealmOptional has been deprecated, use RealmProperty<T?> instead.")
-extension RealmOptional: Codable where Value: Codable {
+extension RealmOptional: Codable where Value: Codable, Value: _RealmSchemaDiscoverable {
```
jsflax, line 77: Why has this sneaked in to the PR?

RealmSwift/SwiftUI.swift
```diff
@@ -645,6 +645,285 @@ extension EnvironmentValues {
...
+
+    /// :nodoc:
+    public var wrappedValue: AsyncOpenState {
+        storage.asyncOpenState
+    }
+
+    /**
+     This will cancel any notification from the property wrapper states
+     */
+    public func cancel() {
```
jsflax, line 763: Without exposing the propertyWrapper itself in some way, this method is not available to the user.

RealmSwift/SwiftUI.swift
```diff
@@ -645,6 +645,285 @@ extension EnvironmentValues {
...
+    /**
+     Initialise the property wrapper
+     - parameter appId: The unique identifier of your Realm app.
+     - parameter partitionValue: The `BSON` value the Realm is partitioned on.
+     - parameter baseConfiguration: A Base Configuration use when creating the Realm,
+       if empty the configuraion is setted to the `defaultConfiguration`
+     - parameter timeout: The maximum number of milliseconds to allow for a connection to
+       become fully established., if empty or `nil` no connection timeout is setted.
+     */
+    public init<T: BSON>(appId: String, partitionValue: T, baseConfiguration: Realm.Configuration = Realm.Configuration.defaultConfiguration, timeout: UInt? = nil) {
```
jsflax, line 777: We can probably derive AppId from the last used AppId as most applications will be singular. This would allow the property to potentially have a default value, though this may be a chunk more of work.

I would also change the name of baseConfiguration to configuration.

RealmSwift/SwiftUI.swift
```diff
@@ -645,6 +645,285 @@ extension EnvironmentValues {
...
+            app.objectWillChange.sink { [self] app in
+                if let currentUser = app.currentUser {
+                    asyncOpenForUser(currentUser, partitionValue: partitionValue, configuration: baseConfiguration)
+                }
+            }.store(in: &storage.cancellables)
+        }
+    }
+
+    public mutating func update() {
+        if storage.realm == nil || storage.realm != realm {
```
jsflax, line 797: I believe we should be checking for configuration here, and also adding a Realm EnvironmentValue for partitionValue. There are use cases where partitionValues are derived and not static. This also should be tested.

RealmSwift/SwiftUI.swift
```diff
@@ -645,6 +645,285 @@ extension EnvironmentValues {
...
+    /**
+     Initialise the property wrapper
+     - parameter appId: The unique identifier of your Realm app.
+     - parameter partitionValue: The `BSON` value the Realm is partitioned on.
+     - parameter baseConfiguration: A Base Configuration use when creating the Realm,
+       if empty the configuraion is setted to the `defaultConfiguration`
+     - parameter timeout: The maximum number of milliseconds to allow for a connection to
+       become fully established, if empty or `nil` no connection timeout is setted.
+     */
+    public init<T: BSON>(appId: String, partitionValue: T, baseConfiguration: Realm.Configuration = Realm.Configuration.defaultConfiguration, timeout: UInt? = nil) {
```
jsflax, line 901: Same comments as above.

RealmSwift/SwiftUI.swift
```diff
@@ -645,6 +645,285 @@ extension EnvironmentValues {
...
+        let app = App(id: appId)
+        if let timeout = timeout {
+            let syncTimeoutOptions = SyncTimeoutOptions()
+            syncTimeoutOptions.connectTimeout = timeout
+            app.syncManager.timeoutOptions = syncTimeoutOptions
+        }
+        if app.currentUser?.isLoggedIn ?? false,
+           let currentUser = app.currentUser {
```
jsflax, line 909: Same comments as above.

RealmSwift/SwiftUI.swift
```diff
@@ -645,6 +645,285 @@ extension EnvironmentValues {
...
+            app.objectWillChange.sink { [self] app in
+                if let currentUser = app.currentUser {
+                    asyncOpenForUser(currentUser, partitionValue: partitionValue, configuration: baseConfiguration)
+                }
+            }.store(in: &storage.cancellables)
+        }
+    }
+
+    public mutating func update() {
+        if storage.realm == nil || storage.realm != realm {
```
jsflax, line 921: Same comments as above.

jsflax (commented):
Great job so far, few remarks here and there. It also appears that a recent commit to master may have snuck in.

RealmSwift/Optional.swift
dianaafanador3: This changes are already in master, I don't know why they are showing on this pull request; I rebase my barch with a master a week ago, any clues why @jsflax ?

Realm/ObjectServerTests/SwiftSyncTestCase.swift
```diff
@@ -324,6 +164,24 @@ open class SwiftSyncTestCase: RLMSyncTestCase {
                 file: (fileName), line: lineNumber)
         }
     }
+
+    public let bigObjectCount = 2
```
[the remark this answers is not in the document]
dianaafanador3: Solved on commit Refactor https://github.com/realm/realm-cocoa/commit/c5897e7788eb0ef78bf17f61209ff5ab0d032504

Realm/ObjectServerTests/SwiftUIServerTests.swift
```diff
@@ -0,0 +1,258 @@
...
+import SwiftUI
+
+#if canImport(RealmTestSupport)
+import RealmSwiftSyncTestSupport
+import RealmSyncTestSupport
+#endif
+
+@available(OSX 11, *)
+@objc(SwiftUIServerTests)
+class SwiftUIServerTests: SwiftSyncTestCase {
```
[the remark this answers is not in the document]
dianaafanador3: @objc is not needed, solved on https://github.com/realm/realm-cocoa/commit/48e7222f26193634154917b615e1106027757f8a

RealmSwift/SwiftUI.swift
dianaafanador3: So I'm doing at some test for the `EnvironmentValue` for a partitionValue, the only issues with this is will be that we cannot have a generic environment value, but we can use AnyBSON, so the user would have to do something like this `.enviroment(.\partitionValue, .string("partitionValue"))` instead of been able to do this `enviroment(.\partitionValue, "partitionValue")` to inject it into the property wrapper

RealmSwift/SwiftUI.swift
dianaafanador3: mmmm maybe i'm not understanding this correctly but is there a way to retrieve the appId from the latest used one?

RealmSwift/SwiftUI.swift
```diff
@@ -645,6 +645,285 @@ extension EnvironmentValues {
...
+                }
+            }
+    }
+}
+
+/**
+`AsyncOpenState`is an enum representing different states from `AsyncOpen` and `AutoOpen` process
+*/
+public enum AsyncOpenState {
+    /// Initial state
```
tgoyne, line 678: This needs more elaboration. I assume it's the state we're in while we're waiting for a user to be logged in or while attempting to establish the connection? It might also be worth splitting those into `.waitingForUser` and `.connecting` if that's easy to do.

RealmSwift/SwiftUI.swift
```diff
@@ -645,6 +645,285 @@ extension EnvironmentValues {
...
+    case notOpen
+    /// Returns a realm when the async open process is done or in case there is no internet connection in AutoOpen
```
tgoyne, line 680: An enum case doesn't do things or return values, and the docs should explain what it means to be in that state. Some rough attempts at that:

open: "The Realm has been opened and is ready for use. For AsyncOpen this means that the Realm has been fully downloaded, but for AutoOpen the existing local file may have been used if the device is offline."
progress: "The Realm is currently being downloaded from the server."
error: "Opening the Realm failed."

RealmSwift/SwiftUI.swift
```diff
@@ -645,6 +645,285 @@ extension EnvironmentValues {
...
+    case error(Error)
+}
+
+// MARK: - AsyncOpen
+
+/// A property wrapper type that initiates a `Realm.asyncOpen()` for the current user which asynchronously open a Realm,
+/// and notifies states for the given process
+///
+/// Add AsyncOpen to your ``SwiftUI/View`` or ``SwiftUI/App``,  after a user is already logged,
+/// or if a user is going to be logged
```
tgoyne, line 694: Missing "in" after both "logged" and a period.

RealmSwift/SwiftUI.swift
```diff
@@ -645,6 +645,285 @@ extension EnvironmentValues {
...
+            } receiveValue: { realm in
+                self.storage.realm = realm
+                self.storage.asyncOpenState = .open(realm)
+            }.store(in: &storage.cancellables)
+    }
+
+    /// :nodoc:
+    public var projectedValue: Published<AsyncOpenState>.Publisher {
+        return storage.$asyncOpenState
+    }
```
tgoyne, line 753: This probably does need to be documented.

RealmSwift/SwiftUI.swift
tgoyne: It's currently not exposed in any way, but we internally have a cache of all Apps which have been initialized in the current run of the application. We could make the appId optional, and then if it's `nil` check if there's exactly one RLMApp in the cache, and if so use that (and otherwise throw an exception).

RealmSwift/SwiftUI.swift
```diff
@@ -645,6 +645,285 @@ extension EnvironmentValues {
...
+        storage.cancellables.forEach { $0.cancel() }
+        storage.cancellables = []
+    }
+
+    /**
+     Initialise the property wrapper
+     - parameter appId: The unique identifier of your Realm app.
+     - parameter partitionValue: The `BSON` value the Realm is partitioned on.
+     - parameter baseConfiguration: A Base Configuration use when creating the Realm,
+     if empty the configuraion is setted to the `defaultConfiguration`
```
tgoyne, line 773: This needs a bit more explanation because we don't just use the supplied configuration directly. From reading this I would assume that I have to set the syncConfiguration on the config that I pass in.

RealmSwift/SwiftUI.swift
```diff
@@ -645,6 +645,285 @@ extension EnvironmentValues {
...
+    /**
+     This will cancel any notification from the property wrapper states
+     */
+    public func cancel() {
+        storage.cancellables.forEach { $0.cancel() }
+        storage.cancellables = []
+    }
+
+    /**
+     Initialise the property wrapper
```
tgoyne, line 769: We use US spelling in the docs, so "Initialize".

RealmSwift/SwiftUI.swift
```diff
@@ -645,6 +645,285 @@ extension EnvironmentValues {
...
+/// Add AutoOpen to your ``SwiftUI/View`` or ``SwiftUI/App``,  after a user is already logged,
+/// or if a user is going to be logged
+///
+///     @AutoOpen(appId: "app_id", partitionValue: <partition_value>, timeout: 4000) var autoOpen
+///
+/// This will inmeadiatly initiates a `Realm.asyncOpen()` operation which will perform all work needed to get the Realm to
+/// a usable state. (see Realm.asyncOpen() documentation)
+///
+/// This property wrapper will publish states of the current `Realm.asyncOpen()` process like progress, errors and an opened realm,
+/// which can be used to update the view
```
tgoyne, line 819: There's several typos in here. Xcode has a built-in spellcheck for comments which you can enable at Edit -> Format -> Spelling and Grammar -> Check Spelling While Typing.

Leading off the explanation with "this is very similar to this other thing" isn't very useful to a user who has alt-clicked a type to find out what it is. It's worth noting the similarity to AsyncOpen, but the docs should start by explaining what this is and what is does.

Realm/ObjectServerTests/SwiftUIServerTests.swift
```diff
@@ -0,0 +1,345 @@
...
+    override func tearDown() {
+        cancellables.forEach { $0.cancel() }
+        cancellables = []
+        super.tearDown()
+    }
+
+    var cancellables: Set<AnyCancellable> = []
+
+    // MARK: - AsyncOpen
+    func testAsyncOpenOpenRealm() {
```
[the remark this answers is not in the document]
dianaafanador3: Thanks for letting me know

RealmSwift/SwiftUI.swift
dianaafanador3: Thanks for the tip

RealmSwift/SwiftUI.swift
dianaafanador3: ok I see that we internally cache all the apps in s_apps, this is private, should we expose this internally on `RLMApp_Private`?

RealmSwift/SwiftUI.swift
dianaafanador3: What do you mean by not exposing the property wrapper?

jsflax: The property wrapper, when used as a property wrapper (which should be 99.9% of use cases), does not expose this method in any way.

RealmSwift/SwiftUI.swift
jsflax: Exposing s_apps to RLMApp_Private.h seems reasonable enough.

RealmSwift/SwiftUI.swift
jsflax: In hindsight, making `BSON` equatable probably wasn't worth the hassle of having to use `AnyBSON` everywhere. Unfortunately, removing conformance would break semver, so there are two options: 

- Either we introduce a lower level protocol with no associated type that `BSON` conforms to, and then use that for places where PartitionValue makes sense:
```swift
protocol PartitionValue {}
protocol BSON: PartitionValue, Equatable {}
```
- Or we use AnyBSON as you suggested.
 
I attempted to get this working with opaque types, but sadly there seems to be a regression with opaque types that allows this to compile.

RealmSwift/SwiftUI.swift
jsflax: E.g.,:
```swift
struct MyView {
    @AsyncOpen var asyncOpen

    // you cannot call `cancel` when the property is declared this way
}
```

dianaafanador3: but the user can so something like this 
```
.onDisappear {
            asyncOpen.cancel()
        }
```
in case they want to cancel the operation

jsflax: They would have to do:
```swift
.onDisappear {
    _asyncOpen.cancel()
}
```
Which is fine, but notably awkward.

dianaafanador3: yep, you are right, users may not use this that much, but who should provide a way to cancel the subscription to the states in a way

Realm/Tests/SwiftUISyncTestHostUITests/SwiftUISyncTestHostUITests.swift
```diff
@@ -0,0 +1,197 @@
...
+        application.terminate()
+        deleteApplicationData()
+        super.tearDown()
+    }
+
+    func deleteApplicationData() {
+        do {
+            let fileManager = FileManager.default
+            let applicationSupportDir = try fileManager.url(for: .applicationSupportDirectory,
+                                                               in: .userDomainMask,
```
jsflax, line 48: Bad indentation

RealmSwift/BSON.swift
```diff
@@ -62,15 +62,23 @@ extension UUID: BSON {
...
+extension Array: BSON, PartitionValue where Element == AnyBSON? {
 }
 
 extension NSRegularExpression: BSON {
 }
 
+/// Implementation for `PartitionValue` conformance.
+extension BSON {
+    /// Returns a `AnyBSON` for any `BSON` value.
+    public var anyBSON: AnyBSON {
```
jsflax, line 77: I would rather we don't pollute the public API with this, as it will be on many Foundation types.

RealmSwift/SwiftUI.swift
```diff
@@ -619,10 +619,20 @@ extension ThreadConfined where Self: ObjectBase {
     }
 }
 
+/// Protocol representing a PartitionValue..
+public protocol PartitionValue {
+    /// Returns a `AnyBSON` for any type conforming to it.
+    var anyBSON: AnyBSON { get }
```
jsflax, line 625: This protocol should be empty.

RealmSwift/SwiftUI.swift
```diff
@@ -643,6 +653,390 @@ extension EnvironmentValues {
...
+            self[PartitionValueEnvironmentKey] = newValue
+        }
+    }
+}
+
+@available(iOS 13.0, macOS 10.15, tvOS 13.0, watchOS 6.0, *)
+private class ObservableAsyncOpenStorage: ObservableObject {
+    var app: App?
+    var configuration: Realm.Configuration?
+    var partitionValue: AnyBSON?
```
jsflax, line 671: None of these need to be nullable– you can initialise the storage in the propertyWrapper initialiser instead of these being nullable. It also appears that `app` isn't used.

RealmSwift/SwiftUI.swift
```diff
@@ -643,6 +653,390 @@ extension EnvironmentValues {
...
+     */
+    public init<T: BSON>(appId: String? = nil, partitionValue: T, configuration: Realm.Configuration = Realm.Configuration.defaultConfiguration, timeout: UInt? = nil) {
+        var app: App
+        if let appId = appId {
+            app = App(id: appId)
+        } else if RLMApp.apps().allKeys.count == 1, // Check if there is a singular cached app
+            let cachedAppId = RLMApp.apps().allKeys.first as? String {
+            app = App(id: cachedAppId)
+        } else {
+            throwRealmException("There is no appId, either provided by the user on the property wrapper or 'any/more than 1' cached RLMApp")
```
jsflax, line 821: I would break this up into different error messages for each case to better guide the user.

RealmSwift/SwiftUI.swift
```diff
@@ -643,6 +653,390 @@ extension EnvironmentValues {
...
```
jsflax, line 988: Same comments as above.

Realm/Tests/SwiftUISyncTestHostUITests/SwiftUISyncTestHostUITests.swift
```diff
@@ -0,0 +1,197 @@
...
+// limitations under the License.
+//
+////////////////////////////////////////////////////////////////////////////
+
+import XCTest
+import RealmSwift
+
+class SwiftUISyncTestHostUITests: SwiftSyncTestCase {
+    let application = XCUIApplication()
+    static var isDataCreated = false
```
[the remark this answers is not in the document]
dianaafanador3: TODO was added to fix this later

CHANGELOG.md
```diff
@@ -43,6 +43,15 @@
 * Improve performance of calculating changesets for notifications, particularly
   for deeply nested object graphs and objects which have List or Set properties
   with small numbers of objects in the collection.
+* Add two new property wrappers for opening a Realm asynchronously in a 
+  SwiftUI View:
+    - `AsyncOpen` is a property wrapped that initiates Realm.asyncOpen
```
jsflax, line 48:
```suggestion
    - `AsyncOpen` is a property wrapper that initiates Realm.asyncOpen 
```

CHANGELOG.md
```diff
@@ -43,6 +43,15 @@
 * Improve performance of calculating changesets for notifications, particularly
   for deeply nested object graphs and objects which have List or Set properties
   with small numbers of objects in the collection.
+* Add two new property wrappers for opening a Realm asynchronously in a 
+  SwiftUI View:
+    - `AsyncOpen` is a property wrapped that initiates Realm.asyncOpen 
+       for the current user, notifying the view when there is a change in Realm asyncOpen state.
+    - `AutoOpen` behaves similar to `AsyncOpen`, but in case of no internet
```
jsflax, line 50:
```suggestion
    - `AutoOpen` behaves similarly to `AsyncOpen`, but in the case of no internet
```

CHANGELOG.md
```diff
@@ -43,6 +43,15 @@
...
+       connection this will return an opened realm. 
+* Add `EnvironmentValues.partitionValue`, this value can be injected into any view using one of
```
jsflax, line 52:
```suggestion
* Add `EnvironmentValues.partitionValue`. This value can be injected into any view using one of 
```

CHANGELOG.md
```diff
@@ -43,6 +43,15 @@
...
   for deeply nested object graphs and objects which have List or Set properties
   with small numbers of objects in the collection.
+* Add two new property wrappers for opening a Realm asynchronously in a 
+  SwiftUI View:
+    - `AsyncOpen` is a property wrapped that initiates Realm.asyncOpen 
+       for the current user, notifying the view when there is a change in Realm asyncOpen state.
+    - `AutoOpen` behaves similar to `AsyncOpen`, but in case of no internet
+       connection this will return an opened realm. 
+* Add `EnvironmentValues.partitionValue`, this value can be injected into any view using one of 
+  our new property wrappers `AsyncOpen` and `AutoOpen`.
```
jsflax, line 53:
```suggestion
  our new property wrappers `AsyncOpen` and `AutoOpen`:
```

CHANGELOG.md
```diff
@@ -51,6 +60,8 @@
 * `@Persisted<T?>` would crash when decoding a `null` value.
   ([#7332](https://github.com/realm/realm-cocoa/issues/7332), since v10.10.0).
 * Sync user profiles now correctly persist between runs.
+* Fixed `configuration(partitionValue: AnyBSON)` will set always a nil partition value
```
jsflax, line 63:
```suggestion
* Fix `configuration(partitionValue: AnyBSON)`. Previously, this was always a nil partition value
```

Realm/ObjectServerTests/SwiftServerObjects.swift
```diff
@@ -0,0 +1,188 @@
+//
+//  SwiftServerObjects.swift
+//  Realm
+//
+//  Created by Diana Maria Perez Afanador on 1/7/21.
+//  Copyright © 2021 Realm. All rights reserved.
+//
```
jsflax, line 8: Add License here and remove template.

Realm/ObjectServerTests/SwiftSyncTestCase.swift
```diff
@@ -324,6 +164,24 @@ open class SwiftSyncTestCase: RLMSyncTestCase {
                 file: (fileName), line: lineNumber)
         }
     }
+
+    public static let bigObjectCount = 2
+    public func populateRealm(user: User, partitionValue: String) {
```
jsflax, line 169: This parameter is not used. It should either be optional and set on nil or removed entirely.

Realm/ObjectServerTests/SwiftUIServerTests.swift
```diff
@@ -0,0 +1,358 @@
...
+
+        let appId = try! RealmServer.shared.createApp()
+        let appConfig = AppConfiguration(baseURL: "http://localhost:5678",
+                                         transport: AsyncOpenConnectionTimeoutTransport(),
+                                         localAppName: nil,
+                                         localAppVersion: nil)
+        let app = App(id: appId, configuration: appConfig)
+        _ = try logInUser(for: basicCredentials(app: app), app: app)
+
+        autoreleasepool {
```
jsflax, line 117: Should see if we can remove this.

RealmSwift/BSON.swift
```diff
@@ -18,9 +18,13 @@
 
 import Realm
 
+/// Protocol representing a PartitionValue..
```
jsflax, line 21:
```suggestion
/// Protocol representing a PartitionValue for synced Realms.
```

RealmSwift/SwiftUI.swift
```diff
@@ -643,6 +647,368 @@ extension EnvironmentValues {
...
+        }
+        set {
+            self[PartitionValueEnvironmentKey] = newValue
+        }
+    }
+}
+
+@available(iOS 14.0, macOS 11.0, tvOS 14.0, watchOS 7.0, *)
+private class ObservableAsyncOpenStorage: ObservableObject {
+    var app: App
```
jsflax, line 663: Should make these private.

Realm/ObjectServerTests/SwiftUIServerTests.swift
```diff
@@ -0,0 +1,358 @@
...
+        asyncOpen.projectedValue
+            .sink { asyncOpenState in
+                if case let .open(realm) = asyncOpenState {
+                    XCTAssertNotNil(realm)
+                    ex.fulfill()
+                }
+            }
+            .store(in: &cancellables)
+        waitForExpectations(timeout: 10.0)
+        asyncOpen.cancel()
```
tgoyne, line 60: I think the bulk of these tests could be extracted to a helper function pretty easily. Something like `func asyncOpen(partitionValue: String, handler: (AsyncOpenState, XCTestExpectation) -> Void)` which contains these lines and then calls the provided callback inside the `.sink()` function would reduce a bunch of these tests to just the code that's different between them.

Realm/Tests/SwiftUISyncTestHost/ContentView.swift
```diff
@@ -0,0 +1,310 @@
...
+////////////////////////////////////////////////////////////////////////////
+
+import SwiftUI
+import RealmSwift
+import Combine
+
+enum LoggingViewState {
+    case initial
+    case logging
+    case logged
```
tgoyne, line 26:
```suggestion
    case loggingIn
    case loggedIn
```

Realm/Tests/SwiftUISyncTestHost/ContentView.swift
```diff
@@ -0,0 +1,310 @@
...
+              partitionValue: ProcessInfo.processInfo.environment["function_name"]!,
+              timeout: 2000) var autoOpen
+
+    var body: some View {
+        VStack {
+            switch autoOpen {
+            case .connecting:
+                ProgressView()
+            case .waitingForUser:
+                ProgressView("Waiting for user to logged in...")
```
tgoyne, line 192:
```suggestion
                ProgressView("Waiting for user to log in...")
```

Realm/Tests/SwiftUISyncTestHost/ContentView.swift
```diff
@@ -0,0 +1,310 @@
...
+                    }
+            }
+        }
+    }
+}
+
+struct AsyncOpenPartitionView: View {
+    @AsyncOpen(appId: ProcessInfo.processInfo.environment["app_id"]!,
+               partitionValue: "wrong_partition_value",
+               timeout: 2000) var asyncOpen
```
tgoyne, line 228: `var` should be at the start of a new line when the property wrapper is multiple lines.

Realm/Tests/SwiftUISyncTestHostUITests/SwiftUISyncTestHostUITests.swift
```diff
@@ -0,0 +1,198 @@
...
+    override func setUp() {
+        continueAfterFailure = false
+        super.setUp()
+        // TODO: This should be a static method
+        if !SwiftUISyncTestHostUITests.isDataCreated {
+            let user = logInUser(for: basicCredentials(withName: #function, register: true))
+            populateRealm(user: user, partitionValue: #function)
```
tgoyne, line 31: `#function` isn't doing something sensible here as it's always just "setUp()". In other places we use `#function` as the partition value as an easy way to use the test name, but here if it's intending to use a single partition for every test then it should just have a sensible partition value in a constant.

CHANGELOG.md
```diff
@@ -1,11 +1,21 @@
 x.y.z Release notes (yyyy-MM-dd)
 =============================================================
 ### Enhancements
-* None.
+
```
jsflax, line 4: Extra new line

CHANGELOG.md
```diff
@@ -1,11 +1,21 @@
...
+    - `AutoOpen` behaves similarly to `AsyncOpen`, but in the case of no internet
+       connection this will return an opened realm. 
+* Add `EnvironmentValues.partitionValue`. This value can be injected into any view using one of 
+  our new property wrappers `AsyncOpen` and `AutoOpen`:
+  `MyView().environment(\.partitionValue, "partitionValue")`. 
 
 ### Fixed
-* <How to hit and notice issue? what was the impact?> ([#????](https://github.com/realm/realm-cocoa/issues/????), since v?.?.?)
-* None.
+
```
jsflax, line 16: Extra newline

RealmSwift/SwiftUI.swift
```diff
@@ -643,6 +647,368 @@ extension EnvironmentValues {
...
+        set {
+            self[PartitionValueEnvironmentKey.self] = newValue
+        }
+    }
+}
+
+@available(iOS 14.0, macOS 11.0, tvOS 14.0, watchOS 7.0, *)
+private class ObservableAsyncOpenStorage: ObservableObject {
+    private var app: App
+    var configuration: Realm.Configuration
```
jsflax, line 664: These should be private unless needed elsewhere.

RealmSwift/SwiftUI.swift
```diff
@@ -643,6 +647,368 @@ extension EnvironmentValues {
...
+    }
+
+    func asyncOpen() -> AnyPublisher<RealmPublishers.AsyncOpenPublisher.Output, RealmPublishers.AsyncOpenPublisher.Failure> {
+        if let currentUser = app.currentUser,
+           currentUser.isLoggedIn {
+            return asyncOpenForUser(app.currentUser!, partitionValue: partitionValue, configuration: configuration)
+        } else {
+            asyncOpenState = .waitingForUser
+            return app.objectWillChange
+                .filter { $0.currentUser != nil }
```
jsflax, line 682: Can probably use `compactMap` here:
`.compactMap { $0.currentUser }`
to make the code simpler/more legible.

RealmSwift/SwiftUI.swift
```diff
@@ -643,6 +647,368 @@ extension EnvironmentValues {
...
+    @Environment(\.partitionValue) var partitionValue
+    @ObservedObject private var storage: ObservableAsyncOpenStorage
+
+    private func asyncOpen() {
+        storage.asyncOpen()
+            .sink { completion in
+                if case .failure(let error) = completion {
+                    self.storage.asyncOpenState = .error(error)
+                }
+                cancel()
```
jsflax, line 777: Why does this cancel even though it's completed?

RealmSwift/SwiftUI.swift
```diff
@@ -643,6 +647,368 @@ extension EnvironmentValues {
...
+@propertyWrapper public struct AsyncOpen<Partition>: DynamicProperty where Partition: BSON {
+    @Environment(\.realmConfiguration) var configuration
+    @Environment(\.partitionValue) var partitionValue
+    @ObservedObject private var storage: ObservableAsyncOpenStorage
+
+    private func asyncOpen() {
+        storage.asyncOpen()
+            .sink { completion in
+                if case .failure(let error) = completion {
+                    self.storage.asyncOpenState = .error(error)
```
jsflax, line 775: It's a bit awkward to me that the failure case sets the asyncOpenState on the storage but the other cases don't. Feels like this logic should all exist in one place (probably on the Storage itself).

RealmSwift/SwiftUI.swift
```diff
@@ -643,6 +647,368 @@ extension EnvironmentValues {
...
+        storage = ObservableAsyncOpenStorage(app: app, configuration: configuration, partitionValue: bsonValue)
+        asyncOpen()
+    }
+
+    public mutating func update() {
+        if let partitionValue = partitionValue as? Partition {
+            let bsonValue = AnyBSON(partitionValue)
+            if storage.partitionValue != bsonValue {
+                storage.partitionValue = bsonValue
+
```
jsflax, line 847: Newline seems odd here.

RealmSwift/SwiftUI.swift
dianaafanador3: configuration and partitionValue are overrided when we use them as an environment value, that's why this is not set to private to been able to do that

RealmSwift/SwiftUI.swift
dianaafanador3: This behaves different for each property wrapper, that's why this logic is done by the property wrapper instead of the storage

RealmSwift/SwiftUI.swift
dianaafanador3: you are right, this shouldn't be done after completed

RealmSwift/SwiftUI.swift
jsflax: Or even better:
`.compactMap(\.currentUser)`

dianaafanador3: Pretty nice, thanks

RealmSwift/SwiftUI.swift
```diff
@@ -643,6 +647,364 @@ extension EnvironmentValues {
...
+                partitionValue: Partition,
+                configuration: Realm.Configuration = Realm.Configuration.defaultConfiguration,
+                timeout: UInt? = nil) {
+        var app: App
+        if let appId = appId {
+            app = App(id: appId)
+        } else if RLMApp.appIds().count == 1, // Check if there is a singular cached app
+            let cachedAppId = RLMApp.appIds().first as? String {
+            app = App(id: cachedAppId)
+        } else if RLMApp.appIds().count > 1 {
```
tgoyne, line 822: This should call `RLMApp.appIds()` exactly once and then reuse the local value. Multiple calls to it could potentially produce different values if a new App is created on a different thread while this is happening, and while it appears that it'll currently work reasonably if that does happen, that's the sort of thing that can produce weird and confused bugs.

RealmSwift/BSON.swift
```diff
@@ -18,9 +18,13 @@
 
 import Realm
 
+/// Protocol representing a PartitionValue for synced Realms.
```
tgoyne, line 21:
```suggestion
/// A tag protocol which marks types that can be used as the partition value
/// for synchronized Realms.
```

I'm confused about how this could be used. AFAICT it's only used by the environment key, and I don't see how you can actually open a Realm from the value read from the environment key? If there's somewhere testing the env key I couldn't find it.

RealmSwift/SwiftUI.swift
```diff
@@ -643,6 +647,364 @@ extension EnvironmentValues {
...
+            throwRealmException("There is no appId, either provided by the user on the property wrapper or more than 1 cached RLMApp")
+        } else {
+            throwRealmException("There is no appId, either provided by the user on the property wrapper or any cached RLMApp")
```
tgoyne, line 825: Some stabs at some better wording:

"No appId was provided to AsyncOpen and no Apps have been initialized."
"No appId was provided to AsyncOpen and more than one App has been initialized."

"Cannot AsyncOpen the Realm because no appId was found. You must either explicitly pass an appId to @AsyncOpen() or initialize an App before displaying your View."
"Cannot AsyncOpen the Realm because more than one appId was found. When using multiple Apps you must explicitly pass an appId to @AsyncOpen to indicate which to use."

RealmSwift/SwiftUI.swift
```diff
@@ -643,6 +647,364 @@ extension EnvironmentValues {
...
+        }
```
tgoyne, line 973: Similar notes apply as above. It'd also be nice to extract what we can from these initializers into a private helper function rather than duplicating them.

RealmSwift/BSON.swift
dianaafanador3: There are some UI tests testing this, You can see it implemented on the view https://github.com/realm/realm-cocoa/blob/496105721836de76f5a8a37614d75e383f10cf3d/Realm/Tests/SwiftUISyncTestHost/ContentView.swift#L69 and the test https://github.com/realm/realm-cocoa/blob/496105721836de76f5a8a37614d75e383f10cf3d/Realm/Tests/SwiftUISyncTestHostUITests/SwiftUISyncTestHostUITests.swift#L92 for `AsyncOpen`. You can find implementation and tests for `AutoOpen` in the same files

RealmSwift/SwiftUI.swift
dianaafanador3: Thanks for the suggestions

RealmSwift/SwiftUI.swift
dianaafanador3: I added a helper method in the property wrapper storage, https://github.com/realm/realm-cocoa/blob/75896163bf278bceb62e997b2e75b317ae4371dd/RealmSwift/SwiftUI.swift#L710

tgoyne (approved):

jsflax (approved):

PR #7342, opened 2021-07-14
Add type safe methods for aggregatable and sortable operations on collections

jsflax:
Add type safe methods for:
    - `RealmCollection.min(ofProperty:)`
    - `RealmCollection.max(ofProperty:)`
    - `RealmCollection.average(ofProperty:)`
    - `RealmCollection.sum(ofProperty:)`
    - `RealmCollection.sorted(byKeyPath:, ascending:)`
    - `Results.distinct(by:)`
    - `SortDescriptor.init(keyPath:, ascending:)
    
  Calling these methods can now be done via Swift keyPaths, like so:
  ```swift
  class Person: Object {
    @Persisted var name: String
    @Persisted var age: Int
  }
  
  let persons = realm.objects(Person.self)
  persons.min(ofProperty: \.age)
  persons.max(ofProperty: \.age)
  persons.average(ofProperty: \.age)
  persons.sum(ofProperty: \.age)
  persons.sorted(byKeyPath: \.age)
  persons.sorted(by: [SortDescriptor(keyPath: \Person.age)])
  persons.distinct(by: [\Person.age])
  ``` 

~Deprecated:~
    ~- `RealmCollection.min(ofProperty:)`~
    ~- `RealmCollection.max(ofProperty:)`~
    ~- `RealmCollection.average(ofProperty:)`~
    ~- `RealmCollection.sum(ofProperty:)`~
    ~- `RealmCollection.sorted(byKeyPath:, ascending:)`~
 ~In favor of their type safe counterparts. String based keyPaths are more error prone,~
 ~and the current versions of these methods allow for using string based keyPaths on types~
 ~that were potentially unsupported.~
  
  ~If previously using any of the above methods by passing in "self" (a valid keyPath for~
  ~`AddableType`, `MinMaxType`, and `Comparable` types), migrate to using the above methods~
  ~with no arguments, e.g., `myIntList.min()`.~


tgoyne:
The string versions should not be deprecated. They still allow doing things which inherently cannot be done using typed keypaths, such as aggregating collections inside migrations (e.g. if you replaced a List property with just the minimum value of that List because it turned out it was the only thing you needed) and when opening a Realm in fully dynamic mode.

jsflax:
Fair enough. I'm happy to un-deprecate them. Though the old `sorted(byKeyPath:)` on Map returns Results<Element?> in the case of Value being of type ObjectBase, so I do feel that method should be deprecated in favour of another that return Results<Element>.

jsflax:
Removed deprecations.

CHANGELOG.md
```diff
@@ -1,14 +1,38 @@
 x.y.z Release notes (yyyy-MM-dd)
 =============================================================
 ### Enhancements
-* None.
+* Add type safe methods for:
+    - `RealmCollection.min(ofProperty:)`
+    - `RealmCollection.max(ofProperty:)`
+    - `RealmCollection.average(ofProperty:)`
+    - `RealmCollection.sum(ofProperty:)`
```
tgoyne, line 8: I think these should just be `min(of:)` etc. The `Property` is there to indicate what the string is and so isn't needed with a better type.

CHANGELOG.md
```diff
@@ -1,14 +1,38 @@
...
+    - `RealmCollection.sorted(byKeyPath:, ascending:)`
+    - `Results.distinct(by:)`
+    - `SortDescriptor.init(keyPath:, ascending:)
```
tgoyne, line 11: `.init` is not normally written here.

Realm/Tests/DictionaryPropertyTests.m
```diff
@@ -587,6 +587,25 @@ - (void)testObjectAggregate {
     test();
 }
 
+- (void)testPrimitiveAggregate {
```
tgoyne, line 590: This is tested very thoroughly in PrimitiveDictionaryPropertyTests.m

RealmSwift/List.swift
```diff
@@ -686,3 +686,33 @@ extension List: PropertyNameConvertible {
...
+        self.max(ofProperty: _name(for: property))
+    }
+
+    public func sum<T>(ofProperty property: KeyPath<Element, T>) -> T where T: AddableType {
+        self.sum(ofProperty: _name(for: property))
+    }
+
+    public func average<T>(ofProperty property: KeyPath<Element, T>) -> T? where T: AddableType {
+        self.average(ofProperty: _name(for: property))
+    }
```
tgoyne, line 707: These look like they should be implementable as a default protocol extension rather than on each type.

RealmSwift/RealmCollection.swift
```diff
@@ -530,6 +530,103 @@ public protocol RealmCollection: RealmCollectionBase {
     func thaw() -> Self?
 }
 
+// MARK: Aggregatable
+
+/**
+ Protocol for RealmCollections where the Value is of an Object type that
+ enables aggregatable operations.
+ */
+public protocol Aggregatable: RealmCollection where Element: ObjectBase {
```
tgoyne, line 539: Why does this need to be a new protocol instead of just a conditional extension on RealmCollection?

RealmSwift/RealmKeyedCollection.swift
```diff
@@ -308,6 +308,87 @@ public protocol RealmKeyedCollection: Sequence, ThreadConfined, CustomStringConv
...
+     youngest to oldest based on their `age` property, you might call
+     `students.sorted(byKeyPath: "age", ascending: true)`.
+
+     - warning: Collections may only be sorted by properties of boolean, `Date`, `NSDate`, single and double-precision
+                floating point, integer, and string types.
+
+     - parameter keyPath:   The key path to sort by.
+     - parameter ascending: The direction to sort in.
+     */
+    func sorted<T: Comparable>(byKeyPath keyPath: KeyPath<Value.Wrapped, T>, ascending: Bool) -> Results<Value.Wrapped>
```
tgoyne, line 376: Does this drop null values when sorting? That seems like a weird thing for it to do, but unless it is doing that then the return type needs to be `Results<Value>`

CHANGELOG.md
```diff
@@ -1,14 +1,38 @@
...
+    
+  Calling these methods can now be done via Swift keyPaths, like so:
```
leemaguire, line 13: Details on Map are missing

Realm/Tests/DictionaryPropertyTests.m
```diff
@@ -587,6 +587,25 @@ - (void)testObjectAggregate {
...
+    [obj.intDictionary minOfProperty:@"self"];
+
+    [realm transactionWithBlock:^{
+        DictionaryPropertyObject *managed = [DictionaryPropertyObject createInRealm:realm withValue:@[]];
+        managed.intDictionary[@"age"] = @20;
+        managed.intDictionary[@"weight"] = @400;
+        obj = managed;
+    }];
+
+    [obj.intDictionary minOfProperty:@"self"];
```
leemaguire, line 605: Is this left over code from when you were experimenting with the aggregates?

RealmSwift/Tests/MapTests.swift
```diff
@@ -547,6 +547,43 @@ class MapTests: TestCase {
...
+    }
+
+    func testKeyedAggregatable() {
+        let map = realm.create(SwiftMapPropertyObject.self, value: [])
+        map.intMap["key"] = SwiftIntObject(value: [1])
+        map.intMap["key2"] = SwiftIntObject(value: [2])
+        map.intMap["key3"] = SwiftIntObject(value: [3])
+        try! realm.commitWrite()
+
+        if map.realm != nil {
```
leemaguire, line 574: should test the unmanaged path too

Realm/Tests/DictionaryPropertyTests.m
jsflax: Yes.

RealmSwift/RealmKeyedCollection.swift
jsflax: Why does that seem like a weird thing for it to do? I considered this a nicer alternative. What does a sorted Results with nil values mean semantically?

RealmSwift/RealmCollection.swift
jsflax: It made more sense to me when I was planning on deprecating the string based methods, but since we've reversed that, I agree that it no longer makes sense.

jsflax:
@leemaguire @tgoyne all comments should be addressed.

RealmSwift/Tests/MapTests.swift
```diff
@@ -547,6 +547,48 @@ class MapTests: TestCase {
...
+            XCTAssertEqual(results2[0]?.stringCol, "apples")
+            XCTAssertEqual(results2[1]?.stringCol, "bananas")
+            XCTAssertEqual(results2[2]?.stringCol, "cockroach")
+        } else {
+            assertThrows(map.sorted(by: \.stringCol, ascending: true))
+        }
+    }
+
+    func testKeyedAggregatable() {
+        let map = realm.create(SwiftMapPropertyObject.self, value: [])
```
leemaguire, line 570: won't this always be managed?

jsflax: Ah right, should add in an add and not create it through the Realm. Though I suppose it's technically not managed until the commit.

RealmSwift/RealmKeyedCollection.swift
tgoyne: Normally a sorted collection is always the same length as the unsorted one because sorting just rearranges values without removing any. We've taken the general position that null is less than all other values.

I assume that this will currently just crash at runtime if there's any null values since there's nothing actually filtering them out.

jsflax: I have since removed it and have gone back to returning the nullable element type.

leemaguire (approved):
🦄

RealmSwift/RealmCollection.swift
```diff
@@ -530,6 +530,103 @@ public protocol RealmCollection: RealmCollectionBase {
...
+ enables aggregatable operations.
+ */
+public extension RealmCollection where Element: ObjectBase {
+    /**
+     Returns the minimum (lowest) value of the given property among all the objects in the collection, or `nil` if the
+     collection is empty.
+
+     - warning: Only a property whose type conforms to the `MinMaxType` protocol can be specified.
+
+     - parameter keyPath: The name of a property whose minimum value is desired.
```
tgoyne, line 546: These don't take a name.

tgoyne (approved):

PR #7351, opened 2021-07-20
Update core to v11.1.0

ericjordanmossman:


ericjordanmossman:
I looked at older PRs that updated core. Here's a list of things I didn't on purpose and why:
- I didn't change cocoa version in either Package.swift or dependencies.list because that's to be determined by what other changes are added before release.
- I didn't add any of the core 11.1.0 change log entries because they didn't appear relevant to a developer using cocoa directly. For example, there is a small core api breaking change, but since we don't hook into it, it should affect someone developing with realm-cocoa.
- I didn't update build.sh because it should change build compatibility.
- I didn't update Realm-Info.plist because I didn't change the cocoa version.

Let me know if they're needed.

tgoyne:
The user profile data being persisted between runs is a user-relevant change that should be in the changelog unless we just don't expose the user profile data at all currently? That change is also broken because it changed the schema of the metadata realm without bumping the schema version, so we can't actually use this release.

jsflax:
It is user relevant but persisting the data was the expected behaviour and a bug otherwise. I'll file a PR tomorrow to address the schema issue.

ericjordanmossman:
Got it, thanks.

PR #7354, opened 2021-07-21
Update core to v11.1.1

ericjordanmossman:
Changelog entries not carried over from core v11.1.1 and v11.1.0:
> Opening a synchronized Realm created with a version older than v11.1.0 would fail due to a schema change in the metadata Realm which did not bump the schema version (since v11.1.0).

- This wouldn't apply because cocoa never has or ever will release with 11.1.0.

> User profile fields are now methods instead of fields.

- This part of the core api was never exposed to users. The fact that user profiles persist between runs is the relevant part, which is part of the changelog.

> Change notifications can now be filtered via a key path. This keypath is passed via the add_notification_callback on Object, Set, List and Result. If such a key path was provided when adding a notification callback it will only ever be executed when a changed property was covered by this filter. Multiple key path filters can be provided when adding a single notification callback.

- This should be added with a cocoa specific entry once the feature is added.

Does this make sense? Is anything else missing?


CHANGELOG.md
```diff
@@ -5,12 +5,19 @@ x.y.z Release notes (yyyy-MM-dd)
   This allows you to select elements in a collection with a given IndexSet (Cocoa [#7298](https://github.com/realm/realm-cocoa/issues/7298).
 * Add `App.emailPasswordAuth.retryCustomConfirmation(email:completion:)` and `[App.emailPasswordAuth retryCustomConfirmation:completion:]`. 
   These functions support retrying a [custom confirmation](https://docs.mongodb.com/realm/authentication/email-password/#run-a-confirmation-function) function.
+* Improve performance of creating collection notifiers for Realms with a complex schema. 
+  This means that the first run of a synchronous query, first call to observe() on a collection, 
+  or any call to find_async() will do significantly less work on the calling thread.
```
tgoyne, line 10: The wording on this needs to be adjusted a little. find_async() isn't something we have in Cocoa and we don't distinguish between sync and async queries like Java does, so it should be something like "the first run of a query or first call to observe() on a collection will do significant less work".

CHANGELOG.md
```diff
@@ -5,12 +5,19 @@ x.y.z Release notes (yyyy-MM-dd)
...
+* Improve performance of calculating changesets for notifications, particularly 
+  for deeply nested object graphs and objects which have List or Set properties 
+  with small numbers of objects in the collection.
 
 ### Fixed
 * `RealmProperty<T?>` would crash when decoding a `null` json value.
   ([Cocoa #7323](https://github.com/realm/realm-cocoa/issues/7323), since v10.8.0)
 * `@Persisted<T?>` would crash when decoding a `null` value.
   ([#7332](https://github.com/realm/realm-cocoa/issues/7332), since v10.10.0).
+* User profile now correctly persists between runs.
```
tgoyne, line 20: This should probably mention that it's the profile for sync users.

ericjordanmossman:
Are the build errors for 13.0 an issue with this version?

tgoyne:
Xcode 13 beta 3 has a broken Combine.framework for iOS, which is responsible for all of the failures except for swiftpm-thread. The swiftpm-thread failure is probably just a networking hiccup.

tgoyne (approved):

Issue #7344, opened 2021-07-17
Build error "Header 'Realm/RLMDictionary_Private.h' not found" in module.modulemap


nicolas255:
## Goals

I am trying to use version of Realm in a SPM package.

## Expected Results

Xcode should resolve the dependencies successfully and then the package should build successfully.

## Actual Results

The dependency is resolved successfully but then the build fails with two errors :

![Screenshot 2021-07-17 at 21 04 43](https://user-images.githubusercontent.com/7323838/126047174-2594b0b4-a49b-4903-83bf-a45641e97406.png)

## Steps for others to Reproduce

1. Create a swift package with Xcode.
2. Add `realm-cocoa` as a dependency.
3. Start building.

## Code Sample

This is the `Package.swift` file :
```swift
let package = Package(
    name: "MyLibrary",
    platforms: [.iOS(.v14)],
    products: [
        .library(
            name: "MyLibrary",
            targets: ["MyLibrary"]
        ),
    ],
    dependencies: [
        .package(name: "Realm", url: "https://github.com/realm/realm-cocoa", .upToNextMajor(from: "10.10.0")),
    ],
    targets: [
        .target(
            name: "MyLibrary",
            dependencies: [
                .product(name: "Realm", package: "Realm"),
                .product(name: "RealmSwift", package: "Realm"),
            ]
        ),
        .testTarget(
            name: "MyLibraryTests",
            dependencies: [
                .target(name: "MyLibrary"),
                .product(name: "Realm", package: "Realm"),
                .product(name: "RealmSwift", package: "Realm"),
            ]
        ),
    ]
)
```

## Version of Realm and Tooling

Realm framework version:  
From `10.8.0` to `10.10.0`.  
Build succeeds with `10.7.7` or earlier.

Xcode version:  
`12.5.1`

iOS/OSX version:  
I tried targeting an iPhone running on iOS 14.6 and simulator running iOS 14.3.

Dependency manager + version:  
SPM, bundled with Xcode

leemaguire:
Hi @nicolas255 I cannot reproduce your issue, have you cleaned your build and derived data?

nicolas255:
Hi @leemaguire, I tested all of that but no luck there…

In the meantime, I also tried to upgrade from 10.7.4 to 10.10.0 in an existing project where packages are managed by Xcode (it's an iOS app project) and I have the same two issues. Then I tried to replicate them in a fresh iOS project and same issues again.

[...]

PR #7360, opened 2021-07-25
Fix the RLMDictionary_Private.h symlink

nicolas255:
~~Fixes the symlink to RLMDictionary_Private.h which causes build errors on case sensitive file systems.~~

See #7344.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,24 @@
+x.y.z Release notes (yyyy-MM-dd)
+=============================================================
+### Enhancements
+* None.
+
+### Fixed
+* Fixed a symlink which prevented Realm from building on case sensitive file systems. 
+  ([#7344](https://github.com/realm/realm-cocoa/issues/7344), since v10.8.0)
+
+<!-- ### Breaking Changes - ONLY INCLUDE FOR NEW MAJOR version -->
+
+### Compatibility
+* Realm Studio: 11.0.0 or later.
+* APIs are backwards compatible with all previous releases in the 10.x.y series.
+* Carthage release for Swift is built with Xcode 12.5.1.
+* CocoaPods: 1.10 or later.
+* Xcode: 12.2-13.0 beta 3.
+
+### Internal
+* Upgraded realm-core from ? to ?
+
 10.11.0 Release notes (2021-07-22)
 =============================================================
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,9 +1,22 @@
 x.y.z Release notes (yyyy-MM-dd)
 =============================================================
 ### Enhancements
-* None.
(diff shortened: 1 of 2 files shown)


nicolas255:
So this PR is not ready to be merge and actually breaks the build (at least on a case sensitive file system) with the same error as the one I described in #7344. But the reason is not the same and I have absolutely no idea how this happened…

If you manually clone the project, everything is fine. But when the project is added to a Swift package, the symlink in the checkout in the derived data is broken in a weird way:

Screenshot 2021-07-25 at 22 47 33

As you can see here the "?" at the end of the symlink. It's actually a newline character and I have absolutely no idea how it got there…

Here is a hex dump of the symlink after a `git clone`:
```
➜  Realm git:(master) readlink -n RLMDictionary_Private.h > foo
➜  Realm git:(master) hexdump -C foo
00000000  2e 2e 2f 2e 2e 2f 52 65  61 6c 6d 2f 52 4c 4d 44  |../../Realm/RLMD|
00000010  69 63 74 69 6f 6e 61 72  79 5f 50 72 69 76 61 74  |ictionary_Privat|
00000020  65 2e 68                                          |e.h|
00000023
```

And here is a hex dump of the same symlink from the derived data:
```
➜  Realm git:(dba32b4f7) readlink -n RLMDictionary_Private.h > foo
➜  Realm git:(dba32b4f7) hexdump -C foo
00000000  2e 2e 2f 2e 2e 2f 52 65  61 6c 6d 2f 52 4c 4d 44  |../../Realm/RLMD|
00000010  69 63 74 69 6f 6e 61 72  79 5f 50 72 69 76 61 74  |ictionary_Privat|
00000020  65 2e 68 0a                                       |e.h.|
00000024
```

leemaguire:
@nicolas255 does the newline / `?` appear in a fresh project with the same setup?

nicolas255:
@leemaguire I'm not sure what kind of sorcery was at play here but somehow, Xcode kept checking out the wrong commit when specifying `https://github.com/nicolas255/realm-cocoa` on `master` event after clearing out the derived data and reseting the package cache. It was only after restarting Xcode that it decided to check out the correct commit and then it worked… So never mind, the new symlink seems to be working just fine (even if I'm still confused).

And to answer your question, yes, creating another new project actually worked and that is what pushed me to check the commits directly 🙏.

leemaguire (approved):
We will need to provide a changelog entry for this

leemaguire:
@nicolas255 Could you please sign the CLA located at the bottom of this doc so that we may proceed https://github.com/realm/realm-cocoa/blob/master/CONTRIBUTING.md

nicolas255:
@leemaguire I already did before I submitted the PR (filled up the form using the `nicolas255`).

PR #7364, opened 2021-07-27
Fix compile warning in Xcode 13 beta 4

cltnschlosser:
image


jsflax:
Looks good @cltnschlosser. Thanks for the fix.

PR #7367, opened 2021-07-28
Add async Realm.asyncOpen and App.login methods

jsflax:
@dianaafanador3 this should get you started for `async/await`.


RealmSwift/Realm.swift
```diff
@@ -1005,3 +1005,35 @@ extension Realm {
...
+     - parameter configuration: A configuration object to use when opening the Realm.
+     - parameter callbackQueue: The dispatch queue on which the callback should be run.
+     - returns: An open Realm.
+     */
+    @discardableResult
+    public static func asyncOpen(configuration: Realm.Configuration = .defaultConfiguration) async throws -> Realm {
+        try await withCheckedThrowingContinuation { (continuation: CheckedContinuation<Realm, Swift.Error>) in
+            RLMRealm.asyncOpen(with: configuration.rlmConfiguration, callbackQueue: .main, callback: { rlmRealm, error in
+                if let realm = rlmRealm.flatMap(Realm.init) {
+                    continuation.resume(with: .success(realm.freeze()))
```
tgoyne, line 1032: Why is this freezing the Realm?

jsflax: Before I added freeze/thaw, using the returned Realm at the call site resulted in a `Realm accessed from incorrect thread` error.

tgoyne: Freezing and thawing the Realm is pretty much the least efficient way of handing it over to a different thread possible. Just opening the Realm using the passed-in config after the asyncOpen completes would be better. It also shouldn't be using .main for the callback queue since it then dispatches back to the continuation afterwards anyway.

jsflax: Fair point, will fix.

jsflax: Fixed now.

Realm/ObjectServerTests/SwiftObjectServerTests.swift
```diff
@@ -2931,4 +2931,45 @@ class CombineObjectServerTests: SwiftSyncTestCase {
     }
 }
 
+@available(macOS 12.0, *)
+class AsyncAwaitObjectServerTests: SwiftSyncTestCase {
+    func testAsyncOpenStandalone() async throws {
+        try autoreleasepool {
+            let realm = try Realm()
+            try! realm.write {
+                (0..<10000).forEach { _ in realm.add(SwiftPerson(firstName: "Charlie", lastName: "Bucket")) }
```
tgoyne, line 2940: Adding this many objects seems like it'd just make the test slow without making it any better at catching bugs.

Realm/ObjectServerTests/SwiftObjectServerTests.swift
```diff
@@ -2931,4 +2931,45 @@ class CombineObjectServerTests: SwiftSyncTestCase {
...
+            try! realm.write {
+                realm.add(SwiftHugeSyncObject.create())
+                realm.add(SwiftHugeSyncObject.create())
+            }
+            let progressEx = self.expectation(description: "Should upload")
+            let token = realm.syncSession!.addProgressNotification(for: .upload, mode: .forCurrentlyOutstandingWork) {
+                if $0.isTransferComplete {
+                    progressEx.fulfill()
+                }
+            }
```
tgoyne, line 2963: Any particular reason for using this rather than just `waitForUploads(for: realm)`?

RealmSwift/Realm.swift
```diff
@@ -1005,3 +1005,33 @@ extension Realm {
 
 /// The type of a block to run for notification purposes when the data in a Realm is modified.
 public typealias NotificationBlock = (_ notification: Realm.Notification, _ realm: Realm) -> Void
+
+@available(macOS 12.0, tvOS 15.0, iOS 15.0, watchOS 8.0, *)
+extension Realm {
+    private static let asyncOpenQueue = DispatchQueue(label: "io.realm.swift.asyncOpenDispatchQueue")
```
tgoyne, line 1011: This looks unused?

RealmSwift/Realm.swift
```diff
@@ -1005,3 +1005,33 @@ extension Realm {
...
+     Opening a Realm asynchronously will perform all work needed to get the Realm to
+     a usable state (such as running potentially time-consuming migrations) on a
+     background thread before dispatching back to the continiuation. In addition,
+     synchronized Realms wait for all remote content available at the time the
+     operation began to be downloaded and available locally.
+
+     - parameter configuration: A configuration object to use when opening the Realm.
+     - returns: An open Realm.
+     */
+    @discardableResult
```
tgoyne, line 1024: This shouldn't be discardable. The callback-based asyncOpen is because it returns an object that can be used to observe the progress, but using that object isn't required.

RealmSwift/Realm.swift
```diff
@@ -1005,3 +1005,33 @@ extension Realm {
...
+    public static func asyncOpen(configuration: Realm.Configuration = .defaultConfiguration) async throws -> Realm {
```
tgoyne, line 1025: Should this maybe just be `open()`? The fact that the function is `async` indicates that it's async, and I think the Swift team is hoping to avoid having function names with async prefixes.

jsflax: In that case, should we just scrap the static method and add an async init instead? Thoughts being that having a method named open (even if marked async) seems a bit odd to have alongside your standard init.

Suppose it would need to be well documented as a replacement for asyncOpen.

tgoyne: Would that even be possible? I don't think you can overload on just asyncness.

jsflax: You can: https://github.com/apple/swift-evolution/blob/main/proposals/0296-async-await.md#overloading-and-overload-resolution

Swift resolves which version of the method to use based on if you are calling from an async context. So anybody opening a Realm in an async context would incidentally be using async open if we overloaded `init`.

tgoyne: That seems like probably an issue because asyncOpen isn't just an async version of init. Maybe we should use this opportunity to rethink this a bit and make it so that it is? We could have `init(configuration: Realm.Config, behavior: DownloadBehavior = .openImmediately) async throws` so that `try! Realm()` and `await try! Realm()` do the same thing other than the async part, and then you'd pass `behavior: .alwaysDownloadLatest` or `behavior: .downloadOnFirstOpen` to opt in to other behavior.

(I think "download behavior" is what other SDKs are calling this but it's a really awkward name)

Realm/ObjectServerTests/SwiftObjectServerTests.swift
jsflax: No, believe I picked it up from another test.

RealmSwift/Realm.swift
```diff
@@ -1005,3 +1005,60 @@ extension Realm {
...
+     Opening a Realm asynchronously will perform all work needed to get the Realm to
+     a usable state (such as running potentially time-consuming migrations) on a
+     background thread before dispatching back to the continiuation. In addition,
+     synchronized Realms wait for all remote content available at the time the
+     operation began to be downloaded and available locally.
+
+     - parameter configuration: A configuration object to use when opening the Realm.
+     - parameter downloadBehavior: The behavior in how to open the realm.
+     - returns: An open Realm.
+     */
```
tgoyne, line 1038: I took a stab at writing more correct docs for this:

```suggestion
    /// Options for when to download all data from the server before opening
    /// a synchronized Realm.
    @frozen public enum OpenBehavior {
        /// Immediately return the Realm as if the synchronous initializer was
        /// used. If this is the first time that the Realm has been opened on
        /// this device, the Realm file will initially be empty. Synchronized
        /// Realms will contact the server and download new data in the
        /// background.
        case never

        /// Always open the Realm asynchronously and download all data from the
        /// server before returning the Realm. This mode will fail to open the
        /// Realm if the device is currently offline.
        case always

        /// Open the Realm asynchronously the first time it is opened on the
        /// current device, and then synchronously afterwards. This mode is
        /// suitable if you wish to wait to download the server-side data the
        /// first time your app is launched on each device, but afterwards
        /// support offline launches using the existing local data.
        ///
        /// Note that if .once is used multiple times simultaneously then calls
        /// after the first may see partial local data from the first call and
        /// not wait for the download.
        case once
    }
    /**
     Obtains a `Realm` instance with the given configuration, possibly asynchronously.

     By default this simply returns the Realm instance exactly as if the
     synchronous initializer was used. It optionally can instead open the Realm
     asynchronously, performing all work needed to get the Realm to a usable
     state on a background thread. For local Realms, this means that migrations
     will be run in the background, and for synchronized Realms all data will
     be downloaded from the server before the Realm is returned.

     - parameter configuration: A configuration object to use when opening the Realm.
     - parameter downloadBehavior: When opening the Realm should first download
                                   all data from the server.
     - parameter queue: An optional dispatch queue to confine the Realm to. If
                        given, this Realm instance can be used from within
                        blocks dispatched to the given queue rather than on the
                        current thread.
     - throws: An `NSError` if the Realm could not be initialized.
     - returns: An open Realm.
     */
```

While doing this I noticed that it's missing the queue parameter, which it probably should have.

Realm/ObjectServerTests/SwiftObjectServerTests.swift
```diff
@@ -2912,4 +2912,197 @@ class CombineObjectServerTests: SwiftSyncTestCase {
...
+        } else {
+            let user = try await app.login(credentials: .anonymous)
+            let realm = try await Realm(configuration: user.configuration(testName: #function),
+                                        downloadBeforeOpen: .once)
+            XCTAssertEqual(realm.objects(SwiftHugeSyncObject.self).count, 2)
+        }
+    }
+
+    func testAsyncOpenDownloadBehaviorNever() async throws {
+        // this test will test how the `once` behavior responds
```
tgoyne, line 2952: This isn't testing .once

Realm/ObjectServerTests/SwiftObjectServerTests.swift
```diff
@@ -2912,4 +2912,197 @@ class CombineObjectServerTests: SwiftSyncTestCase {
...
+        let realm = try await Realm()
+        XCTAssertEqual(realm.objects(SwiftPerson.self).count, 10)
+    }
+
+    func testAsyncOpenSync() async throws {
+        if isParent {
+            let email = "realm_tests_do_autoverify\(randomString(7))@\(randomString(7)).com"
+            let password = randomString(10)
+            try await app.emailPasswordAuth.registerUser(email: email, password: password)
+            let user = try await self.app.login(credentials: .emailPassword(email: email, password: password))
```
tgoyne, line 2935: These tests should all use the helper functions for doing this.

CHANGELOG.md
```diff
@@ -1,6 +1,8 @@
 x.y.z Release notes (yyyy-MM-dd)
 =============================================================
 ### Enhancements
+* Add `async` `Realm.asyncOpen` and `App.login` methods.
+* Add additional `observe` methods for Objects and RealmCollections which take a `PartialKeyPath` type key path parameter.
```
tgoyne, line 5: Bad changelog merge.

Realm.xcodeproj/project.pbxproj
```diff
@@ -2083,6 +2113,8 @@
 			);
 			dependencies = (
 				534DF4D025B86F3A00655AE2 /* PBXTargetDependency */,
+				5345F0A626E390D600827AC9 /* PBXTargetDependency */,
+				5345F0A826E390D800827AC9 /* PBXTargetDependency */,
```
tgoyne, line 2117: Bad merge

Realm/ObjectServerTests/SwiftObjectServerTests.swift
```diff
@@ -2912,4 +2912,188 @@ class CombineObjectServerTests: SwiftSyncTestCase {
...
+                                            downloadBeforeOpen: .once)
+                XCTAssertEqual(realm.objects(SwiftHugeSyncObject.self).count, 2)
+            case .second:
+                let user = try await app
+                    .login(credentials: .emailPassword(email: environment.email!,
+                                                       password: environment.password!))
+                XCTAssertTrue(Realm.fileExists(for: user.configuration(testName: #function)))
+                let realm = try await Realm(configuration: user.configuration(testName: #function),
+                                            downloadBeforeOpen: .once)
+                XCTAssertEqual(realm.objects(SwiftHugeSyncObject.self).count, 2)
```
tgoyne, line 3036: It looks like the only thing that changes between these two is the `XCTAssertTrue(Realm.fileExists(for: user.configuration(testName: #function)))` and duplicating the rest just makes it less clear.

Realm/ObjectServerTests/SwiftObjectServerTests.swift
```diff
@@ -2912,4 +2912,188 @@ class CombineObjectServerTests: SwiftSyncTestCase {
...
+                                            downloadBeforeOpen: .always)
+                XCTAssertEqual(realm.objects(SwiftHugeSyncObject.self).count, 2)
+            case .second:
+                let user = try await app
+                    .login(credentials: .emailPassword(email: environment.email!,
+                                                       password: environment.password!))
+                XCTAssertTrue(Realm.fileExists(for: user.configuration(testName: #function)))
+                let realm = try await Realm(configuration: user.configuration(testName: #function),
+                                            downloadBeforeOpen: .always)
+                XCTAssertEqual(realm.objects(SwiftHugeSyncObject.self).count, 4)
```
tgoyne, line 3092: Similarly the user at least could be pulled out here.

tgoyne (approved):

PR #7373, opened 2021-08-02
Update to core 11.2.0

tgoyne:


CHANGELOG.md
```diff
@@ -1,22 +1,36 @@
...
+       connection this will return an opened realm.
+* Add `EnvironmentValues.partitionValue`. This value can be injected into any view using one of
   our new property wrappers `AsyncOpen` and `AutoOpen`:
-  `MyView().environment(\.partitionValue, "partitionValue")`. 
+  `MyView().environment(\.partitionValue, "partitionValue")`.
+* Shift more of the work done when first initializing a collection notifier to the background worker thread rather than doing it on the main thread.
 
 ### Fixed
-* Fix `configuration(partitionValue: AnyBSON)` will set always a nil partition value
+* Fix `configuration(partitionValue: AnyBSON)` would always set a nil partition value
```
jsflax, line 16:
```suggestion
* Fix `configuration(partitionValue: AnyBSON)`– it would always set a nil partition value
```

Realm/RLMManagedDictionary.mm
```diff
@@ -368,29 +368,16 @@ - (void)removeObjectsForKeys:(NSArray *)keyArray {
...
-    }
-    catch (realm::KeyNotFound const&) {
-        return;
-    }
-    catch (...) {
-        throwError(nil, nil);
-    }
+    changeDictionary(self, ^{
+        RLMAccessorContext context(*_objectInfo);
+        _backingCollection.try_erase(context.unbox<realm::StringData>(key));
```
jsflax, line 379: What is the consequence of swallowing this error?

jsflax (approved):

Realm/RLMManagedDictionary.mm
tgoyne: Trying to remove a key from a dictionary which doesn't have that key isn't an error to begin with.

PR #7375, opened 2021-08-03
Add Projections

pavel-ship-it:
Projection is a effortless way to create a light weight view model out of Realm Object

Example of usage:
```
public class Person: Object {
    @Persisted var firstName = ""
    @Persisted var lastName = ""
    @Persisted var address: Address? = nil
    @Persisted var friends = List<Person>()
}

public class Address: EmbeddedObject {
    @Persisted var city: String = ""
    @Persisted var country = ""
}

struct PersonProjection: Projection {
    @Projected(\Person.firstName) var firstName
    @Projected(\Person.address.city) var homeCity
    @Projected(\Person.friends.projectTo.firstName) var firstFriendsName: ProjectedCollection<String>
}

let people: Results<PersonProjection> = realm.objects(PersonProjection.self)
```


RealmSwift/Projection.swift
```diff
@@ -0,0 +1,357 @@
...
+            ref[keyPath: projectedKeyPath as! WritableKeyPath<T, Value>] = $1
+        }
+    }
+}
+
+public protocol Projection: ThreadConfined, RealmCollectionValue {
+    associatedtype Root: ObjectBase
+
+    init()
+    func objectClassName() -> String
```
jsflax, line 99: What is the point of this method? Seems excessive.

RealmSwift/Projection.swift
```diff
@@ -0,0 +1,357 @@
...
+
+public enum ProjectionChange {
+    /**
+     If an error occurs, notification blocks are called one time with a `.error`
+     result and an `NSError` containing details about the error. Currently the
+     only errors which can occur are when opening the Realm on a background
+     worker thread to calculate the change set. The callback will never be
+     called again after `.error` is delivered.
+     */
+    case error(_ error: NSError)
```
jsflax, line 163: Why is this an `NSError` instead `Error`?

RealmSwift/Projection.swift
```diff
@@ -0,0 +1,357 @@
...
+    /**
+     One or more of the properties of the object have been changed.
+     */
+    case change(_: ObjectBase, _: [PropertyChange])
```
jsflax, line 167: Shouldn't this be the `Projection`?

RealmSwift/Projection.swift
```diff
@@ -0,0 +1,357 @@
...
+    }
+    // Must also conform to `AssistedObjectiveCBridgeable`
+    /**
+     The Realm which manages the object, or `nil` if the object is unmanaged.
+     Note: Projection can be instantiated for the managed objects only therefore realm will never be nil.
+     Unmanaged objects are not confined to a thread and cannot be passed to methods expecting a
+     `ThreadConfined` object.
+     */
+    public var realm: Realm? {
+        if let object = self.realmObject as? Object {
```
jsflax, line 192: This will fail for `EmbeddedObject`.

RealmSwift/Projection.swift
```diff
@@ -0,0 +1,357 @@
...
+            return object.realm
+        }
+        fatalError("Realm cannot be nil")
+    }
+
+    /// Indicates if the object can no longer be accessed because it is now invalid.
+    public var isInvalidated: Bool {
+        return false
```
jsflax, line 200: This should operate off the base object.

RealmSwift/Projection.swift
```diff
@@ -0,0 +1,357 @@
...
+    }
+    /**
+     Indicates if the object is frozen.
+     Frozen objects are not confined to their source thread. Forming a `ThreadSafeReference` to a
+     frozen object is allowed, but is unlikely to be useful.
+     */
+    public var isFrozen: Bool {
+        return false
```
jsflax, line 208: As should all the `ThreadConfined` methods.

jsflax (commented):
All public methods should be being tested on projections. Observation is also (obviously) missing.

RealmSwift/Projection.swift
pavel-ship-it: not required any more. Removed

RealmSwift/Projection.swift
pavel-ship-it: It must be a copy paste from `ObjectChange`

RealmSwift/Projection.swift
pavel-ship-it: Yep, `Projection` with the following `ProjectedChanges`

pavel-ship-it: It should be `Projection`

RealmSwift/Tests/ProjectionTests.swift
```diff
@@ -0,0 +1,1792 @@
...
+        "\(int) - \(bool)"
+    }
+}
+
+class AdvancedProjection: Projection<AdvancedObject> {
+    @Projected(\AdvancedObject.commonArray.count) var arrayLen
+    @Projected(\AdvancedObject.commonArray) var renamedArray
+    @Projected(\AdvancedObject.objectsArray.projectTo.stringify) var projectedArray: ProjectedList<String>
+    @Projected(\AdvancedObject.commonSet.first) var firstElement
+    @Projected(\AdvancedObject.objectsSet.projectTo.bool) var projectedSet: ProjectedMutableSet<Bool>
```
leemaguire, line 191: Is every possible type of advanced projection present in this class?

RealmSwift/Realm.swift
```diff
@@ -650,10 +650,23 @@ import Realm.Private
 
      - returns: A `Results` containing the objects.
      */
-    public func objects<Element: Object>(_ type: Element.Type) -> Results<Element> {
+    public func objects<Element: ObjectBase>(_ type: Element.Type) -> Results<Element> {
```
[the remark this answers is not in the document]
pavel-ship-it: it's not required. Just happened to be working. I'll revert this change.

RealmSwift/Tests/ProjectedMutableSetTests.swift
```diff
@@ -0,0 +1,300 @@
...
+        let chandedObject = realm.objects(CommonPerson.self).filter("firstName == 'Overwrite'").first
+        XCTAssertNotNil(chandedObject)
+    }
+
+    func testFreezeThawProjectedList() {
+        let realm = realmWithTestPath()
+        let johnSnow = realm.objects(PersonProjection.self).first!
+        let projectedList = collection.freeze()
+
+        XCTAssertTrue(projectedList.isFrozen)
```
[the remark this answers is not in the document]
pavel-ship-it: variable names are fixed. Yes, the test's logic remains same for MutableSet

RealmSwift/Tests/ProjectionTests.swift
pavel-ship-it: Yes

RealmSwift/Object.swift
```diff
@@ -646,21 +646,21 @@ internal func dynamicSet(object: ObjectBase, key: String, value: Any?) {
...
     }
 
-    internal var bridged: (objectiveCValue: Any, metadata: Any?) {
+    var bridged: (objectiveCValue: Any, metadata: Any?) {
         return (objectiveCValue: unsafeCastToRLMObject(), metadata: nil)
     }
 }
 
+// FIXME: Remove when `as! Self` can be written
+private func forceCastToInferred<T, V>(_ x: T) -> V {
```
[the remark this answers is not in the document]
pavel-ship-it: it's used un this file

RealmSwift/SwiftUI.swift
```diff
@@ -342,16 +381,29 @@ private class ObservableStorage<ObservedType>: ObservableObject where ObservedTy
 }
 
 // MARK: ObservedResults
+/**
+ A type which can be used with @ObservedResults propperty wrapper. Children class of Realm Object or Projection.
```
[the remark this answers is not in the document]
pavel-ship-it: Fixed

examples/ios/swift/Projections/ContentView.swift
```diff
@@ -0,0 +1,126 @@
...
+
+class SeriesModel: Projection<Series> {
+    @Projected(\Series.title) var title
+    @Projected(\Series.episodes.count) var epCount
+    @Projected(\Series.episodes.first?.title.quoted) var firstEpisode
+    @Projected(\Series.episodes) var episodes
+}
+
+extension String {
+    var quoted: String {
```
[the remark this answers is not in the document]
pavel-ship-it: it's for Projection example only. Now it's interpolated

RealmSwift/Object.swift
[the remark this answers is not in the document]
pavel-ship-it: There is a conflict with the recent change if AssistedObjectiveCBridgeable

RealmSwift/Object.swift
```diff
@@ -516,7 +516,7 @@ extension Object: ThreadConfined {
  Information about the changes made to an object which is passed to `Object`'s
  notification blocks.
  */
-@frozen public enum ObjectChange<T: ObjectBase> {
+@frozen public enum ObjectChange<T> {
```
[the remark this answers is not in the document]
tgoyne: We don't actually make any sort of ABI stability guarantees and breaking ABI compat is fine.

CHANGELOG.md
```diff
@@ -1,6 +1,33 @@
...
+    // There will be the only String for `city` of the original object `Address`
+    @Projected(\Person.address.city) var homeCity 
+    // List<Person> will be mapped to list of firstNames
+    @Projected(\Person.friends.projectTo.firstName) var firstFriendsName: ProjectedList<String>
+}
+
+// `people` will contain projections for every `Person` object in the `realm`
+let people: Results<PersonProjection> = realm.objects(PersonProjection.self)
+```
+* Add `.searchable()` SwiftUI View Modifier which allows us to filter
```
tgoyne, line 30: Bad changelog merge here.

RealmSwift/Combine.swift
```diff
@@ -57,6 +57,18 @@ extension ObjectKeyIdentifiable {
...
 
+@available(macOS 10.15, iOS 13, tvOS 13, watchOS 6, *)
+extension Projection: Identifiable, Hashable {
+    /// A stable identifier for this projection. This value will be
+    /// the same for all projections of this object
+    public var id: UInt64 {
+        // UInt64(bitPattern:) to properly cast negative Int64 values
+        // It will use the same memory representation, the numeric value
+        // may be different which is fine in this case
+        UInt64(bitPattern: Int64(ObjectIdentifier(type(of: self)).hashValue))
```
tgoyne, line 68: I don't think this is valid. `hashValue` is allowed to have collisions, but `id` is not. `return withUnsafePointer(to: self) { UInt(bitPattern: $0) }` looks to be the correct way to do this.

RealmSwift/Combine.swift
```diff
@@ -57,6 +57,18 @@ extension ObjectKeyIdentifiable {
     }
 }
 
+@available(macOS 10.15, iOS 13, tvOS 13, watchOS 6, *)
+extension Projection: Identifiable, Hashable {
```
tgoyne, line 61: Unconditionally conforming to Identifiable means that users can't define their own `id` property. For normal objects we make this opt-in with ObjectKeyIdentifiable.

RealmSwift/Combine.swift
```diff
@@ -57,6 +57,18 @@ extension ObjectKeyIdentifiable {
...
+    /// A stable identifier for this projection. This value will be
+    /// the same for all projections of this object
```
tgoyne, line 63: This doesn't seem true? Two different projections of the same object will have different ObjectIdentifiers. Should the  `id` be derived from the object key?

RealmSwift/Results.swift
```diff
@@ -164,14 +176,27 @@ extension AnyRealmValue: AddableType {}
      */
     public subscript(position: Int) -> Element {
         throwForNegativeIndex(position)
+        if let projector = projector {
```
tgoyne, line 179: This seems like something that could just be done in `dynamicBridgeCast` without having the projector at all. `Projection` needs to conform to a protocol that allows constructing it from an ObjectBase, and then `dynamicBridgeCast()` does `if let object = x as? ObjectBase, type = T.self as? ProjectionProtocol.Type { return type(x) as! T }`. This would make it so that none of the getter functions need changes.

RealmSwift/SwiftUI.swift
```diff
@@ -176,10 +176,24 @@ private final class ObservableStoragePublisher<ObjectType>: Publisher where Obje
     private var subscribers = [AnySubscriber<Void, Never>]()
     private let value: ObjectType
     private let keyPaths: [String]?
+    private let unwrappedValue: () -> (ObjectBase?)
```
tgoyne, line 179: Why is this a closure and not just a value?

RealmSwift/SwiftUI.swift
```diff
@@ -211,27 +225,43 @@ private final class ObservableStoragePublisher<ObjectType>: Publisher where Obje
...
     }
 
+    let instance: () -> (ObservableStoragePublisher<ObservedType>)
     var objectWillChange: ObservableStoragePublisher<ObservedType>
     var keyPaths: [String]?
 
     init(_ value: ObservedType, _ keyPaths: [String]? = nil) {
         self.value = value.realm != nil && !value.isInvalidated ? value.thaw() ?? value : value
-        self.objectWillChange = ObservableStoragePublisher(value, keyPaths)
+        self.instance = { ObservableStoragePublisher(value, keyPaths) }
```
tgoyne, line 246: Why is this a closure?

RealmSwift/SwiftUI.swift
```diff
@@ -342,16 +381,30 @@ private class ObservableStorage<ObservedType>: ObservableObject where ObservedTy
...
 ///
 /// The results use the realm configuration provided by
 /// the environment value `EnvironmentValues/realmConfiguration`.
 @available(iOS 13.0, macOS 10.15, tvOS 13.0, watchOS 6.0, *)
-@propertyWrapper public struct ObservedResults<ResultType>: DynamicProperty, BoundCollection where ResultType: Object & Identifiable {
+@propertyWrapper public struct ObservedResults<ResultType>: DynamicProperty, BoundCollection where ResultType: _ObservedResultsValue {
     private class Storage: ObservableStorage<Results<ResultType>> {
         var setupHasRun = false
-
+        var resultFactory: ((Realm) -> Results<ResultType>)!
```
tgoyne, line 407: This appears to be set to the same thing in every place it's used.

RealmSwift/Projection.swift
```diff
@@ -0,0 +1,1342 @@
...
+        storage storageKeyPath: ReferenceWritableKeyPath<EnclosingSelf, Self>
+        ) -> Value {
+        get {
+            let storage = observed[keyPath: storageKeyPath]
+            return observed.rootObject[keyPath: storage._projectedKeyPath]
+        }
+        set {
+            precondition(observed[keyPath: storageKeyPath].projectedKeyPath is WritableKeyPath<T, Value>,
+                         "KeyPath is not writable")
+            observed.rootObject[keyPath: observed[keyPath: storageKeyPath].projectedKeyPath as! WritableKeyPath<T, Value>] = newValue
```
tgoyne, line 84: Rather than repeating the lookup and cast this should be a `guard let ... else { preconditionFailure() }`

RealmSwift/Projection.swift
```diff
@@ -0,0 +1,1342 @@
...
+/// You can retrieve all Projections of a given type from a Realm by calling the `objects(_:)` of Realm or `init(projecting:)`
+/// of Projection's class:
+/// ```swift
+/// let projections = realm.object(PersonProjection.self)
+/// let personObject = realm.create(Person.self)
+/// let singleProjection = PersonProjection(projecting: personObject)
+/// ```
+open class Projection<Root: ObjectBase>: RealmCollectionValue, ProjectionObservable {
+    /// The object being projected
+    public var rootObject: Root
```
tgoyne, line 207: This should be `let` right?

RealmSwift/Projection.swift
```diff
@@ -0,0 +1,1342 @@
...
+}
+
+// MARK: Projection Schema
+private struct ProjectedMetadata {
+    let projectedKeyPath: AnyKeyPath
+    let originPropertyKeyPathString: String
+    let label: String
+}
+
+private var schema = [ObjectIdentifier: [ProjectedMetadata]]()
```
tgoyne, line 101: The code touching this needs locking. Currently initializing projections on multiple threads at once will explode.

RealmSwift/Projection.swift
```diff
@@ -0,0 +1,1342 @@
...
+    public func thaw() -> Self? {
+        if let thawedObject = rootObject.thaw() {
+            return Self(projecting: thawedObject)
+        }
+        return nil
+    }
+}
+
+// - MARK: _RealmSchemaDiscoverable
+extension Projection {
```
tgoyne, line 539: This looks unnecessary?

RealmSwift/Projection.swift
```diff
@@ -0,0 +1,1342 @@
...
+        guard let backingList = backingList.thaw() else {
+            return nil
+        }
+        return anyCtor(backingList)
+    }
+
+    private var backingList: List<Object>
+    private let keyPath: AnyKeyPath
+    private let propertyName: String
+    private let anyCtor: (List<Object>) -> Self
```
tgoyne, line 885: There doesn't appear to be any reason for this to be a closure.

RealmSwift/Projection.swift
```diff
@@ -0,0 +1,1342 @@
...
+}
+#endif // !(os(iOS) && (arch(i386) || arch(arm)))
+
+#endif // canImport(Combine)
+
+// MARK: Projected Mutable Set
+
+/// ProjectedMutableSet is a special type of collection for Projection's properties
+/// You don't need to instantiate this type manually.
+public struct ProjectedMutableSet<NewElement>: RandomAccessCollection where NewElement: RealmCollectionValue {
```
tgoyne, line 1044: Is this actually a different type from ProjectedList? At first glance it appears that both could be ProjectedCollection.

RealmSwift/Projection.swift
pavel-ship-it: Conformance to `RealmCollectionValue` needs conformance to `_RealmSchemaDiscoverable`

RealmSwift/Projection.swift
pavel-ship-it: it's a proper init now

RealmSwift/Projection.swift
pavel-ship-it: var, but with private setter to access `observed.rootObject[keyPath: *WritableKeyPath*]`

RealmSwift/SwiftUI.swift
pavel-ship-it: it's a value now

RealmSwift/SwiftUI.swift
pavel-ship-it: both places has specific definition of ResultType which can be lost outside of init method

tgoyne: The captured ResultType is always the same as the ResultType generic parameter.

RealmSwift/Projection.swift
pavel-ship-it: Interface similar, but isn't matching, like other our collections

RealmSwift/SwiftUI.swift
pavel-ship-it: `realm.object(type)` will be
`objects<Element: Object>(_ type: Element.Type) -> Results<Element>` or
`objects<Root, T: Projection<Root>>(_ type: T.Type) -> Results<T> where Root: ThreadConfined`

RealmSwift/SwiftUI.swift
pavel-ship-it: it should be a value

RealmSwift/SwiftUI.swift
pavel-ship-it: Added `RealmFetchable` protocol to eliminate uncertainty. Looking for a better name

Realm.xcodeproj/xcshareddata/xcschemes/RealmSwift.xcscheme
```diff
@@ -64,6 +64,14 @@
                BlueprintName = "RealmSwift Tests"
                ReferencedContainer = "container:Realm.xcodeproj">
             </BuildableReference>
+            <SkippedTests>
+               <Test
+                  Identifier = "ProjectedCollectTestsTemplate">
```
tgoyne, line 69: Should these actually be skipped here? If these are base classes that shouldn't be run then this won't work for spm.

RealmSwift/Combine.swift
```diff
@@ -57,6 +57,18 @@ extension ObjectKeyIdentifiable {
...
 
+/// :nodoc:
+@available(macOS 10.15, iOS 13, tvOS 13, watchOS 6, *)
+extension ObjectKeyIdentifiable where Self: ProjectionObservable {
+    /// A stable identifier for this projection.
+    public var id: UInt64 {
+        // UInt64(bitPattern:) to properly cast negative Int64 values
+        // It will use the same memory representation, the numeric value
+        // may be different which is fine in this case
+        UInt64(withUnsafePointer(to: self, { UInt(bitPattern: $0) }))
```
tgoyne, line 68: This is not a valid conformance to `Identifiable`. It needs to use the object key from the projected object.

RealmSwift/Projection.swift
```diff
@@ -0,0 +1,642 @@
...
+///     @Projected(\Person.address.city) var homeCity
+///     @Projected(\Person.friends.projectTo.firstName) var firstFriendsName: ProjectedCollection<String>
+/// }
+///
+/// let people: Results<PersonProjection> = realm.objects(PersonProjection.self)
+/// ```
+@propertyWrapper
+public struct Projected<T: ObjectBase, Value>: AnyProjected {
+
+    fileprivate var _projectedKeyPath: KeyPath<T, Value>!
```
tgoyne, line 56: This doesn't appear to need to be a optional.

RealmSwift/Projection.swift
```diff
@@ -0,0 +1,642 @@
...
+        let schema = projection._schema
+        switch objectChange {
+        case .error(let error):
+            return .error(error)
+        case .change(let object, let objectPropertyChanges):
+            let newProjection = T(projecting: object)
+            let projectedPropertyChanges: [PropertyChange] = objectPropertyChanges.map { propChange in
+                // read the metadata for the property whose origin name matches
+                // the changed property's name
+                let lock = NSLock()
```
tgoyne, line 131: This lock doesn't do anything. Each call to this function is creating a new local lock that isn't shared with other simultaneous calls. There needs to be a single global lock which is used by everything which touches the schema global.

RealmSwift/Projection.swift
```diff
@@ -0,0 +1,642 @@
...
+                lock.lock()
+                let propertyMetadata = schema.first(where: {
+                    $0.originPropertyKeyPathString == propChange.name
+                })!
+                lock.unlock()
+                var change: (name: String?, oldValue: Any?, newValue: Any?) = (nil, nil, nil)
```
tgoyne, line 137: This is never actually used as a tuple and should probably just be three local variables?

RealmSwift/Projection.swift
```diff
@@ -0,0 +1,642 @@
...
+                if let oldValue = propChange.oldValue {
+                    // if there is an oldValue in the change, construct an empty Root
+                    let newRoot = T.Root()
```
tgoyne, line 140: Why is this constructing a new Root rather than using the one passed to this function?

RealmSwift/Tests/ProjectionTests.swift
```diff
@@ -0,0 +1,1792 @@
...
+    @Projected(\CommonPerson.address?.city) var homeCity
+    @Projected(\CommonPerson.friends.projectTo.firstName) var firstFriendsName: ProjectedCollection<String>
+}
+
+public class SimpleObject: Object {
+    @Persisted var int: Int
+    @Persisted var bool: Bool
+}
+
+@available(OSX 10.15, watchOS 6.0, iOS 13.0, iOSApplicationExtension 13.0, OSXApplicationExtension 10.15, tvOS 13.0, *)
```
tgoyne, line 227: What about these require a higher deployment target?

RealmSwift/Tests/ProjectionTests.swift
```diff
@@ -0,0 +1,1792 @@
...
+////////////////////////////////////////////////////////////////////////////
+
+import Foundation
+import Realm
+import RealmSwift
+import XCTest
+
+#if canImport(RealmTestSupport)
+import RealmTestSupport
+import SwiftUI
```
tgoyne, line 26: There doesn't appear to be anything using SwiftUI in this test.

Realm.xcodeproj/xcshareddata/xcschemes/RealmSwift.xcscheme
pavel-ship-it: fixed

RealmSwift/Projection.swift
pavel-ship-it: fixed

RealmSwift/Projection.swift
pavel-ship-it: fixed

RealmSwift/Projection.swift
pavel-ship-it: fixed

RealmSwift/Tests/ProjectionTests.swift
pavel-ship-it: fixed

RealmSwift/Projection.swift
pavel-ship-it: it'll be mutated to collect the old value.

RealmSwift/Tests/ProjectionTests.swift
pavel-ship-it: not required. fixed.

RealmSwift/Combine.swift
pavel-ship-it: fixed

Realm.xcodeproj/xcshareddata/xcschemes/RealmSwift.xcscheme
tgoyne: This test case appears to not just not exist at all, so this doesn't need to be here.

RealmSwift/Projection.swift
```diff
@@ -0,0 +1,643 @@
...
+                    projectionSchemaLock.unlock()
+                    throwRealmException("@Projected property '\(child.label!)' must be a part of Realm object")
+                }
+                return ProjectedMetadata(projectedKeyPath: projected.projectedKeyPath,
+                                         originPropertyKeyPathString: originPropertyLabel,
+                                         label: child.label!)
+            }
+            schema[ObjectIdentifier(type(of: self))] = metadatas
+        }
+        return schema[ObjectIdentifier(type(of: self))]!
```
tgoyne, line 457:
```suggestion
        let identifier = ObjectIdentifier(type(of: self))
        if let schema = schema[identifier] {
            return schema
        }

        let mirror = Mirror(reflecting: self)
        let metadatas: [ProjectedMetadata] = mirror.children.compactMap { child in
            guard let projected = child.value as? AnyProjected else {
                return nil
            }
            let originPropertyLabel = _name(for: projected.projectedKeyPath as! PartialKeyPath<Root>)
            guard !originPropertyLabel.isEmpty else {
                projectionSchemaLock.unlock()
                throwRealmException("@Projected property '\(child.label!)' must be a part of Realm object")
            }
            return ProjectedMetadata(projectedKeyPath: projected.projectedKeyPath,
                                     originPropertyKeyPathString: originPropertyLabel,
                                     label: child.label!)
        }
        schema[identifier] = metadatas
        return metadatas
```

RealmSwift/RealmCollection.swift
```diff
@@ -1946,3 +1946,428 @@ internal protocol PropertyNameConvertible {
...
+
+ class ListProjection: Projection<PersistedListObject> {
+     @Projected(\PersistedListObject.people.projectTo.firstName) var strings: ProjectedCollection<String>
+ }
+ ```
+*/
+public struct ProjectedCollection<NewElement>: RandomAccessCollection, ThreadConfined where NewElement: RealmCollectionValue {
+    public typealias Element = NewElement
+    public typealias Index = Int
+/**
```
tgoyne, line 1967: Missing indentation.

RealmSwift/RealmCollection.swift
```diff
@@ -1946,3 +1946,428 @@ internal protocol PropertyNameConvertible {
...
+    }
+    /// A human-readable description of the object.
+    public var description: String {
+        var elements = ""
+        for (i, o) in self.enumerated() {
+            elements += "\t[\(i)] \(o)\n"
+        }
+        return "\(type(of: self))<\(Element.self)> <\(self)> {\n" +
+        elements +
+        "}"
```
tgoyne, line 2239:
```suggestion
        return "\(type(of: self))<\(Element.self)> <\(self)> {\n\(elements)}"
```

tgoyne (approved):
A few very minor comments but otherwise LGTM.

PR #7379, opened 2021-08-04
Upgrade APIs for `async/await`

dianaafanador3:
Update some APIs for `async await` 

- [x] Fix documentation
- [x] Changelog
- [x] Pull request https://github.com/realm/realm-cocoa/pull/7367 to be merged


chrsomle (approved):

tgoyne:
The change to copy the stitch dylib to the bundle breaks the SPM build, and I'm not sure why it's being done. The object server tests on master work for me on Monterey.

tgoyne:
The code changes and tests look fine to me, so this just needs to have conflicts resolved and the build fixed.

Realm/ObjectServerTests/SwiftMongoClientTests.swift
```diff
@@ -0,0 +1,1222 @@
...
+import RealmSwift
+import XCTest
+
+#if canImport(RealmTestSupport)
+import RealmSwiftSyncTestSupport
+import RealmSyncTestSupport
+import RealmTestSupport
+#endif
+
+@objc(SwiftMongoClientTests)
```
tgoyne, line 30: Giving test suites explicit objc names is only needed for suites which use child processes.

Realm/ObjectServerTests/SwiftObjectServerTests.swift
```diff
@@ -3079,6 +2223,124 @@ class AsyncAwaitObjectServerTests: SwiftSyncTestCase {
...
+    func testCallResetPasswordAsyncAwait() async throws {
+        let email = "realm_tests_do_autoverify\(randomString(7))@\(randomString(7)).com"
+        let password = randomString(10)
+        try await app.emailPasswordAuth.registerUser(email: email, password: password)
+        do {
+            try await app.emailPasswordAuth.callResetPasswordFunction(email: email,
+                                                                      password: randomString(10),
+                                                                      args: [[:]])
+        } catch {
+            XCTAssertNotNil(error)
```
tgoyne, line 2236: I'm not sure what this test is trying to check. It appears to pass regardless of if callResetPasswordFunction succeeds or fails as long as the failure isn't a nil error, which seems to be a weird thing to be testing for?

Realm/ObjectServerTests/SwiftObjectServerTests.swift
```diff
@@ -3079,6 +2223,124 @@ class AsyncAwaitObjectServerTests: SwiftSyncTestCase {
...
+
+    func testEmailPasswordProviderClientAsyncAwait() async throws {
+        let email = "realm_tests_do_autoverify\(randomString(7))@\(randomString(7)).com"
+        let password = randomString(10)
+        try await app.emailPasswordAuth.registerUser(email: email, password: password)
+
+        do {
+            try await app.emailPasswordAuth.confirmUser("atoken", tokenId: "atokenid")
+        } catch {
+            XCTAssertNotNil(error)
```
tgoyne, line 2277: Are these trying to verify that the call does throw? If so there needs to be a XCTFail() in the do block after the thing that should fail.

Realm/ObjectServerTests/SwiftObjectServerTests.swift
```diff
@@ -3079,6 +2223,124 @@ class AsyncAwaitObjectServerTests: SwiftSyncTestCase {
...
+        try await app.emailPasswordAuth.registerUser(email: email, password: password)
+
+        let syncUser = try await self.app.login(credentials: Credentials.anonymous)
+        XCTAssertNotNil(syncUser)
+
+        let userDataEx = expectation(description: "Update user data")
+        app.currentUser?.functions.updateUserData([["favourite_colour": "green", "apples": 10]]) { _, error  in
+            XCTAssertNil(error)
+            userDataEx.fulfill()
+        }
```
tgoyne, line 2337: Shouldn't this have an async version too?

Realm/ObjectServerTests/SwiftMongoClientTests.swift
```diff
@@ -0,0 +1,1222 @@
...
+class SwiftMongoClientTests: SwiftSyncTestCase {
+    override func tearDown() {
+        _ = setupMongoCollection()
+        super.tearDown()
+    }
+    func testMongoClient() {
+        let user = try! logInUser(for: Credentials.anonymous)
```
jsflax, line 37: Can just do .anonymous.

Realm/ObjectServerTests/SwiftMongoClientTests.swift
```diff
@@ -0,0 +1,1222 @@
...
+#if swift(>=5.5)
+@available(macOS 12.0, *)
+class AsyncAwaitMongoClientTests: SwiftSyncTestCase {
+    func setupMongoCollection() async throws -> MongoCollection {
+        let email = "realm_tests_do_autoverify\(randomString(7))@\(randomString(7)).com"
+        let password = randomString(10)
+        try await app.emailPasswordAuth.registerUser(email: email, password: password)
+
+        let credentials = Credentials.emailPassword(email: email, password: password)
+        let user = try await app.login(credentials: credentials)
```
jsflax, line 881: This should use our convenience method to create and login a user, which will eliminate the above 6 lines of code.

RealmSwift/MongoClient.swift
```diff
@@ -703,6 +703,240 @@ extension MongoCollection {
...
+    ///   - update: A bson `Document` representing the update to be applied to a matching document.
+    ///   - upsert: When true, creates a new document if no document matches the query.
+    /// - Returns: `UpdateResult`result of the `updateOne` operation.
+    public func updateOneDocument(filter: Document,
+                                  update: Document,
+                                  upsert: Bool? = nil) async throws -> UpdateResult {
+        return try await withCheckedThrowingContinuation { continuation in
+            if let upsert = upsert {
+                updateOneDocument(filter: filter,
+                                       update: update,
```
jsflax, line 824: Bad indents.

RealmSwift/MongoClient.swift
```diff
@@ -703,6 +703,240 @@ extension MongoCollection {
...
+    ///   - update: A bson `Document` representing the update to be applied to a matching document.
+    ///   - upsert: When true, creates a new document if no document matches the query.
+    /// - Returns:`UpdateResult`result of the `updateMany` operation.
+    public func updateManyDocuments(filter: Document,
+                                    update: Document,
+                                    upsert: Bool? = nil) async throws -> UpdateResult {
+        return try await withCheckedThrowingContinuation { continuation in
+            if let upsert = upsert {
+                updateManyDocuments(filter: filter,
+                                         update: update,
```
jsflax, line 847: Bad indents.

RealmSwift/MongoClient.swift
```diff
@@ -703,6 +703,240 @@ extension MongoCollection {
...
+    ///   - update: A bson `Document` representing the update to be applied to a matching document.
+    ///   - options: `RemoteFindOneAndModifyOptions` to use when executing the command.
+    /// - Returns: `Document` result of the attempt to update a document  or `nil` if document wasn't found.
+    public func findOneAndUpdate(filter: Document,
+                                 update: Document,
+                                 options: FindOneAndModifyOptions? = nil) async throws -> Document? {
+        return try await withCheckedThrowingContinuation { continuation in
+            if let options = options {
+                findOneAndUpdate(filter: filter,
+                                      update: update,
```
jsflax, line 875: Bad indents.

RealmSwift/MongoClient.swift
```diff
@@ -703,6 +703,240 @@ extension MongoCollection {
...
+    ///   - replacement: A `Document` describing the replacement.
+    ///   - options: `FindOneAndModifyOptions` to use when executing the command.
+    /// - Returns: `Document`result of the attempt to reaplce a document   or `nil` if document wasn't found.
+    public func findOneAndReplace(filter: Document,
+                                  replacement: Document,
+                                  options: FindOneAndModifyOptions? = nil) async throws -> Document? {
+        return try await withCheckedThrowingContinuation { continuation in
+            if let options = options {
+                findOneAndReplace(filter: filter,
+                                       replacement: replacement,
```
jsflax, line 903: Bad indents.

RealmSwift/MongoClient.swift
```diff
@@ -703,6 +703,240 @@ extension MongoCollection {
...
+    /// - Parameters:
+    ///   - filter: A `Document` that should match the query.
+    ///   - options: `FindOneAndModifyOptions` to use when executing the command.
+    /// - Returns: `Document` result of the attempt to delete a document  or `nil` if document wasn't found.
+    public func findOneAndDelete(filter: Document,
+                                 options: FindOneAndModifyOptions? = nil) async throws -> Document? {
+        return try await withCheckedThrowingContinuation { continuation in
+            if let options = options {
+                findOneAndDelete(filter: filter,
+                                      options: options,
```
jsflax, line 929: Bad indents.

Realm/ObjectServerTests/SwiftMongoClientTests.swift
```diff
@@ -0,0 +1,1222 @@
...
+                     reason: "Cannot set 'deleteRealmIfMigrationNeeded' when sync is enabled ('syncConfig' is set).")
+
+        var localConfiguration = Realm.Configuration.defaultConfiguration
+        assertSucceeds {
+            localConfiguration.deleteRealmIfMigrationNeeded = true
+        }
+    }
+}
+
+#if swift(>=5.5)
```
jsflax, line 872: Should add a MARK here.

Realm/ObjectServerTests/SwiftMongoClientTests.swift
```diff
@@ -0,0 +1,1222 @@
...
+
+        let document: Document = ["name": "tomas", "breed": "jack rusell"]
+        let objectId = try await collection.insertOne(document)
+        XCTAssertNotNil(objectId)
+        let fetchedDocument = try await collection.find(filter: document)
+        XCTAssertEqual(fetchedDocument[0]["name"]??.stringValue, "tomas")
+
+        let document1: Document = ["name": "lucas", "breed": "german shepard"]
+        let document2: Document = ["name": "fito", "breed": "goberian"]
+        let objectIds = try await collection.insertMany([document1, document2])
```
jsflax, line 901: Testing insertMany should be another test.

Realm/ObjectServerTests/SwiftMongoClientTests.swift
```diff
@@ -0,0 +1,1222 @@
...
+        XCTAssertEqual(objectIds.count, 4)
+
+        // Test findOne all
+        let fetchedOneDocument = try await collection.findOneDocument(filter: document)
+        XCTAssertNotNil(fetchedOneDocument)
+        XCTAssertEqual(fetchedOneDocument?["name"]??.stringValue, "tomas")
+        XCTAssertEqual(fetchedOneDocument?["breed"]??.stringValue, "jack rusell")
+
+        // Test findOne all with option limit to one
+        let findOptions = FindOptions(2, nil, nil)
```
jsflax, line 962: Option limit is not one here.

Realm/ObjectServerTests/SwiftMongoClientTests.swift
```diff
@@ -0,0 +1,1222 @@
...
+        let document: Document = ["name": "tomas", "breed": "jack rusell"]
+        let document1: Document = ["name": "lucas", "breed": "german shepard"]
+        let document2: Document = ["name": "fito", "breed": "goberian"]
+        let document3: Document = ["name": "fosca", "breed": "labradoodle"]
+
+        let objectIds = try await collection.insertMany([document, document1, document2, document3])
+        XCTAssertNotNil(objectIds)
+        XCTAssertEqual(objectIds.count, 4)
+
+        // Returning a document means that it was founded and deleted
```
jsflax, line 1041:
```suggestion
        // Returning a document means that it was found and deleted
```

Realm/ObjectServerTests/SwiftMongoClientTests.swift
```diff
@@ -0,0 +1,1222 @@
...
+        let deletedDocument = try await collection.findOneAndDelete(filter: document)
+        XCTAssertNotNil(deletedDocument)
+        XCTAssertEqual(deletedDocument?["name"]??.stringValue, "tomas") // shouldReturnNewDocument is false that is why returns old document
```
jsflax, line 1044: What does this comment mean?

Realm/ObjectServerTests/SwiftMongoClientTests.swift
```diff
@@ -0,0 +1,1222 @@
...
+        XCTAssertEqual(updatedResult2.modifiedCount, 0)
+
+        let updatedResult3 = try await collection.updateOneDocument(filter: document5,
+                                                                    update: document2)
+        XCTAssertNil(updatedResult3.objectId) // No upsert
+        XCTAssertEqual(updatedResult3.matchedCount, 0)
+        XCTAssertEqual(updatedResult3.modifiedCount, 0)
+    }
+
+    func testMongoCollectionUpdateManyAsyncAwait() async throws {
```
jsflax, line 1095: I think this test is doing too much? Seems like the content can be cut in half and handled with one or two updates instead of four.

Realm/ObjectServerTests/SwiftMongoClientTests.swift
```diff
@@ -0,0 +1,1222 @@
...
+        let document3: Document = ["name": "balo", "breed": "pug"]
+
+        let deletedDocumentCount = try await collection.deleteManyDocuments(filter: document)
+        XCTAssertEqual(deletedDocumentCount, 0)
+
+        let objectIds = try await collection.insertMany([document, document1, document2, document3])
+        XCTAssertNotNil(objectIds)
+        XCTAssertEqual(objectIds.count, 4)
+
+        let deletedDocumentCount2 = try await collection.deleteManyDocuments(filter: document)
```
jsflax, line 1171: Should probably test that this delete more than one document.

Realm/ObjectServerTests/SwiftMongoClientTests.swift
```diff
@@ -0,0 +1,1222 @@
...
+        XCTAssertEqual(count, 4)
+
+        let count2 = try await collection.count(filter: document)
+        XCTAssertEqual(count2, 1)
+
+        let count3 = try await collection.count(filter: [:], limit: 2)
+        XCTAssertEqual(count3, 2)
+
+        let count4 = try await collection.count(filter: document3, limit: 1)
+        XCTAssertEqual(count4, 1)
```
jsflax, line 1217: I'm not sure this adds anything to the test.

Realm/ObjectServerTests/SwiftObjectServerTests.swift
dianaafanador3: As mentioned on the internal slack channel testing this casuistic will need to implement handler that intercepts the email from the server and it is a lot of work, because of this we are testing at least for this to fail. I'm adding a  XCTFail() in case of success.

Realm/ObjectServerTests/SwiftObjectServerTests.swift
dianaafanador3: looks like subscripts cannot be `async` as mentioned in the proposal https://github.com/apple/swift-evolution/blob/main/proposals/0296-async-await.md#:~:text=for%20such%20references.-,Special,-functions%20like%20deinit, maybe they will allow it in the future for subscripts that only have a getter which is the case here

Realm/ObjectServerTests/SwiftObjectServerTests.swift
dianaafanador3: same as the above comment

Realm/ObjectServerTests/SwiftMongoClientTests.swift
dianaafanador3: comment deleted

Realm/ObjectServerTests/SwiftObjectServerTests.swift
jsflax: There is a clever, non-breaking way to make this `async` that needs to be accomplished in this PR. Since our Functions struct only returned pseudo completionHandler continuations, we can overload another structure that leverages `dynamicCallable` that can then be async:

```swift
// end result
func testCallFunctionAsync() async throws {
    let user = try await app.login(credentials: basicCredentials())

    guard case let .int32(sum) = try await user.functions.sum([1, 2, 3, 4, 5]) else {
        return XCTFail("Should be int32")
    }
    XCTAssertEqual(sum, 15)

    guard case let .bool(upd) = try await user.functions.updateUserData([
        ["favourite_colour": "green", "apples": 10]
    ]) else {
        return XCTFail("Should be bool")
    }
    XCTAssertTrue(upd)
}


// implementation
@dynamicMemberLookup
@frozen public struct Functions {

    private let user: User

    fileprivate init(user: User) {
        self.user = user
    }

    /// A closure type for receiving the completion of a remote function call.
    public typealias FunctionCompletionHandler = (AnyBSON?, Error?) -> Void

    /// A closure type for the dynamic remote function type.
    public typealias Function = ([AnyBSON], @escaping FunctionCompletionHandler) -> Void

    /// The implementation of @dynamicMemberLookup that allows for dynamic remote function calls.
    public subscript(dynamicMember string: String) -> Function {
        return { (arguments: [AnyBSON], completionHandler: @escaping FunctionCompletionHandler) in
            let objcArgs = arguments.map(ObjectiveCSupport.convertBson)
            self.user.__callFunctionNamed(string, arguments: objcArgs) { (bson: RLMBSON?, error: Error?) in
                completionHandler(bson.map(ObjectiveCSupport.convertBson) ?? .none, error)
            }
        }
    }

    /// The implementation of @dynamicMemberLookup that allows for dynamic remote function calls.
    public subscript(dynamicMember string: String) -> FunctionCallable {
        FunctionCallable(name: string, user: user)
    }
}

@dynamicCallable
public struct FunctionCallable {
    fileprivate let name: String
    fileprivate let user: User

    public func dynamicallyCall(withArguments args: [([AnyBSON], ((Result<AnyBSON, Error>) -> Void))]) -> Void {
        let objcArgs = args.first!.0.map(ObjectiveCSupport.convertBson)
        self.user.__callFunctionNamed(name, arguments: objcArgs) { (bson: RLMBSON?, error: Error?) in
            if let b = bson.map(ObjectiveCSupport.convertBson), let bson = b {
                args.first!.1(.success(bson))
            } else {
                args.first!.1(.failure(error ?? Realm.Error.callFailed))
            }
        }
    }

    #if !(os(iOS) && (arch(i386) || arch(arm)))
    /// The implementation of @dynamicMemberLookup that allows for dynamic remote function calls.
    @available(OSX 10.15, watchOS 6.0, iOS 13.0, iOSApplicationExtension 13.0, OSXApplicationExtension 10.15, tvOS 13.0, macCatalyst 13.0, macCatalystApplicationExtension 13.0, *)
    public func dynamicallyCall(withArguments args: [[AnyBSON]]) -> Future<AnyBSON, Error> {
        return Future<AnyBSON, Error> { promise in
            let objcArgs = args.first!.map(ObjectiveCSupport.convertBson)
            self.user.__callFunctionNamed(name, arguments: objcArgs) { (bson: RLMBSON?, error: Error?) in
                if let b = bson.map(ObjectiveCSupport.convertBson), let bson = b {
                    promise(.success(bson))
                } else {
                    promise(.failure(error ?? Realm.Error.callFailed))
                }
            }
        }
    }
    #endif
}

@available(macOS 12.0, tvOS 15.0, iOS 15.0, watchOS 8.0, *)
extension FunctionCallable {
    public func dynamicallyCall(withArguments args: [[AnyBSON]]) async throws -> AnyBSON {
        try await withCheckedThrowingContinuation { continuation in
            let objcArgs = args.first!.map(ObjectiveCSupport.convertBson)
            self.user.__callFunctionNamed(name, arguments: objcArgs) { (bson: RLMBSON?, error: Error?) in
                if let b = bson.map(ObjectiveCSupport.convertBson), let bson = b {
                    continuation.resume(returning: bson)
                } else {
                    continuation.resume(throwing: error ?? Realm.Error.callFailed)
                }
            }
        }
    }
}
```

Note that this code has some unnecessary repetition that could use cleaning up, but the rest stands.

Realm/ObjectServerTests/SwiftMongoClientTests.swift
dianaafanador3: changed this to test limit option on deleteManyDocuments function

Realm/ObjectServerTests/SwiftObjectServerTests.swift
```diff
@@ -3124,6 +2287,131 @@ class AsyncAwaitObjectServerTests: SwiftSyncTestCase {
...
     }
+
+    func testCallResetPasswordAsyncAwait() async throws {
+        let email = "realm_tests_do_autoverify\(randomString(7))@\(randomString(7)).com"
+        let password = randomString(10)
+        try await app.emailPasswordAuth.registerUser(email: email, password: password)
+        do {
+            try await app.emailPasswordAuth.callResetPasswordFunction(email: email,
+                                                                      password: randomString(10),
+                                                                      args: [[:]])
```
jsflax, line 2298: Extra new line

Realm/ObjectServerTests/SwiftObjectServerTests.swift
```diff
@@ -1208,944 +1208,108 @@ class SwiftObjectServerTests: SwiftSyncTestCase {
...
-        }
-
-        DispatchQueue.global().async {
-            watchTestUtility.isOpenSemaphore.wait()
-            for _ in 0..<3 {
-                collection.insertOne(document) { result in
-                    if case .failure = result {
-                        XCTFail("Should insert")
-                    }
+    func testCallFunctionResult() throws {
```
jsflax, line 1211: Did this get moved? I thought we already had a test for this.

RealmSwift/Sync.swift
```diff
@@ -311,12 +311,29 @@ public typealias Provider = RLMIdentityProvider
         }
     }
 
-    #if !(os(iOS) && (arch(i386) || arch(arm)))
     /// The implementation of @dynamicMemberLookup that allows for dynamic remote function calls.
+    public subscript(dynamicMember string: String) -> FunctionCallable {
```
jsflax, line 315: The above doc string needs updating.

RealmSwift/Sync.swift
```diff
@@ -311,12 +311,29 @@ public typealias Provider = RLMIdentityProvider
...
+        FunctionCallable(name: string, user: user)
+    }
+}
+
+@dynamicCallable
+public struct FunctionCallable {
```
jsflax, line 321: These new public structs and methods need docstrings.

Realm/ObjectServerTests/SwiftObjectServerTests.swift
dianaafanador3: I added a new test but the the diff makes it look like I deleted some tests, I merge it into the callFunction test

Realm/ObjectServerTests/SwiftObjectServerTests.swift
dianaafanador3: No extra line

jsflax (approved):
Not sure about the CI failures– I'm able to compile it in Xcode 12.2. Looks good otherwise!

jsflax:
Re-requesting my own review. The cocoa pods failures seem legit. Otherwise, looks good.

dianaafanador3:
Added a dummy implementation for `dynamicallyCall(withArguments:)` for builds with older architecture which do not support `Combine` and/or `Async Await`

dianaafanador3:
Added #canImport(_Concurrency) flag to code that uses the Concurrency module

dianaafanador3:
CI passing now @jsflax

RealmSwift/Sync.swift
```diff
@@ -311,12 +311,43 @@ public typealias Provider = RLMIdentityProvider
         }
     }
 
+    /// The implementation of @dynamicMemberLookup that allows for dynamic remote function calls with a `callable` return.
+    public subscript(dynamicMember string: String) -> FunctionCallable {
+        FunctionCallable(name: string, user: user)
+    }
+}
+
+/// Structure providing an interface to been able to return a `callable` for a MongoDB Realm function.
```
jsflax, line 320: I would reword this since it's not really saying anything. Why does it exist? What are we enabling? I would call out the syntactic sugar here similarly to line 763.

RealmSwift/Sync.swift
```diff
@@ -311,12 +311,43 @@ public typealias Provider = RLMIdentityProvider
...
+@dynamicCallable
+public struct FunctionCallable {
+    fileprivate let name: String
+    fileprivate let user: User
+
+    /// :nodoc:
+    public func dynamicallyCall(withArguments args: [Never]) {
+        /// Dummy implementation for @dynamicCallable
```
jsflax, line 328: I wouldn't actually use the word "Dummy" here 😅 . I would simply say `/// noop` and see if we can place it in the #else block beneath the #if block to better telegraph what this is doing.

jsflax (commented):
Two more comments! Then LGTM.

jsflax (approved):
🎉

leemaguire (approved):
LGTM 🎉

PR #7391, opened 2021-08-13
Loosen protocol requirements from ObjectKeyIdentifiable to Identifiable

tgoyne:
ObjectKeyIdentifiable is just a helper type to make objects without primary keys conform to Identifiable, and isn't something that should ever be required.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@ x.y.z Release notes (yyyy-MM-dd)
 ### Fixed
 * Fix crash in `MongoCollection.findOneDocument(filter:)` that occurred when no results were 
   found for a given filter.
+* Some of the SwiftUI property wrappers incorrectly required objects to conform
+  to ObjectKeyIdentifiable rather than Identifiable.
 
 <!-- ### Breaking Changes - ONLY INCLUDE FOR NEW MAJOR version -->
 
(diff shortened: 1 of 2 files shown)


rromanchuk:
This fixed everything, thanks! 

This was hard to debug, but mostly because of bad luck. When I upgraded my very limited swiftui views to new property wrappers, the exiting `Identifiable` objects now triggered compiler hint to `ObjectKeyIdentifiable`. Added the new protocol and everything compiled and happy.  Then during run time, a create call with `update: .modified`  threw for missing primary key. I assumed there was an invalid json response with null/missing id. It wasn't until i actually opened the realm in studio i realized primary key was missing from the schema, and unrelated to my response payload. The runtime fatal was in my generic CRUD operations making it feel very random. 

```swift
    @discardableResult
    static func success<K: Any>(_ unmanaged: [K]) -> (Realm, [RealmType]) {
        try! autoreleasepool {
            var managedObjects = [RealmType]()
            do {
                let realm = try Realm()
                realm.beginWrite()
                unmanaged.forEach { (unmanagedObject) in
                    let object = realm.create(RealmType.self, value: unmanagedObject, update: .modified) 
                    (object as? WithSaveHooks)?.afterCreate()
                    managedObjects.append(object)
                }
                try realm.commitWrite()
                return (realm, managedObjects)
            } catch let error {
                log.error(error)
                SentrySDK.capture(error: error)
                throw error
            }
        }
    }
```

leemaguire (approved):

PR #7398, opened 2021-08-17
Add callback queue parameter to async open

ericjordanmossman:


leemaguire:
This looks like you just need to use `receive(on:)`? e.g

```swift
Realm.asyncOpen().receive(on: DispatchQueue.main).sink { _ in

        } receiveValue: { realm in

        }
```

ericjordanmossman:
> This looks like you just need to use receive(on:)

Hmm, yes it does. I'll post that in the issue this came from.
Ultimately though do you think callback queue should be added here?

leemaguire:
> > This looks like you just need to use receive(on:)
> 
> Hmm, yes it does. I'll post that in the issue this came from.
> Ultimately though do you think callback queue should be added here?

Nope, due to the fact it can already be added within the chain.

PR #7400, opened 2021-08-17
Fix a rare crash when running the sync tests

tgoyne:
Capturing a strong reference to the app in the error callback results in things breaking when that's the last remaining reference to the App. This isn't something that can happen in normal uses but our tests would occasionally hit it.

### DIFF
--- a/Realm/ObjectServerTests/RealmServer.swift
+++ b/Realm/ObjectServerTests/RealmServer.swift
@@ -519,15 +519,16 @@ public class RealmServer: NSObject {
         let binDir = Self.buildDir.appendingPathComponent("bin").path
         let libDir = Self.buildDir.appendingPathComponent("lib").path
         let binPath = "$PATH:\(binDir)"
+        let env = [
+            "PATH": binPath,
+            "DYLD_LIBRARY_PATH": libDir
+        ]
 
         let stitchRoot = RealmServer.buildDir.path + "/go/src/github.com/10gen/stitch"
 
         // create the admin user
         let userProcess = Process()
-        userProcess.environment = [
-            "PATH": binPath,
-            "LD_LIBRARY_PATH": libDir
-        ]
+        userProcess.environment = env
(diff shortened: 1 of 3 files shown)


Realm/ObjectServerTests/RealmServer.swift
```diff
@@ -519,15 +519,16 @@ public class RealmServer: NSObject {
         let binDir = Self.buildDir.appendingPathComponent("bin").path
         let libDir = Self.buildDir.appendingPathComponent("lib").path
         let binPath = "$PATH:\(binDir)"
+        let env = [
+            "PATH": binPath,
+            "DYLD_LIBRARY_PATH": libDir
```
tgoyne, line 524: DYLD_LIBRARY_PATH was always the correct thing here but LD_LIBRARY_PATH worked until macOS 12.

Realm/RLMSyncConfiguration.mm
```diff
@@ -183,8 +184,10 @@ - (instancetype)initWithUser:(RLMUser *)user
                     break;
             }
 
-            RLMSyncManager *manager = [app syncManager];
-            auto errorHandler = manager.errorHandler;
+            RLMSyncErrorReportingBlock errorHandler;
+            @autoreleasepool {
```
leemaguire, line 188: Is there a need for `@autoreleasepool` here?

leemaguire (approved):
LGTM, one minor comment

Realm/RLMSyncConfiguration.mm
tgoyne: Accessing a weak pointer autoreleases the value. This code's running on the sync worker thread, so letting an autoreleased value escape from the function could result in it living a very long time. I don't think it'd lead to a retain cycle or the crash being fixed here, but minimizing how long we're holding a strong reference to the sync manager seems the most likely to avoid problems.

PR #7401, opened 2021-08-17
Work around canImport(Combine) being broken in Xcode 13

tgoyne:
Xcode 13 beta 3 added a swiftinterface file for armv7 Combine which makes canImport(Combine) return true, but the swiftinterface file doesn't actually compile (and no armv7 devices can run OS versions with Combine). Work around this by replacing the `canImport(Combine)` checks with a define we manually set when building for armv7.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,10 +4,11 @@ x.y.z Release notes (yyyy-MM-dd)
 * None.
 
 ### Fixed
-* Fix crash in `MongoCollection.findOneDocument(filter:)` that occurred when no results were 
+* Fix crash in `MongoCollection.findOneDocument(filter:)` that occurred when no results were
   found for a given filter.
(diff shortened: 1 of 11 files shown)


leemaguire:
There still seems to be issues with Combine and it doesn't seem related to armv7?

❌  /Applications/Xcode-13.app/Contents/Developer/Platforms/iPhoneSimulator.platform/Developer/SDKs/iPhoneSimulator15.0.sdk/System/Library/Frameworks/Combine.framework/Modules/Combine.swiftmodule/i386-apple-ios-simulator.swiftinterface:57:59: 'AnyCancellable' is only available in iOS 13.0 or newer

tgoyne:
Oh, right, i386 is effected too. It's 32-bit iOS specifically that doesn't have Combine, so both device and simulator need to exclude it.

dianaafanador3 (approved):

leemaguire (approved):

PR #7402, opened 2021-08-17
Threadsafe Wrapper

ericjordanmossman:
Adds a new `@ThreadSafe` property wrapper which can passed between threads.
```
@ThreadSafe var stringObject: SwiftStringObject?
```

This uses a new protocol called `ThreadSafeWrappable`. `@Threadsafe` uses `ThreadSafeWrappable` instead of `ThreadConfined` because `@Threadsafe` is incompatible with `LinkingObjects`, which is `ThreadConfined`. 

This excludes `LinkingObjects` because a thread safe reference to `LinkingObjects` can't happen within a write transaction. Whenever the wrapped value is accessed, a new thread-safe reference to the wrapped value is created. Instead of realm-cocoa developers hitting this exception the hard, confusing way, they'll get a compiler error.

Also most (if not all) cases would be better served using the origin `List` or `Object`, instead of the `LinkingObjects`


RealmSwift/List.swift
```diff
@@ -834,3 +834,10 @@ extension List: PropertyNameConvertible {
         return (key: rlmArray.propertyKey, isLegacy: rlmArray.isLegacyProperty)
     }
 }
+
+// MARK: Thread Safe Wrapper
+
+extension List: ThreadSafeWrappable {
+    /// :nodoc:
+    public static var confinedType: PropertyType { Element._rlmType }
```
tgoyne, line 842: This doesn't appear to be used anywhere.

RealmSwift/ThreadSafeReference.swift
```diff
@@ -122,6 +122,72 @@ public protocol ThreadConfined {
...
+            // As of Swift 5.5 property wrappers can't have throwing accessors.
+            } catch let error as NSError { throwRealmException(error.localizedDescription) }
+        }
+        set {
+            guard let newValue = newValue else {
+                threadSafeReference = nil
+                return
+            }
+            guard let configuration = newValue.realm?.configuration else {
+                throwRealmException("No realm configuration on this object found. Only managed objects may be wrapped as thread safe.")
```
tgoyne, line 169: The first half of this error message is just an implementation detail. The user doesn't know or care that we happen to grab the configuration off of the Realm, so there's no reason for the exception to mention that.

RealmSwift/ThreadSafeReference.swift
```diff
@@ -122,6 +122,72 @@ public protocol ThreadConfined {
...
+    A property type wrapper type that may be passed between threads.
+
+    A `@ThreadSafe` property contains a thread-safe reference to the underlying wrapped value. This reference is resolved to the thread on which the wrapped value is accessed. A new thread safe reference is created each time the property is accessed.
+
+ - see: `ThreadConfined`
+ - see: `ThreadSafeReference`
+*/
+@propertyWrapper public class ThreadSafe<T: ThreadSafeWrappable> {
+    var threadSafeReference: ThreadSafeReference<T>?
+    var configuration: Realm.Configuration?
```
tgoyne, line 145: Grabbing the RLMRealmConfiguration, opening a RLMRealm and then wrapping that in a Realm would eliminate a nontrivial bit of overhead.

tgoyne (commented):
A ThreadSafeReference can't be constructed for a Results which was created by sorting/filtering/etc. a collection while inside a write transaction either; that isn't a quirk specific to LinkingObjects.

RealmSwift/List.swift
ericjordanmossman: Is it okay to have an empty protocol?

ericjordanmossman: Results will need to be excluded too.

tgoyne: Yes.

RealmSwift/ThreadSafeReference.swift
ericjordanmossman: Does [`5004efb`](https://github.com/realm/realm-cocoa/pull/7402/commits/5004efb3144e73316e0a38d014a6f42407d0038d) do what you mean? And if so, what is the effect?
I was mainly looking at `RLMGetThreadLocalCachedRealmForPath`, but Swift `Realm` opens also hit this path. 
Is the main benefit that `Realm(_ rlmRealm: RLMRealm)` is one less call than `Realm(configuration: Configuration, queue: DispatchQueue? = nil)`?
Or is it significant that `Realm(_ rlmRealm:)` doesn't throw?

nirinchev:
I'm guessing this design has been discussed within the Cocoa team, but since this functionality would be really useful for other SDKs as well, I'm curious to hear what your thoughts are on the effect this will have on filesize growth. As far as I can tell from the code, setting a `@ThreadSafe` property will eagerly create a ThreadSafeReference to the object and keep it around until the property is accessed at which point it will resolve the reference and create a new one. If my understanding is correct, wouldn't this run a risk of having a rarely used property pin a version for a long time, thus causing excessive filesize growth? Is there something I'm missing about how property wrappers or the Cocoa SDK works that would alleviate that or do you think that it's not a real-world concern?

RealmSwift/ThreadSafeReference.swift
tgoyne: The main benefit is that converting Realm.Configuration to RLMRealmConfiguration does a decent amount of work so it's better to do that only once.

tgoyne:
The use-case where this PW will work well is for function arguments and local variables that get captured by an async dispatched block. Using them for properties on long-lived objects probably is a footgun that should be explicitly called out in the docs, but that's sort of a pre-existing footgun.

tgoyne:
On a related note, most of the ThreadSafeReference types don't actually need to pin the version any more but still are. I think that'd be a really quick and easy thing to fix.

RealmSwift/ThreadSafeReference.swift
```diff
@@ -122,6 +122,78 @@ public protocol ThreadConfined {
     }
 }
 
+/**
+ Object types conforming to `ThreadSafeWrappable` are able to have the `@ThreadSafe` property wrapper applied to them.
+
+ Note that only types defined by Realm can meaningfully conform to this protocol, and defining new
+ classes which attempt to conform to it will not make them work with `@ThreadSafe`
+ */
+public protocol ThreadSafeWrappable: ThreadConfined {}
```
tgoyne, line 131: https://github.com/realm/realm-core/pull/4857 fixes the problem with using @ThreadSafe on Results and LinkingObjects, so this protocol should no longer be needed

CHANGELOG.md
```diff
@@ -1,7 +1,7 @@
 x.y.z Release notes (yyyy-MM-dd)
 =============================================================
 ### Enhancements
-* None.
+* Add `@ThreadSafe` property wrapper. This property wrapper type may be passed between threads.
```
tgoyne, line 4: This doesn't really tell me what I'd use it for. It should mention that you use it on local variables and function parameters to make it so that the value can be used on multiple threads.

RealmSwift/ThreadSafeReference.swift
```diff
@@ -122,6 +122,78 @@ public protocol ThreadConfined {
...
+            // FIXME: wrappedValue should throw
+            // As of Swift 5.5 property wrappers can't have throwing accessors.
+            } catch let error as NSError { throwRealmException(error.localizedDescription) }
+        }
+        set {
+            guard let newValue = newValue else {
+                threadSafeReference = nil
+                return
+            }
+            guard let configuration = newValue.realm?.configuration else {
```
tgoyne, line 174: `newValue.realm?.rlmRealm.configuration`

Currently this is converting the RLMRealmConfiguration to a Realm.Configuration and then back, and we want to instead grab the RLMRealmConfiguration from the RLMRealm directly.

RealmSwift/ThreadSafeReference.swift
```diff
@@ -122,6 +122,73 @@ public protocol ThreadConfined {
...
+                let realm = Realm(rlmRealm)
+                guard let value = threadSafeReference.resolve(in: realm) else {
+                    self.threadSafeReference = nil
+                    return nil
+                }
+                self.threadSafeReference = ThreadSafeReference(to: value)
+                return value
+            // FIXME: wrappedValue should throw
+            // As of Swift 5.5 property wrappers can't have throwing accessors.
+            } catch let error as NSError { throwRealmException(error.localizedDescription)
```
jsflax, line 161: Missing new line before `throwRealmException`.

RealmSwift/ThreadSafeReference.swift
```diff
@@ -122,6 +122,73 @@ public protocol ThreadConfined {
     }
 }
 
+/**
```
jsflax, line 125: This should have a MARK above it:
```
// MARK: ThreadSafe propertyWrapper
```

jsflax:
@tgoyne do you think it would be worth it for us to make `@ThreadSafe` into a struct with a mutating getter? 

An argument can be made that value semantics actually _do_ make sense here, as we are modifying the underlying ThreadSafeReference. It also signals to the user that getting this value is doing "more" under the hood.

On the other hand, that may just be a hassle for the end user. Curious to hear your thoughts.

tgoyne:
Using a struct instead of a class would be a small perf improvement. I don't think it'd have any change on the behavior.

RealmSwift/ThreadSafeReference.swift
ericjordanmossman: Should be changed in latest commit. Not sure if github ui picked it up as outdated.

RealmSwift/ThreadSafeReference.swift
```diff
@@ -122,6 +122,76 @@ public protocol ThreadConfined {
...
+    }
+
+    /// :nodoc:
+    public init(wrappedValue: T?) {
+        guard let wrappedValue = wrappedValue else {
+            self.threadSafeReference = nil
+            self.rlmConfiguration = nil
+            return
+        }
+        guard let config = wrappedValue.realm?.configuration else {
```
tgoyne, line 187: This should grab the configuration in the same way as the above code (or this should be just `self.wrappedValue = wrappedValue` if that's legal).

CHANGELOG.md
```diff
@@ -1,7 +1,9 @@
 x.y.z Release notes (yyyy-MM-dd)
 =============================================================
 ### Enhancements
-* None.
+* Add `@ThreadSafe` property wrapper. This property wrapper type may be passed between threads. It's
```
jsflax, line 4: Bad grammar.

RealmSwift/App.swift
```diff
@@ -542,7 +542,7 @@ public extension APIKeyAuth {
     }
 }
 
-#if swift(>=5.5)
+#if swift(>=5.5) && canImport(_Concurrency)
```
jsflax, line 545: Should pull latest master.

RealmSwift/Tests/ThreadSafeReferenceTests.swift
```diff
@@ -330,3 +330,456 @@ class ThreadSafeReferenceTests: TestCase {
...
+    @ThreadSafe var stringObject: SwiftStringObject?
+    @ThreadSafe var intObject: SwiftIntObject?
+    @ThreadSafe var employees: List<SwiftEmployeeObject>?
+    @ThreadSafe var employeeSet: MutableSet<SwiftEmployeeObject>?
+    @ThreadSafe var owners: LinkingObjects<SwiftOwnerObject>?
+    @ThreadSafe var results: Results<SwiftStringObject>?
+
+    @ThreadSafe var arcResults: AnyRealmCollection<SwiftEmployeeObject>?
+    @ThreadSafe var arcList: AnyRealmCollection<SwiftEmployeeObject>?
+    @ThreadSafe var arcSet: AnyRealmCollection<SwiftEmployeeObject>?
```
leemaguire, line 345: This is missing `Map` tests

leemaguire: Should add a test with `@ThreadSafe var fooMap: Map<String, SwiftIntObject?>`

leemaguire (approved):
LGTM 👍🏼

RealmSwift/Tests/ThreadSafeReferenceTests.swift
```diff
@@ -330,3 +330,480 @@ class ThreadSafeReferenceTests: TestCase {
...
+                obj?.stringCol = "middle"
+            }
+            mainEx.fulfill()
+        }
+        backgroundQueue.async {
+            try! obj?.realm?.write {
+                obj?.stringCol = "after"
+            }
+            backgroundEx.fulfill()
+        }
```
tgoyne, line 804: `concurrentPerform(iterations:execute:)` with something like 100 iterations would exercise this a bit better.

RealmSwift/Tests/ThreadSafeReferenceTests.swift
```diff
@@ -330,3 +330,480 @@ class ThreadSafeReferenceTests: TestCase {
...
+        waitForExpectations(timeout: 0.1, handler: nil)
```
tgoyne, line 805: Even if 100ms is normally totally overkill, anything under a second is likely to have some occasional spurious failures on CI.

RealmSwift/Tests/ThreadSafeReferenceTests.swift
```diff
@@ -330,3 +330,480 @@ class ThreadSafeReferenceTests: TestCase {
...
+        guard let stringObj = stringObj else {
+            XCTFail("no object found")
+            return
+        }
+        try! stringObj.realm!.write {
+            stringObj.stringCol = "after"
+        }
+    }
+
+    func testThreadSafeFunctionArgument() {
```
tgoyne, line 762: This test would pass even if mutateStringCol()'s argument wasn't `@ThreadSafe`. The dispatch should be happening inside the function using `@ThreadSafe`, not the caller.

RealmSwift/ThreadSafeReference.swift
```diff
@@ -122,6 +122,76 @@ public protocol ThreadConfined {
...
+            wrapper is **better suited for function arguments and local variables**
+            **that get captured by an aynchronously dispatched block.**
+
+ - see: `ThreadSafeReference`
+ - see: `ThreadConfined`
+*/
+@propertyWrapper public class ThreadSafe<T: ThreadConfined> {
+    private var threadSafeReference: ThreadSafeReference<T>?
+    private var rlmConfiguration: RLMRealmConfiguration?
+    private var barrier = DispatchQueue(label: "thread-safe-property-wrapper")
```
tgoyne, line 146: Creating a queue for every property wrapper is a *lot* of overhead. GCD queues were supposed to be cheap to create, but that didn't work out in practice. https://gist.github.com/tclementdev/6af616354912b0347cdf6db159c37057 suggests that this should either be a static queue shared between all of them or just a NSLock.

RealmSwift/ThreadSafeReference.swift
```diff
@@ -122,6 +122,76 @@ public protocol ThreadConfined {
...
+
+    /// :nodoc:
+    public var wrappedValue: T? {
+        get {
+            barrier.sync(flags: .barrier) {
```
tgoyne, line 151: `.barrier` doesn't do anything here. It's only applicable to concurrent queues.

tgoyne (approved):

PR #7405, opened 2021-08-23
Remove comment

ericjordanmossman:


jsflax:
Why are we removing this? Seems like we want that test to exist at some point?

ericjordanmossman:
I created the PR because I left it in by accident. Now I wasn't sure whether it's okay to leave in. But now I am. :)

Issue #7399, opened 2021-08-17
Crash when deleting/disabling user from MongoDB dashboard


zcohan:
## Goals
When deleting or disabling a signed-in user from the MongoDB Realm dashboard, my application crashes

## Expected Results
I'm not sure what the correct way for handling a user being deleted is, but perhaps there's a notification of some kind? Certainly not a hard crash.

## Actual Results
*** Terminating app due to uncaught exception 'NSInvalidArgumentException', reason: '*** -[__NSPlaceholderDictionary initWithObjects:forKeys:count:]: attempt to insert nil object from objects[0]'
*** First throw call stack:
(
	0   CoreFoundation                      0x00007fff206d387b __exceptionPreprocess + 242
	1   libobjc.A.dylib                     0x00007fff2040bd92 objc_exception_throw + 48
	2   CoreFoundation                      0x00007fff207892ea _CFThrowFormattedException + 202
	3   CoreFoundation                      0x00007fff20794dd1 -[__NSPlaceholderDictionary initWithCapacity:].cold.1 + 0
	4   CoreFoundation                      0x00007fff205e2ecd -[__NSPlaceholderDictionary initWithObjects:forKeys:count:] + 251
	5   CoreFoundation                      0x00007fff205e2db8 +[NSDictionary dictionaryWithObjects:forKeys:count:] + 49
	6   Realm                               0x0000000104dc1b7f _ZNSt3__110__function6__funcIZ77-[RLMSyncConfiguration initWithUser:partitionValue:customFileURL:stopPolicy:]E3$_0NS_9allocatorIS2_EEFvNS_10shared_ptrIN5realm11SyncSessionEEENS6_9SyncErrorEEEclEOS8_OS9_ + 987
	7   Realm                               0x00000001051b3761 _ZNSt3__110__function6__funcIZN5realm11SyncSession14handle_refreshENS_10shared_ptrIS3_EEE3$_0NS_9allocatorIS6_EEFvNS2_4util8OptionalINS2_3app8AppErrorEEEEEclEOSD_ + 1073
	8   Realm                               0x00000001051bfc1f _ZNSt3__110__function6__funcIZN5realm8SyncUser19refresh_custom_dataENS_8functionIFvNS2_4util8OptionalINS2_3app8AppErrorEEEEEEE3$_8NS_9allocatorISC_EESA_EclEOS9_ + 223
[...]

PR #7408, opened 2021-08-24
Fix crash in `RLMSyncConfiguration.initWithUser` error mapping when a user is disabled/deleted from MongoDB Realm

dianaafanador3:
Fix crash in `RLMSyncConfiguration.initWithUser` error mapping when a user is disabled/deleted from MongoDB Realm dashboard (https://github.com/realm/realm-cocoa/issues/7399)


jsflax (commented):
There is probably a way to test this using the admin API– would be good to see a test for it.

dianaafanador3:
> There is probably a way to test this using the admin API– would be good to see a test for it.

Added a Test for this specific use case, deleting the user using the Admin API

Realm/ObjectServerTests/SwiftObjectServerTests.swift
```diff
@@ -33,6 +33,12 @@ import RealmTestSupport
 @available(OSX 10.14, *)
 @objc(SwiftObjectServerTests)
 class SwiftObjectServerTests: SwiftSyncTestCase {
+    var notificationToken: NotificationToken?
```
leemaguire, line 36: [minor] space between property and method

Realm/ObjectServerTests/SwiftObjectServerTests.swift
```diff
@@ -872,6 +878,75 @@ class SwiftObjectServerTests: SwiftSyncTestCase {
...
+        wait(for: [loginEx], timeout: 4.0)
+
+        let realm = try openRealm(partitionValue: #function, user: syncUser!)
+        XCTAssertNotNil(realm)
+        if isParent {
+            executeChild()
+            waitForDownloads(for: realm)
+            checkCount(expected: 3, realm, SwiftPerson.self)
+
+            notificationToken = realm.objects(SwiftPerson.self).observe(on: DispatchQueue.main, { _ in
```
leemaguire, line 912: This could do with a comment explaining why you are observing

Realm/ObjectServerTests/SwiftObjectServerTests.swift
```diff
@@ -872,6 +878,75 @@ class SwiftObjectServerTests: SwiftSyncTestCase {
         XCTAssertEqual(app.allUsers.count, 1)
     }
 
+    func testSafelyRemoveUserFromMongoDBRealm() throws {
```
leemaguire, line 881: I think you can omit `FromMongoDBRealm` here

Realm/ObjectServerTests/RealmServer.swift
```diff
@@ -809,6 +817,41 @@ public class RealmServer: NSObject {
...
         try createAppForBSONType("string")
     }
+
+    // Retrieve MongoDB Realm AppId with ClientAppId using the Admin API
+    public func retrieveAppId(clientAppId: String) throws -> String {
+        guard let session = session else {
+            throw URLError(.unknown)
+        }
+
+        let appsListInfo = try session.apps.get().get()
```
leemaguire, line 827: Why is this `get().get()`?

Realm/ObjectServerTests/SwiftObjectServerTests.swift
```diff
@@ -872,6 +872,56 @@ class SwiftObjectServerTests: SwiftSyncTestCase {
...
+            let ex = expectation(description: "Error callback should fire upon receiving an error")
+            app.syncManager.errorHandler = { (error, _) in
+                XCTAssertNotNil(error)
+                blockCalled = true
+                ex.fulfill()
+            }
+
+            let appServerId = try RealmServer.shared.retrieveAppId(clientAppId: appId)
+            let deleteUserEx = expectation(description: "delete-user")
+            RealmServer.shared.removeUserForApp(appServerId, userId: app.currentUser!.id) { result in
```
leemaguire, line 904:
```suggestion
            RealmServer.shared.removeUserForApp(appServerId, userId: user.id) { result in
```

Realm/ObjectServerTests/RealmServer.swift
dianaafanador3: first get() use the `GET` as the https method, and the second one returns the success as a throwing expression instead of using a block here

leemaguire (approved):
LGTM once changelog conflict is resolved

Realm/ObjectServerTests/RealmServer.swift
```diff
@@ -809,6 +817,41 @@ public class RealmServer: NSObject {
     @objc public func createApp() throws -> AppId {
         try createAppForBSONType("string")
     }
+
+    // Retrieve MongoDB Realm AppId with ClientAppId using the Admin API
+    public func retrieveAppId(clientAppId: String) throws -> String {
```
jsflax, line 822: This concept should not be exposed as a public method. `removeUserForApp` should take the clientAppId as that is the only concept that should be exposed to the non-server caller.

Realm/ObjectServerTests/RealmServer.swift
```diff
@@ -809,6 +817,41 @@ public class RealmServer: NSObject {
...
+        guard let session = session else {
+            throw URLError(.unknown)
+        }
+
+        let appsListInfo = try session.apps.get().get()
+        guard let appsList = appsListInfo as? [[String: Any]] else {
+            throw URLError(.badServerResponse)
+        }
+
+        let app = appsList.filter {
```
jsflax, line 832: This should just use `first`.

Realm/ObjectServerTests/RealmServer.swift
```diff
@@ -809,6 +817,41 @@ public class RealmServer: NSObject {
...
+        }.first
+
+        guard let appId = app?["_id"] as? String else {
+            throw URLError(.badServerResponse)
+        }
+        return appId
+    }
+
+    // Remove User from MongoDB Realm using the Admin API
+    public func removeUserForApp(_ appId: String, userId: String, _ completion: @escaping (Result<Any?, Error>) -> Void) {
```
jsflax, line 847: See comments above.

jsflax (approved):

PR #7419, opened 2021-08-27
Add support for type safe queries

leemaguire:
This PR adds the ability to use Swift style expressions when querying a collection of `Results`. These expressions are what Swift developers would be accustomed to when using the `.filter(_:)` method on collections.

This PR does not deprecate filtering with `NSPredicate`

Usage:
```swift
class Person: Object {
     @Persisted var name: String
     @Persisted var hobbies: MutableSet<String>
     @Persisted var pets: List<Pet>
}
class Pet: Object {
     @Persisted var name: String
     @Persisted var age: 
}

let persons = realm.objects(Person.self).where {
     $0.hobbies.contains("music") || $0.hobbies.contains("baseball")
}

persons = realm.objects(Person.self).where {
     ($0.pets.age >= 2) && $0.pets.name.starts(with: "L")
}
```

### Comparisons
| Type               | Expression    |
| ------------- | ------------- |
| Equals | `==` |
| Not Equals | `!=` |
| Greater than | `>`  |
| Greater than or equal | `>=` |
| Less than | `<`  |
| Less than or equal | `<=` |
| Between | `.contains(_ range:)` |

### Collections
| Type               | Expression    |
| ------------- | ------------- |
| IN | `.contains(_ element:)` |
| Between | `.contains(_ range:)` |

### Map
| Type               | Expression    |
| ------------- | ------------- |
| `@allKeys` | `.keys` |
| `@allValues` | `.values` |

### Compound
| Type               | Expression    |
| ------------- | ------------- |
| And | `&&` |
| Or | `\|\|` |

### Collection Aggregation
| Type               | Expression    |
| ------------- | ------------- |
| `@avg` | `.avg` |
| `@min` | `.min` |
| `@max` | `.max` |
| `@sum` | `.sum` |
| `@count` | `.count` |

### Other
| Type               | Expression    |
| ------------- | ------------- |
| Not | `!` |
| Subquery | `($0.fooList.intCol >= 5).count > n` |

### DIFF
--- a/Realm/RLMQueryUtil.mm
+++ b/Realm/RLMQueryUtil.mm
@@ -1336,7 +1336,6 @@ struct KeyPath {
     std::vector<RLMProperty *> links;
     RLMProperty *property;
     bool containsToManyRelationship;
-    bool isDictionary;
 };
 
 KeyPath key_path_from_string(RLMSchema *schema, RLMObjectSchema *objectSchema, NSString *keyPath)
@@ -1345,7 +1344,6 @@ KeyPath key_path_from_string(RLMSchema *schema, RLMObjectSchema *objectSchema, N
     std::vector<RLMProperty *> links;
 
     bool keyPathContainsToManyRelationship = false;
-    bool isDictionary = false;
 
     NSUInteger start = 0, length = keyPath.length, end = NSNotFound;
     do {
@@ -1358,8 +1356,6 @@ KeyPath key_path_from_string(RLMSchema *schema, RLMObjectSchema *objectSchema, N
 
         if (property.collection)
             keyPathContainsToManyRelationship = true;
-        if (property.dictionary)
-            isDictionary = true;
 
         if (end != NSNotFound) {
             RLMPrecondition(property.type == RLMPropertyTypeObject || property.type == RLMPropertyTypeLinkingObjects,
@@ -1374,7 +1370,7 @@ KeyPath key_path_from_string(RLMSchema *schema, RLMObjectSchema *objectSchema, N
         start = end + 1;
     } while (end != NSNotFound);
 
-    return {std::move(links), property, keyPathContainsToManyRelationship, isDictionary};
+    return {std::move(links), property, keyPathContainsToManyRelationship};
 }
 
 ColumnReference QueryBuilder::column_reference_from_key_path(RLMObjectSchema *objectSchema,
@@ -1385,7 +1381,7 @@ ColumnReference QueryBuilder::column_reference_from_key_path(RLMObjectSchema *ob
     if (isAggregate && !keyPath.containsToManyRelationship) {
         throwException(@"Invalid predicate",
                        @"Aggregate operations can only be used on key paths that include an collection property");
-    } else if (!isAggregate && keyPath.containsToManyRelationship && !keyPath.isDictionary) {
+    } else if (!isAggregate && keyPath.containsToManyRelationship) {
         throwException(@"Invalid predicate",
                        @"Key paths that include a collection property must use aggregate operations");
     }
@@ -1874,20 +1870,22 @@ NSExpression *simplify_self_value_for_key_path_function_expression(NSExpression 
 void QueryBuilder::apply_map_expression(RLMObjectSchema *objectSchema, NSExpression *functionExpression,
                                         NSComparisonPredicateOptions options, NSPredicateOperatorType operatorType,
                                         NSExpression *right) {
-    NSExpression *keyPathExpression;
+    NSString *keyPath;
     NSString *mapKey;
     if (functionExpression.operand.expressionType == NSKeyPathExpressionType) {
         NSExpression *mapItems = [functionExpression.arguments firstObject];
         NSExpression *linkCol = [[functionExpression.operand arguments] firstObject];
         NSExpression *mapCol = [mapItems.arguments firstObject];
         mapKey = [mapItems.arguments[1] constantValue];
-        keyPathExpression = [NSExpression expressionForKeyPath:[NSString stringWithFormat:@"%@.%@", linkCol.keyPath, mapCol.keyPath]];
+        keyPath = [NSString stringWithFormat:@"%@.%@", linkCol.keyPath, mapCol.keyPath];
     } else {
-        keyPathExpression = functionExpression.arguments.firstObject;
+        keyPath = [functionExpression.arguments.firstObject keyPath];
         mapKey = [functionExpression.arguments[1] constantValue];
     }
 
-    ColumnReference collectionColumn = column_reference_from_key_path(objectSchema, [keyPathExpression keyPath], true);
+    ColumnReference collectionColumn = column_reference_from_key_path(objectSchema, keyPath, true);
+    RLMPrecondition(collectionColumn.property().dictionary,
+                    @"Invalid predicate", @"Only dictionaries support subscript predicates.");
     add_mixed_constraint(operatorType, options, collectionColumn.resolve<Dictionary>().key([mapKey UTF8String]), [right constantValue]);
 }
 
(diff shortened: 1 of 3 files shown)


RealmSwift/Query.swift
```diff
@@ -0,0 +1,856 @@
...
+            case let .rhs(v):
+                predicateString.append(" %@")
+                arguments.append(v.objCValue)
+            case let .subquery(col, str, args):
+                predicateString.append("SUBQUERY(\(col), $obj, \(str)).@count")
+                arguments.append(contentsOf: args)
+            case let .collectionAggregation(agg):
+                predicateString.append(agg.rawValue)
+            case let .keypathCollectionAggregation(agg):
+                predicateString.insert(agg.rawValue, at: predicateString.count-2)
```
leemaguire, line 352: @dianaafanador3 
This will crash if the query expression only contains one keyPath. e.g this test will crash:
```swift
assertCollectionQuery(on: \.intCol,
                              isList: true,
                              predicate: "intCol.@avg > %@",
                              values: [1],
                              expectedCount: 1) {
            $0.intCol.avg > 1
        }
```

RealmSwift/Query.swift
```diff
@@ -0,0 +1,878 @@
...
+            case let .rhs(v):
+                predicateString.append(" %@")
+                arguments.append(v.objCValue)
+            case let .subquery(col, str, args):
+                predicateString.append("SUBQUERY(\(col), $obj, \(str)).@count")
+                arguments.append(contentsOf: args)
+            case let .collectionAggregation(agg):
+                predicateString.append(agg.rawValue)
+            case let .keypathCollectionAggregation(agg):
+                if idx-2 >= 0,
```
leemaguire, line 362: @dianaafanador3 we need comments explaining this logic

Realm/RLMQueryUtil.mm
```diff
@@ -1378,7 +1382,7 @@ KeyPath key_path_from_string(RLMSchema *schema, RLMObjectSchema *objectSchema, N
     if (isAggregate && !keyPath.containsToManyRelationship) {
         throwException(@"Invalid predicate",
                        @"Aggregate operations can only be used on key paths that include an array property");
-    } else if (!isAggregate && keyPath.containsToManyRelationship) {
+    } else if (!isAggregate && keyPath.containsToManyRelationship && !keyPath.isDictionary) {
```
leemaguire, line 1385: Loosening this requirement allows us to create more flexible queries with Map. Core has no issue with using a map property as the lhs of a query expression when querying for a value inside of a Map.

Realm/Tests/QueryTests.m
```diff
@@ -2155,7 +2155,6 @@ - (void)testDictionaryMultiLevelLinkQuery {
     XCTAssertThrows([CircleDictionaryObject objectsInRealm:realm where:@"ANY data = '2'"]);
     XCTAssertThrows([CircleDictionaryObject objectsInRealm:realm where:@"ANY circles.next = '2'"]);
     XCTAssertThrows([CircleDictionaryObject objectsInRealm:realm where:@"ANY data.circles = '2'"]);
-    XCTAssertThrows([CircleDictionaryObject objectsInRealm:realm where:@"circles.data = '2'"]);
```
tgoyne, line 2158: What does this query now do instead of throwing an exception? There should be a new test replacing this one.

RealmSwift/LinkingObjects.swift
```diff
@@ -96,6 +96,15 @@ import Realm
         return notFoundToNil(index: rlmResults.indexOfObject(with: predicate))
     }
 
+    /**
+     Returns the index of the first object matching the given query, or `nil` if no objects match.
+
+     - parameter query: The query with which to filter the objects.
```
tgoyne, line 102: The parameter to this function isn't a query. In general these docs aren't really going to give anyone any clue how to call this function.

RealmSwift/Query.swift
```diff
@@ -0,0 +1,1075 @@
...
+
+ */
+@dynamicMemberLookup
+public struct Query<T: _RealmSchemaDiscoverable> {
+
+    private var context: QueryContext
+
+    /// Initializes a `Query` object.
+    /// - Parameter isPrimitive: States is the query is on 'self' and will have no key path context.
+    public init(isPrimitive: Bool = false) {
```
tgoyne, line 206: This constructor should probably be internal?

RealmSwift/Query.swift
```diff
@@ -0,0 +1,1075 @@
...
+
+/// Enum representing an option for `String` queries.
+public enum StringOptions {
+    /// A case-insensitive search.
+    case caseInsensitive
+    /// Search ignores diacritic marks.
+    case diacriticInsensitive
+}
+
+private enum QueryExpression {
```
tgoyne, line 29: This representation of an expression makes all of the code working with it very complicated. Building a tree rather than a flat array with some complicated non-end insertions to mimic a tree structure would be a lot less error-prone. It's also not obvious to me that we need our own data type for this at all. Is there anything which would be difficult to do if the builder functions simply assembled a NSPredicate directly?

RealmSwift/Query.swift
```diff
@@ -0,0 +1,1075 @@
...
+        let name = _name(for: member)
+        return append(expression: [.keyPath(name: name, isCollection: true)])
+    }
+
+    // MARK: Query Construction
+
+    /// Creates an NSPredicate compatibe string.
+    /// - Parameter isSubquery: States if expression need to be arraged in a special way to cater to subqueries.
+    /// - Returns: A tuple containing the predicate string and an array of arguments.
+    public func _constructPredicate(_ isSubquery: Bool = false) -> QueryPredicate {
```
tgoyne, line 327: Creating a NSPredicate by serializing our representation to a string and then asking NSPredicate to parse that is pretty roundabout. This all seems like it'd be a lot simpler if it just constructed a NSPredicate.

RealmSwift/Query.swift
```diff
@@ -0,0 +1,1075 @@
...
+                if closedRange {
+                    predicateString.append(" BETWEEN {%@, %@}")
+                    arguments.append(contentsOf: [low, high])
+                } else if idx > 0, case let .keyPath(name, _) = context.expression[idx-1] {
+                    predicateString.append(" >= %@")
+                    arguments.append(low)
+                    predicateString.append(" && \(name) <\(closedRange ? "=" : "") %@")
+                    arguments.append(high)
+                } else {
+                    throwRealmException("Could not construct .contains(_:) predicate")
```
tgoyne, line 359: This error message mentions the wrong thing, but more importantly isn't actionable. If a user writes some code which hits this, the error doesn't tell them what the problem is or give them hints to what they need to change to fix it.

RealmSwift/Query.swift
leemaguire: There are a few cases which make assembling an NSPredicate directly tricky, the two main ones are subqueries and some Map expressions. I'm happy to take a look and see if I can achieve anything with the tree approach and if that leads to better NSPredicate generation.

RealmSwift/Query.swift
leemaguire: I agree, the reason for building the string out like this was because it offered more granular control. For example we need that granular control when construction a collection aggregate predicate– we need to transform `$0.arrayCol.intCol.avg` to `arrayCol.@avg.intCol` so we need to insert the `@avg` operator at a certain position of the string. I'll look into seeing if we can build the NSPredicate along the way instead of passing a large string to the NSPredicate constructor at the end.

RealmSwift/Query.swift
leemaguire: I'm doing a bit of refactoring replacing `QueryExpression` with NSExpression and NSCompoundPredicate etc.

Issue #7390, opened 2021-08-12
RLMException crash: Index out of bounds in results observer


miklselsoe:
I am seeing a crash coming in from the field where I am reacting to a change in an observer on a `Results` object.

This is the exception:
```
*** Terminating app due to uncaught exception 'RLMException', reason: 'Index 2 is out of bounds (must be less than 2).'
```

This is the code that crashes:

```swift
    let orders = realm.objects(Order.self)
    self.observerToken = orders.observe(on: self.processingQueue) { [weak self] (change) in

        switch change {
            case .update(let ordersAfterUpdate, _, _, let modifications):
                let updatedOrders = modifications.map { ordersAfterUpdate[$0] } // <- Crash
                if updatedOrders.isEmpty == false {
                    self?.updateOrders(updatedOrders)
                }
            default:
                break
        }
    }
```

It was my understanding that the indices in the `modifications` array matched what was available in the `ordersAfterUpdate` list, although the documentation isn't very clear on that. I have also tried to open a realm and fetch the orders after an update, but indicies can still be out of bounds. Since this is happening in the field, I don't what leads to this state.

How am I supposed to use the `modifications` list if not as indices on the `ordersAfterUpdate` list? I just want to find the orders that was updated.

## Version of Realm and Tooling
Realm framework version: 10.12.0

Xcode version: 12.5.0

iOS/OSX version: 14.6

Dependency manager + version: Carthage 0.38.0

tgoyne:
The modifications indices are the indices in the previous version of the collection which were modified, as that's what UITableView requires for its batch update functions and the API was designed to make it easy to drive a UITableView. If any of the objects were deleted or you perform non-end insertions then they will be incorrect.

Translating these into indices in the new collection is moderately awkward. There's probably a better way to write this, but the naive version is:
```
func modificationInNew(index: Int, deletions: [Int], insertions: [Int]) -> Int {
    var i = index - deletions.prefix { $0 < index }.count
    for insertion in insertions {
        if insertion > i { break }
        i += 1
    }
    return i
} 
```

If you never do non-end insertions then `index - deletions.prefix { $0 < index }.count` is the only part you need.

miklselsoe:
It seems like this is a documentation issue then. Documentation:
https://github.com/realm/realm-cocoa/blob/83a07f6a508c3427058d9e2c466208d0b6a960fa/RealmSwift/RealmCollection.swift#L158

This is what it says about the `modifications` parameter:
```
[...]

PR #7424, opened 2021-09-01
Fix KVO observation in @StateRealmObject

leemaguire:
# Problem
Currently if you try to add an unmanaged object that is declared with `@StateRealmObject` to a Realm or collection you will get an exception stating that objects with observers cannot be added to the Realm. 

# Solution
This PR addresses that issue by adding a check inside `RLMAccessorContext::createObject` which sees if the object being added to the Realm has any existing `Subscriptions` in `SwiftUIKVO`, if there are then we remove the observers for that object and once the object is successfully managed by the Realm we add back a set of new observers for the properties on that object. The new observers will continue to use the same `SwiftUIKVO` object for receiving the notifications.

I don't see a need to use KVO context pointers here but maybe I'm wrong.

This PR also sneaks in the doc change for https://github.com/realm/realm-cocoa/issues/7390


Realm/RLMAccessor.mm
```diff
@@ -1034,6 +1034,7 @@ static auto toOptional(__unsafe_unretained id const value) {
 
     RLMObjectBase *objBase = RLMDynamicCast<RLMObjectBase>(value);
     realm::Obj obj, *outObj = nullptr;
+    bool requiresSwiftUIObservers;
```
tgoyne, line 1037: This variable needs to be initialized to a value.

Realm/RLMObservation.hpp
```diff
@@ -188,3 +188,12 @@ void RLMDidChange(std::vector<realm::BindingContext::ObserverState> const& obser
 realm::KeyPathArray RLMKeyPathArrayFromStringArray(RLMRealm *realm,
                                                    RLMClassInfo *info,
                                                    NSArray<NSString *> *keyPath);
+
+// Used for checking if `Object`s declared with `@StateRealmObject` need to have
+// their accessors temporarily removed and added back so that the `Object` can be
+// managed be the Realm.
+[[clang::objc_runtime_visible]]
+@interface _SwiftUIKVO : NSObject
```
tgoyne, line 196: This is not a legal name for a type in C. All names starting with an underscore and then followed by a capital letter are reserved for the standard library. It should be RLMSwiftUIKVO.

Realm/RLMObservation.hpp
```diff
@@ -188,3 +188,12 @@ void RLMDidChange(std::vector<realm::BindingContext::ObserverState> const& obser
...
++ (BOOL)removeObserversWithObject:(NSObject *)object;
++ (void)addObserversWithObject:(NSObject *)object;
```
tgoyne, line 198: "FromObject" and "ToObject" rather than "With"

RealmSwift/SwiftUI.swift
```diff
@@ -67,41 +67,64 @@ private func createBinding<T: ThreadConfined, V>(_ value: T,
     })
 }
 
-// MARK: SwiftUIKVO
+// MARK: _SwiftUIKVO
 
+// swiftlint:disable type_name
 @available(iOS 13.0, macOS 10.15, tvOS 13.0, watchOS 6.0, *)
-internal final class SwiftUIKVO: NSObject {
+@objc(_SwiftUIKVO) public class _SwiftUIKVO: NSObject {
```
tgoyne, line 74: This should still be internal and named SwiftUIKVO.

jsflax (approved):

tgoyne (approved):

PR #7425, opened 2021-09-01
Allow indexing PersistableEnum properties

### LINKED ISSUE #7404: Could swift enums + index be made possible with @Persisted?
!!! MANDATORY TO FILL OUT !!!

## Goals

It would be great to make it possible to add index to a swift enum property. (Or if someone can confirm we have no plan to support this)

## Expected Results

Make it possible to use `@Persisted(indexed: true) var property: EnumType` with `enum EnumType: String/Int/..., PersistableEnum`.

## Actual Results

An compile error was produced:
`Referencing initializer 'init(wrappedValue:indexed:)' on 'Persisted' requires that 'EnumType' conform to '_Indexable'`

## Code Sample

```
class Entity: Object {
    @Persisted(indexed: true) var category: Category
    enum Category: String, PersistableEnum {
        case a, b, c, d
    }
}
```

Realm framework version: 10.12

Xcode version: 12.5.1

iOS/OSX version: iOS 14.7

Dependency manager + version: SwiftPackageManager

tgoyne:
The obvious straightforward way to do this would be `extension PersistableEnum: _Indexable where RawValue: _Indexable {}`, but Swift doesn't support conditional protocol inheritance. Instead requires a backport of a bit of the custom type mapping code so that we can check if the persisted type is indexable.

This doesn't really have any tests because I'm not sure what there even is to test other than that the new model classes compile and don't produce schema validation errors.

Fixes #7404.

### DIFF
--- a/RealmSwift/Impl/Persistable.swift
+++ b/RealmSwift/Impl/Persistable.swift
@@ -49,12 +49,21 @@ public protocol _Persistable: _RealmSchemaDiscoverable {
     // If we are in key path tracing mode, instantiate an empty object and forward
     // the lastAccessedNames array.
     static func _rlmKeyPathRecorder(with lastAccessedNames: NSMutableArray) -> Self
+    // The type which is actually stored in the Realm. This is Self for types
+    // we support directly, but may be a different type for enums and mapped types.
+    associatedtype _RealmValue: _Persistable
 }
 
 extension _Persistable {
     public static var _rlmRequiresCaching: Bool {
         false
     }
+}
+
+// A helper protocol for types which we support directly rather than via
+// mapping to a supported type.
+public protocol _BuiltInPersistable {
+    associatedtype _RealmValue = Self
 }
 
 extension _RealmSchemaDiscoverable where Self: _Persistable {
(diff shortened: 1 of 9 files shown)


RealmSwift/Tests/ModernObjectTests.swift
```diff
@@ -134,6 +134,9 @@ class ModernObjectTests: TestCase {
         test(ModernPrimaryObjectIdObject(), ObjectId.generate(), ObjectId.generate())
         test(ModernPrimaryOptionalObjectIdObject(), ObjectId.generate(), nil)
 
+        test(ModernPrimaryIntEnumObject(), .value1, .value2)
+        test(ModernPrimaryOptionalIntEnumObject(), .value1, nil)
```
leemaguire, line 138: Would it be worth adding another modern test object which declares all indexable types and then iterates over the schema to make sure they are indeed indexed?

tgoyne: I've added a test which verifies that all indexable types are marked as indexed in the schema when they should be.

RealmSwift/Impl/Persistable.swift
```diff
@@ -57,6 +60,10 @@ extension _Persistable {
     }
 }
 
+public protocol _BuiltInPersistable {
```
jsflax, line 63: Would be nice to have a comment explaining what this is. I can assume what `BuiltIn` means in this context, but it would nice to be sure.

leemaguire (approved):
I agree with Jasons comment about adding a comment for `_BuiltInPersistable` but LGTM 👍🏼

RealmSwift/Impl/Persistable.swift
tgoyne: Added a short note.

PR #7426, opened 2021-09-03
Resume building releases with Xcode 13

tgoyne:
https://github.com/realm/realm-cocoa/pull/7401 worked around the problem so we want to start packaging it again. I also fixed the remaining things making cocoa-pipeline fail now that the cocoapods release required for that is out, and this branch built successfully: https://ci.realm.io/blue/organizations/jenkins/cocoa-pipeline/detail/cocoa-pipeline/1497/pipeline/

### DIFF
--- a/RealmSwift/BSON.swift
+++ b/RealmSwift/BSON.swift
@@ -379,18 +379,20 @@ extension MinKey: BSON {
     /// Return this BSON as a `Decimal128` if possible.
     /// This will coerce numeric cases (e.g. `.double`) into a `Decimal128` if such coercion would be lossless.
     public func asDecimal128() -> Decimal128? {
+        let str: String
         switch self {
         case let .decimal128(d):
             return d
         case let .int64(i):
-            return try? Decimal128(string: String(i))
+            str = String(i)
         case let .int32(i):
-            return try? Decimal128(string: String(i))
+            str = String(i)
         case let .double(d):
-            return try? Decimal128(string: String(d))
+            str = String(d)
         default:
             return nil
         }
+        return try? Decimal128(string: str)
     }
 
     /// Return this BSON as a `T` if possible, otherwise nil.
--- a/RealmSwift/Decimal128.swift
+++ b/RealmSwift/Decimal128.swift
@@ -87,12 +87,12 @@ public final class Decimal128: RLMDecimal128, Decodable {
 
     /// The mininum value for Decimal128
     public static var min: Decimal128 {
-        __minimumDecimalNumber as! Self
+        unsafeDowncast(__minimumDecimalNumber, to: Self.self)
     }
 
     /// The maximum value for Decimal128
     public static var max: Decimal128 {
-        __maximumDecimalNumber as! Self
+        unsafeDowncast(__maximumDecimalNumber, to: Self.self)
     }
 }
 
@@ -201,7 +201,7 @@ extension Decimal128 {
 
     /// The magnitude of this Decimal128.
     public var magnitude: Magnitude {
-        self.__magnitude as! Magnitude
+        unsafeDowncast(self.__magnitude, to: Magnitude.self)
     }
 
     /// Adds two decimal128 values and produces their sum.
@@ -210,7 +210,7 @@ extension Decimal128 {
     ///   - lhs: The first Decimal128 value to add.
     ///   - rhs: The second Decimal128 value to add.
     public static func + (lhs: Decimal128, rhs: Decimal128) -> Decimal128 {
-        lhs.decimalNumber(byAdding: rhs) as! Decimal128
+        unsafeDowncast(lhs.decimalNumber(byAdding: rhs), to: Decimal128.self)
     }
 
     /// Subtracts one Decimal128 value from another and produces their difference.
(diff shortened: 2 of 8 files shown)


RealmSwift/BSON.swift
```diff
@@ -379,18 +379,20 @@ extension MinKey: BSON {
...
         case let .int32(i):
-            return try? Decimal128(string: String(i))
+            str = String(i)
         case let .double(d):
-            return try? Decimal128(string: String(d))
+            str = String(d)
         default:
             return nil
         }
+        return try? Decimal128(string: str)
```
tgoyne, line 395: This change made this function compile in 5ms rather than 150ms (as reported by `-debug-time-function-bodies`)

RealmSwift/Decimal128.swift
```diff
@@ -250,8 +250,8 @@ extension Decimal128 {
     ///
     /// - Parameter other: The Decimal128 value to calculate the distance to.
     /// - Returns: The distance from this value to `other`.
-    public func distance(to other: Decimal128) -> Stride {
-        other - self
+    public func distance(to other: Decimal128) -> Decimal128 {
+        unsafeDowncast(other.decimalNumber(bySubtracting: self), to: Decimal128.self)
```
tgoyne, line 254: Overloaded operators take a really long time to type-check so each of these were taking close to 100ms to build and now are <1ms. `unsafeDowncast()` ended up not speeding up the build but made the release binary a bit smaller (but not enough to justify listing it in the changelog).

RealmSwift/ObjectiveCSupport+BSON.swift
```diff
@@ -165,7 +165,7 @@ public extension ObjectiveCSupport {
                     return convertBson(object: value)
                 }
                 return .null
-            }.map { $0 == .null ? nil : $0 })
+            }.map { (v: AnyBSON) -> AnyBSON? in v == .null ? nil : v })
```
tgoyne, line 168: Similarly this type hint made this function a lot faster to build.

build.sh
```diff
@@ -649,8 +649,7 @@ case "$COMMAND" in
             export ASAN_OPTIONS='check_initialization_order=true:detect_stack_use_after_return=true'
         fi
         xcrun swift package resolve
-        find .build -name views.cpp -delete
-        xcrun swift test --configuration "$(echo "$CONFIGURATION" | tr "[:upper:]" "[:lower:]")" $SANITIZER
+        xcrun swift test -Xcc -g0 --configuration "$(echo "$CONFIGURATION" | tr "[:upper:]" "[:lower:]")" $SANITIZER
```
tgoyne, line 652: The non-SPM pull request jobs were already disabling debug symbols so this brings SPM in line with them. This knocks ~20% off the build time and cuts the total size of all the object files and such in half, which might fix some of the jobs which were timing out.

examples/installation/build.sh
```diff
@@ -106,7 +106,7 @@ xctest() {
         )
     elif [[ $NAME == SwiftPackageManager* ]]; then
         if [ -n "$sha" ]; then
-            sed -i '' 's@branch = "master"@branch = "'"$sha"'"@' "$DIRECTORY/$NAME.xcodeproj/project.pbxproj"
+            ex '+%s@branch = "master"@branch = "'"$sha"'"@' -scwq "$DIRECTORY/$NAME.xcodeproj/project.pbxproj"
```
tgoyne, line 109: `sed -i` doesn't work on symlinks and this file has been a symlink for a while. `ex` (the line-editor part of vim) does and is a reasonably drop-in replacement.

examples/installation/build.sh
```diff
@@ -133,7 +133,9 @@ xctest() {
     local scheme=(-scheme "$NAME")
 
     # Ensure that dynamic framework tests try to use the correct version of the prebuilt libraries.
-    sed -i '' 's@/realm-swift-latest@/realm-swift-latest/'"${REALM_XCODE_VERSION}"'@' "$DIRECTORY/$NAME.xcodeproj/project.pbxproj"
+    if grep '/realm-swift-latest' "$DIRECTORY/$NAME.xcodeproj/project.pbxproj"; then
```
tgoyne, line 136: Unlike sed, ex reports failure if the expression isn't present in the file so check first.

jsflax (approved):
This is cool and the compilation improvements are unexpected. Did you run `-debug-time-function-bodies` on the entire codebase?

leemaguire (approved):
I learned a fair bit from this 👍🏼

RealmSwift/Decimal128.swift
leemaguire: Is it worth adding some comments to say why you use `unsafeDowncast`? Its not initially apparent that it improves build times.

leemaguire: But I suppose you could just always trace the git blame

tgoyne:
> Did you run -debug-time-function-bodies on the entire codebase?

Yes, and the main thing is showed was that the Swift frontend isn't a meaningful portion of our build time and there isn't much room for improvement there. Everything in non-test code left that's >10ms adds up to a total of ~300ms and most of that is in the long tail of 12ms functions. None of the test code appears to be hitting any particularly bad cases for the compiler either (at least with Xcode 13).

PR #7427, opened 2021-09-03
Upgrade to core 11.4.1

tgoyne:
When using Combine to receive notifications on a different queue, we need to ensure that the changeset version matches the object version, which requires pinning that version. TSRs no longer do that pinning, so use a frozen object instead.

Also do some refactoring that turned out to be unneeded but eliminates some duplicated code.


tgoyne:
I think this is now down to just pre-existing issues.

Realm/RLMManagedSet.mm
```diff
@@ -507,32 +507,27 @@ - (BOOL)isFrozen {
...
 }
 
-- (instancetype)freeze {
-    if (self.frozen) {
-        return self;
-    }
-
-    RLMRealm *frozenRealm = [_realm freeze];
-    auto& parentInfo = _ownerInfo->resolve(frozenRealm);
+- (instancetype)resolveInRealm:(RLMRealm *)realm {
```
leemaguire, line 510: It would be nice to at some stage create a common home for all of these repeated methods in the collections

leemaguire (approved):
LGTM  👍🏼

Issue #7393, opened 2021-08-14
Warnings on realm-cocoa (version 5.5.2) when using Xcode13


nacho4d:
## Goals

Build realm-cocoa 5.5.2 (added via SPM) in Xcode13 without warnings

## Expected Results

No warnings at compile time

## Actual Results

See warnings:
<img width="1593" alt="Screen Shot 2021-08-14 at 16 42 34" src="https://user-images.githubusercontent.com/111651/129439180-16e76193-58f4-4749-bec2-375a3b526ae3.png">

Also see this related issue. It was closed however warnings still happens:
https://github.com/realm/realm-cocoa/issues/7295

## Steps for others to Reproduce

Just add the library via SPM and build.

## Version of Realm and Tooling

Realm framework version: 5.5.2

Xcode version: 13 beta5 (AFAIK problem happens since beta1)

iOS/OSX version: All?

Dependency manager + version: SPM

ericjordanmossman:
The changes mentioned in #7295 were released in 10.8.1.
We'd have to backport to v5, which is on a longer release cycle.

nacho4d:
@ericjordanmossman 
Yes please.

PR #7433, opened 2021-09-09
Fix v5 warning

ericjordanmossman:
Remove invalid source from Package.swift
See: https://github.com/realm/realm-cocoa/issues/7393


tgoyne (approved):
The CI failures here are because we no longer have those versions of Xcode installed, and it built successfully on the version which we do still have.

PR #7434, opened 2021-09-10
Fix the SwiftUI and Catalyst tests

tgoyne:
This turned out to be a whole pile of problems:

1. The project file had some weird incorrect dependencies such as SwiftUITestHostTests depending on SwiftUISyncTestHost and not RealmSwift. The build happened to work if things were built in the correct order but broke if you tried to run the tests directly from a clean state.
2. iOS 14.5 did some weird things to NavigationLink. There's a bug where if a List has exactly two NavigationLinks then following one will immediately pop back to the parent, and a simulator-specific bug where if the parent refreshes when you're 2+ NavigationLinks deep it reloads the intermediate View even if the ID matches. Since we need our UI test to work on the simulator, I worked around that by using keypath filtering to avoid refreshing the top-level view when editing in the detail view.
3. It turned out that keypath filtering with @ObservedResults didn't work. The fix was trivial, but this suggests that it's entirely untested.
4. `tap()` by default taps in the exact middle of the element. On iOS 14.5+, this meant that accessiblity taps on the NavigationLink in the top-level View happens to land on the TextField and switch to editing mode instead of going to the next View. I worked around this by setting the tap location to 0,0 which happens to work at the moment.
5. Each character typed in a TextField is setting the property *twice*. I'm not sure why, but I made createBinding() only perform a write if the value has actually changed, which makes TextFields a lot less laggy.
6. There were some half-implemented ideas in the test app (such as the Save button in the detail edit form) that I removed while trying to figure out what was going on and didn't restore afterwards.
7. Unrelated to this, using a test host app seemed to be causing problems for catalyst. When catalyst is running with the app hidden, catalyst itself seems to hang sometimes for 20+ seconds, which makes our CI jobs take a very long time and often outright time out. For whatever reason, this only happened when launching the tests via xcodebuild and not Xcode. Attaching Xcode to the app after launching it via xcodebuild would make the window for the app appear, and so also sidestep the problem. Using xctest rather than a test host app seems to fix this. We only actually need to use a test host for on-device testing, and only used it everywhere because that used to be simpler (but now isn't with how TEST_HOST is now set).

### DIFF
--- a/RealmSwift/SwiftUI.swift
+++ b/RealmSwift/SwiftUI.swift
@@ -26,20 +26,20 @@ import Realm.Private
 
 private func safeWrite<Value>(_ value: Value, _ block: (Value) -> Void) where Value: ThreadConfined {
     let thawed = value.realm == nil ? value : value.thaw() ?? value
-    var didStartWrite = false
-    if thawed.realm?.isInWriteTransaction == false {
-        didStartWrite = true
-        thawed.realm?.beginWrite()
-    }
-    block(thawed)
-    if didStartWrite {
-        try! thawed.realm?.commitWrite()
+    if let realm = thawed.realm, !realm.isInWriteTransaction {
+        try! realm.write {
+            block(thawed)
+        }
+    } else {
+        block(thawed)
     }
 }
 
 @available(iOS 13.0, macOS 10.15, tvOS 13.0, watchOS 6.0, *)
-private func createBinding<T: ThreadConfined, V>(_ value: T,
-                                                 forKeyPath keyPath: ReferenceWritableKeyPath<T, V>) -> Binding<V> {
+private func createBinding<T: ThreadConfined, V>(
+    _ value: T,
+    forKeyPath keyPath: ReferenceWritableKeyPath<T, V>) -> Binding<V> {
+
     guard let value = value.isFrozen ? value.thaw() : value else {
         throwRealmException("Could not bind value")
     }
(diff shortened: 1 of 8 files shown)


RealmSwift/SwiftUI.swift
```diff
@@ -471,6 +526,14 @@ extension Binding where Value: ObjectBase & ThreadConfined {
     public subscript<V>(dynamicMember member: ReferenceWritableKeyPath<Value, V>) -> Binding<V> where V: _Persistable {
         createBinding(wrappedValue, forKeyPath: member)
     }
+    /// :nodoc:
+    public subscript<V>(dynamicMember member: ReferenceWritableKeyPath<Value, V>) -> Binding<V> where V: _Persistable & RLMSwiftCollectionBase & ThreadConfined {
```
leemaguire, line 530: I thought that compounding protocol conformance like this was the undesired way to do it? and instead should be like:
```swift
where V: _Persistable, V: RLMSwiftCollectionBase, V: RLMSwiftCollectionBase ...
```

RealmSwift/SwiftUI.swift
```diff
@@ -48,19 +48,59 @@ private func createBinding<T: ThreadConfined, V>(_ value: T,
...
+    }, set: { newValue in
+        guard !value.isInvalidated else { return }
+        safeWrite(value) { value in
+            value[keyPath: keyPath] = newValue
         }
+    })
+}
+
+@available(iOS 13.0, macOS 10.15, tvOS 13.0, watchOS 6.0, *)
+private func createCollectionBinding<T: ThreadConfined, V: RLMSwiftCollectionBase & ThreadConfined>(
```
jsflax, line 63: Nice.

jsflax (approved):
👏

PR #7448, opened 2021-09-22
[Breaking] Unlock Self for observable types

jsflax:
This PR does a number of things. The downside of these things is that `observe` will be less discoverable in our docs, but it does consolidate a lot of code, makes our API more type safe, and removes functionality that didn't really work:
* Disallow key path observation on collections where the Element is not of Object type. Attempting to observe keyPaths on non-object  types yields no changes, so this functionality was already broken.
  * @tgoyne I am not sure how we feel about this. It breaks the API for those that were using it incorrectly.
*  Unlock `Self` for observable types. By moving object observation behind a private protocol, and collection observation
  cleaned up, key path filtering and Object/CollectionChanges have become more type safe. Previously, keyPath filtering
  was called as:
  ```swift
  person.observe(keyPaths: [\Person.name, \Person.age]) { change in /* do stuff */ }
  ```
  Now the class can be dropped, as it is derived from the calling type:
  ```swift
  person.observe(keyPaths: [\.name, \.age]) { change in /* do stuff */ }
  ```
  Non-keyPath filtering has also been enhanced. Previously:
  ```swift
  person.observe { change in
	  switch change {
	  case .initial(let object):
		// type is of ObjectBase and needs to be casted
		guard let person = object as? Person else { fatalError() }
		// do stuff
	  default: break
	  }
  }
  ```
  Now:
  ```swift
  person.observe { change in
	  switch change {
	  case .initial(let person):
		// person is already of `Person` type
	  }
  }
  ```
The above is all unlocked for collection observation as well.
* Consolidate observation logic to `RealmCollection`.


jsflax:
According to the CI, this seems to fail for Xcode < 13. I will have to investigate and potentially guard this behind directives.

jsflax:
Replicated the compiler error on Swift 5.4. It would _appear_ https://github.com/apple/swift-evolution/blob/main/proposals/0299-extend-generic-static-member-lookup.md enables the `Self` behaviour that this PR unlocks. So users on Xcode < 13 will not be able to take advantage of the generic type lookup for keyPath filtering, but they can still take advantage of `Object/CollectionChanges` passing back the fully typed object.

Issue #7450, opened 2021-09-23
Crash during migration when accessing ObjectSchema


michalrentka:
### How frequently does the bug occur?

All the time

### Description

Hi,

I migrated my Realm models to use `EmbeddedObject`s and I received some crash reports from this migration. 

I have a simple `safeGet(key: String) -> Any?` function which checks object schema whether it contains property with given name and then returns the value:
```
fileprivate func safeGet(key: String) -> Any? {
    guard self.objectSchema.properties.contains(where: { $0.name == key }) else {
        DDLogError("Database: trying to get '\(key)' but it's not in schema.")
        return nil
    }
    return self[key]
}
```

The app crashes when trying to access this object schema. Should I also check for `isInvalidated` flag before accessing the object or is this related to something else entirely?

### Stacktrace & log output

```shell
Incident Identifier: 6F7DC0D0-FD03-45EF-858B-37C2DBA9FC15
Hardware Model:      iPad13,11
Process:             Zotero [1598]
Path:                /private/var/containers/Bundle/Application/D4E24841-4108-455B-92F2-995F594ECA9E/Zotero.app/Zotero
Identifier:          org.zotero.ios.Zotero
Version:             1.0 (151)
AppStoreTools:       13A227
AppVariant:          1:iPad13,11:15
Beta:                YES
Code Type:           ARM-64 (Native)
Role:                unknown
Parent Process:      launchd [1]
Coalition:           org.zotero.ios.Zotero [923]

Date/Time:           2021-09-22 18:06:00.0597 -0700
Launch Time:         2021-09-22 18:05:58.0544 -0700
OS Version:          iPhone OS 15.0 (19A346)
Release Type:        User
Baseband Version:    2.09.10
Report Version:      104

Exception Type:  EXC_CRASH (SIGKILL)
Exception Codes: 0x0000000000000000, 0x0000000000000000
Exception Note:  EXC_CORPSE_NOTIFY
Triggered by Thread:  0


Kernel Triage:
VM - pmap_enter failed with resource shortage
VM - pmap_enter failed with resource shortage
VM - pmap_enter failed with resource shortage
VM - pmap_enter failed with resource shortage
VM - pmap_enter failed with resource shortage


Thread 0 name:
Thread 0 Crashed:
[...]

PR #7451, opened 2021-09-23
Add a basic test for the .objectSchema property on MigrationObject

tgoyne:
Prompted by #7450 as I couldn't find any tests that actually exercised that, and it's plausible that reading the schema from a DynamicObject could fail as they're kinda weird. Appears to work fine in the simple case, though.


RealmSwift/Tests/MigrationTests.swift
```diff
@@ -1023,6 +1028,31 @@ class MigrationTests: TestCase {
...
+            migration.enumerateObjects(ofType: "SwiftObject") { oldObj, newObj in
+                guard let oldObj = oldObj, let newObj = newObj else {
+                    XCTFail("did not get objects in enumerate")
+                    return
+                }
+                enumerated = true
+                XCTAssertEqual(oldObj.objectSchema.properties, properties)
+                XCTAssertEqual(newObj.objectSchema.properties, properties)
+            }
+            XCTAssertTrue(enumerated)
```
leemaguire, line 1052: Is it worth also testing `migration.oldSchema[type].properties.contains { $0.name == key }` after or before this block too?

leemaguire:
@tgoyne is it worth finishing this PR?

PR #7456, opened 2021-09-24
Fix change notifications from sync writes when using asyncOpen

### LINKED ISSUE #7447: RealmSwift: @ObservedResults not updating with Realm Sync

All the time

### Description

I’m trying to use Realm Sync for a basic CRUD app using SwiftUI, but for some reason the @ObservedResults property wrapper is not updating my view when I add/delete objects.

The Realm App is set up on the free shared tier. When I insert/delete entries, I can see in the Realm App logs that the data is being inserted/deleted appropriately in the collection. The problem is that the SwiftUI View (using the @ObservedResults wrapper) does not update appropriately. On the other hand, if I quit and restart the iOS app, any insertions/deletions I’ve made DO show up.

I brought this up on the forums and was asked to file an issue here.

### Stacktrace & log output

_No response_

Yes, always

### Reproduction Steps

Relevant portions of code:

```

let app = RealmSwift.App(id: "[XXXXXXXXX]")

class AppState: ObservableObject {
    @Published var userID: String?
}

@main
struct RepositoryRealm_App: SwiftUI.App {
    @StateObject var appState = AppState()
    var body: some Scene {
        WindowGroup {
            switch appState.userID {
            case nil:
                Text("XXXX")
                    .onAppear {
                        switch app.currentUser {
                        case nil:
                            app.login(credentials: .anonymous) { result in
                                DispatchQueue.main.async {
                                    switch result {
                                    case .failure(let error):
                                        fatalError(error.localizedDescription)
                                    case .success(let user):
                                        let partitionValue = user.id
                                        appState.userID = partitionValue
                                    }
                                }
                            }
                            
                        case let currentUser?:
                            let partitionValue = currentUser.id
                            appState.userID = partitionValue
                        }
                    }
                
            case let userID:
                ContentView()
                    .environmentObject(appState)
                    .environment(\.partitionValue, userID)
            }
        }
    }
}

struct ContentView: View {
    @AsyncOpen(appId: "[XXXXXXXXX]", partitionValue: "", timeout: 5000) var asyncOpen
    var body: some View {
        switch asyncOpen {
        case .connecting:
            Text("Connecting...")
        case .waitingForUser:
            Text("Waiting for user...")
        case .open(let userRealm):
            EntryObjectListView()
                .environment(\.realm, userRealm)
        case .progress(let progress):
            ProgressView(progress)
        case .error(let error):
            fatalError(error.localizedDescription)
        }
    }
}

struct EntryObjectListView: View {
    @EnvironmentObject var appState: AppState
    @ObservedResults(EntryObject.self) var entryObjects
    
    var body: some View {
        NavigationView {
            VStack {

                List {
                    ForEach(entryObjects) { entryObject in
                        Text(entryObject.title)
                    }
                    .onDelete(perform: $entryObjects.remove)
                }
                Spacer()
                Button("add text", systemImage: "plus") {

                    let entryObject = EntryObject(dataType: TextData.self)
                    entryObject.title = "hello text!"
                    entryObject._partition = appState.userID!

                    $entryObjects.append(entryObject)
                }
                Button("delete all", systemImage: "minus") {
                    withAnimation {
                        for entryObject in entryObjects {
                            $entryObjects.remove(entryObject)
                        }
                    }
                }
                Button("print count", systemImage: "minus") {
                    print(entryObjects.count)
                }
            }
        }
    }
}

final class EntryObject: Object, ObjectKeyIdentifiable {
    
    @Persisted(primaryKey: true) var _id = UUID()
    @Persisted var _partition: String = ""
    @Persisted var type: EntryType = .undefined
    @Persisted var dateCreated: Date = Date()
    @Persisted var dateModified: Date = Date()
    @Persisted var title: String
    
    @Persisted var textData: TextData?
    @Persisted var urlData: URLData?
    @Persisted var objectData: ObjectData?

// Some convenience methods / inits omitted
}
```

### Version

Realm 10.15.1

MongoDB Realm (i.e. Sync, auth, functions)

### Are you using encryption?

No, not using encryption

iOS 14.7.1

### Build environment

Xcode version: 13.0

### LINKED ISSUE #7453: Sync realm Result.observe is not fired by remote change but on next write

All the time

### Description

Using the tutorial task tracker app, the expected sync behavior (which is observed on the .NET and Android tutorial apps) is that when i make a change to a task in the backend the observe callback is called immediately on the client iOS SDK Results<Task> 

What actually happens is that the observe callback is not called immediately. However if I change the status of another task on the iOS client, I get all the previous expected observe callbacks in one burst which seems to indicate that the sync occurred correctly but the callback was not called. 

I might be doing something wrong as I'm not too familiar with the system but this is unchanged tutorial app, tried waiting for realm sync session connection and different ways of opening the realms but still observe the same behavior.   

### Stacktrace & log output

_No response_

Yes, always

### Reproduction Steps

_No response_

### Version

Realm 10.15.1 RealmDatabase 11.4.1

MongoDB Realm (i.e. Sync, auth, functions)

### Are you using encryption?

No, not using encryption

iPadOS 15.0

### Build environment

Xcode version: 13.0
Dependency manager and version: Realm 10.15.1 RealmDatabase 11.4.1

tgoyne:
Destroying and then immediately recreating a RealmCoordinator with an active sync session can result in the new RealmCoordinator using the old SyncSession, which almost entirely works other than that it fails to set the sync notification callback correctly. That bug needs to be fixed in core, but we don't want to be destroying and recreating the RealmCoordinator in asyncOpen anyway as that has a lot of overhead.

Fixes #7447. Fixes #7453. Fixes https://github.com/realm/realm-core/issues/4909.

### DIFF
--- a/Realm/RLMRealm.mm
+++ b/Realm/RLMRealm.mm
@@ -257,7 +257,7 @@ static RLMAsyncOpenTask *openAsync(RLMRealmConfiguration *configuration, void (^
 + (RLMAsyncOpenTask *)asyncOpenWithConfiguration:(RLMRealmConfiguration *)configuration
                                    callbackQueue:(dispatch_queue_t)callbackQueue
                                         callback:(RLMAsyncOpenRealmCallback)callback {
-    return openAsync(configuration, [=](ThreadSafeReference, std::exception_ptr err) {
+    return openAsync(configuration, [=](ThreadSafeReference ref, std::exception_ptr err) {
         @autoreleasepool {
             if (err) {
                 try {
@@ -272,12 +272,18 @@ static RLMAsyncOpenTask *openAsync(RLMRealmConfiguration *configuration, void (^
                 }
                 return;
             }
-            dispatch_async(callbackQueue, ^{
+            // Ensure that we keep the Realm open until we've initialized the
+            // RLMRealm on the target queue. This ensures that the existing
+            // RealmCoordinator and DB are reused rather than being reopened.
+            // We can't just capture the ThreadSafeReference as dispatch_async()
+            // requires a copyable block.
+            dispatch_async(callbackQueue, [=, ref = ref.resolve<std::shared_ptr<Realm>>(nullptr)]() mutable {
                 @autoreleasepool {
                     NSError *error;
                     RLMRealm *localRealm = [RLMRealm realmWithConfiguration:configuration
                                                                       queue:callbackQueue
                                                                       error:&error];
+                    ref.reset();
                     callback(localRealm, error);
                 }
             });
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,11 @@ x.y.z Release notes (yyyy-MM-dd)
 
 ### Fixed
 * <How to hit and notice issue? what was the impact?> ([#????](https://github.com/realm/realm-cocoa/issues/????), since v?.?.?)
-* None.
+* A race condition in Realm.asyncOpen() sometimes resulted in subsequent writes
+  from Realm Sync failing to produce notifications
+  ([#7447](https://github.com/realm/realm-cocoa/issues/7447),
+  [#7453](https://github.com/realm/realm-cocoa/issues/7453),
+  [Core #4909](https://github.com/realm/realm-core/issues/4909), since v10.15.0).
 
 <!-- ### Breaking Changes - ONLY INCLUDE FOR NEW MAJOR version -->
 
--- a/Realm/ObjectServerTests/RLMObjectServerTests.mm
+++ b/Realm/ObjectServerTests/RLMObjectServerTests.mm
@@ -831,6 +831,35 @@ static NSString *randomEmail() {
     }
 }
 
+- (void)testIncomingSyncWritesTriggerNotifications {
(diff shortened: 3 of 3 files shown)


Realm/RLMRealm.mm
```diff
@@ -272,12 +272,18 @@ + (RLMAsyncOpenTask *)asyncOpenWithConfiguration:(RLMRealmConfiguration *)config
                 }
                 return;
             }
-            dispatch_async(callbackQueue, ^{
+            // Ensure that we keep the Realm open until we've initialized the
+            // RLMRealm on the target queue. This ensures that the existing
+            // RealmCoordinator and DB are reused rather than being reopened.
+            // We can't just capture the ThreadSafeReference as dispatch_async()
+            // requires a copyable block.
+            dispatch_async(callbackQueue, [=, ref = ref.resolve<std::shared_ptr<Realm>>(nullptr)]() mutable {
```
jsflax, line 280: Why are we resolving with a nullptr payload?

tgoyne: The argument there is the realm to resolve the TSR in, which isn't applicable when resolving a TSR to a realm.

jsflax:
Lgtm

PR #7469, opened 2021-10-05
Add UserProfile

jsflax:


Realm/RLMUser.mm
```diff
@@ -323,3 +330,78 @@ - (instancetype)initUserIdentityWithProviderType:(NSString *)providerType
...
+@interface RLMUserProfile () {
+    SyncUserProfile _userProfile;
+}
+@end
+
+@implementation RLMUserProfile
+
+- (instancetype)initWithUserProfile:(SyncUserProfile)userProfile {
+    if (self = [super init]) {
+        _userProfile = userProfile;
```
tgoyne, line 345: This should move the argument.

Realm/RLMUser.mm
```diff
@@ -323,3 +330,78 @@ - (instancetype)initUserIdentityWithProviderType:(NSString *)providerType
...
+    }
+    return self;
+}
+
+- (NSString *)name {
+    _userProfile.data();
+    if (_userProfile.name() == util::none) {
+        return nil;
+    }
+    return @(_userProfile.name()->c_str());
```
tgoyne, line 355: This is a lot of duplicated logic which could easily be extracted to a function.

PR #7487, opened 2021-10-18
Change configuration logic in ObservedResults

leemaguire:
In this PR I'm omitting the condition to check if the stored realm config in `Storage` matches the config stored in the SwiftUI environment values. 

This is because a user can pass a realm configuration to `@ObservedResults` and that Realm config will always be overridden  by whatever config is stored in the environment value.

This change may only affect a user who is passing their config via `.environment(\.realmConfiguration, ...)` in a dynamic fashion on a view update.

May fix: https://github.com/realm/realm-cocoa/issues/7463

### DIFF
--- a/RealmSwift/SwiftUI.swift
+++ b/RealmSwift/SwiftUI.swift
@@ -603,7 +603,7 @@ extension Projection: _ObservedResultsValue { }
     public mutating func update() {
         // When the view updates, it will inject the @Environment
         // into the propertyWrapper
-        if storage.configuration == nil || storage.configuration != configuration {
+        if storage.configuration == nil {
             storage.configuration = configuration
         }
     }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,8 @@ x.y.z Release notes (yyyy-MM-dd)
 ### Fixed
 * Adding a Realm Object to a `ObservedResults` or a collections using `StateRealmObject` that is managed by the same Realm 
   would throw if the Object was frozen and not thawed before hand.
+* Setting a Realm Configuration for @ObservedResults using it's initializer would be overrode by the Realm Configuration stored in
+  `.environment(\.realmConfiguration, ...)` if they did not match ([Cocoa #7463](https://github.com/realm/realm-swift/issues/7463), since v10.6.0).
 
 <!-- ### Breaking Changes - ONLY INCLUDE FOR NEW MAJOR version -->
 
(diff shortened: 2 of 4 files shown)


jsflax:
@leemaguire

bmunkholm:
Missing tests and changelog update

RealmSwift/SwiftUI.swift
```diff
@@ -603,7 +603,7 @@ extension Projection: _ObservedResultsValue { }
     public mutating func update() {
         // When the view updates, it will inject the @Environment
         // into the propertyWrapper
-        if storage.configuration == nil || storage.configuration != configuration {
```
jsflax, line 606: Why did this fix it?

leemaguire: If the configuration stored in `.environment(\.realmConfiguration, ...)`did not match a config that was passed to the initialiser it would override `storage.configuration` with whatever the `.environment` stored, which is likely just a default realm configuration.

jsflax (approved):

RealmSwift/SwiftUI.swift
dianaafanador3: Can we just add the schema to the Equatable function?, this will probably reset the configuration every time we render the view.

jsflax: Good point Diana.

dianaafanador3: Yep, you're right

tgoyne: Is there any way for the configuration from the Environment to change between calls to `update()` without a new View being constructed? The documentation implies it's possible (it says that `update()` is called each time the value for the environment key changes), but I don't see any way for it to actually happen.

leemaguire: I can't see a way to update the value without it reinitialising the entire view (using `print(Self._printChanges())` to debug that). The docs also mention `You can use this property wrapper to read — but not set` which is kinda contradictory to `If the value changes, SwiftUI updates any parts of your view that depend on the value`. Maybe this means if you modify from the top level with something like?

```swift
    @State var changeConfig = false

    var body: some Scene {
        WindowGroup {
            ContentView(changeConfig: $changeConfig)
                .environment(\.realmConfiguration, changeConfig ? Realm.Configuration(inMemoryIdentifier: "foo") : Realm.Configuration(inMemoryIdentifier: "bar"))
        }
    }
```

tgoyne (approved):

PR #7488, opened 2021-10-19
Upgrade to core 11.6.0

tgoyne:
Requires https://github.com/realm/realm-core/pull/4972 first.


tgoyne:
Also needs https://github.com/realm/realm-core/pull/5008.

PR #7504, opened 2021-10-29
Make Realm async init @MainActor

tgoyne:
It currently doesn't appear to be possible for async init to work without being confined to a global actor. That global actor doesn't need to be MainActor, but there also doesn't appear to be a way for it to be a user-supplied one and MainActor is probably the only one people care about for now.

I'm calling this a bug fix instead of a breaking change because the thing we're currently shipping doesn't actually work.

### DIFF
--- a/Realm/ObjectServerTests/RLMSyncTestCase.mm
+++ b/Realm/ObjectServerTests/RLMSyncTestCase.mm
@@ -591,7 +591,7 @@ static NSURL *syncDirectoryForChildProcess() {
     NSAssert(session, @"Cannot call with invalid partition value");
     XCTestExpectation *ex = expectation ?: [self expectationWithDescription:@"Wait for download completion"];
     __block NSError *theError = nil;
-    BOOL queued = [session waitForDownloadCompletionOnQueue:nil callback:^(NSError *err) {
+    BOOL queued = [session waitForDownloadCompletionOnQueue:dispatch_get_global_queue(0, 0) callback:^(NSError *err) {
         theError = err;
         [ex fulfill];
     }];
@@ -610,7 +610,7 @@ static NSURL *syncDirectoryForChildProcess() {
     NSAssert(session, @"Cannot call with invalid Realm");
     XCTestExpectation *ex = [self expectationWithDescription:@"Wait for upload completion"];
     __block NSError *completionError;
-    BOOL queued = [session waitForUploadCompletionOnQueue:nil callback:^(NSError *error) {
+    BOOL queued = [session waitForUploadCompletionOnQueue:dispatch_get_global_queue(0, 0) callback:^(NSError *error) {
         completionError = error;
         [ex fulfill];
     }];
--- a/Realm/ObjectServerTests/SwiftObjectServerTests.swift
+++ b/Realm/ObjectServerTests/SwiftObjectServerTests.swift
@@ -755,8 +755,6 @@ class SwiftObjectServerTests: SwiftSyncTestCase {
 
     // MARK: - App tests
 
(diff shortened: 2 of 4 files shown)


Realm/ObjectServerTests/RLMSyncTestCase.mm
```diff
@@ -591,7 +591,7 @@ - (void)waitForDownloadsForUser:(RLMUser *)user
     NSAssert(session, @"Cannot call with invalid partition value");
     XCTestExpectation *ex = expectation ?: [self expectationWithDescription:@"Wait for download completion"];
     __block NSError *theError = nil;
-    BOOL queued = [session waitForDownloadCompletionOnQueue:nil callback:^(NSError *err) {
+    BOOL queued = [session waitForDownloadCompletionOnQueue:dispatch_get_global_queue(0, 0) callback:^(NSError *err) {
```
tgoyne, line 594: These are because running the runloop (via `waitForExpectations`) in an async context on the main thread seems to not dispatch the main queue, so these need to be delivered to a different queue.

Realm/ObjectServerTests/SwiftObjectServerTests.swift
```diff
@@ -765,6 +763,8 @@ class SwiftObjectServerTests: SwiftSyncTestCase {
     }
 
     func testAppInit() {
+        let appName = "translate-utwuv"
```
jsflax, line 766: Any particular reason this was made local?

tgoyne: This test is the only place it's used so there's no reason for it to have a larger scope.

Realm/ObjectServerTests/SwiftObjectServerTests.swift
```diff
@@ -2148,157 +2148,97 @@ class AsyncAwaitObjectServerTests: SwiftSyncTestCase {
...
-                                            downloadBeforeOpen: .always)
-                XCTAssertEqual(realm.objects(SwiftHugeSyncObject.self).count, 2)
-            case .second:
-                XCTAssertTrue(Realm.fileExists(for: user.configuration(testName: #function)))
-                let realm = try await Realm(configuration: user.configuration(testName: #function),
-                                            downloadBeforeOpen: .always)
-                XCTAssertEqual(realm.objects(SwiftHugeSyncObject.self).count, 4)
-            }
-        }
+    @MainActor func testAsyncOpenSync() async throws {
```
jsflax, line 2151: I guess multiple users makes it easier to test than multiple processes 😅 .

RealmSwift/Realm.swift
```diff
@@ -1057,17 +1054,20 @@ extension Realm {
                 fallthrough
             }
         case .always:
-            try await withCheckedThrowingContinuation { (continuation: CheckedContinuation<Void, Swift.Error>) in
-                RLMRealm.asyncOpen(with: configuration.rlmConfiguration, callback: { error in
+            rlmRealm = try await withCheckedThrowingContinuation { continuation in
+                RLMRealm.asyncOpen(with: configuration.rlmConfiguration, callbackQueue: .main) { (realm, error) in
```
jsflax, line 1058: Is specifying the callback queue needed here?

tgoyne: That is what makes it so that we can use the RLMRealm instance from the asyncOpen rather than obtaining a new one.

jsflax: Right, I thought `@MainActor` would've covered this for us, but fair enough.

jsflax (approved):

PR #7505, opened 2021-10-30
Conform `@ThreadSafe` and `ThreadSafeReference` to `Sendable`

jsflax:
Title is self explanatory.

To make `ThreadSafeReference` conform to `Sendable`, its underlying properties need to be `Sendable` as well. We probably could have gotten away with using `unchecked Sendable` (which is what we're doing for `@ThreadSafe`), but this seemed like a good opportunity to clean up some of the top level (but internal) gnarly bits of our TSR implementation.

Conforms `RLMSwiftCollectionBase` to `AssistedObjectiveCBridgeable`. This deletes the need for each `RealmCollection` to conform to `AssistedObjectiveCBridgeable` since they all bridge to `RLMCollection` anyway. It does force us to implement a previously unimplemented initialiser on inheriting types– but it should be noted that these initialisers were already callable and would [abort](https://stackoverflow.com/questions/38334776/fatal-error-use-of-unimplemented-initializer-in-custom-navigationcontroller) if called.

Removes the metadata field in `AssistedObjectiveCBridgeable` calls. This field was originally added to be able to retain the `RealmCollection` type information for `AnyRealmCollection` types. Since `AnyRealmCollection`'s bridged type is an `RLMCollection`, we can instead deduce the type from that value: if `RLMCollection` is an `RLMArray`, it can be deduced that the underlying type is a List. For `MutableSet`, RLMSet, and `LinkingObjects` and `Results, `RLMResults`.

Makes `@ThreadSafe` `final` to conform to `@unchecked Sendable`.


RealmSwift/RealmCollection.swift
```diff
@@ -1236,6 +1235,17 @@ public struct AnyRealmCollection<Element: RealmCollectionValue>: RealmCollection
...
+        if let rlmCollection = rlmCollection as? RLMArray<AnyObject> {
+            self.base = _AnyRealmCollection(base: List<Element>(objc: rlmCollection))
+        } else if let rlmCollection = rlmCollection as? RLMSet<AnyObject> {
+            self.base = _AnyRealmCollection(base: MutableSet<Element>(objc: rlmCollection))
+        } else if let rlmCollection = rlmCollection as? RLMResults<AnyObject> {
+            self.base = _AnyRealmCollection(base: Results<Element>(objc: rlmCollection))
+        } else {
+            preconditionFailure()
+        }
+    }
```
ericjordanmossman, line 1248: Linking objects also extends RealmCollection. Does it need to be here?

RealmSwift/EmbeddedObject.swift
```diff
@@ -272,11 +272,11 @@ private func forceCastToInferred<T, V>(_ x: T) -> V {
...
 extension EmbeddedObject: AssistedObjectiveCBridgeable {
-    static func bridging(from objectiveCValue: Any, with metadata: Any?) -> Self {
+    static func bridging(from objectiveCValue: Any) -> Self {
         return forceCastToInferred(objectiveCValue)
     }
 
-    var bridged: (objectiveCValue: Any, metadata: Any?) {
-        return (objectiveCValue: unsafeBitCast(self, to: RLMObject.self), metadata: nil)
+    var bridged: Any {
+        return unsafeBitCast(self, to: RLMObject.self)
```
ericjordanmossman, line 280: Just wondering:
Is there a difference between this and unsafeCastToRLMObject?

RealmSwift/RealmCollection.swift
jsflax: `LinkingObjects` is bridged to `RLMCollection` as an `RLMResults`.

RealmSwift/EmbeddedObject.swift
jsflax: I don't think so anymore.

ericjordanmossman:
Is there a way to test this?

jsflax:
Not really (yet). The compilation checks won't be included until until Swift 6.0. And our existing tests test the new TSR implementation.

ericjordanmossman (approved):
If that's the case about testing, then LGTM.

tgoyne (approved):

PR #7510, opened 2021-11-03
Allow collections of all types in migrations.

leemaguire:
Previously when you accessed a collection inside a migration it would only cast to `List<DynamicObject>`. This caused obvious issues because collections can be of a variety of types inside a migration. This PR addresses the issue by looking at what the underlying type the `RLMArray`/`RLMSet`/`RLMMap` is and then casting to the appropriate type.


RealmSwift/Object.swift
```diff
@@ -542,13 +542,13 @@ public final class DynamicObject: Object {
...
             if let array = value as? RLMArray<AnyObject> {
-                return List<DynamicObject>(objc: array)
+                return list(from: array)
             }
             if let set = value as? RLMSet<AnyObject> {
-                return MutableSet<DynamicObject>(objc: set)
+                return mutableSet(from: set)
             }
             if let dictionary = value as? RLMDictionary<AnyObject, AnyObject> {
-                return Map<String, DynamicObject>(objc: dictionary)
```
leemaguire, line 551: This is an issue, Map only supports optional objects. So changing this here and on `dynamicMap` will be a breaking change.

jsflax (approved):

ericjordanmossman (approved):

PR #7520, opened 2021-11-09
Add history flag for writeCopy

ericjordanmossman:
Introduces a flag on `writeCopy` which allows a copied realm to include sync history. This allows a copied realm to synchronize data. 

Introduces a seedFilePath option on `RLMRealmConfiguration` and `Realm.Configuration`. If this 
option is set then instead of creating an empty, the realm at the `seedFilePath` will 
be copied to the `fileURL` of the new realm.

This enables use cases like bundling a synchronized realm with an application.


Realm/ObjectServerTests/SwiftObjectServerTests.swift
```diff
@@ -692,6 +692,137 @@ class SwiftObjectServerTests: SwiftSyncTestCase {
...
+        do {
+            let user1 = try logInUser(for: basicCredentials())
+            if !isParent {
+                populateRealm(user: user1, partitionValue: #function)
+                return
+            }
+
+            // Wait for the child process to upload all the data.
+            // DELETE COMMENT: Does this need to be a child process? I doubt it.
+            executeChild()
```
ericjordanmossman, line 707: Does this need to be a different process?

Realm/RLMRealm.mm
```diff
@@ -944,16 +945,24 @@ - (RLMObject *)createObject:(NSString *)className withValue:(id)value {
...
     NSString *path = fileURL.path;
 
     try {
-        _realm->verify_thread();
-        try {
-            _realm->read_group().write(path.UTF8String, static_cast<const char *>(key.bytes));
-        }
-        catch (...) {
-            _impl::translate_file_exception(path.UTF8String);
+        if (shouldSync) {
```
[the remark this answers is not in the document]
leemaguire: It does not make sense to wait for uploads because the Realm is always currently up to date (w.r.t data it is going to upload). The only argument is that we should wait for uploads to make sure that the sync history is full up to date before copy, but this would then make `writeCopy` a blocking call and would require network access.

Realm/ObjectServerTests/SwiftObjectServerTests.swift
```diff
@@ -692,6 +692,137 @@ class SwiftObjectServerTests: SwiftSyncTestCase {
...
+            XCTFail("Got an error: \(error) (process: \(isParent ? "parent" : "child"))")
+        }
+    }
+
+    // Bundling a sync realm and opening it with a different partition isn't supported
+    // But writeCopy takes a file path, not a configuration. So in theory someone could copy a realm, then
+    // open the new path with a configuration that has a different partition.
+    // There's no way to check if partitions
+    // are different when writeCopy is called with .writeCopy alone(?)
+//    func testWriteCopyDifferentPartition() {
```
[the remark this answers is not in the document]
leemaguire: A realm should be able to be copied with any arbitrary file name. If another sync user choses to use the copied synced realm then they need to specify the `fileUrl` in the Realm configuration.

Realm/ObjectServerTests/SwiftObjectServerTests.swift
leemaguire: Executing this is a separate process isn't doing much by the way of testing `writeCopy`. What this is doing is just syncing the data from another user and then coping the Realm file.

Realm/RLMRealm.h
```diff
@@ -225,20 +225,41 @@ NS_ASSUME_NONNULL_BEGIN
...
+ the *current* data is written, not the data from the point when the previous write
  transaction was committed.
 
  @param fileURL Local URL to save the Realm to.
  @param key     Optional 64-byte encryption key to encrypt the new file with.
- @param error   If an error occurs, upon return contains an `NSError` object
- that describes the problem. If you are not interested in
- possible errors, pass in `NULL`.
+ @param error  If an error occurs, upon return contains an `NSError` object
+              that describes the problem. If you are not interested in
```
leemaguire, line 235: Fix indentation

Realm/RLMRealm.h
```diff
@@ -225,20 +225,41 @@ NS_ASSUME_NONNULL_BEGIN
...
+ Note that if this method is called from within a write transaction,
+ the *current* data is written, not the data from the point when the previous write
+ transaction was committed.
+
+ @param fileURL Local URL to save the Realm to.
+ @param key     Optional 64-byte encryption key to encrypt the new file with.
+ @param enableHistory  Indicates whether the copied Realm will have sync history enabled
+                      and synchronize data with the Realm server. All changes from the
+                      original realm must be uploaded to the server prior to calling
+                      `writeCopyToURL`, otherwise and error is thrown. If `enableHistory`
```
leemaguire, line 256: typo

Realm/RLMRealm.mm
```diff
@@ -944,16 +945,24 @@ - (RLMObject *)createObject:(NSString *)className withValue:(id)value {
...
 
     try {
-        _realm->verify_thread();
-        try {
-            _realm->read_group().write(path.UTF8String, static_cast<const char *>(key.bytes));
-        }
-        catch (...) {
-            _impl::translate_file_exception(path.UTF8String);
+        if (enableHistory) {
+            _realm->write_copy(path.UTF8String, {static_cast<const char *>(key.bytes), key.length});
```
leemaguire, line 957: If this is called on a different thread what happens?

Realm/RLMRealm.h
```diff
@@ -225,20 +225,41 @@ NS_ASSUME_NONNULL_BEGIN
...
+                      is false, the copied Realm functions as a local realm.
+ @param error  If an error occurs, upon return contains an `NSError` object
+              that describes the problem. If you are not interested in
+              possible errors, pass in `NULL`.
+ */
+- (BOOL)writeCopyToURL:(NSURL *)fileURL encryptionKey:(nullable NSData *)key enableHistory:(BOOL)enableHistory error:(NSError **)error;
```
tgoyne, line 262: "History" as a concept is sort of an implementation detail. The thing the user cares about is that the end result can be opened as a sync Realm, not the specific details about what's different about sync/non-sync Realms.

I don't have great ideas for what the argument name should be for that though. The obvious things like `enableSync:` or `synchronizedRealm:` seem bad.

tgoyne (commented):
A separate function which waits for upload completion, writes a copy, then calls a completion handler might make sense. The requirement that all of the local history be uploaded is actually fairly tricky to handle and something that would be nice to solve for the user.

Solving this shouldn't be part of this PR, but currently it's also a bit difficult to actually use the copy you made. We probably need something like Java's feature where you can set a bundled Realm path in your config that'll be copied to the appropriate location on open if we don't already have a local Realm.

Realm/RLMRealm.h
nirinchev: I guess `writeSynchronizableCopyToURL` is a mouthful, but also reasonably descriptive.

ericjordanmossman:
> We probably need something like Java's feature where you can set a bundled Realm path in your config that'll be copied to the appropriate location on open if we don't already have a local Realm.

So this would be a configuration option, on the file that is to be copied? Would this only be used for local realms? Since sync realm URLs are generated differently?

Realm/ObjectServerTests/SwiftObjectServerTests.swift
jsflax: Correct, this can just be a multi user test.

jsflax:
This is being reworked.

ericjordanmossman:
I tried creating something similar to `assetFile` in realm-java but felt the extra code wasn’t necessarily worth the benefit. 

- I couldn’t figure out a good way without changing the hot path `realmWithConfiguration`. 
- There was some weirdness with toggling back and forth between in-memory. 
- It would obscure a developer’s ability to wait for uploads, should they need.
- Didn't believe the immediate usefulness justified adding to public API of RLMRealm.

Most of what is necessary is just the ease of access to the path the realm was copied to. So I wanted to expose the ability to tell where a realm would be opened before it is opened. But this is already available after constructing a sync configuration and before opening the realm, in `syncConfig.fileURL`.

Realm/RLMRealm.h
ericjordanmossman: In the latest commit I did go with `enableSync` param. I think a new method `writeSyncCopyToURL` might be alright.
What do you think?

ericjordanmossman:
Needs changelog

tgoyne:
`syncConfig.fileURL` is not public. Even if it was, we'd just be asking developers to copy/paste some boilerplate code to do something we could just be taking care of for them.

I don't understand the relevance of in-memory Realms or why this'd have any effect on waiting for uploads. The non-cache hit path of realmWithConfiguration isn't a hot-path; it's actually pretty expensive to open a Realm from scratch already.

ericjordanmossman:
`config.fileURL` contains the path of the sync realm.
The `inMemoryIdentifier` option is relevant because it's not compatible with an assetFile path. Switching a configuration to and from inMemoryIdentifier would not return back to the previous state easily.

tgoyne:
Clearing the asset file path when a configuration is set to in-memory and not restoring it when the configuration is set to a non-in-memory mode is how everything else on RLMConfiguration works.

jsflax:
@tgoyne I don't really agree with adding an assetFileURL to the config. It doesn't save the user much boilerplate and adds confusion– notably because they will not be able to use the same configuration (containing the assetFileURL) to use their new Realm.

Considering writeCopy is not a hot path, I do not see an issue with this just being a flag in writeCopy (albeit not named `history` since that exposes an implementation a detail). So the steps would be:
1. Open asset realm A.
2. Write Copy asset realm A to destination 1.
3. Open realm at destination 1 with configuration with fileURL pointing to destination 1.

Why complicate that?

tgoyne:
That's not how you use bundled Realms. You use writeCopy() to produce the Realm which you will include in your app bundle, not to open the bundled Realm and copy it to the writeable location. The full flow is:

1. Either use Studio or writeCopy() in a simulator to produce a Realm with no client ID.
2. Move that file into your app's project and add it to the bundled files.
3. On app startup, construct your app's Realm.Configuration normally.
4. Do `if !Realm.fileExists(for: config) { try! FileManager.default.copy(from: Bundle.main.url(forResource: "bundledRealm", ext: "realm")!, config.fileURL) }`
5. Open the Realm normally

Java's `assetURL` just replaces step 4 with `config.assetURL = Bundle.main.url(forResource: "bundledRealm", ext: "realm")`. We didn't bother adding it when they did because it is just one line of code, but that one line of code is incorrect (it'll crash and/or corrupt data if another thread tries to open the Realm while the copy is happening) and it seems sort of silly to have some code in the documentation for people to copy/paste when we could just take care of it for them.

jsflax:
That makes sense, I assumed writeCopy was being used in the same process. `assetURL` makes sense given what you're saying, but the name is confusing. An asset is congruous to a file so having `assetURL` *and* `fileURL` seem super error prone. I'm open to better names.

tgoyne:
Yeah, I don't like the name very much.

ericjordanmossman:
Ready for review again.
Added the ability to set a file path that will be copied.

I went with `seedFilePath`.
Other options included:
`assetFile`: To confusing with `fileURL`. Maybe `assetFile` would work better if we didn't already have `fileURL`
`assetFilePath`
`inputFilePath`
`originFilePath`
`copyFilePath`

RealmSwift/RealmConfiguration.swift
```diff
@@ -276,6 +294,7 @@ extension Realm {
             configuration.setCustomSchemaWithoutCopying(self.customSchema)
             configuration.disableFormatUpgrade = self.disableFormatUpgrade
             configuration.maximumNumberOfActiveVersions = self.maximumNumberOfActiveVersions ?? 0
+            configuration.seedFilePath = self.seedFilePath
```
ericjordanmossman, line 297: This will nil out inMemoryIdentifier when you don't want it to be nil.

Realm/Tests/RealmTests.mm
```diff
@@ -1871,6 +1871,67 @@ - (void)testWritingCopyUsesWriteTransactionInProgress
...
+    // Create a seed realm with two objects
+    RLMRealmConfiguration *seedConfig = [RLMRealmConfiguration defaultConfiguration];
+    seedConfig.fileURL = RLMTestRealmURL();
+    RLMRealm *realm = [RLMRealm realmWithConfiguration:seedConfig error:nil];
+    [realm beginWriteTransaction];
+    [IntObject createInRealm: realm withValue:@[@1]];
+    [IntObject createInRealm: realm withValue:@[@2]];
+    [realm commitWriteTransaction];
+
+    NSURL *fileURL = [NSURL fileURLWithPath:RLMRealmPathForFile(@"filename.realm")];
```
leemaguire, line 1886: I ran this test and I didn't see `filename.realm` at the file path.

Realm/Tests/RealmTests.mm
```diff
@@ -1871,6 +1871,67 @@ - (void)testWritingCopyUsesWriteTransactionInProgress
...
```
leemaguire, line 1911: Again, `filename.realm` doesn't exist before its removed.

Realm/ObjectServerTests/SwiftObjectServerTests.swift
```diff
@@ -692,6 +692,94 @@ class SwiftObjectServerTests: SwiftSyncTestCase {
...
+            config.objectTypes = [SwiftHugeSyncObject.self]
+            let realm = try Realm(configuration: config)
+            try! realm.write {
+                for _ in 0..<SwiftSyncTestCase.bigObjectCount {
+                    realm.add(SwiftHugeSyncObject.create())
+                }
+            }
+            waitForUploads(for: realm)
+
+            try realm.writeCopy(toFile: RLMTestRealmURL(), enableSync: true)
```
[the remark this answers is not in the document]
ericjordanmossman: You would receive an exception:
```
Incompatible histories. Expected a Realm with no or in-realm history, but found history type 3
```

Realm/Tests/RealmTests.mm
ericjordanmossman: The removal is just extra test assurance that there is no file at the path at that point in time. The writeCopy that occurs later is expecting there to be no file. I added lines asserting this explicitly.

Realm/Tests/RealmTests.mm
ericjordanmossman: Similar to the other test, except these ones can be removed.

Realm/Tests/RealmTests.mm
ericjordanmossman: I guess the assert is all that's needed. The removal is unnecessary.

Realm/RLMRealm.mm
ericjordanmossman: I believe it functions similarly. Ultimately the file is written in the format. And once that file is accessed, what thread it was written from is irrelevant. It's possible a developer could run into an error like calling writeCopy before all changes are synced to the server, or accessing a realm from the wrong thread.

I changed up `testSeedFilePathOpen` to write copy from a background thread:
```
            let queue = DispatchQueue(label: "background")
            queue.async {
                autoreleasepool {
                    do {
                        try realm.writeCopy(toFile: RLMTestRealmURL(), enableSync: true)
                    } catch {
                        XCTFail("Got an error: \(error)")
                    }
                }
            }
            queue.sync { }
```

I can keep it this way. I wrote another `testWriteCopyDifferentThread` but it was substantially similar testSeedFilePathOpen, and they take a bit of time to run. We could have `testSeedFilePathOpen ` just do the write copy from a background thread.

Realm/Tests/RealmTests.mm
ericjordanmossman: I removed the "removeItemAtPath", kept the asserts.

Realm/ObjectServerTests/SwiftObjectServerTests.swift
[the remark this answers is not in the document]
leemaguire: Can we test this and try give a better exception back to the user?

Realm/Tests/RealmTests.mm
```diff
@@ -1871,6 +1871,65 @@ - (void)testWritingCopyUsesWriteTransactionInProgress
...
+    assert(![[NSFileManager defaultManager] fileExistsAtPath:fileURL.path]);
+
+    // Create a realm at destination path; add one object
+    RLMRealmConfiguration *configuration = [RLMRealmConfiguration defaultConfiguration];
+    configuration.fileURL = fileURL;
+    {
```
leemaguire, line 1916: The objects here in this scope won't necessarily be released on exit of the scope. So use `@autorelesasepool`

Realm/RLMRealm.mm
leemaguire: The above will succeed because it was opened on the background thread in the first place. My concern is do we need to do similar with the non `enableSync` route and prepend `_realm->verify_thread()`.

ericjordanmossman: I'm not sure I understand. Since writeCopy is a method on the Realm, if the realm were being accessed from an incorrect thread, then the operation catches with "Realm accessed from incorrect thread".

ericjordanmossman: The verify thread check happens in Realm::write_copy.

Realm/ObjectServerTests/SwiftObjectServerTests.swift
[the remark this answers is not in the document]
ericjordanmossman: We can't check history before the realm open. Offline decided to document what incorrect usage could receive "incompatible history" error.

leemaguire (approved):
Some small comments left due to incorrect indentation on comments but otherwise LGTM

Realm/RLMRealm.mm
```diff
@@ -944,16 +950,24 @@ - (RLMObject *)createObject:(NSString *)className withValue:(id)value {
...
 
     try {
-        _realm->verify_thread();
-        try {
-            _realm->read_group().write(path.UTF8String, static_cast<const char *>(key.bytes));
-        }
-        catch (...) {
-            _impl::translate_file_exception(path.UTF8String);
+        if (enableSync) {
+            _realm->write_copy(path.UTF8String, {static_cast<const char *>(key.bytes), key.length});
```
tgoyne, line 962: This call needs to be inside the try/catch which translates file exceptions. Currently if any errors occur this'll result in an c++ exception escaping, and we need to translate it to an obj-c exception.

Realm/ObjectServerTests/SwiftObjectServerTests.swift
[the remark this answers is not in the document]
tgoyne: We can catch the IncompatibleHistory exception and throw a new one with a better error message. This should just require adding a clause to `RLMRealmTranslateException()`.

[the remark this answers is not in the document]
ericjordanmossman: What kind of clause would be added? Should I be looking for what `kind` the exception is?
I can't switch case on a new kind, since the exception is classified as `AccessError`, which already exists. What property in the error could I branch off of? `m_underlying` seems like the only unique thing? But that's just a string.

Screen Shot 2022-01-10 at 3 06 20 PM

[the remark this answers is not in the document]
tgoyne: It throws an `IncompatibleHistories` exception. If you catch that it'll take precedence over the catch clause for the base class.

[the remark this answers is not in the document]
tgoyne: Or nevermind, it looks like it's being translated incorrectly in the object store code before that. We do check the string message in a few spots where there's no better way, but really the OS code should be adjusted to throw something better.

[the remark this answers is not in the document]
ericjordanmossman: Do you know a good spot where we check the string message? So that I can see how do it in other places.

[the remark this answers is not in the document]
ericjordanmossman: > it looks like it's being translated incorrectly in the object store code before that.

Where does this happen? So I can create a core ticket.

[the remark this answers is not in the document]
ericjordanmossman: I wasn’t sure the best way to check the underlying, so I committed a possible example to see what you think.

As an aside, the test I added started failing because of a session sticking around whenever I added string check. I couldn't tell why.
Screen Shot 2022-01-10 at 4 13 40 PM

ericjordanmossman:
The CI tests should fail with an exception until this is merged and updated to: https://github.com/realm/realm-core/pull/5161

After that mismatching `enableSync` to the kind of configuration that's ultimately used should throw:
```
  Realm file’s history format is incompatible with the 
  settings in the configuration object being used to open 
  the Realm. Note that Realms configured for sync cannot be 
  opened as non-synced Realms, and vice versa. Otherwise, the 
  file may be corrupt. Path: xyz
```

ericjordanmossman:
Core updated.
Ready for review again.

leemaguire:
We will deprecate this PR in favour of https://github.com/realm/realm-swift/pull/7613

ericjordanmossman:
@leemaguire No problem - should I close this PR?

tgoyne:
The `seedFilePath` part of this is still relevant after #7613.

PR #7524, opened 2021-11-09
Add post clone script for SwiftLint

jsflax:


ci_scripts/ci_post_clone.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/sh
+
+# Set the -e flag to stop running the script in case a command returns
+# a non-zero exit code.
+set -e
+
+if [[ $CI_WORKFLOW == "SwiftLint" ]]
+then
+    brew install swiftlint
+fi
```
tgoyne, line 10: Any new shell scripts we're adding should pass shellcheck, which in this case needs:

```suggestion
if [ "$CI_WORKFLOW" = "SwiftLint" ]; then
    brew install swiftlint
fi
```

tgoyne (approved):

PR #7529, opened 2021-11-12
Rework obj-c bridging and RealmCollection

tgoyne:
This some refactoring extracted from custom type mappings, and there's three sort of but not really related pieces which are in separate commits.

1. CustomObjectiveCBridgeable was designed to be `internal`, both to avoid cluttering public types with methods and to make it so that users can't declare their own types as conforming to it. This is no longer possible with custom type mappings, as we need a public version that their mapped types conform to. Rather than having several different protocols related to bridging (CustomObjectiveCBridgeable, RealmEnum, ThreadConfined) plus also types which were implicitly bridged, now all types we pass to and from obj-c conform to _ObjcBridgeable. Like RealmEnum, the obj-c to Swift conversion is failable, and `staticBridgeCast` is now the single place where we force-unwrap the bridged values for places that should never fail.

    This eliminates lots of dynamic casting and type checking in favor of a single virtual call on the type, so this probably improves performance somewhat, and it makes a lot of bits of the implementation simpler.

2. While updating the collection types to use the new bridging, I got annoyed at having to make the same changes in nine places, and I tried to see if I could extract the duplicated code. The answer to that was yes. Swift protocol extensions have gotten much more powerful in the time since RealmCollection was originally implemented, and now we can implement nearly all of the RealmCollection protocol via extensions rather than in each collection type. The convenience wrappers are defined in RealmCollection extensions, while the functions which need to actually access the backing collection are extensions on an internal RealmCollectionImpl protocol. This means that most of the collection access functions are now defined in exactly one place. In addition, they're also _documented_ in only one place. We used to have to redocument them all over the place to work with option-click, but both that and DocC now inherit documentation comments from the protocol which declared a function.

    This also made it possible to dramatically simplify AnyRealmCollection. Rather than wrapping a RealmCollection, it now just wraps a RLMCollection. Previously this would have required re-implementing every function on it, but now the implementations are all just inherited from the protocol.

    Doing this exposed the fact that RLMCollection was missing a few methods (such as distinct and thaw).

3. While all of the tests passed with the above changes (except for `description` due to the AnyRealmCollection change and a few error messages that changed wording slightly), the AnyRealmCollection change made it so that the collection tests no longer test what they're supposed to test. Before Swift 2, a subclass of a generic class had to have the same generic types, and you couldn't write things like `class ResultsTests: RealmCollectionTests<Results>`. This meant that `RealmCollectionTypeTests` had to operate on a type-erased collection rather than a generic collection. Changing AnyRealmCollection to wrap a RLMCollection directly rather than a RealmCollection meant that a test which uses `AnyRealmCollection(List<T>())` no longer calls any of `List<T>`'s functions.

    Fortunately, these days we can just make RealmCollectionTypeTests generic and operate on the desired collection directly. I also renamed it to RealmCollectionTests to match the fact that we dropped the `Type` suffix for protocols when the standard library did.


tgoyne:
Perf-wise this turned out to be neutral for types which don't require custom bridging and roughly twice as fast for types which do.

Realm/RLMCollection.h
```diff
@@ -154,40 +172,63 @@ typedef RLM_CLOSED_ENUM(int32_t, RLMPropertyType);
 /**
  Returns all objects matching the given predicate in the collection.
 
- @param predicate   The predicate with which to filter the objects.
+ This is only supported for managed collections.
```
jsflax, line 175: Why is this on the base protocol then? We have managed types.

Realm/RLMDictionary.mm
```diff
@@ -493,6 +493,22 @@ - (id)objectAtIndex:(NSUInteger)index {
     @throw RLMException(@"This method is not available on RLMDictionary.");
 }
 
+- (nullable id)firstObject {
```
jsflax, line 496: Why not just mark these as `@optional`?

RealmSwift/Impl/ObjcBridgeable.swift
```diff
@@ -0,0 +1,180 @@
...
+//
+////////////////////////////////////////////////////////////////////////////
+
+import Foundation
+import Realm
+
+/// A type which can be bridged to and from Objective-C.
+///
+/// Do not use this protocol or the functions it adds directly.
+public protocol _ObjcBridgeable {
```
jsflax, line 25: Why is this public? I know that the public protocols that use cannot refine the protocol if it were internal, but why not have the final types implement it?

jsflax: Ah, the final types can be custom user types. Disregard.

tgoyne: We also can't use it as a protocol requirement on public functions without it being public, which lead to a bunch of awful dynamic casts which would always succeed.

Realm/RLMCollection.h
tgoyne: This is just documenting existing behavior. The existence of the managed subclasses isn't part of the API, and the base class exposes all of the functions, including the ones which will just throw an exception in the unmanaged case.

Realm/RLMDictionary.mm
tgoyne: It maybe should be. This is weirdness resulting from RLMDictionary conforming to RLMCollection but Map not conforming to RealmCollection. Everything which conforms to RealmCollection has first/last so it's sort of awkward to have to handle those being optional on the RLMCollection.

CHANGELOG.md
```diff
@@ -27,6 +28,10 @@ x.y.z Release notes (yyyy-MM-dd)
 * Conform `@ThreadSafe` and `ThreadSafeReference` to `Sendable`.
 * Allow using Swift enums which conform to `PersistableEnum` as the value type
   for all Realm collections.
+* `AnyRealmCollection` now conforms to `Encodable`.
+* Greatly improve performance of reading AnyRealmValue and enum types from
```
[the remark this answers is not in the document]
tgoyne: Just a bad merge; I moved it between sections and then ended up with it in both accidentally.

Realm/RLMDictionary.mm
tgoyne: I've switched the ones which previously were optional back to being optional.

jsflax (approved):

RealmSwift/Tests/RealmCollectionTypeTests.swift
```diff
@@ -69,18 +69,47 @@ class CTTStringSet: Object {
...
+                                                          CTTLinkTarget.self,
+                                                          CTTStringList.self,
+                                                          CTTStringSet.self,
+
+                                                          SwiftDoubleListOfSwiftObject.self,
+                                                          SwiftListOfSwiftObject.self,
+                                                          SwiftObject.self,
+                                                          SwiftBoolObject.self])
+}
+class RealmCollectionTests<Collection: RealmCollection, AggregateCollection: RealmCollection>: TestCase where
```
leemaguire, line 89: This file should also be renamed to `RealmCollectionTests.swift`

leemaguire (approved):

PR #7537, opened 2021-11-17
Exposed RLMArray  initWithObjectType constructor to create primitive type array using swift language in realm Objective-c Library

tusharmutreja300:
I am working on the project that has the combination of both objective-c and Swift. While working with realm-coca, using the swift language in same project , I found that below mentioned constructor is not exposed in RLMArray.h file to create the primitive type Array.
`- (instancetype)initWithObjectType:(RLMPropertyType)type optional:(BOOL)optional;`

Only `- (instancetype)initWithObjectClassName:(NSString *)objectClassName;` is exposed to create the array of RLMObject.

As With the requirement of project ,we require this constructor to be exposed .


jsflax:
This is achievable using generics, eg., `RLMArray<RLMInt>`.

PR #7551, opened 2021-11-29
Async write transactions ObjC API

pavel-ship-it:
This pull request to enable asynchronous transactions in Objective-С SDK. The feature will be exposed to Swift SDK in the later pull requests.

```
    [realm asyncTransactionWithBlock:^{
        [StringObject createInRealm:realm withValue:@[@"string"]];
    } onComplete:^{
        // optional handling
    }];

    [realm beginAsyncWriteTransaction:^{
        [StringObject createInRealm:realm withValue:@[@"string"]];
        [realm commitAsyncWriteTransaction];
    }];

    RLMAsyncTransactionId asyncTransactionId = [realm beginAsyncWriteTransaction:^{
        // ...
    }];
    [realm cancelAsyncTransaction:asyncTransactionId];
```


Realm/RLMRealm.h
```diff
@@ -500,6 +500,69 @@ typedef void (^RLMNotificationBlock)(RLMNotification notification, RLMRealm *rea
...
 - (BOOL)transactionWithoutNotifying:(NSArray<RLMNotificationToken *> *)tokens block:(__attribute__((noescape)) void(^)(void))block error:(NSError **)error;
 
+#ifdef REALM_ASYNC_WRITES
+
+/**
+ AsyncHandle A handle for asynchronous transaction returned
+ by `beginAsyncWriteTransaction:` or `commitAsyncWriteTransaction:` calls.
+ Can be used to cancel the transaction.
+*/
+typedef unsigned AsyncHandle;
```
jsflax, line 510: Bad grammar/punctuation on doc string. I'm also not sure "handle" is idiomatic for Realm. I would consider `AsyncTransactionToken` or `AsyncTransactionId`.

Realm/RLMRealm.h
```diff
@@ -500,6 +500,69 @@ typedef void (^RLMNotificationBlock)(RLMNotification notification, RLMRealm *rea
...
+/**
+ Indicates if the Realm is currently engaged in an async write transaction.
+ @warning Do not simply check this property and then start a write transaction whenever an object needs to be
+ created, updated, or removed. Doing so might cause a large number of write transactions to be created,
+ degrading performance. Instead, always prefer performing multiple updates during a single transaction.
+ */
+@property (nonatomic, readonly) BOOL inAsyncWriteTransaction;
+/**
```
jsflax, line 518: Docstrings should have new line between fields.

Realm/RLMRealm.mm
```diff
@@ -706,6 +706,53 @@ - (void)cancelWriteTransaction {
...
+
+- (AsyncHandle)commitAsyncWriteTransaction {
+    return _realm->async_commit_transaction();
+}
+
+- (AsyncHandle)commitAsyncWriteTransaction:(nullable void(^)())doneBlock {
+    if (doneBlock) {
+        return _realm->async_commit_transaction(doneBlock);
+    }
+    return _realm->async_commit_transaction();
```
jsflax, line 727: Do you even need to check if `doneBlock` is not nil? It looks like you can pass nil into the core function.

Realm/RLMRealm.mm
```diff
@@ -706,6 +706,53 @@ - (void)cancelWriteTransaction {
...
+}
+
+- (AsyncHandle)commitAsyncWriteTransaction:(nullable void(^)())doneBlock isGroupingAllowed:(BOOL)isGroupingAllowed {
+    if (doneBlock) {
+        return _realm->async_commit_transaction(doneBlock, isGroupingAllowed);
+    }
+    return _realm->async_commit_transaction(nil, isGroupingAllowed);
```
jsflax, line 734: Same comment as above.

Realm/RLMRealm.h
```diff
@@ -500,6 +500,69 @@ typedef void (^RLMNotificationBlock)(RLMNotification notification, RLMRealm *rea
...
+ @note Cancelling a commit will not abort the commit, it will only
+ cancel the callback informing of commit completion.
+*/
+- (void)cancelAsyncTransaction:(AsyncHandle)handle;
+/** Commits a code block to asynchronous transactions queue.
+ @param block  The block containing actions to perform.
+ @param doneBlock  is queued for execution on the scheduler associated with
+ the current realm. It will run after the commit has reached stable storage.
+*/
+- (void)asyncTransactionWithBlock:(void(^)())block onComplete:(nullable void(^)())doneBlock;
```
jsflax, line 561: Why doesn't this return an `AsyncHandle`?

Realm/RLMRealm.mm
```diff
@@ -706,6 +706,53 @@ - (void)cancelWriteTransaction {
...
+}
+
+- (void)cancelAsyncTransaction:(AsyncHandle)handle {
+    _realm->async_cancel_transaction(handle);
+}
+
+- (void)asyncTransactionWithBlock:(void(^)())block onComplete:(nullable void (^)())doneBlock {
+    AsyncHandle transaction = [self beginAsyncWriteTransaction: ^{
```
jsflax, line 742: Why are you storing this in a local variable?

Realm/RLMRealm.mm
```diff
@@ -706,6 +706,53 @@ - (void)cancelWriteTransaction {
...
+        block();
+        if (_realm->is_in_async_transaction()) {
```
jsflax, line 744: This looks like there is undefined behavior?  How could they not be in an async transaction and if they aren't, should it throw instead of never committing?

Realm/Tests/RealmTests.mm
```diff
@@ -1455,6 +1455,381 @@ - (void)testActiveVersionLimit {
...
+- (void)testAsyncTransactionShouldWrite {
+    RLMRealm *realm = RLMRealm.defaultRealm;
+    XCTestExpectation *asyncComplete = [self expectationWithDescription:@"async transaction complete"];
+
+    XCTAssertEqual(0, [StringObject allObjectsInRealm:realm].count);
+
+    [realm asyncTransactionWithBlock:^{
+        [realm createObject:StringObject.className withValue:@[@"string"]];
+    }
+                          onComplete:^{
```
jsflax, line 1470: Weird formatting.

Realm/Tests/RealmTests.mm
```diff
@@ -1455,6 +1455,381 @@ - (void)testActiveVersionLimit {
...
+
+    [realm beginAsyncWriteTransaction:^{
+        [StringObject createInRealm:realm withValue:@[@"string"]];
+        [expectation fulfill];
+    }];
+
+    AsyncHandle handle = [realm beginAsyncWriteTransaction:^{
+        [StringObject createInRealm:realm withValue:@[@"string"]];
+        [expectation fulfill];
+        [realm cancelAsyncTransaction:handle];
```
jsflax, line 1783: Does this actually compile?

Realm/RLMRealm.h
pavel-ship-it: I’d say it’s `AsyncTransactionId` as it really holds nothing but a number of a transaction. Token is something we’re use to retain the object (`NotificationToken`)

Realm/RLMRealm.h
pavel-ship-it: yes, this can return and it can be useful

Realm/RLMRealm.mm
pavel-ship-it: You're right. It's something I added looking at sync transactions

Realm/RLMRealm.mm
pavel-ship-it: `async_commit_transaction` expecting done block as `&& the_done_block` and passing nil value will not be same as passing nil

Realm/Tests/RealmTests.mm
pavel-ship-it: Yes, and working perfectly. It was an issue with variable capturing by the closure in the Swift implementation but it's OK for objc.

Realm/RLMRealm.mm
jsflax: In that case, these blocks should not be `nullable` since you've already overloaded the method to not take a block param.

jsflax:
Why is there a projection test failing?

CHANGELOG.md
```diff
@@ -1,6 +1,26 @@
 x.y.z Release notes (yyyy-MM-dd)
 =============================================================
 ### Enhancements
+* Conform `@ThreadSafe` and `ThreadSafeReference` to `Sendable`.
```
ericjordanmossman, line 4: I believe this is from a different PR. Should it be here?

Realm/RLMRealm.h
```diff
@@ -500,6 +505,96 @@ typedef void (^RLMNotificationBlock)(RLMNotification notification, RLMRealm *rea
...
+ */
+- (RLMAsyncTransactionId)beginAsyncWriteTransaction:(void(^)())block;
+
+/** Commit asynchronous transaction.
+ 
+ @param doneBlock  is queued for execution on the scheduler associated with
+                  the current realm. It will run after the commit has reached stable storage.
+ 
+ @param isGroupingAllowed If `true`, the next `commitAsyncWriteTransaction` *may* run without an
+                         intervening synchronization of stable storage.  Such a sequence of commits
```
ericjordanmossman, line 557: Indentation. Same with `doneBlock` above.

CHANGELOG.md
```diff
@@ -1,6 +1,26 @@
...
+* Add an API for asynchronous transactions on ObjC.
```
ericjordanmossman, line 5: Really small change: "Add Obj-C API for asynchronous transactions".
Just a suggestion.

Realm/RLMRealm.h
```diff
@@ -500,6 +505,96 @@ typedef void (^RLMNotificationBlock)(RLMNotification notification, RLMRealm *rea
...
+@property (nonatomic, readonly) BOOL inAsyncWriteTransaction;
+
+/// A block type representing a block which can be used to report an async transaction related error to the application.
+typedef void(^RLMRealmAsyncErrorHandler)(RLMAsyncTransactionId, NSError *);
+
+/**
+ Set the error handler for the asynbhronous transactions.
+ Synchronous try/catch hadling will not work with the asynchronous transactions.
+ Asynchronous exceptions will be dispatched to the handler as `NSErrors`.
+ @note pass `nil` as a parameter to remove the handler
```
ericjordanmossman, line 527: Typos:
asynchronous
handling

Realm/RLMRealm.h
```diff
@@ -500,6 +505,96 @@ typedef void (^RLMNotificationBlock)(RLMNotification notification, RLMRealm *rea
...
+
+/** Commit asynchronous transaction.
+ 
+ @param doneBlock  is queued for execution on the scheduler associated with
+                  the current realm. It will run after the commit has reached stable storage.
+ 
+ @param isGroupingAllowed If `true`, the next `commitAsyncWriteTransaction` *may* run without an
+                         intervening synchronization of stable storage.  Such a sequence of commits
+                         form a group. In case of a platform crash, either none or all of the commits
+                         in a group will reach stable storage.
```
ericjordanmossman, line 559: I may have missed a larger design discussion - but would this violate atomicity?
I suppose since this is opt-in, and the developer is warned it's fine?

Realm/RLMRealm.h
```diff
@@ -500,6 +505,96 @@ typedef void (^RLMNotificationBlock)(RLMNotification notification, RLMRealm *rea
...
+ @param RLMAsyncTransactionId The transaction Id returned by `beginAsyncWriteTransaction` or  `commitAsyncWriteTransaction`
+*/
+- (void)cancelAsyncTransaction:(RLMAsyncTransactionId)asyncTransactionId;
+
+/** Commits a code block to asynchronous transactions queue.
+ 
+ @param block  The block containing actions to perform.
+ 
+ @param doneBlock  is queued for execution on the scheduler associated with
+                  the current realm. It will run after the commit has reached stable storage.
```
ericjordanmossman, line 588: Indentation.

Realm/Tests/RealmTests.mm
```diff
@@ -1455,6 +1455,478 @@ - (void)testActiveVersionLimit {
...
+
+- (void)testAsyncTransactionShouldWriteOnCommit {
+    RLMRealm *realm = RLMRealm.defaultRealm;
+    XCTestExpectation *asyncComplete = [self expectationWithDescription:@"async transaction complete"];
+
+    XCTAssertEqual(0, [StringObject allObjectsInRealm:realm].count);
+
+    [realm beginAsyncWriteTransaction:^{
+        [realm createObject:StringObject.className withValue:@[@"string"]];
+        [realm commitAsyncWriteTransaction:^{
```
ericjordanmossman, line 1486: Should the developer always be putting the commit _in_ the transaction block?

ericjordanmossman: I see this now: 
```
     [realm createObject:StringObject.className withValue:@[@"string 2"]];
     [realm commitAsyncWriteTransaction:^{
         [asyncComplete fulfill];
     }];
```

Realm/RLMRealm.h
pavel-ship-it: I add  "false by default" to make it obvious.

ericjordanmossman (approved):
The three failing tests (testAddObjectsWithNilPartitionValue, SyncUI tests, and Xcode build crash) aren't related AFAIC.
LGTM in this state. Make sure to get a second opinion, since others may have a better understanding of the design doc.

jsflax:
You need to add `REALM_ASYNC_WRITES` to the Package.swift file testCxxSettings so that your tests are ran at least for SPM.

Realm/RLMRealm.h
```diff
@@ -500,6 +505,96 @@ typedef void (^RLMNotificationBlock)(RLMNotification notification, RLMRealm *rea
...
+/** Commits a code block to asynchronous transactions queue.
+ 
+ @param block  The block containing actions to perform.
+ 
+ @param doneBlock  is queued for execution on the scheduler associated with
+        the current realm. It will run after the commit has reached stable storage.
+
+ @return Asynchronous transaction's Id.
+ */
+- (RLMAsyncTransactionId)asyncTransactionWithBlock:(void(^)())block onComplete:(void(^)())doneBlock;
```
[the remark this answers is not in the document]
pavel-ship-it: `completionBlock` sounds better. Will use it for `commitAsyncWriteTransaction` too

Realm/RLMRealm.mm
```diff
@@ -706,6 +706,72 @@ - (void)cancelWriteTransaction {
...
+    return _realm->async_commit_transaction();
+}
+
+- (RLMAsyncTransactionId)commitAsyncWriteTransaction:(void(^)())doneBlock {
+    return _realm->async_commit_transaction(doneBlock);
+}
+
+- (RLMAsyncTransactionId)commitAsyncWriteTransaction:(nullable void(^)())doneBlock isGroupingAllowed:(BOOL)isGroupingAllowed {
+    if (doneBlock) {
+        return _realm->async_commit_transaction(doneBlock, isGroupingAllowed);
```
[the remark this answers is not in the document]
pavel-ship-it: Non-pointer operand type 'void (^ _Nullable)()' incompatible with nullptr

Realm/RLMRealm.h
```diff
@@ -500,6 +505,96 @@ typedef void (^RLMNotificationBlock)(RLMNotification notification, RLMRealm *rea
...
+ on a dedicated background thread.
+ 
+ @param completionBlock  is queued for execution on the scheduler associated with
+        the current realm. It will run after the commit has reached stable storage.
+        Callbacks to `completionBlock` will occur in the order of `commitAsyncWriteTransaction`
+ 
+ @param isGroupingAllowed If `true`, the next `commitAsyncWriteTransaction` *may* run without an
+        intervening synchronization of stable storage.  Such a sequence of commits
+        form a group. In case of a platform crash, either none or all of the commits
+        in a group will reach stable storage. `false` by default.
```
tgoyne, line 563: This needs to explain why you'd want to set this.

Realm/RLMRealm.h
```diff
@@ -500,6 +505,96 @@ typedef void (^RLMNotificationBlock)(RLMNotification notification, RLMRealm *rea
...
+/// :nodoc:
+- (RLMAsyncTransactionId)commitAsyncWriteTransaction:(void(^)())completionBlock;
+/// :nodoc:
+- (RLMAsyncTransactionId)commitAsyncWriteTransaction;
+
+/** Cancel a queued code block (either for an `asyncTransactionWithBlock`
+ or for an `commitAsyncWriteTransaction`)
+ 
+ @note Cancelling a commit will not abort the commit, it will only cancel the callback
+       informing of commit completion.
```
tgoyne, line 577: So it doesn't cancel it? I don't understand what this function actually does.

Realm/RLMRealm.h
```diff
@@ -500,6 +505,96 @@ typedef void (^RLMNotificationBlock)(RLMNotification notification, RLMRealm *rea
...
+ 
+ @param RLMAsyncTransactionId The transaction Id returned by `beginAsyncWriteTransaction` or  `commitAsyncWriteTransaction`
+*/
+- (void)cancelAsyncTransaction:(RLMAsyncTransactionId)asyncTransactionId;
+
+/** Commits a code block to asynchronous transactions queue.
```
tgoyne, line 583: Using the word "commit" here is confusing because it's a very different meaning of the word from what we're usually talking about when we use "commit" related to a write.

Realm/RLMRealm.h
```diff
@@ -500,6 +505,96 @@ typedef void (^RLMNotificationBlock)(RLMNotification notification, RLMRealm *rea
...
+ 
+ @param isGroupingAllowed If `true`, the next `commitAsyncWriteTransaction` *may* run without an
+        intervening synchronization of stable storage.  Such a sequence of commits
+        form a group. In case of a platform crash, either none or all of the commits
+        in a group will reach stable storage. `false` by default.
+
+ @return Asynchronous transaction's Id.
+*/
+- (RLMAsyncTransactionId)commitAsyncWriteTransaction:(nullable void(^)())completionBlock isGroupingAllowed:(BOOL)isGroupingAllowed;
+/// :nodoc:
```
tgoyne, line 568: These do need to be documented.

Realm/RLMRealm.h
```diff
@@ -500,6 +505,96 @@ typedef void (^RLMNotificationBlock)(RLMNotification notification, RLMRealm *rea
...
+ @param completionBlock  is queued for execution on the scheduler associated with
+        the current realm. It will run after the commit has reached stable storage.
+        Callbacks to `completionBlock` will occur in the order of `commitAsyncWriteTransaction`
+ 
+ @param isGroupingAllowed If `true`, the next `commitAsyncWriteTransaction` *may* run without an
+        intervening synchronization of stable storage.  Such a sequence of commits
+        form a group. In case of a platform crash, either none or all of the commits
+        in a group will reach stable storage. `false` by default.
+
+ @return Asynchronous transaction's Id.
```
tgoyne, line 565: Is this the same Id as was returned from `beginAsyncWriteTransaction:`?

Realm/RLMRealm.mm
```diff
@@ -706,6 +710,72 @@ - (void)cancelWriteTransaction {
...
+                    RLMRealmTranslateException(&error);
+                    block(asyncTransactionId, error);
+                }
+            }
+        });
+    }
+}
+
+- (RLMAsyncTransactionId)beginAsyncWriteTransaction:(void(^)())block {
+    return _realm->async_begin_transaction(block);
```
tgoyne, line 739: These can all throw exceptions and need to translate the exceptions from c++ to obj-c.

Realm/RLMRealm.mm
```diff
@@ -706,6 +710,72 @@ - (void)cancelWriteTransaction {
...
+}
+
+- (RLMAsyncTransactionId)commitAsyncWriteTransaction:(void(^)())completionBlock {
+    return _realm->async_commit_transaction(completionBlock);
+}
+
+- (RLMAsyncTransactionId)commitAsyncWriteTransaction:(nullable void(^)())completionBlock isGroupingAllowed:(BOOL)isGroupingAllowed {
+    if (completionBlock) {
+        return _realm->async_commit_transaction(completionBlock, isGroupingAllowed);
+    }
```
tgoyne, line 753: This block isn't doing anything.

Realm/RLMRealm.mm
```diff
@@ -706,6 +710,72 @@ - (void)cancelWriteTransaction {
...
+    return _realm->async_commit_transaction(nil, isGroupingAllowed);
+}
+
+- (void)cancelAsyncTransaction:(RLMAsyncTransactionId)asyncTransactionId {
+    _realm->async_cancel_transaction(asyncTransactionId);
+}
+
+- (RLMAsyncTransactionId)asyncTransactionWithBlock:(void(^)())block onComplete:(void (^)())completionBlock {
+    RLMAsyncTransactionId asyncTransactionId = [self beginAsyncWriteTransaction: ^{
```
tgoyne, line 762: No space after colons in obj-c message sends.

Realm/Tests/RealmTests.mm
```diff
@@ -1455,6 +1455,471 @@ - (void)testActiveVersionLimit {
...
+            StringObject *stringObject = [StringObject allObjectsInRealm:realm].firstObject;
+            XCTAssertEqualObjects(stringObject.stringCol, @"string");
```
tgoyne, line 1488: This test would pass even if it was moved out of this callback to before the call to `commitAsyncWriteTransaction:`, so it isn't actually testing much. A more effective way to test it would be to wrap it in `dispatchAsyncAndWait:` and obtain a new RLMRealm instance so that it's actually verifying that a new reader sees the committed data.

Package.swift
```diff
@@ -122,7 +125,7 @@ let package = Package(
             targets: ["Realm", "RealmSwift"]),
     ],
     dependencies: [
-        .package(name: "RealmDatabase", url: "https://github.com/realm/realm-core", .exact(Version(coreVersionStr)!))
+        .package(name: "RealmDatabase", url: "https://github.com/realm/realm-core", .branch("master"))
```
jsflax, line 128: This should be using the correct version.

Realm/RLMRealm.h
```diff
@@ -500,6 +505,144 @@ typedef void (^RLMNotificationBlock)(RLMNotification notification, RLMRealm *rea
...
+        If you are not interested in possible errors, pass in `NULL`.
+
+ @return The Id for the asynchronous transaction or 0 in case of error.
+
+ @note The Id for the asynchronous transaction can be used to cancel the transaction. You can get it
+       as a return from `beginAsyncWriteTransaction`, `commitAsyncWriteTransaction` or `asyncTransactionWithBlock`.
+
+ @note The immediate return without `error` is not a guarantee of the future successful transaction commit.
+*/
+- (RLMAsyncTransactionId)commitAsyncWriteTransaction:(nullable void(^)())completionBlock isGroupingAllowed:(BOOL)isGroupingAllowed error:(NSError **)error;
```
jsflax, line 568: Why did you end up choosing an out error here instead of having the error be a block param?

pavel-ship-it: block param is better. Fixed

Package.swift
```diff
@@ -27,6 +27,7 @@ let cxxSettings: [CXXSetting] = [
     .define("REALM_VERSION_PATCH", to: String(coreVersionExtra[0])),
     .define("REALM_VERSION_EXTRA", to: "\"\(coreVersionExtra.count > 1 ? String(coreVersionExtra[1]) : "")\""),
     .define("REALM_VERSION_STRING", to: "\"\(coreVersionStr)\""),
+    .define("REALM_ASYNC_WRITES", .when(configuration: .debug)),
```
jsflax, line 30: Why isn't this just in the test setting for now?

jsflax (approved):
One Q, otherwise LGTM.

Package.swift
pavel-ship-it: Only to reduce the availability of a feature before making it fully accessible.

jsflax: I worry that this may end up confusing for users, as this will expose it to them until they go to release their app. The idea of feature flagging was that users would have to explicitly opt in.

pavel-ship-it: This is required as a value I add for tests will take effect on tests build only. The framework build will have no async code.

tgoyne: When Realm is installed via SPM, debug builds of the app use debug builds of Realm, and this will be set.

tgoyne: Actually I think this might be fine? It'll be defined when building Realm itself but not for the user's code which imports RLMRealm.h, so while the methods are defined the user won't be able to see them.

Realm/Tests/RealmTests.mm
```diff
@@ -1455,6 +1455,438 @@ - (void)testActiveVersionLimit {
...
+    RLMRealm *realm = RLMRealm.defaultRealm;
+    __block XCTestExpectation *asyncComplete = [self expectationWithDescription:@"async transaction complete"];
+    asyncComplete.expectedFulfillmentCount = 4;
+
+    RLMResults<RLMObject *> *resultsUnderTest = [StringObject allObjects];
+    RLMNotificationToken *token = [resultsUnderTest addNotificationBlock:^(RLMResults<RLMObject *> * _Nullable results, RLMCollectionChange * _Nullable change, NSError *error) {
+        if (!change)
+            return; // ignore initial
+        XCTAssertNotNil(change);
+        [asyncComplete fulfill];
```
tgoyne, line 1713: There's a race condition here. The first (or both) writes can happen before the initial notification, which results in fewer than four total calls to `fulfill`. You have to wait for the initial notification to be delivered before performing the writes.

Realm/Tests/RealmTests.mm
```diff
@@ -1455,6 +1455,438 @@ - (void)testActiveVersionLimit {
...
+    [realm beginAsyncWriteTransaction:^{
+        [StringObject createInRealm:realm withValue:@[@"string"]];
+        [expectation fulfill];
+        [realm commitAsyncWriteTransaction];
+    }];
+
+    RLMAsyncTransactionId asyncTransactionIdA = [realm beginAsyncWriteTransaction:^{
+        [StringObject createInRealm:realm withValue:@[@"string"]];
+        [expectation fulfill];
+        [realm cancelAsyncTransaction:asyncTransactionIdA];
```
tgoyne, line 1828: This doesn't actually do anything; the transaction is just being implicitly cancelled due to to commit.

Realm/RLMRealm.mm
```diff
@@ -706,6 +706,83 @@ - (void)cancelWriteTransaction {
...
+- (RLMAsyncTransactionId)beginAsyncWriteTransaction:(void(^)())block {
+    try {
+        _realm->async_begin_transaction(block);
+    }
+    catch (std::exception &ex) {
+        @throw RLMException(ex);
+    }
+}
+
+- (RLMAsyncTransactionId)commitAsyncWriteTransaction {
```
tgoyne, line 724: This should probably just return void? You can't do anything with the returned transaction id as there's no completion callback to cancel.

tgoyne (commented):
Some more things to test are that all of the new functions throw appropriate exceptions when called in invalid ways (wrong thread, on a read-only Realm, commit while not in a transaction), that invalidate works and does sensible things from each of the async write callbacks, and async writes when using queue-confined Realms.

Realm/RLMRealm.h
```diff
@@ -500,6 +505,138 @@ typedef void (^RLMNotificationBlock)(RLMNotification notification, RLMRealm *rea
...
+ Performs actions contained within the given block inside an asynctonous write transaction.
+ 
+ @param block  The block containing actions to perform.
+ 
+ @return The Id for the asynchronous transaction or 0 in case of error.
+
+ @note The Id for the asynchronous transaction can be used to cancel the transaction. You can get it
+       as a return from `beginAsyncWriteTransaction`, `commitAsyncWriteTransaction` or `asyncTransactionWithBlock`.
+ */
+- (RLMAsyncTransactionId)asyncTransactionWithBlock:(void(^)())block;
```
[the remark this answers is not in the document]
pavel-ship-it: Duplicate

Realm/RLMRealm.mm
```diff
@@ -706,6 +706,83 @@ - (void)cancelWriteTransaction {
...
+        if (completionBlock) {
+            return _realm->async_commit_transaction(^{ completionBlock(nil); }, isGroupingAllowed);
+        }
+        return _realm->async_commit_transaction(nullptr, isGroupingAllowed);
+    }
+    catch (...) {
+        NSError *error = nil;
+        RLMRealmTranslateException(&error);
+        completionBlock(error);
+        return 0;
```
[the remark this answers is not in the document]
tgoyne: This is still applicable. `async_commit_transaction()` throws exceptions for precondition violations, which should not result in a call to the completion handler.

tgoyne (approved):

PR #7554, opened 2021-12-02
Adjust where bridging is done for query values

tgoyne:
The next piece of refactoring to support custom type mapping, this time improving support for mapped types in queries.

Use RLMStatelessAccessor::unbox() when reading values for queries rather than a more limited query-specific version of that. This both eliminates some duplicated code and makes the query builder support types which require bridging, such as PersistableEnums and custom type mappings in collections.

This lets us drop most of the explicit handling for PersistableEnum in Query in favor of checks on the PersistedType, and adds support for queries on collections of enums via Query. Some operations can't really be performed on the enum type (e.g. sum and average) but might still be meaningful queries to write, so I added add `.rawValue` modifier to `Query` to support querying the raw value instead of the enum. This has no effect on the constructed NSPredicate as the above change makes them interchangable at that layer.

While writing tests for this, I noticed that the return type for many of the operations on Query was incorrect. All of the comparison operations returned Query<V> rather than Query<Bool>, which allowed many types of nonsensical queries to be expressed (e.g. `{ $0.intCol > 5 > 6}` and `{ $0.intCol }`).  There were also a few incidental incorrect constraints, such as `.count` on `Map` requiring the value type to be numeric.

Most of the changes in this PR are refactoring QueryTests.swift.gyb to reduce the size of QueryTests.swift. After adding tests for collections of enums, the generated file was over 50k lines of code, took multiple minutes to compile, and locked up Xcode if you made the mistake of opening the file. To fix this I shifted a lot of the looping from GYB to Swift. This results in more complicated tests that are harder to understand, but the status quo was pretty unworkable and custom type mappings will make the file even bigger.

This PR will currently fail on CI as it exposed some core bugs (https://github.com/realm/realm-core/pull/5057, https://github.com/realm/realm-core/pull/5080, https://github.com/realm/realm-core/pull/5081).


RealmSwift/Impl/Persistable.swift
```diff
@@ -60,12 +60,6 @@ extension _Persistable {
     }
 }
 
-// A helper protocol for types which we support directly rather than via
-// mapping to a supported type.
-public protocol _BuiltInPersistable {
-    associatedtype _RealmValue = Self
-}
```
tgoyne, line 67: _BuiltInPersistable didn't actually work and I gave up on figuring out why.

RealmSwift/Impl/Persistable.swift
```diff
@@ -51,7 +51,7 @@ public protocol _Persistable: _RealmSchemaDiscoverable, _ObjcBridgeable {
     static func _rlmKeyPathRecorder(with lastAccessedNames: NSMutableArray) -> Self
     // The type which is actually stored in the Realm. This is Self for types
     // we support directly, but may be a different type for enums and mapped types.
-    associatedtype _RealmValue: _Persistable
+    associatedtype PersistedType: _Persistable
```
tgoyne, line 54: This renaming is aligning it to what the public API for custom type mappings will use. It's not strictly part of this PR but renaming everything back to _RealmValue would have just produced annoying merge conflicts.

leemaguire (approved):
Nice cleanup on the tests, LGTM once CI is passing.

PR #7560, opened 2021-12-08
Allow initializing dictionaries entries with null object values

tgoyne:
Creating an unmanaged object with `Object(value:)` accidentally didn't let you set dictionary values to null for dictionaries of objects. The other collections of objects don't support nulls, so the similar code which this was adapted from didn't need to handle that.

### DIFF
--- a/Realm/Tests/DictionaryPropertyTests.m
+++ b/Realm/Tests/DictionaryPropertyTests.m
@@ -1667,4 +1667,29 @@ static RLMDictionary<NSString *, IntObject *><RLMString, IntObject> *managedTest
     [dict.realm cancelWriteTransaction];
 }
 
+- (void)testInitWithNullLink {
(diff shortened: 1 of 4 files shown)


Realm/Tests/DictionaryPropertyTests.m
```diff
@@ -1667,4 +1667,29 @@ - (void)testSetDictionaryManaged {
     [dict.realm cancelWriteTransaction];
 }
 
+- (void)testInitWithNullLink {
```
leemaguire, line 1670: This could probably use a test in Swift too.

jsflax (approved):

Realm/Tests/DictionaryPropertyTests.m
tgoyne: Added a test.

PR #7565, opened 2021-12-11
Rework querying on dictionaries

tgoyne:
While implementing memberwise comparisons for types mapped to embedded objects I found that some query things didn't work when the object was inside a dictionary. Fixing this ended up requiring significantly reworking how dictionaries on queries is implemented. `@allValues` being a collection operation meant that it couldn't have a keypath after it, so we couldn't query on the properties of objects in a dictionary. Changing it to be instead be treated as a normal keypath helps eliminate a lot of duplicated code and enables some more forms of queries. Our query keypath parsing didn't really support doing that, so I rewrote that. We now parse the keypath ASAP and operate on that parsed struct rather than passing around the string form and doing some odd logic on that.

Along the way I also found and fixed a few other dictionary bugs while writing tests for the bits of RLMQueryUtil which weren't covered.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -38,6 +38,10 @@ let people: Results<PersonProjection> = realm.objects(PersonProjection.self)
 * Queries on collections of PersistableEnums can now be performed with `where()`.
 * Add support for querying on the rawValue of an enum with `where()`.
 * `.count` is supported for Maps of all types rather than just numeric types in `where()`.
+* Add support for querying on the properties of objects contained in
+  dictionaries (e.g. "dictProperty.@allValues.name CONTAINS 'a'").
+* Improve the error message for many types of invalid predicates in queries.
+* Add support for comparing `@allKeys` to another property on the same object.
 
 ### Fixed
 * <How to hit and notice issue? what was the impact?> ([#????](https://github.com/realm/realm-cocoa/issues/????), since v?.?.?)
@@ -45,6 +49,10 @@ let people: Results<PersonProjection> = realm.objects(PersonProjection.self)
 * Add missing `Indexable` support for UUID. 
   ([Cocoa #7545](https://github.com/realm/realm-cocoa/issues/7545), since v10.10.0)
 * `where()` allowed constructing some nonsensical queries due to boolean comparisons returning `Query<T>` rather than `Query<Bool>`.
+* `@allValues` queries on dictionaries accidentally did not require "ANY".
+* Case-insensitive and diacritic-insensitive modifiers were ignored when
+  comparing the result of an aggregate operation to another property in a
+  query.
 
 <!-- ### Breaking Changes - ONLY INCLUDE FOR NEW MAJOR version -->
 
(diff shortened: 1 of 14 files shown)


Realm/RLMQueryUtil.mm
```diff
@@ -247,19 +250,28 @@ bool isNSNull(T) {
 // A reference to a column within a query. Can be resolved to a Columns<T> for use in query expressions.
 class ColumnReference {
 public:
-    ColumnReference(Query& query, Group& group, RLMSchema *schema, RLMProperty* property, const std::vector<RLMProperty*>& links = {})
-    : m_links(links), m_property(property), m_schema(schema), m_group(&group), m_query(&query), m_table(query.get_table())
+    ColumnReference(Query& query, Group& group, RLMSchema *schema, RLMProperty* property, std::vector<RLMProperty*>&& links = {})
```
leemaguire, line 253: what is the reason for changing this to use move semantics instead of pass by ref?

Realm/RLMQueryUtil.mm
```diff
@@ -1312,6 +1081,8 @@ void convert_null(T&& value, Fn&& fn) {
 
 template<typename T>
 using Identity = T;
+template<typename>
+using AlwaysDictionary = Dictionary;
```
leemaguire, line 1085: It looks like this is only used in one spot, what is the benefit of using this approach?

leemaguire (approved):
This is quite a nice PR. I have left minor comments, but LGTM.

CHANGELOG.md
```diff
@@ -4,11 +4,18 @@ x.y.z Release notes (yyyy-MM-dd)
...
+* Improve the error message for many types of invalid predicates in queries.
+* Add support for comparing `@allKeys` to another property on the same object.
 
 ### Fixed
 * Add missing `Indexable` support for UUID. 
   ([Cocoa #7545](https://github.com/realm/realm-cocoa/issues/7545), since v10.10.0)
+* `@allValues` queries on dictionaries accidentally did not require "ANY".
+* Case-insensitive and diacritic-insensitive modifiers were ignored when
+  comparing the result of an aggregate operation to another property in a
+  query.
```
bmunkholm, line 18: Since ?

Realm/RLMQueryUtil.mm
tgoyne: `do_add_constraint` takes a template template parameter which is used to map a type `T` into `Columns<Lst<T>>`, `Columns<Set<T>>`, or `Columns<T>`. This lets us use the same code path for Dictionary as well rather than duplicating it. Unlike the other collections Dictionary doesn't have a template parameter, so the dictionary template template ignores its template parameter and always produces `Columns<Dictionary>` regardless of `T`. Note that `T` is still used for the other parameter in the expression being built, so we can't skip the switch on type for Dictionaries even though the column is always resolved to the same type.

CHANGELOG.md
tgoyne: Whenever we first implemented collection aggregates sometime in the 0.x version range, but it's sorta unclear if it ever actually mattered before now.

Issue #7562, opened 2021-12-10
Only TestFlight Crash (iOS ... <15)


AdlibBeats:
{"app_name":"test","app_cohort":"2|date=1639096200000&sf=143469&tid=ce768aa66a3d199a9fca7ebbc6d5301ffc41fd4426e910468d2e396485202005&ttype=i","app_version":"1.1.6","timestamp":"2021-12-10 03:45:14.54 +0300","slice_uuid":"aa178c58-89ec-38b1-905b-4134f9770992","adam_id":1533641502,"build_version":"108","bundleID":"xxx.test","share_with_app_devs":false,"is_first_party":false,"bug_type":"109","os_version":"iPhone OS 12.5.5 (16H62)","incident_id":"6C234363-78C3-4962-A52D-0F7C276F8E40","is_beta":true,"name":"test"}
Incident Identifier: 6C234363-78C3-4962-A52D-0F7C276F8E40
Beta Identifier:     7E045794-93A6-4129-84B4-2C8A5D9EA5FF
Hardware Model:      iPhone6,2
Process:             test [147]
Path:                /private/var/containers/Bundle/Application/7EECE82C-35E1-42B1-B8E3-710F040DD726/test.app/test
Identifier:          xxx.test
Version:             108 (1.1.6)
AppStoreTools:       13C88a
AppVariant:          1:iPhone6,2:12.2
Beta:                YES
Code Type:           ARM-64 (Native)
Role:                Foreground
Parent Process:      launchd [1]
Coalition:           xxx.test [248]

Date/Time:           2021-12-10 03:45:11.8437 +0300
Launch Time:         2021-12-10 03:45:04.7125 +0300
OS Version:          iPhone OS 12.5.5 (16H62)
Baseband Version:    10.80.02
Report Version:      104

Exception Type:  EXC_CRASH (SIGABRT)
Exception Codes: 0x0000000000000000, 0x0000000000000000
Exception Note:  EXC_CORPSE_NOTIFY
Termination Description: DYLD, Library not loaded: @rpath/libswift_Concurrency.dylib | Referenced from: /private/var/containers/Bundle/Application/7EECE82C-35E1-42B1-B8E3-710F040DD726/test.app/Frameworks/RealmSwift.framework/RealmSwift | Reason: image not found
Triggered by Thread:  0

coutar-t:
Got exactly the same here. We had to ask our users to update to iOS 15.1.

WarSunsOfPeace:
[...]

PR #7568, opened 2021-12-13
Require Xcode 13.2 for async/await

tgoyne:
A workaround for #7562. Apple is trying to use bitcode to rebuild existing apps to now use async/await on iOS 14 but they're currently doing so incorrectly and actually just making things crash.

This is probably required anyway to properly support concurrency backdeployment so it's worth doing even if Apple fixes the bug. Presumably everyone using async/await is using bleeding-edge Xcode versions anyway.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,17 +7,20 @@ Xcode 12.4 is now the minimum supported version of Xcode.
 * None.
 
 ### Fixed
-* Add missing `Indexable` support for UUID. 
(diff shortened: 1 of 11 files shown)


marcusziade (approved):

jsflax (approved):

anlaital:
@tgoyne do you have any idea as to when a release containing this fix and prebuilt Xcode 13.2 binaries would be available?

tgoyne:
Today assuming no problems pop up.

PR #7569, opened 2021-12-15
Some assorting refactorings to support custom type mappings

tgoyne:
Hopefully the last of the refactoring for custom type mappings. I had to redo some of this to resolve conflicts with projections, and doing that incrementally happened to make it easy to extract to a separate PR.

Some of these may not make sense without context, so I'll attempt to describe them:

<dl>
<dt>Decouple _RealmSchemaDiscoverable from RealmCollectionValue</dt>
<dd>Projections must conform to RealmCollectionValue, but they can't appear as a property on an object and the schema discovery functions were unreachable. This means that RealmCollectionValue should no longer inherit from _RealmSchemaDiscoverable. Along the way I pulled keypath strings out to a separate file and shifted things around a bit so that RealmCollectionValue would no longer need _rlmDefaultValue().</dd>
<dt>Rename _OptionalPersistable to _PersistableInsideOptional</dt>
<dd> "OptionalPersistable" kept confusing me and making me think it was Optional<_Persistable> when it's actually the thing that can go inside Optional on a @Persisted property.</dd>
<dt>Move _rlmGetPropertyOptional to _PersistableInsideOptional</dt>
<dd>This is just a more correct place for it. This eliminates some unreachable functions which previously had to be fatalErrors().</dd>
<dt>Detangle _HasPersistedType from _Persistable</dt>
<dd>The next commit makes it so that RealmCollectionValue needs PersistedType, along with two types that are neither RealmCollectionValue nor _Persistable (NSNumber and NSDate). Defining it in three different protocols sorta worked but was a mess.</dd>
<dt>Rework the protocols used for collection elements</dt>
<dd>
This has a few parts:
<ol>
<li>Splitting RealmCollectionValue into RealmCollectionValue and _RealmCollectionValueInsideOptional makes it so that we can reject some invalid declarations such optional AnyRealmValue or any double-optional inside a collection at compile time.
<li>The aggregate functions on collections now check if Element.PersistedType conforms to MinMaxType/etc. rather than if Element does. This is required for custom type mappings because CustomPersistable can't conditionally conform to MinMaxType/etc. and it would be awkward to tell users to do that manually.
<li>The sorted(ascending:) function for non-object collections required that Element conform to Comparable, but we actually can sort all non-object types.
<li>Optional now conforms to MinMaxType/etc. when the wrapped type does. This made it possible to delete a bunch of duplicated functions for optional and non-optional versions.
<li>Rejecting double-optionals in collections broke TestValueFactory, which had unreachable functions with types like KeyPath<Object, List<T??>>. Making it work again ended up being a net reduction in code, partially because the above change eliminated the need for the Optional versions of some tests.
<li>Making Decimal128 conform to Numeric simplifies NumericTestValueFactory and is something the type should have done originally.
</ol>
</dd>
<dt>Use @_marker where applicable</dt>
<dd>@_marker indicates that a protocol has no runtime effects and just statically tags a type. Adding this where applicable has minor binary size benefits and makes some accidental misuse impossible.</dd>
</dl>


RealmSwift/Tests/ProjectedCollectTests.swift
```diff
@@ -37,7 +31,6 @@ class ProjectedCollections: Projection<PersistedCollections> {
 }
 
 class ProjectedCollectionsTestsTemplate: TestCase {
```
leemaguire, line 33: Very minor but it looks like the file name for this should be `ProjectedCollectionTests.swift`

leemaguire (approved):

Issue #5073, opened 2017-06-27
Use custom initializer on an Realm object inside a class with generic type


klemenzagar91:
I have a class called `ObjectManager` which has generic type `O: Object, Persistable`. This way I can have one class that can operate with multiple different subclasses of realm `Object`. But the problem is in the last line  `cache = O(message: "hello world")` the error is:  **Incorrect argument label in call (have 'message:', expected 'value:')**

When typing it suggests me that I can use initializer of `Persistable` protocol, but complain during the compiling, that it expect an initializer from `Object` class. Is it possible to cast Persistable initializer?
```swift
protocol Persistable {
  init(message: String)
}

class CustomObject: Object, Persistable {
  dynamic var message: String = ""
  required convenience init(message: String) {
    self.init()
    self.message = message
  }
}

class ObjectManager<O>: NSObject where O: Object, O: Persistable {
  var cache: O?
  
  func didReceive(message: String) {
    cache = O(message: "hello world")
  }
}

```

klemenzagar91:
One workaround I found is to replace init fit a function that I apply on the object right after initialization. 
Can somebody tell me I there is a better way?
```
protocol Persistable {
  func fillWith(message: String)
}

class CustomObject: Object, Persistable {
  dynamic var message: String = ""

  
  func fillWith(message: String) {
    self.message = message
  }
}

class ObjectManager<O>: NSObject where O: Object, O: Persistable {
  var cache: O?
  
  func didReceive(message: String) {
    let object = O()
    object.fillWith(message: message)
  }
}
```

bdash:
@austinzheng provided an [answer on your StackOverflow question](https://stackoverflow.com/a/44791896/1991710).

Issue #5069, opened 2017-06-25
Realm notifications don't automatically sorted with new added object


peihsendoyle:
I have a table view and results get from Realm query with a sorted criteria as "createdAt" ascending. I integrate the results array with Notification so that it automatically update the tableView.

The problem here is when I add new object with the "createdAt" property as Date() (it means today time) to Realm, the notification get fired but it update the new object to the last row of tableView (although I sort the results with: realm.objects().sorted(byKeyPath: "createdAt) ). 
I want it to the first row (first index) of tableView.

Maybe I don't know exactly how Realm notifications work, anyone can help me clear this case? 

P/s: Because I'm on board now to go home, so I'm sorry that I cannot provide the code I used on tableView. I will provide it when I come home, thanks all.

nirinchev:
If `createdAt` behaves as expected, then I would assume the newly created object will have the largest `createdAt` value (because it was created last). This means that an ascending (smallest to largest) ordering will correctly put it as the last object in the collection. Perhaps you meant to order them in descending order?

peihsendoyle:
@nirinchev thanks for your help.

When I put the order in `realm.objects().sorted(byKeyPath: "created_at", ascending: true)` ( this is ascending order, right?). The collection is in the right order (from nearest time to furthest time as below picture). But when the new notification come in (I add new object to Realm), it is put at the bottom of the collection.
<img width="314" alt="screen shot 2017-06-26 at 8 18 11 am" src="https://user-images.githubusercontent.com/7514128/27521528-1cfa4072-5a48-11e7-9c24-bdfd6490e56f.png">

Vice versa, when I put `realm.objects().sorted(byKeyPath: "created_at", ascending: false)` (descending), the collection is reversed by time as below second picture (i don't want it show from farthest to nearest time). But when the new notification come, it is put at the first row (I want it).
<img width="320" alt="screen shot 2017-06-26 at 8 27 32 am" src="https://user-images.githubusercontent.com/7514128/27521613-51682256-5a49-11e7-9758-45f1819ba745.png">

So could you tell me what was wrong here, I want the collection show from nearest time to furthest time, and the new object put at the first order. Thanks.

nirinchev:
It's not immediately obvious from your screenshots if the values are in the past or in the future (e.g. 2 months ago or 2 months from now). I would assume in the past, but then the ordering doesn't make sense - for example if I have 3 dates:
1. June 5, 2017
2. June 15, 2017
[...]

PR #7572, opened 2021-12-15
Upgrade core to 10.7.0

ericjordanmossman:
Upgrade core to 10.7.0.

Changelog notes:
I didn't include the core "Internal" section in the changelog because they are internal.

I didn't include the following core change "Fix" notes:

The client reset callbacks now pass out SharedRealm objects instead of TransactionRefs. (#5048, since v11.5.0)
    - Previous transactionRef format hadn't been exposed to the developer yet.

A client reset in DiscardLocal mode would revert to Manual mode on the next restart of the session. (#5050, since v11.5.0)
    - DiscardLocal hadn't been exposed in realm-cocoa yet.

A client reset in DiscardLocal mode would assert if the server added embedded object tables. (#5069, since v11.5.0)
    - DiscardLocal hadn't been exposed in realm-cocoa yet.

The client reset callbacks have changed so that the pre and post Realm state are passed to the 'after' callback and thee 'before' callback only has the local state. (#5066, since 11.5.0).
    - Before and after callbacks hadn't been exposed in realm-cocoa yet.

In the C-API, query count() did not apply descriptors such as limit/distinct. (#5073, since the beginning of the C-API in v10.4.0)
    - C-API


tgoyne:
11.7.0, not 10.7.0.

The 11.7.0 release is broken and is missing a bunch of headers so the non-spm builds don't work.

PR #7577, opened 2021-12-17
Don't declare all Objects as conforming to Identifiable

tgoyne:
Making Object conform to Identifiable results in an invalid default implementation for it being supplied, which is not overridden by the correct one from ObjectKeyIdentifiable on subclasses.

No changelog entry since we haven't released this yet.


pavel-ship-it (approved):
LGTM

jsflax (approved):

PR #7590, opened 2021-12-25
Add Flexible Sync

dianaafanador3:
Add Support for Flexible Sync

**Objective-C**

```objective-c
// Obtain a flexible sync configuration
RLMRealmConfiguration *config = [user flexibleSyncConfiguration];
// Open a realm with a flexible sync configuration
RLMRealm *realm = [RLMRealm realmWithConfiguration:config error:nil];

// Get the current Subscription Set
RLMSyncSubscriptionSet *subs = realm.subscriptions;

// All `Subscription`s transactions (Add/Remove/Update) should be done within a write transaction
[subs write:^{
        // Add a Subscription to the Subscription Set
        [subs addSubscriptionWithClassName:Person.className
                          subscriptionName:@"initial_subscription"
                                     where:@"age > 15"];;
}];

// You can query a `Subscription` (By name , or query)
RLMSyncSubscription *foundSubscription = [subs subscriptionWithName:@"person_older_15"];

// If you want to wait for the server synchronisation to be complete, you can use `onComplete`
[subs write:^{
        // Update a subscription
        [foundSubscription updateSubscriptionWithClassName:Person.className where:@"age > 20"]

        // Remove a subscription (By name, query or subscription)
        [subs removeSubscriptionWithName:@"initial_subscription"];

       // Or remove several `Subscription`s by type 
       [subs removeAllSubscriptionsWithClassName:Person.className];

       // Or remove  all the `Subscription`s on the Subscription Set
       [subs removeAllSubscriptions];
} onComplete: ^(NSError *error){
        if (error == nil) {
           // Synchronisation completed
        }
}];
```

**Swift**

```swift
// Obtain a flexible sync configuration
let config = user.flexibleSyncConfiguration()
// Open a realm with a flexible sync configuration

// Get the current Subscription Set
let subscriptions = realm.subscriptions

// All `Subscription`s transactions (Add/Remove/Update) should be done within a write transaction
subscriptions.write {
           // Append a `Subscription` or several `Subscription`s for the same object type to the Subscription Set 
            subscriptions.append {
                QuerySubscription<SwiftPerson>(name: "person_age_15") {
                    $0.age > 15
                }
                QuerySubscription<SwiftPerson>(name: "person_age_20") {
                    $0.age > 20
                }
            }
}

// You can query a `Subscription` (By name , or query)
let foundSubscription1 = subscriptions.first(named: "person_age_15")

// If you want to wait for the server synchronisation to be complete, you can use `onComplete`
subscriptions.write({
           // Update a subscription
           foundSubscription1?.update { QuerySubscription<SwiftPerson> { $0.age > 0 } }

           // Remove a subscription (By name, query or subscription)
           subscriptions.remove(named: "person_firstname_james")
}, onComplete: { error in
            if error == nil {
                 // Synchronisation completed
            }
})

// Or you can use our `await` API
try await subscriptions.write {
               // Or remove several `Subscription`s by type 
                subscriptions.removeAll(ofType: SwiftTypesSyncObject.self)
               // Or remove  all the `Subscription`s on the Subscription Set
                subscriptions.removeAll(ofType: SwiftPerson.self)
}
```


Package.swift
```diff
@@ -221,7 +221,8 @@ let package = Package(
                 "Realm/RLMSyncUtil.mm",
                 "Realm/RLMUpdateResult.mm",
                 "Realm/RLMUser.mm",
-                "Realm/RLMUserAPIKey.mm"
+                "Realm/RLMUserAPIKey.mm",
+                "Realm/RLMSyncSubscription.mm"
```
tgoyne, line 225: New files should be added in sorted order.

Realm/RLMObject.h
```diff
@@ -812,4 +812,6 @@ __attribute__((deprecated("RLM_ARRAY_TYPE has been deprecated. Use RLM_COLLECTIO
 @protocol RLM_OBJECT_SUBCLASS <NSObject>   \
 @end
 
+@protocol RLMAbstract
```
tgoyne, line 815: Appears to be unused?

Realm/RLMRealm.h
```diff
@@ -695,6 +695,18 @@ NS_REFINED_FOR_SWIFT;
...
+#pragma mark - Sync Subscriptions
+
+/**
+ Returns an instance of `SyncSubscriptionSet`, representing the active subscriptions
+ for this realm, which can be used to add/remove/update and search flexible sync subscriptions.
+ Getting the subscriptions from a local or partition-based configured realm will return a nil subscription
+ set.
+
+ @return A `RLMSyncSubscriptionSet`.
+ */
```
tgoyne, line 707: This is written as documentation for a method which returns a value rather than a property.

Realm/RLMRealm.mm
```diff
@@ -1043,4 +1045,29 @@ - (void)detachAllEnumerators {
...
+}
+
+- (RLMSyncSubscriptionSet *)subscriptionsWithError:(NSError **)error {
+    if (_realm->config().sync_config) {
+        if (_realm->config().sync_config->flx_sync_requested) {
+#if REALM_ENABLE_SYNC
+            return [[RLMSyncSubscriptionSet alloc] initWithSubscriptionSet:_realm->get_latest_subscription_set() realm:self];
+#else
+            NSLog(@"Realm was not built with sync enabled");
+            return NULL;
```
tgoyne, line 1059: `nil` for obj-c pointers.

Realm/RLMRealm.mm
```diff
@@ -1043,4 +1045,29 @@ - (void)detachAllEnumerators {
...
+    return [self subscriptionsWithError:nil];
+}
+
+- (RLMSyncSubscriptionSet *)subscriptionsWithError:(NSError **)error {
+    if (_realm->config().sync_config) {
+        if (_realm->config().sync_config->flx_sync_requested) {
+#if REALM_ENABLE_SYNC
+            return [[RLMSyncSubscriptionSet alloc] initWithSubscriptionSet:_realm->get_latest_subscription_set() realm:self];
+#else
+            NSLog(@"Realm was not built with sync enabled");
```
tgoyne, line 1058: If this is erroneous then it should throw an exception, and if it's fine then it shouldn't log anything.

Realm/RLMRealmConfiguration+Sync.mm
```diff
@@ -53,8 +53,11 @@ - (void)setSyncConfiguration:(RLMSyncConfiguration *)syncConfiguration {
 
     if (syncConfiguration.customFileURL) {
         self.config.path = syncConfiguration.customFileURL.path.UTF8String;
+    } else if (self.config.sync_config->flx_sync_requested) {
+        std::string str = std::string([@"default" UTF8String]);
```
tgoyne, line 57: This is a really roundabout way of writing `"default"`.

Can a single app support both flexible sync and partition-based sync? If so using a legal partition name as the name of the flexible sync Realm seems like an issue.

Realm/RLMSyncConfiguration_Private.h
```diff
@@ -34,11 +34,17 @@ typedef RLM_CLOSED_ENUM(NSUInteger, RLMSyncStopPolicy) {
               partitionValue:(nullable id<RLMBSON>)partitionValue
                   stopPolicy:(RLMSyncStopPolicy)stopPolicy;
 
+- (instancetype)initWithUser:(RLMUser *)user
+                  stopPolicy:(RLMSyncStopPolicy)stopPolicy
+              isFlexibleSync:(BOOL)isFlexibleSync;
```
tgoyne, line 39: `enableFlexibleSync:` seems like a better name when setting the config.

Realm/RLMSyncConfiguration.mm
```diff
@@ -115,35 +115,57 @@ - (void)setCancelAsyncOpenOnNonFatalErrors:(bool)cancelAsyncOpenOnNonFatalErrors
...
-        _config = std::make_unique<SyncConfig>(
-            [user _syncUser],
-            s.str()
-        );
+        if (isFlexibleSync) {
+            _config = std::make_unique<SyncConfig>([user _syncUser], SyncConfig::FLXSyncEnabled{});
+        } else {
+            std::stringstream s;
+            s << RLMConvertRLMBSONToBson(partitionValue);
+            _config = std::make_unique<SyncConfig>(
```
tgoyne, line 164: Newline here is weird.

Realm/RLMSyncSubscription.h
```diff
@@ -0,0 +1,336 @@
...
+//
+////////////////////////////////////////////////////////////////////////////
+
+#import <Foundation/Foundation.h>
+
+@class RLMObjectId;
+
+#pragma mark - Subscription States
+
+typedef NS_ENUM(NSUInteger, RLMSyncSubscriptionState) {
```
tgoyne, line 25: Enum itself needs docs too.

Realm/RLMSyncSubscription.h
```diff
@@ -0,0 +1,336 @@
...
+    /// still use the current subscription set to write a subscription.
+    RLMSyncSubscriptionStateError,
+    /// The subscription is persisted locally but not yet processed by the server, which means
+    /// the server hasn't yet returned all the data that matched the updated subscription queries.
+    RLMSyncSubscriptionStatePending,
+    /// The subscription set has been super-ceded by an updated one, this typically means that
+    /// someone is trying to write a subscription on a different instance of the subscription set.
+    /// You should not use a super-ceded subscription set and instead obtain a new instance of
+    /// the subscription set to write a subscription.
+    RLMSyncSubscriptionStateSuperceded
```
tgoyne, line 41: superseded

Realm/RLMSyncSubscription.h
```diff
@@ -0,0 +1,336 @@
...
+};
+
+NS_ASSUME_NONNULL_BEGIN
+
+@interface RLMSyncSubscription : NSObject
```
tgoyne, line 46: Each of these classes need docs.

Realm/RLMSyncSubscription.h
```diff
@@ -0,0 +1,336 @@
...
+                              predicate:(NSPredicate *)predicate;
+
+@end
+
+@interface RLMSyncSubscriptionSet : NSObject <NSFastEnumeration>
+
+/// The number of subscriptions in the subscription set.
+@property (readonly) NSUInteger count;
+
+/// Gets the error associated to the subscription set, this will return a no nil in case the current
```
tgoyne, line 96: This comma should be a period. Same thing in the next few docs as well.

"This will be non-nil in case"

Realm/RLMSyncSubscription.h
```diff
@@ -0,0 +1,336 @@
...
+- (void)write:(__attribute__((noescape)) void(^)(void))block;
+
+/**
+ Synchronously performs any transactions (add/remove/update) to the subscription set within the block,
+ this will not wait for the server to acknowledge and see all the data associated with this collection of subscriptions,
+ and will return after committing the subscription transactions.
+
+ @param block The block containing actions to perform to the subscription set.
+ */
+- (void)write:(__attribute__((noescape)) void(^)(void))block onComplete:(void(^)(NSError**))block;
```
tgoyne, line 122: Need docs for onComplete.

Realm/RLMSyncSubscription.h
```diff
@@ -0,0 +1,336 @@
...
+    /// the subscription set to write a subscription.
+    RLMSyncSubscriptionStateSuperceded
+};
+
+NS_ASSUME_NONNULL_BEGIN
+
+@interface RLMSyncSubscription : NSObject
+
+/// Identifier of the subscription.
+@property (nonatomic, readonly) RLMObjectId *identifier;
```
tgoyne, line 49: There doesn't appear to be anything you can actually do with this identifier. I'd expect there to be a lookup function using it and maybe for adding a subscription to return the identifier.

Realm/RLMSyncSubscription.h
```diff
@@ -0,0 +1,336 @@
...
+/**
+ Updates a Flexible Sync's subscription query with an allowed query which will be used to bootstrap data
+ from the server when committed.
+
+ @warning This method may only be called during a write subscription block.
+
+ @param objectClassName The class name for the model class to be queried.
+ @param predicateFormat A predicate format string, optionally followed by a variable number of arguments.
+ */
+- (void)updateSubscriptionWithClassName:(NSString *)objectClassName
```
tgoyne, line 69: I'm confused by this API. This seems to suggest that a Subscription has a query on one or more classes? Or exactly one class at a time, and this changes what class the Subscription is for?

Realm/RLMSyncSubscription.mm
```diff
@@ -0,0 +1,483 @@
...
+    if (self = [super init]) {
+        _subscription = std::make_unique<realm::sync::Subscription>(subscription);
+        _subscriptionSet = subscriptionSet;
+        return self;
+    }
+    return nil;
+}
+
+- (RLMObjectId *)identifier {
+    return [[RLMObjectId alloc]initWithValue:_subscription->id()];
```
tgoyne, line 50:
```suggestion
    return [[RLMObjectId alloc] initWithValue:_subscription->id()];
```

Realm/RLMSyncSubscription.mm
```diff
@@ -0,0 +1,483 @@
...
+- (nullable NSString *)name {
+    const std::string_view str_view = _subscription->name();
+    std::string str = std::string(str_view);
+    const char * characters = str.c_str();
+    NSString *name = [NSString stringWithCString:characters
+                                        encoding:[NSString defaultCStringEncoding]];
+    if ([name length] == 0) {
+        return NULL;
+    }
+    return name;
```
tgoyne, line 62:
```suggestion
    auto strView = _subscription->name();
    if (strView.size() == 0) {
        return nil;
    }
    return [[NSString alloc] initWithBytes:strView.data()
                                    length:strView.size()
                                  encoding:NSUTF8StringEncoding];
```

Converting the std::string_view to a std::string doesn't do anything useful. `[NSString defaultCStringEncoding]` does complicated things which are incorrect here; strings passed to and from core must always be UTF-8. Passing the size is required to support embedded nuls in the name; if we don't want to support that then it can be just `return @(strView.data());`

Realm/RLMSyncSubscription.mm
```diff
@@ -0,0 +1,483 @@
...
+- (NSDate *)createdAt {
+    return RLMTimestampToNSDate(_subscription->created_at());
+}
+
+- (NSDate *)updatedAt {
+    return RLMTimestampToNSDate(_subscription->updated_at());
+}
+
+- (NSString *)queryString {
+    return [NSString stringWithCString:std::string(_subscription->query_string()).c_str()
```
tgoyne, line 74: See above notes. Since we're converting std::string_view to NSString in many places, we want a std::string_view version of RLMStringDataToNSString().

Realm/RLMSyncSubscription.mm
```diff
@@ -0,0 +1,483 @@
...
+
+- (NSString *)objectClassName {
+    return [NSString stringWithCString:std::string(_subscription->object_class_name()).c_str()
+                              encoding:[NSString defaultCStringEncoding]];
+}
+
+- (void)updateSubscriptionWithClassName:(NSString *)objectClassName
+                                  where:(NSString *)predicateFormat, ... {
+    va_list args;
+    va_start(args, predicateFormat);
```
tgoyne, line 86: va_start must always be paired with va_end.

Realm/RLMSyncSubscription.mm
```diff
@@ -0,0 +1,483 @@
...
+                                  where:(NSString *)predicateFormat
+                                   args:(va_list)args {
+    [self updateSubscriptionWithClassName:objectClassName
+                                predicate:[NSPredicate predicateWithFormat:predicateFormat arguments:args]];
+    
+}
+
+- (void)updateSubscriptionWithClassName:(NSString *)objectClassName
+                              predicate:(NSPredicate *)predicate {
+    [_subscriptionSet verifyInWriteTransaction];
```
tgoyne, line 102: Shouldn't the RLMSubscriptionSet method be checking this?

Realm/RLMSyncSubscription.mm
```diff
@@ -0,0 +1,483 @@
...
+#pragma mark - SubscriptionSet
+
+@interface RLMSyncSubscriptionSet () {
+    std::unique_ptr<realm::sync::SubscriptionSet> _subscriptionSet;
+    std::unique_ptr<realm::sync::MutableSubscriptionSet> _mutableSubscriptionSet;
+}
+@end
+
+@implementation RLMSyncSubscriptionSet {
+    BOOL isInWriteTransaction;
```
tgoyne, line 120: Ivar names should have an underscore prefix.

Realm/RLMSyncSubscription.mm
```diff
@@ -0,0 +1,483 @@
...
+    RLMRealm *_realm;
+    id _strongBuffer[16];
```
tgoyne, line 122: While it's admittedly a very silly thing to do, storing this here means that nested for loops on a single RLMSyncSubscriptionSet object will break.

Realm/RLMSyncSubscription.mm
```diff
@@ -0,0 +1,483 @@
...
+}
+
+- (instancetype)initWithSubscriptionSet:(realm::sync::SubscriptionSet)subscriptionSet
+                                  realm:(RLMRealm *)realm {
+    if (self = [super init]) {
+        _subscriptionSet = std::make_unique<realm::sync::SubscriptionSet>(subscriptionSet);
+        _realm = realm;
+        self->isInWriteTransaction = false;
```
tgoyne, line 130: Ivars in obj-c objects are automatically initialized to 0/false/nil.

Realm/RLMSyncSubscription.mm
```diff
@@ -0,0 +1,483 @@
...
+
+- (nullable NSError *)error {
+    NSString *errorMessage = RLMStringDataToNSString(_subscriptionSet->error_str());
+    if ([errorMessage length] == 0) {
+        return NULL;
+    }
+    return [[NSError alloc]initWithDomain:RLMSyncErrorDomain
+                                               code:0
+                                           userInfo:@{NSLocalizedDescriptionKey : errorMessage}];
+}
```
tgoyne, line 148:
```suggestion
- (nullable NSError *)error {
    NSString *errorMessage = RLMStringDataToNSString(_subscriptionSet->error_str());
    if (errorMessage.length == 0) {
        return nil;
    }
    return [[NSError alloc] initWithDomain:RLMSyncErrorDomain
                                      code:0
                                  userInfo:@{NSLocalizedDescriptionKey: errorMessage}];
}
```

Realm/RLMSyncSubscription.mm
```diff
@@ -0,0 +1,483 @@
...
+
+#pragma mark - Batch Update subscriptions
+
+- (void)write:(__attribute__((noescape)) void(^)(void))block {
+    return [self write:block onComplete:^(NSError**){}];
+}
+
+- (void)write:(__attribute__((noescape)) void(^)(void))block
+   onComplete:(void(^)(NSError**))completionBlock {
+    [self secureWrite];
```
tgoyne, line 173: "secure" is a strange word to use here.

Realm/RLMSyncSubscription.mm
```diff
@@ -0,0 +1,483 @@
...
+    _mutableSubscriptionSet = std::make_unique<realm::sync::MutableSubscriptionSet>(_subscriptionSet->make_mutable_copy());
+    self->isInWriteTransaction = true;
+    block();
+    try {
+        _subscriptionSet = std::make_unique<realm::sync::SubscriptionSet>(std::move(*_mutableSubscriptionSet).commit());
+        _mutableSubscriptionSet = nullptr;
+        self->isInWriteTransaction = false;
+    }
+    catch (std::exception error) {
```
tgoyne, line 182: Exceptions need to always be caught by reference or it'll do very bad things.

Realm/RLMSyncSubscription.mm
```diff
@@ -0,0 +1,483 @@
...
+        NSError *err = [[NSError alloc] initWithDomain:@"subscription_set" code:-1 userInfo:@{@"reason":@(error.what())}];
```
tgoyne, line 183: `@"subscription_set"` is not a sensible error domain to be using. The error domain should always be a constant exposed as part of the API, and the string chosen needs to have a RLM prefix.

Realm/RLMSyncSubscription.h
tgoyne: `NSError **` is used for out parameters, which this is not.

Realm/RLMSyncSubscription.mm
```diff
@@ -0,0 +1,483 @@
...
+- (BOOL)writeAsync:(__attribute__((noescape)) void(^)(void))block
+          callback:(RLMSyncSubscriptionCallback)callback {
+    [NSException raise:@"NotImplemented" format:@"Needs Implementation"];
+    return NULL;
+}
+
+#pragma mark - Find subscription
+
+- (nullable RLMSyncSubscription *)subscriptionWithName:(NSString *)name {
+    std::string str = std::string([name UTF8String]);
```
tgoyne, line 209: Converting this to a std::string rather than passing `name.UTF8String` directly to `find()` is just a pessimization.

Realm/RLMSyncSubscription.mm
```diff
@@ -0,0 +1,483 @@
...
+- (nullable RLMSyncSubscription *)subscriptionWithClassName:(NSString *)objectClassName
+                                                  predicate:(NSPredicate *)predicate {
+    RLMClassInfo& info = _realm->_info[objectClassName];
+    auto query = RLMPredicateToQuery(predicate, info.rlmObjectSchema, _realm.schema, _realm.group);
+    auto iterator = _subscriptionSet->find(query);
+    if (iterator != _subscriptionSet->end()) {
+        return [[RLMSyncSubscription alloc] initWithSubscription:*iterator
+                                                 subscriptionSet:self];
+    }
+    return NULL;
```
tgoyne, line 244:
```suggestion
    if (auto it = _subscriptionSet->find(query); it != _subscriptionSet->end()) {
        return [[RLMSyncSubscription alloc] initWithSubscription:*it
                                                 subscriptionSet:self];
    }
    return nil;
```

`it` is the idiomatic name for an iterator variable when there's just one and you don't need to distinguish different iterators from each other. When possible the find should be inside the if, and `nil` should be used for obj-c pointers.

Realm/RLMSyncSubscription.mm
```diff
@@ -0,0 +1,483 @@
...
+    
+    if (name != nil) {
+        std::string str = std::string([name UTF8String]);
+        auto iterator = _mutableSubscriptionSet->find(str);
+        
+        if (iterator == _mutableSubscriptionSet->end() || updateExisting) {
+            _mutableSubscriptionSet->insert_or_assign(str, query);
+        }
+        else {
+            // Should this throw??
```
tgoyne, line 321: This should either throw or return a value indicating whether or not it actually added a new subscription.

Realm/RLMSyncSubscription.mm
```diff
@@ -0,0 +1,483 @@
...
+    [self verifyInWriteTransaction];
+
+    for (auto it = _mutableSubscriptionSet->begin(); it != _mutableSubscriptionSet->end();) {
+        if (it->id() == subscription.identifier.value) {
+            it = _mutableSubscriptionSet->erase(it);
+            return;
+        }
+        else {
+            it++;
+        }
```
tgoyne, line 380: This can just be in the `for`.

Realm/RLMSyncSubscription.mm
```diff
@@ -0,0 +1,483 @@
...
+- (void)removeAllSubscriptions {
+    [self verifyInWriteTransaction];
+    _mutableSubscriptionSet->clear();
+}
+
+- (void)removeAllSubscriptionsWithClassName:(NSString *)className {
+    [self verifyInWriteTransaction];
+    
+    for (auto it = _mutableSubscriptionSet->begin(); it != _mutableSubscriptionSet->end();) {
+        if (it->object_class_name() == std::string([className UTF8String])) {
```
tgoyne, line 395: Converting this to std::string is a pessimization.

Realm/RLMSyncSubscription.mm
```diff
@@ -0,0 +1,483 @@
...
+
+#pragma mark - NSFastEnumerator
+
+- (NSUInteger)countByEnumeratingWithState:(NSFastEnumerationState *)state
+                                  objects:(__unused __unsafe_unretained id [])buffer
+                                    count:(NSUInteger)len {
+    NSUInteger batchCount = 0, count = _subscriptionSet->size();
+    for (NSUInteger index = state->state; index < count && batchCount < len; ++index) {
+        auto iterator = _subscriptionSet->at(size_t(index));
+        RLMSyncSubscription *subscription = [[RLMSyncSubscription alloc]initWithSubscription:iterator
```
tgoyne, line 412: There's a lot of places missing spaces after ]

RealmSwift.podspec
```diff
@@ -26,7 +26,8 @@ Pod::Spec.new do |s|
 
   s.dependency 'Realm', "= #{s.version}"
   s.source_files = 'RealmSwift/*.swift', 'RealmSwift/Impl/*.swift'
-  s.exclude_files = 'RealmSwift/Nonsync.swift'
+  s.exclude_files = 'RealmSwift/Nonsync.swift',
+                    'RealmSwift/SyncSubscription.swift'
```
tgoyne, line 30: Why is this excluded?

Realm/RLMObject.h
jsflax: Oops, test code for something completely unrelated.

Realm/RLMRealmConfiguration+Sync.mm
jsflax: No, a single app cannot support both and I believe there is no intention for that to ever be the case.

Realm/RLMUser_Private.hpp
```diff
@@ -45,7 +45,7 @@ class CocoaSyncUserContext : public realm::SyncUserContext {
 
 @interface RLMUser ()
 - (instancetype)initWithUser:(std::shared_ptr<realm::SyncUser>)user app:(RLMApp *)app;
-- (std::string)pathForPartitionValue:(std::string const&)partitionValue;
+- (std::string)pathForValue:(std::string const&)partitionValue;
```
leemaguire, line 48: `pathForPartitionValue` sounds more descriptive?

RealmSwift/Realm.swift
```diff
@@ -960,6 +960,26 @@ import Realm.Private
     }
 }
 
+// MARK: Sync Subscriptions
+
+extension Realm {
+    /**
+     Returns an instance of `SyncSubscriptionSet`, representing the active subscriptions
```
leemaguire, line 967: bad indent

RealmSwift/SyncSubscription.swift
```diff
@@ -0,0 +1,397 @@
...
+    /// queries at the time the subscription set was updated. The server is now in a steady-state
+    /// synchronization mode where it will stream update as they come.
+    case complete
+    /// The subscription encountered an error and synchronization is paused for this Realm. You can
+    /// still use the current subscription set to write a subscription.
+    case error(Error)
+    /// The subscription is persisted locally but not yet processed by the server, which means
+    /// the server hasn't yet returned all the data that matched the updated subscription queries.
+    case pending
+    /// The subscription set has been super-ceded by an updated one, this typically means that
```
leemaguire, line 35:
```suggestion
    /// The subscription set has been superseded by an updated one, this typically means that
```

RealmSwift/SyncSubscription.swift
```diff
@@ -0,0 +1,397 @@
...
+    public func remove(named: String) {
+        rlmSyncSubscriptionSet.removeSubscription(withName: named)
+    }
+
+    /**
+     Removes all subscriptions from the subscription set.
+
+     - warning: This method may only be called during a write subscription block.
+     - warning: Removing all subscriptions will result in an error if no new subscription is added. Server should
+              acknowledge at least one subscription.
```
leemaguire, line 299: bad indent

Realm/RLMRealm.mm
```diff
@@ -56,6 +57,7 @@
 
 #import <realm/object-store/sync/async_open_task.hpp>
 #import <realm/object-store/sync/sync_session.hpp>
+#import <realm/object-store/shared_realm.hpp>
```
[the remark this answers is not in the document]
jsflax: Not sure I understand this suggestion.

RealmSwift/Realm.swift
jsflax: Xcode disagrees.

Realm/RLMRealm.mm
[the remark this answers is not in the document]
leemaguire: It was just about ordering, leaving the more nested files lower down

Realm/RLMSyncSubscription.h
jsflax: cc @dianaafanador3

Realm/RLMSyncSubscription.h
jsflax: cc @dianaafanador3

Realm/RLMSyncSubscription.mm
jsflax: cc @dianaafanador3

Realm/RLMSyncSubscription.h
dianaafanador3: A subscription only have one associated class at a time, it doesn't mean that this class cannot be changed. We can update a subscription with a query associated to a different class that the previous one. For example you can have a subscription named `initial_query` which can have an initial query `Project.id == 'XXXXXX'` which can be updated to `User.id == 'XXXXXX'`, which is using a different object class for the same subscription.

Realm/RLMSyncSubscription.mm
dianaafanador3: yep you are right, this is already been verified by the `RLMSubscriptionSet`

Realm/RLMSyncSubscription.mm
dianaafanador3: Fixed

Package.swift
```diff
@@ -122,7 +122,7 @@ let package = Package(
             targets: ["Realm", "RealmSwift"]),
     ],
     dependencies: [
-        .package(name: "RealmDatabase", url: "https://github.com/realm/realm-core", .exact(Version(coreVersionStr)!))
+        .package(name: "RealmDatabase", url: "https://github.com/realm/realm-core", .branch("master"))
```
leemaguire, line 125: Has core not released flexible sync yet?

Realm/RLMRealmConfiguration+Sync.mm
```diff
@@ -53,6 +53,8 @@ - (void)setSyncConfiguration:(RLMSyncConfiguration *)syncConfiguration {
 
     if (syncConfiguration.customFileURL) {
         self.config.path = syncConfiguration.customFileURL.path.UTF8String;
+    } else if (self.config.sync_config->flx_sync_requested) {
+        self.config.path = [user pathForPartitionValue:"default"];
```
leemaguire, line 57: It would be good to have a descriptive constant here instead of `"default"`

Realm/RLMSyncManager.mm
```diff
@@ -133,6 +132,7 @@ + (SyncClientConfig)configurationWithRootDirectory:(NSURL *)rootDirectory appId:
         config.user_agent_application_info = RLMStringDataWithNSString(appId);
     }
 
+    config.log_level = Level::trace;
```
leemaguire, line 135: Is the log level intentionally set to trace?

Realm/RLMSyncSubscription.mm
```diff
@@ -0,0 +1,528 @@
...
+    _mutableSubscriptionSet = std::make_unique<realm::sync::MutableSubscriptionSet>(_subscriptionSet->make_mutable_copy());
+    self->_isInWriteTransaction = true;
+    block();
+    try {
+        _subscriptionSet = std::make_unique<realm::sync::SubscriptionSet>(std::move(*_mutableSubscriptionSet).commit());
+        _mutableSubscriptionSet = nullptr;
+        self->_isInWriteTransaction = false;
+    }
+    catch (const std::exception& error) {
+        NSError *err = [[NSError alloc] initWithDomain:RLMFlexibleSyncErrorDomain code:-1 userInfo:@{@"reason":@(error.what())}];
```
leemaguire, line 257: We should list out error codes instead of passing back `-1`

Realm/RLMSyncSubscription.mm
```diff
@@ -0,0 +1,528 @@
...
+                completionBlock(error);
+            }
+        });
+}
+
+typedef void(^RLMSyncSubscriptionCallback)(NSError * _Nullable error);
+
+- (BOOL)writeAsync:(__attribute__((noescape)) void(^)(void))block
+          callback:(RLMSyncSubscriptionCallback)callback {
+    [NSException raise:@"NotImplemented" format:@"Needs Implementation"];
```
leemaguire, line 275: Do we plan to implement this for the initial release?

Realm/RLMUser.h
```diff
@@ -100,13 +100,25 @@ NS_ASSUME_NONNULL_BEGIN
...
- Create a query-based configuration instance for the given url.
+ Create a partition-based sync configuration instance for the given partition value.
 
  @param partitionValue The `RLMBSON` value the Realm is partitioned on.
  @return A default configuration object with the sync configuration set to use the given partition value.
  */
 - (RLMRealmConfiguration *)configurationWithPartitionValue:(nullable id<RLMBSON>)partitionValue NS_REFINED_FOR_SWIFT;
 
+/**
+ Create a flexible sync configuration instance, which can be used to open a Realm with a which
```
leemaguire, line 111:
```suggestion
 Create a flexible sync configuration instance, which can be used to open a Realm that
```

RealmSwift/Realm.swift
```diff
@@ -960,6 +960,24 @@ import Realm.Private
...
+     for this realm, which can be used to add/remove/update and search flexible sync subscriptions.
+     Getting the subscriptions from a local or partition-based configured realm will return a nil subscription
```
leemaguire, line 969: nil or empty because this isn't declared as optional?

RealmSwift/SyncSubscription.swift
```diff
@@ -0,0 +1,395 @@
...
+    public func removeAll<T: Object>(ofType type: T.Type) {
+        rlmSyncSubscriptionSet.removeAllSubscriptions(withClassName: type.className())
+    }
+
+    // MARK: Subscription Retrieval
+
+    /**
+     Returns the subscription at the given `index`.
+
+     - parameter index: The index.
```
leemaguire, line 319: Parameter name is incorrect, also the doc could be a bit more descriptive.

Realm/RLMSyncSubscription.mm
dianaafanador3: This is not included on the scope for this first beta release

Realm/RLMRealmConfiguration+Sync.mm
dianaafanador3: I changed this to use the path obtained from core for flexible sync

Package.swift
dianaafanador3: Changed to use latest core release

Realm/RLMUser.mm
```diff
@@ -132,6 +141,15 @@ - (void)invalidate {
     return path;
 }
 
+- (std::string)pathForFlexibleSync {
+    if (!_user) {
+        return "";
```
leemaguire, line 146: this should really be returning an optional string

Realm/RLMSyncSubscription.mm
leemaguire: We should just omit this for now and add it back on the next version on flex sync

Realm/RLMUser.mm
jsflax: Or this can be an exception, as this is arguably an exceptional state.

Realm/ObjectServerTests/RLMFlexibleSyncServerTests.mm
```diff
@@ -0,0 +1,1014 @@
...
+
+    XCTAssertEqual(subs.version, 1UL);
+    XCTAssertEqual(subs.count, (unsigned long)numberOfSubs);
+
+    __weak id objects[(unsigned long)pow(numberOfSubs, 2.0) + (unsigned long)numberOfSubs];
+    NSInteger count = 0;
+    for (RLMSyncSubscription *sub in subs) {
+        XCTAssertNotNil(sub);
+        objects[count++] = sub;
+        for (RLMSyncSubscription *sub in subs) {
```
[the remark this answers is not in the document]
jsflax: The nested for-loop tests that we implemented NSFastEnumeration correctly. It was previously implemented incorrectly and a nested for-loop would have broken it. The use of `pow` was to dynamically calculate count since the loop is O(n²).

Realm/ObjectServerTests/RLMObjectServerTests.mm
```diff
@@ -700,18 +639,18 @@ - (void)testAddObjects {
 - (void)testAddObjectsWithNilPartitionValue {
     RLMRealm *realm = [self openRealmForPartitionValue:nil user:self.anonymousUser];
 
+    // Other tests expect the nil partition to be empty so we need to clean up
+    [realm transactionWithBlock:^{
```
leemaguire, line 643: I feel like this also needs to be called at the end of the test too?

Realm/RLMUser.mm
```diff
@@ -26,11 +26,13 @@
 #import "RLMRealmConfiguration+Sync.h"
 #import "RLMSyncConfiguration_Private.hpp"
 #import "RLMSyncSession_Private.hpp"
+#import "RLMUtil.hpp"
 
 #import <realm/object-store/sync/sync_manager.hpp>
 #import <realm/object-store/sync/sync_session.hpp>
 #import <realm/object-store/sync/sync_user.hpp>
 #import <realm/object-store/util/bson/bson.hpp>
+#import <realm/util/optional.hpp>
```
leemaguire, line 35: unused?

leemaguire (approved):
LGTM 🎉 I have left just a few small comments

Realm/ObjectServerTests/RLMObjectServerTests.mm
dianaafanador3: Other test with partition value null, are cleaning the database at the beginning of the test as well

Realm/RLMRealm.mm
```diff
@@ -53,7 +53,9 @@
 #import "RLMSyncManager_Private.hpp"
 #import "RLMSyncSession_Private.hpp"
 #import "RLMSyncUtil_Private.hpp"
+#import "RLMSyncSubscription_Private.hpp"
 
+#import <realm/object-store/shared_realm.hpp>
```
tgoyne, line 58: This is already imported above.

Realm/RLMSyncSubscription.mm
```diff
@@ -0,0 +1,523 @@
...
+    [_enumerators removeObject:enumerator];
+}
+
+- (void)registerEnumerator:(RLMSyncSubscriptionEnumerator *)enumerator {
+    std::lock_guard lock(_collectionEnumeratorMutex);
+    if (!_enumerators) {
+        _enumerators = [NSHashTable hashTableWithOptions:NSPointerFunctionsWeakMemory];
+    }
+    [_enumerators addObject:enumerator];
+}
```
tgoyne, line 207: There doesn't appear to be any reason for this? RLMRealm keeps track of its active enumerators because it needs to notify them when a write transaction is started, but `_enumerators` is never actually used here.

Realm/RLMSyncSubscription.mm
```diff
@@ -0,0 +1,523 @@
...
+    return [self write:block onComplete:^(NSError*){}];
+}
+
+- (void)write:(__attribute__((noescape)) void(^)(void))block
+   onComplete:(void(^)(NSError *))completionBlock {
+    if (self->_isInWriteTransaction) {
+        @throw RLMException(@"Cannot initiate a write transaction on subscription set that is already been updated.");
+    }
+    _mutableSubscriptionSet = std::make_unique<realm::sync::MutableSubscriptionSet>(_subscriptionSet->make_mutable_copy());
+    self->_isInWriteTransaction = true;
```
tgoyne, line 251: Having both `_mutableSubscriptionSet` and `_isInWriteTransaction` is redundant, and introduces the opportunity for bugs from them getting out of sync. We could instead check `_mutableSubscriptionSet` everywhere that we're currently checking `_isInWriteTransaction`.

Realm/RLMSyncSubscription.mm
```diff
@@ -0,0 +1,523 @@
...
+    auto query = RLMPredicateToQuery(predicate, info.rlmObjectSchema, _realm.schema, _realm.group);
+    
+    if (name != nil) {
+        auto iterator = _mutableSubscriptionSet->find([name UTF8String]);
+        
+        if (iterator == _mutableSubscriptionSet->end() || updateExisting) {
+            _mutableSubscriptionSet->insert_or_assign([name UTF8String], query);
+        }
+        else {
+            @throw RLMException(@"Cannot duplicate a subscription, if you meant to update the subscription please use the `update` method.");
```
tgoyne, line 388: comma should be a period.

Realm/RLMSyncSubscription.mm
```diff
@@ -0,0 +1,523 @@
...
+    [self verifyInWriteTransaction];
+    
+    RLMClassInfo& info = _realm->_info[objectClassName];
+    auto query = RLMPredicateToQuery(predicate, info.rlmObjectSchema, _realm.schema, _realm.group);
+    
+    if (name != nil) {
+        auto iterator = _mutableSubscriptionSet->find([name UTF8String]);
+        
+        if (iterator == _mutableSubscriptionSet->end() || updateExisting) {
+            _mutableSubscriptionSet->insert_or_assign([name UTF8String], query);
```
tgoyne, line 385: Doing the lookup only to ignore the result when updateExisting is true is some extra work.

Realm/RLMSyncSubscription.mm
```diff
@@ -0,0 +1,523 @@
...
+    if (iterator != _mutableSubscriptionSet->end()) {
+        _mutableSubscriptionSet->erase(iterator);
+    }
+}
+
+- (void)removeSubscriptionWithClassName:(NSString *)objectClassName
+                                  where:(NSString *)predicateFormat, ... {
+    va_list args;
+    va_start(args, predicateFormat);
+    [self removeSubscriptionWithClassName:objectClassName
```
tgoyne, line 411: missing va_end

Realm/RLMSyncSubscription.mm
```diff
@@ -0,0 +1,523 @@
...
+                            (unsigned long long)index, (unsigned long long)size);
+    }
+    
+    return [[RLMSyncSubscription alloc]initWithSubscription:_subscriptionSet->at(size_t(index))
+                                            subscriptionSet:self];
+}
+
+- (RLMSyncSubscription *)firstObject {
+    if (_subscriptionSet->size() < 1) {
+        return 0;
```
tgoyne, line 490: return `nil` when it's an obj-c pointer.

RealmSwift/SyncSubscription.swift
```diff
@@ -0,0 +1,415 @@
...
+        self.query = query
+    }
+}
+
+#if swift(>=5.5)
+/**
+ Result builder which allows to add more than one query to a subscription,
+ within an `append` or `remove` block.
+ */
+@resultBuilder public struct QueryBuilder {
```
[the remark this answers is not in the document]
jsflax: The idea was to allow them to add multiple subscriptions without having to call multiple appends, which... is less verbose than what you're saying? But I agree that passing in `Query` would be nicer as opposed to `QuerySubscription`.

Realm/RLMSyncSubscription.mm
jsflax: My mistake– I ported over the RLMRealm code assuming that holding onto the enumerators had something to do with holding onto the reference.

RealmSwift/SyncSubscription.swift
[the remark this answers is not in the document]
tgoyne: The current thing is just as verbose as just not doing it. It's replacing `subscriptions.append(QuerySubscription(...), QuerySubscription(...))` with `subscriptions.append { QuerySubscription(...) QuerySubscription(...) }`.

Realm/ObjectServerTests/RealmServer.swift
```diff
@@ -701,43 +712,83 @@ public class RealmServer: NSObject {
...
+        if case .pbs(let bsonType) = syncMode {
+            partitionKeyType = bsonType
+        } else {
+            partitionKeyType = nil
+        }
         var ruleCreations = [Result<Any?, Error>]()
         for objectSchema in syncTypes {
-            ruleCreations.append(rules.post(objectSchema.stitchRule(bsonType, schema)))
+            let validSyncClasses = ["Dog", "Person", "SwiftPerson", "SwiftTypesSyncObject"]
+            if partitionKeyType != nil ||
```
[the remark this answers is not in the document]
dianaafanador3: This will be possible if all the added objects were Swift Objects, but because is not the case I was not able to add this.I simplified the code as much as possible so we don't have to change this twice.

RealmSwift/SyncSubscription.swift
```diff
@@ -0,0 +1,459 @@
...
+     collection of subscription.
+
+     - parameter block: The block containing the subscriptions transactions to perform.
+
+     - throws: An `NSError` if the transaction could not be completed successfully.
+               If `block` throws, the function throws the propagated `ErrorType` instead.
+     */
+    public func write(_ block: (() -> Void)) async throws {
+        try await withCheckedThrowingContinuation { (continuation: CheckedContinuation<Void, Error>) in
+            lock.lock()
```
tgoyne, line 446: Locking here doesn't really do much. The most immediate problem is that every entry point needs locking, not just one, or calling `await subscriptions.write()` and then continuing to use the `subscriptions` object on the calling thread will break. Adding that locking still wouldn't work, though, because RLMSyncSubscriptionSet is ultimately backed by a thread-confined Realm, and so accessing it from other threads is illegal. Each of RLMSyncSubscriptionSet's public entry points really should be calling `[_realm verifyThread]`.

This currently will often work by coincidence because `withCheckedThrowingContinuation()` doesn't _guarantee_ that it hops to a different thread, but as soon as it does it'll break.

The simplest way to provide something that'll work for now is to just make it `@MainActor` and only support using this from the main thread. A better solution might be to re-obtain the Realm and SyncSubscriptionSet from within the continuation to get a valid thread-local one and pass it to the user's block.

CHANGELOG.md
```diff
@@ -17,7 +17,7 @@ x.y.z Release notes (yyyy-MM-dd)
 * Xcode: 12.4-13.2.1.
 
 ### Internal
-* Upgraded realm-core from ? to ?
+* Upgraded realm-core from 11.8.0 to 11.9.0
```
tgoyne, line 20: Also need changelog entries for the changes which come from bumping the core version.

RealmSwift/SyncSubscription.swift
dianaafanador3: I can just make this to run only on the main thread for today's release and fix this for a next release.
If we add the `@MainActor` annotation this will mean that everything we are doing on the writeOnComplete method will block the main thread, because the only thing I'm worried is if we are waiting for all the data to be synced and during this time the main thread is blocked. @tgoyne

CHANGELOG.md
```diff
@@ -1,11 +1,11 @@
 x.y.z Release notes (yyyy-MM-dd)
 =============================================================
 ### Enhancements
-* None.
+* Add beta support for flexible sync. See the [backend](https://docs.mongodb.com/realm/sync/data-access-patterns/flexible-sync/) and [SDK](https://docs.mongodb.com/realm/sdk/swift/examples/flexible-sync/) documentation for more information. Please report any issues with the beta through Github.
 
 ### Fixed
-* <How to hit and notice issue? what was the impact?> ([#????](https://github.com/realm/realm-swift/issues/????), since v?.?.?)
-* None.
+* UserIdentity metadata table grows indefinitely. ([#5152](https://github.com/realm/realm-core/issues/5152), since v10.0.0)
```
tgoyne, line 7: The bug being fixed here was introduced in v10.20.0.

CHANGELOG.md
```diff
@@ -1,11 +1,11 @@
...
+* Improved error messaging when opening a Realm with `IncompatibleHistories` when translating file exceptions ([#5161](https://github.com/realm/realm-core/pull/5161), since v6.0.0).
```
tgoyne, line 8: v6.0.0 doesn't exist. This was probably actually v5.0.0.

This changelog entry doesn't really mean anything to users of Realm. The user-visible change here is that we once again report a useful error message when opening a sync Realm in non-sync mode or vice-versa.

PR #7613, opened 2022-01-14
Add ability to export local Realm to synced Realm

leemaguire:
This PR adds `Realm.writeCopy(configuration:)`/`[RLMRealm writeCopyForConfiguration:]` which allows you to export a given instance of a local Realm for usage with Realm Sync. The local Realm will be copied to the directory specified by the `RLMRealmConfiguration`.

You can also use `writeCopyForConfiguration` to copy a synced Realm to another destination and use it with another sync user.

Usage:
```swift
var localConfig = Realm.Configuration()
let localRealm = try! Realm(configuration: localConfig)
try! localRealm.write {
    localRealm.add(SwiftPerson(firstName: "John", lastName: "Doe"))
}

var syncConfig = myUser.configuration(partitionValue: .null)
try! localRealm.writeCopy(configuration: syncConfig)

let syncedRealm = try! Realm(configuration: syncConfig)
// syncedRealm will contain `SwiftPerson(firstName: "John", lastName: "Doe")`
```

This PR also adds the ability to set a `seedFilePath` parameter on the RLMRealm. This allows you to pre-populate an instance of a Realm with data if the file did not already exist.

```swift
var config = Realm.Configuration()
config.seedFilePath = // Location of the Realm that will initially populate this instance.
```


tgoyne:
Isn't this `writeCopy()` but taking a configuration instead of a file URL and encryption key?

leemaguire:
> Isn't this `writeCopy()` but taking a configuration instead of a file URL and encryption key?

Pretty much. I didn't go the route of extending `writeCopy` due to the feature being called `export` in other SDK's.

tgoyne:
It seems very weird to me to have two totally different names for the same operation. I'd expect to just have `writeCopy(forConfiguration:)` and have it do the right thing for that configuration (which is what `Realm::export_to()` does internally already).

Realm/ObjectServerTests/RLMObjectServerTests.mm
```diff
@@ -1979,6 +1979,133 @@ - (void)testSyncConfigShouldNotMigrate {
...
+
+    RLMRealm *localRealm = [RLMRealm realmWithConfiguration:localConfig error:nil];
+    [localRealm transactionWithBlock:^{
+        [localRealm addObject:[Person ringo]];
+    }];
+
+    [localRealm writeCopyWithConfiguration:syncConfig error:nil];
+
+    RLMRealm *syncedRealm = [RLMRealm realmWithConfiguration:syncConfig error:nil];
+    XCTAssertEqual([[Person allObjectsInRealm:syncedRealm] objectsWhere:@"firstName = 'Ringo'"].count, 1);
```
ericjordanmossman, line 2002: 1U

Realm/ObjectServerTests/RLMObjectServerTests.mm
```diff
@@ -1979,6 +1979,133 @@ - (void)testSyncConfigShouldNotMigrate {
...
+    RLMUser *user = [self userForTest:_cmd];
+    RLMRealmConfiguration *syncConfig = [user configurationWithPartitionValue:NSStringFromSelector(_cmd)];
+    syncConfig.objectClasses = @[Person.class];
+
+    RLMUser *user2 = [self logInUserForCredentials:[self basicCredentialsWithName:@"SyncToSyncUser"
+                                                                         register:YES]];
+    RLMRealmConfiguration *syncConfig2 = [user2 configurationWithPartitionValue:NSStringFromSelector(_cmd)];
+
+    RLMRealm *syncedRealm = [RLMRealm realmWithConfiguration:syncConfig error:nil];
+    [syncedRealm.syncSession suspend];
```
ericjordanmossman, line 2043: Clever - this is to make sure that changes aren't synced for testing?

Realm/ObjectServerTests/RLMObjectServerTests.mm
```diff
@@ -1979,6 +1979,133 @@ - (void)testSyncConfigShouldNotMigrate {
...
+    [self waitForDownloadsForRealm:syncedRealm];
+    XCTAssertEqual([syncedRealm allObjects:@"Person"].count, 3);
+    [syncedRealm transactionWithBlock:^{
+        [syncedRealm addObject:[Person stuart]];
+    }];
+
+    [self waitForUploadsForRealm:syncedRealm];
+    RLMResults<Person *> *syncedResults = [Person allObjectsInRealm:syncedRealm];
+
+    NSPredicate *p = [NSPredicate predicateWithFormat:@"firstName = 'John' AND lastName = 'Doe' AND _id = %@", conflictingObjectId];
```
ericjordanmossman, line 2104: Just wondering - should behavior be John Doe overwrite ringo, or should ringo stay over John Doe.
I don't know one way or another.

jsflax (approved):

leemaguire:
@tgoyne PTAL

Realm/RLMRealm.mm
```diff
@@ -447,6 +447,12 @@ + (instancetype)realmWithConfiguration:(RLMRealmConfiguration *)configuration
         }
     }
 
+    if (configuration.seedFilePath && ![RLMRealm fileExistsForConfiguration:configuration]) {
```
tgoyne, line 450: This is a race condition if multiple processes hit this. We need to do this inside a call to `DB::call_with_lock()` to ensure that only one process tries to copy the file, and that the second process doesn't open the file while the copy is in progress.

Realm/RLMRealm.mm
```diff
@@ -447,6 +447,12 @@ + (instancetype)realmWithConfiguration:(RLMRealmConfiguration *)configuration
...
+        [[NSFileManager defaultManager] copyItemAtURL:configuration.seedFilePath
```
tgoyne, line 451: We need to be checking the return value of this.

leemaguire: What should we do if it returns false?

leemaguire: I have added that it will throw if the return value is `NO`

Realm/RLMRealm.mm
```diff
@@ -448,9 +448,19 @@ + (instancetype)realmWithConfiguration:(RLMRealmConfiguration *)configuration
...
-                                                error:error];
+        bool didCopySeed = false;
+        bool didLock = DB::call_with_lock(configuration.seedFilePath.path.UTF8String,
+                                          [&configuration, &error, &didCopySeed](auto const&) {
+            didCopySeed = [[NSFileManager defaultManager] copyItemAtURL:configuration.seedFilePath
+                                                                  toURL:configuration.fileURL
+                                                                  error:error];
+        });
+        if (!didLock) {
+            @throw RLMException(@"Cannot copy seed Realm at path '%@' while it is already open.", configuration.seedFilePath);
```
tgoyne, line 459: This isn't an error. If call_with_lock() returns false, it just means that someone else created the file between when we checked if it existed and when we tried to get the lock. It's fine if that happens, and we should just open the file.

Realm/RLMRealm.mm
```diff
@@ -448,9 +448,19 @@ + (instancetype)realmWithConfiguration:(RLMRealmConfiguration *)configuration
     }
 
     if (configuration.seedFilePath && ![RLMRealm fileExistsForConfiguration:configuration]) {
-        [[NSFileManager defaultManager] copyItemAtURL:configuration.seedFilePath
-                                                toURL:[configuration fileURL]
-                                                error:error];
+        bool didCopySeed = false;
+        bool didLock = DB::call_with_lock(configuration.seedFilePath.path.UTF8String,
```
tgoyne, line 452: We need to lock the target path, not the source path.

Realm/RLMRealm.mm
```diff
@@ -448,9 +448,19 @@ + (instancetype)realmWithConfiguration:(RLMRealmConfiguration *)configuration
...
+                                          [&configuration, &error, &didCopySeed](auto const&) {
+            didCopySeed = [[NSFileManager defaultManager] copyItemAtURL:configuration.seedFilePath
+                                                                  toURL:configuration.fileURL
+                                                                  error:error];
+        });
+        if (!didLock) {
+            @throw RLMException(@"Cannot copy seed Realm at path '%@' while it is already open.", configuration.seedFilePath);
+        }
+        if (!didCopySeed) {
+            @throw RLMException(@"Failed to copy seed Realm at path '%@' to destination path '%@'.", configuration.seedFilePath, configuration.fileURL);
```
tgoyne, line 462: This is a runtime error and not a precondition, so it should set the `error` out parameter if that's non-nil and only throw an exception if it's nil (i.e. RLMSetErrorOrThrow). Either way the error message should include the error from `copyItemAtURL:toURL:` rather than discarding the underlying error.

Realm/RLMRealm.mm
```diff
@@ -447,6 +447,24 @@ + (instancetype)realmWithConfiguration:(RLMRealmConfiguration *)configuration
...
+                           [&configuration, &error](auto const&) {
+            if (![RLMRealm fileExistsForConfiguration:configuration]) {
+                @autoreleasepool {
+                    bool didCopySeed = false;
+                    NSError *copyError;
+                    didCopySeed = [[NSFileManager defaultManager] copyItemAtURL:configuration.seedFilePath
+                                                                          toURL:configuration.fileURL
+                                                                          error:&copyError];
+                    if (!didCopySeed) {
+                        RLMSetErrorOrThrow(copyError, error);
```
tgoyne, line 461: Need to return nil after this, or we'll go on to try to open the file after failing to copy if `error` is non-nil.

leemaguire:
@tgoyne PTAL

tgoyne (approved):

tgoyne:
Oh actually it isn't compiling at all.

Issue #7624, opened 2022-01-20
Xcode 13.1 "invalid exclude" warning


bbauman1:
### How frequently does the bug occur?

All the time

### Description

When compiling with Realm through SPM a warning is generated for an Invalid Exclude

### Stacktrace & log output

```shell
Invalid Exclude '/Users/redacted/Library/Developer/Xcode/DerivedData/redacted/SourcePackages/checkouts/realm-cocoa/tools': File not found.
```

### Can you reproduce the bug?

Yes, always

### Version

10.21.1

### What SDK flavour are you using?

Local Database only

### Are you using encryption?

No, not using encryption

### Platform OS and version(s)

Xcode 13.1. iOS any-version

### Build environment

Xcode version: 13.1
Dependency manager and version: SPM

bbauman1:
I was able to get a more descriptive output 

```Invalid Exclude '/Users/redacted/Library/Developer/Xcode/DerivedData/redacted/SourcePackages/checkouts/realm-core/src/realm/object-store/impl/set_notifier.hpp': File not found.```

leemaguire:
Hi @bbauman1 We will fix this, thanks!

drmarkpowell:
seeing this in Xcode 13.2.1 as well

zbeckman:
Same issue. Xcode 13.2.1, Realm 10.21.1 (and RealmDatabase 11.8.0).
> Invalid Exclude '/Users/zbeckman/Library/Developer/Xcode/DerivedData/NourishMe-gfsnyvfufdqacwdlkrekaaxqhgsm/SourcePackages/checkouts/realm-swift/tools': File not found.

aehlke:
Does latest Xcode work with 10.21.0 or an earlier version instead?

OlegEremenko991:
Same issue:
Invalid Exclude '/Users/oleg991/Library/Developer/Xcode/DerivedData/Days_2.1-cuqmlgwmfaqgzgdzjktdrpjgkshc/SourcePackages/checkouts/realm-cocoa/tools': File not found.

pavel-ship-it:
Fixed in #7646

PR #7646, opened 2022-02-01
[SPM] Fix invalid exclude of "tools" directory

rzulkoski:
#7624 - Fix warning introduced in Xcode 13.1 by removing invalid exclude of non-existent "tools" directory from Realm target in Package.swift.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -159,7 +159,6 @@ let package = Package(
                 "logo.png",
                 "plugin",
                 "scripts",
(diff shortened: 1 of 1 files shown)


rzulkoski:
Looks like the "tools" directory was removed in this commit: [8d8a2801](https://github.com/realm/realm-swift/commit/8d8a2801510b60a8ad0aa95d008a983348ff10a4)

PR #7661, opened 2022-02-09
Simplify cocoapods setup

### LINKED ISSUE #7660: 'No Xcode found with version' Error on Jenkins on pod install
Installing pod on Jenkins always results in 'No Xcode found with version' error even with all the variables set and Xcode installed.

tgoyne:
We no longer need the shared setup stuff like detecting the Xcode and Swift version for cocoapods setup, so extract cocoapods setup (and core downloading) from build.sh into its own script.

Fixes #7660.

### DIFF
--- a/scripts/download-core.sh
+++ b/scripts/download-core.sh
@@ -1,0 +1,92 @@
+#!/usr/bin/env bash
+
+set -euo pipefail
+source_root="$(dirname "$0")/.."
+readonly source_root
+
+# override this env variable if you need to download from a private mirror
+: "${REALM_BASE_URL:="https://static.realm.io/downloads"}"
+# set to "current" to always use the current build
+: "${REALM_CORE_VERSION:=$(sed -n 's/^REALM_CORE_VERSION=\(.*\)$/\1/p' "${source_root}/dependencies.list")}"
+
+readonly dst="$source_root/core"
+copy_core() {
+    local src="$1"
+    rm -rf "$dst"
+    mkdir "$dst"
+    ditto "$src" "$dst"
+
+    # XCFramework processing only copies the "realm" headers, so put the third-party ones in a known location
+    mkdir -p "$dst/include"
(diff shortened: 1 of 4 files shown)


scripts/download-core.sh
```diff
@@ -0,0 +1,92 @@
+#!/usr/bin/env bash
+
+set -euo pipefail
+source_root="$(dirname "$0")/.."
+readonly source_root
```
leemaguire, line 5:
```suggestion
readonly source_root="$(dirname "$0")/.."
```

scripts/download-core.sh
```diff
@@ -0,0 +1,92 @@
...
+    exit 0
+fi
+
+echo "Downloading dependency: ${version} from ${url}"
+
+if [ -z "$TMPDIR" ]; then
+    TMPDIR='/tmp'
+fi
+temp_dir=$(dirname "$TMPDIR/waste")/realm-core-tmp
+readonly temp_dir
```
leemaguire, line 61:
```suggestion
readonly temp_dir=$(dirname "$TMPDIR/waste")/realm-core-tmp
```

scripts/setup-cocoapods.sh
```diff
@@ -0,0 +1,17 @@
+#!/usr/bin/env bash
+
+set -euo pipefail
+source_root="$(dirname "$0")/.."
+readonly source_root
```
leemaguire, line 5: Same comment as above, does the `readonly` prefix need to be on a new line?

leemaguire (commented):
I have left some minor comments, LGTM otherwise.

scripts/setup-cocoapods.sh
tgoyne: https://github.com/koalaman/shellcheck/wiki/SC2155

Shellcheck complains about combining them because it results in ignoring failures in the subshell.

jsflax (approved):

PR #7662, opened 2022-02-10
Add .discardLocal client reset mode.

ericjordanmossman:
- Adds a new clientResetMode property to the SyncConfiguration.
- Adds two notification callback properties to SyncConfiguration. `.notifyBeforeClientReset`, `.notifyAfterClientReset`.
- Adds a handful of admin api testing methods for fully integrated client reset tests.

Unfortunately, as of right now, it takes any where between 3-5 minutes to run a full client reset test. I ran into the same issues [here](https://github.com/realm/realm-core/blob/master/test/object-store/sync/sync_test_utils.cpp#L459) and [here](https://github.com/realm/realm-core/blob/master/test/object-store/sync/sync_test_utils.cpp#L512). This is commented in line with the tests.


ericjordanmossman:
Needs Changelog. Should also edit the PR description to include sample usage.

Realm/RLMSyncConfiguration.h
```diff
@@ -18,13 +18,40 @@
...
+    RLMClientResetModeDiscardLocal
+};
+
+/**
+ A block type used to report before a client reset will occur.
+ The RlMRealm argument contains the local database state prior to client reset.
+ */
+typedef void(^RLMClientResetBeforeBlock)(RLMRealm * _Nonnull);
+
+// ???: Is there really no way to label these arguments?
```
ericjordanmossman, line 47: Is there really no way to label these arguments? It'd be nice if there were auto filled template in xcode.

Realm/RLMSyncConfiguration.h
```diff
@@ -18,13 +18,40 @@
...
 
+/**  Determines file behavior during a client reset.
+
+ - see: https://docs.mongodb.com/realm/sync/error-handling/client-resets/
+*/
+typedef NS_ENUM(NSUInteger, RLMClientResetMode) {
+    /// The SDK will create a back up of unsynced data.The client reset error handler may be manually overwritten to
+    /// transfer data from the backup copy to a new destination. Otherwise no effort to transfer the data from the backup is carried out.
+    RLMClientResetModeManual,
+    /// The SDK will overwrite the client database with the server database. Object accessors remain bound so Realm notifications are not disrupted.
```
tgoyne, line 37:
```suggestion
    /// All unsynchronized local changes are automatically discarded and the
    /// local state is automatically reverted to the most recent state from the
    /// server. Unsynchronized changes can then be recovered in the
    /// post-client-reset callback block.
```

Realm/RLMSyncConfiguration.mm
```diff
@@ -56,6 +58,27 @@ RLMSyncSystemErrorKind errorKindForSyncError(SyncError error) {
 }
 }
 
+struct BeforeClientResetWrapper {
+    RLMClientResetBeforeBlock block;
+    void operator()(std::shared_ptr<Realm> local) {
+        RLMSchema *schema = [RLMSchema dynamicSchemaFromObjectStoreSchema:local->schema()];
+        RLMRealm *realm = [RLMRealm realmWithSharedRealm:local schema:schema];
```
tgoyne, line 65: This doesn't work for Swift. The dynamic schema read from object store is missing information required to use RealmSwiftObject subclasses, and only works for the dynamic API we expose for migrations.

Realm/RLMSyncConfiguration.mm
```diff
@@ -219,7 +329,10 @@ - (instancetype)initWithUser:(RLMUser *)user
                 errorHandler(nsError, session);
             });
         };
-        _config->client_resync_mode = realm::ClientResyncMode::Manual;
+        // Default to manual mode
+        _config->client_resync_mode = (clientResetMode) ? translateClientResetMode(clientResetMode) : realm::ClientResyncMode::Manual;
```
tgoyne, line 333: This conditional doesn't make any sense. The result is identical without it.

RealmSwift/Sync.swift
```diff
@@ -228,6 +236,58 @@ public typealias Provider = RLMIdentityProvider
...
+     - note: Defaults to `.manual`
+
+     - see: `RLMClientResetMode`
+     - see: https://docs.mongodb.com/realm/sync/error-handling/client-resets/
+    */
+    public let clientResetMode: ClientResetMode
+
+    /**
+     A callback which notifies prior to  prior to a client reset occurring.
+     The Realm argument contains the local database state prior to client reset.
```
tgoyne, line 250: This should specifically say that the Realm passed in is frozen.

RealmSwift/Sync.swift
```diff
@@ -228,6 +236,58 @@ public typealias Provider = RLMIdentityProvider
...
+        set {
+            self.asConfig().beforeClientReset = ObjectiveCSupport.convert(object: newValue)
+            self._notifyBeforeClientReset = newValue
+        }
+    }
+    private var _notifyBeforeClientReset: ((Realm) -> Void)?
+    /**
+     A callback which notifies after a client reset has occurred.
+     The first realm argument contins the local database state prior to client reset.
+     The second argument contains the server database state prior to client reset.
```
tgoyne, line 271: The second argument is the local Realm after the client reset.

RealmSwift/Sync.swift
```diff
@@ -228,6 +236,58 @@ public typealias Provider = RLMIdentityProvider
...
+    */
+    public let clientResetMode: ClientResetMode
+
+    /**
+     A callback which notifies prior to  prior to a client reset occurring.
+     The Realm argument contains the local database state prior to client reset.
+     ```
+     configuration.syncConfiguration?.notifyBeforeClientReset = { localRealm in
+        let results = localRealm.objects(MyClass.self)
+        ...
```
tgoyne, line 254: `...` is carrying a lot of weight here. It's not clear what if anything anyone would actually write here.

RealmSwift/Sync.swift
```diff
@@ -228,6 +236,58 @@ public typealias Provider = RLMIdentityProvider
...
+     }
+     ```
+     */
+    public var notifyBeforeClientReset: ((Realm) -> Void)? {
+        get {
+            return self._notifyBeforeClientReset
+        }
+        set {
+            self.asConfig().beforeClientReset = ObjectiveCSupport.convert(object: newValue)
```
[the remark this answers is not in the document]
ericjordanmossman: If a developer wants to use these callbacks, should they do so only at user.configuration(...)?
I was imagining something like 
```
let config = user.configuration(partitionValue:clientResetMode)
config.SyncConfiguration!.notifyBeforeClientReset = { ... }
config.SyncConfiguration!.notifyAfterClientReset = { ... }
```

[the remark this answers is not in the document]
tgoyne: Yes, it should be set when creating the config.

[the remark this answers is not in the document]
ericjordanmossman: Okay perfect, that'll clear up a lot of problems.

Realm/RLMSyncConfiguration.mm
ericjordanmossman: What information is missing?

ericjordanmossman: I'm wondering what else I could do differently with `objectSchemaForObjectStoreSchema` that `dynamicSchemaFromObjectStoreSchema` isn't doing already.

ericjordanmossman: Or if we have any other methods for converting between object-store and objc.

RealmSwift/ObjectiveCSupport.swift
```diff
@@ -163,4 +163,45 @@ import Realm
             return object(Int(totalBytes), Int(usedBytes))
         }
     }
+
+    // ???: Should the swift blocks be a named type?
```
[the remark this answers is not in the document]
ericjordanmossman: My goal was labelling the arguments for Swift. But this is not possible, so using a typealias wouldn't add that value.

Realm/RLMSyncConfiguration.mm
tgoyne: It just isn't possible to construct a valid schema from the information object store has. You have to use the schema passed in from the user (possibly implicitly) in the original RLMRealmConfiguration.

ericjordanmossman: I tried passing the RLMSchema from the RLMConfiguration to the  RLMSyncConfigration at the time it's initialized. One issue is that RLMConfiguration and it's RLMSchema isn't created until _after_ the RLMSyncConfiguration is created. So it's not available at the time of initialization.

I was also thinking there might be a way to get the RLMSchema at the time of client reset. A problem with this is that there isn't a way to retrieve the RLMRealmConfiguration (and it's customSchema) that the RLMSyncConfiguration is attached to.

What other ways could we get an accurate RLMSchema to open these realms?

RealmSwift/Sync.swift
```diff
@@ -389,6 +452,34 @@ public extension User {
...
+     - parameter partitionValue: The `BSON` value the Realm is partitioned on.
+     - parameter clientResetMode: Determines file recovery behavior during a client reset.
+     - parameter notifyBeforeClientReset: A callback which notifies prior to a client reset occurring. See: `notifyBeforeClientReset`.
+     - parameter notifyAfterClientReset: A callback which notifies after a client reset has occurred. See: `notifyAfterClientReset`.
+
+
+     Additional settings can be optionally specified. Descriptions of these
+     settings follow.
+
+     `enableSSLValidation` is true by default. It can be disabled for debugging
```
leemaguire, line 467: does `enableSSLValidation` exist anymore?

ericjordanmossman: I don't think it does? It was previously on the sync manager, then got moved to syncConfiguration. But I no longer see it all. The docstring must have been left over?

Realm/RLMRealm_Private.hpp
```diff
@@ -21,6 +21,7 @@
 #import "RLMClassInfo.hpp"
 
 #import <realm/object-store/object_schema.hpp>
+#import <realm/object-store/shared_realm.hpp>
```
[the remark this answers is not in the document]
ericjordanmossman: Can you forward declare a cpp type? I can't seem to @class SharedRealm because it conflicts with the typedef. And if I try to declare it as typedef in the namespace, compiler says all cpp types require type specifier.

Realm/ObjectServerTests/SwiftObjectServerTests.swift
```diff
@@ -375,6 +463,152 @@ class SwiftObjectServerTests: SwiftSyncTestCase {
...
+                            XCTAssertEqual(documentCount, 1)
+                            XCTAssertEqual(documents[0]["firstName"]??.stringValue, "Paul")
+                            break
+                        }
+                    case .failure:
+                        XCTFail("Should find")
+                    }
+                }
+                if documentCount == 0 {
+                    if requestCount * timeBetweenRequests > 300 {
```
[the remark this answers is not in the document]
ericjordanmossman: Yes. This can take anywhere between 30 seconds and 4 minutes. Similar to what @ironage [experienced](https://github.com/realm/realm-core/blob/master/test/object-store/sync/sync_test_utils.cpp#L464).

I was previously using SwiftSyncHugeObject, which would take well beyond five minutes in some cases. So I switched to something smaller.

Realm/ObjectServerTests/SwiftObjectServerTests.swift
```diff
@@ -313,6 +333,74 @@ class SwiftObjectServerTests: SwiftSyncTestCase {
...
+                XCTFail("Error: \(error.localizedDescription)")
+            }
+        }
+        waitForExpectations(timeout: 6, handler: nil)
+        XCTAssertTrue(try RealmServer.shared.syncEnabled(appServerId: appServerId, syncServiceId: syncServiceId))
+    }
+
+    func waitForDevModeEnabled(appServerId: String, syncServiceId: String, syncServiceConfig: [String: Any]) {
+        let devModeEnabled = try! RealmServer.shared.devModeEnabled(appServerId: appServerId, syncServiceId: syncServiceId)
+        let exp3 = expectation(description: "enable dev mode")
```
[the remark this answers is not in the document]
ericjordanmossman: Good catch.

Realm/ObjectServerTests/RealmServer.swift
```diff
@@ -922,6 +930,97 @@ public class RealmServer: NSObject {
         return appId
     }
 
+    public func retrieveSyncServiceId(appServerId: String) throws -> String {
+        guard let session = session else {
+            throw URLError(.unknown)
```
tgoyne, line 935: These should be `fatalError()`, as the session will only be `nil` due to writing invalid code calling this (specifically, trying to call it in a child process).

Realm/ObjectServerTests/RealmServer.swift
```diff
@@ -922,6 +930,97 @@ public class RealmServer: NSObject {
...
+
+    public func syncEnabled(appServerId: String, syncServiceId: String) throws -> Bool {
+        guard let session = session else { throw URLError(.unknown) }
+        let app = session.apps[appServerId]
+        let response = try app.services[syncServiceId].config.get().get() as? [String: Any]
+        guard let syncInfo = response?["sync"] as? [String: Any] else {
+            return false
+        }
+
+        if syncInfo["state"] as? String == "enabled" {
```
tgoyne, line 973: The if/else here is redundant.

Realm/ObjectServerTests/SwiftObjectServerTests.swift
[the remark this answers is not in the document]
tgoyne: I've seen the translator lag 10+ minutes behind. It's significantly slower than the sync server, and our sync tests are heavily bounded in speed by the sync server, so the translator ends up getting progressively further and further behind.

Disabling sync, deleting all docs via the mongo API, then re-enabling sync at the start of the test might be a net performance gain since it (might?) cancel all of the pending work from previous tests. If that doesn't work out we may need to launch a separate server process just for the client reset tests.

Realm/RLMRealm_Private.hpp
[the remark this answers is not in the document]
tgoyne: It should use `std::shared_ptr<Realm>` instead of `SharedRealm`.

Realm/RLMSyncConfiguration.mm
tgoyne: The objectstore API here mostly just doesn't really work; we can't actually accept a `std::shared_ptr<Realm>` and construct a `RLMRealm` because that bypasses the `RLMRealm` cache. We also don't determine what the `RLMSchema` is for a `RLMRealm` until *after*  we call `Realm::get_shared_realm()` (and this is an important optimization for getting `RLMRealm` instances when the Realm is already open on a different thread).

Realm/ObjectServerTests/SwiftObjectServerTests.swift
```diff
@@ -375,6 +463,152 @@ class SwiftObjectServerTests: SwiftSyncTestCase {
...
+                let results = local.objects(SwiftPerson.self)
+                XCTAssertEqual(results.count, 2)
+                let paul = results.filter("firstName == 'Paul'")
+                let john = results.filter("firstName == 'John'")
+                XCTAssertNotNil(paul)
+                XCTAssertNotNil(john)
+            }
+            let afterClientResetBlock: (Realm, Realm) -> Void = { before, after in
+                // Expect the local realm that was overwritten to have had 2 objects before it was overwritten.
+                let results = before.objects(SwiftPerson.self)
```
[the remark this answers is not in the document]
ericjordanmossman: The github ui doesn't say outdated because this line stayed the same. See latest commit.

ericjordanmossman:
@tgoyne 
I re-requested review because I think addressed all but one point. Note that I'm still thinking about the objectStore schema to swift Realm issue.

Realm/RLMSyncConfiguration.mm
ericjordanmossman: What would prevent us from calling `RLMGetAnyCachedRealmForPath` within the wrapper, then retrieving the schema from that? Is it roundabout? Lead to a migration problem?

ericjordanmossman: I suppose the path doesn't exist yet, either?

ericjordanmossman: What would java be doing differently than us?
It appears they take an [OS schema and OS realm](https://github.com/realm/realm-java/blob/c612779423630959ff867a37d0e76d3ede8fe536/realm/realm-library/src/objectServer/java/io/realm/internal/SyncObjectServerFacade.java#L125).
But I don't totally get [how they do this](https://github.com/realm/realm-java/blob/c612779423630959ff867a37d0e76d3ede8fe536/realm/realm-library/src/main/java/io/realm/Realm.java#L181).

tgoyne: They either don't have all the schema data that's not part of the objectstore Schema which we do, or they have some way of looking it up. They also presumably do Realm instance caching differently.

ericjordanmossman:
```
struct BeforeClientResetWrapper {
    RLMClientResetBeforeBlock block;
    void operator()(std::shared_ptr<Realm> local) {
        RLMSchema *schema = [RLMSchema dynamicSchemaFromObjectStoreSchema:local->schema()];
        RLMRealm *realm = [RLMRealm realmWithSharedRealm:local schema:schema];
        // This is the only way I've been able to get path or config info inside this wrapper
        // Everything else is a catch-22. I need to get RLMConfig or schema info into this wrapper. The RLMConfig needs to be created and opened for this be generated. But in order to create and open the RLMConfig,
        // the RLMSyncConfiguration needs to be created before. But if the RLMSyncConfiguration is created beforehand, then it won't have the schema info.
        // I haven't found a way really retroactively pass the schema info back, because this information is needed at initialization.
        std::string str = std::string([realm.configuration.fileURL.absoluteString UTF8String]);
        RLMRealm *cached = RLMGetAnyCachedRealmForPath(str);
        if (cached) { // can this ever be false? There's have to be a cached realm because one was just opened?
            block(cached);
        } else {
            block(realm);
        }
    }
};
```
Is checking for a cached realm a viable option? Seems really wonky.

ericjordanmossman: @tgoyne 
When using a `realm::SyncConfig` to set [here](https://github.com/realm/realm-swift/commit/f69f6e30088d1622343e167c91bb890ec99c4970#diff-36067fa7e0f5b4726384d8cdd6aa9392cf67bfc8bb94904f10bd434d9627e2c4R198), I'm still getting nil [here](https://github.com/realm/realm-swift/commit/f69f6e30088d1622343e167c91bb890ec99c4970#diff-36067fa7e0f5b4726384d8cdd6aa9392cf67bfc8bb94904f10bd434d9627e2c4R72). I thought that no longer using `initWithRawConfiguration` would resolve the issue. What else could be missing? Is something being lost [here](https://github.com/realm/realm-swift/blob/em/discard-local/Realm/RLMRealmConfiguration%2BSync.mm#L52)?

Realm/ObjectServerTests/SwiftObjectServerTests.swift
```diff
@@ -375,6 +460,160 @@ class SwiftObjectServerTests: SwiftSyncTestCase {
...
+                                                   notifyAfterClientReset: afterClientResetBlock)
+            configuration.objectTypes = [SwiftPerson.self]
+
+            XCTAssertEqual(configuration.syncConfiguration?.clientResetMode, .discardLocal)
+            XCTAssertNotNil(configuration.syncConfiguration?.notifyBeforeClientReset)
+            XCTAssertNotNil(configuration.syncConfiguration?.notifyAfterClientReset)
+
+            // Seed 1 object, Upload to server
+            try autoreleasepool {
+                let realm = try Realm(configuration: configuration) // 1.  2. okay *now* opening realm for first time, now the config is set to wrapper
```
ericjordanmossman, line 513: remove comment

ericjordanmossman:
@tgoyne @jsflax @leemaguire Ready for review

Realm/ObjectServerTests/SwiftObjectServerTests.swift
```diff
@@ -33,6 +33,23 @@ import RealmTestSupport
...
+        let database = mongoClient.database(named: "test_data")
+        let collection = database.collection(withName: collectionName)
+        removeAllFromCollection(collection)
+        return collection
+    }
+
+    func removeAllFromCollection(_ collection: MongoCollection) {
+        let ex = expectation(description: "delete objects")
+        collection.deleteManyDocuments(filter: [:]) { result in
+            if case .success = result {
```
tgoyne, line 47: This should log something if it gets an error.

Realm/ObjectServerTests/SwiftObjectServerTests.swift
```diff
@@ -375,6 +460,160 @@ class SwiftObjectServerTests: SwiftSyncTestCase {
...
+                let john2 = results2.filter("firstName == 'John'")
+                XCTAssertFalse(paul2.isEmpty)
+                XCTAssert(john2.isEmpty)
+                afterCalledCount += 1
+            }
+
+            var configuration = user.configuration(partitionValue: #function,
+                                                   clientResetMode: .discardLocal,
+                                                   notifyBeforeClientReset: beforeClientResetBlock,
+                                                   notifyAfterClientReset: afterClientResetBlock)
```
tgoyne, line 504: Should this maybe be `clientResetMode: .discardLocal(beforeReset: beforeClientResetBlock, afterReset: afterClientResetBlock)`? It would eliminate one of the ways to call `configuration()` incorrectly.

Realm/RLMSyncConfiguration.mm
```diff
@@ -56,6 +61,50 @@ RLMSyncSystemErrorKind errorKindForSyncError(SyncError error) {
 }
 }
 
+struct BeforeClientResetWrapper {
+    RLMClientResetBeforeBlock block;
+    RLMRealmConfiguration *rlmConfig;
```
tgoyne, line 66: This doesn't appear to need the entire RLMRealmConfiguration? Storing just the path and custom schema would fix the retain cycle and reduce memory usage.

Realm/RLMSyncConfiguration.mm
```diff
@@ -56,6 +61,50 @@ RLMSyncSystemErrorKind errorKindForSyncError(SyncError error) {
...
+    void operator()(std::shared_ptr<Realm> local) {
+        RLMSchema *schema;
+        if (rlmConfig.dynamic) {
+            schema = [RLMSchema dynamicSchemaFromObjectStoreSchema:local->schema()];
+        }
+        if (auto cached = RLMGetAnyCachedRealmForPath(rlmConfig.config.path)) { // crash, no rlmConfig, despite being set when realm is open?
```
tgoyne, line 72: I assume this comment is out of date.

This is missing an `else` and will always overwrite the dynamic schema.

Realm/RLMSyncConfiguration.mm
```diff
@@ -56,6 +61,50 @@ RLMSyncSystemErrorKind errorKindForSyncError(SyncError error) {
 }
 }
 
+struct BeforeClientResetWrapper {
+    RLMClientResetBeforeBlock block;
+    RLMRealmConfiguration *rlmConfig;
+    void operator()(std::shared_ptr<Realm> local) {
```
tgoyne, line 67: This is a point where we're making a C++ -> obj-c call so the body of this function needs to be wrapped in an `@autoreleasepool` or the autoreleased objects created in this function may live a very long time.

RealmSwift/Sync.swift
```diff
@@ -389,6 +450,31 @@ public extension User {
...
+     - parameter partitionValue: The `BSON` value the Realm is partitioned on.
+     - parameter clientResetMode: Determines file recovery behavior during a client reset.
+     - parameter notifyBeforeClientReset: A callback which notifies prior to a client reset occurring. See: `notifyBeforeClientReset`.
+     - parameter notifyAfterClientReset: A callback which notifies after a client reset has occurred. See: `notifyAfterClientReset`.
+
+
+     Additional settings can be optionally specified. Descriptions of these
+     settings follow.
+
+     - warning: NEVER disable SSL validation for a system running in production.
```
tgoyne, line 465: None of this is applicable here.

Realm/RLMSyncConfiguration_Private.hpp
```diff
@@ -36,6 +36,13 @@ NS_ASSUME_NONNULL_BEGIN
 
 - (realm::SyncConfig&)rawConfiguration;
 
+// Pass the RLMRealmConfiguration to it's sync configuration so client reset callbacks
+// can access schema, dynamic, and path properties.
+
```
ericjordanmossman, line 41: extra newline

RealmSwift/Sync.swift
```diff
@@ -207,6 +207,55 @@ public typealias Provider = RLMIdentityProvider
...
+ - see: `RLMClientResetMode`
+ - see: https://docs.mongodb.com/realm/sync/error-handling/client-resets/
+*/
+public enum ClientResetMode {
+    /// - see: `RLMClientResetModeManual`
+    case manual
+    /// - see: `RLMClientResetModeDiscardLocal` for more details on `.discardLocal` behavior
+    ///
+    /// The first function `.discardLocal` argument notifies prior to a client reset occurring.
+    /// This functions `Realm` argument contains a frozen copy of the Realm state prior to client reset.
```
ericjordanmossman, line 222: Weird wording. Will change.

RealmSwift/Sync.swift
```diff
@@ -228,6 +277,15 @@ public typealias Provider = RLMIdentityProvider
      */
     internal let stopPolicy: RLMSyncStopPolicy
 
+    /**
+     An enum which determines file recovery behvaior in the event of a client reset.
+     - note: Defaults to `.manual`
+
+     - see: `RLMClientResetMode`
```
ericjordanmossman, line 284: Will refer to swift defintion too in future commit.

ericjordanmossman:
@tgoyne Ready for review again, thanks!

Realm/RLMSyncConfiguration.mm
ericjordanmossman: Changed [here](https://github.com/realm/realm-swift/pull/7662/commits/1acd21cb3a162ec702f9fd49b697f2cea50ef2d8). The github UI didn't update.

tgoyne (approved):

PR #7673, opened 2022-02-23
Add Xcode 13.3 to CI

tgoyne:
And also fix the always-failing tests and make the sometimes-failing tests fail in better ways (i.e. don't make the job hit 45 minutes and time out).


Realm/Tests/KVOTests.mm
```diff
@@ -1532,6 +1528,7 @@ - (void)testSetKVO {
 
     id mutator = [obj mutableSetValueForKey:@"set"];
     id mutator2 = [obj2 mutableSetValueForKey:@"set"];
+    id set2 = [obj2 valueForKey:@"set"];
```
leemaguire, line 1531: What is this test doing differently with just `valueForKey`?

tgoyne: `unionSet:` calls `copy` on the thing passed into it. With `NSSet` and `RLMSet`, this is a shallow copy. However, when running in Catalyst (and presumably everywhere with a recently deployment target), in the unmanaged case `mutator` is a `NSKeyValueSlowMutableSet`, which performs a deep copy and requires the objects in the set to conform to `NSCopying`. I started adding conformances to this to our types, but `-[RLMObject copy]` seems questionable (would people ever expect it to actually copy the object inside the Realm?), so I then instead shifted to avoiding deep copies.

leemaguire (approved):

PR #7677, opened 2022-02-24
Don't set a TEST_HOST for the UI tests

tgoyne:
c989dbb92 refactored the xcconfig files in a way that resulted in it being set
for all targets, but UI tests require it to be unset.


dianaafanador3 (approved):
Other than xcsheme files seems to be missing for `SwiftUISyncTests` and `SwiftUISyncTest` so we can run this test from Realm.xcodeproj everything looks perfect. We still have a failing test in SwiftUITests but It looks like it is not related to this changes.

PR #7690, opened 2022-03-04
Add a TMPDIR fallback for download-core.sh

### LINKED ISSUE #7688: download-core.sh script doesn't work on non interactive shells

All the time

### Description

Your download-core.sh script has the following lines

```bash
set -euo pipefail
```
which indicates that this script should fail if there's an undefined variable. However, below code causes a failure on non-interactive shells.

https://github.com/realm/realm-swift/blob/c6d8ff61221f2ef8a4cdc53a36f44e700c61b286/scripts/download-core.sh#L57-L59
On macOS, this TMDIR is sandboxed so it is created per user.
Even though your script covers the scenario of undefined `TMPDIR` variable, because of the first line, this script fails. There are two ways to fix this issue.

1. You can remove the `u` part from t`set -euo pipefail`. 
2. You can put https://github.com/realm/realm-swift/blob/c6d8ff61221f2ef8a4cdc53a36f44e700c61b286/scripts/download-core.sh#L57-L59 at the beginning of the script before the 
```bash
set -euo pipefail
```

### Stacktrace & log output

_No response_

Yes, always

### Reproduction Steps

Run the pod install step with non interactive shells or undefined the TMPDIR

### Version

10.23.0

Local Database only

### Are you using encryption?

No, not using encryption

Monterey 

### Build environment

Xcode version: ...
Dependency manager and version: ...

tgoyne:
I missed this variable when splitting download-core.sh out of build.sh in bf5c04619.

Fixes #7688.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,8 @@ x.y.z Release notes (yyyy-MM-dd)
 * The Swift package set the linker flags on the wrong target, resulting in
   linker errors when SPM decides to build the core library as a dynamic library
   ([#7266](https://github.com/realm/realm-swift/issues/7266)).
+* The download-core task failed if run in an environment without TMPDIR set
+  ([#7688](https://github.com/realm/realm-swift/issues/7688), since v10.23.0).
 
 <!-- ### Breaking Changes - ONLY INCLUDE FOR NEW MAJOR version -->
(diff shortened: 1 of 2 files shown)


pavel-ship-it (approved):
LGTM

PR #7720, opened 2022-03-25
Allow async await to be used in all supported OS's

leemaguire:
Allows apps targeting iOS 13 to use async await with Realm. Note Xcode 13.3 RC must be used. Fixes https://github.com/realm/realm-swift/issues/7599


ericjordanmossman:
Is Xcode cloud is right - It looks like searchable is only in iOS15. Or am I missing something?
https://developer.apple.com/documentation/swiftui/emptyview/searchable(text:placement:)

tgoyne:
The changelog also needs to explicitly say that async/await is no longer supported when building with Xcode 13.2.

leemaguire:
@tgoyne I have added a breaking changes section to the changelog

tgoyne (approved):

PR #7724, opened 2022-03-28
Remove object from list

ericjordanmossman:
Adds a list remove method which takes an object instead of an index.


RealmSwift/List.swift
```diff
@@ -154,6 +154,33 @@ public final class List<Element: RealmCollectionValue>: RLMSwiftCollectionBase,
...
+    /**
+     Removes an object in the list if present, otherwise throws an error. The object is not removed from the Realm that manages it.
+
+     - warning: This method may only be called during a write transaction.
+
+     - parameter object: The object to remove.
+     */
+    public func remove(_ object: Element) {
+        guard let index = self.index(of: object) else {
+            throwRealmException("Object does not exist in List") // ???: Is this too harsh? Should method just return a bool or something?
```
ericjordanmossman, line 166: Is an exception much? Should nothing happen if the object isn't present?

Naveen-C-Ramachandrappa (approved):

PR #7727, opened 2022-03-30
`@AsyncOpen`/`@AutoOpen` property wrappers with flexible sync.

dianaafanador3:
Allow to open a flexible synced realm with @AsyncOpen and @AutopOpen in SwiftIUI. 
Previously, using both property wrappers was very coupled to provide a partition value.

### DIFF
--- a/Realm/ObjectServerTests/RLMUser+ObjectServerTests.mm
+++ b/Realm/ObjectServerTests/RLMUser+ObjectServerTests.mm
@@ -58,6 +58,45 @@ using namespace realm;
     return dispatch_semaphore_wait(sema, dispatch_time(DISPATCH_TIME_NOW, (int64_t)(timeout * NSEC_PER_SEC))) == 0;
 }
 
+- (BOOL)waitForUploadsForRealm:(RLMRealm *)realm error:(NSError **)error {
+    const NSTimeInterval timeout = 20;
+    dispatch_semaphore_t sema = dispatch_semaphore_create(0);
+    RLMSyncSession *session = realm.syncSession;
+    NSAssert(session, @"Cannot call with invalid Realm");
+    __block NSError *completionError;
+    BOOL couldWait = [session waitForUploadCompletionOnQueue:dispatch_get_global_queue(DISPATCH_QUEUE_PRIORITY_LOW, 0) callback:^(NSError *error) {
+        completionError = error;
+        dispatch_semaphore_signal(sema);
+    }];
+    if (!couldWait) {
+        return NO;
+    }
+
+    if (error)
+        *error = completionError;
+
+    return dispatch_semaphore_wait(sema, dispatch_time(DISPATCH_TIME_NOW, (int64_t)(timeout * NSEC_PER_SEC))) == 0;
+}
+
+- (BOOL)waitForDownloadsForRealm:(RLMRealm *)realm error:(NSError **)error {
+    const NSTimeInterval timeout = 20;
+    dispatch_semaphore_t sema = dispatch_semaphore_create(0);
+    RLMSyncSession *session = realm.syncSession;
+    NSAssert(session, @"Cannot call with invalid Realm");
+    __block NSError *completionError;
+    BOOL couldWait = [session waitForDownloadCompletionOnQueue:dispatch_get_global_queue(DISPATCH_QUEUE_PRIORITY_LOW, 0) callback:^(NSError *error) {
+        completionError = error;
+    }];
+    if (!couldWait) {
+        return NO;
+    }
+
+    if (error)
+        *error = completionError;
+
+    return dispatch_semaphore_wait(sema, dispatch_time(DISPATCH_TIME_NOW, (int64_t)(timeout * NSEC_PER_SEC))) == 0;
+}
+
 - (void)simulateClientResetErrorForSession:(NSString *)partitionValue {
     RLMSyncSession *session = [self sessionForPartitionValue:partitionValue];
     NSAssert(session, @"Cannot call with invalid URL");
--- a/Realm/ObjectServerTests/RLMUser+ObjectServerTests.mm
+++ b/Realm/ObjectServerTests/RLMUser+ObjectServerTests.mm
@@ -58,41 +58,32 @@ using namespace realm;
     return dispatch_semaphore_wait(sema, dispatch_time(DISPATCH_TIME_NOW, (int64_t)(timeout * NSEC_PER_SEC))) == 0;
 }
 
-- (BOOL)waitForUploadsForRealm:(RLMRealm *)realm error:(NSError **)error {
+- (BOOL)waitForUploadsForRealm:(RLMRealm *)realm {
     const NSTimeInterval timeout = 20;
     dispatch_semaphore_t sema = dispatch_semaphore_create(0);
     RLMSyncSession *session = realm.syncSession;
     NSAssert(session, @"Cannot call with invalid Realm");
-    __block NSError *completionError;
-    BOOL couldWait = [session waitForUploadCompletionOnQueue:dispatch_get_global_queue(DISPATCH_QUEUE_PRIORITY_LOW, 0) callback:^(NSError *error) {
-        completionError = error;
+    BOOL couldWait = [session waitForUploadCompletionOnQueue:dispatch_get_global_queue(DISPATCH_QUEUE_PRIORITY_LOW, 0) callback:^(NSError *) {
         dispatch_semaphore_signal(sema);
     }];
     if (!couldWait) {
         return NO;
     }
 
-    if (error)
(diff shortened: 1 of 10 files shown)


RealmSwift/BSON.swift
```diff
@@ -200,6 +200,49 @@ extension MinKey: BSON {
         }
     }
 
+    /// Initialize a `BSON` from a type which conforms to `PartitionValue`. If this is not a valid `BSON` type,
+    /// if will be considered `BSON` null type and will return `nil`.
+    public init(_ bson: PartitionValue) {
```
leemaguire, line 205:
```suggestion
    static func convert<T>(_ bson: T) -> AnyBSON {
        switch bson {
        case let val as Int:
            return .int64(Int64(val))
        case let val as Int32:
            return .int32(val)
        case let val as Int64:
            return .int64(val)
        case let val as Double:
            return .double(val)
        case let val as String:
            return .string(val)
        case let val as Data:
            return .binary(val)
        case let val as Date:
            return .datetime(val)
        case let val as Decimal128:
            return .decimal128(val)
        case let val as UUID:
            return .uuid(val)
        case let val as ObjectId:
            return .objectId(val)
        case let val as Document:
            return .document(val)
        case let val as Array<AnyBSON?>:
            return .array(val)
        case let val as Bool:
            return .bool(val)
        case is MaxKey:
            return .maxKey
        case is MinKey:
            return .minKey
        case let val as NSRegularExpression:
            return .regex(val)
        case let val as AnyBSON:
            return val
        default:
            return .null
        }
    }

    /// Initialize a `BSON` from a type `T`. If this is not a valid `BSON` type,
    /// if will be considered `BSON` null type and will return `nil`.
    public init<T: BSON>(_ bson: T) {
        self = Self.convert(bson)
    }

    /// Initialize a `BSON` from a type which conforms to `PartitionValue`. If this is not a valid `BSON` type,
    /// if will be considered `BSON` null type and will return `nil`.
    public init(_ bson: PartitionValue) {
        self = Self.convert(bson)
    }
```

RealmSwift/SwiftUI.swift
```diff
@@ -1374,17 +1398,34 @@ private class ObservableAsyncOpenStorage: ObservableObject {
...
+                 become fully established., if empty or `nil` no connection timeout is set.
+     */
     public init(appId: String? = nil,
-                partitionValue: Partition,
                 configuration: Realm.Configuration? = nil,
                 timeout: UInt? = nil) {
         let app = ObservableAsyncOpenStorage.configureApp(appId: appId, withTimeout: timeout)
         // Store property wrapper values on the storage
-        storage = ObservableAsyncOpenStorage(asyncOpenKind: .autoOpen, app: app, configuration: configuration, partitionValue: AnyBSON(partitionValue))
+        storage = ObservableAsyncOpenStorage(asyncOpenKind: .asyncOpen, app: app, configuration: configuration, partitionValue: nil)
```
leemaguire, line 1424: Why is this `.asyncOpen` and not `.autoOpen`?

Realm/ObjectServerTests/RLMUser+ObjectServerTests.mm
```diff
@@ -58,6 +58,36 @@ - (BOOL)waitForDownloadToFinish:(NSString *)partitionValue {
     return dispatch_semaphore_wait(sema, dispatch_time(DISPATCH_TIME_NOW, (int64_t)(timeout * NSEC_PER_SEC))) == 0;
 }
 
+- (BOOL)waitForUploadsForRealm:(RLMRealm *)realm {
```
tgoyne, line 61: These are almost but not quite identical to the version in RLMSyncTestCase. I think the version there is functionally better (if the wait times out, this version silently continues without reporting an error), and the two should be unified rather than having the functionality in two places.

Realm/Tests/SwiftUISyncTestHostUITests/SwiftUISyncTestHostUITests.swift
```diff
@@ -548,3 +557,134 @@ extension SwiftUISyncTestHostUITests {
...
+        try? FileManager.default.createDirectory(at: clientDataRoot, withIntermediateDirectories: true)
+        try? FileManager.default.createDirectory(at: appClientDataRoot, withIntermediateDirectories: true)
+
+        // Create App once for this Test Suite
+        if SwiftUISyncTestHostUITests.flexibleSyncAppId == nil {
+            do {
+                let appId = try RealmServer.shared.createAppWithQueryableFields(["age", "firstName"])
+                SwiftUISyncTestHostUITests.flexibleSyncAppId = appId
+            } catch {
+                XCTFail("Cannot initialise test without a creating an App on the server")
```
tgoyne, line 574: This error doesn't make much sense to me? It's also discarding the actual error message.

Realm/Tests/SwiftUISyncTestHostUITests/SwiftUISyncTestHostUITests.swift
```diff
@@ -548,3 +557,134 @@ extension SwiftUISyncTestHostUITests {
...
+            }
+        }
+
+        // Instantiate App from appId after
+        do {
+            flexibleSyncApp = try getApp()
+        } catch {
+            print("Error creating user \(error)")
```
tgoyne, line 582: This should be XCTFail rather than print.

Realm/Tests/SwiftUISyncTestHostUITests/SwiftUISyncTestHostUITests.swift
```diff
@@ -222,18 +180,69 @@ extension SwiftUISyncTestHostUITests {
...
+        // Create app in current process
+        return App(id: appId!, configuration: appConfiguration, rootDirectory: clientDataRoot)
+    }
+
+    private func createUsers(email: String, password: String, n: Int) throws -> User {
+        let user = try registerAndLoginUser(email: email, password: password, for: app!)
+        let config = user.configuration(partitionValue: user.id)
+        let realm = try openRealm(configuration: config, for: user)
+        try realm.write {
+            (1...n).forEach { _ in
```
tgoyne, line 237: This could just be a `for`.

Realm/Tests/SwiftUISyncTestHostUITests/SwiftUISyncTestHostUITests.swift
```diff
@@ -222,18 +180,69 @@ extension SwiftUISyncTestHostUITests {
...
+        do {
+            app = try getApp()
+        } catch {
+            print("Error creating user \(error)")
+        }
+    }
+
+    private func getApp() throws -> App {
+        // Setup App for Testing
+        let appConfiguration = RLMAppConfiguration(baseURL: "http://localhost:9090",
```
tgoyne, line 223: Swift code should use the non-RLM-prefixed name.

There's a lot of incidentally duplicated code in each of the subclasses which would be easy to extract.

RealmSwift/BSON.swift
```diff
@@ -157,49 +157,59 @@ extension MinKey: BSON {
...
     }
 
+    /// Initialize a `BSON` from a type `T`. If this is not a valid `BSON` type,
+    /// it will be considered `BSON` null type and will return `nil`.
+    public init<T: BSON>(_ bson: T) {
+        self = Self.convert(bson)
+    }
+    /// Initialize a `BSON` from a type which conforms to `PartitionValue`. If this is not a valid `BSON` type,
+    /// it will be considered `BSON` null type and will return `nil`.
+    public init(_ bson: PartitionValue) {
```
tgoyne, line 209: Adding this overload does bad things. The existential is considered a "better" match than a generic function, so everything goes through this overload instead of the generic one. The type erasure then means that every call to `AnyBSON` does a linear scan over the possible types to determine which it is.

examples/ios/swift/ListSwiftUI/Views/ContentView.swift
```diff
@@ -268,9 +268,7 @@ struct ContentView: View {
     var body: some View {
         NavigationView {
             VStack {
-                if #available(iOS 15.0, macOS 12.0, tvOS 15.0, watchOS 8.0, *) {
-                    // Don't add a SearchView in case searchable is available
-                } else {
+                if #unavailable(iOS 15.0, macOS 12.0, tvOS 15.0, watchOS 8.0) {
```
tgoyne, line 271: #unavailable was added in Swift 5.6 and our minimum supported version is 5.5.

dianaafanador3: I disable the swift lint error while we are on swift 5.5 minimum version.

leemaguire (approved):

tgoyne (approved):

PR #7742, opened 2022-04-10
Init Randomized tests

jsflax:
The basic gist is that I log in 3 different users and dispatch 1000 random RealmActions to be executed concurrently for each user. The schema and objects are dynamically generated.

- adding to a realm will add a random object type
- removing from a realm will remove a random object of each type
- modifying in a realm will take a random object of each type and modify each of its properties in a randomised way
    - for primitives, this is a pretty straight forward “set the value to a random value”
    - for collections, these are considered separate operations and take a different path, and lists/sets have their own types of random modifications
- reading from a realm is basically a full scan of the db, reading every property of every object

Notably missing things:
- it doesn’t intentionally attempt to provide garbage input anywhere, which is i guess more what fuzz testing is meant to be
- testing lists of primitives, which is rote work i just haven’t done yet
- testing maps which is nontrivial
- subscriptions are currently hardcoded to all be TRUEPREDICATE to get the broadest scope of things
- opening a realm with randomised configurations
- and obviously a ton of infrastructurey things. once the test suite fails somewhere, the test suit just aborts.
- cleanup


jsflax:
Note that this is more _randomized_ testing as opposed to _fuzz_ testing, but I do believe this is at least the start to _something_.

mpobrien:
@jsflax can this be closed out?

PR #7753, opened 2022-04-20
Fix XCFramework with library evolution

leemaguire:
Fixes: https://github.com/realm/realm-swift/issues/7313

This does not test against all platforms because it would hog the CI resources.

### The flow of the test is as follows:

1. XCFramework is built with the oldest available Xcode version.
2. The XCFramework is referenced in the project located in `examples/installation/iOS/swift/XCFrameworkExample`
3. Switch to the latest Xcode version.
4. The test in that project is run.

Tested on Cocoa Pipeline on CI


build.sh
```diff
@@ -440,6 +442,44 @@ case "$COMMAND" in
...
+        fi 
+
+        exit 0
+        ;;
+
+    "verify-xcframework-evolution-mode")
+        export REALM_EXTRA_BUILD_ARGUMENTS="$REALM_EXTRA_BUILD_ARGUMENTS REALM_BUILD_LIBRARY_FOR_DISTRIBUTION=YES"
+        # set the Xcode version to the oldest
+        export REALM_XCODE_VERSION=$REALM_XCODE_OLDEST_VERSION
+        export REALM_SWIFT_VERSION=$REALM_SWIFT_OLDEST_VERSION
```
tgoyne, line 472: Setting both REALM_SWIFT_VERSION and REALM_XCODE_VERSION is unnecessary. The Swift version is automatically set from the Xcode version, and it being a separate thing is just old cruft from when Xcode packaged multiple versions of the Swift toolchain.

tgoyne (approved):

PR #7755, opened 2022-04-21
Flexible Sync V2 (Proposed 1)

dianaafanador3:
This PR contains a complete new API for Flexible Sync, intended to make flexible sync less confusing to the developer.

```swift
let app = App(id: "my-app-id")
let user = try await app.login(credentials: .anonymous)
let realm = try user.realm(configuration: .defaultConfiguration)

// You can subscribe and get `QueryResults` from 1-12 queries at the same time, this subscriptions will be done within a batch write
let (persons, dogs, birds) = try await realm.subscriptions.subscribe(to: 
QuerySubscription<Person> { $0.age >= 18}, 
QuerySubscription<Dog>(), //Intended to query all elements for this object 
QuerySubscription<Bird> { $0.species == .magpie })

print(persons.count) //20

let newPerson = Person()
newPerson.age = 18
try realm.write {
    realm.add(newPerson)
}

// QueryResults will be updated if the values are within the subscription query 
print(persons.count) //21 

// Unsubscribe will remove the values from the `QueryResult` and remove the subscription from the subscription set.
try await persons.unsubscribe()

// Reusing a `QueryResults` is as easy as calling again subscribe for the same query
let persons: QueryResults<Person>= try await realm.subscriptions.subscribe(to: { $0.age >= 18 })
// Or search in your subscription set
let foundedSubscription = realm.subscriptions.first(ofType: SwiftPerson.self, where: { { $0.age >= 18 })

// You can use unsubscribeAll to remove all subscriptions from the subscription set
try await realm.subscriptions.unsubscribeAll()

// Or you can remove them by type
try await realm.subscriptions.unsubscribeAll(ofType: Person.self)
```

SwiftUI
You can use `@ObservedResults` to add a subscription and retrieve data from it
We are adding a new property to `@ObservedResults`which includes a state value, this state will publish any subscription state changes to view.
```swift
public enum SubscriptionState {
    case pending //Subscription has been added and waiting for data to bootstrap.
    case error(Error) // An error has occurred while adding the subscription (client or server side)
    case completed // Data has been bootstrapped and query results updated.
}
```
```swift
@available(macOS 12.0, *)
struct ObservedQueryResultsStateView: View {
    @ObservedResults(Person.self, where: { $0.age > 18 && $0.firstName == ProcessInfo.processInfo.environment["firstName"]! })
    var persons

    var body: some View {
        VStack {
            switch $persons.state {
            case .pending:
                ProgressView()
            case .completed:
                List {
                    ForEach(persons) { person in
                        Text("\(person.firstName)")
                    }
                }
            case .error(let error):
                ErrorView(error: error)
                    .background(Color.red)
                    .transition(AnyTransition.move(edge: .trailing)).animation(.default)
            }
        }
    }
}

@available(macOS 12.0, *)
struct ObservedQueryResultsView: View {
    @ObservedResults(Person.self,  where: { $0.age >= 15 && $0.firstName == ProcessInfo.processInfo.environment["firstName"]! })
    var persons
    @State var searchFilter: String = ""

    var body: some View {
        VStack {
            if persons.isEmpty {
                ProgressView()
            } else {
                List {
                    ForEach(persons) { person in
                        HStack {
                            Text("\(person.firstName)")
                            Spacer()
                            Text("\(person.age)")
                        }
                    }
                }
            }
        }
        .onDisappear {
            Task {
                do {
                    try await $persons.unsubscribe()
                }
            }
        }
    }
}
```

Notes
* AnyQueryResults??, didn't add the implementation for using this as a collection, which can be added if needed but I don’t consider this to be a priority, which brings me to the issue if the user is really going to use this instead of doing a search for the query or subscribe again to the query which will retrieve the results without adding another subscription.

### DIFF
--- a/Realm/ObjectServerTests/SwiftFlexibleSyncServerTests.swift
+++ b/Realm/ObjectServerTests/SwiftFlexibleSyncServerTests.swift
@@ -522,6 +522,146 @@ extension SwiftFlexibleSyncTests {
         XCTAssertEqual(originalPersons.count, 0)
         XCTAssertEqual(persons!.count, 0)
     }
+
+    @MainActor
+    func testFlexibleSyncInitialSubscriptionsAsync() async throws {
+        try await populateFlexibleSyncDataForType(SwiftPerson.self) { realm in
+            for i in 1...20 {
+                let person = SwiftPerson(firstName: "\(#function)", lastName: "lastname_\(i)", age: i)
+                realm.add(person)
+            }
+        }
+
+        let user = try await logInUser(for: basicCredentials(app: self.flexibleSyncApp), app: self.flexibleSyncApp)
+        var config = user.flexibleSyncConfiguration(initialSubscriptions: { subscriptions in
+            subscriptions.append(ofType: SwiftPerson.self,
+                                 where: { $0.age > 10 && $0.firstName == "\(#function)" })
+        })
+
+        if config.objectTypes == nil {
+            config.objectTypes = [SwiftPerson.self]
+        }
+        let realm = try await Realm(configuration: config, downloadBeforeOpen: .once)
+        XCTAssertNotNil(realm)
+
+        XCTAssertEqual(realm.subscriptions.count, 1)
+        // Adding this sleep, because there seems to be a timing issue after this commit in baas
+        // https://github.com/10gen/baas/commit/64e75b3f1fe8a6f8704d1597de60f9dda401ccce,
+        // data take a little longer to be downloaded to the realm even though the
+        // sync client changed the subscription state to completed.
+        sleep(1)
+        checkCount(expected: 10, realm, SwiftPerson.self)
+    }
+
+    @MainActor
+    func testFlexibleSyncInitialSubscriptionsNotRerunOnOpen() async throws {
+        let user = try await logInUser(for: basicCredentials(app: self.flexibleSyncApp), app: self.flexibleSyncApp)
+        var config = user.flexibleSyncConfiguration(initialSubscriptions: { subscriptions in
+            subscriptions.append(ofType: SwiftPerson.self,
+                                 where: { $0.age > 10 && $0.firstName == "\(#function)" })
+        })
+
+        if config.objectTypes == nil {
+            config.objectTypes = [SwiftPerson.self]
+        }
+        let realm = try await Realm(configuration: config, downloadBeforeOpen: .once)
+        XCTAssertNotNil(realm)
+        XCTAssertEqual(realm.subscriptions.count, 1)
+
+        let realm2 = try await Realm(configuration: config, downloadBeforeOpen: .once)
+        XCTAssertNotNil(realm2)
+        XCTAssertEqual(realm.subscriptions.count, 1)
+    }
+
+    @MainActor
+    func testFlexibleSyncInitialSubscriptionsRerunOnOpen() async throws {
+        try await populateFlexibleSyncDataForType(SwiftTypesSyncObject.self) { realm in
+            for i in 1...30 {
+                let object = SwiftTypesSyncObject()
+                object.dateCol = Calendar.current.date(
+                    byAdding: .hour,
+                    value: -i,
+                    to: Date())!
+                realm.add(object)
+            }
+        }
+
+        let user = try await logInUser(for: basicCredentials(app: self.flexibleSyncApp), app: self.flexibleSyncApp)
+        var isFirstOpen = true
+        var config = user.flexibleSyncConfiguration(initialSubscriptions: { subscriptions in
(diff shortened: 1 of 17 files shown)


RealmSwift/SyncSubscription.swift
```diff
@@ -50,159 +50,255 @@ import Realm.Private
...
-                                  which will be used to modify the query.
+    - parameter type: The type of the results to return.
      */
-    public func update(to predicateFormat: String, _ args: Any...) {
-        _rlmSyncSubscription.update(with: NSPredicate(format: predicateFormat, argumentArray: unwrapOptionals(in: args)))
+    public func `as`<T: Object>(type: T.Type) -> QueryResults<T>? {
+        guard _rlmSyncSubscription!.objectClassName == "\(T.self)" else {
+            return nil
+        }
+        return QueryResults(_rlmSyncSubscription!, results.realm!.objects(T.self).filter(_rlmSyncSubscription!.queryString))
```
leemaguire, line 158:
```suggestion
        return QueryResults(_rlmSyncSubscription!, unsafeBitCast(results, to: Results<T>.self))
```

RealmSwift/SyncSubscription.swift
```diff
@@ -50,159 +50,255 @@ import Realm.Private
...
+
+     - throws: An `NSError` if t
+     */
+    func unsubscribe() async throws
+    #endif // canImport(_Concurrency)
+}
+
+// `SyncSubscription` includes all the common implementation for a subscription.
+internal protocol SyncSubscription: _SyncSubscription {
+    var _rlmSyncSubscription: RLMSyncSubscription? { get }
```
leemaguire, line 80: There is no need for this to be optional

RealmSwift/SwiftUI.swift
```diff
@@ -617,6 +617,292 @@ extension Projection: _ObservedResultsValue { }
...
+        var searchString: String = ""
+        var searchFilter: NSPredicate? {
+            didSet {
+                didSet()
+            }
+        }
+
+        func subscribe() {
+            guard let configuration = configuration else {
+                throwRealmException("")
```
leemaguire, line 702: These need actual errors

Realm/ObjectServerTests/SwiftFlexibleSyncServerTests.swift
```diff
@@ -52,999 +57,468 @@ class SwiftFlexibleSyncTests: SwiftSyncTestCase {
...
-            if error == nil {
-                ex2.fulfill()
-            } else {
-                XCTFail("Subscription Set could not complete with \(error!)")
-            }
-        })
-        waitForExpectations(timeout: 20.0, handler: nil)
+        var count = 0
+        for subscription in realm.subscriptions {
+            XCTAssertNotNil(subscription)
```
leemaguire, line 417: should assert the properties of the object the iterator produces too

Realm/ObjectServerTests/SwiftFlexibleSyncServerTests.swift
```diff
@@ -52,999 +57,468 @@ class SwiftFlexibleSyncTests: SwiftSyncTestCase {
...
-                QuerySubscription<SwiftTypesSyncObject>(name: "object_int_more_than_zero") {
-                    $0.intCol > 0
-                },
-                QuerySubscription<SwiftTypesSyncObject>(name: "object_string") {
-                    $0.stringCol == "John" || $0.stringCol == "Tom"
-                })
-        }
-        XCTAssertEqual(subscriptions.count, 3)
+        // This will trigger a client reset, which will result in the server not responding to any instruction, this is not removing the object from the database.
+//        let newPerson = SwiftPerson()
```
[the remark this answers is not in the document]
dianaafanador3: this will work as it is supposed when compensating writes project is finished, when this is finished on server we will have to update our baas version and test it against it

RealmSwift/SwiftUI.swift
dianaafanador3: I removed all the code and went with the ObservedResults approach

RealmSwift/SyncSubscription.swift
```diff
@@ -220,235 +316,597 @@ import Realm.Private
...
+        try await write {
+            subscriptions.forEach { subscription in
+                self.append(subscription)
+            }
+        }
     }
 
     /**
-     Removes a subscription with the specified query.
+     Appends the query to the current subscription set and wait for the server to acknowledge the subscription,
```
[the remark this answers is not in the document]
dianaafanador3: Because we are extending from `RealmCollectionImpl` on QueryResults to make use of the default implementation of a realm  ollection, we need to this to be optional.

mpobrien:
@dianaafanador3 is this PR something we still want to move forward with?

PR #7766, opened 2022-04-22
(work in progress) Add Ownable<T> and OwnableResults<T>

jasperblues:
PR is too messy to merge as-is (added some of our utility classes for testing) however using it to serve as an executable discussion: 

# The Problem 

Modern iOS and Mac apps depend on very high levels of concurrency. Realm provides utilities for passing objects between threads, however the current (lack of) documentation and approaches mean that it takes a lot of research and then boiler-plate code to get it done. The upshot is that most apps using Realm objects to update the UI end up using Realm on the main thread. Often the performance is pretty good, but for large datasets, it can be an issue. 

## Proposal: Simple Way to Pass Between Threads. 

I introduce an `Ownable<T>` and `OwnableResults<T>` that can be returned from concurrent code, like: 

* Promises
* Async Functions
* Rx____ and Combine, etc 

Use it like so: 

```swift 
    func loadSeveral() async throws -> OwnableResults<Message> {
        let messages = try Realm.defaultRealm.objects(Message.self)
            .filter("color = %@", "Orange")
        return OwnableResults(messages)
    }
```

Above function will run in its own (unknown) thread. We can then pass ownership of the returned Realm objects to another (including the main thread, when calling it) like this: 

```
let results = try await loadSeveral().take() // Now use the results on a new thread
```

Similarly we have an `Ownable<T>` type for a single item. (We could probably combine the one class to support anything that's Realm-backed) if we wanted. 

## Implementation Details

The `Ownable<T>` is just a pure swift class: 

```
public class Ownable<T: ThreadConfined> {

    private(set) var realm: Realm?
    private(set) var threadConfined: ThreadSafeReference<T>?
    private(set) var unmanaged: T?

    public convenience init(item: T) {
        if let realm = item.realm {
            let wrappedObj = ThreadSafeReference(to: item)
            self.init(threadConfined: wrappedObj, unmanaged: nil, realm: realm)
        } else {
            self.init(threadConfined: nil, unmanaged: item, realm: nil)
        }
    }

    public init(threadConfined: ThreadSafeReference<T>?, unmanaged: T?, realm: Realm?) {
        self.threadConfined = threadConfined
        self.unmanaged = unmanaged
        self.realm = realm
    }

    public func take() throws -> T {
        if let threadConfined = threadConfined, let realm = realm {
            if let result = realm.resolve(threadConfined) {
                return result
            } else {
                // TODO: Throw errors the Realm way
                throw RealmError("Object was deleted in another thread!!")
            }
        } else if let unmanaged = unmanaged {
            return unmanaged
        } else {
            throw RealmError("Require either a realm bound or managed object")
        }
    }
}
```

### RLMOwnableCollection 

For Results<T> it seems that the Swift API didn't expose the internals needed, so I created an Objective-C class: 

```objc
@class RLMRealm;
@class RLMThreadSafeReference;
@protocol RLMThreadConfined;
@protocol RLMCollection;

@interface RLMOwnableCollection<__covariant Confined : id <RLMThreadConfined>> : NSObject

@property(nonatomic, strong, readonly) RLMThreadSafeReference *__nonnull threadConfined;
@property(nonatomic, strong, readonly) RLMRealm *__nonnull realm;

- (__nonnull instancetype)initWithItems:(id<RLMCollection> __nonnull)items;

- (__nonnull instancetype)initWithThreadConfined:(RLMThreadSafeReference *__nonnull)threadConfined
                                           realm:(RLMRealm *__nonnull)realm;

- (__nonnull Confined)take;

@end
```

With implementation . . . 

```objc
@implementation RLMOwnableCollection

- (instancetype)initWithItems:(id<RLMCollection>)items {
    if (items.realm != nil) {
        id reference = [RLMThreadSafeReference referenceWithThreadConfined:items];
        return [self initWithThreadConfined:reference realm:items.realm];
    } else {
        [NSException raise:NSInternalInconsistencyException format:@"Item realm can't be nil"];
        return nil;
    }
}

- (instancetype)initWithThreadConfined:(RLMThreadSafeReference *)threadConfined realm:(RLMRealm *)realm {
    self = [super init];
    if (self) {
        _threadConfined = threadConfined;
        _realm = realm;
    }
    return self;
}

- (id)take {
    id result = [_realm resolveThreadSafeReference:_threadConfined];
    if (result) {
        return result;
    } else {
        [NSException raise:NSInternalInconsistencyException format:@"Object was deleted in another thread"];
        return nil;
    }
}


@end
```

And it has a little Swift wrapper around it: 

```swift
public class OwnableResults<T: RealmCollectionValue> {

    private(set) var ownableCollection: RLMOwnableCollection<RLMCollection>

    public init(_ results: Results<T>) {
        ownableCollection = RLMOwnableCollection(items: results.collection)
    }

    public func take() throws -> Results<T> {
        Results<T>(collection: ownableCollection.take())
    }
}
```

Again we could probably have a single `Ownable<T>` class that works for anything that's bound to a realm, but this is a proof of concept. 

### Notes

Note that docs state you can pass an object to between threads, but they don't address what to do if the object is not yet in a realm, so users get confused, like in this StackOverflow issue: 

* https://stackoverflow.com/questions/57617719/how-to-save-normal-array-into-realm-in-background-thread-swift

^-- We can see how its useful to address whether the object is already in a Realm or not. However doing this every time is a lot of boiler-plate, so its useful to have a wrapper to handle it. 

Here are some other issues where folks are struggling with threading: 

* https://stackoverflow.com/questions/67336606/reaching-realm-results-object-from-different-thread-gives-crash
* https://stackoverflow.com/questions/37525964/how-to-deal-with-thread-and-realm-ios

### DIFF
--- a/Realm.podspec
+++ b/Realm.podspec
@@ -44,6 +44,7 @@ Pod::Spec.new do |s|
                               'include/RLMSwiftValueStorage.h',
                               'include/RLMThreadSafeReference.h',
                               'include/RLMValue.h',
+                              'include/RLMOwnableCollection.h',
 
                               # Sync
                               'include/NSError+RLMSync.h',
--- a/Realm.xcodeproj/project.pbxproj
+++ b/Realm.xcodeproj/project.pbxproj
@@ -464,6 +464,17 @@
 		ACC6BEE62796EA78006C1728 /* RLMSyncSubscription.h in Headers */ = {isa = PBXBuildFile; fileRef = 53E308CA27905794002A8D91 /* RLMSyncSubscription.h */; settings = {ATTRIBUTES = (Public, ); }; };
 		ACC6BEE72796ED6C006C1728 /* RLMSyncSubscription.mm in Sources */ = {isa = PBXBuildFile; fileRef = 531F956B27906F7600E497F1 /* RLMSyncSubscription.mm */; };
 		ACF08B6726DD936200686CBC /* Query.swift in Sources */ = {isa = PBXBuildFile; fileRef = ACF08B6626DD936200686CBC /* Query.swift */; };
+		BA79815D9647D01921938C5A /* RLMOwnableCollection.h in Headers */ = {isa = PBXBuildFile; fileRef = BA798B95AA33448D6A7008A0 /* RLMOwnableCollection.h */; settings = {ATTRIBUTES = (Public, ); }; };
+		BA7983CB4661565356800EDA /* Ownable.swift in Sources */ = {isa = PBXBuildFile; fileRef = BA798EF69698FEF55B403301 /* Ownable.swift */; };
+		BA79851F3F507B88B52A8BE7 /* BusyTown.swift in Sources */ = {isa = PBXBuildFile; fileRef = BA7989238B03E75B5A8A483F /* BusyTown.swift */; };
+		BA79859EEE05196EEFB5331B /* BusyTownTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = BA7988D3D3663B21BC36581B /* BusyTownTests.swift */; };
+		BA79893551C6DA98CDC294B6 /* RealmProvider.swift in Sources */ = {isa = PBXBuildFile; fileRef = BA798DF10142DDB73AFE15D9 /* RealmProvider.swift */; };
+		BA7989E011BBEE5E788F07BE /* Message.swift in Sources */ = {isa = PBXBuildFile; fileRef = BA798E7F2A7899A2719C3954 /* Message.swift */; };
+		BA798A18F33D842D8E8EFD72 /* RLMOwnableCollection.m in Sources */ = {isa = PBXBuildFile; fileRef = BA798C70EDE6781C01F32606 /* RLMOwnableCollection.m */; };
+		BA798B204BEA767074BF3F6D /* OwnableResults.swift in Sources */ = {isa = PBXBuildFile; fileRef = BA7980FE54040A4B84689038 /* OwnableResults.swift */; };
+		BA798EE7176C80CAD69D63CD /* RealmError.swift in Sources */ = {isa = PBXBuildFile; fileRef = BA7986D5AD8A9B811252A459 /* RealmError.swift */; };
+		BA798F8FC9E7554DBD28A423 /* Realm+Extensions.swift in Sources */ = {isa = PBXBuildFile; fileRef = BA7982B21912163B3CE0F00E /* Realm+Extensions.swift */; };
+		BA798FA37D65BAE3382B2885 /* RLMOwnableCollection.m in Sources */ = {isa = PBXBuildFile; fileRef = BA798C70EDE6781C01F32606 /* RLMOwnableCollection.m */; };
 		C042A48D1B7522A900771ED2 /* RealmConfigurationTests.mm in Sources */ = {isa = PBXBuildFile; fileRef = C042A48C1B7522A900771ED2 /* RealmConfigurationTests.mm */; };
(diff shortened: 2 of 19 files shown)


rydamckinney:
+1 on the necessity of this abstraction

wonder2011:
@jasperblues awesome work with this. I love realm and with some like this i will love more.

jsflax:
Hey @jasperblues , we really appreciate the pull request! Always nice to see community submissions.

It is important to understand your motivation here though– to be clear it seems that what you're adding here is twofold– a wrapper for various classes/collections that removes the boilerplate for ThreadSafeReferences and a writeAsync method?

We actually recently added both of these things 😅. Check out our `@ThreadSafe` property wrapper: https://github.com/realm/realm-swift/blob/master/RealmSwift/Tests/ThreadSafeReferenceTests.swift#L719

```swift
        @ThreadSafe var swiftStringObject = results!.first
        queue.async {
            guard let swiftStringObject = swiftStringObject else {
                return XCTFail("no results")
            }

            XCTAssertEqual(swiftStringObject.stringCol, "A")
        }
```

And for `writeAsync`, we've also just recently added that as a feature flagged feature: https://github.com/realm/realm-swift/blob/master/RealmSwift/Tests/RealmTests.swift#L985

```swift
let realm = try! Realm()
        let asyncComplete = expectation(description: "async transaction complete")

        realm.writeAsync {
            realm.create(SwiftStringObject.self, value: ["string"])
        } onComplete: { _ in
            let object = realm.objects(SwiftStringObject.self).first
            XCTAssertEqual(object?.stringCol, "string")
        }
```

We are aware concurrency + Realm can be cumbersome, which is why we released these features. What would be great for us is if you could check out these new features and see if you feel that anything is missing or if there is room for improvement.

jasperblues:
Hi @jsflax you're correct about the intention of the PR. To remove the boilerplate and allow working at a level of abstraction close to the persistence domain (having to do with getting things in and out of realm). 

It was intended to open a discussion and show how things might look, not a final solution. But ah yes! I see that we do have new stuff that addresses these. 

I will try them out and let you know how they work for us. 

As for my own approach: I noticed it worked in my test app, but didn't (still got a 'ream accessed from incorrect thread' error) in our own more complex app. I'm not sure if the root cause of that was in Realm or my solution, but will keep you posted there too. 

Let's close this PR then, and I'll open new tickets/PRs if I see something still unaddressed or not working as expected in the new approaches. 

Thanks everyone from Realm 👍

jasperblues:
@jsflax I finally got time to try the @ThreadSafe wrapper. Sorry for delays. 

* tldr1: It would be useful to be able to pass `Results<T>` in a thread-safe way. 
* tldr2: If this is supported are there any tests or docs showing such usage? 

--
*Details:*

Its clear from referenced tests how to use `@ThreadSafe` to extract a single result, at a time, from a collection. This is good, for manually handling results.  After I created my wrapper for a single result, we tested general utility and found it works in some cases, but not for eg observers of a whole collection, where we use them like so: 

swift
```
let liveQuery: Results<MyDomainType> = ....
self.notificationToken = liveQuery.observe { [weak self] (changes: RealmCollectionChange) in
        switch changes {
        case .initial(let results):
            //populate the view
        case .update(_, let deletions, let insertions, let modifications):
            //update the view
            self?.refreshViewModels()
        case .error(let error):
            //handle errors
        }
    }
```

Above is a useful pattern for collections in both SwiftUI and UIKit apps. 

I tested the `@ThreadSafe` wrapper (btw this is much cooler than a class-based wrapper) and noticed the same. Did I miss something? Is there a way to use the `@ThreadSafe` wrapper for `Results<T>` like I addressed in the PR, or is it something still pending here too? 

// cc @rydamckinney

PR #7773, opened 2022-04-27
Upgrade to core v11.15.0

tgoyne:


Realm/ObjectServerTests/RLMObjectServerTests.mm
```diff
@@ -2823,20 +2825,18 @@ - (void)testSyncFilePaths {
...
-    s << RLMConvertRLMBSONToBson(partitionValue);
-    NSString *encodedPartitionValue = [@(s.str().c_str()) stringByAddingPercentEncodingWithAllowedCharacters:[NSCharacterSet URLPathAllowedCharacterSet]];
-    realm::SyncConfig config(user._syncUser,
-                             [[NSString alloc] initWithFormat:@"%@/%@", user.identifier, encodedPartitionValue].UTF8String);
-    return @(user._syncUser->sync_manager()->path_for_realm(config).c_str());
+static NSString *oldPathForPartitionValue(RLMUser *user, id<RLMBSON> partitionValue, NSString *oldName) {
+    realm::SyncConfig config(user._syncUser, "null");
+    return [NSString stringWithFormat:@"%@/%s%@.realm",
+            [@(user._syncUser->sync_manager()->path_for_realm(config).c_str()) stringByDeletingLastPathComponent],
+            user._syncUser->identity().c_str(), oldName];
```
tgoyne, line 2832: The previous version of this test was broken by core-side changes to be a bit more strict about invalid input. I switched it to hardcoding more of the expected value because the whole point of this test is that the expected values should never change.

leemaguire (approved):

PR #7778, opened 2022-04-29
Fix some minor race conditions in client reset callbacks and speed up the sync tests

tgoyne:
While refactoring config handling for audit stuff (similarly to client reset callbacks, it needs to use the parent config to open RLMRealm instances), I found some minor bugs in the current implementation. The SyncConfig wasn't actually copied when the parent config was, so we were mutating the input SyncConfig when configuring the client reset callbacks. This was mostly harmless: we don't expose the ability for the user to mutate the SyncConfig attached to a RLMRealmConfiguration and they can only replace it with a new one. However, this meant that opening a RLMRealm on one thread while a client reset was happening on another thread would be a data race. Probably impossible for this to actually break, but it's still UB and should be avoided.

The callbacks also were initialized too late for async open. They were configured when the async open completed rather than at the start, which in practice was probably always fine but in theory the client reset could happen before the async open completed.

I added a test for client reset handlers with async open and refactor the client reset test to write the data which is expected to be preset post-reset directly to mongo, which makes it so that we don't have to wait for the translator to catch up and knocks multiple minutes off the test's runtime.

Changing the polling interval for baas makes it so that it doesn't sometimes sit there doing nothing for an entire minute. This speeds up the tests a bit and hopefully will make it so that the flx sync tests stop sometimes timing out.

I then got sidetracked and spent some time making other sync tests faster. In total this cuts the runtime of the complete SPM test suite on my machine from 1120 seconds to 850 seconds, so hopefully it'll make those jobs stop timing out on CI.


tgoyne:
Remaining CI failures look unrelated.

Realm/ObjectServerTests/SwiftCollectionSyncTests.swift
```diff
@@ -351,72 +296,48 @@ class MapSyncTests: CollectionSyncTestCase {
...
                                                    UUID(uuidString: "6b28ec45-b29a-4b0a-bd6a-343c7f6d90fe")!,
                                                    UUID(uuidString: "6b28ec45-b29a-4b0a-bd6a-343c7f6d90ff")!,
                                                    UUID(uuidString: "6b28ec45-b29a-4b0a-bd6a-343c7f6d90ae")!,
                                                    UUID(uuidString: "6b28ec45-b29a-4b0a-bd6a-343c7f6d90bf")!])
-    }
-
-    // FIXME: We need to add a test where a value in a map of objects is `null`. currently the server
-    // is throwing a bad changeset error when that happens.
-    func testObjectMap() throws {
+        // FIXME: We need to add a test where a value in a map of objects is `null`. currently the server
```
leemaguire, line 334: Do we know if this is fixed yet?

ericjordanmossman (approved):
Looks good to me. Comments were just thoughts I had.

Realm/ObjectServerTests/SwiftCollectionSyncTests.swift
tgoyne: I have no idea.

leemaguire (approved):

PR #7783, opened 2022-05-04
Implement the Events API

tgoyne:
Not mergeable yet as it depends on https://github.com/realm/realm-core/pull/5436, but the code changes here are reviewable now.


leemaguire (commented):
This is looking pretty good, I left a few small comments. I also notice that there are no Objective-C tests, but given this project's use case and implementation I think just Swift tests should be enough?

Realm/RLMAccessor.mm
```diff
@@ -921,6 +923,11 @@ void RLMSetSwiftPropertyAny(__unsafe_unretained RLMObjectBase *const obj, uint16
 }
 
 id RLMAccessorContext::box(realm::Obj&& r) {
+    if (!currentProperty) {
+        // If currentProperty is set then the caller took care of reporting
+        // the audit event
+        realm::Object(_realm->_realm, *_info.objectSchema, r, _parentObject, _colKey);
```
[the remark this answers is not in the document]
tgoyne: Creating an `Object` instance reports an audit read. I'll make the comment more descriptive.

Realm/RLMAudit.h
```diff
@@ -0,0 +1,59 @@
...
+                                              NSDictionary<NSString *, NSString *> *newMetadata);
+
+@interface RLMAuditConfiguration : NSObject
+@property (nonatomic) NSString *partitionPrefix;
+@property (nonatomic, nullable) RLMUser *syncUser;
+@property (nonatomic, nullable) NSDictionary<NSString *, NSString *> *metadata;
+@property (nonatomic, nullable) void (^logger)(RLMSyncLogLevel, NSString *);
+@property (nonatomic, nullable) void (^errorHandler)(NSError *);
+
+#ifdef __cplusplus
```
jsflax, line 54: Why is there cpp code exposed in this header?

Realm/RLMEvent.h
```diff
@@ -0,0 +1,59 @@
...
+////////////////////////////////////////////////////////////////////////////
+
+#import <Foundation/Foundation.h>
+#import <Realm/RLMConstants.h>
+
+#ifdef __cplusplus
+#include <memory>
+
+namespace realm {
+struct AuditConfig;
```
bmunkholm, line 26: Shouldn't this be "EventConfig"?

Realm/RLMEvent.h
```diff
@@ -0,0 +1,59 @@
...
+
+@interface RLMEventConfiguration : NSObject
+@property (nonatomic) NSString *partitionPrefix;
+@property (nonatomic, nullable) RLMUser *syncUser;
+@property (nonatomic, nullable) NSDictionary<NSString *, NSString *> *metadata;
+@property (nonatomic, nullable) void (^logger)(RLMSyncLogLevel, NSString *);
+@property (nonatomic, nullable) void (^errorHandler)(NSError *);
+
+#ifdef __cplusplus
+- (std::shared_ptr<realm::AuditConfig>)auditConfigWithRealmConfiguration:(RLMRealmConfiguration *)realmConfig;
```
bmunkholm, line 55: auditConfigWithRealmConfiguration -> eventConfigWithRealmConfiguration ?

bmunkholm (commented):
It looks like the "audit" naming hasn't been completely updated to "event". But also that it's not external facing so likely ok for now. It would be good to rename the internals at some point for internal consistency.

Realm/RLMEvent.h
tgoyne: No, all of the c++ stuff is still named Audit because that renaming came after it was merged.

PR #7789, opened 2022-05-10
Fix issue with creating a Frozen Realm after writeCopy

leemaguire:
Fixes: https://github.com/realm/realm-swift/issues/7697

The issue is that a freshly created realm will not have a ‘metadata’ table. The importance of this is that this table holds the schema version number as m_schema_version will be ’NotVersioned’ a write will be committed and the schema will be cached.

The error occurring using a `writeCopy` Realm after trying to freeze it is that because this copied realm already has a metadata table there is no need to run a write transaction and cache the schema. The write transaction version from the original Realm also will not be copied to the new one.

The freeze method in cocoa relies on the transaction version to be bumped in order to track the new Realm to be frozen. Because the transaction number will always be the same (because it is never bumped) the Realm stored in the NSMapTable will be returned for that key and a new frozen Realm will never be created.

### Solution
Hash both the current transaction version number and a flag of `true` to create a unique key for the frozen Realm. (This is the same as just artificially advancing the version number by one which should be adequate)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -49,6 +49,8 @@ x.y.z Release notes (yyyy-MM-dd)
 * If an async open of a Realm triggered a client reset, the callbacks for
   `discardLocal` could theoretically fail to be called due to a race condition.
   The timing for this was probably not possible to hit in practice (since 10.25.0).
+* Calling `[RLMRealm freeze]`/`Realm.freeze` on a Realm which had been created from `writeCopy`
+  would not produce a frozen Realm. ([#7697](https://github.com/realm/realm-swift/issues/7697), since v5.0.0)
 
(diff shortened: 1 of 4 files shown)


leemaguire:
CI failures are unrelated.

tgoyne:
The problem here is just that the main thread/queue Realm is using the cache key 1, which is also a valid version number. The main thread Realm needs to instead use a cache key that isn't a valid version or a valid pointer, such as `std::numeric_limits<uintptr_t>::max()`.

Hashing the version is just working around this by coincidence, and makes it so that frozen versions can conflict with non-frozen Realms. The cache relies on the idea that the smallest valid heap pointer is larger than the largest possible frozen version, and hashing the version projects frozen keys into the key space used by live keys. Adding `hasher(reinterpret_cast<void *>(true)))` to the result is just adding a large constant number, which doesn't really do much of anything.

PR #7792, opened 2022-05-11
Lower minimum version for more concurrency extensions

jsflax:
It would appear that we missed two more places where we didn't reduce the minimum version of the concurrency module.


dianaafanador3 (approved):
LGTM, probably this needs a changelog so users know when this was fixed.

tgoyne (approved):

PR #7829, opened 2022-06-07
Add Xcode 14 support

### LINKED ISSUE #7828: Xcode 14Beta Build Failed

All the time

### Description

When I build project,  RealmSwift/SwiftUI.swift has some error

### Stacktrace & log output

```shell
Type 'V' does not conform to protocol 'RealmCollectionValue'
Reference to member 'remove' cannot be resolved without a contextual type
```

Yes, always

### Reproduction Steps

use Xcode 14Beta building

### Version

10.27.0

Local Database only

### Are you using encryption?

No, not using encryption

MacOS 12.4

### Build environment

Xcode version: 14.0Beta

tgoyne:
Fixes #7828.

Swift 5.7 doesn't like the clever thing we're doing with generics in the SwiftUI extensions (and gives contradictory errors), but we can just not do that and split them into separate extensions.

Our use of sprintf was safe but Xcode 14 has quite reasonably started warning on all uses of it.

tvOS and watchOS are now separate downloads from Xcode and I haven't updated the CI VM build script to install them yet, so we can currently only build them on Xcode 13 on CI. They work fine for me locally with the SDKs installed.

The catalyst build was breaking in weird ways with the iOS deployment target set to 9. We probably want to bump it anyway now that 32-bit iOS is gone, so I bumped it to 12 when building with Xcode 14.

The `cocoapods-ios` CI failures are a CI configuration issue of some sort and those tests work for me locally.

The UI tests required some assorted changes because SwiftUI uses different UI elements for some things on iOS 16.


tgoyne:
I think the remaining CI failures are preexisting CI flakeyness.

alexmx:
Hi, is there a reason why the iOS 11 min deployment target for Xcode 14 can't be kept? The bump from 9 to 12 is quite an aggressive one.

tgoyne:
We held back updating the deployment target for several years to maintain support for 32-bit devices, but Xcode 14 has finally dropped support for those. iOS 12 supports all devices that can run iOS 11 and all devices which Xcode 14 can build for.

tgoyne:
If there's a good reason to continue to support iOS 11 that is an option, but we do benefit from going up to 12 (that's when they shipped most of the c++17 standard library) and there doesn't appear to be a downside.

alexmx:
> If there's a good reason to continue to support iOS 11 that is an option, but we do benefit from going up to 12 (that's when they shipped most of the c++17 standard library) and there doesn't appear to be a downside.

@tgoyne My company can't drop iOS 11 that easily as we're in a highly regulated market (banking app) and dropping an OS version is a lengthy process. We still have a significant amount of customers using iOS 11 devices, forcing them to update their OSes all at once is not an option for us as we're still supporting iOS 10.3. We could drop iOS 10 with plenty of notice but not both 10 and 11 together.

If Realm stops supporting iOS 11 we would probably need to seek for alternatives unfortunately.

laosb:
> If Realm stops supporting iOS 11 we would probably need to seek for alternatives unfortunately.

I mean, for legacy usage there's nothing can stop you from using older versions of Realm.

tgoyne:
Can you actually upgrade to Xcode 14, then? Dropping armv7 is a very big deal in terms of how many devices out there it drops (e.g. all those iPad 2s still out there), but there's a lot fewer devices running iOS 11 than 10.

Bumping to 11 and then waiting until next year for 12 wouldn't really be a big problem since we can't directly take advantage of things which require 12 until we drop Xcode 13 support (which will happen next year in the leadup to WWDC).

alexmx:
@tgoyne we're planning to drop iOS 10.3, hopefully before we're forced to switch to Xcode 14. We would really appreciate if support for iOS 11 can be kept for at least one more year. This will give us plenty of time and notice to our customers to update their OSes / change devices.

Configuration/Base.xcconfig
```diff
@@ -70,6 +78,6 @@ EXCLUDED_ARCHS[sdk=watchsimulator*] = $(REALM_ARM_ARCHS);
 EXCLUDED_ARCHS[sdk=iphonesimulator*] = $(REALM_ARM_ARCHS);
 EXCLUDED_ARCHS[sdk=appletvsimulator*] = $(REALM_ARM_ARCHS);
 
-SWIFT_VERSION = 5.0;
+SWIFT_VERSION = 5.5;
```
jsflax, line 81: Why 5.5 instead of 5.7?

Realm.xcodeproj/project.pbxproj
```diff
@@ -2567,6 +2567,7 @@
 /* Begin PBXShellScriptBuildPhase section */
 		3F1DDB5825155E33007A9630 /* Carthage post-build workaround */ = {
 			isa = PBXShellScriptBuildPhase;
+			alwaysOutOfDate = 1;
```
jsflax, line 2570: What is this?

Realm/Tests/SwiftUITestHostUITests/SwiftUITestHostUITests.swift
```diff
@@ -150,11 +178,11 @@ class SwiftUITests: XCTestCase {
         }
 
         let searchBar = app.textFields["searchField"]
-        let table = app.tables.firstMatch
+        let table = tables.firstMatch
 
         searchBar.tap()
 
-        searchBar.typeText("reminder list 1")
+        searchBar.typeText("reminder list 1\n") // \n to dismiss keyboard
```
jsflax, line 185: Clever.

Realm.xcodeproj/project.pbxproj
tgoyne: Xcode 14 now warns when you have a run script task which is triggered based on dependency analysis but doesn't actually declare any dependencies. This tells Xcode to just run it after every build, which is what we need for this.

.jenkins.yml
```diff
@@ -52,6 +53,9 @@ exclude:
   - xcode_version: 13.3.1
     target: docs
 
+  - xcode_version: 13.4.1
```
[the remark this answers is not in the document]
tgoyne: This is in the excludes section, so it's running in neither 13.3.1 or 13.4.1. I recommend just looking at the ruby script which generates this file as the output yaml file is hard to scan.

jsflax (approved):

Realm/Tests/SwiftUITestHostUITests/SwiftUITestHostUITests.swift
```diff
@@ -89,9 +97,14 @@ class SwiftUITests: XCTestCase {
         XCTAssertTrue(app.navigationBars.staticTexts["My Reminder"].waitForExistence(timeout: 1.0))
         let myReminder = realm.objects(ReminderList.self).first!.reminders.first!
         XCTAssertEqual(myReminder.priority, .low)
-        app.buttons["picker"].tap()
-        app.tables.switches["medium"].tap()
-        XCTAssertEqual(myReminder.priority, .medium)
+        if #available(iOS 16, *) {
```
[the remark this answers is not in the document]
tgoyne: I couldn't find anything which would match the cell, and the recorder just broke when I tried it.

dianaafanador3 (approved):

bdkjones:
If it's possible to get this PR released quickly, that would be very helpful. Eliminating "clever" things is always a good idea!

gongzhang:
Hi, @tgoyne,

Is watchOS not currently supported? Xcode 14 seems to report an error when building against the watchOS simulator. (Install via SPM)

Issue #7840, opened 2022-06-14
Trying to copy seed when the database exists


Truba:
### How frequently does the bug occur?

All the time

### Description

When the database exists and we are also providing the `seedFilePath` the app crashes trying to write seed over the main database file.

This was working fine in the `2.25.2` but is failing after the upgrade to `2.28.1`
It seems that [this](https://github.com/realm/realm-swift/commit/c4b36c55ebf5e7f37746f80de50b9b4389d7e548#diff-a98a6c3989e1668cf28dbcb829199452024a880dced23fc11060ba13b113989d) change in `Realm/RLMRealm.mm` which caused this behavior.
![Screenshot 2022-06-14 at 12 06 03](https://user-images.githubusercontent.com/5460328/173552286-8a54b60e-0c0b-4b3c-ad34-66b731bec461.png)

My current workaround is to check if there is a database file on the disc, and send `nil` instead of real URL for `seedFilePath`.

### Stacktrace & log output

```shell
Error Domain=NSCocoaErrorDomain Code=516 "“seed.realm” couldn’t be copied to “Documents” because an item with the same name already exists." UserInfo={NSSourceFilePathErrorKey=/Users/truba/Library/Developer/CoreSimulator/Devices/6A6FC623-0CCD-4DC2-9EFA-1BC91DE7C851/data/Containers/Bundle/Application/573FACFA-0143-40BB-8D77-447204E4D311/RealmExample.app/seed.realm, NSUserStringVariant=(
    Copy
), NSDestinationFilePath=/Users/truba/Library/Developer/CoreSimulator/Devices/6A6FC623-0CCD-4DC2-9EFA-1BC91DE7C851/data/Containers/Data/Application/A2372DE6-3227-4404-9EAC-A090C8C13F3C/Documents/default.realm, NSFilePath=/Users/truba/Library/Developer/CoreSimulator/Devices/6A6FC623-0CCD-4DC2-9EFA-1BC91DE7C851/data/Containers/Bundle/Application/573FACFA-0143-40BB-8D77-447204E4D311/RealmExample.app/seed.realm, NSUnderlyingError=0x6000023f7570 {Error Domain=NSPOSIXErrorDomain Code=17 "File exists"}}
```

### Can you reproduce the bug?

Yes, always

### Reproduction Steps

Code to reproduce:
```swift
static let seedURL: URL = Bundle.main.url(forResource: "seed", withExtension: "realm")!
    
    static func setup() throws {
[...]

PR #7841, opened 2022-06-14
Possible fix for #7840

Truba:


tgoyne:
Checking if the file exists before trying to copy is an example of a [TOCTOU race](https://en.wikipedia.org/wiki/Time-of-check_to_time-of-use), so the correct thing is to try to copy and ignore the "file already exists" error. The change in v10.26.0 just kinda forgot the second half of that, so #7843 adds that along with the missing test.

PR #7843, opened 2022-06-14
Fix opening existing Realms with seedFilePath set

### LINKED ISSUE #7840: Trying to copy seed when the database exists

tgoyne:
Fixes #7840.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,8 +4,9 @@ x.y.z Release notes (yyyy-MM-dd)
 * None.
 
 ### Fixed
-* <How to hit and notice issue? what was the impact?> ([#????](https://github.com/realm/realm-swift/issues/????), since v?.?.?)
-* None.
+* Using `seedFilePath` threw an exception if the Realm file being opened
+  already existed ([#7840](https://github.com/realm/realm-swift/issues/7840),
+  since v10.26.0).
 
 <!-- ### Breaking Changes - ONLY INCLUDE FOR NEW MAJOR version -->
 
--- a/Realm/RLMRealm.mm
+++ b/Realm/RLMRealm.mm
@@ -495,7 +495,7 @@ static RLMRealm *getCachedRealm(RLMRealmConfiguration *configuration, void *cach
                                                                       toURL:configuration.fileURL
                                                                       error:&copyError];
             });
-            if (!didCopySeed && copyError != nil) {
+            if (!didCopySeed && copyError != nil && copyError.code != NSFileWriteFileExistsError) {
                 RLMSetErrorOrThrow(copyError, error);
                 return nil;
             }
--- a/RealmSwift/Tests/RealmTests.swift
+++ b/RealmSwift/Tests/RealmTests.swift
@@ -964,33 +964,46 @@ class RealmTests: TestCase {
         }
     }
 
-    func testSeedFilePath() {
-        do {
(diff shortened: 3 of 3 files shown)


ericjordanmossman (commented):
So I have this straight: when a seed file path is used, but a realm already exists at the destination path, the existing realm is used. No attempted overwrite, no attempted merge, no error thrown , and no exception (now).

I can add a test for this, if you want.

ericjordanmossman (approved):

tgoyne:
Yes, the seed file should be copied over if the Realm file doesn't exist on open, and ignored entirely if it does. This should now actually be tested.

PR #7851, opened 2022-06-16
Only call the `initialSubscriptions` callback when it should be called

tgoyne:
It was called on every Realm open because the `initializationFunction` which attempted to check if this was the first open was never actually used so `isFirstOpen` was always `false`, and then the code which checked that value inverted the condition and did the logic for first open if `isFirstOpen` was `false` rather than `true`.

The test which attempted to test this did not actually open the Realm a second time.

### DIFF
--- a/Realm/ObjectServerTests/RLMFlexibleSyncServerTests.mm
+++ b/Realm/ObjectServerTests/RLMFlexibleSyncServerTests.mm
@@ -229,7 +229,7 @@
         RLMAssertThrowsWithReason([subs addSubscriptionWithClassName:Person.className
                                                     subscriptionName:@"person_age"
                                                                where:@"age > 20"],
-                                  @"Cannot duplicate a subscription. If you meant to update the subscription please use the `update` method.");
+                                  @"A subscription named 'person_age' already exists. If you meant to update the existing subscription please use the `update` method.");
     }];
 
     XCTAssertEqual(subs.version, 1UL);
@@ -918,10 +918,7 @@
 }
 
 - (void)testFlexibleSyncInitialSubscription {
-    RLMUser *user = [self logInUserForCredentials:[self basicCredentialsWithName:NSStringFromSelector(_cmd)
-                                                                        register:YES
-                                                                             app:self.flexibleSyncApp]
-                                              app:self.flexibleSyncApp];
+    RLMUser *user = [self flexibleSyncUser:_cmd];
     RLMRealmConfiguration *config = [user flexibleSyncConfigurationWithInitialSubscriptions:^(RLMSyncSubscriptionSet *subscriptions) {
         [subscriptions addSubscriptionWithClassName:Person.className
                                    subscriptionName:@"person_age"
@@ -940,10 +937,7 @@
         return;
     }
 
-    RLMUser *user = [self logInUserForCredentials:[self basicCredentialsWithName:NSStringFromSelector(_cmd)
-                                                                        register:YES
-                                                                             app:self.flexibleSyncApp]
-                                              app:self.flexibleSyncApp];
+    RLMUser *user = [self flexibleSyncUser:_cmd];
     RLMRealmConfiguration *config = [user flexibleSyncConfigurationWithInitialSubscriptions:^(RLMSyncSubscriptionSet *subscriptions) {
         [subscriptions addSubscriptionWithClassName:Person.className
                                    subscriptionName:@"person_age"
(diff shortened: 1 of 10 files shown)


Realm/RLMRealm.mm
```diff
@@ -456,52 +448,76 @@ + (instancetype)realmWithConfiguration:(RLMRealmConfiguration *)configuration er
...
     // this is just an artifact of NSMapTable's strange API.
-    void *cacheKey = reinterpret_cast<void *>(std::numeric_limits<uintptr_t>::max());
-    if (queue) {
-        if (queue != dispatch_get_main_queue()) {
-            cacheKey = (__bridge void *)queue;
-        }
+    return reinterpret_cast<void *>(std::numeric_limits<uintptr_t>::max());
+}
+
+static bool copySeedFile(RLMRealmConfiguration *configuration, NSError **error) {
```
dianaafanador3, line 468: Is this supposed to be on this PR. It doesn't make sense in the initialSubscriptions` context

Realm/RLMRealm.mm
```diff
@@ -581,13 +584,15 @@ + (instancetype)realmWithConfiguration:(RLMRealmConfiguration *)configuration
...
+        if (!configuration.rerunOnOpen && configuration.initialSubscriptions) {
+            initializationFunction = [&isFirstOpen](SharedRealm) {
+                isFirstOpen = true;
+            };
+        }
 
         try {
             realm->_realm->update_schema(schema.objectStoreCopy, config.schema_version,
-                                         std::move(migrationFunction));
+                                         std::move(migrationFunction), std::move(initializationFunction));
```
dianaafanador3, line 595: why do we need to add this, if the initialisationFunction is used?

Realm/ObjectServerTests/RLMFlexibleSyncServerTests.mm
```diff
@@ -1013,19 +1008,28 @@ - (void)testFlexibleSyncInitialSubscriptionRerunOnOpen {
...
         XCTAssertEqual(realm.subscriptions.count, 2UL);
         CHECK_COUNT(16, Person, realm);
         [ex2 fulfill];
     }];
     [self waitForExpectationsWithTimeout:10.0 handler:nil];
+    XCTAssertEqual(openCount, 2);
+
+    [self dispatchAsyncAndWait:^{
+        [RLMRealm realmWithConfiguration:config error:nil];
+        // Should not have called initial subscriptions despite rerunOnOpen being
```
dianaafanador3, line 1029: I don't understand this message, in this case initialSubscription should be called? which is correct when compare openCount == 2 on the next line, or am I missing something?

tgoyne: The Realm is already open on the main thread, and initialSubscriptions is only supposed to be called when opening the Realm from scratch.

Realm/ObjectServerTests/SwiftFlexibleSyncServerTests.swift
```diff
@@ -1242,18 +1242,20 @@ extension SwiftFlexibleSyncServerTests {
...
-        XCTAssertEqual(realm.subscriptions.count, 2)
-        XCTAssertEqual(realm2.subscriptions.count, 2)
-        checkCount(expected: 19, realm2, SwiftTypesSyncObject.self)
+        let c = config
+        _ = try await Task { @MainActor in
+            let realm = try await Realm(configuration: c, downloadBeforeOpen: .always)
+            XCTAssertNotNil(realm)
+            XCTAssertEqual(realm.subscriptions.count, 1)
+            checkCount(expected: 9, realm, SwiftTypesSyncObject.self)
+        }.value
```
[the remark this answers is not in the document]
tgoyne: Obtaining the return value from the Task (which is Void in this case) is how you wait for the Task to complete.

Realm/RLMRealm.mm
tgoyne: I don't understand the question. This is the only place that the initializationFunction is used.

dianaafanador3 (approved):

Realm/ObjectServerTests/SwiftFlexibleSyncServerTests.swift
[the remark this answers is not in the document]
theoks: @tgoyne I don't think that you need to assign the Void result of the Task to an underscore variable. Just the `.value` and `await` keyword make sure that the Task is awaited to return from the suspension point.

Issue #7831, opened 2022-06-08
RLMProgressNotificationToken released without unregistering a notification. You must hold on to the RLMProgressNotificationToken and call -[RLMProgressNotificationToken invalidate] when you no longer wish to receive progress update notifications.


anton-plebanovich:
### Description

The error is falsely produced when a session is released before `invalidate` for a token is called causing an inability to reset the `_token` value and so producing a misleading error message. On the screenshot is the proposed solution but I don't know the Realm source base good enough so it's up to you.
<img width="714" alt="image" src="https://user-images.githubusercontent.com/10807812/172525067-eae60bdf-6a7b-41da-b697-f17630ac3cf1.png">

### Version

10.27.0

anton-plebanovich:
> Do you know what's causing the session to be released that makes you think the error falsely produced? That may be the issue. I think never having a progress_notifier exist whose token was released is intentional.

Well, I did what the warning says, I hold the token and then called `invalidate` when I no longer need it but still received the warning on the `dealloc` so I assumed it is missleading because I spent some time to investigate my logic while it was working fine.

This is happening on the `App` switch during ongoing load and overall is minor since there are no other issues.

ericjordanmossman:
What is the "`App` switch"? How are you storing the token? 
It sounds like you're able to reproduce reliably -- could you share a .zip so the team can check it out?

That'd be super helpful.

anton-plebanovich:
Please check the reproduction example here https://github.com/anton-plebanovich/SingleView/tree/realm/notification-token-bug

PR #7864, opened 2022-06-28
Do not show an error message in cases where  session is already deallocated

dianaafanador3:
Do not show an error message in cases where `RLMProgressNotificationToken` session is already deallocated ([#7831](https://github.com/realm/realm-swift/issues/7831), since v2.3.0).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -340,8 +340,6 @@ The prebuilt binary for Carthage is now build with Xcode 14.0.1.
   removing the destination link object.
   ([Core #5574](https://github.com/realm/realm-core/pull/5573), since the introduction of AnyRealmValue in v10.8.0)
 
-<!-- ### Breaking Changes - ONLY INCLUDE FOR NEW MAJOR version -->
-
 ### Compatibility
 
 * Realm Studio: 12.0.0 or later.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,8 @@ The prebuilt binary for Carthage is now build with Xcode 14.0.1.
 
 ### Fixed
 * Setting a `List` property with `Results` no longer throws an unrecognized selector exception (since 10.8.0-beta.2)
+* Do not show an error message in cases where `RLMProgressNotificationToken` session is already deallocated ([#7831](https://github.com/realm/realm-swift/issues/7831),
+  since v2.3.0).
 
 <!-- ### Breaking Changes - ONLY INCLUDE FOR NEW MAJOR version -->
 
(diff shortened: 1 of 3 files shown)


tgoyne:
I wonder if the token should actually store a strong pointer to the session rather than a weak one? All of our other tokens keep the thing they're tracking notifications for alive, and it's sort of weird that this one doesn't.

dianaafanador3:
@tgoyne updated the PR to have a reference so the SyncSession

CHANGELOG.md
```diff
@@ -338,6 +340,8 @@ The prebuilt binary for Carthage is now build with Xcode 14.0.1.
   removing the destination link object.
   ([Core #5574](https://github.com/realm/realm-core/pull/5573), since the introduction of AnyRealmValue in v10.8.0)
 
+<!-- ### Breaking Changes - ONLY INCLUDE FOR NEW MAJOR version -->
```
tgoyne, line 343: Bad merge.

Realm/RLMSyncSession.mm
```diff
@@ -52,20 +52,9 @@ - (void)suppressNextNotification {
...
-}
-
-- (void)dealloc {
-    if (_token != 0) {
-        NSLog(@"RLMProgressNotificationToken released without unregistering a notification. "
-              @"You must hold on to the RLMProgressNotificationToken and call "
-              @"-[RLMProgressNotificationToken invalidate] when you no longer wish to receive "
-              @"progress update notifications.");
-    }
+    _session->unregister_progress_notifier(_token);
```
tgoyne, line 55: This needs to check if `_session` is non-null so that calling `invalidate` twice isn't a use-after-free.

CHANGELOG.md
```diff
@@ -8,6 +8,8 @@ The prebuilt binary for Carthage is now build with Xcode 14.0.1.
 
 ### Fixed
 * Setting a `List` property with `Results` no longer throws an unrecognized selector exception (since 10.8.0-beta.2)
+* Do not show an error message in cases where `RLMProgressNotificationToken` session is already deallocated ([#7831](https://github.com/realm/realm-swift/issues/7831),
```
tgoyne, line 11: It's worth mentioning too that the token now keeps the sync session alive.

CHANGELOG.md
tgoyne: This still needs to be removed.

dianaafanador3: Sorry, didn't see it, done

tgoyne (approved):

PR #7871, opened 2022-07-04
Style: "Fix XCTFail Message"

wi-seong-cheol:
The message in XCTFail was corrected because the parenthesis was incorrect. The modified message is the same as the message used by other existing XCTFail.


wi-seong-cheol:
@cla-bot check

dianaafanador3 (approved):

PR #7873, opened 2022-07-05
Flx Sync - Add InitialSubscriptions and rerunOnOpen to the copy method

dianaafanador3:
Add InitialSubscriptions and rerunOnOpen to the copy method


tgoyne:
Needs a changelog entry.

leemaguire (approved):

PR #7880, opened 2022-07-12
Asymmetric Sync

dianaafanador3:
Add Support for asymmetric sync. When an object inherit from `AsymmetricObject` is immediately
   marked as asymmetric, upon creation this objects is sync'd unidirectionally and cannot be query
   locally.

 ```swift
     class PersonObject: AsymmetricObject {
        @Persisted(primaryKey: true) var _id: ObjectId = ObjectId.generate()
        @Persisted var name: String
        @Persisted var age: Int
     }

     try realm.write {
        let person = PersonObject(name: "Dylan", age: 20, isFavorite: true)
        realm.create(person) // This will create the object and sync'd unidirectionally.
     }
 ```


Realm/RLMObjectStore.mm
```diff
@@ -138,6 +138,18 @@ void RLMAddObjectToRealm(__unsafe_unretained RLMObjectBase *const object,
     return object;
 }
 
+void RLMCreateAsymmetricObjectToRealm(__unsafe_unretained RLMObjectBase *const object,
+                                      __unsafe_unretained RLMRealm *const realm) {
+    RLMVerifyInWriteTransaction(realm);
+
+    CreatePolicy createPolicy = {.create = true, .copy = false, .diff = false, .update = false};
+    createPolicy.copy = true;
```
pavel-ship-it, line 146: Why not set during declaration?

Realm/ObjectServerTests/RLMAsymmetricSyncServerTests.mm
```diff
@@ -0,0 +1,163 @@
...
+    [super tearDown];
+}
+
+- (void)checkCountInMongo:(unsigned long)count {
+    RLMUser *user = [self logInUserForCredentials:[RLMCredentials anonymousCredentials]
+                                              app:self.flexibleSyncApp];
+    RLMMongoClient *client = [user mongoClientWithServiceName:@"mongodb1"];
+    RLMMongoDatabase *database = [client databaseWithName:@"test_data"];
+    RLMMongoCollection *collection = [database collectionWithName:@"PersonAsymmetric"];
+    XCTestExpectation *findExpectation4 = [self expectationWithDescription:@"should not find documents"];
```
pavel-ship-it, line 84: "Should find"?

RealmSwift/ObjectSchema.swift
```diff
@@ -48,6 +48,12 @@ import Realm.Private
     /// The object class the schema describes.
     public var objectClass: AnyClass { return rlmObjectSchema.objectClass }
 
+    /// Whether this object is embedded.
+    public var isEmbedded: Bool { return rlmObjectSchema.isEmbedded }
+
+    /// Whether this object is asymmetric.
+    public var isAymmetric: Bool { return rlmObjectSchema.isAsymmetric }
```
leemaguire, line 55:
```suggestion
    public var isAsymmetric: Bool { return rlmObjectSchema.isAsymmetric }
```

RealmSwift/Impl/ComplexTypes.swift
```diff
@@ -83,6 +83,33 @@ extension EmbeddedObject: SchemaDiscoverable, _PersistableInsideOptional, _Defau
     }
 }
 
+extension AsymmetricObject: SchemaDiscoverable, _PersistableInsideOptional, _DefaultConstructible {
+    public typealias PersistedType = EmbeddedObject
```
leemaguire, line 87:
```suggestion
    public typealias PersistedType = AsymmetricObject
```

RealmSwift/AsymmetricObject.swift
```diff
@@ -0,0 +1,125 @@
...
+// limitations under the License.
+//
+////////////////////////////////////////////////////////////////////////////
+
+import Foundation
+import Realm
+import Realm.Private
+
+/**
+ `AsymmetricObject` is a base class used to define asymmetric Realm model objects.
```
leemaguire, line 24:
```suggestion
 `AsymmetricObject` is a base class used to define asymmetric Realm objects.
```

Realm/RLMObjectStore.mm
```diff
@@ -138,6 +138,18 @@ void RLMAddObjectToRealm(__unsafe_unretained RLMObjectBase *const object,
     return object;
 }
 
+void RLMCreateAsymmetricObjectToRealm(__unsafe_unretained RLMObjectBase *const object,
```
leemaguire, line 141:
```suggestion
void RLMAddAsymmetricObjectToRealm(__unsafe_unretained RLMObjectBase *const object,
```

Realm/RLMAsymmetricObject.mm
```diff
@@ -0,0 +1,97 @@
...
+    return [object isKindOfClass:RLMObjectBase.class] && RLMObjectBaseAreEqual(self, object);
+}
+
++ (NSString *)className {
+    return [super className];
+}
+
+#pragma mark - Default values for schema definition
+
++ (NSString *)primaryKey {
```
leemaguire, line 66: Is there a need to redeclare these getters?

Realm/RLMAsymmetricObject.h
```diff
@@ -0,0 +1,150 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+////////////////////////////////////////////////////////////////////////////
+
+#import <Foundation/Foundation.h>
```
leemaguire, line 19: You can omit this import

Realm/RLMAsymmetricObject.h
```diff
@@ -0,0 +1,150 @@
...
+
+ Each property of type `RLMLinkingObjects` must have a key in the dictionary returned by this method consisting
+ of the property name. The corresponding value must be an instance of `RLMPropertyDescriptor` that describes the class
+ and property that the property is linked to.
+
+     return @{ @"owners": [RLMPropertyDescriptor descriptorWithClass:Owner.class propertyName:@"dogs"] };
+
+ @return     A dictionary mapping property names to `RLMPropertyDescriptor` instances.
+ */
++ (NSDictionary<NSString *, RLMPropertyDescriptor *> *)linkingObjectsProperties;
```
leemaguire, line 121: will Asymmetric sync even support this?

Realm/ObjectServerTests/RLMAsymmetricSyncServerTests.mm
```diff
@@ -0,0 +1,163 @@
...
+    [self waitForUploadsForRealm:realm];
+    [self checkCountInMongo:1];
+
+    [realm beginWriteTransaction];
+    [realm createObject:person];
+    [realm commitWriteTransaction];
+    [self waitForUploadsForRealm:realm];
+    [self checkCountInMongo:1];
+}
+@end
```
leemaguire, line 161: We should also test that Asymmetric objects with invalid schemas throw.

Realm/ObjectServerTests/RLMAsymmetricSyncServerTests.mm
```diff
@@ -0,0 +1,163 @@
...
+//
+////////////////////////////////////////////////////////////////////////////
+
+#import "RLMSyncTestCase.h"
+
+// We add this compilation flag, because we don't want to compile this code when we run
+// `xcrun swift test`. Adding an asymmetric table to a local realm will throw a fatal error
+// when opening the realm, excluding this from compilation will not add the object to
+// the schema. We are only running this tests on the ObjectServer target.
+#if REALM_SERVER_TEST_TARGET
```
tgoyne, line 25: We need to run all of the tests via SPM as that's the only way for us to test with asan and tsan. If the problem is that the type ends up in the default schema, the class should just override the class function to exclude it.

CHANGELOG.md
```diff
@@ -1,7 +1,22 @@
 x.y.z Release notes (yyyy-MM-dd)
 =============================================================
 ### Enhancements
-* None.
+* Add Support for asymmetric sync. When an object inherit from `AsymmetricObject` is immediately
+  marked as asymmetric, upon creation this objects is sync'd unidirectionally and cannot be query
+  locally.
```
tgoyne, line 6:
```suggestion
* Add support for asymmetric sync. When a class inherits from
  `AsymmetricObject`, objects created are synced unidirectionally to the server
  and cannot be queried or read locally.
```

All the "sync'd" all over the place should be "synced".

CHANGELOG.md
```diff
@@ -1,7 +1,22 @@
...
+  
+```swift
+    class PersonObject: AsymmetricObject {
+       @Persisted(primaryKey: true) var _id: ObjectId = ObjectId.generate()
```
tgoyne, line 10:
```suggestion
       @Persisted(primaryKey: true) var _id: ObjectId
```

CHANGELOG.md
```diff
@@ -1,7 +1,22 @@
...
+       @Persisted var name: String
+       @Persisted var age: Int
+    }
+    
+    try realm.write {
+       let person = PersonObject(name: "Dylan", age: 20, isFavorite: true)
+       realm.create(person) // This will create the object and sync'd unidirectionally.
```
tgoyne, line 17:
```suggestion
        // This will create the object on the server but not locally
       realm.create(person) 
```

Realm/ObjectServerTests/RLMAsymmetricSyncServerTests.mm
```diff
@@ -0,0 +1,163 @@
...
+}
+
+- (void)testCreateAsymmetricObjectWithPBSConfigurationShouldNotSync {
+    RLMUser *user = [self userForTest:_cmd];
+    RLMRealmConfiguration *configuration = [user configurationWithPartitionValue:NSStringFromSelector(_cmd)];
+    configuration.objectClasses = @[PersonAsymmetric.self];
+    RLMRealm *realm = [RLMRealm realmWithConfiguration:configuration error:nil];
+    [realm beginWriteTransaction];
+    PersonAsymmetric *person = [[PersonAsymmetric alloc] initWithPrimaryKey:[RLMObjectId objectId]
+                                                                  firstName:[NSString stringWithFormat:@"firstname"]
```
tgoyne, line 109: Why the stringWithFormats?

Realm/ObjectServerTests/RLMAsymmetricSyncServerTests.mm
```diff
@@ -0,0 +1,163 @@
...
+
+- (void)testOpenLocalRealmWithAsymmetricObjectError {
+    RLMRealmConfiguration *configuration = [RLMRealmConfiguration defaultConfiguration];
+    NSError *error;
+    RLMRealm *realm = [RLMRealm realmWithConfiguration:configuration error:&error];
+    XCTAssertNil(realm);
+    XCTAssertNotNil(error);
+}
+
+- (void)testCreateAsymmetricObjectWithPBSConfigurationShouldNotSync {
```
tgoyne, line 102: Shouldn't this be an error instead of just silently not working?

Realm/ObjectServerTests/SwiftAsymmetricSyncServerTests.swift
```diff
@@ -0,0 +1,233 @@
...
+import SwiftUI
+#endif
+
+// We add this compilation flag, because we don't want to compile this code when we run
+// `xcrun swift test`. Adding an asymmetric table to a local realm will throw a fatal error
+// when opening the realm, excluding this from compilation will not add the object to
+// the schema. We are only running this tests on the ObjectServer target.
+#if REALM_SERVER_TEST_TARGET
+public class SwiftObjectAsymmetric: AsymmetricObject {
+    @Persisted(primaryKey: true) var _id: ObjectId = ObjectId.generate()
```
tgoyne, line 38:
```suggestion
    @Persisted(primaryKey: true) var _id: ObjectId
```

Realm/RLMAsymmetricObject.h
```diff
@@ -0,0 +1,150 @@
...
+
+NS_ASSUME_NONNULL_BEGIN
+
+@class RLMObjectSchema, RLMPropertyDescriptor;
+/**
+ `RLMAsymmetricObject` is a base class used to define asymmetric Realm model objects.
+
+ Asymmetric objects can only be created using the `create(_ object:)`
+ function, and cannot be added, removed or queried.
+ When created asymmetric objects will sync'd unidirectionally to the  MongoDB
```
tgoyne, line 31:
```suggestion
 When created asymmetric objects will synced unidirectionally to the MongoDB
```

Realm/RLMAsymmetricObject.h
```diff
@@ -0,0 +1,150 @@
...
+
+#import <Realm/RLMObjectBase.h>
+
+NS_ASSUME_NONNULL_BEGIN
+
+@class RLMObjectSchema, RLMPropertyDescriptor;
+/**
+ `RLMAsymmetricObject` is a base class used to define asymmetric Realm model objects.
+
+ Asymmetric objects can only be created using the `create(_ object:)`
```
tgoyne, line 29: These docs are for the obj-c type and need to use obj-c syntax and names.

Realm/RLMAsymmetricObject.h
```diff
@@ -0,0 +1,150 @@
...
+ For frozen objects and object types with a primary key, `isEqual:` is
+ overridden to use the same logic as this method (along with a corresponding
+ implementation for `hash`). Non-frozen objects without primary keys use
+ pointer identity for `isEqual:` and `hash`.
+
+ @param object  The object to compare the receiver to.
+
+ @return    Whether the object represents the same object as the receiver.
+ */
+- (BOOL)isEqualToObject:(RLMAsymmetricObject *)object;
```
tgoyne, line 138: This is not applicable to asymmetric objects.

Realm/RLMObjectSchema.mm
```diff
@@ -138,9 +138,17 @@ + (instancetype)schemaForObjectClass:(Class)objectClass {
     schema.accessorClass = objectClass;
     schema.unmanagedClass = objectClass;
     schema.isSwiftClass = isSwift;
-    schema.isEmbedded = [(id)objectClass isEmbedded];
     schema.hasCustomEventSerialization = [objectClass conformsToProtocol:@protocol(RLMCustomEventRepresentable)];
 
+    bool isEmbedded = [(id)objectClass isEmbedded];
+    bool isAsymmetric = [(id)objectClass isAsymmetric];
+    if (isEmbedded && isAsymmetric) {
+        @throw RLMException(@"Object '%@' is declared both as `Embedded` and `Asymmetric`, which is not supported", className);
```
tgoyne, line 146: This should be an assertion, as it's not something that the user can do via API misuse.

Realm/RLMObjectStore.mm
tgoyne: Create is correct here, but it should be `InRealm` not `ToRealm`.

Realm/RLMRealm.h
```diff
@@ -828,6 +828,20 @@ NS_REFINED_FOR_SWIFT;
...
+ can be created using this method.
+
+ Objects created using this method will not be added to the Realm.
+
+ @warning This method may only be called during a write transaction.
+
+ @param object The asymmetric object to create.
+
+ */
+- (void)createObject:(RLMAsymmetricObject *)object;
```
tgoyne, line 843: This function is weird and different from the normal API for no obvious reason.

Realm/RLMAsymmetricObject.mm
dianaafanador3: I guess it is not necessary, because there are not specific default values for this, I'll remove it.

dianaafanador3: this is needed for schema discovery purposes

Realm/ObjectServerTests/RLMAsymmetricSyncServerTests.mm
dianaafanador3: After discussing this with core, the validation is gonna be done on their side.  I'm going to comment both tests and add a FIXME so when the core release is ready we can enable this tests.

dianaafanador3:
@tgoyne seems like the CI failures doesn't have anything to do with the PR, I don't know the reason the test are not able to execute.

Realm/RLMRealm.h
dianaafanador3: I changed to instead of using the realm to create the object, you can do it like this `[PersonAsymmetric createInRealm:realm withObject:person]` like the others APIs

Realm/ObjectServerTests/RLMAsymmetricSyncServerTests.mm
dianaafanador3: Changed this completely so instead of using a build configuration flag,  I'm creating an app exclusively for asymmetric testing which is not using the sharedSchema to create the app.

Realm/ObjectServerTests/RLMAsymmetricSyncServerTests.mm
dianaafanador3: Added a few tests for links within an asymmetric object and a top level object linking an asymmetric object.   The validation is done by core and the errors are thrown when the realm is opened.

Realm/RLMAsymmetricObject.h
dianaafanador3: This is not supported so I removed it

Realm/ObjectServerTests/RLMAsymmetricSyncServerTests.mm
```diff
@@ -0,0 +1,322 @@
...
+////////////////////////////////////////////////////////////////////////////
+
+#import "RLMSyncTestCase.h"
+#import "RLMObject_Private.h"
+#import "RLMApp_Private.h"
+#import "RLMObjectSchema_Private.h"
+
+// These are defined in Swift. Importing the auto-generated header doesn't work
+// when building with SPM, so just redeclare the bits we need.
+@interface RealmServer : NSObject
```
leemaguire, line 26: `RLMFlexibleSyncTests` also defines an interface for `RealmServer`. I think we should create a helper header file that defines this interface so that both test suites can use it, and removes the need to repeat the interface in both places.

Realm/ObjectServerTests/SwiftAsymmetricSyncServerTests.swift
```diff
@@ -0,0 +1,249 @@
...
+        do {
+            _ = try Realm(configuration: configuration)
+            XCTFail("Opening a local Realm with an `Asymmetric` table should fail")
+        } catch {
+            XCTAssertNotNil(error)
+        }
+    }
+
+    // FIXME: Enable this test when this is implemented on core. Core should validate if the schema includes an asymmetric table for a PBS configuration and throw an error.
+//    func testOpenPBSConfigurationRealmWithAsymmetricObjectError() throws {
```
leemaguire, line 154: To make sure the test still compiles and not make it run, just prefix `fixme` to the function name.

Realm/RLMAsymmetricObject.h
```diff
@@ -0,0 +1,98 @@
...
+
+ Incoming links from and asymmetric table are not allowed, meaning embedding
+ an asymmetric object within an `RLMObject` will throw an error.
+
+ The property types supported on `RLMAsymmetricObject` are the same as for `RLMObject`,
+ except for that asymmetric objects can only link to embedded objects, so `RLMObject`
+ and `RLMArray<RLMObject>` properties are not supported (`RLMEmbeddedObject` and
+ `RLMArray<RLEmbeddedObject>` *are*).
+ */
+
```
leemaguire, line 40: Remove newline

Realm/RLMAsymmetricObject.h
```diff
@@ -0,0 +1,98 @@
...
+//
+////////////////////////////////////////////////////////////////////////////
+
+#import <Realm/RLMObjectBase.h>
+
+NS_ASSUME_NONNULL_BEGIN
+
+@class RLMObjectSchema, RLMPropertyDescriptor;
+/**
+ `RLMAsymmetricObject`  is a base class used to define asymmetric Realm objects.
```
leemaguire, line 25:
```suggestion
 `RLMAsymmetricObject` is a base class used to define asymmetric Realm objects.
```

Realm/RLMAsymmetricObject.h
```diff
@@ -0,0 +1,98 @@
...
+
+ Asymmetric objects can only be created using the `createObject:`
+ function, and cannot be added, removed or queried.
+ When created asymmetric objects will synced unidirectionally to the  MongoDB
```
leemaguire, line 29:
```suggestion
 When created, asymmetric objects will be synced unidirectionally to the MongoDB
```

Realm/RLMAsymmetricObject.mm
```diff
@@ -0,0 +1,100 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+////////////////////////////////////////////////////////////////////////////
+
+#import "RLMAsymmetricObject.h"
+
+#import "RLMObject_Private.hpp"
+#import "RLMSchema_Private.h"
+#import "RLMObjectStore.h"
```
leemaguire, line 23:
```suggestion
#import "RLMObject_Private.hpp"
#import "RLMObjectStore.h"
#import "RLMSchema_Private.h"
```

RealmSwift/AsymmetricObject.swift
```diff
@@ -0,0 +1,105 @@
...
+     @Persisted(primaryKey: true) var _id: ObjectId = ObjectId.generate()
+     @Persisted var name: String
+     @Persisted var age: Int
+ }
+ ```
+ */
+public typealias AsymmetricObject = RealmSwiftAsymmetricObject
+extension AsymmetricObject: _RealmCollectionValueInsideOptional {
+    /// :nodoc:
+    public class override final func isAsymmetric() -> Bool {
```
leemaguire, line 50: I don't see the need to override this getter?

RealmSwift/Realm.swift
```diff
@@ -1101,6 +1101,25 @@ extension Realm {
     }
 }
 
+// MARK: Asymmetric Sync
+
+extension Realm {
+    /**
+     Creates an Asymmetric object, which will be synced unidirectionally and
+     cannot be query locally, only objects which inherits from `AsymmetricObject`
```
leemaguire, line 1109:
```suggestion
     cannot be query locally, only objects which inherit from `AsymmetricObject`
```

leemaguire (commented):
Implementation is looking good, just minor comments

RealmSwift/AsymmetricObject.swift
dianaafanador3: we need to do this for schema discovery, only asymmetric objects (object that inherit from `AsymmetricObject`), will have this set as true, and when we are getting the object schema for the specific class https://github.com/realm/realm-swift/blob/e6b763cf9bcf2359c9c423336e08aee277f04635/Realm/RLMObjectSchema.mm#L144, only the objects which have this true will be set as asymmetric, we set this flag for schema validation with the server and core's using this to allow/disallow some operation with asymmetric objects.

Realm/ObjectServerTests/RLMAsymmetricSyncServerTests.mm
dianaafanador3: I just centralised everything in `RLMSyncTestCases.h`https://github.com/realm/realm-swift/blob/0b3b8720e0de3ab2561d585e126f6549c006e079/Realm/ObjectServerTests/RLMSyncTestCase.h#L28  file which is added to all sync test files.

RealmSwift/Realm.swift
```diff
@@ -1101,6 +1101,25 @@ extension Realm {
     }
 }
 
+// MARK: Asymmetric Sync
+
+extension Realm {
+    /**
+     Creates an Asymmetric object, which will be synced unidirectionally and
+     cannot be query locally, only objects which inherit from `AsymmetricObject`
```
leemaguire, line 1109:
```suggestion
     cannot be queried locally, only objects which inherit from `AsymmetricObject`
```

leemaguire (approved):

Realm/ObjectServerTests/RLMAsymmetricSyncServerTests.mm
```diff
@@ -0,0 +1,316 @@
...
+
+- (void)testUnsupportedAsymmetricLinkAsymmetricThrowsError {
+    RLMUser *user = [self userForSelector:_cmd];
+    RLMRealmConfiguration *configuration = [user flexibleSyncConfiguration];
+    configuration.objectClasses = @[UnsupportedLinkAsymmetric.self, PersonAsymmetric.self];
+    NSError *error;
+    RLMRealm *realm = [RLMRealm realmWithConfiguration:configuration error:&error];
+    XCTAssertNil(realm);
+    // This error comes from core, we are not even adding this objects to the server schema.
+    XCTAssertNotNil(error);
```
tgoyne, line 227: We should verify that the error is something sensible and not just that there is an error in each of these tests.

Realm/ObjectServerTests/RLMAsymmetricSyncServerTests.mm
```diff
@@ -0,0 +1,316 @@
...
+    RLMMongoCollection *collection = [database collectionWithName:@"PersonAsymmetric"];
+    XCTestExpectation *findExpectation4 = [self expectationWithDescription:@"Should find documents"];
+    [collection findWhere:@{}
+               completion:^(NSArray<NSDictionary *> *documents, NSError *error) {
+        XCTAssertEqual(documents.count, count);
+        XCTAssertNil(error);
+        [findExpectation4 fulfill];
+    }];
+    [self waitForExpectationsWithTimeout:60.0 handler:nil];
+}
```
tgoyne, line 209: This is going to fail sometimes due to timing. Asymmetric objects use a fast-path on the server that bypasses the normal translator, but it still doesn't guarantee that the objects will be present in the mongo collection by the time the upload completes. We have to instead poll for completion and keep retrying until we get the expected count (with a maximum time, of course).

It also should use `countWhere:` rather than `findWhere:` to avoid pointlessly transferring and parsing the documents when we only need the count.

Realm/ObjectServerTests/SwiftAsymmetricSyncServerTests.swift
```diff
@@ -0,0 +1,249 @@
...
+        XCTAssertTrue(realm.schema.objectSchema[0].isAsymmetric)
+    }
+
+    func testOpenLocalRealmWithAsymmetricObjectError() throws {
+        let configuration = Realm.Configuration(objectTypes: [SwiftObjectAsymmetric.self])
+        do {
+            _ = try Realm(configuration: configuration)
+            XCTFail("Opening a local Realm with an `Asymmetric` table should fail")
+        } catch {
+            XCTAssertNotNil(error)
```
tgoyne, line 149: As with the obj-c versions these should be testing that the error is the correct error.

Realm/ObjectServerTests/SwiftObjectServerTests.swift
```diff
@@ -2424,12 +2409,15 @@ class CombineObjectServerTests: SwiftSyncTestCase {
 
     func testAsyncOpenStandaloneCombine() throws {
         try autoreleasepool {
-            let realm = try Realm()
+            let configuration = Realm.Configuration(objectTypes: [SwiftPerson.self])
```
tgoyne, line 2412: The configuration doesn't need to be defined twice in these two tests.

RealmSwift/AsymmetricObject.swift
```diff
@@ -0,0 +1,100 @@
...
+ ```swift
+ class Person: AsymmetricObject {
+     @Persisted(primaryKey: true) var _id: ObjectId = ObjectId.generate()
+     @Persisted var name: String
+     @Persisted var age: Int
+ }
+ ```
+ */
+public typealias AsymmetricObject = RealmSwiftAsymmetricObject
+extension AsymmetricObject: _RealmCollectionValueInsideOptional {
```
tgoyne, line 48: I don't think `_RealmCollectionValueInsideOptional` is correct? I may be missing something, but I don't think that AsymmetricObject should conform to `_Persistable` or `RealmCollectionValue`, as you can't link to an asymmetric object or have a collection of them.

RealmSwift/Impl/ComplexTypes.swift
```diff
@@ -83,6 +83,33 @@ extension EmbeddedObject: SchemaDiscoverable, _PersistableInsideOptional, _Defau
...
 
+extension AsymmetricObject: SchemaDiscoverable, _PersistableInsideOptional, _DefaultConstructible {
+    public typealias PersistedType = AsymmetricObject
+    public static var _rlmType: PropertyType { .object }
+    public static func _rlmPopulateProperty(_ prop: RLMProperty) {
+        Object._rlmPopulateProperty(prop)
+        prop.objectClassName = className()
+    }
+
+    public static func _rlmGetProperty(_ obj: ObjectBase, _ key: UInt16) -> Self {
```
tgoyne, line 94: Similarly these don't apply to asymmetric objects, as you can't read one from the Realm.

The schema discovery parts do apply, though.

RealmSwift/Realm.swift
```diff
@@ -1101,6 +1101,25 @@ extension Realm {
...
+     cannot be queried locally, only objects which inherit from `AsymmetricObject`
+     can be created using this method.
+
+     Objects created using this method will not be added to the Realm.
+
+     - warning: This method may only be called during a write transaction.
+
+     - parameter object: The asymmetricObject to create.
+     */
+    public func create(_ object: AsymmetricObject) {
```
tgoyne, line 1118: This should be the same API as the normal `create()` function other than returning void.

RealmSwift/AsymmetricObject.swift
dianaafanador3: You are right, we are not supposed to use AsymmetricObjects within other top level object. Removing conformance to _RealmCollectionValueInsideOptional and _Persistable disallows using an asymmetric object within other Objects

RealmSwift/Realm.swift
dianaafanador3: Do you mean instead of passing an Asymmetric Object, we pass the values? @tgoyne

RealmSwift/Impl/ComplexTypes.swift
dianaafanador3: I guess we don't even need to conform to SchemaDiscovery as we only need to know if the table is asymmetric and that comes from the override `isAsymmetric`, or am I missing something?. I removed the conformance and everything works fine.

tgoyne: No, you're right and we don't need the schema discovery part either.

Realm/ObjectServerTests/SwiftAsymmetricSyncServerTests.swift
```diff
@@ -28,7 +28,8 @@ import RealmSyncTestSupport
 import RealmTestSupport
 #endif
 
-public class SwiftObjectAsymmetric: AsymmetricObject {
+let object = SwiftObjectAsymmetric()
```
tgoyne, line 31: Accidental global?

Realm/ObjectServerTests/SwiftAsymmetricSyncServerTests.swift
```diff
@@ -180,36 +191,43 @@ extension SwiftAsymmetricSyncTests {
...
     }
 
+    @MainActor
+    func checkCountInMongo(_ expectedCount: Int, forCollection collection: String) async throws {
+        let waitStart = Date()
+        let collection = try await setupCollection(collection)
+        while collection.count(filter: [:]).await(self) < expectedCount && waitStart.timeIntervalSinceNow > -600.0 {
+            sleep(5)
+        }
+    }
```
tgoyne, line 209: Needs to fail if the wait times out.

Realm/RLMAsymmetricObject.h
```diff
@@ -72,10 +72,12 @@ NS_ASSUME_NONNULL_BEGIN
...
  @warning This method may only be called during a write transaction.
 
- @param object The asymmetric object to create.
+ @param realm    The Realm to be used to create the asymmetric object..
+ @param value    The value used to populate the object.
 
+ @return  This will return `nil`.
  */
-+ (void)createInRealm:(RLMRealm *)realm withObject:(RLMAsymmetricObject *)object;
++ (instancetype)createInRealm:(RLMRealm *)realm withValue:(id)value __attribute__((warn_unused_result));
```
tgoyne, line 80: `void` return was correct here.

RealmSwift/AsymmetricObject.swift
```diff
@@ -61,7 +61,7 @@ extension AsymmetricObject: _RealmCollectionValueInsideOptional {
 
      - parameter value:  The value used to populate the object.
      */
-    public convenience init(value: Any) {
+    internal convenience init(value: Any) {
```
tgoyne, line 64: This should be public. Unmanaged asymmetric objects are a little weird, but they are something we allow.

Realm/RLMAsymmetricObject.h
leemaguire: somehow this is causing confusion in the `QueryTests`

```
Aug 15 08:01:32 /Users/realm/workspace/objc_pr/target/osx/xcode_version/13.2.1/Realm/Tests/QueryTests.m:1680:18: error: initializing 'QueryObject *__strong' with an expression of incompatible type 'void'
Aug 15 08:01:32     QueryObject *q1 = [self.queryObjectClass createInRealm:realm withValue:values];
```

leemaguire (commented):
Needs to be rebased onto latest master

Realm/ObjectServerTests/SwiftAsymmetricSyncServerTests.swift
```diff
@@ -0,0 +1,278 @@
...
+        XCTAssertTrue(realm.schema.objectSchema[0].isAsymmetric)
+    }
+
+    func testOpenLocalRealmWithAsymmetricObjectError() throws {
+        let configuration = Realm.Configuration(objectTypes: [SwiftObjectAsymmetric.self])
+        do {
+            _ = try Realm(configuration: configuration)
+            XCTFail("Opening a local Realm with an `Asymmetric` table should fail")
+        } catch {
+            XCTAssertEqual(error.localizedDescription, "Schema validation failed due to the following errors:\n- Asymmetric table \'SwiftObjectAsymmetric\' not allowed in a local Realm")
```
tgoyne, line 159: These should use XCTAssertThrowsError.

Realm/RLMAsymmetricObject.h
```diff
@@ -0,0 +1,99 @@
...
+ can be created using this method.
+
+ Objects created using this method will not be added to the Realm.
+
+ @warning This method may only be called during a write transaction.
+
+ @param realm    The Realm to be used to create the asymmetric object..
+ @param value    The value used to populate the object.
+
+ @return  This will return `nil`.
```
tgoyne, line 78: It does not return nil.

Realm/RLMAsymmetricObject.h
```diff
@@ -0,0 +1,99 @@
...
+          name.
+
+ @return  The class name for the model class.
+ */
++ (NSString *)className;
+
+/**
+ Creates an Asymmetric object, which will be synced unidirectionally and
+ cannot be queried locally, only objects which inherits from `RLMAsymmetricObject`
+ can be created using this method.
```
tgoyne, line 69: The second half of this sentence doesn't really make any sense.

RealmSwift/AsymmetricObject.swift
```diff
@@ -0,0 +1,100 @@
...
+/**
+ `AsymmetricObject` is a base class used to define asymmetric Realm objects.
+
+ Asymmetric objects can only be created using the `create(_ object:)`
+ function, and cannot be added, removed or queried.
+ When created, asymmetric objects will be synced unidirectionally to the MongoDB
+ database and cannot be accessed locally.
+
+ Incoming links from any asymmetric table are not allowed, meaning embedding
+ an asymmetric object within an `Object` will throw an error.
```
tgoyne, line 32: "table" is an implementation detail and not a word that should appear in the docs.

I think this sentence is backwards; an incoming link from an asymmetric object would be an asymmetric object linking to another object, which is the opposite of embedding an asymmetric object within an `Object`.

RealmSwift/AsymmetricObject.swift
```diff
@@ -0,0 +1,100 @@
...
+
+ The property types supported on `AsymmetricObject` are the same as for `Object`,
+ except for that asymmetric objects can only link to embedded objects, so `Object`
+ and `List<Object>` properties are not supported (`EmbeddedObject` and
+ `List<EmbeddedObject>` *are*).
+
+ ```swift
+ class Person: AsymmetricObject {
+     @Persisted(primaryKey: true) var _id: ObjectId = ObjectId.generate()
```
tgoyne, line 41:
```suggestion
     @Persisted(primaryKey: true) var _id: ObjectId
```

RealmSwift/Realm.swift
```diff
@@ -1101,6 +1101,27 @@ extension Realm {
     }
 }
 
+// MARK: Asymmetric Sync
+
+extension Realm {
+    /**
+     Creates an Asymmetric object, which will be synced unidirectionally and
+     cannot be queried locally, only objects which inherit from `AsymmetricObject`
+     can be created using this method.
```
tgoyne, line 1110: This is two sentences and should have a period, not a comma.

tgoyne (approved):

PR #7888, opened 2022-07-16
Add Sectioned Results

leemaguire:
Sectioned results allows you to take a `Results`, `List` & `MutableSet` collection and split the data into individual sections determined by a key generated from a property on the object it represents.

This works for both `Object` and primitive types of collections.

**Usage:**
```swift
class DemoObject: Object {
    @Persisted var title: String
    @Persisted var date: Date
    var firstLetter: String? {
        return title.first.map(String(_:)) // returns first letter of `sectionTitle`
    }
}

var sectionedResults: SectionedResults<String?, DemoObject>
...
sectionedResults = realm.objects(DemoObject.self)
    .sectioned(by: \.firstLetter, ascending: true)
```

The results can now be represented by individual sections in a tableview, where the section key is the first letter of `sectionTitle`.

*Note*: This is using a nightly Core version until `12.4.0` is available.


RealmSwift/SectionedResults.swift
```diff
@@ -0,0 +1,1043 @@
...
+     // ...
+     let dogs = realm.objects(Dog.self)
+     let sectionedResults = dogs.sectioned(by: \.age, ascending: true)
+     let token = sectionedResults.observe(keyPaths: ["name"]) { changes in
+         switch changes {
+         case .initial(let sectionedResults):
+            // ...
+         case .update:
+            // This case is hit:
+            // - after the token is intialized
```
dianaafanador3, line 252:
```suggestion
            // - after the token is initialized
```

RealmSwift/SectionedResults.swift
```diff
@@ -0,0 +1,1043 @@
...
```
dianaafanador3, line 377:
```suggestion
            // - after the token is initialized
```

RealmSwift/SectionedResults.swift
```diff
@@ -0,0 +1,1043 @@
...
```
dianaafanador3, line 495:
```suggestion
            // - after the token is initialized
```

RealmSwift/SectionedResults.swift
```diff
@@ -0,0 +1,1043 @@
...
```
dianaafanador3, line 658:
```suggestion
            // - after the token is initialized
```

RealmSwift/SectionedResults.swift
```diff
@@ -0,0 +1,1043 @@
...
```
dianaafanador3, line 782:
```suggestion
            // - after the token is initialized
```

RealmSwift/SectionedResults.swift
```diff
@@ -0,0 +1,1043 @@
...
```
dianaafanador3, line 899:
```suggestion
            // - after the token is initialized
```

RealmSwift/SectionedResults.swift
```diff
@@ -0,0 +1,1043 @@
...
+     - parameter modifications: The indexPaths of the objects which were modified in the previous version of this collection.
+     - parameter sectionsToInsert: The indexSet of the sections which were newly inserted into the sectioned results collection.
+     - parameter sectionsToDelete: The indexSet of the sections which were recently deleted from the previous sectioned results collection.
+     */
+    case update(Collection, deletions: [IndexPath], insertions: [IndexPath], modifications: [IndexPath],
+                sectionsToInsert: IndexSet, sectionsToDelete: IndexSet)
+    /**
+     If an error occurs, notification blocks are called one time with a `.error`
+     result and an `NSError` containing details about the error. This can only
+     currently happen if opening the Realm on a background thread to calcuate
```
dianaafanador3, line 989:
```suggestion
     currently happen if opening the Realm on a background thread to calculate
```

RealmSwift/SectionedResults.swift
```diff
@@ -0,0 +1,1043 @@
...
+     the change set fails. The callback will never be called again after it is
+     invoked with a .error value.
```
dianaafanador3, line 991:
```suggestion
     invoked with a `.error` value.
```

RealmSwift/SectionedResults.swift
```diff
@@ -0,0 +1,1043 @@
...
+        return RLMSectionedResultsIterator(collection: collection)
+    }
+
+    /// :nodoc:
+    public static func == (lhs: SectionedResults<Key, T>, rhs: SectionedResults<Key, T>) -> Bool {
+        return lhs.collection == rhs.collection
+    }
+}
+
+public struct ResultsSection<Key: _Persistable & Hashable, T: RealmCollectionValue>: RealmSectionedResultImpl, Identifiable {
```
dianaafanador3, line 559: It may be nice to have some documentation in `SectionResults` and `ResultsSection`

Realm/Tests/SwiftUITestHostUITests/SwiftUITestHostUITests.swift
```diff
@@ -350,4 +350,207 @@ class SwiftUITests: XCTestCase {
         let tableB = tables["ListB"]
         XCTAssertEqual(tableB.cells.count, 5)
     }
+
+    func testKeyPathObservedSectionedResults() {
```
dianaafanador3, line 354: It would be interesting to see a test, `@ObservedSectionResults` using sync.

Realm/RLMSectionedResults.mm
```diff
@@ -0,0 +1,753 @@
...
+#pragma clang diagnostic pop
+
+#pragma mark - Thread Confined Protocol Conformance
+
+- (realm::ThreadSafeReference)makeThreadSafeReference {
+    return _parent->_results;
+}
+
+- (id)objectiveCMetadata {
+    return @{
```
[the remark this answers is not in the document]
leemaguire: Changed to be a class. Is it safe to pass a reference to a block to be called on another thread?

Realm/RLMSectionedResults.mm
```diff
@@ -0,0 +1,753 @@
...
+    }
+
+    for (NSUInteger i = batchCount; i < len; ++i) {
+        _strongBuffer[i] = nil;
+    }
+
+    if (batchCount == 0) {
+        // Release our data if we're done, as we're autoreleased and so may
+        // stick around for a while
+        if (_sectionedResults) {
```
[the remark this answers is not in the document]
leemaguire: Fixed by consolidating both _sectionedResults & _resultsSection to now be _sectionedResult

RealmSwift/SwiftUI.swift
```diff
@@ -614,6 +614,397 @@ extension Projection: _ObservedResultsValue { }
...
+     must be responsible for determining the section key.
+     - parameter filter: Observations will be made only for passing objects.
+     If no filter given - all objects will be observed
+     - parameter keyPaths: Only properties contained in the key paths array will be observed.
+     If `nil`, notifications will be delivered for any property change on the object.
+     String key paths which do not correspond to a valid a property will throw an exception.
+     - parameter configuration: The `Realm.Configuration` used when creating the Realm.
+     If empty the configuration is set to the `defaultConfiguration`
+     */
+    public init(_ type: ResultType.Type,
```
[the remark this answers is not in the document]
leemaguire: projections does not support TSQ

Realm/RLMDictionary.mm
```diff
@@ -498,6 +498,17 @@ - (nullable NSArray *)objectsAtIndexes:(nonnull NSIndexSet *)indexes {
     @throw RLMException(@"This method is not available on RLMDictionary.");
 }
 
+- (RLMSectionedResults *)sectionedResultsSortedUsingKeyPath:(NSString *)keyPath
```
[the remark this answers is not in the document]
leemaguire: Given that RLMDictionary conforms to RLMCollection, we need to make sure we throw with a message in the case that a user tries to use this unsupported method.

CHANGELOG.md
```diff
@@ -5,7 +5,45 @@
...
+  var sectionedResults: SectionedResults<String, DemoObject>
+  ...
+  sectionedResults = realm.objects(DemoObject.self)
+      .sectioned(by: \.firstLetter, ascending: true)
+  ```
+* Add `@ObservedSectionedResults` for SwiftUI support. This property wrapper type retrieves sectioned results 
+  from a Realm using a keyPath or callback to determine the section key.
+  ```swift
+  struct DemoView: View {
+    @ObservedSectionedResults(DemoObject.self,
```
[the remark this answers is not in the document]
leemaguire: Should that be left to the docs?, my concern is that it will bloat the changelog

Realm/RLMSectionedResults.h
```diff
@@ -0,0 +1,956 @@
...
+ thread.
+
+ @warning This method cannot be called during a write transaction, or when the
+          containing Realm is read-only.
+ @warning Holding onto a frozen collection for an extended period while performing
+          write transaction on the Realm may result in the Realm file growing
+          to large sizes. See `RLMRealmConfiguration.maximumNumberOfActiveVersions`
+          for more information.
+ */
+- (instancetype)freeze;
```
ericjordanmossman, line 96: If you can freeze an individual section, what happens when a change is made to the sectionedResults that would cause a recalculation of the section keys, and the frozen section would be affected?

RealmSwift/Tests/SectionedResultsTests.swift
```diff
@@ -0,0 +1,1911 @@
...
+                    XCTAssertTrue(modifications.isEmpty)
+                    XCTAssertEqual(insertions, [IndexPath(item: 0, section: 0), IndexPath(item: 1, section: 0)])
+                }
+            case .error:
+                XCTFail("Shouldn't happen")
+            }
+        }
+
+        try! realm.write {
+            realm.delete([results[2], results[0]]) // banana, apple
```
ericjordanmossman, line 535: After this delete is made, it looks like the "a" section has the delete notify as a `sectionsToDelete` (490), because apple was the last "a". But the "b" section notification propagates as `deletions`(518) with an index, because "box" remains in the section. Is that right? 
Is there a way for the developer to see what the last object(s) of a section were in a deletion?

Realm/RLMSectionedResults.h
leemaguire: A frozen section will get a copy of the parent sectioned results, so any changes to the original parent won't do anything to frozen sections. A thawed section will produce a new up to date copy of the sectioned results, and if in the meantime that section has been deleted, the thaw method will throw.

RealmSwift/Tests/SectionedResultsTests.swift
leemaguire: 'Is that right?'– Yes that is correct.
'Is there a way for the developer to see what the last object(s) of a section were in a deletion?'- This would require setting up a realm with a transaction ref pointing to the previous state. We don't expose anything like that. Maybe if they really needed that they could hold a frozen instance of the sectioned results before their modifications and use that later on.

RealmSwift/SectionedResults.swift
```diff
@@ -0,0 +1,1050 @@
...
+    public static func == (lhs: SectionedResults<Key, T>, rhs: SectionedResults<Key, T>) -> Bool {
+        return lhs.collection == rhs.collection
+    }
+}
+
+
+/// `ResultsSection` is a collection which allows access  to objects that belong to a given section key.
+/// The collection is lazily evaluated, meaning that if the underlying collection has changed a full recalculation of the section keys will take place.
+/// A `ResultsSection` instance can be observed and it also conforms to `ThreadConfined`.
+public struct ResultsSection<Key: _Persistable & Hashable, T: RealmCollectionValue>: RealmSectionedResultImpl, Identifiable {
```
tgoyne, line 566: `Identifiable` has a deployment target of iOS 13 and so needs to be in an extension guarded by `@available`. I think the reason it doesn't give any warnings here is that our tests are built with a higher deployment target.

tgoyne:
The interaction of keyPath filtering and sectioning where the field used for sectioning isn't in the filter (i.e. something like `objects.sectioned(by: \.firstLetter).observe(keyPaths: [\.age])`) doesn't appear to be documented or tested. I think the implementation handles it correctly, but it'd be worth writing a test to verify and the documentation should explicitly mention what the behavior will be.

RealmSwift/SectionedResults.swift
```diff
@@ -0,0 +1,1050 @@
...
+     - parameter queue: The serial dispatch queue to receive notification on. If
+                        `nil`, notifications are delivered to the current thread.
+     - parameter block: The block to be called whenever a change occurs.
+     - returns: A token which must be held for as long as you want updates to be delivered.
+     */
+    public func observe<T: ObjectBase>(keyPaths: [PartialKeyPath<T>],
+                                       on queue: DispatchQueue? = nil,
+                                       _ block: @escaping (RealmSectionedResultsChange<Self>) -> Void) -> NotificationToken {
+        return collection.addNotificationBlock(wrapObserveBlock(block), keyPaths: keyPaths.map(_name(for:)), queue: queue)
+    }
```
tgoyne, line 957: This needs to be:

```
extension ResultsSection where Element: RLMObjectBase {
    public func observe(keyPaths: [PartialKeyPath<Element>],
                        on queue: DispatchQueue? = nil,
                        _ block: @escaping (RealmSectionedResultsChange<Self>) -> Void) -> NotificationToken {
        return collection.addNotificationBlock(wrapObserveBlock(block), keyPaths: keyPaths.map(_name(for:)), queue: queue)
    }
}
```

The current definition accepts nonsense (such as a keypath on an unrelated type), and that's why passing bare `\.stringCol` doesn't work without explicitly specifying the base type.

RealmSwift/SectionedResults.swift
```diff
@@ -0,0 +1,1050 @@
...
+            }
+            block(RealmSectionedResultsChange.fromObjc(value: col, change: change, error: error))
+        }
+    }
+}
+
+/// `SectionedResults` is a type safe collection which holds individual `ResultsSection`s as its elements.
+/// The container is lazily evaluated, meaning that if the underlying collection has changed a full recalculation of the section keys will take place.
+/// A `SectionedResults` instance can be observed and it also conforms to `ThreadConfined`.
+public struct SectionedResults<Key: _Persistable & Hashable, T: RealmCollectionValue>: RealmSectionedResultImpl {
```
tgoyne, line 150: Both of these types need a `Key` typealias so that the generic parameter can be accessed from outside the type. We get `Element` for free from the base collection protocol.

dianaafanador3 (approved):
Very good job with this!

tgoyne (approved):

PR #7894, opened 2022-07-25
Recovery mode for partition realms

ericjordanmossman:
Adds the `.recover`, `.recoverOrDiscard`, `RLMClientResetModeRecover`, and `RLMClientResetModeRecoverOrDiscard` to `ClientResetMode` and `RLMClientResetMode` respectively.

The new recover modes function by downloading a realm with objects which reflect the latest version of the server after a client reset. The recovery process is run locally in an attempt to integrate the server version with any changes from 
the local realm before the client reset occurred.
1. Objects created locally that were not synced before client reset will be integrated.
2. If an object has been deleted on the server, but was modified on the client, the delete takes precedence and the update is discarded
3. If an object was deleted on the client, but not the server, then the client delete instruction is applied.
4. In the case of conflicting updates to the same field, the most recent update is applied.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,18 +2,18 @@ x.y.z Release notes (yyyy-MM-dd)
 =============================================================
 ### Enhancements
 * Add `.recover` and `.recoverOrDiscardLocal` behaviors to `ClientResetMode`. See below for detail.
-* Add `RLMClientResetModeRecover` and `RLMClientResetModeRecoverOrDiscard` behaviors to `RLMClientResetMode` enums.  See below for detail.
+* Add `RLMClientResetModeRecover` and `RLMClientResetModeRecoverOrDiscard` behaviors to `RLMClientResetMode`.  See below for detail.
 * The new recover modes function by downloading a realm with objects which reflect 
   the latest version of the server after a client reset. The recovery process is run 
   locally in an attempt to integrate the server version with any local changes from 
   before the client reset occurred.
   The changes are integrated with the following rules:
-  1. Objects created locally that were not integrated before client reset will be integrated.
-  2. If an object has been deleted on the server, but was modified on the recovering client, the delete takes precedence and the update is discarded
-  3. If an object was deleted on the recovering client, but not the server, then the client delete instruction is applied.
+  1. Objects created locally that were not synced before client reset will be integrated.
+  2. If an object has been deleted on the server, but was modified on the client, the delete takes precedence and the update is discarded
+  3. If an object was deleted on the client, but not the server, then the client delete instruction is applied.
   4. In the case of conflicting updates to the same field, the most recent update is applied.
-  - The two new swift recovery also support client reset callbacks: `.recover(((Realm) -> Void)? = nil, ((Realm, Realm) -> Void)? = nil)`.
-  - The two new Obj-C recovery modes support client reset callback via the `notifyBeforeReset` and `notifyAfterReset` parameters in `[RLMUser configurationWithPartitionValue]`
+  - The two new swift recovery modes support client reset callbacks: `.recover(((Realm) -> Void)? = nil, ((Realm, Realm) -> Void)? = nil)`.
+  - The two new Obj-C recovery modes support client reset callbacks via the `notifyBeforeReset` and `notifyAfterReset` parameters in `[RLMUser configurationWithPartitionValue]`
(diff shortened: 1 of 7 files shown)


Realm/RLMSyncConfiguration.h
```diff
@@ -50,7 +50,18 @@ typedef NS_ENUM(NSUInteger, RLMClientResetMode) {
...
+    /// The client device will download a realm with objects reflecting the latest version of the server. A recovery
+    /// process is run locally in an attempt to integrate the server version with any local changes from before the
+    /// client reset occurred.
+    /// The changes are integrated with the following rules:
+    /// 1. Objects created locally that were not integrated before client reset will be integrated.
+    /// 2. If an object has been deleted on the server, but was modified on the recovering client, the delete takes precedence and the update is discarded
+    /// 3. If an object was deleted on the recovering client, but not the server, then the client delete instruction is applied.
+    /// 4. In the case of conflicting updates to the same field, the most recent update is applied.
+    RLMClientResetModeRecover,
+    // TODO: docs
```
ericjordanmossman, line 63: Docs. Need to make a note that .recoverOrDiscardLocal falls back to discard local if the recovery process fails. And .recover falls back to manual client reset handler.

ericjordanmossman, line 63: Docs. Need to make a note that .recoverOrDiscardLocal falls back to discard local if the recovery process fails. And .recover falls back to manual client reset handler.

CHANGELOG.md
```diff
@@ -1,7 +1,20 @@
...
 ### Enhancements
-* None.
+* Add `.recover` and `.recoverOrDiscardLocal` behaviors to `ClientResetMode`. See below for detail.
+* Add `RLMClientResetModeRecover` and `RLMClientResetModeRecoverOrDiscard` behaviors to `RLMClientResetMode` enums.  See below for detail.
+* The new recover modes function by downloading a realm with objects which reflect 
+  the latest version of the server after a client reset. The recovery process is run 
+  locally in an attempt to integrate the server version with any local changes from 
+  before the client reset occurred.
+  The changes are integrated with the following rules:
+  1. Objects created locally that were not integrated before client reset will be integrated.
```
ericjordanmossman, line 11: "not synced before client reset will be integrated"

Realm/ObjectServerTests/SwiftObjectServerTests.swift
```diff
@@ -508,6 +526,10 @@ class SwiftObjectServerTests: SwiftSyncTestCase {
         case .discardLocal(let before, let after):
             XCTAssertNotNil(before)
             XCTAssertNotNil(after)
+        case .recover:
```
tgoyne, line 529: Could be just a default case rather than three separate invalid ones.

Realm/RLMSyncConfiguration.h
```diff
@@ -50,7 +50,40 @@ typedef NS_ENUM(NSUInteger, RLMClientResetMode) {
...
+    /// process is run locally in an attempt to integrate the server version with any local changes from before the
+    /// client reset occurred.
+    ///
+    /// The changes are integrated with the following rules:
+    /// 1. Objects created locally that were not synced before client reset will be integrated.
+    /// 2. If an object has been deleted on the server, but was modified on the client, the delete takes precedence and the update is discarded
+    /// 3. If an object was deleted on the client, but not the server, then the client delete instruction is applied.
+    /// 4. In the case of conflicting updates to the same field, the most recent update is applied.
+    ///
+    /// If the recovery process fails, the client reset process falls back to RLMClientResetModeDiscardLocal.
```
tgoyne, line 82: It'd be good to mention *why* it could fail. Is it just if recovery mode is disabled on the server?

RealmSwift/Sync.swift
```diff
@@ -218,7 +218,7 @@ public enum ClientResetMode {
     case manual
     /// - see: `RLMClientResetModeDiscardLocal` for more details on `.discardLocal` behavior
     ///
-    /// The first `.discardLocal` function argument notifies prior to a client reset occurring.
+    /// The first `.discardLocal` function argument, `((Realm) -> Void)? = nil`, is executed prior to a client reset occurring.
```
tgoyne, line 221: I don't think it's useful to repeat the type of the argument in the docs, as that's already displayed.

ericjordanmossman: Okay will take it out. I was on the fence since in swift these blocks don't have a proper name.

Realm/RLMSyncConfiguration.h
ericjordanmossman: Yes - that and if an incompatible schema change exists.

PR #7895, opened 2022-07-25
Ensure sync sessions stay alive while there is a pending error

tgoyne:
If the last external reference to a sync session was dropped between when the error handler on the sync thread ran and when we delivered the error to the user on the main thread, the user would see a sync session in a weird invalid state in the error handler due to that RLMSyncSession only holds a weak reference.

### DIFF
--- a/Realm/ObjectServerTests/SwiftObjectServerTests.swift
+++ b/Realm/ObjectServerTests/SwiftObjectServerTests.swift
@@ -449,7 +449,7 @@ class SwiftObjectServerTests: SwiftSyncTestCase {
 
         // After restarting sync, the sync history translator service needs time
         // to resynthesize the new history from existing objects on the server
-        // The following creates a new realm with the same parition and wait for
+        // The following creates a new realm with the same partition and wait for
         // downloads to ensure the the new history has been created.
         try autoreleasepool {
             var newConfig = user.configuration(partitionValue: partition)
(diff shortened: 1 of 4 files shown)


Realm/ObjectServerTests/SwiftObjectServerTests.swift
```diff
@@ -472,6 +472,90 @@ class SwiftObjectServerTests: SwiftSyncTestCase {
...
+        let syncManager = self.app.syncManager
+        syncManager.errorHandler = { error, session in
+            guard let error = error as? SyncError else {
+                XCTFail("Bad error type: \(error)")
+                return
+            }
+            XCTAssertEqual(error.code, .clientResetError)
+            XCTAssertEqual(session?.state, .inactive)
+            XCTAssertEqual(session?.connectionState, .disconnected)
+            XCTAssertEqual(session?.parentUser()?.id, user.id)
```
ericjordanmossman, line 542: So I understand this test correctly:
If the change in RLMSyncConfiguration.mm weren't there then the `session` in 540-542 would be in an invalid state/not exist because it would have been deallocated because there were no strong references.

ericjordanmossman (approved):
LGTM 
The CI failures look like processes aborting and not related to PR.

Realm/ObjectServerTests/SwiftObjectServerTests.swift
tgoyne: Yes, without the fix the third check on 542 failed because the `id` would be `nil`. It was technically a "valid" state, but just not a very useful one.

leemaguire (approved):

PR #7938, opened 2022-09-01
Recovery client reset

ericjordanmossman:
Add `.recoverUnsyncedChanges` (`RLMClientResetModeRecoverUnsyncedChanges`) and
`.recoverOrDiscardUnsyncedChanges` behaviors to `ClientResetMode` (`RLMClientResetMode`). 
More on [client resets](https://www.mongodb.com/docs/atlas/app-services/sync/error-handling/client-resets/).

***
The default `clientResetMode` (`RLMClientResetMode`) is switched from `.manual` (`RLMClientResetModeManual`)
  to `.recoverUnsyncedChanges` (`RLMClientResetModeRecoverUnsyncedChanges`).

This is a breaking change for any applications using a manual client reset handler.
***

The new recover modes function by downloading a realm with objects which reflect the latest
    version of the server after a client reset. A recovery process is run locally in an
    attempt to integrate the server version with any local changes from before the
    client reset occurred.
    The changes are integrated with the following rules:
    1. Objects created locally that were not synced before client reset will be integrated.
    2. If an object has been deleted on the server, but was modified on the client, the delete takes precedence and the update is discarded
    3. If an object was deleted on the client, but not the server, then the client delete instruction is applied.
    4. In the case of conflicting updates to the same field, the most recent update is applied.

***

This also adds the option to set a manual client reset handler on the `.manual` mode enum.
`ClientResetMode.manual` -> `ClientResetMode.manual(ErrorReportingBlock? = nil)`
Previously the manual client reset handler could only be set on the SyncManager.errorHandler. The intention is that the new `.manual(ErrorReportingBlock? = nil)` interface more closely mirrors the other modes.


Realm/RLMSyncConfiguration.mm
```diff
@@ -270,32 +285,43 @@ - (BOOL)enableFlexibleSync {
...
+- (void)setErrorHandler:(RLMUser *)user {
     RLMSyncManager *manager = [user.app syncManager];
     __weak RLMSyncManager *weakManager = manager;
-    config.error_handler = [weakManager](std::shared_ptr<SyncSession> errored_session, SyncError error) {
+    _config->error_handler = [weakManager, self](std::shared_ptr<SyncSession> errored_session, SyncError error) {
         RLMSyncErrorReportingBlock errorHandler;
         @autoreleasepool {
             errorHandler = weakManager.errorHandler;
         }
-        if (!errorHandler) {
```
dianaafanador3, line 284: why remove this, I would probably first check if both errorHandler and _manualClientResetHandler are not null, in that case is not even worth it to check for the error

Realm/RLMSyncConfiguration.mm
```diff
@@ -270,32 +285,43 @@ - (BOOL)enableFlexibleSync {
...
-            return;
-        }
         NSError *nsError = RLMTranslateSyncError(std::move(error));
         if (!nsError) {
             return;
         }
+        if (!errorHandler && nsError.code != RLMSyncErrorClientResetError) {
```
dianaafanador3, line 300: MaybeI'm not seen it but both this conditionals seems unnecessary

Realm/RLMUser.h
```diff
@@ -155,13 +212,20 @@ NS_ASSUME_NONNULL_BEGIN
...
+ @param beforeResetBlock A callback which notifies prior to a client reset occurring. See: `RLMClientResetBeforeBlock`
+ @param afterResetBlock A callback which notifies after a client reset has occurred. See: `RLMClientResetAfterBlock`
  @param rerunOnOpen If true, allows to run the initial set of subscriptions specified, on every app startup.
                     This can be used to re-run dynamic time ranges and other queries that require a
                     re-computation of a static variable.
 
  @return A `RLMRealmConfiguration` instance with a flexible sync configuration.
  */
 - (RLMRealmConfiguration *)flexibleSyncConfigurationWithInitialSubscriptions:(RLMFlexibleSyncInitialSubscriptionsBlock)initialSubscriptions
+                                                             clientResetMode:(RLMClientResetMode)clientResetMode
```
dianaafanador3, line 226: I will probably keep the order of the arguments, because rerunOnOpen makes sense after InitialSubscriptions

Realm/RLMUser.h
```diff
@@ -155,13 +212,20 @@ NS_ASSUME_NONNULL_BEGIN
...
+                                                           notifyBeforeReset:(nullable RLMClientResetBeforeBlock)beforeResetBlock
+                                                            notifyAfterReset:(nullable RLMClientResetAfterBlock)afterResetBlock
```
dianaafanador3, line 228: This initialiser are missing  `flexibleSyncConfiguration` with manual `manualClientResetHandler`and `flexibleSyncConfigurationWithInitialSubscriptions`with `manualClientResetHandler` as well.

Realm/RLMSyncUtil.mm
```diff
@@ -45,8 +45,12 @@
 
 using namespace realm;
 
-static_assert((int)RLMClientResetModeManual == (int)realm::ClientResyncMode::Manual);
```
dianaafanador3, line 48: Can we convert RLMClientResetModeManual to RLMClientResetModeDiscardUnsyncedChanges, so we don't have this warning on the code

RealmSwift/Sync.swift
```diff
@@ -253,7 +252,35 @@ public enum ClientResetMode {
     /// ```
     ///  For more details on the second block: ((Realm, Realm) -> Void)? = nil,
     /// - see: `RLMClientResetAfterBlock`
+    @available(*, deprecated, message: "Use discardUnsyncedChanges")
     case discardLocal(((Realm) -> Void)? = nil, ((Realm, Realm) -> Void)? = nil)
+    /// Identical in behavior to `.discardLocal.` Will fully replace
+    /// `.discardLocal.` once it's removed.
+    case discardUnsyncedChanges(((Realm) -> Void)? = nil, ((Realm, Realm) -> Void)? = nil)
+    /// - see: `RLMClientResetModeRecoverUnsyncedChanges` for more details on `.recover` behavior
```
dianaafanador3, line 260: This will better to have its own dosctrings instead of linking to the Objective-C documentation

RealmSwift/Sync.swift
```diff
@@ -253,7 +252,35 @@ public enum ClientResetMode {
     /// ```
     ///  For more details on the second block: ((Realm, Realm) -> Void)? = nil,
     /// - see: `RLMClientResetAfterBlock`
+    @available(*, deprecated, message: "Use discardUnsyncedChanges")
     case discardLocal(((Realm) -> Void)? = nil, ((Realm, Realm) -> Void)? = nil)
+    /// Identical in behavior to `.discardLocal.` Will fully replace
+    /// `.discardLocal.` once it's removed.
+    case discardUnsyncedChanges(((Realm) -> Void)? = nil, ((Realm, Realm) -> Void)? = nil)
```
dianaafanador3, line 259: This will better to have its own dosctrings instead of linking to the Objective-C documentation, given that there is a callback which needs more detailed documentation

dianaafanador3, line 259: I supposed this callback are beforeClientReset and afterClientReset, but are not fully documented here and probably needs an argument name

RealmSwift/Sync.swift
```diff
@@ -253,7 +252,35 @@ public enum ClientResetMode {
...
+    /// - see: `RLMClientResetModeRecoverUnsyncedChanges` for more details on `.recover` behavior
+    ///
+    /// - see `ClientResetMode.discardLocal` above for a detailed explanation of the
+    ///   two callback arguments: `((Realm) -> Void)? = nil, ((Realm, Realm) -> Void)? = nil`
+    case recoverUnsyncedChanges(((Realm) -> Void)? = nil, ((Realm, Realm) -> Void)? = nil)
```
dianaafanador3, line 264: Same as `discardUnsyncedChanges`

RealmSwift/Sync.swift
```diff
@@ -253,7 +252,35 @@ public enum ClientResetMode {
...
+    /// - see: `RLMClientResetModeRecoverOrDiscardUnsyncedChanges` for more details on `.recoverOrDiscard` behavior
+    ///
+    /// - see `ClientResetMode.discardLocal` for a detailed explanation of the
+    ///   two callback arguments: `((Realm) -> Void)? = nil, ((Realm, Realm) -> Void)? = nil`
+    case recoverOrDiscardUnsyncedChanges(((Realm) -> Void)? = nil, ((Realm, Realm) -> Void)? = nil)
```
dianaafanador3, line 269: Same `discardUnsyncedChanges`

RealmSwift/Sync.swift
```diff
@@ -277,17 +304,23 @@ public enum ClientResetMode {
...
 
      - see: `ClientResetMode` and `RLMClientResetMode`
      - see: https://docs.mongodb.com/realm/sync/error-handling/client-resets/
     */
     public var clientResetMode: ClientResetMode {
         switch config.clientResetMode {
         case .manual:
-            return .manual
+            return .manual(config.manualClientResetHandler)
         case .discardLocal:
```
dianaafanador3, line 316: I would probably change this to
```
case .discardLocal, .discardUnsyncedChanges:
            return .discardUnsyncedChanges(ObjectiveCSupport.convert(object: config.beforeClientReset), ObjectiveCSupport.convert(object: config.afterClientReset))
```

RealmSwift/Sync.swift
```diff
@@ -425,7 +458,7 @@ public extension User {
      Additional settings can be optionally specified. Descriptions of these
      settings follow.
 
-     `ClientResetMode` is `.manual` by default.
+     `ClientResetMode` is `.recover` by default.
```
dianaafanador3, line 461:
```suggestion
     `ClientResetMode` is `.recoverUnsyncedChanges` by default.
```

RealmSwift/Sync.swift
```diff
@@ -438,22 +471,38 @@ public extension User {
      Create a sync configuration instance.
 
      - parameter partitionValue: The `BSON` value the Realm is partitioned on.
-     - parameter clientResetMode: Determines file recovery behavior during a client reset.
+     - parameter clientResetMode: Determines file recovery behavior during a client reset. `.recover` by default.
```
dianaafanador3, line 474:
```suggestion
     - parameter clientResetMode: Determines file recovery behavior during a client reset. `. recoverUnsyncedChanges ` by default.
```

RealmSwift/Sync.swift
```diff
@@ -209,13 +209,12 @@ public typealias Provider = RLMIdentityProvider
 
 /**
  An enum used to determines file recovery behavior in the event of a client reset.
+ Defaults to  `.recover`.
```
dianaafanador3, line 212:
```suggestion
 Defaults to  `. recoverUnsyncedChanges `.
```

Realm/ObjectServerTests/SwiftObjectServerTests.swift
```diff
@@ -521,27 +669,123 @@ class SwiftObjectServerTests: SwiftSyncTestCase {
...
+
+        try autoreleasepool {
+            let realm = try Realm(configuration: configuration)
+            waitForDownloads(for: realm)
+            // After reopening, the old Realm file should have been moved aside
+            // and we should now have the data from the server
+            XCTAssertEqual(realm.objects(SwiftPerson.self).count, 1)
+            XCTAssertEqual(realm.objects(SwiftPerson.self)[0].firstName, "Paul")
         }
+    }
```
dianaafanador3, line 777: We may need an extra test that makes sure the manual mode callback is called even if the error handler is set as well

dianaafanador3 (commented):
This branch needs rebase now that asymmetric sync is merged to see is everything is working correctly after the baas update

CHANGELOG.md
```diff
@@ -1,15 +1,57 @@
...
+    you must explicitly make is designating manual mode in
+    your `Realm.Configuration.SyncConfiguration`s, since they will now default to `.recoverUnsyncedChanges`.
+  - You may choose to define your manual client reset handler in the newly
+    introduced `manual(ErrorReportingBlock? = nil)`
+    or `RLMSyncConfiguration.manualClientResetHandler`, but this is not required.
+    The `SyncManager.errorHandler` will still be invoked during a client reset if
+    no callback is passed into these new interfaces.
+
+### Compatibility
+* This release introduces breaking changes for applications handling client resets manually.
```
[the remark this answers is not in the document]
ericjordanmossman: No, I'll take this out. I added a '### Deprecations', which we've used in the past.

CHANGELOG.md
```diff
@@ -1,15 +1,57 @@
 x.y.z Release notes (yyyy-MM-dd)
 =============================================================
 ### Enhancements
-* None.
+* Add `.recoverUnsyncedChanges` (`RLMClientResetModeRecoverUnsyncedChanges`) and
+`.recoverOrDiscardUnsyncedChanges` behaviors to `ClientResetMode` (`RLMClientResetMode`).
+  - The new recover modes function by downloading a realm with objects which reflect the latest
+    version of the server after a client reset. A recovery process is run locally in an
```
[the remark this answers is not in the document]
ericjordanmossman: The device asks for the server realm after the client reset error thrown. So I'll edit this to say "latest state of the server after a client reset." 

And yes, core downloads a completely a new realm on disk, then carries out the recovery process entirely on device.

Realm/ObjectServerTests/RLMObjectServerTests.mm
```diff
@@ -1473,9 +1475,9 @@ - (void)testSetClientResetMode {
     RLMRealmConfiguration *config = [user configurationWithPartitionValue:partitionValue clientResetMode:RLMClientResetModeDiscardLocal];
```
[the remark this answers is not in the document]
ericjordanmossman: I left the deprecated discardLocal in some obj-c and swift tests so that I could be sure I wasn't breaking it, just deprecating. I imagine once it's removed from the api we can completely scrub it from all tests.

Realm/ObjectServerTests/RealmServer.swift
```diff
@@ -1014,28 +1015,66 @@ public class RealmServer: NSObject {
         app.sync.config.put(data: ["development_mode_enabled": true], completion)
     }
 
-    public func disableSync(appServerId: String, syncServiceId: String, completion: @escaping (Result<Any?, Error>) -> Void) {
```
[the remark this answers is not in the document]
ericjordanmossman: It's possible. disableSync patches the config with a "". enableSync patches the config with whatever config is passed into the method. The different param `syncServiceConfiguration: [String: Any]` in enableSync makes me think to keep them separate and leave that param non-optional.

Realm/RLMSyncConfiguration.h
```diff
@@ -84,25 +138,31 @@ typedef void(^RLMClientResetAfterBlock)(RLMRealm * _Nonnull beforeFrozen, RLMRea
...
  A callback which notifies after a client reset has occurred.
- -see: `RLMClientResetAfterBlock`
+ @see: `RLMClientResetAfterBlock`
  */
 @property (nonatomic, nullable) RLMClientResetAfterBlock afterClientReset;
 
+/**
+ 
+ */
+@property (nonatomic, nullable) RLMSyncErrorReportingBlock manualClientResetHandler;
```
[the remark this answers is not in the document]
ericjordanmossman: I think you're right. The idea is that all the other SDKs are trying to make the manual mode handler set similar to the other three modes. It'd be odd to have four modes, but one of the modes would have a handler set elsewhere. But having this manualClientResetHandler kind of makes it uglier. I suppose we do need this at least privately so that `setErrorHandler` can invoke the callback passed down from the swift layer.

Realm/RLMSyncConfiguration.h
```diff
@@ -41,16 +43,68 @@ typedef NS_ENUM(NSUInteger, RLMClientResetMode) {
...
+    /// process is run locally in an attempt to integrate the server version with any local changes from before the
+    /// client reset occurred.
+    ///
+    /// The changes are integrated with the following rules:
+    /// 1. Objects created locally that were not synced before client reset will be integrated.
+    /// 2. If an object has been deleted on the server, but was modified on the client, the delete takes precedence and the update is discarded
+    /// 3. If an object was deleted on the client, but not the server, then the client delete instruction is applied.
+    /// 4. In the case of conflicting updates to the same field, the most recent update is applied.
+    ///
+    /// If the recovery integration fails, the client reset process falls back to RLMClientResetModeDiscardLocal.
```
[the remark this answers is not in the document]
ericjordanmossman: It is - I'll add this to changelog too to address another comment.

RealmSwift/Sync.swift
ericjordanmossman: RLMClientResetDiscardLocal and RLMClientResetDiscardUnsyncedChanges both assert to the same core enum. So the switch cases get obscured. Is someone is using .discardLocal and have switch cases for discardLocal, then those cases won't hit. That'd be a breaking change for them.

But that means for anyone using discardUnsyncedChanges, their cases hit on discardLocal but not discardUnsyncedChanges (see testClientResetDiscardUnsyncedChanges).

How can we differentiate? I tried using raw values in the swift enum, but you can't have a raw value _and_ enum parameters.

Realm/RLMSyncUtil.mm
ericjordanmossman: RLMClientResetModeManual will still need to assert to manual. For RLMClientResetModeDiscardLocal do you mean like convert in our objc convert file?

Realm/RLMSyncConfiguration.h
```diff
@@ -41,16 +43,68 @@ typedef NS_ENUM(NSUInteger, RLMClientResetMode) {
...
+    /// client reset occurred.
+    ///
+    /// The changes are integrated with the following rules:
+    /// 1. Objects created locally that were not synced before client reset will be integrated.
+    /// 2. If an object has been deleted on the server, but was modified on the client, the delete takes precedence and the update is discarded
+    /// 3. If an object was deleted on the client, but not the server, then the client delete instruction is applied.
+    /// 4. In the case of conflicting updates to the same field, the most recent update is applied.
+    ///
+    /// If the recovery integration fails, the client reset process falls back to RLMClientResetModeManual.
+    /// The recovery integration will fail if the "Client Recovery" setting is not enabled on the server.
```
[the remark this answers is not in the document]
ericjordanmossman: No, I don't think so :(  I tried looking.

CHANGELOG.md
```diff
@@ -1,15 +1,57 @@
...
+    attempt to integrate the server version with any local changes from before the
+    client reset occurred.
+    The changes are integrated with the following rules:
```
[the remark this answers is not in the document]
ericjordanmossman: Added bullet points below

RealmSwift/Sync.swift
```diff
@@ -438,22 +471,38 @@ public extension User {
...
      */
     func configuration<T: BSON>(partitionValue: T,
                                 clientResetMode: ClientResetMode) -> Realm.Configuration {
         var config: RLMRealmConfiguration
         switch clientResetMode {
-        case .manual:
+        case .manual(let manualClientReset):
             config = self.__configuration(withPartitionValue: ObjectiveCSupport.convert(object: AnyBSON(partitionValue)),
-                                              clientResetMode: .manual)
+                                          clientResetMode: .manual,
```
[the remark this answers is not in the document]
ericjordanmossman: See: https://github.com/realm/realm-swift/pull/7938/files#r964123264

RealmSwift/Sync.swift
```diff
@@ -895,8 +944,20 @@ extension User {
...
      @return A `Realm.Configuration` instance with a flexible sync configuration.
      */
-    public func flexibleSyncConfiguration() -> Realm.Configuration {
-        let config = self.__flexibleSyncConfiguration()
+    public func flexibleSyncConfiguration(clientResetMode: ClientResetMode = .recoverUnsyncedChanges(nil, nil)) -> Realm.Configuration {
+        var config: RLMRealmConfiguration
+        switch clientResetMode {
+        case .manual:
+            throwRealmException("Manual mode not supported for flexible sync configurations")
+        case .discardLocal(let beforeBlock, let afterBlock):
```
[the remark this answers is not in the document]
ericjordanmossman: https://github.com/realm/realm-swift/pull/7938/files#r964123264

Realm/RLMSyncConfiguration.mm
ericjordanmossman: Something like
```
if (!errorHandler && !_manualClientResetHandler) {
            return;
        }

        RLMSyncSession *session = [[RLMSyncSession alloc] initWithSyncSession:errored_session];
        dispatch_async(dispatch_get_main_queue(), ^{
            // Keep the SyncSession alive until the callback completes as
            // RLMSyncSession only holds a weak reference
            static_cast<void>(errored_session);
            if (_manualClientResetHandler && nsError.code == RLMSyncErrorClientResetError) {
                self.manualClientResetHandler(nsError, session);
            } else {
                errorHandler(nsError, session);
            }
        });
```
Would crash on  errorHandler(nsError, session); for sync errors other than client reset, I think.

Realm/RLMSyncConfiguration.mm
ericjordanmossman: https://github.com/realm/realm-swift/pull/7938/files#r965220073

Realm/RLMSyncConfiguration.mm
```diff
@@ -188,12 +189,26 @@ - (void)setAfterClientReset:(RLMClientResetAfterBlock)afterClientReset {
...
     }
 }
 
+- (RLMSyncErrorReportingBlock)manualClientResetHandler {
+    return _manualClientResetHandler;
+}
+
+- (void)setManualClientResetHandler:(RLMSyncErrorReportingBlock)manualClientReset {
+    if (!manualClientReset) {
+        return;
```
tgoyne, line 204: Why is this making it so that you can't set the block back to `nil` after it's set?

Realm/RLMSyncConfiguration.mm
```diff
@@ -270,32 +285,43 @@ - (BOOL)enableFlexibleSync {
...
+                self.manualClientResetHandler(nsError, session);
+            } else {
+                errorHandler(nsError, session);
+            }
         });
     };
+};
+
+static void setDefaults(SyncConfig& config, RLMUser *user) {
+    config.client_resync_mode = ClientResyncMode::Manual;
```
tgoyne, line 321: The changelog claims this is no longer the default.

Realm/RLMSyncConfiguration.mm
```diff
@@ -270,32 +285,43 @@ - (BOOL)enableFlexibleSync {
...
 
-static void setDefaults(SyncConfig& config, RLMUser *user) {
-    config.client_resync_mode = ClientResyncMode::Manual;
-    config.stop_policy = SyncSessionStopPolicy::AfterChangesUploaded;
-
+- (void)setErrorHandler:(RLMUser *)user {
     RLMSyncManager *manager = [user.app syncManager];
     __weak RLMSyncManager *weakManager = manager;
-    config.error_handler = [weakManager](std::shared_ptr<SyncSession> errored_session, SyncError error) {
+    _config->error_handler = [weakManager, self](std::shared_ptr<SyncSession> errored_session, SyncError error) {
```
tgoyne, line 291: This is a retain cycle.

Realm/RLMSyncConfiguration.mm
```diff
@@ -270,32 +285,43 @@ - (BOOL)enableFlexibleSync {
...
+        }
+
         RLMSyncSession *session = [[RLMSyncSession alloc] initWithSyncSession:errored_session];
         dispatch_async(dispatch_get_main_queue(), ^{
             // Keep the SyncSession alive until the callback completes as
             // RLMSyncSession only holds a weak reference
             static_cast<void>(errored_session);
-            errorHandler(nsError, session);
+            if (_manualClientResetHandler && nsError.code == RLMSyncErrorClientResetError) {
+                self.manualClientResetHandler(nsError, session);
```
tgoyne, line 312: Mixing self and ivar use for manualClientResetHandler is weird.

Realm/RLMSyncConfiguration.mm
```diff
@@ -270,32 +285,43 @@ - (BOOL)enableFlexibleSync {
     return make_sync_error(errorClass, @(error.message.c_str()), error.error_code.value(), custom);
 }
 
-static void setDefaults(SyncConfig& config, RLMUser *user) {
-    config.client_resync_mode = ClientResyncMode::Manual;
-    config.stop_policy = SyncSessionStopPolicy::AfterChangesUploaded;
-
+- (void)setErrorHandler:(RLMUser *)user {
```
tgoyne, line 288: `setErrorHandler:` is a bad name for this function, as it's not a setter for the `errorHandler` property.

Realm/RLMSyncUtil.mm
```diff
@@ -45,8 +45,12 @@
...
 static_assert((int)RLMClientResetModeDiscardLocal == (int)realm::ClientResyncMode::DiscardLocal);
+static_assert((int)RLMClientResetModeDiscardUnsyncedChanges == (int)realm::ClientResyncMode::DiscardLocal);
```
tgoyne, line 49: Things using deprecated symbols need to be wrapped in
```
#pragma clang diagnostic push
#pragma clang diagnostic ignored "-Wdeprecated-declarations"
// thing
#pragma clang diagnostic pop
```

Realm/RLMUser.h
```diff
@@ -114,6 +114,8 @@ NS_ASSUME_NONNULL_BEGIN
  @param partitionValue The `RLMBSON` value the Realm is partitioned on.
  @param clientResetMode Determines file recovery behavior in the event of a client reset.
                         See: https://docs.mongodb.com/realm/sync/error-handling/client-resets/
+ 
+ @return A default configuration object with the sync configuration set to use the given partition value.
```
tgoyne, line 118: We should avoid using the word `default` here as we have a concept of a "default configuration" that this doesn't interact with.

Realm/RLMUser.h
```diff
@@ -144,6 +163,42 @@ NS_ASSUME_NONNULL_BEGIN
  */
 - (RLMRealmConfiguration *)flexibleSyncConfiguration NS_REFINED_FOR_SWIFT;
 
+/**
+ Create a flexible sync configuration instance, which can be used to open a Realm that
+ supports flexible sync.
+ 
+ It won't possible to combine flexible and partition sync in the same app, which means if you open
+ a realm with a flexible sync configuration, you won't be able to open a realm with a PBS configuration
+ and the other way around.
```
tgoyne, line 172: "won't" is the wrong word here.

This paragraph is confusing because of how we've overloaded the word "app". You can use both PBS and FLX in the same app and they just need different server-side apps. It's also misleading because opening a Realm with a flexible sync configuration isn't what makes it invalid to open that same realm in partition-based mode.

RealmSwift/Sync.swift
tgoyne: Adding cases to an enum is already a breaking change because any code which switches on the enum will no longer work. As a result, returning a different enum case than we used to isn't really "more breaking".

Realm/RLMSyncUtil.mm
ericjordanmossman: Thanks. From what I searched there is no equivalent for swift :/

Realm/RLMSyncConfiguration.mm
ericjordanmossman: I tried a couple of things. There does have to be some kind of reference to self in order to use the ivar.
I mainly tried a strong self/weak self kind of thing like we do with the weakManager, but couldn't get it working. I'd pass a weak self into the lamdba, then try to make a strong reference before the weak self could possibly become null.

Is that the right track?

Realm/ObjectServerTests/SwiftSyncTestCase.swift
```diff
@@ -272,7 +278,7 @@ open class SwiftSyncTestCase: RLMSyncTestCase {
 #if swift(>=5.6) && canImport(_Concurrency)
 @available(macOS 12.0, *)
 extension SwiftSyncTestCase {
-    public func basicCredentials(usernameSuffix: String = "", app: App? = nil) async throws -> Credentials {
+    public func basicCredentials(usernameSuffix: String = "", app: RealmSwift.App? = nil) async throws -> Credentials {
```
dianaafanador3, line 281: I think after removing the SwiftUI dependency here, there is no need to RealmSwift before App

RealmSwift/Sync.swift
```diff
@@ -253,7 +252,166 @@ public enum ClientResetMode {
...
     ///  For more details on the second block: ((Realm, Realm) -> Void)? = nil,
     /// - see: `RLMClientResetAfterBlock`
+    @available(*, deprecated, message: "Use discardUnsyncedChanges")
     case discardLocal(((Realm) -> Void)? = nil, ((Realm, Realm) -> Void)? = nil)
+    /// All unsynchronized local changes are automatically discarded and the local state is
+    /// automatically reverted to the most recent state from the server. Unsynchronized changes
+    /// can then be recovered in the post-client-reset callback block.
+    ///
+    /// If `.discardUnsyncedChanges` is enabled but the client reset operation is unable to complete
+    /// then the client reset process reverts to manual mode. Example) During a destructive schema change this
```
dianaafanador3, line 262:
```suggestion
    /// then the client reset process reverts to manual mode. Example: During a destructive schema change this
```

RealmSwift/Sync.swift
```diff
@@ -895,8 +1068,18 @@ extension User {
 
      @return A `Realm.Configuration` instance with a flexible sync configuration.
      */
-    public func flexibleSyncConfiguration() -> Realm.Configuration {
-        let config = self.__flexibleSyncConfiguration()
+    public func flexibleSyncConfiguration(clientResetMode: ClientResetMode = .recoverUnsyncedChanges(nil, nil)) -> Realm.Configuration {
+        var config: RLMRealmConfiguration
+        switch clientResetMode {
```
dianaafanador3, line 1073: Can we extract this to a helper function given that this is done here, the flx initialiser with initial subscription  and in the PBS initialiser

Realm/ObjectServerTests/SwiftObjectServerTests.swift
```diff
@@ -596,37 +893,350 @@ class SwiftObjectServerTests: SwiftSyncTestCase {
...
+            waitForExpectations(timeout: 15.0)
+            XCTAssertEqual(realm.objects(SwiftPerson.self).count, 1)
+            // The Person created locally ("John") should have been discarded,
+            // while the one from the server ("Paul") should be present.
+            XCTAssertEqual(realm.objects(SwiftPerson.self)[0].firstName, "Paul")
+        }
+        try waitForEditRecoveryMode(flexibleSync: true, appId: self.flexibleSyncAppId, disable: false)
+    }
+
+    // A stray session is left during teardown. Can't find out where.
```
dianaafanador3, line 1180: Any reason for this to be commented?, can put fix_ to the test and a TODO if this needs to be fixed later

dianaafanador3 (commented):
Just a few comments, almost there

Realm/RLMUser.h
```diff
@@ -126,31 +128,88 @@ NS_ASSUME_NONNULL_BEGIN
...
+ @param partitionValue The `RLMBSON` value the Realm is partitioned on.
+ @param clientResetMode Determines file recovery behavior in the event of a client reset.
+                        See: https://docs.mongodb.com/realm/sync/error-handling/client-resets/
+ @param manualClientResetHandler An error reporting block that is invoked during a client reset.
+                                 See `RLMSyncErrorReportingBlock` and `RLMClientResetInfo`
+ 
+ @return A configuration object with the sync configuration set to use the given partition value.
+ */
+- (RLMRealmConfiguration *)configurationWithPartitionValue:(nullable id<RLMBSON>)partitionValue
+                                           clientResetMode:(RLMClientResetMode)clientResetMode
```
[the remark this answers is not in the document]
ericjordanmossman: An exception should be thrown in `setManualClientResetHandler`. I now added a test for this.

Realm/ObjectServerTests/SwiftObjectServerTests.swift
ericjordanmossman: No reason, I un-commented. 
Any ideas on stray session? I thought I needed to reset the sync manager in between opens, but resetting hits the commented exception.

RealmSwift/Sync.swift
ericjordanmossman: I think we need to use the mostly repetitive callsites because we need to extract the client reset blocks.

If we move them to the same method I don't think we'd save much repetition. Three different objc methods are called (w/ partition, w/ flex, w/ flex and initial subs). There'd still be 15 cases, each method with 5 modes. The only difference is that all 15 cases would be in one method.

CHANGELOG.md
```diff
@@ -1,15 +1,63 @@
...
-* None.
+
+* Add `.recoverUnsyncedChanges` (`RLMClientResetModeRecoverUnsyncedChanges`) and
+`.recoverOrDiscardUnsyncedChanges` (`RLMClientResetModeRecoverOrDiscardUnsyncedChanges`) behaviors to `ClientResetMode` (`RLMClientResetMode`).
+  - The newly added recover modes function by downloading a realm which reflects the latest
+    state of the server after a client reset. A recovery process is run locally in an
+    attempt to integrate the server state with any local changes from before the
+    client reset occurred.
+    The changes are integrated with the following rules:
+    1. Objects created locally not synced before client reset, will be integrated.
```
tgoyne, line 12: I think this sentence is missing a word or something.

Realm/ObjectServerTests/SwiftSyncTestCase.swift
```diff
@@ -79,9 +79,15 @@ open class SwiftSyncTestCase: RLMSyncTestCase {
 
     public func openRealm<T: BSON>(partitionValue: T,
                                    user: User,
+                                   clientResetMode: ClientResetMode? = nil,
```
tgoyne, line 82: Maybe just `clientResetMode: ClientResetMode = . recoverUnsyncedChanges` rather than dancing around with nil?

Realm/RLMRealmConfiguration.mm
```diff
@@ -66,6 +66,7 @@
 
 @implementation RLMRealmConfiguration {
     realm::Realm::Config _config;
+    RLMSyncConfiguration *_syncConfiguration;
```
tgoyne, line 69: If `manualClientResetHandler` is the only thing being used from the sync config we could just store that instead of the entire thing.

Realm/RLMSyncConfiguration.mm
```diff
@@ -270,32 +286,44 @@ - (BOOL)enableFlexibleSync {
...
+- (void)assignConfigErrorHandler:(RLMUser *)user {
     RLMSyncManager *manager = [user.app syncManager];
     __weak RLMSyncManager *weakManager = manager;
-    config.error_handler = [weakManager](std::shared_ptr<SyncSession> errored_session, SyncError error) {
+    RLMSyncErrorReportingBlock resetHandler = self.manualClientResetHandler;
+    _config->error_handler = [weakManager, resetHandler](std::shared_ptr<SyncSession> errored_session, SyncError error) {
         RLMSyncErrorReportingBlock errorHandler;
         @autoreleasepool {
             errorHandler = weakManager.errorHandler;
         }
```
tgoyne, line 297: I think consolidating the decision on which callback to use at the start makes this function a little simpler. It also avoids capturing the unused callback, which probably doesn't really matter.

```suggestion
    _config->error_handler = [weakManager, resetHandler](std::shared_ptr<SyncSession> errored_session, SyncError error) {
        RLMSyncErrorReportingBlock errorHandler;
        if (error.is_client_reset_requested()) {
            errorHandler = resetHandler;
        }
        if (!errorHandler) {
            @autoreleasepool {
                errorHandler = weakManager.errorHandler;
            }
        }
        if (!errorHandler) {
            return;
        }
```

Realm/RLMSyncConfiguration.mm
```diff
@@ -270,32 +286,44 @@ - (BOOL)enableFlexibleSync {
...
+            }
         });
     };
+};
+
+static void setDefaults(SyncConfig& config, RLMUser *user) {
+    config.client_resync_mode = ClientResyncMode::Recover;
+    config.stop_policy = SyncSessionStopPolicy::AfterChangesUploaded;
+
+    RLMSyncManager *manager = [user.app syncManager];
```
tgoyne, line 326: Appears to be unused?

Realm/RLMUser.h
```diff
@@ -126,31 +128,84 @@ NS_ASSUME_NONNULL_BEGIN
...
+
 /**
  Create a flexible sync configuration instance, which can be used to open a Realm that
  supports flexible sync.
-
- It won't possible to combine flexible and partition sync in the same app, which means if you open
- a realm with a flexible sync configuration, you won't be able to open a realm with a PBS configuration
- and the other way around.
+ 
+ @note A single server-side Device Sync App may connect to only flexible or partition-based realms.
```
tgoyne, line 158: The wording's off a bit here. Server-side apps don't connection to Realms.

Realm/ObjectServerTests/SwiftObjectServerTests.swift
```diff
@@ -596,39 +907,353 @@ class SwiftObjectServerTests: SwiftSyncTestCase {
...
+            }
+            try autoreleasepool {
+                let realm = try Realm(configuration: config)
+                waitForExpectations(timeout: 15.0)
+                // The locally created object should still be present as we didn't
+                // actually handle the client reset
+                XCTAssertEqual(realm.objects(SwiftPerson.self).count, 1)
+                XCTAssertEqual(realm.objects(SwiftPerson.self)[0].firstName, "John")
+            }
+//            resetSyncManager() Need to reset the syncManager. But path_for_realm is nil when creating new flex config?
```
tgoyne, line 1220: The session is torn down asynchronously after the last Realm instance is released. `syncManager.waitForSessionTermination()` waits for all sessions to finish being destroyed.

dianaafanador3 (commented):
Almost there

Realm/RLMSyncConfiguration.mm
ericjordanmossman: Used for the manager.authorizationHeaderName and manager.customRequestHeaders in 336 and 339.

dianaafanador3 (approved):
🎉

Realm/RLMSyncConfiguration.h
```diff
@@ -41,16 +43,76 @@ typedef NS_ENUM(NSUInteger, RLMClientResetMode) {
...
     /// can then be recovered in the post-client-reset callback block.
     ///
-    /// If RLMClientResetModeDiscardLocal is enabled but the client reset operation is unable to complete
+    /// If `RLMClientResetModeDiscardLocal` is enabled but the client reset operation is unable to complete
     /// then the client reset process reverts to manual mode. Example) During a destructive schema change this
     /// mode will fail and invoke the manual client reset handler.
-    RLMClientResetModeDiscardLocal
+    ///
+    /// The RLMClientResetModeDiscardLocal mode supports two client reset callbacks -- `RLMClientResetBeforeBlock`, `RLMClientResetAfterBlock` -- which can be passed as arguments when creating the `RLMSyncConfiguration`.
+    ///- see: `RLMClientResetAfterBlock` and `RLMClientResetBeforeBlock`
```
tgoyne, line 68: Some `- see`s which should be `@see`, and Jazzy needs a blank line before it to render well.

Linking to other documented symbols in DocC requires surrounding the name in double backticks. We haven't gone through and updated all the existing docs for this, but anything new should do it.

RealmSwift/Sync.swift
tgoyne: I think argument labels might be a good idea here. `case discardUnsyncedChanges(beforeReset: ((Realm) -> Void)? = nil, afterReset: ((Realm, Realm) -> Void)? = nil)` still isn't super clear but it's a bit more obvious?

Might also be better to spell out Synchronized? We do abbreviate it a lot, but using "synced" in a sentence is always kinda weird.

tgoyne:
Looks good to me other than those minor nits.

Realm/RLMSyncConfiguration.h
ericjordanmossman: Do these go for swift files also?

ericjordanmossman: The double back tick, that is.

tgoyne: yes

RealmSwift/Sync.swift
ericjordanmossman: I don't remember why, but there was a point in time I didn't think a typealias would work here. I was wrong. What about this?

public typealias beforeBlock = (_ beforeRealm: Realm) -> Void
public typealias afterBlock = (_ beforeRealm: Realm, _ afterRealm: Realm) -> Void
...
then
...
case discardUnsyncedChanges(beforeBlock? = nil, afterBlock? = nil)

That way I can pull out all the swift resetblock docs to one place, the typealias docstrings.

ericjordanmossman: Still need the underscore because function types cannot have argument labels.

ericjordanmossman: alternative names:

public typealias beforeResetBlock = (_ beforeResetRealm: Realm) -> Void
public typealias afterResetBlock = (_ beforeResetRealm: Realm, _ afterResetRealm: Realm) -> Void

ericjordanmossman: I'm going with 
case recoverUnsyncedChanges(beforeResetBlock? = nil, afterResetBlock? = nil)

tgoyne: No "Block". Argument/tuple labels in Swift shouldn't repeat the type of the thing they're labeling (unlike in objc).

ericjordanmossman: Okay if I understand correctly I'll go with:
public typealias beforeReset = (_ beforeReset: Realm) -> Void
public typealias afterReset = (_ beforeReset: Realm, _ afterResetRealm: Realm) -> Void

case recoverUnsyncedChanges(beforeReset? = nil, afterReset? = nil)

tgoyne: Oh, sorry, I don't mean that there should be typealiases here; there should be labels for the tuple elements. The problem is that at the place where the enum is being instantiated it's not obvious what the two functions are to someone reading it who isn't super familiar with the API, and having it be `. discardUnsyncedChanges(beforeReset: { realm in ...}, afterReset: { oldRealm, newRealm in ...})` would hopefully help with that.

RealmSwift/Sync.swift
```diff
@@ -209,18 +209,27 @@ public typealias Provider = RLMIdentityProvider
...
-    /// The first `.discardLocal` function argument notifies prior to a client reset occurring.
-    /// The `Realm` argument contains a frozen copy of the Realm state prior to client reset.
+    /// - parameter beforeReset: a function invoked prior to a client reset occurring.
+    /// - parameter before: a frozen copy of the local Realm state prior to client reset.
+    /// - seeAlso ``RLMClientResetBeforeBlock``
+
+    /// Example Usage
     /// ```
-    /// user.configuration(partitionValue: "myPartition", clientResetMode: .discardLocal({ beforeRealm in
+    /// user.configuration(partitionValue: "myPartition", clientResetMode: .discardUnsyncedChanges({ beforeRealm in
```
tgoyne, line 232: Usage examples need the argument labels added.

tgoyne (approved):

PR #7939, opened 2022-09-02
Add Xcode 14 binaries to the release package

tgoyne:
We now actually have tvos/watchos installed on the CI machines.

I also fixed all the pre-existing failing installation tests and now they all pass: https://ci.realm.io/blue/organizations/jenkins/cocoa-pipeline/detail/cocoa-pipeline/1790/pipeline/609. There were a few things that just got out of date; occasionally xcodebuild failed to connect to xctest and killing xctest first seems to fix that; and sometimes setting up the iOS simulator didn't work so I switched the evolution mode test to just testing macOS (which is a bit faster to do too).


leemaguire (approved):
LGTM

PR #7946, opened 2022-09-06
Update to core 12.6.0

tgoyne:
Most of the changes here are related to that core removed the error parameter for notifications (as they could never actually fail) and I updated the docs to reflect that. I also took advantage of that sectioned results haven't shipped yet to remove the error parameter from the public API for those.


tgoyne:
CI failure looks like a real core bug.

PR #7951, opened 2022-09-08
Rename authenticationDidCompleteWithError:

tgoyne:
Apple has banned the use of that selector name, so we have to pick a new name.

Fixes https://github.com/realm/realm-swift/issues/7945.


Infinity-James:
This doesn't build for me because of the following error:
`Cannot find type 'RealmSwiftAsymmetricObject' in scope`

PR #7960, opened 2022-09-16
Fix moving `List` items to a higher index in SwiftUI results in wrong  destination index.

dianaafanador3:
Fix moving `List` items to a higher index in SwiftUI results in wrong  destination index.
https://github.com/realm/realm-swift/issues/7956
Fixed some of the SwiftUI test, the other two are failing because the hardware keyboard is  enabled and is causing issues when typing on a staticText field.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,8 @@ x.y.z Release notes (yyyy-MM-dd)
 * None.
 
 ### Fixed
-* Fix moving `List` items to a higher index in SwiftUI results in wrong destination index ([#7956](https://github.com/realm/realm-swift/issues/7956), since v10.6.0).
+* Fix moving `List` items to a higher index in SwiftUI results in wrong destination index
+  ([#7956](https://github.com/realm/realm-swift/issues/7956), since v10.6.0).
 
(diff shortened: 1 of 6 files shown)


RealmSwift/SwiftUI.swift
```diff
@@ -1206,7 +1206,12 @@ public extension BoundCollection where Value == List<Element> {
     /// :nodoc:
     func move(fromOffsets offsets: IndexSet, toOffset destination: Int) {
         safeWrite(self.wrappedValue) { list in
-            list.move(fromOffsets: offsets, toOffset: destination)
+            var d = destination
+            if let firstOffset = offsets.first,
+               destination > firstOffset {
+                d = destination - 1
+            }
+            list.move(fromOffsets: offsets, toOffset: d)
```
tgoyne, line 1214: `list.move(fromOffsets:toOffset:)` is our function and not something built-in, so if it's doing the wrong thing then it should be fixed.

ericjordanmossman (approved):

RealmSwift/SwiftUI.swift
dianaafanador3: I moved the fix to list.move() @tgoyne

leemaguire (approved):

tgoyne (approved):

PR #7966, opened 2022-09-21
Check conformance to NSFastEnumeration before iterating

ericjordanmossman:
This will now check that a collection conforms to NSFastEnumeration before iterating in RLMObjectBase.
Previously a __SwiftValue could run into an unrecognized selector exception. RLMAsFastEnumeration checks that the collection conforms to NSFastEnumeration. If it does, it returns. If it doesn't, it bridges the value.

This should allow something like:
Screen Shot 2022-09-20 at 1 17 12 PM

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,8 +4,7 @@ x.y.z Release notes (yyyy-MM-dd)
 * None.
 
 ### Fixed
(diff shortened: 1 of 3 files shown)


tgoyne:
The Dictionary case above this looks like it needs the same thing.

tgoyne (approved):

PR #7981, opened 2022-10-03
Add custom column names API

dianaafanador3:
Add custom column names API, which allows to set a different column name in the realm 
   from the one used in your project.
   ```swift
   class Person: Object {
       @Persisted var firtName: String
       @Persisted var birthDate: Date
       @Persisted var age: Int

       override class public func propertiesMapping() -> [String : String] {
           ["firtName"; "first_name",
            "lastName"; "last_name"]
       }
   }
   ```
   This very helpful in cases where you want a property named differently from you `Device Sync` JSON schema.
   This will only affects the Swift API for old and modern declaration syntax.

This need this Core [PR](https://github.com/realm/realm-core/pull/5952) to be merged.
Some CI jobs are not passing because we are using a core branch.

Notes:

- This only can be merged when there is core release with the following commit https://github.com/realm/realm-core/commit/8e55fe842664ca99e467784db050369187debbcb.

- Objective-C tests  that use a key path using `LinkToRenamedProperties1` and `LinkToRenamedProperties2` were changed to use a model  that doesn't use `_realmObjectName`. `_realmObjectName` may retrieve the incorrect object schema if there are two objects named equally.


Realm/RLMSwiftValueStorage.mm
```diff
@@ -71,7 +71,7 @@ void attach(__unsafe_unretained RLMObjectBase *const obj, NSString *property) {
     ManagedSwiftValueStorage(RLMObjectBase *obj, RLMProperty *prop)
     : _realm(obj->_realm)
     , _object(obj->_realm->_realm, *obj->_info->objectSchema, obj->_row)
-    , _propertyName(prop.name.UTF8String)
+    , _propertyName(prop.columnName.UTF8String)
```
tgoyne, line 74: Should rename this to `_columnName` for clarity.

Realm/RLMProperty.mm
```diff
@@ -722,6 +724,9 @@ - (NSString *)columnName {
 - (realm::Property)objectStoreCopy:(RLMSchema *)schema {
     realm::Property p;
     p.name = self.columnName.UTF8String;
+    if (self.columnName != _name) {
```
tgoyne, line 727: I think `if (_columnName)` is clearer here. Comparing NSStrings by pointer is something that inherently looks very suspicious, even though it is functionally correct here.

RealmSwift/AsymmetricObject.swift
```diff
@@ -85,6 +85,40 @@ extension AsymmetricObject {
...
+         override class public func propertiesMapping() -> [String : String] {
+             ["firtName"; "first_name",
+              "lastName"; "last_name"]
+         }
+     }
+     ```
+
+     - note: Only property that have a different column name have to be added to the properties mapping
+     dictionary.
+     - note: This API is enable for both old and modern property declaration syntax only.
```
tgoyne, line 112: I don't understand this sentence. It's only enabled for both of the ways you can define properties?

RealmSwift/AsymmetricObject.swift
```diff
@@ -85,6 +85,40 @@ extension AsymmetricObject {
...
     }
 
+    // MARK: Object Customization
+
+    /**
+     Override this method to specify a map of public-private property names.
+     This will set a different property column name on the Realm, and allows using the public name
+     for any operation with the property. (Ex: Queries, Sorting, ...).
+     This very helpful if you need to map property names from your `Device Sync` JSON schema
+     to local property names.
```
tgoyne, line 95: I think we want to avoid using the word "column" in anything public. .NET mentions a "persisted property name", while JS uses "underlying".

RealmSwift/Impl/SchemaDiscovery.swift
```diff
@@ -62,14 +62,16 @@ internal extension RLMProperty {
     }
 }
 
-private func getModernProperties(_ object: ObjectBase) -> [RLMProperty] {
+private func getModernProperties(_ object: ObjectBase, _ cls: ObjectBase.Type) -> [RLMProperty] {
```
tgoyne, line 65: New parameter appears to be unused.

Realm/RLMObservation.mm
```diff
@@ -571,15 +571,15 @@ static KeyPath keyPathFromString(RLMRealm *realm,
         TableKey tk = info->objectSchema->table_key;
         ColKey ck;
         if (property.type == RLMPropertyTypeObject) {
-            ck = info->tableColumn(property.columnName);
+            ck = info->tableColumn(property.name);
```
tgoyne, line 574: I think this change is correct, but it suggests we're missing some KVO tests that should be added.

RealmSwift/Tests/PrimitiveMapTests.swift
```diff
@@ -346,6 +346,7 @@ class ManagedPrimitiveMapTests: TestCase {
         SortablePrimitiveMapTests<ManagedObjectFactory, Double>.defaultTestSuite.tests.forEach(suite.addTest)
         SortablePrimitiveMapTests<ManagedObjectFactory, String>.defaultTestSuite.tests.forEach(suite.addTest)
         SortablePrimitiveMapTests<ManagedObjectFactory, Date>.defaultTestSuite.tests.forEach(suite.addTest)
+        SortablePrimitiveListTests<ManagedObjectFactory, Data>.defaultTestSuite.tests.forEach(suite.addTest)
```
tgoyne, line 349: Wrong test class name here and in Set tests?

Realm/RLMProperty.mm
dianaafanador3: Thanks for the tip

leemaguire (commented):
Few small comments, but this is looking good.

Realm/RLMObservation.mm
dianaafanador3: Added Test for this in objective-C, and there were some already in Swift in the new ColumnNameTest suite

tgoyne (approved):
LGTM once we have a proper core release.

RealmSwift/AsymmetricObject.swift
```diff
@@ -85,6 +85,39 @@ extension AsymmetricObject {
...
+         @Persisted var age: Int
+
+         override class public func propertiesMapping() -> [String : String] {
+             ["firstName"; "first_name",
+              "birthDate"; "birth_date"]
+         }
+     }
+     ```
+
+     - note: Only property that have a different column name have to be added to the properties mapping
```
leemaguire, line 110:
```suggestion
     - note: Only properties that have a different column name must be added to the properties mapping
```

leemaguire (approved):
Small comment, LGTM once Core release is ready.

PR #7993, opened 2022-10-17
Fix a pile of issues with sync/app error reporting

tgoyne:
Many sync and app errors had error codes from `RLMSyncSystemErrorKind` (a private enum), `realm::ProtocolError`, `realm::ServiceErrorCode` or `realm::ClientErrorCode` (not exposed in the SDK at all), and often add error domains which were just arbitrary strings from core. This meant that most of the errors weren't actually handleable without relying on internal implementation details. This changes it so that (almost) all errors have a well-defined domain+error code pair, and adds many missing error codes to make that possible.

RLMAppError is a subset of the error codes in `ServiceErrorCode` which includes only the ones which can actually be hit via our API. There's a bunch which are only applicable to the admin API, or not at all (e.g. GCM related things, which Google shut down a few years ago).

`Realm.Error` predates `NS_ERROR_ENUM` and manually defines exactly what the automatic bridging now does, so change it to just be a typealias.

All of the tests which assert that an error occurred now also validate that it's the expected error. In the process I switched many of the tests over to using the Combine versions of the App APIs as it makes basic operations like logging in one line instead of 5+, and makes it easier to verify that every single operation is actually checking appropriately for success/failure.

`ServerValidationPolicy` and `RLMSyncAuthError ` have been unused since v10 and should have been deleted then.

The sectioned results tests created `default.realm` but didn't clean it up, which made some other tests fail depending on the order the test cases run. I switched them over to using an in-memory realm to solve this.

### DIFF
--- a/Realm/ObjectServerTests/RLMFlexibleSyncServerTests.mm
+++ b/Realm/ObjectServerTests/RLMFlexibleSyncServerTests.mm
@@ -901,15 +901,19 @@
     }];
     CHECK_COUNT(3, Person, realm);
 
-    auto ex = [self expectationWithDescription:@"should client reset"];
-    [self.flexibleSyncApp syncManager].errorHandler = ^(NSError * error, RLMSyncSession *) {
-        XCTAssertNotNil(error);
+    RLMObjectId *invalidObjectPK = [RLMObjectId objectId];
+    auto ex = [self expectationWithDescription:@"should revert write"];
+    self.flexibleSyncApp.syncManager.errorHandler = ^(NSError *error, RLMSyncSession *) {
+        RLMValidateError(error, RLMSyncErrorDomain, RLMSyncErrorWriteRejected,
+                         @"Client attempted a write that is outside of permissions or query filters; it has been reverted");
         [ex fulfill];
     };
     [realm transactionWithBlock:^{
-        [realm addObject:[[Person alloc] initWithPrimaryKey:[RLMObjectId objectId] age:10 firstName:@"Nic" lastName:@"Cages"]];
+        [realm addObject:[[Person alloc] initWithPrimaryKey:invalidObjectPK age:10 firstName:@"Nic" lastName:@"Cages"]];
     }];
     [self waitForExpectations:@[ex] timeout:20];
+    [self waitForDownloadsForRealm:realm];
+    CHECK_COUNT(3, Person, realm);
 }
 
 - (void)testFlexibleSyncInitialSubscription {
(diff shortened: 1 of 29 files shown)


ericjordanmossman (approved):
Awesome. Thanks for refactoring a ton of realm server testing methods too.

Realm/TestUtils/include/RLMAssertions.h
```diff
@@ -118,6 +118,44 @@ FOUNDATION_EXTERN bool RLMHasCachedRealmForPath(NSString *path);
     XCTAssertEqual([exception.userInfo[NSUnderlyingErrorKey] code], expectedCode, __VA_ARGS__); \
 })
 
+#define RLMValidateError(error, errDomain, errCode, msg) do {                                           \
```
dianaafanador3, line 121: What are the advantages of defining this as a macro than as a function? in TestCases for example.

dianaafanador3 (approved):
LGTM, Just a few small comments

Realm/ObjectServerTests/RLMFlexibleSyncServerTests.mm
```diff
@@ -901,15 +901,19 @@ - (void)testFlexibleSyncAddObjectOutsideQuery {
     }];
     CHECK_COUNT(3, Person, realm);
 
-    auto ex = [self expectationWithDescription:@"should client reset"];
-    [self.flexibleSyncApp syncManager].errorHandler = ^(NSError * error, RLMSyncSession *) {
-        XCTAssertNotNil(error);
+    RLMObjectId *invalidObjectPK = [RLMObjectId objectId];
```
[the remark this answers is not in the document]
tgoyne: In the exception unification branch this test also checks that the correct object PK is reported in the error's extended info. Here it's pointless but I was trying to minimize merge conflicts while splitting this off.

Realm/TestUtils/include/RLMAssertions.h
tgoyne: These are macros so that they can report the failing line as the line containing the `RLMValidateError()` call rather than the line within the test function which made the failing assertion. We can use regular functions in Swift because the Swift version of `XCTAssert()` is itself just a function which takes a line number as an argument, so we can pass that along explicitly. In obj-c `XCTAssert()` is a macro which uses `__LINE__` internally, so anything which wraps it must also be macros for `__LINE__` and `__FILE__` to have the correct values.

dianaafanador3: Ohh that's good to know, thanks

PR #8022, opened 2022-11-08
Fix issue with UserAPIKey.key returning incorrect value

leemaguire:
Fixes: https://github.com/realm/realm-swift/issues/8021

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,8 @@ x.y.z Release notes (yyyy-MM-dd)
 * The prebuilt watchOS library in the objective-c release package was missing
   an arm64 slice. The Swift release package was uneffected
   ([PR #8016](https://github.com/realm/realm-swift/pull/8016)).
+* Fix issue where `RLMUserAPIKey.key`/`UserAPIKey.key` incorrectly returned the name of the API 
+  key instead of the key itself. ([#????](https://github.com/realm/realm-swift/issues/8021), since v10.0.0)
 
 <!-- ### Breaking Changes - ONLY INCLUDE FOR NEW MAJOR version -->
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,7 @@ x.y.z Release notes (yyyy-MM-dd)
   an arm64 slice. The Swift release package was uneffected
   ([PR #8016](https://github.com/realm/realm-swift/pull/8016)).
 * Fix issue where `RLMUserAPIKey.key`/`UserAPIKey.key` incorrectly returned the name of the API 
-  key instead of the key itself. ([#????](https://github.com/realm/realm-swift/issues/8021), since v10.0.0)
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -10,6 +10,8 @@ x.y.z Release notes (yyyy-MM-dd)
 * The prebuilt watchOS library in the objective-c release package was missing
   an arm64 slice. The Swift release package was uneffected
   ([PR #8016](https://github.com/realm/realm-swift/pull/8016)).
+* Fix issue where `RLMUserAPIKey.key`/`UserAPIKey.key` incorrectly returned the name of the API 
+  key instead of the key itself. ([#????](https://github.com/realm/realm-swift/issues/8021), since v10.0.0)
```
dianaafanador3, line 14:
```suggestion
  key instead of the key itself. ([#8021](https://github.com/realm/realm-swift/issues/8021), since v10.0.0)
```

Realm/RLMUserAPIKey.mm
```diff
@@ -51,7 +51,7 @@ - (NSString *)name {
 // the response when an API key is first created.
```
dianaafanador3, line 51: I'm a little curious about this text over here, this will only return only when we call the `create_api_key` object store API, and will return nil when we fetch `fetch_api_key` and `fetch_api_keys`

Realm/ObjectServerTests/SwiftObjectServerTests.swift
```diff
@@ -1767,6 +1767,10 @@ class SwiftObjectServerTests: SwiftSyncTestCase {
         let syncUser = app.login(credentials: credentials).await(self)
 
         let apiKey = syncUser.apiKeysAuth.createAPIKey(named: "my-api-key").await(self)
+        XCTAssertEqual(apiKey.name, "my-api-key")
+        XCTAssertNotNil(apiKey.key)
+        XCTAssertNotEqual(apiKey.key!, "my-api-key")
+        XCTAssertFalse(apiKey.key!.isEmpty)
 
         syncUser.apiKeysAuth.fetchAPIKey(apiKey.objectId).await(self)
```
dianaafanador3, line 1775: Related to the other comment, but If fetchApiKey return the key as well, can we check it here, that it is in fact the same that returns the create API

dianaafanador3:
Would be interesting to test againts the admin API if the server API key is the same as the one we receive in our end, but because this is a short fix is not necessary and probably something that will be better to be tested in core.

dianaafanador3 (approved):
Nice quick bug fix!!!. LGTM, just a small change and some questions.

Realm/RLMUserAPIKey.mm
matthewmorek: This is intentional, as the API key is not supposed to be revealed outside of the `create` action, as it's a secret. See docs here: https://www.mongodb.com/docs/realm/sdk/swift/app-services/manage-user-api-keys/

dianaafanador3: Thanks @matthewmorek for the info.

PR #8034, opened 2022-11-16
Add progress callback to asynchronous `async await` initialiser for Realm

dianaafanador3:
This is extending the API for `Realm.init()` async to have a callback for progress notifications. Feature request here https://github.com/realm/realm-swift/issues/7763.

### DIFF
--- a/RealmSwift/Realm.swift
+++ b/RealmSwift/Realm.swift
@@ -1206,7 +1206,8 @@ extension Realm {
      */
     @MainActor
     public init(configuration: Realm.Configuration = .defaultConfiguration,
-                downloadBeforeOpen: OpenBehavior = .never) async throws {
+                downloadBeforeOpen: OpenBehavior = .never,
+                progress block: ((SyncSession.Progress) -> Void)? = nil) async throws {
         var rlmRealm: RLMRealm?
         switch downloadBeforeOpen {
         case .never:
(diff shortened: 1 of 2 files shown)


RealmSwift/Realm.swift
```diff
@@ -1217,13 +1218,18 @@ extension Realm {
             }
         case .always:
             rlmRealm = try await withCheckedThrowingContinuation { continuation in
-                RLMRealm.asyncOpen(with: configuration.rlmConfiguration, callbackQueue: .main) { (realm, error) in
+                let rlmTask = RLMRealm.asyncOpen(with: configuration.rlmConfiguration, callbackQueue: .main) { (realm, error) in
```
rowungiles, line 1221: Hey @dianaafanador3 thanks for the PR. Looking good.

What was your reasoning for fixing the callbackQueue to `.main` rather than letting an implementer pass in their preffered queue?

dianaafanador3: Because you can only run `let realm = try await Realm()` at the moment, in a @MainActor. 
We are still discussing if this is the bet API for this, but we have a project in progress which should give better support for actor-confined Realms and offer a better experience with Swift concurrency in general, which should include exposing recurring callbacks as `AsyncSequences` for sync progress notification.

rowungiles: That's great, thanks @dianaafanador3

PR #8038, opened 2022-11-18
Eliminate some redundant updates in SwiftUI components

tgoyne:
While adding Sendable annotations to the SwiftUI stuff I extracted some duplicated code and fixed some minor issues. Updating the search text for searchable triggered two updates because it modified both `filter` and `where`, which each triggered an update. The `where` property on the storage appears to be just totally redundant.

Changing anything on ObservableResults would trigger up to three updates due to storing temporary values in the `@StateObject` property.

AsyncOpen and AutoOpen would start an async open twice if both the partitionValue and configuration environment keys were set.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,11 +32,13 @@ x.y.z Release notes (yyyy-MM-dd)
   terminated at a certain point in the process of comitting a write
   transaciton. ([Core PR #5993](https://github.com/realm/realm-core/pull/5993), since v10.21.1)
 * `@AsyncOpen` and `@AutoOpen` would begin and then cancel a second async open
-  operation (since v10.12.0).
+  operation ([PR #8038](https://github.com/realm/realm-swift/pull/8038), since v10.12.0).
 * Changing the search text when using the searchable SwiftUI extension would
-  trigger multiple updates on the View for each change (since v10.19.0).
-* Changing the filter or search properties of an `@ObservedResults` would
-  trigger up to three updates on the View (since v10.6.0).
+  trigger multiple updates on the View for each change
+  ([PR #8038](https://github.com/realm/realm-swift/pull/8038), since v10.19.0).
+* Changing the filter or search properties of an `@ObservedResults` or
+  `@ObservedSectionedResults` would trigger up to three updates on the View
+  ([PR #8038](https://github.com/realm/realm-swift/pull/8038), since v10.6.0).
 
 <!-- ### Breaking Changes - ONLY INCLUDE FOR NEW MAJOR version -->
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,22 +9,34 @@ x.y.z Release notes (yyyy-MM-dd)
       @Persisted var firstName: String
       @Persisted var birthDate: Date
       @Persisted var age: Int
-      
+
       override class public func propertiesMapping() -> [String : String] {
           ["firstName"; "first_name",
            "birthDate"; "birth_date"]
       }
   }
   ```
-  This is very helpful in cases where you want to name a property differently 
+  This is very helpful in cases where you want to name a property differently
   from your `Device Sync` JSON schema.
   This API is only available for old and modern object declaration syntax on the
   `RealmSwift` SDK.
+* Flexible sync bootstraps now apply 1MB of changesets per write transaction
+  rather than applying all of them in a single write transaction.
+  ([Core PR #5999](https://github.com/realm/realm-core/pull/5999)).
 
 ### Fixed
-* Fix a race condition which could result in "operation cancelled" errors being delivered to async open callbacks rather than the actual sync error which caused things to fail ([PR #5968](https://github.com/realm/realm-core/pull/5968), since the introduction of async open).
-* Bootstraps will not be applied in a single write transaction - they will be applied 1MB of changesets at a time, or as configured by the SDK ([#5999](https://github.com/realm/realm-core/pull/5999), since v10.27.0).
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -9,22 +9,34 @@ x.y.z Release notes (yyyy-MM-dd)
...
-* Bootstraps will not be applied in a single write transaction - they will be applied 1MB of changesets at a time, or as configured by the SDK ([#5999](https://github.com/realm/realm-core/pull/5999), since v10.27.0).
-* Fix database corruption and encryption issues on apple platforms, reported in several bugs listed in the PR. ([PR #5993](https://github.com/realm/realm-core/pull/5993), since v10.21.1)
+* Fix a race condition which could result in "operation cancelled" errors being
+  delivered to async open callbacks rather than the actual sync error which
+  caused things to fail ([Core PR #5968](https://github.com/realm/realm-core/pull/5968), since the introduction of async open).
+* Fix database corruption issues which could happen if an application was
+  terminated at a certain point in the process of comitting a write
+  transaciton. ([Core PR #5993](https://github.com/realm/realm-core/pull/5993), since v10.21.1)
+* `@AsyncOpen` and `@AutoOpen` would begin and then cancel a second async open
```
dianaafanador3, line 34: Should we link the PR for this three fixes.

RealmSwift/SwiftUI.swift
```diff
@@ -265,6 +265,55 @@ private class ObservableStorage<ObservedType>: ObservableObject where ObservedTy
...
+private class ObservableResultsStorage<T>: ObservableStorage<T> where T: RealmSubscribable & ThreadConfined & Equatable {
+    var setupHasRun = false
+    func didSet() {
+        if setupHasRun {
+            setupValue()
+            setupHasRun = true
+        }
+    }
+
+    func setupValue() {
```
dianaafanador3, line 278: I would probably add a message or comment so we know this is been override by the storage on each property wrappers

RealmSwift/SwiftUI.swift
```diff
@@ -421,90 +470,42 @@ extension Projection: _ObservedResultsValue { }
...
-            }
-        }
-
-        func setupValue() {
+    private class Storage: ObservableResultsStorage<Results<ResultType>> {
+        override func setupValue() {
             /// A base value to reset the state of the query if a user reassigns the `filter` or `sortDescriptor`
             let realm = try! Realm(configuration: configuration ?? Realm.Configuration.defaultConfiguration)
-            value = realm.objects(ResultType.self)
+            var value = realm.objects(ResultType.self)
```
dianaafanador3, line 477: Nice catch on this

CHANGELOG.md
```diff
@@ -9,22 +9,34 @@ x.y.z Release notes (yyyy-MM-dd)
...
+  operation (since v10.12.0).
+* Changing the search text when using the searchable SwiftUI extension would
+  trigger multiple updates on the View for each change (since v10.19.0).
+* Changing the filter or search properties of an `@ObservedResults` would
```
dianaafanador3, line 38: I would probably mention this fix also applies to @ObservedSectionedResults

RealmSwift/SwiftUI.swift
```diff
@@ -265,6 +265,55 @@ private class ObservableStorage<ObservedType>: ObservableObject where ObservedTy
...
+            setupValue()
+            setupHasRun = true
+        }
+    }
+
+    func setupValue() {
+        fatalError()
+    }
+
+    var sortDescriptor: SortDescriptor? {
```
dianaafanador3, line 282: Maybe it is too much work, but now that we are unifying storage for @ObservedResults and @ObservedSectionResults, there seems to be a request for extending @ObservedResults API to allow more than one sortDescriptor, https://github.com/realm/realm-swift/issues/7709,

dianaafanador3 (approved):
Nice PR, This would make our SwiftUI implementation more efficient. Just some small comments

RealmSwift/SwiftUI.swift
```diff
@@ -651,47 +645,21 @@ extension Projection: _ObservedResultsValue { }
             }
 
             value = results.sectioned(sortDescriptors: sortDescriptors, sectionBlock)
-
-            setupHasRun = true
```
dianaafanador3, line 655: Removing this and move it to `didSet` results in this function been called endlessly from `wrappedValue`. This also applies to `@ObservedSectionedResults`. I tested this on a project and restoring this and removing it from didSet solves the issue

dianaafanador3 (changes requested):
I tested this PR with a project and the setupValue function is been called recurrently. This is prob

RealmSwift/SwiftUI.swift
tgoyne: Good catch; I accidentally made it so that the only place `setupHasRun = true` was set was inside a `if setupHasRun { }` block. I don't see an obvious way to catch this from a test because I think it still always gave correct results and was just a lot slower than it should be.

RealmSwift/SwiftUI.swift
tgoyne: It looks easy to implement but not so entirely trivial that I might as well just do it here.

dianaafanador3: @tgoyne If you say you'll implement it here?, I'll wait for the final review

tgoyne: No, I won't be implementing it here. I put a few too many conditionals in that sentence.

dianaafanador3 (approved):

PR #8039, opened 2022-11-22
Check initWithData creates valid string

ericjordanmossman:
Declares an NSString instead of a std::stringview and adds a check that the allocated str is valid. Previously could crash by calling .empty() on an invalid str. See internal [ticket.](https://jira.mongodb.org/browse/HELP-38960)


tgoyne:
This code just doesn't really make any sense to begin with. Why is it converting to a NSString and back at all? Presumably it should be something like:

```
    if (_watchStream.state() == realm::app::WatchStream::State::NEED_DATA) {
        [event enumerateByteRangesUsingBlock:^(const void *bytes, NSRange byteRange, BOOL *) {
            _watchStream.feed_buffer(std::string_view(static_cast<const char *>(bytes), byteRange.length));
        }];
    }
```

CHANGELOG.md
```diff
@@ -25,6 +25,7 @@ x.y.z Release notes (yyyy-MM-dd)
 * Fix a race condition which could result in "operation cancelled" errors being delivered to async open callbacks rather than the actual sync error which caused things to fail ([PR #5968](https://github.com/realm/realm-core/pull/5968), since the introduction of async open).
 * Bootstraps will not be applied in a single write transaction - they will be applied 1MB of changesets at a time, or as configured by the SDK ([#5999](https://github.com/realm/realm-core/pull/5999), since v10.27.0).
 * Fix database corruption and encryption issues on apple platforms, reported in several bugs listed in the PR. ([PR #5993](https://github.com/realm/realm-core/pull/5993), since v10.21.1)
+* Add extra check when allocating string when receiving data in a changestream [PR #8039](https://github.com/realm/realm-swift/pull/8039).
```
bmunkholm, line 28: Can we reformulate this from the perspective of users? How would they experience this bug?

tgoyne:
Bad merge on the changelog.

tgoyne (approved):

PR #8050, opened 2022-12-06
Fix concurrency warnings in Projection and related types

tgoyne:
Concurrency checking wants us to use `withLock()` to guard access to the schema cache, and requires marking the schema cache as unchecked sendable. While refactoring to support this I also fixed an unrelated problem (object change notifications were incorrect if a property was projected more than one time) and cut down on some redundant object allocations. The schema lock is a moderately hot path since we hit it every time we allocate an object, so I made it use the faster os_unfair_lock when possible.

The changes to unrelated files are to make the latest version of swiftlint happy.

### DIFF
--- a/.swiftlint.yml
+++ b/.swiftlint.yml
@@ -7,41 +7,16 @@ included:
   - examples/installation/ios/swift
   - examples/ios/swift
   - examples/tvos/swift
+type_name:
+  allowed_symbols:
+    - _
 identifier_name:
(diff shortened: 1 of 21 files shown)


dianaafanador3 (approved):
Very nice catch with the notification fix, and very interesting implementation of the cache lock. I guess, this prepares the code for a Projection property to be sendable, and this will be implemented in another PR?

PR #8051, opened 2022-12-07
Add Sendable annotations to the obj-c API

tgoyne:
This is mostly just a giant mess of boilerplate changes. NS_HEADER_AUDIT_BEGIN/NS_HEADER_AUDIT_END is a generalization of NS_ASSUME_NONNULL_BEGIN/NS_ASSUME_NONNULL_END to enable also marking things as audited for sendability, which changes the rules for warning generation for unannotated things a bit. It was added in Xcode 14, so we need to use our own define which falls back to the old thing with Xcode 13.

Types which are thread-safe are marked with NS_SWIFT_SENDABLE, which is equivalent to marking them as Sendable in Swift. I had to fiddle with RLMSetSwiftBridgeCallback as assigning to a global variable pretty reasonably produces concurrency warnings.

### DIFF
--- a/Realm/RLMUtil.mm
+++ b/Realm/RLMUtil.mm
@@ -73,6 +73,7 @@ static inline bool checkCollectionType(__unsafe_unretained id<RLMCollection> con
         && (type != RLMPropertyTypeObject || [collection.objectClassName isEqualToString:objectClassName]);
 }
 
+static id (*s_bridgeValue)(id);
 id<NSFastEnumeration> RLMAsFastEnumeration(__unsafe_unretained id obj) {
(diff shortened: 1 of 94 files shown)


Realm/RLMUtil.mm
```diff
@@ -18,6 +18,7 @@
 
 #import "RLMUtil.hpp"
 
+#import "RLMApp_Private.hpp"
```
dianaafanador3, line 21: Is this import really needed here?

dianaafanador3 (approved):
LGTM, Just some small comments

PR #8054, opened 2022-12-08
Update swiftui tests for iOS 16.1

tgoyne:
Sectioned table views are now implemented using a collection view rather than a table so those need to be updated. A few behavioral changes in 16.0 appear to have reverted back to the pre-16 behavior as well.

### DIFF
--- a/Realm/Tests/SwiftUITestHostUITests/SwiftUITestHostUITests.swift
+++ b/Realm/Tests/SwiftUITestHostUITests/SwiftUITestHostUITests.swift
@@ -60,6 +60,19 @@ class SwiftUITests: XCTestCase {
         }
     }
 
+    private func checkRowCount(_ table: XCUIElement, _ expected: Int, line: UInt = #line) {
+        let count = table.cells.count
+        guard table.cells.count != expected else { return }
+        XCTAssertGreaterThanOrEqual(expected, count, line: line)
+
+        // Cells are lazily loaded as needed, so any offscreen ones don't exist.
+        // We have fewer cells than expected, so verify that the final cell
+        // extends past the bottom of the screen.
+        let lastCell = table.cells.element(boundBy: count - 1)
+        XCTAssertGreaterThan(lastCell.frame.origin.y + lastCell.frame.height,
+                             table.frame.height)
+    }
+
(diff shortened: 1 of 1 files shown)


dianaafanador3:
Very nice, finally fixing this and get a green check in the CI. 
Two things, seems like the matching for `app.buttons.matching(identifier: "Reorder").` is not working in the latest iOS releases.
Also, I tested this in an iPhone 8 which seems to be the chosen simulator for running this tests and some table cells count are not equal to the one in the test, because of the screen size.
I'm testing on an iPhone 8 (16.1).

dianaafanador3 (commented):
Very nice, finally fixing this and get a green check in the CI.
Two things, seems like the matching for app.buttons.matching(identifier: "Reorder"). is not working in the latest iOS releases.
Also, I tested this in an iPhone 8 which seems to be the chosen simulator for running this tests and some table cells count are not equal to the one in the test, because of the screen size.
I'm testing on an iPhone 8 (16.1).

tgoyne:
The cell count depending on screen size is super awkward. I guess I need to make it check screen size and adjust the expected amount based on that. Another option might be to use a really small text size for the test app size so that they all fit as it doesn't actually need to be usable.

tgoyne:
I've updated the tests to hopefully be screen-size independent.

Realm/Tests/SwiftUITestHostUITests/SwiftUITestHostUITests.swift
```diff
@@ -60,6 +60,19 @@ class SwiftUITests: XCTestCase {
         }
     }
 
+    private func checkRowCount(_ table: XCUIElement, _ expected: Int, line: UInt = #line) {
```
dianaafanador3, line 63: Would be possible to use this function to check all the row counts in all UI tests. Checking if the count is equal to the expected count in cases the cells fit in the screen and if not check the table frame height with the  lastCel origin + height like you do here
```
let lastCell = table.cells.element(boundBy: count - 1)
XCTAssertGreaterThan(lastCell.frame.origin.y + lastCell.frame.height, table.frame.height)
```

Issue #7908, opened 2022-08-11
SwiftUI warning - publishing changes within a view (Xcode b5)


Sonisan:
### How frequently does the bug occur?

All the time

### Description

Using Xcode beta 5 triggers the following SwiftUI warnings:

>  [SwiftUI] Publishing changes from within view updates is not allowed, this will cause undefined behavior.

No warning was present in the previous beta 3.
My project uses Realm Sync with the `@AutoOpen` property wrapper. Besides the warning, the code is working as intended.

### Stacktrace & log output

```shell
Warnings lead to the following block of code:

...
 @Published var asyncOpenState: AsyncOpenState = .connecting {
        willSet {
            objectWillChange.send() // <--- HERE
        }
    }

...
 private func asyncOpenForUser(_ user: User) {
        asyncOpenState = .connecting // <--- HERE

....
```

### Can you reproduce the bug?

Yes, always

### Reproduction Steps

Launch the complete code example from the SwiftUI quickstart on https://www.mongodb.com/docs/realm/sdk/swift/swiftui-tutorial/#complete-code

### Version

10.28.5

### What SDK flavour are you using?

MongoDB Realm (i.e. Sync, auth, functions)

### Are you using encryption?

No, not using encryption

### Platform OS and version(s)

macOS 12.5

### Build environment

Xcode version: 14.0 beta 5

QuietRocket:
I am experiencing the same warning when using `@ObservedResults`. Also seems to have no effect on the resulting program, but mentioning "undefined behavior" is slightly worrisome.

The code causing the warning is line 205 in ObservableStoragePublisher.
```swift
    func send() {
        subscribers.forEach {
            _ = $0.receive() // <- This one
        }
    }
```

It might be worth noting that I am not using any sync features, just a local database.

drmarkpowell:
me too +1

drmarkpowell:
Still happening Xcode 14 beta 6, RealmSwift 10.28.6 macOS Monterey 12.5

Sonisan:
In my case, it seemed to be tied to `@AutoOpen` and `@AsyncOpen` property wrappers. I refactored my authentication flow without them (using `let realm = try await Realm(configuration: configuration`) and the warnings disappeared.

QuietRocket:
Hello. This issue was opened over a month ago. Will it be addressed?

[...]

PR #8055, opened 2022-12-08
Don't emit objectWillChange when an observed Results is first initialized

tgoyne:
This fixes ones of the causes of the "Publishing changes from within view updates is not allowed" warning, which I stumbled over while trying to debug something unrelated. #7908 mentions this happening with several other property wrappers as well, and the fix is probably to do something similar to this for all of them.

### DIFF
--- a/RealmSwift/SwiftUI.swift
+++ b/RealmSwift/SwiftUI.swift
@@ -232,16 +232,7 @@ private final class ObservableStoragePublisher<ObjectType>: Publisher where Obje
 
 @available(iOS 13.0, macOS 10.15, tvOS 13.0, watchOS 6.0, *)
 private class ObservableStorage<ObservedType>: ObservableObject where ObservedType: RealmSubscribable & ThreadConfined & Equatable {
-    @Published var value: ObservedType {
-        willSet {
-            if newValue != value {
-                objectWillChange.subscribers.forEach {
-                    $0.receive(subscription: ObservationSubscription(token: newValue._observe(keyPaths, $0)))
-                }
-                objectWillChange.send()
-            }
-        }
-    }
+    @Published var value: ObservedType
 
     let objectWillChange: ObservableStoragePublisher<ObservedType>
     let keyPaths: [String]?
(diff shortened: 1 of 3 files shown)


RealmSwift/SwiftUI.swift
```diff
@@ -232,16 +232,7 @@ private final class ObservableStoragePublisher<ObjectType>: Publisher where Obje
...
-        willSet {
-            if newValue != value {
-                objectWillChange.subscribers.forEach {
-                    $0.receive(subscription: ObservationSubscription(token: newValue._observe(keyPaths, $0)))
-                }
-                objectWillChange.send()
-            }
-        }
-    }
+    @Published var value: ObservedType
```
dianaafanador3, line 235: Given that `ObservableStorage` is not used exclusively by `@ObservedResults` and `@ObservedSectionedResults`, wouldn't this change affect also `@StateRealmObject` and  `@ObservableRealmObject` in cases when the value is updated and we need to update the subscription. For example
``` swift
struct DetailTicketView: View {
    @StateRealmObject var ticket: Ticket = Ticket()
    @Environment(\.realm) var realm

    var body: some View {
        VStack {
            TextField("Title", text: $ticket.title)
                .font(.headline)
        }
        .onAppear {
            ticket = realm.objects(Ticket.self).first!
        }
    }
}
``` I don’t know if this is a common use case, but may be affecting users that use `@StateRealmObject` as a draft object. I also tested things and the behaviour changes with your solution.  I’ve been investigating this and changing the order of calling `objectWillChange` in the willSet removes the warning
``` swift @Published var value: ObservedType {
        willSet {
            if newValue != value {
                objectWillChange.send()
                objectWillChange.subscribers.forEach {
                    $0.receive(subscription: ObservationSubscription(token: newValue._observe(keyPaths, $0)))
                }
            }
        }
}  ``` Which makes sense here because calling `subscriber.receive()` tells the subscriber that a publisher is ready to receive further requests, and then we tell it to update given subscription by calling `subscriber.receive(subscription)`

PR #8070, opened 2022-12-16
Allow setting `cancelAsyncOpenOnNonFatalErrors` from Swift's `SyncConfiguration`.

dianaafanador3:
Allow setting `cancelAsyncOpenOnNonFatalErrors` from Swift's `SyncConfiguration`, similar to what we already do in the Objective-C API `RLMSyncConfiguration. 
There is already a initialiser which allow this for a partition based configuration, but changing `cancelAsyncOpenOnNonFatalErrors` is not allowed for flexible sync configuration or when inserting a client reset into a user's configuration, and adding a initialiser for each of those cases seems like a too much.
This is added by a user request https://github.com/realm/realm-swift/issues/8071

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,7 @@ x.y.z Release notes (yyyy-MM-dd)
 ### Enhancements
 * Improve performance of creating Projection objects and of change
   notifications on projections ([PR #8050](https://github.com/realm/realm-swift/pull/8050)).
-* Allow setting `cancelAsyncOpenOnNonFatalErrors` from Swift's `SyncConfiguration`.
+* Allow initialising any sync configuration with `cancelAsyncOpenOnNonFatalErrors`.
 
 ### Fixed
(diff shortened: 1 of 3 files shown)


leemaguire:
A flex sync test should be added for this in Swift.

dianaafanador3:
> Making SyncConfiguration mutable is an issue for sendability. Copying a SyncConfiguration doesn't copy the underlying RLMSyncConfiguration, so it's only safe for it to be Sendable because it's immutable.

I was avoiding creating initialiser for all the different combination of sync configurations in user, but It seems that's the only way if we have to maintain syncConfiguration as immutable for it to be Sendable

dianaafanador3:
@tgoyne I changed this to use the initialisers to set `cancelAsyncOpenOnNonFatalErrors`.

tgoyne (approved):

leemaguire (approved):

PR #8106, opened 2023-01-20
Add some missing @preconcurrency annotations

### LINKED ISSUE #8104: RealmSwift 10.34.0 install with SPM, compiling error about @Sendable
- RealmSwift 10.34.0 install with SPM, building compiling error:
     - Converting non-sendable function value to ‘@Sendable (UInt, UInt) → Bool’ may introduce data races
     - image
     - swift-driver version: 1.62.8 Apple Swift version 5.7 (swiftlang-5.7.0.127.4 clang-1400.0.29.50)
Target: x86_64-apple-macosx13.0
     - Xcode Version 14.0.1 (14A400)

tgoyne:
I'm not sure why these missing didn't cause problems with newer versions of Xcode, but they clearly should be there. Xcode 14.0's implementation of sendable checking is still fairly incomplete so our xcodeproj has it disabled, but when importing via SPM or CocoaPods it can end up enabled due to the parent project (and it appears to default to "targeted" with SPM). The swiftpm-ios job would have caught this if we tested it for all Xcode versions and it's a fast job to run (~5 minutes due to not running any tests) so I enabled it for all Xcode versions.

Fixes #8104.


CHANGELOG.md
```diff
@@ -4,8 +4,8 @@ x.y.z Release notes (yyyy-MM-dd)
 * None.
 
 ### Fixed
-* <How to hit and notice issue? what was the impact?> ([#????](https://github.com/realm/realm-swift/issues/????), since v?.?.?)
-* None.
+* Add some missing `@preconcurrency` annotations which lead to build failures
+  with Xcode 14.0 when importing via SPM or CocoaPods (since v10.34.0).
```
dianaafanador3, line 8: Would be nice to have a link to the issues on github as well

dianaafanador3 (approved):
LGTM!, just a small comment

CHANGELOG.md
tgoyne: Added issue link.

PR #8109, opened 2023-01-23
Redesign `@AsyncOpen` and `@AutoOpen` initialisers to allow setting a client reset modes for any sync configuration

dianaafanador3:
Redesign `@AsyncOpen` and `@AutoOpen` initialisers to allow setting a client reset mode for any sync configuration, and avoid confusion with the configuration and the sync configuration from the user on each intialiser.

Currently, setting a client reset mode for @AsyncOpen and @AutoOpen is not working correctly, which mainly has to do with the implementation approach within the property wrappers which uses the sync configuration from the sync type (depending on the initialiser, flx or pbs) and combine it with the injected configuration (which can come from the initialiser or and environment value).
In this PR I'm proposing a redesign of the initials for the both @AsyncOpen and @AutoOpen to avoid confusion of which sync configuration is been used internally. 

This PR has some breaking changes, which only should affect users in cases where they were using this feature incorrectly (for example, if they were injecting a configuration with a different partition value that the one on the initialiser or a configuration with a different sync type that the one from the initialiser).


CHANGELOG.md
```diff
@@ -1,13 +1,30 @@
...
 
-<!-- ### Breaking Changes - ONLY INCLUDE FOR NEW MAJOR version -->
+### Breaking Changes
+* `AsyncOpen.init(appId:partitionValue:configuration:timeout)` and 
+  `AutoOpen.init(appId:partitionValue:configuration:timeout)` will change behavior. Previously 
+  injecting both the configuration and a partition value will make the property wrappers to set
+  the injected configuration (if not nil) combine with the syncConfiguration from the 
+  `user.configuration(partitionValue:)`. This will change to either user the injected configuration (if not nil) 
+  or the configuration obtained from `user.configuration(partitionValue:)` for the current logged user.
+  This API will be deprecated on the next major release.
```
tgoyne, line 18: Both the old and new behavior seems incorrect? If a non-nil configuration is provided I would expect it to use exactly that configuration except with the partition value changed to the supplied one.

RealmSwift/SwiftUI.swift
```diff
@@ -1729,6 +1729,34 @@ private class ObservableAsyncOpenStorage: ObservableObject {
...
+     Initialize the property wrapper for a flexible sync app.
+     - parameter appId: The unique identifier of your Realm app, if empty or `nil` will try to retrieve latest singular cached app.
+     - parameter timeout: The maximum number of milliseconds to allow for a connection to
+                 become fully established., if empty or `nil` no connection timeout is set.
+     */
+    public init(appId: String? = nil,
+                timeout: UInt? = nil) {
+        let app = ObservableAsyncOpenStorage.configureApp(appId: appId, timeout: timeout)
+        // Store property wrapper values on the storage
+        storage = ObservableAsyncOpenStorage(asyncOpenKind: .asyncOpen, app: app, configuration: nil, partitionValue: nil)
```
leemaguire, line 1757: If the configuration is `nil` how does it know about the `initialSubscriptions`?

tgoyne:
This is an unnecessary breaking change which breaks things that seem completely reasonable. `@AutoOpen(partitionValue: "foo", configuration = .init(objectTypes: [Bar.self]))` is different from how configurations are defined elsewhere, but it's not an invalid thing to do.

PR #8114, opened 2023-01-27
Fix issue with ObservedResults and searchable

leemaguire:
Addresses issue where observation subscription is cancelled on initial view refresh.
This PR fixes up the SwiftUITests for iOS 16 too.
Fixes: https://github.com/realm/realm-swift/issues/8096
Fixes: https://github.com/realm/realm-swift/issues/8132

### DIFF
--- a/Realm.xcodeproj/xcshareddata/xcschemes/SwiftUITestHost.xcscheme
+++ b/Realm.xcodeproj/xcshareddata/xcschemes/SwiftUITestHost.xcscheme
@@ -69,6 +69,15 @@
       debugDocumentVersioning = "YES"
       debugServiceExtension = "internal"
       allowLocationSimulation = "YES">
+      <PreActions>
+         <ExecutionAction
+            ActionType = "Xcode.IDEStandardExecutionActionsCore.ExecutionActionType.ShellScriptAction">
+            <ActionContent
+               title = "Run Script"
+               scriptText = "killall Simulator&#10;defaults write com.apple.iphonesimulator ConnectHardwareKeyboard -bool false&#10;">
+            </ActionContent>
+         </ExecutionAction>
+      </PreActions>
       <BuildableProductRunnable
          runnableDebuggingMode = "0">
          <BuildableReference
--- a/RealmSwift/SwiftUI.swift
+++ b/RealmSwift/SwiftUI.swift
@@ -179,7 +179,7 @@ private final class ObservableStoragePublisher<ObjectType>: Publisher where Obje
     public typealias Failure = Never
 
     var subscribers = [AnySubscriber<Void, Never>]()
-    private let value: ObjectType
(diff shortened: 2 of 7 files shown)


Realm.xcodeproj/xcshareddata/xcschemes/SwiftUITestHost.xcscheme
```diff
@@ -69,6 +69,15 @@
       debugDocumentVersioning = "YES"
       debugServiceExtension = "internal"
       allowLocationSimulation = "YES">
+      <PreActions>
+         <ExecutionAction
+            ActionType = "Xcode.IDEStandardExecutionActionsCore.ExecutionActionType.ShellScriptAction">
+            <ActionContent
+               title = "Run Script"
+               scriptText = "killall Simulator&#10;defaults write com.apple.iphonesimulator ConnectHardwareKeyboard -bool false&#10;">
```
leemaguire, line 77: Having the hardware keyboard option enabled will give a high chance of characters being missed when tests are writing to text fields.

RealmSwift/SwiftUI.swift
```diff
@@ -237,6 +242,7 @@ private class ObservableStorage<ObservedType>: ObservableObject where ObservedTy
         willSet {
             if newValue != value {
                 objectWillChange.send()
+                objectWillChange.update(value: newValue)
```
leemaguire, line 245: When `ObservedResults` is setup the publisher is passed `RLMResults<ResultType>.emptyDetached()`. When the actual `Results` to be observed is ready we need to update the publisher to hold a reference to it. Otherwise `receive()` will never setup the `ObservationSubscription` again once the original token is invalidated.

dianaafanador3 (commented):
It would be nice to have a test which really test this particular use case. If I run the SwiftUI tests removing the update function, all the tests are completing successfully (running on ios16), so I cannot reproduce the original issue. The way the issue seems to be happenignwhen we update the realm without using the property wrapper (`$observedResults.append(:)`), or maybe I'm wrong.

tgoyne (approved):

RealmSwift/Sync.swift
```diff
@@ -966,7 +966,9 @@ public class UserPublisher: Publisher {
     /// :nodoc:
     public func receive<S>(subscriber: S) where S: Subscriber, S.Failure == Never, Output == S.Input {
         let token = user.subscribe { _ in
-            _ = subscriber.receive(self.user)
+            DispatchQueue.main.async {
```
tgoyne, line 969: This is not a valid thing to do, and it's producing (correct) sendability warnings.

leemaguire: I'll revert this. Is the correct approach to have the end user bounce to the main thread instead when calling something which mutates the`User`?

PR #8129, opened 2023-02-13
Fix a typo in sync manager docs

nirinchev:


ericjordanmossman (approved):

nirinchev:
Not sure why the xcode cloud tests are failing, but I assume it's not because of my change 😅

dianaafanador3:
> Not sure why the xcode cloud tests are failing, but I assume it's not because of my change 😅

This is because the migration Xcode Cloud branch is still on the making, some of the workflows are enable because I'm doing some tests, but they are failing because they need the code in the branch,

PR #8131, opened 2023-02-13
Expose MongoCollection.watch() as an async sequence

tgoyne:
The actual wrapper is trivial, but this includes some refactoring to let it use `ImmediateScheduler` rather than a pointless extra scheduler hop for each event and a bit of incidental general cleanup.

`withCheckedThrowingContinuation()` is broken on macOS 12 (and presumably iOS 15): it's supposed to inherit the caller's actor, but that doesn't actually work. Using the automatic async bridging for obj-c functions instead of it sidesteps this problem. It turned out to usually simply the code as well, so I updated everything to use that instead of just the places where this bug was a problem.

### DIFF
--- a/Realm/ObjectServerTests/SwiftMongoClientTests.swift
+++ b/Realm/ObjectServerTests/SwiftMongoClientTests.swift
@@ -18,6 +18,7 @@
 
 #if os(macOS)
 
+import Combine
 import RealmSwift
 import XCTest
 
@@ -632,7 +633,7 @@ class SwiftMongoClientTests: SwiftSyncTestCase {
         try performWatchWithMatchFilterTest(.main)
     }
 
-    func testWatchWithMatchFilterAsync() throws {
+    func testWatchWithMatchFilterQueue() throws {
         let queue = DispatchQueue(label: "io.realm.watchQueue", attributes: .concurrent)
         try performWatchWithMatchFilterTest(queue)
     }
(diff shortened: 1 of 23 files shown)


jsflax (approved):

leemaguire (approved):

Realm/ObjectServerTests/SwiftMongoClientTests.swift
```diff
@@ -727,6 +728,76 @@ class SwiftMongoClientTests: SwiftSyncTestCase {
...
+    }
+
+    @available(macOS 12.0, *)
+    @MainActor
+    func testWatchAsync() async throws {
+        try await performAsyncWatchTest()
+    }
+
+    @available(macOS 12.0, *)
+    @MainActor
```
dianaafanador3, line 789: Any reason to run this on the main thread?, or just for testing purposes

dianaafanador3 (approved):
LGTM!, just a few comments and small suggestions.

Realm/ObjectServerTests/SwiftMongoClientTests.swift
tgoyne: It was for `wait(for:)`, but it appears to not need to be MainActor with `fulfillment(of:)`.

Realm/ObjectServerTests/SwiftMongoClientTests.swift
```diff
@@ -727,6 +728,76 @@ class SwiftMongoClientTests: SwiftSyncTestCase {
...
+            await fulfillment(of: [ex], timeout: 2.0)
+        }
+
+        task.cancel()
+        _ = await task.result
+    }
+
+    @available(macOS 12.0, *)
+    @MainActor
+    func testWatchAsync() async throws {
```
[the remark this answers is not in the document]
tgoyne: There isn't a super obvious way to test it. It looks like the only errors that are possible are the http request failing (doing things like passing an invalid match filter just result in no events), and while we should probably be testing that for all of the mongo client stuff we don't currently have anything set up for that.

PR #8140, opened 2023-02-17
Don't bounce to main thread to update User's subscriber.

leemaguire:
`User` incorrectly bounced all notifications to the main thread when being used as an `ObservedObject` in SwiftUI which in turn suppressed sendability warnings. The correct approach is to have the caller which mutates the `User` bounce to the main thread and perform the operation.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,9 +4,9 @@ x.y.z Release notes (yyyy-MM-dd)
 * None.
 
 ### Fixed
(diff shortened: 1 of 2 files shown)


tgoyne:
I think `User.objectWillChange` needs to be `UserPublisher(self).receive(on: DispatchQueue.main)` as SwiftUI does need it to be on the main thread.

leemaguire:
@tgoyne changed to make UserPublisher receive on main thread. PTAL.

tgoyne (approved):

PR #8143, opened 2023-02-17
Fix new warnings introduced in Swift 5.8

tgoyne:
Swift now warns when a collection literal is inferred as being a collection of Any, which resulted in a few hundred warnings in our tests. Deleting unnecessary empty collection literals and adding convenience initializers to Object which explicitly take an array or dictionary of Any eliminates most of these.

For our json literals in RealmServer I added a more specific protocol that we can have a collection of instead of Any, as annotating everything as being `[String: [String: [String: Any]]]` was awful. This relies on the relatively recent protocol existential changes so it's only used for recent swift versions.


Realm/TestUtils/RLMTestCase.m
```diff
@@ -23,7 +23,7 @@
 #import <Realm/RLMSchema_Private.h>
 #import <Realm/RLMRealmConfiguration_Private.h>
 
-static NSString *parentProcessBundleIdentifier()
+static NSString *parentProcessBundleIdentifier(void)
```
tgoyne, line 26: This is a non-C++ source file and `()` in C does bad deprecated things which now produce a warning.

RealmSwift/Realm.swift
```diff
@@ -563,13 +563,6 @@ public typealias AsyncTransactionId = RLMAsyncTransactionId
         }
     }
 
-    /// :nodoc:
-    @discardableResult
-    @available(*, unavailable, message: "Pass .error, .modified or .all rather than a boolean. .error is equivalent to false and .all is equivalent to true.")
-    public func create<T: Object>(_ type: T.Type, value: Any = [:], update: Bool) -> T {
-        fatalError()
-    }
```
tgoyne, line 571: This migration helper is nearly four years old and now produces warnings so I just deleted it. It's not a breaking change since you couldn't call it in the first place.

RealmSwift/Aliases.swift
```diff
@@ -17,7 +17,7 @@
 ////////////////////////////////////////////////////////////////////////////
 
 import Foundation
-import Realm
+@_exported import Realm
```
tgoyne, line 20: typealiasing something from an imported module will require users of the typealias to import that module in Swift 6. This change preserves the old behavior.

RealmSwift/Aliases.swift
```diff
@@ -17,7 +17,7 @@
 ////////////////////////////////////////////////////////////////////////////
 
 import Foundation
-import Realm
+import Realm.Swift
```
tgoyne, line 20: Using an aliased type from a resilient module (which SPM builds some things as) without importing Realm now gives "'Object' aliases 'Realm.RealmSwiftObject' and cannot be used here because 'Realm' was not imported by this file; this is an error in Swift 6". `@_exported import Realm` re-exports way too much, so I moved RealmSwiftObject to a separate submodule which doesn't export too much. I'm not sure why it doesn't need `@_exported`. This might just be making things too complicated for the warning to work rather than fixing the problem.

leemaguire (approved):

PR #8146, opened 2023-02-22
Update BAAS to support Flexible Sync Permissions Changes

dianaafanador3:
This PR changes how we set permissions for the flexible sync service through the Admin API for our server tests.


Realm/ObjectServerTests/RealmServer.swift
```diff
@@ -903,6 +894,29 @@ public class RealmServer: NSObject {
             throw URLError(.badServerResponse)
         }
 
+        switch syncMode {
```
tgoyne, line 897: Isn't this just an if?

tgoyne (approved):

PR #8168, opened 2023-03-13
Import as `RLMRealm_Private.h` as a module would cause issues when using Realm as a subdependency

dianaafanador3:
Import as `RLMRealm_Private.h` as a module would cause issues when using Realm as a subdependency.
Because `<Realm/RLMRealm_Private.h>` is part of the Private Realm Module, there seems to be an issue when you use `Realm` as a sub-dependency with Cocopods which doesn't find the file within the headers associated to the module, even if we expose them as private headers. 
Importing the file relative to the directory using quotation marks fixes the issue.

### DIFF
--- a/.jenkins.yml
+++ b/.jenkins.yml
@@ -34,6 +34,7 @@ target:
  - cocoapods-osx
  - cocoapods-ios
  - cocoapods-ios-dynamic
+ - cocoapods-ios-subdependency
  - cocoapods-watchos
  - swiftui-ios
  - swiftui-server-osx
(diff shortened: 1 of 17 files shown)


dianaafanador3:
This fix seems to be working using Realm as a sub-dependency, and it is working for all the installation methods.

tgoyne:
This is something we should probably have an installation test for if we currently don't have anything which breaks when we include the header incorrectly.

dianaafanador3:
> This is something we should probably have an installation test for if we currently don't have anything which breaks when we include the header incorrectly.

Added installation test using Realm as a subdependency with Cocoapods to test this

tgoyne (approved):

PR #8169, opened 2023-03-13
Enable actor data race detection and fix resulting problems

tgoyne:
This revealed a number of places where callbacks were invoked on a background thread but weren't marked as Sendable. It also unforunately revealed some more compiler bugs related to property wrappers and sendability; the `projectedValue` hack with `@Locked` doesn't work with data race detection enabled because it generates an actor-confined closure as part of the implementation.


tgoyne:
CI failures appear to all be spurious.

RealmSwift/Sync.swift
```diff
@@ -553,7 +554,7 @@ public typealias Provider = RLMIdentityProvider
     }
 
     /// A closure type for receiving the completion of a remote function call.
-    public typealias FunctionCompletionHandler = (AnyBSON?, Error?) -> Void
+    public typealias FunctionCompletionHandler = @Sendable (AnyBSON?, Error?) -> Void
 
     /// A closure type for the dynamic remote function type.
     public typealias Function = ([AnyBSON], @escaping FunctionCompletionHandler) -> Void
```
dianaafanador3, line 560: Shouldn't this be @Sendable as well?

RealmSwift/Tests/TestUtils.swift
```diff
@@ -35,6 +35,10 @@ public class Locked<T>: @unchecked Sendable {
         lock.initialize(to: os_unfair_lock())
     }
```
dianaafanador3, line 37: I would love for this property wrappers to have some sort of documentation/comment, so we can have it mind to use it for testing in new tests.

RealmSwift/Sync.swift
```diff
@@ -905,7 +908,7 @@ public extension User {
     /// Refresh a user's custom data. This will, in effect, refresh the user's auth session.
     /// @returns A publisher that eventually return `Dictionary` with user's data or `Error`.
     func refreshCustomData() -> Future<[AnyHashable: Any], Error> {
```
dianaafanador3, line 910: From my understanding a function should be marked as `@preconcurrency`, to tell Swift the this function was created without modern concurrency, so should' we make this one if we already marked the one using a completion block, maybe I'm misunderstanding this.

RealmSwift/Realm.swift
```diff
@@ -1231,13 +1231,17 @@ extension Realm {
         let task = RLMAsyncOpenTask(configuration: rlmConfiguration, confinedTo: scheduler,
                                     download: shouldAsyncOpen(configuration, downloadBeforeOpen))
         do {
-            try await withTaskCancellationHandler {
-                // Work around https://github.com/apple/swift/issues/61119 by smuggling
-                // the Realm out via a property on task rather than returning it
-                task.localRealm = try await task.waitForOpen()
-            } onCancel: {
-                task.cancel()
+            // Work around https://github.com/apple/swift/issues/61119, which
```
dianaafanador3, line 1234: Should this be fixed in the future , Should we mark this as FIXME or other marker?, same as the one in TestUtils

RealmSwift/Tests/TestUtils.swift
dianaafanador3: Also is the projected value workaround needed anymore here?

Realm/ObjectServerTests/SwiftObjectServerTests.swift
```diff
@@ -2443,11 +2391,11 @@ class CombineObjectServerTests: SwiftSyncTestCase {
         let objIds = collection.insertMany([document, document2, document3, document4]).await(self)
         let objectIds = objIds.map { $0.objectIdValue! }
 
-        var watchEx1 = expectation(description: "Main thread watch")
-        var watchEx2 = expectation(description: "Background thread watch")
+        let watchEx1 = Locked(expectation(description: "Main thread watch"))
```
dianaafanador3, line 2394: Why do we need a lock to access this expectations?

RealmSwift/Sync.swift
tgoyne: It doesn't *need* to be, but it looks like it (and the containing type) can be sendable and so might as well be.

RealmSwift/Tests/TestUtils.swift
tgoyne: There's still a few spots using it where the callback actually will be called on the main thread.

RealmSwift/Sync.swift
tgoyne: `@preconcurrency` is only applicable to functions which take a `@Sendable` closure argument. Changing an argument from non-sendable to sendable is an API breaking change, which `@preconcurrency` fixes by making the function only require a sendable argument when in swift 6/strict concurrency mode, where the user has specifically opted into breaking changes.

Realm/ObjectServerTests/SwiftObjectServerTests.swift
tgoyne: Combine and sendable checking don't really combine well. The main thread one will always be used on the main thread, but the compiler doesn't know that. It could safely just be `Unchecked` instead of `Locked`, but I'm inclined to say that we should always be using `Locked` in tests when we need to silence a sendability warning even if it shouldn't be required simply because it's less likely to result in weird mysterious problems if we're wrong about it not being required. In non-test code the perf overhead is more of a concern.

The background thread expectation actually does require locking as it's written to on the main thread and read from a dispatch queue.

PR #8191, opened 2023-03-28
Add missing fields in `-[RLMRealmConfiguration copy]`

### LINKED ISSUE #8190: Migrating from 10.33.0 to 10.34.0 can no longer cast to `MigrationObject`.

Always

### Description

# Overview

We are looking to migrate from version 10.33.0 to the latest version (10.37.1 at the time of writing) but our tests flagged up an issue affecting migration. It appears we can no longer cast to `MigrationObject/DynamicObject` which worked previously. Here is the migration code with added `print` statements to illustrate what is going wrong:

``` swift
m.enumerateObjects(ofType: Subscription.className()) { (old, new) in

    guard let old, let new else { return }

    if let oldAccount = old["account"] as? DynamicObject {
        print("Old account: \(oldAccount)")
        if let accountId = oldAccount["id"] as? String {
            print("Account ID: \(accountId)")
            new["accountID"] = accountId // 10.33.0 successfully migrates the value
        } else {
            print("Failed to get account ID")
        }
    } else {
         print("Failed to get old account") // 10.34.0+ fails at this point instead of casting correctly
    }
}
```
In 10.33.0 we are able to cast `old["account"]` to a `DynamicObject` or `MigrationObject` and access it's subscript properties without issue. But as of 10.34.0 (I've tested all versions up to the latest and 10.34.0 was where the issue was introduced) the cast fails. I can't see anything within the release notes which would indicate a change in this behaviour.

# Details

What seems to have worked (on 10.34.0+) is casting to `RLMDynamicObject` in the console. At this point I can then access the subscript properties with no issue. Unfortunately this is only possible in the console due to `RLMDynamicObject` being private. 

Screenshot 2023-03-28 at 17 36 05
Screenshot 2023-03-28 at 17 36 57

But if I try to cast to either `DynamicObject` or `MigrationObject` I get `nil`.

Screenshot 2023-03-28 at 17 38 27

The `Subscription` object is quite large, but you can see that `account` is clearly associated to it in this small snippet.

Screenshot 2023-03-28 at 17 42 38

### Stacktrace & log output

_No response_

Always

### Reproduction Steps

The actual code this relates to is sensitive so I've shared what I can publicly for now.

### Version

10.34.0

### What Atlas Services are you using?

Local Database only

### Are you using encryption?

No

All

### Build environment

Xcode version: 14.2
Dependency manager and version: SPM

tgoyne:
Fixes #8190.

Most of the changes here are just refactoring the migration tests to perform each test with four different ways to trigger the migration.

### DIFF
--- a/Realm/RLMRealmConfiguration.mm
+++ b/Realm/RLMRealmConfiguration.mm
@@ -128,6 +128,8 @@ NSString *RLMRealmPathForFile(NSString *fileName) {
     configuration->_migrationBlock = _migrationBlock;
     configuration->_shouldCompactOnLaunch = _shouldCompactOnLaunch;
     configuration->_customSchema = _customSchema;
+    configuration->_eventConfiguration = _eventConfiguration;
+    configuration->_migrationObjectClass = _migrationObjectClass;
     configuration->_initialSubscriptions = _initialSubscriptions;
     configuration->_rerunOnOpen = _rerunOnOpen;
     return configuration;
--- a/RealmSwift/Tests/MigrationTests.swift
+++ b/RealmSwift/Tests/MigrationTests.swift
@@ -45,71 +45,78 @@ private func dynamicRealm(_ fileURL: URL) -> RLMRealm {
 }
 
 class MigrationTests: TestCase {
-
-    // MARK: Utility methods
-
-    // create realm at path and test version is 0
-    private func createAndTestRealmAtURL(_ fileURL: URL) {
-        autoreleasepool {
-            _ = try! Realm(fileURL: fileURL)
-            return
+    private func createDefaultRealm() throws {
+        let config = Realm.Configuration(fileURL: defaultRealmURL())
+        try autoreleasepool {
+            _ = try Realm(configuration: config)
         }
-        XCTAssertEqual(0, try! schemaVersionAtURL(fileURL), "Initial version should be 0")
+        XCTAssertEqual(0, try schemaVersionAtURL(config.fileURL!))
     }
 
-    // migrate realm at path and ensure migration
-    private func migrateAndTestRealm(_ fileURL: URL, shouldRun: Bool = true, schemaVersion: UInt64 = 1,
-                                     autoMigration: Bool = false, block: MigrationBlock? = nil) {
+#if swift(<5.7)
+    // Work around pre-Swift 5.7 multiple trailing closures resolution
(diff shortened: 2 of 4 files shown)


Realm/RLMRealmConfiguration.mm
```diff
@@ -128,6 +128,8 @@ - (instancetype)copyWithZone:(NSZone *)zone {
     configuration->_migrationBlock = _migrationBlock;
     configuration->_shouldCompactOnLaunch = _shouldCompactOnLaunch;
     configuration->_customSchema = _customSchema;
+    configuration->_eventConfiguration = _eventConfiguration;
+    configuration->_migrationObjectClass = _migrationObjectClass;
```
dianaafanador3, line 132: Why do we need to set this `migrationObjectClass` in the configuration?, I see this is set to MigrationObject.self as default and only available to from our Private API, if this is to allow users to setUp their own Migration Object Classes through our private API, or this is for testing purposes?. I don't see this documented anywhere.

RealmSwift/Tests/MigrationTests.swift
```diff
@@ -45,72 +45,79 @@ private func dynamicRealm(_ fileURL: URL) -> RLMRealm {
...
 
-        XCTAssertEqual(didRun.value, shouldRun)
-    }
-
-    private func migrateAndTestDefaultRealm(_ schemaVersion: UInt64 = 1, block: @escaping MigrationBlock) {
-        migrateAndTestRealm(defaultRealmURL(), schemaVersion: schemaVersion, block: block)
-        let config = Realm.Configuration(fileURL: defaultRealmURL(),
-                                         schemaVersion: schemaVersion)
-        Realm.Configuration.defaultConfiguration = config
+        try withTestFile {
```
dianaafanador3, line 96: 👌

dianaafanador3 (approved):
LGTM!

Realm/RLMRealmConfiguration.mm
tgoyne: Inside a migration we need to give the user RLMMigrationObject objects when they're using the obj-c API and MigrationObject objects when they're using the Swift API. The migration code doesn't know which is in use, and even if it did it can't directly reference MigrationObject since that's defined by the Swift framework. Instead we store the class which we want to instantiate in the config, and set it to either the obj-c or swift class depending on how the config is contructed.

dianaafanador3: Ohh I see it, thanks for the explanation

PR #8195, opened 2023-03-30
Address some memory leaks

leemaguire:
Attempts to address some memory leak issues a user is having.

### DIFF
--- a/Realm/RLMBSON.mm
+++ b/Realm/RLMBSON.mm
@@ -370,7 +370,7 @@ id<RLMBSON> RLMConvertBsonToRLMBSON(const Bson& b) {
         case realm::bson::Bson::Type::Double:
             return @(static_cast<double>(b));
         case realm::bson::Bson::Type::String:
-            return RLMStringDataToNSString(static_cast<std::string>(b).c_str());
+            return @(static_cast<std::string>(b).c_str());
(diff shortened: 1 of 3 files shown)


Realm/RLMBSON.mm
```diff
@@ -370,7 +370,7 @@ Bson RLMConvertRLMBSONToBson(id<RLMBSON> b) {
         case realm::bson::Bson::Type::Double:
             return @(static_cast<double>(b));
         case realm::bson::Bson::Type::String:
-            return RLMStringDataToNSString(static_cast<std::string>(b).c_str());
+            return [NSString stringWithUTF8String:static_cast<std::string>(b).c_str()];
```
leemaguire, line 373: This ensures we copy the bytes. We also didn't need `RLMStringDataToNSString` as the value comes in as a `std::string` so we performed an extra conversion too.

tgoyne: RLMStringDataToNSString also copies the input data. Since this doesn't need to handle embedded nulls, this should be just `@(static_cast<std::string>(b).c_str())`.

leemaguire:
@tgoyne comment addressed.

tgoyne (approved):

PR #8197, opened 2023-04-03
Add actor-isolated Realms with async writes

tgoyne:


Realm.xcodeproj/xcshareddata/xcschemes/RealmSwift.xcscheme
```diff
@@ -41,10 +41,11 @@
       </BuildActionEntries>
    </BuildAction>
    <TestAction
-      buildConfiguration = "Debug"
+      buildConfiguration = "Release"
```
leemaguire, line 44: Should this not be `Debug`?

leemaguire (commented):
Looking good, just a couple of comments

Realm/RLMConstants.h
```diff
@@ -55,7 +55,12 @@ RLM_HEADER_AUDIT_BEGIN(nullability, sendability)
 #define RLM_ERROR_ENUM(type, name, domain) NS_ENUM(type, name)
 #endif
 
-#define REALM_HIDDEN __attribute__((visibility("hidden")))
+#define RLM_HIDDEN __attribute__((visibility("hidden")))
+#define RLM_VISIBLE __attribute__((visibility("default")))
+#define RLM_HIDDEN_BEGIN _Pragma("GCC visibility push(hidden)")
+#define RLM_HIDDEN_END _Pragma("GCC visibility pop")
+#define RLM_DIRECT __attribute__((objc_direct))
+#define RLM_DIRECT_MEMBERS __attribute__((objc_direct_members))
```
jsflax, line 63: Are these just for convenience?

Realm/RLMScheduler.mm
```diff
@@ -72,8 +93,81 @@ - (void *)cacheKey {
...
+        });
+    }
+
+    bool is_on_thread() const noexcept override {
+        _verify();
+        return true;
+    }
+
+    bool is_same_as(const Scheduler *) const noexcept override {
+        return false;
```
jsflax, line 116: Is this correct? Shouldn't this check if it's the same actor?

RealmSwift/EmbeddedObject.swift
```diff
@@ -217,6 +217,94 @@ extension EmbeddedObject: _RealmCollectionValueInsideOptional {
...
+
+    - warning: This method cannot be called during a write transaction, or when
+    the containing Realm is read-only.
+    - parameter actor: The actor to isolate notifications to.
+    - parameter block: The block to call with information about changes to the object.
+    - returns: A token which must be held for as long as you want updates to be delivered.
+     */
+    @available(macOS 10.15, tvOS 13.0, iOS 13.0, watchOS 6.0, *)
+    @_unsafeInheritExecutor
+    public func observe<A: Actor, T: Object>(
```
jsflax, line 256: Why can't the actor be assumed? Couldn't we pull the actor from the Realm?

RealmSwift/Realm.swift
```diff
@@ -1205,42 +1230,208 @@ extension Realm {
...
+     either a local actor or a global actor. The calling function does not need
+     to be isolated to the actor passed in, but if it is not it will not be
+     able to use the returned Realm.
+     - parameter downloadBeforeOpen: When opening the Realm should first download
+     all data from the server.
+     - throws: An `NSError` if the Realm could not be initialized.
+               `CancellationError` if the task is cancelled.
+     - returns: An open Realm.
+     */
+    public init<A: Actor>(configuration: Realm.Configuration = .defaultConfiguration,
```
jsflax, line 1269: Adding a note her as per our discussion: it would be nice to have a synchronous init specifically for the case of synchronous actor initialization.

RealmSwift/EmbeddedObject.swift
tgoyne: This function is specifically for observing on a different actor, similar to the option to observe on a different dispatch queue.

Realm/RLMScheduler.mm
tgoyne: This function is only used by objectstore's Realm cache, which we don't use.

Realm/RLMScheduler.mm
```diff
@@ -90,6 +184,14 @@ + (RLMScheduler *)dispatchQueue:(dispatch_queue_t)queue {
     return RLMScheduler.currentRunLoop;
 }
 
++ (RLMScheduler *)actor:(id)actor invoke:(void (^)(dispatch_block_t))invoke verify:(void (^)())verify {
+    auto mainRunLoopScheduler = RLMScheduler.mainRunLoop;
```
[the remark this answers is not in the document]
tgoyne: The one instance of RLMMainLoopScheduler is something like 32 bytes. The class definition uses significantly more resources than the object instance.

jsflax (approved):

RealmSwift/Realm.swift
tgoyne: I think this needs SE-0392 (custom actor executors) to be reasonable. With the current limitations it'd have to be entirely unchecked and would be a super error-prone API.

PR #8199, opened 2023-04-04
Integrate Custom Logger implementation and new Logger API.

dianaafanador3:
* Added support adding a user created default logger, which allows implementing your own logging logic
   and the log threshold level.
   You can define your own logger creating an instance of `Logger` and define the log function which will be
   invoked whenever there is a log message.

  ```swift
  let logger = Logger(level: .all) { level, message in
     print("Realm Log - \(level): \(message)")
  }
  ```

  Set this custom logger as you default logger using `Logger.shared`.
  ```swift
     Logger.shared = inMemoryLogger
  ```
   This logger can be set as default using the new API `Logger.setDefaultLogger(inMemoryLogger)`.
 * It is now possible to change the default log threshold level at any point of the application's lifetime.
   ```swift
   Logger.shared.logLevel = .debug
   ```
   This will override the log level set anytime before by a user created logger.
 * We have set `.warn` as the default log threshold level for Realm. You will now see some 
   log message in your console. To disable use `Logger.shared.level = .off`.


CHANGELOG.md
```diff
@@ -1,12 +1,38 @@
 x.y.z Release notes (yyyy-MM-dd)
 =============================================================
 ### Enhancements
-* None.
+* Added support for creating and add a custom logger, and implement your own logging logic.
```
leemaguire, line 4:
```suggestion
* Added support for creating a custom logger, to allow implementing your own logging logic.
```

Realm/RLMLogger.mm
```diff
@@ -0,0 +1,192 @@
...
+    Logger::set_default_level_threshold(level);
+}
+} // anonymous namespace
+
+@implementation RLMLogger {
+    std::shared_ptr<Logger> _logger;
+}
+
+static RLMLogger *_defaultLogger;
+static RLMLogLevel _defaultLogLevel = RLMLogLevelWarn;
```
tgoyne, line 109: An underscore prefix on a variable indicates that it's a member variable or object ivar, which these are not.

Realm/RLMLogger.mm
```diff
@@ -0,0 +1,192 @@
...
+    return self;
+}
+
+- (instancetype)init {
+    if (self = [super init]) {
+        auto logger = std::make_shared<CustomLogger>();
+        auto wrapper = LoggerFunctionCallbackWrapper{(^(RLMLogLevel level, NSString *message) {
+            [self doLog:level message:message];
+        })};
+        logger->function = wrapper.block;
```
tgoyne, line 132: Why is this constructing a LoggerFunctionCallbackWrapper and then doing nothing with it but reading the property set in the constructor?

Realm/RLMLogger.h
```diff
@@ -0,0 +1,129 @@
...
+
+     // InMemoryLogger.m
+     @implementation InMemoryLogger
+     - (void)doLog:(RLMLogLevel)logLevel message:(NSString *)message {
+        NSString *newLogs = [_logs stringByAppendingFormat:@" %@ %lu %@", [NSDate now], logLevel, message];
+        _logs = newLogs;
+     }
+     @end
+
+ Set this custom logger as you default logger using `[RLMLogger setDefaultLogger:]`.
```
tgoyne, line 70:
```suggestion
 Set this custom logger as your default logger using `[RLMLogger setDefaultLogger:]`.
```

Realm/RLMLogger.h
```diff
@@ -0,0 +1,129 @@
...
+    /// - warning: Will incur a measurable performance impact.
+    RLMLogLevelTrace,
+    /// Log information that can aid in debugging. More events will be logged than with `RLMSyncLogLevelTrace`.
+    ///
+    /// - warning: Will incur a measurable performance impact.
+    RLMLogLevelAll
+} NS_SWIFT_NAME(LogLevel);
+
+/**
+ `RLMLogger` is a base class for creating your own custom logging logic.
```
tgoyne, line 52: Why does this API involve defining a subclass rather than just a logging callback like the old API? I don't really see why we'd want to expose an interface for users to do their own logging via the logger they passed to us, and the base class doesn't add any flexibility.

Realm/RLMLogger.mm
```diff
@@ -0,0 +1,192 @@
...
+    auto level = levelForLogLevel(logLevel);
+    if (_logger->would_log(level)) {
+        _logger->log(level, message.UTF8String);
+    }
+}
+
+#pragma mark Global Logger Setter
+
++ (RLMLogLevel)logLevel {
+    return _defaultLogLevel;
```
tgoyne, line 160: This should be `Logger::get_default_level_threshold()` rather than having a separate duplicated value.

Realm/RLMSyncManager.h
```diff
@@ -22,6 +22,8 @@
 
 RLM_HEADER_AUDIT_BEGIN(nullability, sendability)
 
+// NEXT-MAJOR: This enum needs to be removed when access to the logger is removed
+// from the sync manager.
```
tgoyne, line 26: This type should presumably be deprecated?

Realm/Tests/RealmTests.mm
```diff
@@ -2939,3 +2939,76 @@ - (void)testDeleteOpenRealmFile {
...
+    [RLMLogger setDefaultLogger:inMemoryLogger];
+
+    @autoreleasepool { [RLMRealm defaultRealm]; }
+    XCTAssertTrue([inMemoryLogger.logs length] == 0);
+
+    // Test LogLevel Info
+    inMemoryLogger.level = RLMLogLevelInfo;
+    @autoreleasepool { [RLMRealm defaultRealm]; }
+    XCTAssertTrue([inMemoryLogger.logs length] >= 0);
+    XCTAssertTrue([inMemoryLogger.logs rangeOfString:@"4 DB:" options:NSRegularExpressionSearch].location != NSNotFound); // Info
```
tgoyne, line 2982: Why are these doing regex searches?

RealmSwift/Tests/RealmTests.swift
```diff
@@ -1495,3 +1495,120 @@ extension CancellationError: Equatable {
...
+            return "unknown"
+        }
+    }
+}
+
+@available(macOS 12.0, watchOS 8.0, iOS 15.0, tvOS 15.0, macCatalyst 15.0, *)
+class LoggerTests: TestCase {
+    func testSetDefaultLogLevel() throws {
+        let inMemoryLogger = InMemoryLogger(level: .off)
+        Logger.setDefaultLogger(inMemoryLogger)
```
tgoyne, line 1548: The idiomatic Swift naming for this is `Logger.shared = isMemoryLogger`.

Realm/RLMLogger.h
tgoyne: I think the minimal reasonable API we could do which doesn't lock us into just a single global default logger would be something like:

```swift
class Logger {
    public class var shared: Logger
    public var level: LogLevel
    public init(level: LogLevel, logFunction: (LogLevel, String) -> Void)
}

// Used as
Logger.shared = Logger(level: .info) { level, msg in
  print("\(level): \(msg)")
}
Logger.shared.level = .warn

File: Realm/RLMSyncManager.h (hunk 7 above)
Comment (dianaafanador3) [reply Realm/RLMSyncManager.h:26]: When we deprecate setting the log level and log function from the syncManager and we move it to be set  through the app or the realm. we should instead use `RLMLogLevel`.

File: Realm/Tests/RealmTests.mm (hunk 8 above)
Comment (dianaafanador3) [reply Realm/Tests/RealmTests.mm:2982]: Trying to check if the log for the specific log level are been added to the log string we are using for testing

**Comment by dianaafanador3** [comment]
@tgoyne There is a new proposal for the Core Logging API following your guidelines.

File: Realm/RLMLogger.h [hunk 10]
```diff
@@ -0,0 +1,111 @@
...
+/**
+ `RLMLogger` is used for creating your own custom logging logic.
+
+ You can define your own logger creating an instance of `RLMLogger` and define the log function which will be
+ invoked whenever there is a log message.
+ Set this custom logger as you default logger using `setDefaultLogger`.
+
+     [RLMLogger setDefaultLogger:[[RLMLogger alloc] initWithLevel:RLMLogLevelDebug logFunction:^(RLMLogLevel level,    NSString * _Nonnull message) {
+        NSLog(@"Realm Log - %lu, %@", (unsigned long)level, message);
+     }]];
```
tgoyne, line 67:
```suggestion
     RLMLogger.defaultLogger = [[RLMLogger alloc] initWithLevel:RLMLogLevelDebug
                                                    logFunction:^(RLMLogLevel level, NSString *message) {
        NSLog(@"Realm Log - %lu, %@", (unsigned long)level, message);
     }];
```

Realm/RLMLogger.h
```diff
@@ -0,0 +1,111 @@
...
+
+/**
+  Gets the logging threshold level used by the logger.
+ */
+@property (nonatomic) RLMLogLevel level;
+
+/// :nodoc:
+- (instancetype)init NS_UNAVAILABLE;
+
+/// Creates a  logger with the associated log level and the logic function to define your own logging logic.
```
tgoyne, line 81: Extra space here, and the blank line with no `///` between this one and the next makes this line not part of the doc comment.

Realm/RLMLogger.h
```diff
@@ -0,0 +1,111 @@
...
+ Sets a logger implementation as default, that will be used whenever information must be logged.
+
+ @param logger The `RLMLogger` to be configured as the default logger
+ */
++ (void)setDefaultLogger:(RLMLogger *)logger;
+
+/**
+ Gets the current set default logger.
+ */
++ (instancetype)defaultLogger;
```
tgoyne, line 107: These can be declared as `@property (class) RLMLogger *defaultLogger;`.

Realm/RLMLogger.mm
```diff
@@ -0,0 +1,171 @@
...
+@implementation RLMLogger {
+    std::shared_ptr<Logger> _logger;
+}
+
+static RLMLogger *s_defaultLogger;
+
+typedef void(^LoggerBlock)(RLMLogLevel level, NSString *message);
+
+- (RLMLogLevel)level {
+    return logLevelForLevel(Logger::get_default_level_threshold());
```
tgoyne, line 105: Now that these are instance methods they should be getting and setting the level on the instance's logger.

Realm/RLMLogger.mm
```diff
@@ -0,0 +1,171 @@
...
+    if (!s_defaultLogger) {
+        auto defaultLogger = std::make_shared<CocoaLogger>();
+        defaultLogger->set_level_threshold(Level::warn);
+        s_defaultLogger = [[RLMLogger alloc] initWithLogger:defaultLogger];
+    }
+
+    logger = s_defaultLogger;
+    Logger::set_default_logger(logger->_logger);
+    return logger;
+}
```
tgoyne, line 159: Setting the default default logger should be done in `+[RLMLogger initialize]`, which is invoked by the obj-c runtime the first time the class is referenced in a run of the application. Ensuring it's done before RLMApp and RLMRealm init can be done by calling `[RLMLogger class]` in those places.

Storing our own copy of the RLMLogger adds some complications around threading. It'd be simpler to just do `return [[RLMLogger alloc] initWithLogger:Logger::get_default_logger()]` here, but obviously that initializes a new object each time it's accessed (which probably isn't very often as there's no reason to read the logger other than changing the log level?). Reusing one instance requires adding some locking here.

Realm/RLMLogger.mm
```diff
@@ -0,0 +1,171 @@
...
+
++ (void)setDefaultLogger:(RLMLogger *)logger {
+    Logger::set_default_level_threshold(logger->_logger->get_level_threshold());
```
tgoyne, line 162: No need to call `set_default_level_threshold()`, as it'll be replaced by the log level of the logger instance.

Realm/RLMSyncManager.h
tgoyne: We're deprecating those in this PR, so we should also deprecate the types only used by those.

Realm/Tests/RealmTests.mm
```diff
@@ -2937,5 +2937,90 @@ - (void)testDeleteOpenRealmFile {
...
+        logs = newLogs;
+    }]];
+
+    XCTAssertEqual([RLMLogger defaultLogger].level, RLMLogLevelDebug);
+    @autoreleasepool { [RLMRealm defaultRealm]; }
+    XCTAssertTrue([logs rangeOfString:@"4 DB:" options:NSRegularExpressionSearch].location != NSNotFound); // Info
+    XCTAssertTrue([logs rangeOfString:@"7 DB:" options:NSRegularExpressionSearch].location == NSNotFound); // Trace
+}
+
+- (void)testCustomLoggerLogMessage {
```
tgoyne, line 3010: This test is testing a non-public function which appears to only be used in this test.

Realm/Tests/RealmTests.mm
tgoyne: It's searching for a literal string though, not a regular expression. These can just be `containsString:` rather than a regex search.

Realm/Tests/RealmTests.mm
dianaafanador3: This is used this API `- (void)logWithLevel:(RLMLogLevel)logLevel message:(NSString *)message, ... NS_SWIFT_UNAVAILABLE("");` which can be mistaken for this one `- (void)logLevel:(RLMLogLevel)logLevel message:(NSString *)message` which is explicitly exposed for the Swift API

Realm/RLMLogger.mm
dianaafanador3: Yep you are right the only trade-off of this implementation will be that we are initialising RLMLogger every time we get the default logger, but as you mention this is only done when there is a change in the log level, and I don't think users will be changing the log level frequently on runtime (unless and special use case). In the other hand this will be more thread safe (less error prone), and we wouldn't have to use a lock here, we are basically relaying on the lock in Core.

dianaafanador3:
@tgoyne I'm not deprecating the logger parameter on EventConfiguration for now, given that this used by only one customer. Probably in the future when we can inject a logger into an App and a Realm, this will no longer be necessary. Not deprecating this, is causing a warning at the moment.

Realm/ObjectServerTests/RLMSyncTestCase.mm
```diff
@@ -555,7 +555,8 @@ - (RLMApp *)app {
     if (!_app) {
         _app = [RLMApp appWithId:self.appId configuration:self.defaultAppConfiguration rootDirectory:self.clientDataRoot];
         RLMSyncManager *syncManager = self.app.syncManager;
-        syncManager.logLevel = RLMSyncLogLevelOff;
+        [RLMLogger defaultLogger].level = RLMLogLevelOff;
+#pragma clang diagnostic pop
```
tgoyne, line 559: Accidental paste? This doesn't appear to be linked to anything.

Realm/RLMApp.mm
```diff
@@ -21,6 +21,7 @@
 #import "RLMBSON_Private.hpp"
 #import "RLMCredentials_Private.hpp"
 #import "RLMEmailPasswordAuth.h"
+#import "RLMLogger_Private.h"
```
tgoyne, line 24: This doesn't need the private header.

Realm/RLMLogger.h
tgoyne: This still applies. Using dot syntax to set properties is the normal thing to do in modern obj-c.

Realm/Tests/RealmTests.mm
tgoyne: Why is it exposed for the Swift API? We're exposing a way for users to define a sink for our log messages, not for them to do their own logging.

RealmSwift/Logger.swift
```diff
@@ -0,0 +1,100 @@
...
+
+ Set this custom logger as you default logger using `Logger.shared`.
+
+ ```swift
+    Logger.shared = inMemoryLogger
+ ```
+
+ - note: By default default log threshold level is `.warn`, and logging strings are output to Apple System Logger.
+*/
+class Logger {
```
tgoyne, line 42: Why is this a wrapper type and not a typealias for Logger?

dianaafanador3:
https://github.com/realm/realm-swift/pull/8199#discussion_r1179559610 I expose this as internal, so if we want to log a message in the SDK side this will be helpful

dianaafanador3:
Changing the default level from .warn to .info after the discussion on the project channel, this will be documented and there will be some ground rules for when to use each log level.

Realm/Tests/RealmTests.mm
dianaafanador3: Changed to be used internally in case we want log things at SDK level

Realm/Tests/RealmTests.mm
```diff
@@ -2937,5 +2938,93 @@ - (void)testDeleteOpenRealmFile {
...
+    __block NSString *logs = @"";
+    RLMLogger *logger = [[RLMLogger alloc] initWithLevel:RLMLogLevelAll logFunction:^(RLMLogLevel level, NSString * _Nonnull message) {
+        NSString *newLogs = [logs stringByAppendingFormat:@" %@ %lu %@", [NSDate date], level, message];
+        logs = newLogs;
+    }];
+    RLMLogger.defaultLogger = logger;
+
+    @autoreleasepool { [RLMRealm defaultRealm]; }
+    XCTAssertEqual([RLMLogger defaultLogger].level, RLMLogLevelAll);
+    XCTAssertTrue([logs rangeOfString:@"5 DB:" options:NSRegularExpressionSearch].location != NSNotFound); // Detail
```
tgoyne, line 2961: These still don't need to be regex searches.

Realm/RLMLogger.mm
```diff
@@ -0,0 +1,159 @@
...
+    return self;
+}
+
+- (void)logWithLevel:(RLMLogLevel)logLevel message:(NSString *)message, ... {
+    auto level = levelForLogLevel(logLevel);
+    if (_logger->would_log(level)) {
+        va_list args;
+        va_start(args, message);
+        _logger->log(level, message.UTF8String, args);
+        va_end(args);
```
tgoyne, line 139: This doesn't work, and I'm surprised it even compiles. `Logger::log()` doesn't work on a va_list (and it can't, as it needs the static types of arguments to know how to format). This'd need to be something like `_logger->log(level, "%1", [[NSString alloc] initWithFormat:message arguments:args]);`

Realm/RLMLogger.mm
```diff
@@ -0,0 +1,159 @@
...
+    }
+}
+
+- (void)logLevel:(RLMLogLevel)logLevel message:(NSString *)message {
+    auto level = levelForLogLevel(logLevel);
+    if (_logger->would_log(level)) {
+        _logger->log(level, message.UTF8String);
```
tgoyne, line 146: The second argument to `Logger:log()` is a format string, so this'll break if `message` contains the character `%`. It needs to instead be `_logger->log(level, "%1", message.UTF8String);` so that it doesn't parse `message`.

RealmSwift/Tests/RealmTests.swift
```diff
@@ -1903,3 +1903,96 @@ extension CancellationError: Equatable {
...
+        case .all:
+            return "All"
+        default:
+            return "unknown"
+        }
+    }
+}
+
+@available(macOS 12.0, watchOS 8.0, iOS 15.0, tvOS 15.0, macCatalyst 15.0, *)
+class LoggerTests: TestCase {
```
tgoyne, line 1936: These test cases which change the shared logger should have a tearDown function to restore it to a normal state afterwards.

tgoyne (approved):

PR #8225, opened 2023-05-03
Normalize Connection parameters for sync

dianaafanador3:
Normalize Connection parameters for sync


Realm/RLMApp.mm
```diff
@@ -128,25 +134,35 @@ - (instancetype)initWithBaseURL:(nullable NSString *)baseURL
...
         return self;
     }
     return nil;
 }
 
+static void RLMConfigureSyncConnectionParameters(realm::app::App::Config config) {
+    // Anonymized BundleId
+    NSString *bundleId = [[NSBundle mainBundle] bundleIdentifier];
+    NSData *bundleIdData = [bundleId dataUsingEncoding:NSUTF8StringEncoding];
+    RLMNSStringToStdString(config.device_info.bundle_id, RLMHashData(bundleIdData.bytes, bundleIdData.length));
```
dianaafanador3, line 147: We are anonymizing the bundleId to match the reported anonymizedBundleId reported on the builder metrics.

dianaafanador3:
Changing this to be merged to Core's 13.12.0 PR

Realm/RLMApp.mm
```diff
@@ -128,21 +134,35 @@ - (instancetype)initWithBaseURL:(nullable NSString *)baseURL
...
-        RLMNSStringToStdString(_config.device_info.platform_version,
-                               [processInfo operatingSystemVersionString] ?: @"unknown");
-        RLMNSStringToStdString(_config.device_info.sdk_version, REALM_COCOA_VERSION);
+        RLMConfigureSyncConnectionParameters(_config);
         return self;
     }
     return nil;
 }
 
+static void RLMConfigureSyncConnectionParameters(realm::app::App::Config config) {
```
tgoyne, line 143: This function doesn't actually do anything. It needs to take a `realm::app::App::Config&` so that it's not just mutating a local variable.

This is now doing some not totally trivial things, so it'd be nice to have a test. I don't think there's a _great_ way to test it, but maybe you could do something like override the logger before opening a synchronized Realm and verifying that all of the expected values are present in the logs.

dianaafanador3: I added a test for this, using the `Logger`, while on it I found a bug which is fixed on this PR

Realm/RLMApp.mm
```diff
@@ -128,21 +135,35 @@ - (instancetype)initWithBaseURL:(nullable NSString *)baseURL
...
+    RLMNSStringToStdString(config.device_info.sdk_version, REALM_COCOA_VERSION);
+
+    auto processInfo = [NSProcessInfo processInfo];
+    RLMNSStringToStdString(config.device_info.platform_version,
+                           [processInfo operatingSystemVersionString] ?: @"unknown");
+
+    RLMNSStringToStdString(config.device_info.framework_version, @__clang_version__);
+
+#ifdef REALM_UIDEVICE_AVAILABLE
+    RLMNSStringToStdString(config.device_info.device_name, [UIDevice currentDevice].model);
```
tgoyne, line 160: This seems like it won't work for CocoaPods or SPM installations?

dianaafanador3: @tgoyne This is defined at the beginning of this file (on line 24) and can be replaced by `#if __has_include(<UIKit/UIDevice.h>)` if it is clearer

tgoyne: The problem is that UIKit is only being linked for the framework build, so this field will not be set for the other installation methods.

dianaafanador3: I thought linking it on the `xcconfig` file will work for all the installation method.

dianaafanador3: I'm linking UIKit on the podspec now, tested later on macos and ios.

Realm/Tests/RealmTests.mm
```diff
@@ -3023,12 +3023,58 @@ - (void)testCustomLoggerLogMessage {
...
+        if (end.location != NSNotFound)
+        {
+            param = [param substringToIndex:end.location];
+        }
+    }
+    XCTAssertTrue([param length] > 0);
+}
+
+- (void)testSyncConnectionMetrics {
+    __block NSString *logs = @"";
```
tgoyne, line 3061: Using a NSMutableString would simplify this.

Realm/Tests/RealmTests.mm
```diff
@@ -3023,12 +3023,58 @@ - (void)testCustomLoggerLogMessage {
...
+    // We don't even need the login to succeed, we only want for the logger
+    // to log the values on device info after trying to login.
+    [app loginWithCredential:[RLMCredentials anonymousCredentials] completion:^(RLMUser * _Nullable, NSError * _Nullable) {}];
+    // Verifying that this values are set on device_info, some of this are set by
+    // the sdk (sdk, sdk version, platform version) and some by core (platform, core version).
+    [self assertString:logs betweenString:@"platform: " endString:@" version:"];
+    [self assertString:logs betweenString:@"version: " endString:@" - sdk:"];
+    [self assertString:logs betweenString:@"sdk: " endString:@" - sdk version:"];
+    [self assertString:logs betweenString:@"sdk version: " endString:@" - core"];
+    [self assertString:logs betweenString:@"core version: " endString:@"  ."];
```
tgoyne, line 3078: This doesn't appear to actually be checking anything about the values logged? It appears to just be a roundabout way of checking that `logs` contains the two other strings in the correct order, so all that it's testing is that core hasn't changed what order it logs things.

dianaafanador3: You are right, I was overthinking this, knowing this will change depending on the CI Target Device. I changed the test

tgoyne (approved):

PR #8236, opened 2023-05-18
Replace Xcode 14.3 with Xcode 14.3.1

tgoyne:


build.sh
```diff
@@ -445,17 +445,18 @@ case "$COMMAND" in
...
-        -exec sed -i '' 's/RealmSwift.Error[[:>:]]/RealmSwift.Realm.Error/g' {} \; \
-        -exec sed -i '' 's/extension Error/extension Realm.Error/g' {} \; \
-        -exec sed -i '' 's/RealmSwift.AsyncOpenTask/RealmSwift.Realm.AsyncOpenTask/g' {} \; \
-        -exec sed -i '' 's/RealmSwift.UpdatePolicy/RealmSwift.Realm.UpdatePolicy/g' {} \; \
-        -exec sed -i '' 's/RealmSwift.Notification[[:>:]]/RealmSwift.Realm.Notification/g' {} \; \
-        -exec sed -i '' 's/RealmSwift.OpenBehavior/RealmSwift.Realm.OpenBehavior/g' {} \; \
-        -exec sed -i '' 's/τ_1_0/V/g' {} \; # Generics will use τ_1_0 which needs to be changed to the correct type name.
+        find build/RealmSwift.xcframework -name "*.swiftinterface" \
+            -exec sed -i '' 's/Realm\.//g' {} \; \
+            -exec sed -i '' 's/import Private/import Realm.Private\nimport Realm.Swift/g' {} \; \
```
dianaafanador3, line 450: I'm curious about what was the issue in Xcode 14.3 which causes this change, and how is solved by this? 🤗

tgoyne: The .swiftinterface generated by 14.3 has a lot of incidental differences from the one generated by 14.2, but a non-incidental one is that it's just missing `import Realm.Swift` entirely and I couldn't figure out why.

PR #8244, opened 2023-05-18
Results.subscribe

ericjordanmossman:
Adds an experimental async `Results.subscribe` and `.unsubscribe` in Swift. And adds `[RLMResults subscribeWithName:waitForSyncMode:onQueue:timeout:completion]` and `unsubscribe` in Obj-c.
```
public func subscribe(name: String? = nil, waitForSync: WaitForSyncMode = .onCreation, timeout: TimeInterval? = nil)
 async throws -> Results<Element>

public func unsubscribe()
```
```
- (void)subscribeWithName:(NSString *_Nullable)name
          waitForSyncMode:(RLMWaitForSyncMode)waitForSyncMode
                  onQueue:(dispatch_queue_t _Nullable)queue
                  timeout:(NSTimeInterval)timeout
               completion:(RLMResultsCompletionBlock)completion

- (void)unsubscribe
```
`Results.subscribe` will create a SyncSubscription with a query which matches the local filter on the Results at the time `subscribe` is called. The API is an attempt to ease overhead maintaining distinct local queries made on an `Results` and server queries created via `SubscriptionSet.update`. 

The subscription API opens a write transaction to commit subscription updates. Opening a write transaction and diffing the subscription against the server is not an unsubstanial workload. Therefore using the subscription API for large batch subscriptions is *discouraged*.

Using both the current API and this experimental API is entirely possible and encouraged. The current `subscription` API is well-suited for creating batch subscriptions for commonly shared data early in the application setup.
This can be combined with dynamic changes via the experimental subscription API to handle user input.

This change also adds an `unnamedOnly` bool flag to `SyncSubscriptionSet.removeAll`. And adds `[RLMSyncSubscriptionSet removeAllUnnamedSubscriptions]`. These can be used for periodic clean up of unnamed subscriptions that could accrue during app usage.

### DIFF
--- a/Realm/ObjectServerTests/RLMFlexibleSyncServerTests.mm
+++ b/Realm/ObjectServerTests/RLMFlexibleSyncServerTests.mm
@@ -1165,7 +1165,7 @@
 
     [[realm syncSession] suspend];
     XCTestExpectation *ex = [self expectationWithDescription:@"expect timeout"];
-    NSTimeInterval ti = 2.0;
+    NSTimeInterval timeout = 2.0;
     RLMResults *res = [[Person allObjectsInRealm:realm] objectsWhere:@"age >= 20"];
     [res subscribeWithName:@"20up" waitForSync:RLMWaitForSyncModeAlways onQueue:dispatch_get_main_queue() timeout:2.0 completion:^(RLMResults *results, NSError *error) {
         XCTAssert(error);
--- a/Realm/ObjectServerTests/RLMFlexibleSyncServerTests.mm
+++ b/Realm/ObjectServerTests/RLMFlexibleSyncServerTests.mm
@@ -1167,9 +1167,9 @@
     XCTestExpectation *ex = [self expectationWithDescription:@"expect timeout"];
     NSTimeInterval timeout = 2.0;
     RLMResults *res = [[Person allObjectsInRealm:realm] objectsWhere:@"age >= 20"];
-    [res subscribeWithName:@"20up" waitForSync:RLMWaitForSyncModeAlways onQueue:dispatch_get_main_queue() timeout:2.0 completion:^(RLMResults *results, NSError *error) {
+    [res subscribeWithName:@"20up" waitForSync:RLMWaitForSyncModeAlways onQueue:dispatch_get_main_queue() timeout:timeout completion:^(RLMResults *results, NSError *error) {
         XCTAssert(error);
-        NSString *expectedDesc = [NSString stringWithFormat:@"Waiting for update timed out after %.01f seconds.", ti];
+        NSString *expectedDesc = [NSString stringWithFormat:@"Waiting for update timed out after %.01f seconds.", timeout];
         XCTAssert([error.localizedDescription isEqualToString:expectedDesc]);
         XCTAssertNil(results);
         [ex fulfill];
@@ -1186,8 +1186,6 @@
     }
     XCTAssertEqual(realm.subscriptions.state, RLMSyncSubscriptionStateComplete);
 }
-
-- (void)testUnsubscribe {}
 
 #if 0 // FIXME: this is no longer an error and needs to be updated to something which is
 - (void)testFlexibleSyncInitialSubscriptionThrowsError {
--- a/Realm/ObjectServerTests/RLMFlexibleSyncServerTests.mm
+++ b/Realm/ObjectServerTests/RLMFlexibleSyncServerTests.mm
@@ -1174,7 +1174,7 @@
         [ex fulfill];
     }];
     [self waitForExpectationsWithTimeout:5.0 handler:nil];
-    XCTAssertEqual(realm.subscriptions.count, 0);
+    XCTAssertEqual(realm.subscriptions.count, 0U);
 }
 
 - (void)testSubscribeOnQueue {
--- a/Realm/ObjectServerTests/RLMFlexibleSyncServerTests.mm
+++ b/Realm/ObjectServerTests/RLMFlexibleSyncServerTests.mm
@@ -1149,6 +1149,34 @@
     [self waitForExpectationsWithTimeout:5.0 handler:nil];
 }
 
+- (void)testUnsubscribeWithinBlock {
+    bool didPopulate = [self populateData:^(RLMRealm *realm) {
+        Person *person = [[Person alloc] initWithPrimaryKey:[RLMObjectId objectId]
+                                                        age:30
+                                                  firstName:@"Joe"
+                                                   lastName:@"Doe"];
+        person.partition = NSStringFromSelector(_cmd);
+        [realm addObject:person];
+    }];
+    if (!didPopulate) {
+        return;
+    }
+
+    RLMRealm *realm = [self getFlexibleSyncRealm:_cmd];
+    XCTAssertNotNil(realm);
+    CHECK_COUNT(0, Person, realm);
+
+    XCTestExpectation *ex = [self expectationWithDescription:@"wait for download"];
+    [[[Person allObjectsInRealm:realm] objectsWhere:@"lastName == 'Doe'"] subscribeWithName:@"unknown" onQueue:dispatch_get_main_queue() completion:^(RLMResults *results, NSError *error) {
+        XCTAssertNil(error);
+        XCTAssertEqual(results.count, 1U);
+        [results unsubscribe];
+        [ex fulfill];
+    }];
+    [self waitForExpectationsWithTimeout:5.0 handler:nil];
+    XCTAssertEqual(realm.subscriptions.count, 0);
+}
+
 - (void)testSubscribeOnQueue {
     bool didPopulate = [self populateData:^(RLMRealm *realm) {
         Person *person = [[Person alloc] initWithPrimaryKey:[RLMObjectId objectId]
--- a/Realm/ObjectServerTests/RLMFlexibleSyncServerTests.mm
+++ b/Realm/ObjectServerTests/RLMFlexibleSyncServerTests.mm
@@ -1146,7 +1146,6 @@
         XCTAssertEqual(results.count, 1U);
         [ex fulfill];
     }];
-    XCTAssertEqual(realm.subscriptions.count, 1UL);
     [self waitForExpectationsWithTimeout:5.0 handler:nil];
 }
 
@@ -1173,13 +1172,15 @@
         XCTAssertNotNil(realm);
         CHECK_COUNT(0, Person, realm);
 
-        [[[Person allObjectsInRealm:realm] objectsWhere:@"lastName == 'Loren'"] subscribeWithCompletion:^(RLMResults *results, NSError *error) {
+
+        [[[Person allObjectsInRealm:realm] objectsWhere:@"lastName == 'Loren'"]
+         subscribeWithCompletionOnQueue:dispatch_get_main_queue()
+         completionBlock:^(RLMResults *results, NSError *error) {
             XCTAssertNil(error);
             XCTAssertEqual(results.count, 1U);
+            XCTAssertEqual(realm.subscriptions.count, 1UL);
             [ex fulfill];
-        } onQueue:dispatch_get_main_queue()];
-
-        XCTAssertEqual(realm.subscriptions.count, 1UL);
+        }];
     }];
     [self waitForExpectationsWithTimeout:5.0 handler:nil];
 }
--- a/Realm/ObjectServerTests/RLMFlexibleSyncServerTests.mm
+++ b/Realm/ObjectServerTests/RLMFlexibleSyncServerTests.mm
@@ -1106,7 +1106,6 @@
     syncConfig.cancelAsyncOpenOnNonFatalErrors = true;
     config.syncConfiguration = syncConfig;
 
-
     // Set delay above the timeout so it should fail
     proxy.delay = 2.0;
 
--- a/Realm/ObjectServerTests/RLMFlexibleSyncServerTests.mm
+++ b/Realm/ObjectServerTests/RLMFlexibleSyncServerTests.mm
@@ -1167,7 +1167,7 @@
     XCTestExpectation *ex = [self expectationWithDescription:@"expect timeout"];
     NSTimeInterval ti = 2.0;
     RLMResults *res = [[Person allObjectsInRealm:realm] objectsWhere:@"age >= 20"];
-    [res subscribeWithName:@"20up" waitForSyncMode:RLMWaitForSyncModeAlways onQueue:dispatch_get_main_queue() timeout:2.0 completion:^(RLMResults *results, NSError *error) {
+    [res subscribeWithName:@"20up" waitForSync:RLMWaitForSyncModeAlways onQueue:dispatch_get_main_queue() timeout:2.0 completion:^(RLMResults *results, NSError *error) {
         XCTAssert(error);
         NSString *expectedDesc = [NSString stringWithFormat:@"Waiting for subscribed data timed out after %.01f seconds.", ti];
         XCTAssert([error.localizedDescription isEqualToString:expectedDesc]);
--- a/Realm/ObjectServerTests/RLMFlexibleSyncServerTests.mm
+++ b/Realm/ObjectServerTests/RLMFlexibleSyncServerTests.mm
@@ -1200,17 +1200,21 @@
         XCTAssertNotNil(realm);
         CHECK_COUNT(0, Person, realm);
 
-
         [[[Person allObjectsInRealm:realm] objectsWhere:@"lastName == 'Loren'"]
          subscribeWithCompletionOnQueue:dispatch_get_main_queue()
          completion:^(RLMResults *results, NSError *error) {
             XCTAssertNil(error);
-            XCTAssertEqual(results.count, 1U);
-            XCTAssertEqual(realm.subscriptions.count, 1UL);
+            XCTAssertEqual(results.realm.subscriptions.count, 1UL);
+            XCTAssertEqual(results.realm.subscriptions.state, RLMSyncSubscriptionStateComplete);
+            CHECK_COUNT(1, Person, results.realm);
             [ex fulfill];
         }];
     }];
-    [self waitForExpectationsWithTimeout:5.0 handler:nil];
+    [self waitForExpectationsWithTimeout:20.0 handler:nil];
+
+    RLMRealm *realm = [RLMRealm realmWithConfiguration:config error:nil];
+    XCTAssertEqual(realm.subscriptions.count, 1UL);
+    CHECK_COUNT(1, Person, realm);
 }
 
 - (void)testSubscribeWithNameAndTimeout {
--- a/Realm/ObjectServerTests/RLMFlexibleSyncServerTests.mm
+++ b/Realm/ObjectServerTests/RLMFlexibleSyncServerTests.mm
@@ -1169,7 +1169,7 @@
     RLMResults *res = [[Person allObjectsInRealm:realm] objectsWhere:@"age >= 20"];
     [res subscribeWithName:@"20up" waitForSync:RLMWaitForSyncModeAlways onQueue:dispatch_get_main_queue() timeout:2.0 completion:^(RLMResults *results, NSError *error) {
         XCTAssert(error);
-        NSString *expectedDesc = [NSString stringWithFormat:@"Waiting for subscribed data timed out after %.01f seconds.", ti];
+        NSString *expectedDesc = [NSString stringWithFormat:@"Waiting for update timed out after %.01f seconds.", ti];
         XCTAssert([error.localizedDescription isEqualToString:expectedDesc]);
         XCTAssertNil(results);
         [ex fulfill];
--- a/Realm/ObjectServerTests/RLMFlexibleSyncServerTests.mm
+++ b/Realm/ObjectServerTests/RLMFlexibleSyncServerTests.mm
@@ -877,12 +877,14 @@
         [subs addSubscriptionWithClassName:Dog.className
                                      where:@"breed == 'Labradoodle' and partition == %@", NSStringFromSelector(_cmd)];
     }];
+    XCTAssertEqual(realm.subscriptions.count, 3U);
     CHECK_COUNT(2, Person, realm);
     CHECK_COUNT(1, Dog, realm);
 
     [self writeQueryAndCompleteForRealm:realm block:^(RLMSyncSubscriptionSet *subs) {
         [subs removeAllUnnamedSubscriptions];
     }];
+    XCTAssertEqual(realm.subscriptions.count, 1U);
     CHECK_COUNT(1, Person, realm);
     CHECK_COUNT(0, Dog, realm);
 }
@@ -1140,11 +1142,11 @@
     CHECK_COUNT(0, Person, realm);
 
     XCTestExpectation *ex = [self expectationWithDescription:@"wait for download"];
-    [[[Person allObjectsInRealm:realm] objectsWhere:@"lastName == 'Epstein'"] subscribeWithCompletion:^(RLMResults *results, NSError *error) {
+    [[[Person allObjectsInRealm:realm] objectsWhere:@"lastName == 'Epstein'"] subscribeWithName:@"5thBeatle" onQueue:dispatch_get_main_queue() completion:^(RLMResults *results, NSError *error) {
         XCTAssertNil(error);
         XCTAssertEqual(results.count, 1U);
         [ex fulfill];
-    } onQueue:dispatch_get_main_queue()];
+    }];
     XCTAssertEqual(realm.subscriptions.count, 1UL);
     [self waitForExpectationsWithTimeout:5.0 handler:nil];
 }
--- a/Realm/ObjectServerTests/RLMFlexibleSyncServerTests.mm
+++ b/Realm/ObjectServerTests/RLMFlexibleSyncServerTests.mm
@@ -1175,7 +1175,7 @@
 
         [[[Person allObjectsInRealm:realm] objectsWhere:@"lastName == 'Loren'"]
          subscribeWithCompletionOnQueue:dispatch_get_main_queue()
-         completionBlock:^(RLMResults *results, NSError *error) {
+         completion:^(RLMResults *results, NSError *error) {
             XCTAssertNil(error);
             XCTAssertEqual(results.count, 1U);
             XCTAssertEqual(realm.subscriptions.count, 1UL);
--- a/Realm/ObjectServerTests/RLMFlexibleSyncServerTests.mm
+++ b/Realm/ObjectServerTests/RLMFlexibleSyncServerTests.mm
@@ -1150,6 +1150,40 @@
     [self waitForExpectationsWithTimeout:5.0 handler:nil];
 }
 
+- (void)testSubscribeOnQueue {
+    bool didPopulate = [self populateData:^(RLMRealm *realm) {
+        Person *person = [[Person alloc] initWithPrimaryKey:[RLMObjectId objectId]
+                                                        age:30
+                                                  firstName:@"Sophia"
+                                                   lastName:@"Loren"];
+        person.partition = NSStringFromSelector(_cmd);
+        [realm addObject:person];
+    }];
+    if (!didPopulate) {
+        return;
+    }
+
+    RLMUser *user = [self flexibleSyncUser:_cmd];
+    RLMRealmConfiguration *config = [user flexibleSyncConfiguration];
+    config.objectClasses = @[Person.self];
+
+    XCTestExpectation *ex = [self expectationWithDescription:@"wait for download"];
+    [self dispatchAsyncAndWait:^{
+        RLMRealm *realm = [RLMRealm realmWithConfiguration:config error:nil];
+        XCTAssertNotNil(realm);
+        CHECK_COUNT(0, Person, realm);
+
+        [[[Person allObjectsInRealm:realm] objectsWhere:@"lastName == 'Loren'"] subscribeWithCompletion:^(RLMResults *results, NSError *error) {
+            XCTAssertNil(error);
+            XCTAssertEqual(results.count, 1U);
+            [ex fulfill];
+        } onQueue:dispatch_get_main_queue()];
+
+        XCTAssertEqual(realm.subscriptions.count, 1UL);
+    }];
+    [self waitForExpectationsWithTimeout:5.0 handler:nil];
+}
+
 - (void)testSubscribeWithNameAndTimeout {
     bool didPopulate = [self populateData:^(RLMRealm *realm) {
         [self createPeople:realm partition:_cmd];
--- a/Realm/ObjectServerTests/RLMFlexibleSyncServerTests.mm
+++ b/Realm/ObjectServerTests/RLMFlexibleSyncServerTests.mm
@@ -1092,7 +1092,12 @@
 
 - (void)testSubscribeWithName {
     bool didPopulate = [self populateData:^(RLMRealm *realm) {
-        [self createPeople:realm partition:_cmd];
+        Person *person = [[Person alloc] initWithPrimaryKey:[RLMObjectId objectId]
+                                                        age:30
+                                                  firstName:@"Brian"
+                                                   lastName:@"Epstein"];
+        person.partition = NSStringFromSelector(_cmd);
+        [realm addObject:person];
     }];
     if (!didPopulate) {
         return;
@@ -1103,11 +1108,11 @@
     CHECK_COUNT(0, Person, realm);
 
     XCTestExpectation *ex = [self expectationWithDescription:@"wait for download"];
-    [[[Person allObjectsInRealm:realm] objectsWhere:@"age > 20"] subscribeWithCompletion:^(RLMResults* results, NSError* error) {
+    [[[Person allObjectsInRealm:realm] objectsWhere:@"lastName == 'Epstein'"] subscribeWithCompletion:^(RLMResults *results, NSError *error) {
         XCTAssertNil(error);
-        XCTAssertEqual(results.count, 2U);
+        XCTAssertEqual(results.count, 1U);
         [ex fulfill];
-    }];
+    } onQueue:dispatch_get_main_queue()];
     XCTAssertEqual(realm.subscriptions.count, 1UL);
     [self waitForExpectationsWithTimeout:5.0 handler:nil];
 }
@@ -1127,11 +1132,8 @@
     [[realm syncSession] suspend];
     XCTestExpectation *ex = [self expectationWithDescription:@"expect timeout"];
     NSTimeInterval ti = 2.0;
-    RLMResults *res = [[Person allObjectsInRealm:realm] objectsWhere:@"age > 20"];
-    [res subscribeWithName:@"20up"
-           waitForSyncMode:RLMWaitForSyncModeAlways
-                   timeout:ti
-                completion:^(RLMResults *results, NSError *error) {
+    RLMResults *res = [[Person allObjectsInRealm:realm] objectsWhere:@"age >= 20"];
+    [res subscribeWithName:@"20up" waitForSyncMode:RLMWaitForSyncModeAlways onQueue:dispatch_get_main_queue() timeout:2.0 completion:^(RLMResults *results, NSError *error) {
         XCTAssert(error);
         NSString *expectedDesc = [NSString stringWithFormat:@"Waiting for subscribed data timedout after %f seconds.", ti];
         XCTAssert([error.localizedDescription isEqualToString:expectedDesc]);
@@ -1141,6 +1143,8 @@
     XCTAssertEqual(realm.subscriptions.count, 1UL);
     [self waitForExpectationsWithTimeout:5.0 handler:nil];
 
+    // resume session and wait for complete
+    // otherwise test will not tear down successfully
     [[realm syncSession] resume];
     NSDate * start = [[NSDate alloc] init];
     while (realm.subscriptions.state != RLMSyncSubscriptionStateComplete && start.timeIntervalSinceNow > -10.0) {
--- a/Realm/ObjectServerTests/RLMFlexibleSyncServerTests.mm
+++ b/Realm/ObjectServerTests/RLMFlexibleSyncServerTests.mm
@@ -1090,6 +1090,67 @@
     [proxy stop];
 }
 
+- (void)testSubscribeWithName {
+    bool didPopulate = [self populateData:^(RLMRealm *realm) {
+        [self createPeople:realm partition:_cmd];
+    }];
+    if (!didPopulate) {
+        return;
+    }
+
+    RLMRealm *realm = [self getFlexibleSyncRealm:_cmd];
+    XCTAssertNotNil(realm);
+    CHECK_COUNT(0, Person, realm);
+
+    XCTestExpectation *ex = [self expectationWithDescription:@"wait for download"];
+    [[[Person allObjectsInRealm:realm] objectsWhere:@"age > 20"] subscribeWithCompletion:^(RLMResults* results, NSError* error) {
+        XCTAssertNil(error);
+        XCTAssertEqual(results.count, 2U);
+        [ex fulfill];
+    }];
+    XCTAssertEqual(realm.subscriptions.count, 1UL);
+    [self waitForExpectationsWithTimeout:5.0 handler:nil];
+}
+
+- (void)testSubscribeWithNameAndTimeout {
+    bool didPopulate = [self populateData:^(RLMRealm *realm) {
+        [self createPeople:realm partition:_cmd];
+    }];
+    if (!didPopulate) {
+        return;
+    }
+
+    RLMRealm *realm = [self getFlexibleSyncRealm:_cmd];
+    XCTAssertNotNil(realm);
+    CHECK_COUNT(0, Person, realm);
+
+    [[realm syncSession] suspend];
+    XCTestExpectation *ex = [self expectationWithDescription:@"expect timeout"];
+    NSTimeInterval ti = 2.0;
+    RLMResults *res = [[Person allObjectsInRealm:realm] objectsWhere:@"age > 20"];
+    [res subscribeWithName:@"20up"
+           waitForSyncMode:RLMWaitForSyncModeAlways
+                   timeout:ti
+                completion:^(RLMResults *results, NSError *error) {
+        XCTAssert(error);
+        NSString *expectedDesc = [NSString stringWithFormat:@"Waiting for subscribed data timedout after %f seconds.", ti];
+        XCTAssert([error.localizedDescription isEqualToString:expectedDesc]);
+        XCTAssertNil(results);
+        [ex fulfill];
+    }];
+    XCTAssertEqual(realm.subscriptions.count, 1UL);
+    [self waitForExpectationsWithTimeout:5.0 handler:nil];
+
+    [[realm syncSession] resume];
+    NSDate * start = [[NSDate alloc] init];
+    while (realm.subscriptions.state != RLMSyncSubscriptionStateComplete && start.timeIntervalSinceNow > -10.0) {
+        sleep(1);
+    }
+    XCTAssertEqual(realm.subscriptions.state, RLMSyncSubscriptionStateComplete);
+}
+
+- (void)testUnsubscribe {}
+
 #if 0 // FIXME: this is no longer an error and needs to be updated to something which is
 - (void)testFlexibleSyncInitialSubscriptionThrowsError {
     RLMUser *user = [self flexibleSyncUser:_cmd];
--- a/Realm/ObjectServerTests/RLMFlexibleSyncServerTests.mm
+++ b/Realm/ObjectServerTests/RLMFlexibleSyncServerTests.mm
@@ -855,6 +855,38 @@
     CHECK_COUNT(1, Dog, realm);
 }
 
+- (void)testRemoveAllUnnamedSubscriptions {
+    bool didPopulate = [self populateData:^(RLMRealm *realm) {
+        [self createPeople:realm partition:_cmd];
+        [self createDog:realm partition:_cmd];
+    }];
+    if (!didPopulate) {
+        return;
+    }
+
+    RLMRealm *realm = [self getFlexibleSyncRealm:_cmd];
+    XCTAssertNotNil(realm);
+    CHECK_COUNT(0, Person, realm);
+
+    [self writeQueryAndCompleteForRealm:realm block:^(RLMSyncSubscriptionSet *subs) {
+        [subs addSubscriptionWithClassName:Person.className
+                                     where:@"age > 20 and partition == %@", NSStringFromSelector(_cmd)];
+        [subs addSubscriptionWithClassName:Person.className
+                          subscriptionName:@"person_age_2"
+                                     where:@"firstName == 'firstname_1' and partition == %@", NSStringFromSelector(_cmd)];
+        [subs addSubscriptionWithClassName:Dog.className
+                                     where:@"breed == 'Labradoodle' and partition == %@", NSStringFromSelector(_cmd)];
+    }];
+    CHECK_COUNT(2, Person, realm);
+    CHECK_COUNT(1, Dog, realm);
+
+    [self writeQueryAndCompleteForRealm:realm block:^(RLMSyncSubscriptionSet *subs) {
+        [subs removeAllUnnamedSubscriptions];
+    }];
+    CHECK_COUNT(1, Person, realm);
+    CHECK_COUNT(0, Dog, realm);
+}
+
 - (void)testFlexibleSyncUpdateQuery {
     bool didPopulate = [self populateData:^(RLMRealm *realm) {
         [self createPeople:realm partition:_cmd];
@@ -1135,7 +1167,7 @@
     RLMResults *res = [[Person allObjectsInRealm:realm] objectsWhere:@"age >= 20"];
     [res subscribeWithName:@"20up" waitForSyncMode:RLMWaitForSyncModeAlways onQueue:dispatch_get_main_queue() timeout:2.0 completion:^(RLMResults *results, NSError *error) {
         XCTAssert(error);
-        NSString *expectedDesc = [NSString stringWithFormat:@"Waiting for subscribed data timedout after %f seconds.", ti];
+        NSString *expectedDesc = [NSString stringWithFormat:@"Waiting for subscribed data timed out after %.01f seconds.", ti];
         XCTAssert([error.localizedDescription isEqualToString:expectedDesc]);
         XCTAssertNil(results);
         [ex fulfill];
--- a/Realm/ObjectServerTests/SwiftFlexibleSyncServerTests.swift
+++ b/Realm/ObjectServerTests/SwiftFlexibleSyncServerTests.swift
@@ -1394,12 +1394,10 @@ extension SwiftFlexibleSyncServerTests {
         let results0 = try await realm.objects(SwiftPerson.self).where { $0.age >= 8 }.subscribe()
         let ex = XCTestExpectation(description: "no attempt to re-create subscription, returns immediately")
         realm.syncSession!.suspend()
-        Task {
-            _ = try await realm.objects(SwiftPerson.self).where { $0.age >= 8 }.subscribe()
-            _ = try await results0.subscribe()
-            XCTAssertEqual(realm.subscriptions.count, 1)
-            ex.fulfill()
-        }
+        _ = try await realm.objects(SwiftPerson.self).where { $0.age >= 8 }.subscribe()
+        _ = try await results0.subscribe()
+        XCTAssertEqual(realm.subscriptions.count, 1)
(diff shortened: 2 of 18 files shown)


ericjordanmossman:
Needs changelog entry.
Need to figure out what's needed for beta release. Similar to something like [this release](https://github.com/realm/realm-swift/releases/tag/v10.8.0-beta.2).
From what I can tell there aren't any macros I need to add? Just changes in the release pipeline?

Realm/RLMResults.h
```diff
@@ -20,6 +20,25 @@
...
+ Determines wait for download behavior when subscribing on RLMResults.
+ @see ``[RLMResults subscribeWithName:waitForSyncMode:onQueue:completion:]``
+*/
+typedef NS_ENUM(NSUInteger, RLMWaitForSyncMode) {
+    /// `subscribeWithName` will return once matching objects are downloaded
+    /// from the server only when the subscription is created the first time. If the
+    /// subscriptions already exists, the method will return without waiting for new downloads.
+    RLMWaitForSyncModeOnCreation,
+    /// `subscribeWithName` will always for downloads before returning.
+    /// The method will not return in this mode unless an internet connection is established or a timeout is set.
```
dianaafanador3, line 36: This probably needs to be warned to the user on the `subscribe` APIs docstings. 
Can we have always a timeout and set a default value, the above warning to never happen

Realm/RLMResults.h
```diff
@@ -20,6 +20,25 @@
 
 RLM_HEADER_AUDIT_BEGIN(nullability, sendability)
 
+/// A block type used for APIs which asynchronously return an `RLMResults`.
```
dianaafanador3, line 23:
```suggestion
/// A block type used for APIs which asynchronously return a `Results`.
```

Realm/RLMResults.h
```diff
@@ -408,6 +427,174 @@ __attribute__((warn_unused_result));
                                                          NSError *_Nullable error))block
                                       keyPaths:(nullable NSArray<NSString *> *)keyPaths
 __attribute__((warn_unused_result));
+
+#pragma mark - Flexible Sync
+
+/**
+ Creates an RLMSyncSubscription matching the RLMResults' local filter.
+
+ This method opens a write transaction that creates or updates a subscription.
```
dianaafanador3, line 436: I'll add this as a note and a link to that API

Realm/RLMResults.h
```diff
@@ -408,6 +427,174 @@ __attribute__((warn_unused_result));
...
+ be removed from the realm.
+
+ Calling unsubscribe on a RLMResults does not remove the local filter from the RLMResults.
+ After calling unsubscribe, RLMResults may still contain objects because
+ other subscriptions may exist in the RLMRealm's subscription set.
+
+ In order for a named subscription to be removed, the RLMResults
+ must have previously created the subscription. For example:
+ ```
+ let results1 = try await realm.objects(Dog.self).where { $0.age >= 2 }.subscribe(name: "dog_2andOver")
```
dianaafanador3, line 589: this code snippet is in Swift, change it to objective-C or remove it

dianaafanador3 (commented):
This needs a changelog, even if we are exposing this only in Preview.

I have some doubts about some corner cases for this API. What will happen if the Results query is updated and then subscribes again to the new query. Than unsubscribes after this. This will leave an orphan subscription which we are not gonna be able to unsubscribe.
Maybe allowing subscribing once for each Result will be a good solution for this.

RealmSwift/Results.swift
```diff
@@ -147,6 +154,77 @@ extension Projection: KeypathSortable {}
...
     }
 
+    // MARK: Flexible Sync
+
+#if swift(>=5.8)
+    /**
+     Creates a SyncSubscription matching the Results' local filter.
+
+     This method opens a write transaction that creates or updates a subscription.
+     It's advised to *not* use this method to batch multiple subscription changes
```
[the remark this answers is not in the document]
ericjordanmossman: In other words, don't do something like:
for predicate in predicateToSubscribe {
  realm.objects(MyClass.self).filter(predicate).subscribe()
}
Better to use current API to create/update many subscriptions

RealmSwift/Results.swift
```diff
@@ -147,6 +154,77 @@ extension Projection: KeypathSortable {}
...
+     This method opens a write transaction that removes a subscription.
+     It is advised to *not* use this method to batch multiple subscription changes
+     to the server.
+     For batch updates use `SyncSubscription.update`
+
+     The method returns after committing the subcsription removal to the realm's
+     local subscriptionset. Calling this method will not wait for objects to
+     be removed from the realm.
+
+     Calling unsubscribe on a Results does not remove the local filter from the Results.
```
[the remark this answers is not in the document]
ericjordanmossman: Nothing -- attempts to unsubscribe from a non-existing subscription. Adding a test called `testUnsubscribeNoExistingMatch` in next commit.
Another case is `testUnsubscribeWithoutSubscriptionExistingNamed` (renamed from testUnsubscribeWithoutSubscription) covers the named case.

Realm/ObjectServerTests/SwiftFlexibleSyncServerTests.swift
```diff
@@ -1315,6 +1356,318 @@ extension SwiftFlexibleSyncServerTests {
...
+    }
+
+    @MainActor
+    func testSubscribeSameQueryNoName() async throws {
+        try await populateSwiftPerson()
+        let realm = try openFlexibleSyncRealm()
+
+        let results0 = try await realm.objects(SwiftPerson.self).where { $0.age >= 8 }.subscribe()
+        let ex = XCTestExpectation(description: "no attempt to re-create subscription, returns immediately")
+        Task {
```
dianaafanador3, line 1404: Any reason to have this in a Task?

Realm/ObjectServerTests/SwiftFlexibleSyncServerTests.swift
```diff
@@ -1315,6 +1356,318 @@ extension SwiftFlexibleSyncServerTests {
...
+        }
+    }
+
+    @MainActor
+    func testUnsubscribe() async throws {
+        try await populateSwiftPerson()
+        let realm = try openFlexibleSyncRealm()
+
+        let results1 = try await realm.objects(SwiftPerson.self).where { $0.lastName == "lastname_3" }.subscribe()
+        XCTAssertEqual(realm.subscriptions.count, 1)
```
dianaafanador3, line 1471: A more complete test will check the results data after the unsubscription

Realm/ObjectServerTests/SwiftFlexibleSyncServerTests.swift
```diff
@@ -1315,6 +1356,318 @@ extension SwiftFlexibleSyncServerTests {
...
+    func testUnsubscribeAfterReassign() async throws {
+        try await populateSwiftPerson()
+        let realm = try openFlexibleSyncRealm()
+
+        var results0 = try await realm.objects(SwiftPerson.self).where { $0.age >= 8 }.subscribe()
+        XCTAssertEqual(results0.count, 3)
+        XCTAssertEqual(realm.subscriptions.count, 1)
+        results0 = try await results0.where { $0.age < 8 }.subscribe() // subscribes to "age >= 8 && age < 8" because that's the local query
+        XCTAssertEqual(results0.count, 0)
+        XCTAssertEqual(realm.subscriptions.count, 2) // "age >= 8" and "age >= 8 && age < 8"
```
dianaafanador3, line 1486:
```suggestion
        XCTAssertEqual(realm.subscriptions.count, 2) // "age >= 8" OR "age >= 8 && age < 8"
```

Realm/RLMResults.mm
```diff
@@ -563,6 +571,130 @@ - (RLMNotificationToken *)addNotificationBlock:(void (^)(RLMResults *, RLMCollec
...
+}
+
+- (void)unsubscribe {
+    RLMSyncSubscriptionSet *subscriptions = self.realm.subscriptions;
+
+    if (self.associatedSubscriptionId) {
+        [subscriptions update:^{
+            [subscriptions removeSubscriptionWithId:self.associatedSubscriptionId];
+        }];
+    } else {
```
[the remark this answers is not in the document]
ericjordanmossman: So if you have two subs with the same query. One named and one unnamed. And remove the named, the unnamed will stay in the set.
But yes in other cases you could have multiple results relying on the same subscription, if you remove one place it could affect the other results.

Realm/RLMResults.h
ericjordanmossman: If we make a mention of .always on that docstring, then I think we should do it for all of them. But putting all of them there would pollute if it's already explained on the enum.

The behavior is similar to asyncOpen and since asyncOpen doesn't have a default timeout (or any) I think keeping close to that behavior is nice. Or at least that's my preferance.

Realm/ObjectServerTests/SwiftFlexibleSyncServerTests.swift
ericjordanmossman: This is saying that the two subs present are 1) "age >= 8" 2) "age >= 8 && age < 8"
Not a logical AND. I can change.

Realm/RLMResults.mm
```diff
@@ -30,6 +31,8 @@
 #import "RLMRealmConfiguration_Private.hpp"
 #import "RLMSchema_Private.h"
 #import "RLMSectionedResults_Private.hpp"
+#import "RLMSyncSubscription.h"
+#import "RLMSyncSubscription_Private.hpp"
```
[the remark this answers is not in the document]
ericjordanmossman: Edited in last commit.

Realm/ObjectServerTests/SwiftFlexibleSyncServerTests.swift
```diff
@@ -1315,6 +1356,318 @@ extension SwiftFlexibleSyncServerTests {
         XCTAssertEqual(realm.subscriptions.count, 1)
     }
 
+    // MARK: Subscribe
+
+#if swift(>=5.8)
+//     wait(for:) doesn't work in async functions because it blocks the calling
+//     thread and doesn't let async tasks run. Xcode 14.3 introduced a new async
```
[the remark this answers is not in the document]
ericjordanmossman: Is that satisfied by just checking for swift 5.8? I was following existing >5.8 tests like https://github.com/realm/realm-swift/blob/master/RealmSwift/Tests/RealmCollectionTypeTests.swift#L658.

Realm/ObjectServerTests/SwiftFlexibleSyncServerTests.swift
```diff
@@ -1315,6 +1356,318 @@ extension SwiftFlexibleSyncServerTests {
...
+        let realm = try openFlexibleSyncRealm()
+
+        var results0 = try await realm.objects(SwiftPerson.self).where { $0.age >= 8 }.subscribe()
+        XCTAssertEqual(results0.count, 3)
+        XCTAssertEqual(realm.subscriptions.count, 1)
+        results0 = try await results0.where { $0.age < 8 }.subscribe() // results0 local query is { $0.age >= 8 AND $0.age < 8 }
+        XCTAssertEqual(results0.count, 0) // no matches because local query is impossible
+        XCTAssertEqual(realm.subscriptions.count, 2) // two subsscriptions: "$0.age >= 8 AND $0.age < 8" and "$0.age >= 8"
+        let results1 = realm.objects(SwiftPerson.self)
+        XCTAssertEqual(results1.count, 3) // three objects on device because subscription "$0.age >= 8" still exists
```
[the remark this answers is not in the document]
ericjordanmossman: Yes. The test is making sure that the server query match when the local query results append.

Realm/ObjectServerTests/SwiftFlexibleSyncServerTests.swift
ericjordanmossman: Yes. Sorry was missing the line 
```
realm.syncSession!.suspend()
        Task {
```
Test should check that subscribe() when an identical query with no name won't attempt to subscribe again.

Realm/ObjectServerTests/SwiftFlexibleSyncServerTests.swift
ericjordanmossman: It would. But unsubscribe() doesn't wait for any data to leave the phone. It's synchronous and returns when the subscriptionSet changes are committed locally. As far as I can tell, waitForDownloads and waitForUploads don't cover the case for data _leaving_ the client. And I couldn't figure a way to do that. The only way I could thnk was an arbitrary wait time. Is there a better way?

RealmSwift/Results.swift
```diff
@@ -147,6 +154,77 @@ extension Projection: KeypathSortable {}
...
+     that's already subscribed to with a name, an additional subscription is created without a name.
+     If `.subscribe()` is called with a name on a query that's
+     already subscribed to without a name, an additional subscription is created
+     with the provided name.
+     If `.subscribe()` is called with a name that's in use on
+     a different query, the old subscription is updated with the new query.
+     If `.subscribe()` is called with the same name and
+     same query of an existing subscription, no new subscription is created.
+     */
+    @available(macOS 10.15, iOS 13.0, tvOS 13.0, watchOS 6.0, *)
```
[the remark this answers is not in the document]
ericjordanmossman: Should be edited in latest commits.

Realm/RLMResults.mm
```diff
@@ -563,6 +571,130 @@ - (RLMNotificationToken *)addNotificationBlock:(void (^)(RLMResults *, RLMCollec
...
+    return [self subscribeWithName:nil onQueue:queue completion:completion];
+};
+
+- (void)subscribeWithName:(NSString *_Nullable)name
+                  onQueue:(dispatch_queue_t _Nullable)queue
+               completion:(RLMResultsCompletionBlock)completion {
+    return [self subscribeWithName:name waitForSyncMode:RLMWaitForSyncModeOnCreation onQueue:queue completion:completion];
+}
+
+- (void)completeOnQueue:(dispatch_queue_t _Nullable)queue
```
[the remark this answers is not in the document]
ericjordanmossman: Edited in last commit.

Realm/RLMResults.mm
```diff
@@ -563,6 +571,130 @@ - (RLMNotificationToken *)addNotificationBlock:(void (^)(RLMResults *, RLMCollec
...
+- (void)subscribeWithName:(NSString *_Nullable)name
+          waitForSyncMode:(RLMWaitForSyncMode)waitForSyncMode
+                  onQueue:(dispatch_queue_t _Nullable)queue
+                  timeout:(NSTimeInterval)timeout
+               completion:(RLMResultsCompletionBlock)completion {
+    // Create an internal completion that will only be called once
+    __block BOOL called = false;
+    void(^methodCompletion)(RLMResults* _Nullable, NSError* _Nullable) = ^(RLMResults* _Nullable results, NSError* _Nullable error) {
+        if (!called) {
+            if (error != nil) {
```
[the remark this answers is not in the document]
ericjordanmossman: Edited in last commit

Realm/RLMResults.mm
```diff
@@ -563,6 +571,130 @@ - (RLMNotificationToken *)addNotificationBlock:(void (^)(RLMResults *, RLMCollec
...
+                completion(results, nil);
+            }
+            called = true;
+        }
+    };
+
+    // Setup timer
+    dispatch_time_t time =  dispatch_time(DISPATCH_TIME_NOW, (int64_t)(timeout * NSEC_PER_SEC));
+    // If the asynchronouns subscribe call below doesn't return after `time` seconds, the internal completion is called with an error.
+    dispatch_after(time, dispatch_get_main_queue(), ^{
```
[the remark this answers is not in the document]
ericjordanmossman: How so? What kind of definition are you thinking?

Realm/RLMResults.h
ericjordanmossman: Sorry - how so? Like a hyperlink? Or a note below that says : 
```- note: The update transaction used is `[RLMSyncSubscription update:queue:onComplete:]`.
```

Realm/ObjectServerTests/SwiftFlexibleSyncServerTests.swift
```diff
@@ -1315,6 +1356,337 @@ extension SwiftFlexibleSyncServerTests {
...
+        let start = Date()
+        while realm.subscriptions.state != .complete && start.timeIntervalSinceNow > -10.0 {
+            sleep(1)
+        }
+        XCTAssertEqual(realm.subscriptions.state, .complete)
+    }
+
+    // fails in teardown "Assertion failed: no_sessions"
+    // can't tell which?
+    func skip_testSubscribeTimeoutSucceeds() async throws {
```
dianaafanador3, line 1673: Probably this is failing because you're not cancelling the dispatch block when completing succesfully, so this will probably going to fail when you set a timeout and successfully completes a subscription. Check https://developer.apple.com/documentation/dispatch/1431058-dispatch_block_cancel

Realm/RLMError.h
```diff
@@ -199,6 +199,14 @@ typedef RLM_ERROR_ENUM(NSInteger, RLMError, RLMErrorDomain) {
...
      */
     RLMErrorNoSubscriptionForWrite = 22,
+
+    /**
+     Denotes an error that occurs when the client ends an operation
+     due to a timeout set by the client. If the canceled operation was waiting for
+     the sync  server, this error *does not* imply a server error occurred or
+     that a server configured timeout had been exceeded.
+     */
+    RLMErrorClientTimeout = 23,
```
tgoyne, line 209: We currently report timeouts as `NSError(domain: NSPOSIXErrorDomain, code: ETIMEDOUT)`. It's a breaking change to change this, and we shouldn't have different types of timeouts reported differently.

Realm/ObjectServerTests/SwiftFlexibleSyncServerTests.swift
```diff
@@ -1315,6 +1356,337 @@ extension SwiftFlexibleSyncServerTests {
...
+
+    @MainActor
+    func testSubscribeSameQueryNoName() async throws {
+        try await populateSwiftPerson()
+        let realm = try openFlexibleSyncRealm()
+
+        let results0 = try await realm.objects(SwiftPerson.self).where { $0.age >= 8 }.subscribe()
+        let ex = XCTestExpectation(description: "no attempt to re-create subscription, returns immediately")
+        realm.syncSession!.suspend()
+        Task {
```
tgoyne, line 1401: This Task doesn't appear to be doing anything to me?

Realm/ObjectServerTests/SwiftFlexibleSyncServerTests.swift
```diff
@@ -1315,6 +1356,337 @@ extension SwiftFlexibleSyncServerTests {
...
+    func testSubscribeSameQueryDifferentName() async throws {
+        try await populateSwiftPerson()
+        let realm = try openFlexibleSyncRealm()
+
+        let results0 = try await realm.objects(SwiftPerson.self).where { $0.age >= 8 }.subscribe()
+        _ = try await realm.objects(SwiftPerson.self).where { $0.age >= 8 }.subscribe(name: "8 or older")
+        _ = try await results0.subscribe(name: "older than 7")
+        XCTAssertEqual(realm.subscriptions.count, 3)
+        let first = realm.subscriptions.first(ofType: SwiftPerson.self) { $0.age >= 8 }
+        XCTAssertNil(first?.name)
```
tgoyne, line 1439: This is asserting that either `first` or `first.name` is `nil`, and would pass if the subscription was missing entirely. `XCTAssertNil(try XCTUnwrap(first).name)` is a more precise test.

Realm/RLMResults.h
```diff
@@ -20,6 +20,25 @@
...
+typedef void(^RLMResultsCompletionBlock)(RLMResults * _Nullable, NSError * _Nullable);
+
+/**
+ Determines wait for download behavior when subscribing on RLMResults.
+ @see ``[RLMResults subscribeWithName:waitForSyncMode:onQueue:completion:]``
+*/
+typedef NS_ENUM(NSUInteger, RLMWaitForSyncMode) {
+    /// `subscribeWithName` will return once matching objects are downloaded
+    /// from the server only when the subscription is created the first time. If the
+    /// subscriptions already exists, the method will return without waiting for new downloads.
```
tgoyne, line 33: `subscribeWithName:` always returns immediately and what this controls is when the callback is invoked.

Realm/RLMResults.h
```diff
@@ -408,6 +427,167 @@ __attribute__((warn_unused_result));
...
+
+#pragma mark - Flexible Sync
+
+/**
+ Creates an RLMSyncSubscription matching the RLMResults' local filter.
+
+ This method opens a write transaction that creates or updates a subscription.
+ It is advised to *not* use this method to batch multiple subscription changes
+ to the server.
+ @see: `[RLMSyncSubscription update:queue:onComplete:]` for batch updates
```
tgoyne, line 439: This wording is confusing to me. This function doesn't do any batching, so it's weird to say that it can do batches and you just shouldn't.

Realm/RLMResults.h
```diff
@@ -408,6 +427,167 @@ __attribute__((warn_unused_result));
...
+
+ @param name The name used  to identify the subscription.
+ @param waitForSyncMode Dictates when the completion handler is called
+ @param queue The queue where the completion disptaches.
+ @param completion The block called after the subscription download is
+ completed. Return behavior is dictated by the RLMWaitForSyncMode.
+ @see ``RLMWaitForSyncMode``
+ */
+- (void)subscribeWithName:(NSString *_Nullable)name
+          waitForSyncMode:(RLMWaitForSyncMode)waitForSyncMode
```
tgoyne, line 525: "Mode" shouldn't be the argument labels.

Realm/RLMResults.mm
```diff
@@ -563,6 +570,137 @@ - (RLMNotificationToken *)addNotificationBlock:(void (^)(RLMResults *, RLMCollec
     return _results.add_notification_callback(RLMWrapCollectionChangeCallback(block, self, true), std::move(keyPaths));
 }
 
+- (void)completeOnQueue:(dispatch_queue_t _Nullable)queue
```
tgoyne, line 573: Taking a completion queue and passing the RLMResults to the completion callback is a bad combination. The RLMResults is only actually usable if it's queue-confined and the user explicitly passed in the same queue as they were already on. We should probably call the completion block on the Realm's scheduler rather than having a separate queue passed in. There probably also isn't really a reason to pass the RLMResults to the completion.

Realm/RLMResults.mm
```diff
@@ -563,6 +570,137 @@ - (RLMNotificationToken *)addNotificationBlock:(void (^)(RLMResults *, RLMCollec
...
+    return false;
+    }
+
+- (void)subscribeWithName:(NSString *_Nullable)name
+          waitForSyncMode:(RLMWaitForSyncMode)waitForSyncMode
+                  onQueue:(dispatch_queue_t _Nullable)queue
+               completion:(RLMResultsCompletionBlock)completion {
+    if ([self checkEarlyReturnSubscribeWithSyncMode:waitForSyncMode name:name onQueue:queue completion:completion]) {
+        return;
+    } else {
```
tgoyne, line 650: The whole point of an early return is to not have the else block, and doing both is just sorta redundant extra code. `} else {` is also not the style used in our obj-c code.

Realm/RLMSyncSubscription.h
```diff
@@ -117,13 +117,11 @@ RLM_HEADER_AUDIT_BEGIN(nullability, sendability)
...
 /**
- Synchronously performs any transactions (add/remove/update) to the subscription set within the block,
- this will not wait for the server to acknowledge and see all the data associated with this collection of subscriptions,
- and will return after committing the subscription transactions.
+ Synchronously performs any transactions (add/remove/update) to the subscription set within the block. The `onComplete` block is executed after waiting for associated data to be downloaded from the server.
 
  @param block The block containing actions to perform to the subscription set.
  @param onComplete A block which is called upon synchronization of
-                   subscriptions to the server. The block will be passed `nil`
+                   subscriptions from the server. The block will be passed `nil`
```
tgoyne, line 124: to is correct here?

Realm/RLMSyncSubscription.h
```diff
@@ -135,6 +133,23 @@ RLM_HEADER_AUDIT_BEGIN(nullability, sendability)
...
+ Synchronously performs any transactions (add/remove/update) to the subscription set within the block. The `onComplete` block is executed after waiting for associated data to be downloaded from the server.
+
+ @param block The block containing actions to perform to the subscription set.
+ @param queue The serial queue to deliver notifications to.
+ @param onComplete A block which is called upon synchronization of
+                   subscriptions from the server. The block will be passed `nil`
+                   if the update succeeded, and an error describing the problem
+                   otherwise.
+ */
+ - (void)update:(__attribute__((noescape)) void(^)(void))block
```
tgoyne, line 146: Exposing something with a completion queue parameter as a swift async function doesn't really make sense. It'll always just immediately hop off that queue and onto a swift async executor.

Realm/RLMSyncSubscription.mm
```diff
@@ -237,10 +241,45 @@ - (RLMSyncSubscriptionState)state {
...
+    [self update:block queue:nil onComplete:completionBlock];
+}
+
+// FIXME: Ultimately needs something cancellable from realm-core instead of sdk-level workaround.
+- (void)update:(__attribute__((noescape)) void(^)(void))block
+         queue:(nullable dispatch_queue_t)queue
+       timeout:(NSTimeInterval)timeout
+    onComplete:(void(^)(NSError *))completionBlock {
+    // Create an internal completion block that will only be called once
+    __block BOOL called = false;
```
tgoyne, line 257: This is a data race. The block can be called on both the main thread and the sync worker thread at the same time if it completes just as it would time out. This also is making it so that the completion block is always captured until the timeout expires.

Realm/RLMSyncSubscription_Private.hpp
```diff
@@ -18,23 +18,51 @@
...
+class Query;
+}
 
 RLM_HEADER_AUDIT_BEGIN(nullability, sendability)
 
 @interface RLMSyncSubscription ()
 
 - (instancetype)initWithSubscription:(realm::sync::Subscription)subscription subscriptionSet:(RLMSyncSubscriptionSet *)subscriptionSet;
 
+@property (nonatomic, readonly) std::string_view stdString;
```
tgoyne, line 37: This property name is just the type of the property and doesn't give any indication as to what it is.

RealmSwift/Results.swift
```diff
@@ -147,6 +147,80 @@ extension Projection: KeypathSortable {}
...
+    @_unsafeInheritExecutor
+    public func subscribe(name: String? = nil, waitForSync: WaitForSyncMode = .onCreation, timeout: TimeInterval? = nil) async throws -> Results<Element> {
+        var rlmResults = ObjectiveCSupport.convert(object: self)
+        guard let timeout = timeout else {
+            rlmResults = try await rlmResults.__subscribe(withName: name, waitForSyncMode: waitForSync, on: nil)
+            return Results(rlmResults)
+        }
+        rlmResults = try await rlmResults.__subscribe(withName: name, waitForSyncMode: waitForSync, on: nil, timeout: timeout)
+        return Results(rlmResults)
```
tgoyne, line 191: `return self`

Realm/RLMResults.mm
ericjordanmossman: So queue is still available from the user facing api? It's only the impl that's using the scheduler? So something like:
```
- (void)completeOnQueue:(dispatch_queue_t _Nullable)queue
             completion:(RLMResultsCompletionBlock)completion
                  error:(NSError *_Nullable)error {
    RLMResults *ret = (error != nil) ? nil : self;
    if (queue) {
        RLMScheduler* scheduler = [RLMScheduler dispatchQueue:queue];
        [scheduler invoke:^{
            completion(ret, error);
        }];
    } else {
        completion(ret, error);
    }
```

or this with the results removed

```
- (void)completeOnQueue:(dispatch_queue_t _Nullable)queue
             completion:(void(^)(NSError *))completion
                  error:(NSError *_Nullable)error {
    // the commented lines can also be removed right?, since [RLMScheduler dispatchQueue:] checks for queue?
    // if (queue) {
        RLMScheduler* scheduler = [RLMScheduler dispatchQueue:queue];
        [scheduler invoke:^{
            completion(error);
        }];
    // } else {
    //    completion(error);
    //}
}
```

Realm/RLMSyncSubscription.h
ericjordanmossman: I was thinking "the data _from_" when I changed it.

ericjordanmossman: Changing it to "...A block which is called upon synchronization of
                   data from the server..."

dianaafanador3:
This is still missing some CHANGELOG entry so users testing this can check the added APIs without going directly to the APIs docstrings

dianaafanador3:
@tgoyne I implemented a `RLMAsyncSubscriptionTask` which handles the timeout and the waitForSubscription asynchronous calls, and has a mutex which would prevent it from running from different threads. Let me know if this what is asked in the above comments.

Realm/RLMResults.mm
```diff
@@ -563,6 +571,137 @@ - (RLMNotificationToken *)addNotificationBlock:(void (^)(RLMResults *, RLMCollec
...
+        RLMSyncSubscriptionSet *subscriptions = self.realm.subscriptions;
+        [subscriptions update: ^{
+            // associated subscription id is nil when no name is provided.
+            self.associatedSubscriptionId = [subscriptions addSubscriptionWithClassName:self.objectClassName
+                                                                       subscriptionName:name
+                                                                                  query:_results.get_query()
+                                                                         updateExisting:true];
+        } queue: queue onComplete:^(NSError *error) {
```
leemaguire, line 660:
```suggestion
        } queue:queue onComplete:^(NSError *error) {
```

Realm/RLMSyncSubscription.mm
```diff
@@ -266,21 +279,21 @@ - (void)update:(__attribute__((noescape)) void(^)(void))block onComplete:(void(^
...
-    _subscriptionSet->get_state_change_notification(realm::sync::SubscriptionSet::State::Complete)
-        .get_async([completionBlock, queue](realm::StatusWith<realm::sync::SubscriptionSet::State> state) noexcept {
-            if (queue) {
-                return dispatch_async(queue, ^{
-                    completionBlock(makeError(state));
-                });
-            }
-            return completionBlock(makeError(state));
-        });
+    RLMAsyncSubscriptionTask *syncSubscriptionTask = [[RLMAsyncSubscriptionTask alloc]
```
leemaguire, line 291: Wrong indentation

Realm/RLMSyncSubscription.mm
```diff
@@ -387,20 +415,35 @@ - (void)addSubscriptionWithClassName:(NSString *)objectClassName
                            predicate:(NSPredicate *)predicate
                       updateExisting:(BOOL)updateExisting {
     [self verifyInWriteTransaction];
-    
+
     RLMClassInfo& info = _realm->_info[objectClassName];
     auto query = RLMPredicateToQuery(predicate, info.rlmObjectSchema, _realm.schema, _realm.group);
-    
+
+    [self addSubscriptionWithClassName:objectClassName
```
leemaguire, line 422: Wrong indentation

Realm/RLMSyncSubscription_Private.hpp
```diff
@@ -18,23 +18,49 @@
 
 #import "RLMSyncSubscription_Private.h"
 
+#import <string>
```
leemaguire, line 21: unused header

Realm/ObjectServerTests/SwiftFlexibleSyncServerTests.swift
```diff
@@ -1311,6 +1352,336 @@ extension SwiftFlexibleSyncServerTests {
...
+        let results0 = try await realm.objects(SwiftPerson.self).where { $0.age >= 6 }.subscribe(timeout: 2)
+        XCTAssertEqual(results0.count, 5)
+        XCTAssertEqual(realm.subscriptions.count, 1)
+        let results1 = try await realm.objects(SwiftPerson.self).where { $0.lastName == "lastname_3" }.subscribe(timeout: 2)
+        XCTAssertEqual(results1.count, 1)
+        XCTAssertEqual(results0.count, 5)
+        XCTAssertEqual(realm.subscriptions.count, 2)
+        let results2 = realm.objects(SwiftPerson.self)
+        XCTAssertEqual(results2.count, 6)
+//        waitForDownloads(for: realm)
```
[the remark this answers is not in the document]
dianaafanador3: we are not expecting for the timeout to be triggered in this test, we are expecting for this to succedd even though there is a timeout.

Realm/RLMAsyncTask.mm
```diff
@@ -474,3 +475,80 @@ - (void)wait:(void (^)())completion {
...
+        _subscriptionSet->_subscriptionSet->get_state_change_notification(realm::sync::SubscriptionSet::State::Complete)
+            .get_async([self](realm::StatusWith<realm::sync::SubscriptionSet::State> state) noexcept {
+                NSError *error = makeError(state);
+                [self invokeCompletionWithError:error];
+            });
+    }
+}
+
+-(void)invokeCompletionWithError:(NSError * _Nullable)error {
+    if (_called) {
```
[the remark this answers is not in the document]
dianaafanador3: I added a mutex check here

Realm/RLMResults.mm
```diff
@@ -563,6 +571,137 @@ - (RLMNotificationToken *)addNotificationBlock:(void (^)(RLMResults *, RLMCollec
...
 }
 
+- (void)completeOnQueue:(dispatch_queue_t _Nullable)queue
+             completion:(RLMResultsCompletionBlock)completion
+                  error:(NSError *_Nullable)error {
+    RLMResults *ret = (error != nil) ? nil : self;
+    if (queue) {
+        RLMScheduler* scheduler = [RLMScheduler dispatchQueue:queue];
+        [scheduler invoke:^{
+            completion(ret, error);
```
leemaguire, line 581: This won't work if the callback queue doesn't match the caller queue.

dianaafanador3:
@leemaguire after adding a test where the calling thread is different than the callback thread some issues with threading came out, the only way to resolve it without changing too much of the implementation is forcing the user to use the MainActor to do the results subscription. I propose releasing this preview API as it is, and check if we any user feedback regarding this part, we can then get deeper into support any actor for this call.

Realm/RLMResults.h
```diff
@@ -408,6 +427,169 @@ __attribute__((warn_unused_result));
...
+ This could create a performance bottleneck by opening multiple unnecessary write transactions.
+ @see: `[RLMSyncSubscription update:queue:onComplete:]` in order to create multiple subscriptions.
+
+ @param completion The block called after the subscription download
+ is completed. The completion is called according to
+ RLMWaitForSyncModeOnCreation behavior.
+ @see ``RLMWaitForSyncModeOnCreation`` and ``RLMWaitForSyncMode``
+ @param queue The queue where the completion disptaches.
+ */
+- (void)subscribeWithCompletion:(RLMResultsCompletionBlock)completion
```
leemaguire, line 456: This should follow the convention of the other API's where the queue comes before the block. So it would look something like:
```
- (void)subscribeWithCompletionOnQueue:(dispatch_queue_t _Nullable)queue 
                                             completionBlock:(RLMResultsCompletionBlock)completion;

File: Realm/RLMResults.mm [hunk 40]
```diff
@@ -563,6 +571,135 @@ - (RLMNotificationToken *)addNotificationBlock:(void (^)(RLMResults *, RLMCollec
     return _results.add_notification_callback(RLMWrapCollectionChangeCallback(block, self, true), std::move(keyPaths));
 }
 
+- (void)completeOnThreadSafeReference:(RLMThreadSafeReference * _Nullable)reference
```
leemaguire, line 574: Should be `completionWithThreadSafeReference`

Realm/RLMResults.h
```diff
@@ -408,6 +427,169 @@ __attribute__((warn_unused_result));
...
+ @see: `[RLMSyncSubscription update:queue:onComplete:]` in order to create multiple subscriptions.
+
+ @param completion The block called after the subscription download
+ is completed. The completion is called according to
+ RLMWaitForSyncModeOnCreation behavior.
+ @see ``RLMWaitForSyncModeOnCreation`` and ``RLMWaitForSyncMode``
+ @param queue The queue where the completion disptaches.
+ */
+- (void)subscribeWithCompletionOnQueue:(dispatch_queue_t _Nullable)queue
+                       completionBlock:(RLMResultsCompletionBlock)completion NS_REFINED_FOR_SWIFT;
```
tgoyne, line 457: Why is this NS_REFINED_FOR_SWIFT? There isn't a Swift version of this function and I don't see any reason for forbid calling it from Swift when using the obj-c API.

Realm/RLMResults.mm
```diff
@@ -563,6 +571,134 @@ - (RLMNotificationToken *)addNotificationBlock:(void (^)(RLMResults *, RLMCollec
...
+- (void)unsubscribe {
+    RLMSyncSubscriptionSet *subscriptions = self.realm.subscriptions;
+
+    if (self.associatedSubscriptionId) {
+        [subscriptions update:^{
+            [subscriptions removeSubscriptionWithId:self.associatedSubscriptionId];
+        }];
+    } else {
+        RLMSyncSubscription *sub = [subscriptions subscriptionWithQuery:_results.get_query()];
+        if (sub != nil  && sub.name == nil) {
```
tgoyne, line 693: This check is redundant and can be just `if (sub.name) {`

Realm/RLMResults.mm
```diff
@@ -563,6 +571,134 @@ - (RLMNotificationToken *)addNotificationBlock:(void (^)(RLMResults *, RLMCollec
...
+    auto tsr = (error != nil) ? nil : reference;
+    RLMRealmConfiguration *configuration = _realm.configuration;
+    if (queue) {
+        RLMScheduler* scheduler = [RLMScheduler dispatchQueue:queue];
+        if (reference) {
+            [scheduler invoke:^{
+                RLMRealm *realm = [RLMRealm realmWithConfiguration:configuration queue:queue error:nil];
+                RLMResults *collection = [realm resolveThreadSafeReference:tsr];
+                collection.associatedSubscriptionId = self.associatedSubscriptionId;
+                [realm refresh];
```
tgoyne, line 587: This should not be calling `refresh`.

Realm/RLMResults.mm
```diff
@@ -563,6 +571,134 @@ - (RLMNotificationToken *)addNotificationBlock:(void (^)(RLMResults *, RLMCollec
...
 }
 
+- (void)completionWithThreadSafeReference:(RLMThreadSafeReference * _Nullable)reference
+                                    queue:(dispatch_queue_t _Nullable)queue
+                               completion:(RLMResultsCompletionBlock)completion
+                                    error:(NSError *_Nullable)error {
+    auto tsr = (error != nil) ? nil : reference;
+    RLMRealmConfiguration *configuration = _realm.configuration;
+    if (queue) {
+        RLMScheduler* scheduler = [RLMScheduler dispatchQueue:queue];
```
tgoyne, line 581: This can just use the queue directly rather than wrapping it in a scheduler.

Realm/RLMSyncSubscription_Private.hpp
```diff
@@ -18,23 +18,49 @@
...
 @interface RLMSyncSubscription ()
-
 - (instancetype)initWithSubscription:(realm::sync::Subscription)subscription subscriptionSet:(RLMSyncSubscriptionSet *)subscriptionSet;
-
 @end
 
-@interface RLMSyncSubscriptionSet ()
+@interface RLMSyncSubscriptionSet () {
+@public 
+    std::unique_ptr<realm::sync::SubscriptionSet> _subscriptionSet;
```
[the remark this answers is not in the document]
dianaafanador3: This is been used by `RLMAsyncSubscriptionTask` to call `get_state_change_notificatio`

Realm/RLMResults.h
```diff
@@ -408,6 +427,169 @@ __attribute__((warn_unused_result));
...
+ The method returns after committing the subscription removal to the
+ realm's local subscription set. Calling this method will not wait for objects to
+ be removed from the realm.
+
+ Calling unsubscribe on a RLMResults does not remove the local filter from the RLMResults.
+ After calling unsubscribe, RLMResults may still contain objects because
+ other subscriptions may exist in the RLMRealm's subscription set.
+
+ In order for a named subscription to be removed, the RLMResults
+ must have previously created the subscription.
```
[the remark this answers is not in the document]
dianaafanador3: actually you can, the `RLMResults` passed on the completion has the associatedId copied.

Realm/RLMResults.mm
dianaafanador3: This is the only way the realm is updated, if we complete the subscription on a different calling thread

Realm/RLMAsyncTask.mm
```diff
@@ -474,3 +475,82 @@ - (void)wait:(void (^)())completion {
...
+        dispatch_block_cancel(_worker);
+        _worker = nil;
+    }
+
+    if (_queue) {
+        return dispatch_async(_queue, ^{
+            _completion(error);
+        });
+    }
+    _completion(error);
```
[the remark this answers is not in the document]
dianaafanador3: I cannot remove the `called` flag, this is the only way to avoid a second call when the `getAsync` is called after the timeout callback, this can happen if the connection is restored or the subscription is taking too much time.

Realm/RLMResults.mm
tgoyne: I do not understand this explanation.

Realm/RLMAsyncTask.mm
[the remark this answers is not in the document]
tgoyne: You can just check if `completion` is non-nil, as it should always be nil once it's been called (or when another thread is about to call it), and if there's ever a case where it isn't then that's a bug.

Realm/RLMResults.h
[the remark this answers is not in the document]
tgoyne: The documentation needs to say that, then.

Realm/RLMResults.h
```diff
@@ -408,6 +427,171 @@ __attribute__((warn_unused_result));
...
+ @param waitForSyncMode Dictates when the completion handler is called
+ @param queue The queue where the completion dispatches.
+ @param completion The block called after the subscription download is
+ completed. Return behavior is dictated by the RLMWaitForSyncMode.
+ @see ``RLMWaitForSyncMode``
+ */
+- (void)subscribeWithName:(NSString *_Nullable)name
+              waitForSync:(RLMWaitForSyncMode)waitForSyncMode
+                  onQueue:(dispatch_queue_t _Nullable)queue
+               completion:(RLMResultsCompletionBlock)completion NS_REFINED_FOR_SWIFT
```
tgoyne, line 529: None of these should be NS_REFINED_FOR_SWIFT.

Realm/RLMResults.mm
dianaafanador3: I though we need to get the callback confined realm to the state of the realm which is running the subscription, (which is the calling thread realm)

Realm/RLMResults.h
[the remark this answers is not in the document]
dianaafanador3: Updated the documentation and added a test for this

Realm/RLMAsyncTask.mm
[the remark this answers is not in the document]
dianaafanador3: updated this to not use `_called`

Realm/RLMResults.h
```diff
@@ -408,6 +427,173 @@ __attribute__((warn_unused_result));
...
+#pragma mark - Flexible Sync
+
+/**
+ Creates an RLMSyncSubscription matching the RLMResults' local filter.
+
+ After committing the subscription to the realm's local subscription set, the method
+ will wait for downloads according to ``RLMWaitForSyncModeOnCreation`` behavior.
+ @see ``RLMWaitForSyncModeOnCreation`` and ``RLMWaitForSyncMode``
+
+ If `subscribeWithCompletion` is called on a query
```
tgoyne, line 440: Obj-c method names are generally written including the trailing colon as that's a functional part of the name.

Realm/RLMResults.h
```diff
@@ -408,6 +427,173 @@ __attribute__((warn_unused_result));
...
+ that's already subscribed to without a name, another subscription is not created.
+ If `subscribeWithName` is called without a name on a query
+ that's already subscribed to with a name, an additional subscription is created without a name.
+ If `subscribeWithName` is called with a name on a query that's
+ already subscribed to without a name, an additional subscription is created
+ with the provided name.
+ If `subscribeWithName` is called with a name that's in use on
+ a different query, the old subscription is updated with the new query.
+ If `subscribeWithName` is called with the same name and
+ same query of an existing subscription, no new subscription is created
```
tgoyne, line 549: This turns into an unreasonably large paragraph in the rendered docs since the newlines are non-functional.

Realm/RLMResults.h
```diff
@@ -408,6 +427,173 @@ __attribute__((warn_unused_result));
...
+
+ @param name The name used  to identify the subscription.
+ @param waitForSyncMode Dictates when the completion handler is called
+ @param queue The queue where the completion dispatches.
+ @param timeout A timeout which ends waiting for downloads
+ via the completion handler. If the timeout is exceeded the completion
+ handler returns an `RLMErrorClientTimeout`.
+ @see ``RLMErrorClientTimeout``
+ @param completion The block called after the subscription download is
+ completed. Return behavior is dictated by the RLMWaitForSyncMode.
```
tgoyne, line 564: I'm not sure what "Return behavior is dictated by the RLMWaitForSyncMode" means. I think it's trying to say that when the callback is invoked depends on the mode?

Realm/RLMResults.h
```diff
@@ -408,6 +427,173 @@ __attribute__((warn_unused_result));
...
+ This could create a performance bottleneck by opening multiple unnecessary write transactions.
+ @see: `[RLMSyncSubscription update:queue:onComplete:]` in order to create multiple subscriptions.
+
+ @param name The name used  to identify the subscription.
+ @param waitForSyncMode Dictates when the completion handler is called
+ @param queue The queue where the completion dispatches.
+ @param timeout A timeout which ends waiting for downloads
+ via the completion handler. If the timeout is exceeded the completion
+ handler returns an `RLMErrorClientTimeout`.
+ @see ``RLMErrorClientTimeout``
```
tgoyne, line 562: RLMErrorClientTimeout isn't a thing.

Realm/RLMResults.h
```diff
@@ -408,6 +427,173 @@ __attribute__((warn_unused_result));
...
+ realm's local subscription set. Calling this method will not wait for objects to
+ be removed from the realm.
+
+ Calling unsubscribe on a RLMResults does not remove the local filter from the RLMResults.
+ After calling unsubscribe, RLMResults may still contain objects because
+ other subscriptions may exist in the RLMRealm's subscription set.
+
+ In order for a named subscription to be removed, the RLMResults
+ must have previously created the subscription.
+ Returned `RLMResults`  from calling `subscribe` can be used later to unsubscribe from the same
```
tgoyne, line 587: Calling subscribe doesn't return a RLMResults.

Realm/RLMResults.mm
```diff
@@ -563,6 +571,140 @@ - (RLMNotificationToken *)addNotificationBlock:(void (^)(RLMResults *, RLMCollec
...
+                               confinedTo:(RLMScheduler *)confinement
+                               completion:(RLMResultsCompletionBlock)completion
+                                    error:(NSError *_Nullable)error {
+    auto tsr = (error != nil) ? nil : reference;
+    RLMRealmConfiguration *configuration = _realm.configuration;
+    if (reference) {
+        RLMRealm *realm = [RLMRealm realmWithConfiguration:configuration confinedTo:confinement error:nil];
+        RLMResults *collection = [realm resolveThreadSafeReference:tsr];
+        collection.associatedSubscriptionId = self.associatedSubscriptionId;
+        [realm refresh];
```
tgoyne, line 584: We should never be calling `refresh` because it is a blocking operation that waits for async notifiers to run.

Realm/RLMResults.mm
```diff
@@ -563,6 +571,140 @@ - (RLMNotificationToken *)addNotificationBlock:(void (^)(RLMResults *, RLMCollec
...
+        [confinement invoke:^{
```
tgoyne, line 585: Opening a Realm requires that we already be on the scheduler that we're opening the Realm for, so this can never be neccesary.

RealmSwift/Results.swift
```diff
@@ -147,6 +147,84 @@ extension Projection: KeypathSortable {}
...
+     - Note: This method opens a write transaction that creates or updates a subscription.
+     It's advised to *not* loop over this method in order to create multiple subscriptions.
+     This could create a performance bottleneck by opening multiple unnecessary write transactions.
+     - note To create multiple subscriptions at once use `SyncSubscription.update`.
+
+     - parameter name: The name applied to the subscription
+     - parameter waitForSync: Determines the download behavior for the subscription. Defaults to `.onCreation`.
+     See `RLMWaitForSyncMode`.
+     - parameter timeout: An optional client timeout. The client will cancel waiting for subscription downloads after this time has elapsed. Reaching this timeout doesn't imply a server error.
+     - returns: A `Results` object.
```
tgoyne, line 180: Yes, the type of the function indicates that. The relevant thing to document here is that it returns `self` and not some other Results.

RealmSwift/Results.swift
```diff
@@ -147,6 +147,84 @@ extension Projection: KeypathSortable {}
         return RLMIterator(collection: collection)
     }
 
+    // MARK: Flexible Sync
+
+#if swift(>=5.8)
+    /**
+     Creates a SyncSubscription matching the Results' local filter.
+
+     After committing the subscription to the realm's local subscription set, the method
```
tgoyne, line 156: The docs for this function are a mess when rendered by docc. Note that Xcode 15 now has an assistant editor for previewing rendered docs so you can see them side-by-side with the source rather than having to build the docs every time.

Realm/ObjectServerTests/RLMFlexibleSyncServerTests.mm
```diff
@@ -1090,6 +1123,136 @@ - (void)testFlexibleSyncInitialOnConnectionTimeout {
...
+
+    XCTestExpectation *ex = [self expectationWithDescription:@"wait for download"];
+    [[[Person allObjectsInRealm:realm] objectsWhere:@"lastName == 'Doe'"] subscribeWithName:@"unknown" onQueue:dispatch_get_main_queue() completion:^(RLMResults *results, NSError *error) {
+        XCTAssertNil(error);
+        XCTAssertEqual(results.count, 1U);
+        [results unsubscribe];
+        [ex fulfill];
+    }];
+    [self waitForExpectationsWithTimeout:5.0 handler:nil];
+    XCTAssertEqual(realm.subscriptions.count, 0);
```
tgoyne, line 1177: `0U` to avoid this warning.

Realm/RLMResults.mm
```diff
@@ -563,6 +571,138 @@ - (RLMNotificationToken *)addNotificationBlock:(void (^)(RLMResults *, RLMCollec
...
 
+- (void)completionWithThreadSafeReference:(RLMThreadSafeReference * _Nullable)reference
+                               confinedTo:(RLMScheduler *)confinement
+                               completion:(RLMResultsCompletionBlock)completion
+                                    error:(NSError *_Nullable)error {
+    auto tsr = (error != nil) ? nil : reference;
+    RLMRealmConfiguration *configuration = _realm.configuration;
+    [confinement invoke:^{
+        if (reference) {
+            RLMRealm *realm = [RLMRealm realmWithConfiguration:configuration error:nil];
```
tgoyne, line 582: This *should* never fail, but since we have the ability to properly report errors from here we might as well do so.

RealmSwift/Results.swift
```diff
@@ -147,6 +147,84 @@ extension Projection: KeypathSortable {}
...
+     - parameter waitForSync: ``WaitForSyncMode`` Determines the download behavior for the subscription. Defaults to `.onCreation`.
+     - parameter timeout: An optional client timeout. The client will cancel waiting for subscription downloads after this time has elapsed. Reaching this timeout doesn't imply a server error.
+     - returns: Returns `self`.
+
+     - warning: This function is only supported for main thread and
+                actor-isolated Realms.
+     */
+    @available(macOS 10.15, iOS 13.0, tvOS 13.0, watchOS 6.0, *)
+    @_unsafeInheritExecutor
+public func subscribe(name: String? = nil, waitForSync: WaitForSyncMode = .onCreation, timeout: TimeInterval? = nil) async throws -> Results<Element> {
```
tgoyne, line 187: Missing indentation

tgoyne (approved):
A few minor things but otherwise looks good.

PR #8253, opened 2023-06-05
Fix compilation with Xcode 15

tgoyne:


RealmSwift/Map.swift
```diff
@@ -814,7 +814,7 @@ extension Map: Sequence {
 
 extension Map {
     /// An adaptor for Map which makes it a sequence of `(key: Key, value: Value)` instead of a sequence of `SingleMapEntry`.
-    public struct KeyValueSequence<Key: _MapKey, Value: RealmCollectionValue>: Sequence {
```
tgoyne, line 817: Duplicating the generic arguments here was incorrect and now produces a warning.

RealmSwift/Util.swift
```diff
@@ -160,14 +160,16 @@ internal func logRuntimeIssue(_ message: StaticString) {
     }
 }
 
-#if swift(<5.9)
```
tgoyne, line 163: The API ended up getting changed after I wrote this, but it turns out we can't unconditionally use the built-in versions anyway due to the deployment target.

adbdehghan (approved):
Please approve this pr

davidkrammer:
Please approve this

dianaafanador3 (approved):

DaniilDavidovich:
this is a working solution

PR #8275, opened 2023-06-15
Remove warning caused by deprecated APIs used internally.

dianaafanador3:


tgoyne:
This seems like an unnecessary breaking change? The only warning related to loglevel I see when building is from the `typedef RLM_CLOSED_ENUM(NSUInteger, RLMSyncLogLevel);` declaration, which can be silenced in the same way as the usage of that typedef below.

JimRoepcke:
@dianaafanador3 when will this be merged and a new release made? These build warnings are somehow still an issue when we're building the Amity SDK which uses the Realm SDK. Our project has warnings-as-errors enabled and it's preventing us from updating to the latest version of the Amity SDK because we somehow can't build even though Amity is using Realm as a binary dependency.

Once this is fixed we can ask Amity to update their SDK.

Thanks!

tgoyne (approved):

PR #8289, opened 2023-06-30
Rework the installation examples

tgoyne:
The installation examples were originally created when there was just iOS and macOS and cross-platform Apps weren't a thing, and the approach that worked then has turned into a giant mess. Consolidating (almost) all the installation tests into a single codebase with separate project files per installation method cuts the amount of code involved by 90% and makes adding more platforms relatively painles cuts the amount of code involved by 90% and makes adding more platforms relatively painless.

Separate obj-c and Swift tests have not proven to have any value. We've never had any bugs caught specifically by the obj-c test which wasn't also caught by the swift test other than ones specifically related to the obj-c release zip, so now everything except for the obj-c only iOS static framework is tested only via Swift.

Some of the tests ended up just plain being redundant; we had separate XCFramework and fat framework tests, but then the fat framework tests had to be ported to use XCFrameworks.

Running the smoke tests in the simulator has never actually caught a bug and doubles the amount of things to build (and adds the requirement to reset the simulators, which takes a while), so ditch that and only test archiving the app for device.

The CocoaPods transitive dependency test is now done as part of the normal cocoapods tests rather than a separate one.

All platforms are now tested via Xcode SPM rather than only iOS. tvOS is now tested for all installation methods (previously it was not tested at all due to how much of a hassle adding platforms was).

visionOS is now tested when using Xcode 15 (i.e. it still isn't actually tested on CI). Currently only the SPM installation method supports visionOS. Cocoapods is actively working on adding support, and XCFramework support is only blocked on us cutting a release with it.


.jenkins.yml
```diff
@@ -17,7 +17,7 @@ target:
  - swiftpm-debug
  - swiftpm-address
  - swiftpm-thread
- - swiftpm-ios
+ - ios-xcode-spm
```
tgoyne, line 20: This renaming is a lazy way to get it to stop installing baas for these tests.

build.sh
```diff
@@ -464,18 +464,14 @@ case "$COMMAND" in
...
-        export REALM_XCODE_VERSION=$REALM_XCODE_LATEST_VERSION
         unset REALM_SWIFT_VERSION
+
+        # Build with the oldest supported Xcode version
+        REALM_XCODE_VERSION=$REALM_XCODE_OLDEST_VERSION sh build.sh xcframework osx
+
+        # Try to import the built framework using the newest supported version
         cd examples/installation
-        sh build.sh "test-osx-swift-xcframework"
+        REALM_XCODE_VERSION=$REALM_XCODE_LATEST_VERSION ./build.rb osx xcframework
```
tgoyne, line 474: The XCFramework test now uses the framework sitting in the `build` directory if it's not asked to fetch a release, so we can reuse it for this test rather than needing a separate target.

build.sh
```diff
@@ -672,50 +669,24 @@ case "$COMMAND" in
...
+        ./build.rb osx cocoapods
+        ./build.rb tvos cocoapods
+        ./build.rb watchos cocoapods
+        ./build.rb catalyst cocoapods
         ;;
 
     verify-cocoapods-ios-dynamic)
-        PLATFORM=$(echo "$COMMAND" | cut -d - -f 3)
-        # https://github.com/CocoaPods/CocoaPods/issues/7708
-        export EXPANDED_CODE_SIGN_IDENTITY=''
```
tgoyne, line 700: The bug this was working around was fixed a while ago.

build.sh
```diff
@@ -735,21 +706,6 @@ case "$COMMAND" in
         exit 0
         ;;
 
-    "verify-osx-swift")
```
tgoyne, line 738: All these verify- targets which just forward to the test- target are now covered by a `verify-*` target.

tgoyne:
Passing release pipeline with these changes: https://ci.realm.io/blue/organizations/jenkins/cocoa-pipeline/detail/cocoa-pipeline/2090/pipeline

.jenkins.yml
```diff
@@ -34,8 +34,9 @@ target:
  - cocoapods-osx
  - cocoapods-ios
  - cocoapods-ios-dynamic
- - cocoapods-ios-subdependency
  - cocoapods-watchos
+ - cocoapods-tvos
```
dianaafanador3, line 38: Would be nice to distinguish which targets are building the framework and which are installation test

dianaafanador3 (approved):
LGTM! Great work, the new build installation script is easier to understand now, and reworking the installation example removes a lot of files which is also nice.

examples/installation/SwiftPackageManager.notxcodeproj/project.pbxproj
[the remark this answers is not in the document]
tgoyne: build.rb symlinks SwiftPackageManager.notxcodeproj to SwiftPackageManager.xcodeproj, which is gitignored so that we don't accidentally commit it.

Jenkinsfile.releasability
```diff
@@ -5,30 +5,23 @@ platformNames = ['osx': 'macOS', 'ios': 'iOS', 'watchos': 'watchOS', 'tvos': 'tv
 carthageXcodeVersion = '14.3.1'
 docsSwiftVersion = '5.8.1'
 
-def installationTest(platform, test, language) {
```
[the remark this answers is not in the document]
tgoyne: This is part of the cocoa-pipeline job, so it happens after building the release but before publishing it, so I just ran it on this branch. There's also post-release testing which validates that the published release works, and that'll probably need some fiddling to fix after merging this since it's hard to validate now.

examples/installation/build.rb
```diff
@@ -0,0 +1,241 @@
...
+    for method in ['cocoapods', 'carthage', 'spm', 'xcframework']
+      next if platform == 'catalyst' && method == 'carthage'
+      next if platform == 'visionos' && method != 'spm'
+      test platform, method, 'dynamic'
+    end
+
+    test platform, 'cocoapods', 'static' unless platform == 'visionos'
+  end
+
+  test 'ios', 'xcframework', 'static'
```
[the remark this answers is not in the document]
tgoyne: For historical reasons we ship an iOS-specific static framework (iOS 7 didn't support dynamic frameworks, and we just kept shipping it even after dropping iOS 7). These changes generalize the build so that the static framework isn't iOS-specific but we're still only shipping it for iOS.

PR #8298, opened 2023-07-11
Fix Observation notification not triggering on nested properties

dianaafanador3:
This fix will create false positive `PropertyChange`s for projection properties which have the same origin property root.

      - warning: For projected properties where the original property has the same root property name,
                 this will trigger a `PropertyChange` for each of the Projected properties even though
                 the change only corresponds to one of them.
                 For the following `Projection` object
                 ```swift
                 class PersonProjection: Projection<Person> {
                     @Projected(\Person.firstName) var name
                     @Projected(\Person.address.country) originCountry
                     @Projected(\Person.address.phone.number) mobile
                 }
                 let token = projectedPerson.observe { changes in
                     if case .change(_, let propertyChanges) = changes {
                         propertyChanges[0].newValue as? String, "Winterfell" // Will notify the new value
                         propertyChanges[1].newValue as? String, "555-555-555" // Will notify with the current value, which hasn't change.
                     }
                 })
                 try realm.write {
                     person.address.country = "Winterfell"
                 }
                 ```


Test.xcworkspace/contents.xcworkspacedata
```diff
@@ -0,0 +1,10 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
tgoyne, line 1: Looks like an accidental addition?

RealmSwift/Projection.swift
```diff
@@ -321,15 +344,21 @@ extension ProjectionObservable {
 
             var projectedChanges = [PropertyChange]()
             for i in 0..<newValues.count {
-                for property in schema.filter({ $0.originPropertyKeyPathString == names[i] }) {
+                for property in schema.filter({ prop in
```
tgoyne, line 347: Now that the filter block is multiple lines of code it should be moved to a separate statement, as putting it inline in the `for in` makes it hard to see where the filter ends and the body of the for begins.

RealmSwift/Projection.swift
```diff
@@ -321,15 +344,21 @@ extension ProjectionObservable {
...
+                    guard let keyPaths = keyPaths, !keyPaths.isEmpty else {
+                        return prop.originPropertyKeyPathString.components(separatedBy: ".").first == names[i]
+                    }
+                    // This will allow us to notify `PropertyChange`s associated only to the keyPaths passed by the user, instead of any Property which has the same root as the notified one.
+                    return prop.originPropertyKeyPathString.components(separatedBy: ".").first == names[i] && keyPaths.contains(prop.originPropertyKeyPathString) }) {
```
tgoyne, line 352: This could avoid duplicating the condition by rearranging the checks slightly:

```swift
if prop.originPropertyKeyPathString.components(separatedBy: ".").first != names[i] {
    return false
}
guard let keyPaths, !keyPaths.isEmpty else {
    return true
}
// This will allow us to notify `PropertyChange`s
// associated only to the keyPaths passed by the user,
// instead of any Property which has the same root as the
// notified one.
return keyPaths.contains(prop.originPropertyKeyPathString)
```

Since we now require swift 5.7 we can just do `if let x {` rather than `if let x = x {`.

dianaafanador3: thanks of the tip, din't know this `if let x {`

tgoyne (approved):

leemaguire (approved):

PR #8307, opened 2023-07-18
Update Package.swift

kadircopur:


tgoyne:
We no longer support Swift 5.6.

Issue #8308, opened 2023-07-18
Linking failure with 10.41.x + SPM


jberkel:
### How frequently does the bug occur?

Always

### Description

It looks like 10.41.x introduced a (questionable) UIKit dependency in eec5f45c2abe6910c47e6ef889d1a6f7ff45d3e1
The podspecs were updated to include `-framework UIKit`, but `Package.swift` remains unchanged.

I now get a linking error for a project that includes Realm via SPM, see log output. Downgrading to `10.40.2` solves the problem.

### Stacktrace & log output

```shell
Testing failed:
	Undefined symbol: _OBJC_CLASS_$_UIDevice
	Linker command failed with exit code 1 (use -v to see invocation)
	Testing cancelled because the build failed.

** TEST FAILED **


The following build commands failed:

	Ld /Users/jberkel/git/xxx/build/Debug-iphonesimulator/PackageFrameworks/Realm.framework/Realm normal (in target 'Realm' from project 'Realm')
(1 failure)
```

### Can you reproduce the bug?

Always

### Version

10.41.x

### What Atlas Services are you using?

Local Database only

### Are you using encryption?

No

### Platform OS and version(s)

N/A

### Build environment

Xcode 14.3
SPM 5.8.0

dianaafanador3:
Hi @jberkel we tested this using SPM, but maybe we missed something. We'll take a look at this and let you know any updates

brewer001:
I have the same problem also. when i use 10.41.1,  the build always error. this is a log.
Checking out 13.17.0 of package ‘realm-core’
[21:44:49 ](https://hxbuild.workstyle.net/buildConfiguration/Hubble_2_HubbleForIOS/8123?hideProblemsFromDependencies=false&hideTestsFromDependencies=false&expandBuildProblemsSection=true&expandBuildChangesSection=true&showLog=8123_174_174&logView=flowAware)      Couldn’t check out revision ‘f1e962cd447f8b69f8f7cf46a188b1c6246923c5’:
[...]

PR #8312, opened 2023-07-20
Fix undefined symbol error for `UIKit` when linking Realm to a framework using SPM.

dianaafanador3:
Fix undefined symbol error for `UIKit` when linking Realm to a framework using SPM. ([#8308](https://github.com/realm/realm-swift/issues/8308), since v10.41.0)

This branches changes will be merged to `tg/spm-dynamic` where there is a test for this.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -249,7 +249,10 @@ let package = Package(
                 "Realm/RLMUserAPIKey.mm"
             ],
             publicHeadersPath: "include",
-            cxxSettings: cxxSettings
+            cxxSettings: cxxSettings,
+            linkerSettings: [
+                .linkedFramework("UIKit", .when(platforms: [.iOS, .macCatalyst, .tvOS, .watchOS]))
+            ]
         ),
         .target(
             name: "RealmSwift",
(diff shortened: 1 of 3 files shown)


build.sh
```diff
@@ -668,7 +668,9 @@ case "$COMMAND" in
         ;;
 
     verify-cocoapods-ios-dynamic)
-        REALM_TEST_BRANCH="$sha" ./build.rb ios cocoapods
+        REALM_TEST_BRANCH="$sha"
```
tgoyne, line 671: This is a no-op that sets the env variable for the command following it on the line (which is nothing).

Package.swift
```diff
@@ -249,7 +249,10 @@ let package = Package(
                 "Realm/RLMUserAPIKey.mm"
             ],
             publicHeadersPath: "include",
-            cxxSettings: cxxSettings
+            cxxSettings: cxxSettings,
+            linkerSettings: [
+                .linkedFramework("UIKit", .when(platforms: [.iOS, .macCatalyst, .tvOS, .watchOS]))
```
dianaafanador3, line 254: @tgoyne I didn't add visionOS to this, given that we don't add support for visionOS until this PR is merged https://github.com/realm/realm-swift/pull/8309/. Is there a way to add visionOS conditional to a Xcode version

build.sh
dianaafanador3: Changed this, added navigation to the `example/installation` directory given that the CI was failing because it was not able to find the `build.rb`. script

leemaguire (approved):

PR #8319, opened 2023-07-27
RCOCOA-2051 Add Static Queries

jsflax:
Adds support to use Codable types to query an atlas cluster, obviating the need to use the old BSON types we introduced.


dianaafanador3:
@jsflax Can you please rebase this, so the CI can be executed

tgoyne:
This is not in a reviewable state. It is a giant mess of unfinished things and unrelated changes.

jsflax:
Some field renames and added methods snuck in.

If I remove the `MongoQuery` addition (it was experimental and was chatted about when the project started), what else is missing @tgoyne ?

tgoyne:
There is a whole bunch of commented code, random unrelated changes, and very few of the new public things are documented.

PR #8390, opened 2023-10-06
Geospatial queries for points

dianaafanador3:
* Added initial support for geospatial queries on points.
   There is no new dedicated type to store Geospatial points, instead points should
   be stored as ([GeoJson-shaped](https://www.mongodb.com/docs/manual/reference/geojson/)) embedded object, as the example below.
   ```swift
   public class Location: EmbeddedObject {
     @Persisted private var coordinates: List<Double>
     @Persisted private var type: String = "Point"

     public var latitude: Double { return coordinates[1] }
     public var longitude: Double { return coordinates[0] }

     convenience init(_ latitude: Double, _ longitude: Double) {
         self.init()
         // Longitude comes first in the coordinates array of a GeoJson document
         coordinates.append(objectsIn: [longitude, latitude])
     }
   }
   ``` 
   Geospatial queries (`geoWithin`) can only be executed in such a type of objects and 
   will throw otherwise.
   The queries can be used to filter objects whose points lie within a certain area, 
   using the following pre-established shapes (`GeoBox`, `GeoPolygon`, `GeoCircle`).
   ```swift
   class Person: Object {
     @Persisted var name: String
     @Persisted var location: Location? // GeoJson embedded object
   }

   let realm = realmWithTestPath()
   try realm.write {
     realm.add(PersonLocation(name: "Maria", location: Location(latitude: 55.6761, longitude: 12.5683)))
   }

   let shape = GeoBox(bottomLeft: (55.6281, 12.0826), topRight: (55.6762, 12.5684))!
   let locations = realm.objects(PersonLocation.self).where { $0.location.geoWithin(shape) })
   ```
   A `filter` or `NSPredicate` can be used as well to perform a Geospatial query.
   ```swift
   let shape = GeoPolygon(outerRing: [(0, 0), (1, 1), (2, 2), (0, 0)], holes: [[(0, 0), (1, 1), (2, 2), (0, 0)]])!
   let locations = realm.objects(PersonLocation.self).filter("location IN %@", shape)

   let arguments = NSMutableArray()
   arguments.add(shape)
   let locations = realm.objects(PersonLocation.self).filter(NSPredicate(format: "location IN %@", argumentArray: arguments as? [Any])
   ```


Realm/RLMGeospatial.h
```diff
@@ -0,0 +1,228 @@
...
+
+ @note There is no dedicated type to store Geospatial points, instead points should be stored as
+ [GeoJson-shaped](https://www.mongodb.com/docs/manual/reference/geojson/) embedded object, as explained below.
+
+ Persisting geo points in Realm is currently done using duck-typing, which means that any model class with a specific **shape**
+ can be queried as though it contained a geographical location. The recommended approach is using an embedded object.
+
+ @warning This structure cannot be persisted and can only be used to build other geospatial shapes such as (`RLMGeospatialBox`, `RLMGeospatialPolygon` and `RLMGeospatialCircle`).
+ */
+@interface RLMGeospatialPoint : NSObject
```
leemaguire, line 43: Optional altitude is missing

Realm/RLMGeospatial.mm
```diff
@@ -0,0 +1,230 @@
...
+//
+////////////////////////////////////////////////////////////////////////////
+
+#import "RLMGeospatial.h"
+
+#import <realm/geospatial.hpp>
+
+@implementation RLMGeospatial
+- (realm::Geospatial)geoSpatial {
+    return realm::Geospatial{realm::GeoBox{realm::GeoPoint{1.1, 2.2}, realm::GeoPoint{1.1, 2,2}}};
```
leemaguire, line 25: What do the values `{1.1, 2.2}` mean?

CHANGELOG.md
```diff
@@ -1,7 +1,51 @@
 x.y.z Release notes (yyyy-MM-dd)
 =============================================================
 ### Enhancements
-* None.
+* Added initial support for geospatial queries on points.
+  There is no new dedicated type to store Geospatial points, instead points should
+  be stored as ([GeoJson-shaped](https://www.mongodb.com/docs/manual/reference/geojson/)) embedded object, as the example below.
+  ```swift
+  public class Location: EmbeddedObject {
```
leemaguire, line 8: Why are we going for this approach instead of creating a dedicated wrapper class?

RealmSwift/Query.swift
```diff
@@ -824,6 +824,14 @@ extension Query where T: _HasPersistedType, T.PersistedType: _QueryNumeric {
     }
 }
 
+public extension Query where T: OptionalProtocol, T.Wrapped: EmbeddedObject {
```
leemaguire, line 827: Without a tag protocol or dedicated type this means that `geoWithin` can be used on any `EmbeddedObject`. We should expose something to allow conformity.

RealmSwift/Query.swift
```diff
@@ -877,6 +885,8 @@ private indirect enum QueryNode {
 
     case subqueryCount(_ child: QueryNode)
     case mapSubscript(_ keyPath: QueryNode, key: Any)
+
```
leemaguire, line 888: extra newline

CHANGELOG.md
dianaafanador3: This is the proposed approach at the moment, before we add a specific type for storing GeoJson data. Documentation for Swift is not yet added but this is what we are gonna suggest our users https://www.mongodb.com/docs/realm/sdk/dotnet/model-data/data-types/geospatials/#creating-a-geojson-compatible-class

Realm/RLMGeospatial.mm
dianaafanador3: Changed this to a protocol, so we don't need to override it and doesn't need a default value

Realm/RLMGeospatial.h
dianaafanador3: Added this and warning that it doesn't need it for calculations

Realm.xcodeproj/project.pbxproj
```diff
@@ -2858,6 +2876,9 @@
 			isa = XCBuildConfiguration;
 			baseConfigurationReference = 5D659E761BE03E0D006515A0 /* Realm.xcconfig */;
 			buildSettings = {
+				CLANG_ENABLE_MODULES = YES;
```
tgoyne, line 2879: None of these changes should be here. If there are any changes to the project settings which should happen, they should be done in the xcconfig file.

CHANGELOG.md
```diff
@@ -1,7 +1,51 @@
...
+  let locations = realm.objects(PersonLocation.self).where { $0.location.geoWithin(shape) })
+  ```
+  A `filter` or `NSPredicate` can be used as well to perform a Geospatial query.
+  ```swift
+  let shape = GeoPolygon(outerRing: [(0, 0), (1, 1), (2, 2), (0, 0)], holes: [[(0, 0), (1, 1), (2, 2), (0, 0)]])!
+  let locations = realm.objects(PersonLocation.self).filter("location IN %@", shape)
+  
+  let arguments = NSMutableArray()
+  arguments.add(shape)
+  let locations = realm.objects(PersonLocation.self).filter(NSPredicate(format: "location IN %@", argumentArray: arguments as? [Any])
```
tgoyne, line 47: This is an extremely roundabout way of writing this. It can be just `realm.objects(PersonLocation.self).filter("location IN %@", shape)`.

Realm/RLMGeospatial.h
```diff
@@ -0,0 +1,228 @@
...
+@property (readonly) double radians;
+
+/**
+Constructs a `Distance`.
+
+Returns `nil` if the value is lower than 0, because we cannot construct negative distances.
+@param kilometers Distance in kilometers.
+@returns A value that represents the provided distance in radians.
+ */
++ (nullable instancetype)initFromKilometers:(double)kilometers NS_SWIFT_NAME(fromKilometers(kilometers:));
```
tgoyne, line 166: A class function named `init...` doesn't really make any sense; that would be a method which initializes the class itself, not something that returns an instance of the class. A class method which returns a RLMDistance would be `distanceFromKilometers:`.

In Swift the argument label shouldn't be repeated, and there isn't any obvious reason for it to be a factory function rather than an initializer.

Realm/RLMGeospatial.mm
```diff
@@ -0,0 +1,230 @@
...
+    return realm::GeoPoint{_longitude, _latitude};
+}
+
+- (BOOL)isEqual:(id)other {
+    if ([other isKindOfClass:[self class]]) {
+        RLMGeospatialPoint *otherCasted = (RLMGeospatialPoint*)other;
+        if (otherCasted.latitude == self.latitude && otherCasted.longitude == self.longitude)
+            return YES;
+    }
+    return NO;
```
tgoyne, line 49:
```suggestion
    if (auto point = RLMDynamicCast<RLMGeospatialPoint>(other)) {
        return point.latitude == self.latitude && point.longitude == self.longitude;
    }
    return NO;
```

Same thing applies to each of these.

Realm/RLMQueryUtil.mm
```diff
@@ -941,6 +944,16 @@ void validate_and_extract_between_range(id value, RLMProperty *prop, id *from, i
     }
 }
 
+#pragma mark Geospatial
+
+void QueryBuilder::add_within_constraint(const ColumnReference& column, id value) {
+    if ([value isKindOfClass:[RLMGeospatial class]]) {
```
tgoyne: This check is done both here and at the callsite.

Realm/RLMQueryUtil.mm
```diff
@@ -1547,11 +1560,15 @@ NSPredicateOperatorType invert_comparison_operator(NSPredicateOperatorType type)
 
     // turn "key.path IN collection" into ored together ==. "collection IN key.path" is handled elsewhere.
     if (pred.predicateOperatorType == NSInPredicateOperatorType) {
-        process_or_group(m_query, value, [&](id item) {
```
tgoyne, line 1550: The comment above needs to be updated.

RealmSwift/Query.swift
dianaafanador3: @leemaguire we'll need to expose a protocol, or a type, and require the user to conform or inherit from it, which is something we don't want to at the moment given we probably gonna build our own type in the future. Also, developers can as well use IN with filter and NSPredicate which is not gonna need conformance and probably confuse the user. 
This feature is gonna be documented to show  (https://www.mongodb.com/docs/realm/sdk/dotnet/model-data/data-types/geospatials/#creating-a-geojson-compatible-class) and will throw an error on runtime for any value if it is done incorrectly, which probably the developer is going to catch before production. 
Adding extra docstrings to warn the user about this.

RealmSwift/Geospatial.swift
```diff
@@ -0,0 +1,201 @@
...
+public typealias GeoPoint = RLMGeospatialPoint
+
+public extension GeoPoint {
+    /// Initialize a `GeoPoint`, with the specific values for latitude and longitude.
+    ///
+    /// Returns `nil` if the values of latitude and longitude are not within the ranges specified.
+    ///
+    /// - Parameter latitude: Latitude in degrees. Ranges between -90 and 90 degrees, inclusive.
+    /// - Parameter longitude: Longitude in degrees. Ranges between -180 and 180 degrees, inclusive.
+    convenience init?(_ latitude: Double, _ longitude: Double) {
```
tgoyne: This is a place where argument labels are pretty important (since both arguments have the same type) and adding a convenience init without them is just making code using it less readable.

RealmSwift/Geospatial.swift
```diff
@@ -0,0 +1,201 @@
...
+
+public extension GeoPolygon {
+    /// Initialize a `GeoPolygon`, with values for bottom left corner and top right corner.
+    ///
+    /// Returns `nil` if the `GeoPoints` representing a polygon (outer ring or holes), don't have at least 4 points.
+    /// Returns `nil` if the first and the last `GeoPoint` in a polygon are not the same.
+    ///
+    /// - Parameter outerRing: The polygon's external (outer) ring.
+    /// - Parameter holes: The holes (if any) in the polygon.
+    convenience init?(outerRing: [(Double, Double)], holes: [[(Double, Double)]]? = nil) {
```
tgoyne, line 130: `holes` could be non-optional and default to [] as there isn't a semantic difference between empty and nil.

RealmSwift/Geospatial.swift
```diff
@@ -0,0 +1,201 @@
...
+        let outerRingPoints = outerRing.map { GeoPoint(latitude: $0.0, longitude: $0.1) }
+        guard !outerRingPoints.contains(nil) else {
+            return nil
+        }
+        var holesPoints: [[GeoPoint]]?
+        if let holes = holes {
+            let holesMapped = holes.map { $0.map { GeoPoint(latitude: $0.0, longitude: $0.1) } }
+            guard !holesMapped.flatMap({ $0 }).contains(nil) else {
```
tgoyne, line 138: This is a really expensive way of writing this check; it's producing a large temporary array only to immediately discard it after checking if it contains nil. One way to avoid this and the duplicated conversion would be `holesPoints = holes.map { $0.compactMap(GeoPoint.init) }; guard zip(holes, holesPoints).allSatisfy({ $0.count == $1.count }) else { return nil }`.

Realm/RLMGeospatial.h
tgoyne: Actually, modeling the API off [Duration](https://developer.apple.com/documentation/swift/duration) might make sense, which would involve static factory functions like `static func kilometers(_ kilometers: Double) -> Distance` resulting in `GeoCircle(center: (0, 0), radius: .miles(100))`.

Realm/RLMGeospatial.h
```diff
@@ -0,0 +1,248 @@
...
+
+ Persisting geo points in Realm is currently done using duck-typing, which means that any model class with a specific **shape**
+ can be queried as though it contained a geographical location. The recommended approach is using an embedded object.
+
+ @warning This structure cannot be persisted and can only be used to build other geospatial shapes
+ such as (`RLMGeospatialBox`, `RLMGeospatialPolygon` and `RLMGeospatialCircle`).
+
+ @warning Altitude is not used in any of the query calculations.
+ */
+@interface RLMGeospatialPoint : NSObject
```
tgoyne, line 48: These types can all be Sendable as they're immutable after construction.

Realm/RLMGeospatial.h
```diff
@@ -0,0 +1,248 @@
...
+@property (readonly) double radians;
+
+/**
+Constructs a `Distance`.
+
+Returns `nil` if the value is lower than 0, because we cannot construct negative distances.
+@param kilometers Distance in kilometers.
+@returns A value that represents the provided distance in radians.
+ */
++ (nullable instancetype)kilometers:(double)kilometers;
```
tgoyne, line 166: `distanceFromKilometers:(double)kilometers NS_SWIFT_NAME(kilometers:)` and similar for each of these.

Realm/RLMGeospatial.h
```diff
@@ -0,0 +1,248 @@
...
+/// Center of the circle.
+@property (readonly, strong) RLMGeospatialPoint *center;
+/// Radius of the circle.
+@property (readonly) double radians;
+
+/**
+Initialize a `RLMGeospatialCircle`, from its center and radius.
+
+@param center Center of the circle.
+@param radiusInRadians Radius of the circle.
```
tgoyne, line 235: The warning here is correct and needs to be fixed.

Realm/RLMGeospatial.mm
```diff
@@ -0,0 +1,232 @@
...
+
+#import <realm/geospatial.hpp>
+@implementation RLMGeospatialPoint
+- (nullable instancetype)initWithLatitude:(double)latitude longitude:(double)longitude {
+    return [self initWithLatitude:latitude longitude:longitude altitude:0];
+}
+
+- (nullable instancetype)initWithLatitude:(double)latitude longitude:(double)longitude altitude:(double)altitude {
+    if (self = [super init]) {
+        if ((latitude < -90 || latitude > 90) || (longitude < -180 || longitude > 180) || (altitude < 0)) {
```
tgoyne, line 30: Since we're explicitly rejecting out-of-bounds values we probably also want to reject NaN, and as written this'll currently accept NaN for each of the fields (comparisons against NaN always return false).

Realm/RLMQueryUtil.mm
```diff
@@ -941,6 +944,14 @@ void validate_and_extract_between_range(id value, RLMProperty *prop, id *from, i
     }
 }
 
+#pragma mark Geospatial
+
+void QueryBuilder::add_within_constraint(const ColumnReference& column, id value) {
+    id<RLMGeospatial_Private> geospatial = (id<RLMGeospatial_Private>)value;
+    auto geoQuery = column.resolve<Link>().geo_within(geospatial.geoSpatial);
+    m_query.and_query(geoQuery);
```
tgoyne, line 952:
```suggestion
    m_query.and_query(std::move(geoQuery));
```

RealmSwift/Query.swift
```diff
@@ -983,6 +1001,8 @@ private func buildPredicate(_ root: QueryNode, subqueryCount: Int = 0) -> (Strin
             build(keyPath)
             formatStr.append("[%@]")
             arguments.add(key)
+        case .geoWithin(let keyPath, let value):
+            buildExpression(keyPath, "\(QueryNode.Operator.in.rawValue)", value, prefix: nil)
```
tgoyne, line 1005:
```suggestion
            buildExpression(keyPath, QueryNode.Operator.in.rawValue, value, prefix: nil)
```

Realm/RLMGeospatial.mm
```diff
@@ -0,0 +1,223 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+////////////////////////////////////////////////////////////////////////////
+
+#import "RLMGeospatial_Private.hpp"
+#import "RLMUtil.hpp"
+
+#import <realm/geospatial.hpp>
```
leemaguire, line 22: [nitpick] needs a newline

Realm/RLMGeospatial.h
```diff
@@ -0,0 +1,253 @@
...
+ @warning Altitude is not used in any of the query calculations.
+ */
+RLM_SWIFT_SENDABLE
+@interface RLMGeospatialPoint : NSObject
+/// Latitude in degrees.
+@property (readonly) double latitude;
+/// Longitude in degrees.
+@property (readonly) double longitude;
+/// Altitude distance.
+@property (readonly) double altitude;
```
leemaguire, line 55: `altitude` should be nullable

Realm/RLMGeospatial.mm
```diff
@@ -0,0 +1,223 @@
...
+    return self;
+}
+
+- (realm::GeoPoint)value {
+    return realm::GeoPoint{_longitude, _latitude};
+}
+
+- (BOOL)isEqual:(id)other {
+    if (auto point = RLMDynamicCast<RLMGeospatialPoint>(other)) {
+        return point.latitude == self.latitude && point.longitude == self.longitude;
```
leemaguire, line 46: `altitude` needs a check here too

Realm/RLMGeospatial.mm
```diff
@@ -0,0 +1,223 @@
...
+
+@interface RLMGeospatialBox () <RLMGeospatial_Private>
+@end
+
+@implementation RLMGeospatialBox
+- (instancetype)initWithBottomLeft:(RLMGeospatialPoint *)bottomLeft topRight:(RLMGeospatialPoint *)topRight {
+    if (self = [super init]) {
+        _bottomLeft = bottomLeft;
+        _topRight = topRight;
+
```
leemaguire, line 60: [nitpick] remove extra newline

Realm/RLMGeospatial.mm
```diff
@@ -0,0 +1,223 @@
...
+        }
+        _latitude = latitude;
+        _longitude = longitude;
+        _altitude = altitude;
+    }
+    return self;
+}
+
+- (realm::GeoPoint)value {
+    return realm::GeoPoint{_longitude, _latitude};
```
leemaguire, line 41: `altitude` is missing

Realm/RLMQueryUtil.mm
```diff
@@ -941,6 +944,14 @@ void validate_and_extract_between_range(id value, RLMProperty *prop, id *from, i
     }
 }
 
+#pragma mark Geospatial
+
+void QueryBuilder::add_within_constraint(const ColumnReference& column, id value) {
+    id<RLMGeospatial_Private> geospatial = (id<RLMGeospatial_Private>)value;
+    auto geoQuery = column.resolve<Link>().geo_within(geospatial.geoSpatial);
+    m_query.and_query(std::move(geoQuery));
+}
```
leemaguire, line 953:
```suggestion
void QueryBuilder::add_within_constraint(const ColumnReference& column, id<RLMGeospatial_Private> geospatial) {
    auto geoQuery = column.resolve<Link>().geo_within(geospatial.geoSpatial);
    m_query.and_query(std::move(geoQuery));
}
```

RealmSwift/Tests/GeospatialTests.swift
```diff
@@ -0,0 +1,374 @@
...
+        assertGeoPoint(0, -9999999, isNull: true)
+
+        assertGeoPoint(Double.nan, 0, isNull: true)
+        assertGeoPoint(0, Double.nan, isNull: true)
+
+        func assertGeoPoint(_ latitude: Double, _ longitude: Double, isNull: Bool = false) {
+            if isNull {
+                XCTAssertNil(GeoPoint(latitude: latitude, longitude: longitude))
+            } else {
+                XCTAssertNotNil(GeoPoint(latitude: latitude, longitude: longitude))
```
leemaguire, line 103: initialiser with altitude is not tested

Realm/RLMGeospatial.h
dianaafanador3: double cannot be nullable, in case of no value we set 0

tgoyne (approved):

leemaguire (approved):

CHANGELOG.md
```diff
@@ -1,7 +1,49 @@
...
+  try realm.write {
+    realm.add(PersonLocation(name: "Maria", location: Location(latitude: 55.6761, longitude: 12.5683)))
+  }
+  
+  let shape = GeoBox(bottomLeft: (55.6281, 12.0826), topRight: (55.6762, 12.5684))!
+  let locations = realm.objects(PersonLocation.self).where { $0.location.geoWithin(shape) })
+  ```
+  A `filter` or `NSPredicate` can be used as well to perform a Geospatial query.
+  ```swift
+  let shape = GeoPolygon(outerRing: [(0, 0), (1, 1), (2, 2), (0, 0)], holes: [[(0, 0), (1, 1), (2, 2), (0, 0)]])!
```
nirinchev, line 42: [nit] This example shows a polygon where the hole matches the polygon precisely, which is not allowed. I know it's just an example and users are unlikely to ever write this exact same code, but might still be a good idea to show something more realistic.

dianaafanador3: @nirinchev thanks for the input, changing this

PR #8411, opened 2023-11-01
Expose waiting for upload/download on SyncSession

jsflax:


RealmSwift/Sync.swift
```diff
@@ -923,6 +923,34 @@ public extension SyncSession {
...
+            }
+        case .download:
+            __waitForDownloadCompletion(on: nil) { error in
+                block(error)
+            }
+        }
+    }
+    
+    @available(macOS 10.15, iOS 13.0, tvOS 13.0, watchOS 6.0, *)
+    @_unsafeInheritExecutor
```
tgoyne, line 942: This shouldn't need `@_unsafeInheritExecutor` since SyncSession is Sendable.

RealmSwift/Sync.swift
```diff
@@ -923,6 +923,34 @@ public extension SyncSession {
                                                 block(Progress(transferred: transferred, transferrable: transferrable))
         }
     }
+    
+    @preconcurrency
```
tgoyne, line 927: `@preconcurrency` doesn't make sense here.

RealmSwift/Sync.swift
```diff
@@ -923,6 +923,34 @@ public extension SyncSession {
...
+    func wait(for direction: ProgressDirection, block: @Sendable @escaping (Error?) -> Void) {
```
tgoyne, line 928: This should expose the queue parameter for non-async usage.

tgoyne (commented):
Needs a changelog entry.

tgoyne:
Also if our idea is that people should mostly not be using these and that it's potentially a bad idea, the docs really should say that.

jsflax:
I'll leave a warning and mark it in the changelog.

dianaafanador3 (approved):
LGTM!, There are some listing issues to be resolved before merging this

PR #8413, opened 2023-11-03
[Build script] Remove double-quotes around variables to fix incorrect array expansion

hisaac:
In order to use Realm within our app, we need to rebuild it with some extra settings enabled. While experimenting, I noticed what appears to be a bug in Realm's build script:

When adding more than one argument to the `REALM_EXTRA_BUILD_ARGUMENTS` variable, the arguments get passed to `xcodebuild` as a _single string_, rather than multiple space-separated strings.

An example:

```shell
# Running the following command:
REALM_EXTRA_BUILD_ARGUMENTS='REALM_BUILD_LIBRARY_FOR_DISTRIBUTION=YES DEBUG_INFORMATION_FORMAT=dwarf-with-dsym' ./build.sh ios-swift

# Results in this `xcodebuild` invocation:
/Applications/Xcode-15.0.1.app/Contents/Developer/usr/bin/xcodebuild -IDECustomDerivedDataLocation=build/DerivedData -destination generic/platform=iOS -scheme Realm -configuration Release build REALM_HIDE_SYMBOLS=YES "REALM_BUILD_LIBRARY_FOR_DISTRIBUTION=YES DEBUG_INFORMATION_FORMAT=dwarf-with-dsym"
```

Note the last two arguments in the `xcodebuild` invocation are surrounded by double-quotes, so `xcodebuild` interprets them as a single argument.

I've run into this exact issue before when writing my own build scripts in Bash. It has to do with how Bash does array expansion. Luckily the fix is pretty simple, just to remove double-quotes from a couple places.

Here's a diff of the `xcodebuild` output before and after the fix is applied:

```diff
Building with command: xcodebuild -IDECustomDerivedDataLocation=build/DerivedData -destination generic/platform=iOS -scheme Realm -configuration Release build REALM_HIDE_SYMBOLS=YES REALM_BUILD_LIBRARY_FOR_DISTRIBUTION=YES DEBUG_INFORMATION_FORMAT=dwarf-with-dsym
Command line invocation:
-    /Applications/Xcode-15.0.1.app/Contents/Developer/usr/bin/xcodebuild -IDECustomDerivedDataLocation=build/DerivedData -destination generic/platform=iOS -scheme Realm -configuration Release build REALM_HIDE_SYMBOLS=YES "REALM_BUILD_LIBRARY_FOR_DISTRIBUTION=YES DEBUG_INFORMATION_FORMAT=dwarf-with-dsym"
+    /Applications/Xcode-15.0.1.app/Contents/Developer/usr/bin/xcodebuild -IDECustomDerivedDataLocation=build/DerivedData -destination generic/platform=iOS -scheme Realm -configuration Release build REALM_HIDE_SYMBOLS=YES REALM_BUILD_LIBRARY_FOR_DISTRIBUTION=YES DEBUG_INFORMATION_FORMAT=dwarf-with-dsym

User defaults from command line:
    IDECustomDerivedDataLocation = build/DerivedData
    IDEPackageSupportUseBuiltinSCM = YES

Build settings from command line:
-    REALM_BUILD_LIBRARY_FOR_DISTRIBUTION = YES DEBUG_INFORMATION_FORMAT=dwarf-with-dsym
+    DEBUG_INFORMATION_FORMAT = dwarf-with-dsym
+    REALM_BUILD_LIBRARY_FOR_DISTRIBUTION = YES
    REALM_HIDE_SYMBOLS = YES
```


hisaac:
Hi @dianaafanador3. Any chance this could get a review? Just adding you here because I see you approved a recent PR. Let me know if there's someone else I should mention instead. Thanks 👋

fealebenpae:
Hey @hisaac,

I don't know why the original array expansion is surrounded by quotes like that, but I tested a few different invocations and it seems that the problem you spotted appears every time and that your patch makes it go away.

hisaac:
Awesome, thanks for testing it out to make sure @fealebenpae. 🙏

PR #8421, opened 2023-11-17
Allow opening synced realm in-memory.

dianaafanador3:
This will allow setting and in-memory identifier on flexible sync realms, this will not create a .realm file or its associated auxiliary files for the synced realm and instead stores objects in memory while the realm is open and discards them immediately when all instances are closed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,10 @@ x.y.z Release notes (yyyy-MM-dd)
 * The `baseURL` field of `AppConfiguration` can now be updated, rather than the
   value being persisted between runs of the application in the metadata
   storage. ([Core #7201](https://github.com/realm/realm-core/issues/7201))
+* Allow in-memory synced Realms. This will allow setting and in-memory identifier on
+  flexible sync realms, this will not create a .realm file or its associated auxiliary 
+  files for the synced realm and instead stores objects in memory while the realm is 
+  open and discards them immediately when all instances are closed.
 
 ### Fixed
 * `@Persisted`'s Encodable implementation did not allow the encoder to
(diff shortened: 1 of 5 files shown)


Realm/RLMRealmConfiguration.h
```diff
@@ -75,13 +75,13 @@ typedef void(^RLMFlexibleSyncInitialSubscriptionsBlock)(RLMSyncSubscriptionSet *
 
 #pragma mark - Properties
 
-/// The local URL of the Realm file. Mutually exclusive with `inMemoryIdentifier`;
+/// The local URL of the Realm file. Mutually exclusive with `inMemoryIdentifier`,
```
tgoyne, line 78: A semicolon is correct here and a comma is not.

RealmSwift/RealmConfiguration.swift
```diff
@@ -51,8 +51,9 @@ extension Realm {
         /**
          Creates a `Configuration` which can be used to create new `Realm` instances.
 
-         - note: The `fileURL`, `inMemoryIdentifier`, and `syncConfiguration` parameters are mutually exclusive. Only
+         - note: The `fileURL`, `inMemoryIdentifier`, parameters are mutually exclusive. Only
```
tgoyne, line 54: missing "and"

CHANGELOG.md
```diff
@@ -1,7 +1,10 @@
 x.y.z Release notes (yyyy-MM-dd)
 =============================================================
 ### Enhancements
-* None.
+* Allow in-memory synced Realms. This will allow setting and in-memory identifier on
+  flexible sync realms, this will not create a .realm file or its associated auxiliary 
+  files for the synced realm and instead stores objects in memory while the realm is 
+  open and discards them immediately when all instances are closed.
```
tgoyne, line 7: This isn't an accurate description of in-memory Realms. We still create a file and just delete it when it's closed, and never sync to disk so it's unlikely to be actually written.

Realm/ObjectServerTests/AsyncSyncTests.swift
```diff
@@ -514,6 +514,52 @@ class AsyncFlexibleSyncTests: SwiftSyncTestCase {
...
 
+    @MainActor
+    func testFlexibleSyncInitInMemory() async throws {
+        try await populateSwiftPerson(5)
+
+        let user = try await createUser()
+        try await Task {
+            var config = user.flexibleSyncConfiguration(initialSubscriptions: { subs in
+                subs.append(QuerySubscription<SwiftPerson> {
+                    $0.age > 0 && $0.firstName == "\(#function)"
```
tgoyne, line 525: Test should be updated to use `self.name`

rowungiles:
I'm super excited to see this. Thanks @dianaafanador3 and @tgoyne.

PR #8439, opened 2023-12-14
Allow creating notifiers inside write transactions before the first change

tgoyne:
Core has allowed this for a while (and we claimed to allow it as of 10.39.1), but we had our own validation which made it not work.


aehlke:
Thank you, makes sense. I hope this can be prioritized as a "hotfix" without waiting on a major release.

PR #8455, opened 2024-01-09
Add a privacy manifest 

### LINKED ISSUE #8428: Add Privacy Manifest
### Problem

We recently became aware that, according to the new Apple privacy regulations, third-party SDKs will also be required to provide a [Privacy Manifest ](https://developer.apple.com/documentation/bundleresources/privacy_manifest_files) to ensure compliance with privacy policies in the future.

### Solution

Add a Privacy Manifest

### Alternatives

_No response_

### How important is this improvement for you?

Would be a major improvement

### Feature would mainly be used with

-- select --

tgoyne:
Since we don't include the analytics code in builds for the app store, the only thing we need to declare is that we call fstat() on files in the app container, which core uses internally to check the size of Realm files. This is assuming that if the app logs into Atlas that's something that the app itself has to declare rather than us, as we don't automatically talk to the server on our own behalf.

The CocoaPods bundling might not actually work. This is the approach that most of the SDKs on the [mandatory privacy manifest list](https://developer.apple.com/support/third-party-SDK-requirements/) are taking, but since Apple isn't actually checking the manifests yet no one knows if they'll accept this fairly weird way of bundling the xcprivacy file. The less weird way breaks building pods as non-framework static libraries.

Fixes #8428.

### DIFF
--- a/Realm.podspec
+++ b/Realm.podspec
@@ -147,6 +147,7 @@ Pod::Spec.new do |s|
                                 'OTHER_LDFLAGS' => '$(REALM_LD_CLASSIC)',
                               }
   s.preserve_paths          = %w(include scripts)
+  s.resource_bundles        = {'realm_objc_privacy' => ['Realm/PrivacyInfo.xcprivacy']}
 
   s.ios.deployment_target   = '11.0'
   s.osx.deployment_target   = '10.13'
--- a/Realm/PrivacyInfo.xcprivacy
+++ b/Realm/PrivacyInfo.xcprivacy
@@ -1,0 +1,23 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
+<plist version="1.0">
+<dict>
+	<key>NSPrivacyTrackingDomains</key>
+	<array/>
+	<key>NSPrivacyCollectedDataTypes</key>
+	<array/>
+	<key>NSPrivacyAccessedAPITypes</key>
+	<array>
+		<dict>
+			<key>NSPrivacyAccessedAPITypeReasons</key>
+			<array>
+				<string>C617.1</string>
+			</array>
+			<key>NSPrivacyAccessedAPIType</key>
+			<string>NSPrivacyAccessedAPICategoryFileTimestamp</string>
+		</dict>
+	</array>
+	<key>NSPrivacyTracking</key>
+	<false/>
+</dict>
+</plist>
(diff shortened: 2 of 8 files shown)


Realm/PrivacyInfo.xcprivacy
```diff
@@ -0,0 +1,23 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
dianaafanador3, line 1: Why not have just one manifest for both frameworks? Also, I don't think is required to have it in the framework folder, and seems a little bit distracting to have this next to the implementation files.

Realm.podspec
```diff
@@ -147,6 +147,7 @@ Pod::Spec.new do |s|
                                 'OTHER_LDFLAGS' => '$(REALM_LD_CLASSIC)',
                               }
   s.preserve_paths          = %w(include scripts)
+  s.resource_bundles        = {'realm_objc_privacy' => ['Realm/PrivacyInfo.xcprivacy']}
```
dianaafanador3, line 150: https://github.com/SDWebImage/SDWebImage/pull/3649/files What I'm seeing in some other SDKs is they use the name of the framework as the key, so shouldn't be 
s.resource_bundles = {'Realm' => ['Realm/PrivacyInfo.xcprivacy']}, or maybe it is irrelevant given that we don't know if it is going to even work

examples/installation/CocoaPods.xcodeproj/xcshareddata/xcschemes/App.xcscheme
```diff
@@ -0,0 +1,77 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
dianaafanador3, line 1: Any reason for adding this new schema to the installation examples project

dianaafanador3 (approved):
LGTM! Just some minor questions

Realm/PrivacyInfo.xcprivacy
tgoyne: SPM requires it to be within the source root for the target.

Realm.podspec
tgoyne: I based this on what [Flutter](https://github.com/flutter/packages/pull/5846) is doing since they appeared to put a bunch of effort into looking into things.

examples/installation/CocoaPods.xcodeproj/xcshareddata/xcschemes/App.xcscheme
tgoyne: Running the installation test locally was sometimes failing due to the App scheme not existing.

Realm.podspec
stuartmorgan: > What I'm seeing in some other SDKs is they use the name of the framework as the key

FYI, see [this comment](https://github.com/flutter/packages/pull/5846#issuecomment-1881788602) for the relevant discussion about why we aren't doing that.

PR #8488, opened 2024-02-13
Fix some bugs with the new release pipeline

tgoyne:
actions/upload-artifact follows symlinks when asked to archive a directory, which results in the release package having three copies of each macos/catalyst library. To fix this, skip extracting the zip before archiving even though it results in an awkward nested zip.

Building RealmSwift also builds Realm, so we don't need the obj-c specific build jobs.

Docs and update checker publishing were failing due to s3cmd not being installed. Docs publishing still doesn't work because the s3 account we're using doesn't have access to the bucket. Passing the token as a CLI argument hits a bug in s3cmd where access denied errors are reported as the config file being malformed, while passing them as env variables gives more accurate errors.

Using setup-ruby plus bundler to install ruby and gems makes it only take 5-10 seconds and lets us pin specific versions of the gems.


build.sh
```diff
@@ -1327,23 +1283,23 @@ case "$COMMAND" in
...
         VERSION="$(sed -n 's/^VERSION=\(.*\)$/\1/p' "${source_root}/dependencies.list")"
         PRERELEASE_REGEX='alpha|beta|rc|preview'
         if [[ $VERSION =~ $PRERELEASE_REGEX ]]; then
           echo "Pre-release version"
           exit 0
         fi
 
-        s3cmd put --recursive --acl-public --access_key=${AWS_ACCESS_KEY_ID} --secret_key=${AWS_SECRET_ACCESS_KEY} docs/swift_output/ s3://realm-sdks/docs/realm-sdks/swift/${VERSION}/
-        s3cmd put --recursive --acl-public --access_key=${AWS_ACCESS_KEY_ID} --secret_key=${AWS_SECRET_ACCESS_KEY} docs/swift_output/ s3://realm-sdks/docs/realm-sdks/swift/latest/
+        s3cmd put --no-mime-magic --guess-mime-type --recursive --acl-public docs/swift_output/ s3://realm-sdks/docs/realm-sdks/swift/${VERSION}/
```
dianaafanador3, line 1298: are we getting the credentials from the environment values here?

.github/workflows/master-push.yml
```diff
@@ -70,24 +68,22 @@ jobs:
...
-            with:
-              path: realm-examples.zip
-              name: realm-examples
+        - uses: actions/checkout@v4
+        - name: Prepare examples for packaging
+          run: sh -x build.sh release-package-examples
+        - name: Upload examples to artifacts library
+          uses: actions/upload-artifact@v4
+          with:
+            path: realm-examples.zip
```
dianaafanador3, line 77: My mistake as well, but to maintain same order it should be name and path instead of path and name

dianaafanador3 (commented):
LGTM!, Thanks for fixing the issues on this, just some minor comments.
Probably wait for Chris to provide docs credentials before merging this.

dianaafanador3 (approved):

tgoyne:
Successful run with the new tokens: https://github.com/realm/realm-swift/actions/runs/7904978638/job/21576536648

PR #8490, opened 2024-02-19
Introduce `_customRealmProperties()` to all Realm Objects to allow for client-generated RLMProperties (via Swift Macros)

rjchatfield:
## Goals

Allow client code to generate RLMProperties for Objects.

## Background

Late last year I started investigating performance problems in our app. Specifically, App Launch was taking ~700ms to perform the first read from a Realm: without migration, without Realm Sync. I isolated it to the creation of `RLMSchema.shared()`. I then attempted to use Swift Macros to try and generate the schema during compile time, but I had no luck.

Reading the GitHub issues and discussions for "macros", I found these:

 - https://github.com/realm/realm-swift/discussions/8194#discussioncomment-7529043
 - https://github.com/realm/realm-swift/issues/8270#issuecomment-1863122863

@tgoyne was kind enough to reply and shared his experimental work from back in July to do exactly what I wanted. His branch also included some explorations into something similar to Apple's `@Model`, which I'm not interested in at this point. But it was a great jumping off point.

## This PR

This PR introduces the minimum set of changes required for the Swift Macro to work.

- [RLMObjectBase] Add `_customRealmProperties()`
- [RealmSwift] Check _customRealmProperties() from within `getProperties()`
- [RLMProperty] Expose convenience init similar in shape to `@Persisted` property wrapper

## More about the Macro

You can find the macro here: https://github.com/rjchatfield/realm-swift-macro

I'm pretty excited about this. I'm new to macros, and used a lot of what @tgoyne built last year. I stripped it down to just generating `class func _customRealmProperties() -> [RLMProperties]`. I tested this out in my own codebase and saw promising improvements.

Here is an example:

```swift
@CompileTimeSchema
open class FooObject: Object {
    @Persisted(primaryKey: true) var id: String
    @Persisted var name: String
    @Persisted(indexed: true) internal var key: String
    @Persisted public internal(set) var nestedObject: NestedObject?
    @Persisted private var embeddedObjects: List<NestedEmbeddedObject>

    var computed: String { "" }
    func method() {}

    @CompileTimeSchema
    @objc(ObjcNestedObject)
    public class NestedObject: Object {
        @Persisted(primaryKey: true) var id: String
        @Persisted var name2: String!
        @Persisted var nestedNeighbour: FooObject.NestedEmbeddedObject?
    }

    @CompileTimeSchema
    @objc(ObjcNestedEmbeddedObject)
    private final class NestedEmbeddedObject: EmbeddedObject {
        @Persisted var name3: String
    }
}
```

The macro's output:

```swift
open class FooObject: Object {
    @Persisted(primaryKey: true) var id: String
    @Persisted var name: String
    @Persisted(indexed: true) internal var key: String
    @Persisted public internal(set) var nestedObject: NestedObject?
    @Persisted private var embeddedObjects: List<NestedEmbeddedObject>

    var computed: String { "" }
    func method() {}
    @objc(ObjcNestedObject)
    public class NestedObject: Object {
        @Persisted(primaryKey: true) var id: String
        @Persisted var name2: String!
        @Persisted var nestedNeighbour: FooObject.NestedEmbeddedObject?

        public override class func _customRealmProperties() -> [RLMProperty]? {
            return [
                RLMProperty(name: "id", objectType: Self.self, valueType: String.self, primaryKey: true),
                RLMProperty(name: "name2", objectType: Self.self, valueType: String?.self),
                RLMProperty(name: "nestedNeighbour", objectType: Self.self, valueType: FooObject.NestedEmbeddedObject?.self)
            ]
        }
    }
    @objc(ObjcNestedEmbeddedObject)
    private final class NestedEmbeddedObject: EmbeddedObject {
        @Persisted var name3: String

        override class func _customRealmProperties() -> [RLMProperty]? {
            return [
                RLMProperty(name: "name3", objectType: Self.self, valueType: String.self)
            ]
        }
    }

    open override class func _customRealmProperties() -> [RLMProperty]? {
        return [
            RLMProperty(name: "id", objectType: Self.self, valueType: String.self, primaryKey: true),
            RLMProperty(name: "name", objectType: Self.self, valueType: String.self),
            RLMProperty(name: "key", objectType: Self.self, valueType: String.self, indexed: true),
            RLMProperty(name: "nestedObject", objectType: Self.self, valueType: NestedObject?.self),
            RLMProperty(name: "embeddedObjects", objectType: Self.self, valueType: List<NestedEmbeddedObject>.self)
        ]
    }
}
```

I then used this in my own codebase. I annotated the 467x Objects & EmbeddedObjects with `@CompileTimeSchema` (in 322 files). There is a bug in Swift which stopped me using it in 7 places, but that is already fixed in the next version of Swift.

The performance is great.

|| Before | After |
|---|---|---|
| `RLMSchema.shared()` | 612ms | 233ms (45%) |
| | image | image | image |
| | image | image |

This Macro SPM package depends on nothing else by Swift Syntax (which may one day be exposed through the Swift Toolchain and avoid re-compilation). I separated out all the tests into a separate module to avoid some SPM resolution issues I had in my codebase. https://github.com/rjchatfield/realm-swift-macro-tests/tree/main

## Requested stats

```
ProductName:		macOS
ProductVersion:		14.3.1
BuildVersion:		23D60

/Applications/Xcode-15.1.0.app/Contents/Developer
Xcode 15.1
Build version 15C65

pod not found
(not in use here)

/bin/bash
GNU bash, version 3.2.57(1)-release (arm64-apple-darwin23)

/opt/homebrew/bin/carthage
0.39.0
(not in use here)

/usr/bin/git
git version 2.39.3 (Apple Git-145)
```

## Thanks for reading

This is my first PR into Realm. I'm open to any and all feedback at this point. I also understand this change may not be wanted by Realm/Mongo, and I accept what ever you want to do with this PR. Thanks. :)

### DIFF
--- a/RealmSwift/Impl/SchemaDiscovery.swift
+++ b/RealmSwift/Impl/SchemaDiscovery.swift
@@ -47,8 +47,8 @@ extension SchemaDiscoverable {
     public static func _rlmPopulateProperty(_ prop: RLMProperty) { }
 }
 
-internal extension RLMProperty {
-    convenience init(name: String, value: _RealmSchemaDiscoverable) {
+extension RLMProperty {
+    internal convenience init(name: String, value: _RealmSchemaDiscoverable) {
         let valueType = Swift.type(of: value)
         self.init()
         self.name = name
@@ -59,6 +59,28 @@ internal extension RLMProperty {
         if valueType._rlmRequireObjc {
             self.updateAccessors()
         }
+    }
+
+    /// Exposed for Macros.
+    /// Important: Keep args in same order & default value as `@Persisted` property wrapper
+    public convenience init<O: ObjectBase, V: _Persistable>(
+        name: String,
+        objectType _: O.Type,
+        valueType _: V.Type,
+        indexed: Bool = false,
+        primaryKey: Bool = false,
+        originProperty: String? = nil
+    ) {
+        self.init()
+        self.name = name
+        self.type = V._rlmType
+        self.optional = V._rlmOptional
+        self.indexed = primaryKey || indexed
+        self.isPrimary = primaryKey
+        self.linkOriginPropertyName = originProperty
+        V._rlmPopulateProperty(self)
+        V._rlmSetAccessor(self)
+        self.swiftIvar = ivar_getOffset(class_getInstanceVariable(O.self, "_" + name)!)
     }
 }
 
@@ -179,6 +201,9 @@ private func getLegacyProperties(_ object: ObjectBase, _ cls: ObjectBase.Type) -
 }
 
 private func getProperties(_ cls: RLMObjectBase.Type) -> [RLMProperty] {
+    if let props = cls._customRealmProperties() {
+        return props
+    }
     // Check for any modern properties and only scan for legacy properties if
     // none are found.
     let object = cls.init()
--- a/RealmSwift/Impl/SchemaDiscovery.swift
+++ b/RealmSwift/Impl/SchemaDiscovery.swift
@@ -39,6 +39,14 @@ public protocol _RealmSchemaDiscoverable {
     static func _rlmPopulateProperty(_ prop: RLMProperty)
 }
 
+extension RLMObjectBase {
+    /// Allow client code to generate properties (ie. via Swift Macros)
+    @_spi(RealmSwiftPrivate)
+    @objc open class func _customRealmProperties() -> [RLMProperty]? {
+        return nil
+    }
+}
+
 internal protocol SchemaDiscoverable: _RealmSchemaDiscoverable {}
 extension SchemaDiscoverable {
     public static var _rlmOptional: Bool { false }
@@ -63,6 +71,7 @@ extension RLMProperty {
 
     /// Exposed for Macros.
     /// Important: Keep args in same order & default value as `@Persisted` property wrapper
+    @_spi(RealmSwiftPrivate)
     public convenience init<O: ObjectBase, V: _Persistable>(
         name: String,
         objectType _: O.Type,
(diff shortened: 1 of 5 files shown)


nirinchev:
Hey, thanks for this. We'll have a meeting on Monday to go over it - the initial feedback from the team is that the approach seems reasonable.

A few things would be needed before we can accept this though:
1. We'd need you to sign a CLA (the form can be found [here](https://docs.google.com/forms/d/e/1FAIpQLSeQ9ROFaTu9pyrmPhXc-dEnLD84DbLuT_-tPNZDOL9J10tOKQ/viewform)).
2. We'd like to have at least some minimal testing of the feature to make sure we don't accidentally break it in the future. You may want to hold off on this until we're able to give you more concrete feedback after Monday, though, to make sure you don't have to redo them in case the implementation needs bigger changes.

RealmSwift/Impl/SchemaDiscovery.swift
```diff
@@ -60,6 +60,28 @@ internal extension RLMProperty {
...
+    ) {
+        self.init()
+        self.name = name
+        self.type = V._rlmType
+        self.optional = V._rlmOptional
+        self.indexed = primaryKey || indexed
+        self.isPrimary = primaryKey
+        self.linkOriginPropertyName = originProperty
+        V._rlmPopulateProperty(self)
+        V._rlmSetAccessor(self)
```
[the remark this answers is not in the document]
tgoyne: This may be one of the things covered by the dyld cache on iOS, which means that it'll only be slow the first time an app is launched after each installation. If not, it is just frontloading work that'll need to happen at some point. Theoretically it could be done lazily, but right now we take advantage of that the schema is immutable after construction to share it between threads without any locking, and introducing locking could be a meaningful perf hit.

RealmSwift/Impl/SchemaDiscovery.swift
```diff
@@ -60,6 +60,28 @@ internal extension RLMProperty {
             self.updateAccessors()
         }
     }
+
+    /// Exposed for Macros.
+    /// Important: Keep args in same order & default value as `@Persisted` property wrapper
+    public convenience init<O: ObjectBase, V: _Persistable>(
```
tgoyne, line 66: I have not actually verified that this works, but if it does this should be `@_spi(RealmSwiftPrivate)`  so that code using it has to do `@_spi(RealmSwiftPrivate) import RealmSwift` to reflect that this isn't a stable part of the API.

tgoyne (commented):
I think this approach is fine. I had hoped that overriding the existing `sharedSchema` class method would work, but it really wouldn't without significant changes.

rjchatfield:
Updates:

1. I've added tests, as requested. I hope they're good enough.
2. I've edited the author to match my GitHub email address.

RealmSwift/Impl/SchemaDiscovery.swift
rjchatfield: Love this. In fact, I've pushed another commit to remove `_customRealmProperties` from the Objective-C files and done it as an extension in RealmSwift and annotated it with this `@_spi`. Since we only want to expose it to Swift Macros, makes sense yeah? 

What do you think?

PR #8496, opened 2024-02-23
Upgrade to core 14

tgoyne:
All tests now pass. Still need to fix some performance regressions.


tgoyne:
Needs another core release to pick up the final perf regression fix before it can actually be merged, but the code changes here are now reviewable.

tgoyne:
Core 14.3.0 should be good for an actual release.

Realm/Tests/PerformanceTests.m
```diff
@@ -591,8 +591,65 @@ - (void)testCommitWriteTransactionWithListNotification {
...
+    CFRunLoopRun();
+
+    [self measureMetrics:self.class.defaultPerformanceMetrics automaticallyStartMeasuring:NO forBlock:^{
+        [realm beginWriteTransaction];
+        for (StringObject *so in [StringObject allObjectsInRealm:realm]) {
+            so.stringCol = @"a";
+        }
+        [realm commitWriteTransaction];
+
+        [self startMeasuring];
```
dianaafanador3, line 617: What are we measuring exactly on this performance test?, The time committing changes to the 5 objects?

Realm/RLMQueryUtil.mm
```diff
@@ -726,10 +724,55 @@ Query make_diacritic_insensitive_constraint(NSPredicateOperatorType operatorType
...
+                                                                C&& column, T&& value) {
+    bool caseSensitive = !(predicateOptions & NSCaseInsensitivePredicateOption);
+    Query condition = make_diacritic_sensitive_constraint(operatorType, caseSensitive, column, value);
+
+    // Queries on Mixed used to coerce Strings to Binary and vice-versa. Core
+    // no longer does this, but we can maintain compatibility by doing the
+    // coercion and checking both
+    if constexpr (is_any_v<C, Columns<Mixed>, Columns<Lst<Mixed>>, Columns<Set<Mixed>>, Columns<Dictionary>>) {
+        Mixed m = value;
+        if (!m.is_null()) {
```
dianaafanador3, line 746: With this changes are we replacing some core removed implementation?, to maintain compatibility. Does that mean this is no longer needed, and we can remove this in major a version?

dianaafanador3 (approved):
LGTM!, Just some minor questions.
Also, changes to the performance tests number of iterations is because we want to evaluate performance for bigger operations?

RealmSwift/Tests/SectionedResultsTests.swift
```diff
@@ -1557,7 +1556,7 @@ struct SectionedResultsTestDataAnyRealmValue: SectionedResultsTestData {
     }
 
     static func orderedKeys(ascending: Bool) -> [String] {
-        return ["alphanumeric", "data"]
+        return ascending ? ["alphanumeric", "data"] : ["data", "alphanumeric"]
```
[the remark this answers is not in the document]
tgoyne: The sort order used to be that the values were ordered ["alphanumeric", "data", "data", "alphanumeric"]. The order of the sections is determined by the order of the values and not by sorting the keys, so it ended up being the same regardless of what order the objects were sorted.

RealmSwift/Tests/PerformanceTests.swift
```diff
@@ -916,29 +904,27 @@ class SwiftPerformanceTests: TestCase {
...
                 realm.create(ModernSwiftStringObject.self, value: [obj.stringCol])
             }
-        })
+        }
 
-        measure {
-            for _ in 0..<500 {
-                _ = realm.objects(ModernSwiftStringProjection.self)
-            }
+        measure(times: 5) {
```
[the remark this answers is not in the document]
tgoyne: This test previously wasn't actually doing much of anything and the changed tests is doing a lot more.

Realm/Tests/PerformanceTests.m
tgoyne: The time to produce the change notification, which will stop the run loop once it's delivered.

Realm/RLMQueryUtil.mm
tgoyne: Correct, this can go away with a major version bump. I'll add a comment noting that.

PR #8502, opened 2024-03-05
Update packaging for Xcode 15.3

tgoyne:
Swift 5.10 fixed some holes in sendability checking resulting in some new warnings. I fixed most of them in the library itself, but didn't try to fix the ones in the tests.

Xcode cloud no longer allows manually starting a disabled workflow (but reports a 500 server error instead of a useful error), but also fixed the bug where you couldn't actually enable workflows via the api.


tgoyne:
Passing release pipeline job with this: https://github.com/realm/realm-swift/actions/runs/8180887917

nirinchev (approved):

ci_scripts/ci_post_clone.sh
```diff
@@ -77,6 +77,11 @@ if [[ "$target" == "release-package-build-"* ]]; then
     sed -i '' "s/REALM_HIDE_SYMBOLS = NO;/REALM_HIDE_SYMBOLS = YES;/" "$filename"
 fi
 
+# Xcode cloud currently doesn't have visionOS installed for Xcode 15.3
+if [[ "$CI_WORKFLOW" == release-package-build-vision*_15.3 ]]; then
```
dianaafanador3, line 81: From experience this was causing some timeout issues when running the workflow. Is that still the case?

RealmSwift/Impl/RealmCollectionImpl.swift
```diff
@@ -185,7 +185,15 @@ internal func with<A: Actor, Value: ThreadConfined, Return: Sendable>(
         let unchecked = Unchecked(wrappedValue: value)
         return try await actor.invoke { actor in
             if !Task.isCancelled {
+#if swift(>=5.10)
```
dianaafanador3, line 188: I'm deducing this is only happening in 5.10, can we exclude future swift releases from this fix?

RealmSwift/SwiftUI.swift
```diff
@@ -134,23 +146,20 @@ private func createEquatableBinding<T: ThreadConfined, V: Equatable>(
         }
 
         func cancel() {
-            removeObservers()
-            SwiftUIKVO.observedObjects.removeValue(forKey: value)
+            SwiftUIKVO.observedObjects.withLock {
```
dianaafanador3, line 149: Couldn't this directly call
SwiftUIKVO.cancel(value)

dianaafanador3 (approved):
LGTM!; Just a minor comments and questions!

RealmSwift/Impl/RealmCollectionImpl.swift
tgoyne: It's fixed on swift main, but until there's a release with the fix we don't know exactly which versions to check.

PR #8511, opened 2024-03-11
Add NSPrivacyAccessedAPICategoryDiskSpace to the privacy manifest

tgoyne:
I thought this was only called in tests, but it turns out to also be used in the backup code.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,26 @@
+x.y.z Release notes (yyyy-MM-dd)
+=============================================================
+### Enhancements
+* None.
+
+### Fixed
+* <How to hit and notice issue? what was the impact?> ([#????](https://github.com/realm/realm-swift/issues/????), since v?.?.?)
+* The Realm.framework privacy manifest was missing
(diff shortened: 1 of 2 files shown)


WGMB:
Yes, we need this to be merged to master and there's a need for a new release to Cocoapods.

dianaafanador3 (approved):

cristiancardosodexcom:
When this is going to be merged https://github.com/realm/realm-core/pull/7570#issuecomment-2052700766

nirinchev:
I'm sorry, but asking the same question in multiple issues/PRs is not going to get the feature released faster. If anything, it's taking time from us to go and respond to those. As discussed in https://github.com/realm/realm-swift/issues/8535, we've committed to releasing the changes before the May 1st deadline. I understand the sense of urgency here, but 1) Apple still accepts submissions without a privacy manifest and 2) there are workarounds proposed in the issue - such as adding your own privacy manifest that describes the missing API.

I'll lock this and similar issues as right now, they're overly noisy and our position hasn't changed - there will be at least one release of the Realm Swift SDK prior to May 1st with the necessary privacy manifest changes.

PR #8515, opened 2024-03-13
Add build-binaries workflow

nirinchev:


nirinchev:
Merging this since I need the workflow in master to be able to run it. The actual implementation will come as a follow-up PR.

PR #8517, opened 2024-03-13
RCOCOA-2308: Add GHA workflow for core prebuilds

nirinchev:
Successful run: https://github.com/realm/realm-swift/actions/runs/8269909586 (the url in the combine xcframework step has a typo, but that's fixed already).

This created prebuilds for Core v13.27.0 which can be found at https://static.realm.io/downloads/core/v13.27.0/cocoa/realm-monorepo-xcframework-v13.27.0.tar.xz.


.github/workflows/build-binaries.yml
```diff
@@ -1,12 +1,22 @@
 name: Build Core binaries
-on: workflow_dispatch
+on:
+  workflow_dispatch:
+    inputs:
+      core-version:
+        type: string
+        required: false
+        default: ''
+        description: Core version (tag/ref) to use to generate the binaries for. If not provided, the version in dependencies.list will be used
```
tgoyne, line 9: (tag/ref) implies that you can build things other than tagged versions with this but I'm pretty sure doing so will fail as `REALM_VERSION` needs to be something that can be parsed as a version. To support that it'd need to do something along the lines of checking out the supplied branch but using the result of `git describe` as the version everywhere else.

dependencies.list
```diff
@@ -1,3 +1,3 @@
-VERSION=10.48.0
-REALM_CORE_VERSION=13.26.0
+VERSION=10.47.0
```
tgoyne, line 1: Reverting the version number here.

dependencies.list
```diff
@@ -1,3 +1,3 @@
...
+REALM_CORE_VERSION=v13.26.0
```
tgoyne, line 2: `set-core-version` needs to be updated to set this correctly.

Package.swift
```diff
@@ -153,7 +153,7 @@ let package = Package(
             targets: ["Realm", "RealmSwift"]),
     ],
     dependencies: [
-        .package(url: "https://github.com/realm/realm-core.git", exact: coreVersion)
+        .package(url: "https://github.com/realm/realm-core.git", revision: "v\(coreVersion)")
```
tgoyne, line 156: This will make it so that the package cannot be imported. SPM has some very strange rules around non-version-based dependencies.

.github/workflows/build-binaries.yml
```diff
@@ -15,7 +25,92 @@ jobs:
...
+            REALM_CORE_VERSION=$(sed -n 's/^REALM_CORE_VERSION=\(.*\)$/\1/p' dependencies.list)
+          fi
+          echo "version=$REALM_CORE_VERSION" >> "$GITHUB_OUTPUT"
+
+      - name: Checkout Core
+        uses: actions/checkout@v4
+        with:
+          repository: realm/realm-core
+          path: core
+          submodules: recursive
```
tgoyne, line 48: We shouldn't need core's submodules

nirinchev: Those are needed to build the prebuilds. Otherwise, I get errors like this one when running `build-apple-device.sh`:

```
CMake Error at external/CMakeLists.txt:4 (add_subdirectory):
  The source directory

    /Users/runner/work/realm-swift/realm-swift/core/external/catch

  does not contain a CMakeLists.txt file.
```

Package.swift
tgoyne: No, a version tag is a hard requirement. We sometimes temporarily set it to a branch or other thing for local development or for running CI tests on PRs, but that has to be reverted before a release.

.github/workflows/build-binaries.yml
nirinchev: I've clarified this now. This should either be a tag or the output of `git describe`. It doesn't appear like we can do actual releases with a `git describe`-d version, but we can still create prebuilds and use them during development if we wanted to run a CI build on all platforms from an untagged version of Core.

Package.swift
nirinchev: Okay, I read the docs and it appears we always have to release with a tagged version of Core.

tgoyne (approved):

PR #8536, opened 2024-04-07
RCOCOA-2310: Fix a crash when receiving 401/403 when opening a watch stream

nirinchev:
It looks like we can end up in a situation where `URLSession:didCompleteWithError` is invoked with `nil` error, which results in a crash if we try to insert it as a value for `NSError.userInfo` dictionary. With this change we insert `[NSNull null]` for `NSUnderlyingErrorKey` instead.

I've added a test that uses the watch functionality as that appears to be the only thing that exercises the `RLMEventSessionDelegate` codepaths. It's not the cleanest test, but I couldn't come up with anything better that reproduces the crash.

Fixes https://github.com/realm/realm-swift/issues/8519

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,8 @@ x.y.z Release notes (yyyy-MM-dd)
 
 ### Fixed
 * <How to hit and notice issue? what was the impact?> ([#????](https://github.com/realm/realm-swift/issues/????), since v?.?.?)
-* Fixed a crash that would occur when an http error 401 or 403 is returned upon opening a watch stream for a MongoDB collection. ([#8519](https://github.com/realm/realm-swift/issues/8519))
+* Fixed a crash that would occur when an http error 401 or 403 is returned upon
+  opening a watch stream for a MongoDB collection. ([#8519](https://github.com/realm/realm-swift/issues/8519))
 
 <!-- ### Breaking Changes - ONLY INCLUDE FOR NEW MAJOR version -->
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,7 @@ x.y.z Release notes (yyyy-MM-dd)
 
 ### Fixed
 * <How to hit and notice issue? what was the impact?> ([#????](https://github.com/realm/realm-swift/issues/????), since v?.?.?)
-* None.
+* Fixed a crash that would occur when an http error 401 or 403 is returned upon opening a watch stream for a MongoDB collection. ([#8519](https://github.com/realm/realm-swift/issues/8519))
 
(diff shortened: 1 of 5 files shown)


Realm/ObjectServerTests/WatchTestUtility.swift
```diff
@@ -26,10 +26,13 @@ final public class WatchTestUtility: ChangeEventDelegate {
     private let openExpectation: XCTestExpectation
     private let closeExpectation: XCTestExpectation
     private var changeExpectation: XCTestExpectation?
+    private let expectError: Bool
+    public var closeError: Error?
```
dianaafanador3, line 30:
```suggestion
    public var didCloseError: Error?
```

CHANGELOG.md
```diff
@@ -5,7 +5,7 @@ x.y.z Release notes (yyyy-MM-dd)
 
 ### Fixed
 * <How to hit and notice issue? what was the impact?> ([#????](https://github.com/realm/realm-swift/issues/????), since v?.?.?)
-* None.
+* Fixed a crash that would occur when an http error 401 or 403 is returned upon opening a watch stream for a MongoDB collection. ([#8519](https://github.com/realm/realm-swift/issues/8519))
```
dianaafanador3, line 8: This probably needs a line jump

Realm/ObjectServerTests/SwiftObjectServerTests.swift
```diff
@@ -1322,6 +1322,32 @@ class SwiftObjectServerTests: SwiftSyncTestCase {
...
+        syncUser.refreshCustomData().awaitFailure(self)
+
+        // This verifies that we don't crash in RLMEventSessionDelegate when creating a watch stream
+        // with a revoked user. See https://github.com/realm/realm-swift/issues/8519
+        let watchTestUtility = WatchTestUtility(testCase: self, expectError: true)
+        _ = syncUser.collection(for: Dog.self, app: app).watch(delegate: watchTestUtility)
+        watchTestUtility.waitForOpen()
+        watchTestUtility.waitForClose()
+
+        XCTAssertEqual(watchTestUtility.closeError?.localizedDescription, "URLSession HTTP error code: 403")
```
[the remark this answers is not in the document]
dianaafanador3: Should we also verify that the error is nil

dianaafanador3 (approved):
LGTM!, Just a few comments

CHANGELOG.md
nirinchev: Sure - I typically don't add line breaks in markdown files, but can see this is how the changelog is written already, so might as well stick to the precedent in the swift repo.

PR #8537, opened 2024-04-07
RCOCOA-2305: Update base url to point to services.cloud.mongodb.com

nirinchev:
Fixes https://github.com/realm/realm-swift/issues/8512


Realm/RLMAsyncTask.mm
```diff
@@ -57,7 +59,7 @@ @implementation RLMAsyncOpenTask {
 }
 
 - (void)addProgressNotificationOnQueue:(dispatch_queue_t)queue block:(RLMProgressNotificationBlock)block {
-    auto wrappedBlock = ^(NSUInteger transferred_bytes, NSUInteger transferrable_bytes) {
+    auto wrappedBlock = ^(NSUInteger transferred_bytes, NSUInteger transferrable_bytes, double estimate) {
         dispatch_async(queue, ^{
```
dianaafanador3, line 63: If not used, this doesn't needs to be a named parameter

Realm/RLMApp.mm
```diff
@@ -223,7 +223,7 @@ static void setOptionalString(std::optional<std::string>& dst, NSString *src) {
 }
 
 - (NSString *)baseURL {
-    return getOptionalString(_config.base_url);
+    return getOptionalString(_config.base_url) ?: RLMStringViewToNSString(app::App::default_base_url());
```
dianaafanador3, line 226: Shouldn't this be done by Core, and return the default url if there is no given url

dianaafanador3 (commented):
just some minor comments

Realm/RLMApp.mm
nirinchev: I guess it could, but they haven't done it 😄 I think it's a fair decision on their part as they probably want to keep the struct fairly simple and have the defaults be managed elsewhere.

Realm/RLMApp.h
```diff
@@ -42,7 +42,7 @@ typedef void(^RLMOptionalErrorBlock)(NSError * _Nullable);
 @interface RLMAppConfiguration : NSObject <NSCopying>
 
 /// A custom base URL to request against.
-@property (nonatomic, strong, nullable) NSString *baseURL;
+@property (nonatomic, strong, null_resettable) NSString *baseURL;
```
dianaafanador3, line 45: Also, the docstring should be updated to reflect that now we return the default url if there is no given base url, if we are going in that direction

dianaafanador3 (approved):
LGTM! Looks like there is still some linting issue. Finally, before merging I'll separate this into two commits, one for the fix and one for the core update

PR #8539, opened 2024-04-08
Update for the app services split branch

tgoyne:
https://github.com/realm/realm-core/pull/7300


Realm/RLMApp.mm
```diff
@@ -334,16 +330,16 @@ - (instancetype)initWithApp:(std::shared_ptr<realm::app::App>&&)app config:(RLMA
 - (instancetype)initWithConfiguration:(RLMAppConfiguration *)configuration {
     if (self = [super init]) {
         _app = RLMTranslateError([&] {
-            return app::App::get_app(app::App::CacheMode::Enabled, configuration.config, configuration.clientConfig);
+            return app::App::get_app(app::App::CacheMode::Disabled, configuration.config);
```
nirinchev, line 333: Why are we disabling the caching here? Is it because we're anyway maintaining an SDK-level cache?

Realm/RLMSyncSession.h
```diff
@@ -219,12 +219,16 @@ NS_REFINED_FOR_SWIFT;
...
+/// :nodoc:
++ (void)immediatelyHandleError:(RLMSyncErrorActionToken *)token syncManager:(RLMSyncManager *)syncManager
+    __attribute__((deprecated("The syncManager: parameter is no longer required")));
+
 /**
  Given an error action token, immediately handle the corresponding action.
- 
- @see `RLMSyncErrorClientResetError`, `RLMSyncErrorPermissionDeniedError`
+
+ @see ```RLMSyncErrorClientResetError``, ``RLMSyncErrorPermissionDeniedError``
```
nirinchev, line 229: Looks like we have an extra backtick here

```suggestion
 @see ``RLMSyncErrorClientResetError``, ``RLMSyncErrorPermissionDeniedError``
```

nirinchev (approved):

dianaafanador3 (approved):

Realm/RLMApp.mm
tgoyne: Yeah, we never actually get a cache hit in the core cache unless something has gone wrong.

PR #8542, opened 2024-04-10
RCOCOA-2333: Add tests for sync schema migrations

nirinchev:
This adds tests for sync schema migrations. The feature is enabled by specifying a non-zero `schemaVersion` on the realm configuration, so we already supported it, so this is only adding test to ensure the value is correctly routed to Core/Sync/Server.


clementetb:
I have pushed an update to get the tests to correctly trigger the schema migration. Originally they weren't triggered correctly because each test step used a different Realm sync configuration (each had a different user), thus no migration was initiated.

We have exposed `Realm.Configuration.cache` because testing these scenarios requires Realm instances to be unique. When cached only the first schema configuration would be taken.

We have exposed `Realm.close` because tests require closing any open Realm instance before triggering a sync migration.

These two properties and methods have been exposed as internal, as they are only required to be accessed from tests.

Finally, a new API `Realm.requiresMigration` tells if the Realm file that points the configuration to would require a migration. It does so by comparing the schema versions of the file and the one in the configuration.

PR #8546, opened 2024-04-12
Collections in Mixed

dianaafanador3:


dianaafanador3:
This is waiting for this [PR](https://github.com/realm/realm-core/pull/7478) to be merged to core. Also, there is one failing test which is working locally, taking a look at it, but putting this on review on the meantime.

Realm/RLMAccessor.mm
```diff
@@ -890,15 +914,17 @@ void RLMSetSwiftPropertyAny(__unsafe_unretained RLMObjectBase *const obj, uint16
 }
 
 id RLMAccessorContext::box(realm::Mixed v) {
-    return RLMMixedToObjc(v, _realm, &_info);
+    auto property = (currentProperty) ? currentProperty : _info.propertyForTableColumn(_colKey);
```
tgoyne, line 917: `currentProperty ?: _info.propertyForTableColumn(_colKey);`

Realm/RLMAccessor.mm
```diff
@@ -1100,7 +1127,7 @@ static auto toOptional(__unsafe_unretained id const value) {
 
     try {
         realm::Object::create(*this, _realm->_realm, *_info.objectSchema,
-                              (id)value, policy, existingKey, outObj);
+                              realm::util::any_cast<id>(value), policy, existingKey, outObj);
```
tgoyne, line 1130: We should not be using std::any anywhere. This is constructing a temporary std::any from value and then reading the value right back out of it.

Realm/RLMArray_Private.hpp
```diff
@@ -39,7 +39,7 @@ class RLMObservationInfo;
     BOOL _optional;
 @public
     // The name of the property which this RLMArray represents
-    NSString *_key;
+    RLMProperty *_property;
```
tgoyne, line 42: Comment needs to be updated.

Realm/RLMConstants.h
```diff
@@ -92,8 +94,12 @@ typedef RLM_CLOSED_ENUM(int32_t, RLMPropertyType) {
     /** Realm linking objects. See [Realm Models](https://www.mongodb.com/docs/realm/sdk/swift/fundamentals/relationships/#inverse-relationship) for more information. */
     RLMPropertyTypeLinkingObjects = 8,
 
-    RLMPropertyTypeObjectId = 10,
-    RLMPropertyTypeDecimal128 = 11
+#pragma mark - Collection types
+
+    /** Dictionary: `RLMDictionary`, `Map` (Swift) */
+    RLMPropertyTypeDictionary = 512,
```
tgoyne, line 100: These aren't property types, and adding these to this enum is a significant violation of the design. This project is making the set of types which can be stored in mixed no longer the same as the primitive property types, so it needs to be introducing a new enum for that.

Realm/RLMManagedArray.mm
```diff
@@ -198,9 +198,19 @@ - (NSUInteger)countByEnumeratingWithState:(NSFastEnumerationState *)state
...
-        RLMAccessorContext context(*_objectInfo);
-        return _backingList.get(context, index);
+    return translateErrors([&]() -> id {
+        realm::Mixed mixed = _backingList.get_any(index);
+        if (mixed.is_type(realm::type_Dictionary)) {
+            return [[RLMManagedDictionary alloc] initWithBackingCollection:_backingList.get_dictionary(index) parentInfo:_ownerInfo property:_property];
+        }
+        else if (mixed.is_type(realm::type_List)) {
+            return [[RLMManagedArray alloc] initWithBackingCollection:_backingList.get_list(index) parentInfo:_ownerInfo property:_property];
+        }
```
tgoyne, line 208: This is all handled by `List::get()` and doesn't need to be special-cased here or in RLMManagedDictionary.

Realm/RLMManagedArray.mm
```diff
@@ -227,7 +238,8 @@ - (void)insertObject:(id)object atIndex:(NSUInteger)index {
 - (void)insertObjects:(id<NSFastEnumeration>)objects atIndexes:(NSIndexSet *)indexes {
     changeArray(self, NSKeyValueChangeInsertion, indexes, ^{
         NSUInteger index = [indexes firstIndex];
-        RLMAccessorContext context(*_objectInfo);
+        RLMAccessorContext context(*_ownerInfo, *_objectInfo);
+        context.currentProperty = _property;
```
tgoyne, line 242: Rather than setting the currentProperty separately each time it should be added as an argument to the constructor.

Realm/RLMManagedDictionary.mm
```diff
@@ -277,17 +295,19 @@ - (void)removeObjectsForKeys:(NSArray *)keyArray {
...
 
 - (void)enumerateKeysAndObjectsUsingBlock:(void (^)(id key, id obj, BOOL *stop))block {
-    RLMAccessorContext c(*_objectInfo);
+    RLMAccessorContext context(*_ownerInfo, *_objectInfo);
+    context.currentProperty = _property;
     BOOL stop = false;
     @autoreleasepool {
         for (auto&& [key, value] : _backingCollection) {
-            block(c.box(key), c.box(value), &stop);
+            block(context.box(key), [self objectForKey:context.box(key)], &stop);
```
tgoyne, line 310: This is probably a significant pessimization to iteration performance?

Realm/RLMQueryUtil.hpp
```diff
@@ -37,3 +38,4 @@ realm::Query RLMPredicateToQuery(NSPredicate *predicate, RLMObjectSchema *object
 
 // return property - throw for invalid column name
 RLMProperty *RLMValidatedProperty(RLMObjectSchema *objectSchema, NSString *columnName);
+void RLMGetPathElements(std::vector<realm::PathElement> &paths, NSExpression *expression);
```
tgoyne, line 41: This is only used in RLMQueryUtil.mm and shouldn't be exposed in the header.

Realm/RLMQueryUtil.mm
```diff
@@ -1684,23 +1691,27 @@ bool is_self_value_for_key_path_function_expression(NSExpression *expression)
...
-        NSExpression *mapItems = [functionExpression.arguments firstObject];
-        NSExpression *linkCol = [[functionExpression.operand arguments] firstObject];
-        NSExpression *mapCol = [mapItems.arguments firstObject];
-        mapKey = [mapItems.arguments[1] constantValue];
-        keyPath = [NSString stringWithFormat:@"%@.%@", linkCol.keyPath, mapCol.keyPath];
-    } else {
-        keyPath = [functionExpression.arguments.firstObject keyPath];
-        mapKey = [functionExpression.arguments[1] constantValue];
+    NSExpression *keyPathExpression = functionExpression;
+    for (int i = 0; i < pathElements.size(); i++) {
```
tgoyne, line 1699: This loop doesn't make any sense? I think this is just doing the following?

```
NSExpression *keyPathExpression = functionExpression;
    if (pathElements.size() > 1 && keyPathExpression.arguments[0].expressionType != NSKeyPathExpressionType) {
        keyPathExpression = keyPathExpression.arguments[0];
    }
    NSString *keyPath = keyPathExpression.arguments[0].description;
```

Realm/RLMQueryUtil.mm
```diff
@@ -1881,3 +1892,30 @@ bool is_self_value_for_key_path_function_expression(NSExpression *expression)
                     @"Property '%@' not found in object of type '%@'", columnName, desc.className);
     return prop;
 }
+
+void RLMGetPathElements(std::vector<PathElement> &paths, NSExpression *expression) {
```
tgoyne, line 1896: I don't see any reason for the paths not to just be a return value here.

It is ver non-obvious what is going on in this function and why, and it needs some comments explaining.

Realm/RLMAccessor.mm
```diff
@@ -126,17 +128,19 @@ void setValueOrNull(__unsafe_unretained RLMObjectBase *const obj, ColKey col,
...
-            if constexpr (std::is_same_v<T, realm::Mixed>) {
-                obj->_row.set(col, RLMObjcToMixed(value, obj->_realm, realm::CreatePolicy::SetLink));
-            }
-            else {
+        if constexpr (std::is_same_v<T, realm::Mixed>) {
+            realm::Object o(obj->_realm->_realm, *obj->_info->objectSchema, obj->_row);
+            RLMAccessorContext ctx(obj);
+            RLMProperty *property = obj->_info->propertyForTableColumn(col);
+            o.set_property_value(ctx, getProperty(obj, property).name, value ?: NSNull.null);
+        } else {
```
tgoyne, line 136: Our obj-c code has a newline before else.

RealmSwift/AnyRealmValue.swift
```diff
@@ -156,4 +178,28 @@ public enum AnyRealmValue: Hashable {
...
+        return AnyRealmValue.dictionary(map)
+    }
+
+    /// Returns a `AnyRealmValue` storing a `List`.
+    ///
+    /// - Parameter array: A Swift's array of `AnyRealmValue`.
+    /// - Returns: Returns a `AnyRealmValue` storing a `List`.
+    public static func fromArray(_ array: Array<AnyRealmValue>) -> AnyRealmValue {
+        let list = List<AnyRealmValue>()
+        array.forEach { item in
```
tgoyne, line 200: `list.append(array)`

RealmSwift/List.swift
```diff
@@ -361,6 +361,14 @@ extension List: MutableCollection {
     }
 }
 
+// MARK: - Hashable
+
+extension List: Hashable {
```
tgoyne, line 366: This is not a valid definition of Hashable for List, and such a thing is probably not possible. Using a List as a dictionary key does not really make any sense too.

RealmSwift/Map.swift
```diff
@@ -779,6 +779,14 @@ public final class Map<Key: _MapKey, Value: RealmCollectionValue>: RLMSwiftColle
     }
 }
 
+// MARK: - Hashable
+
+extension Map: Hashable {
```
tgoyne, line 784: Similarly, this is not at all a valid Hashable implementation.

RealmSwift/Tests/ObjectCreationTests.swift
```diff
@@ -70,54 +70,6 @@ class ObjectCreationTests: TestCase {
             SwiftOptionalDefaultValuesObject.defaultValues(), boolObjectValue: true)
     }
 
-    func testInitWithDictionary() {
```
tgoyne, line 73: Why are these tests deleted? They're still applicable and unrelated to this project.

tgoyne (commented):
A first pass over things.

Realm/RLMQueryUtil.mm
```diff
@@ -1881,3 +1892,30 @@ bool is_self_value_for_key_path_function_expression(NSExpression *expression)
...
+    for (int i = 0; i < 2; i++) {
```
leemaguire, line 1897: `2` is a magic number here, can you create a comment explaining this logic

Realm/RLMValue.h
```diff
@@ -22,6 +22,8 @@
 #import <Realm/RLMObjectBase.h>
 #import <Realm/RLMObjectId.h>
 #import <Realm/RLMProperty.h>
+#import <Realm/RLMDictionary.h>
```
leemaguire, line 25: order the includes alphabetically

Realm/Tests/KVOTests.mm
```diff
@@ -506,14 +506,14 @@ - (void)testRemoveObserver {
     XCTAssertNoThrow([obj addObserver:self forKeyPath:@"int32Col" options:0 context:context1]);
     XCTAssertThrows([obj removeObserver:self forKeyPath:@"int32Col" context:context2]);
     XCTAssertNoThrow([obj removeObserver:self forKeyPath:@"int32Col" context:context1]);
-
+
```
leemaguire, line 509: delete the whitespace

Realm/Tests/NotificationTests.m
```diff
@@ -262,7 +262,7 @@ - (NSInteger)row {
 }
 @end
 
-static RLMCollectionChange *getChange(RLMTestCase<ChangesetTestCase> *self, void (^block)(RLMRealm *)) {
+static RLMCollectionChange *getChange(RLMTestCase<ChangesetTestCase> *self, void (^block)(RLMRealm *realm)) {
```
leemaguire, line 265: `realm` looks unused?

Realm/Tests/ObjectTests.m
```diff
@@ -785,7 +785,7 @@ - (void)testCreateInRealmValidationForDictionary {
         }
 
         if ([keyToInvalidate isEqualToString:@"anyCol"]) {
-            obj = self;
+            continue;
         }
```
leemaguire, line 789: remove this

RealmSwift/AnyRealmValue.swift
```diff
@@ -21,6 +21,8 @@ import Foundation
 import Realm
 
 /// A enum for storing and retrieving values associated with an `AnyRealmValue` property.
+/// `AnyRealmValue` can be as well a collection (List, Dictionary) of `AnyRealmValue`, meaning that you can have
```
leemaguire, line 24:
```suggestion
/// `AnyRealmValue` can also store a collection (List, Dictionary) of `AnyRealmValue`, meaning that you can have
```

RealmSwift/Map.swift
dianaafanador3: Using ObjectIdentifier given that we already conform to Equatable

RealmSwift/Tests/ObjectCreationTests.swift
dianaafanador3: Skipping only invalid check for mixed property given that it now accepts lists and dictionaries

RealmSwift/List.swift
dianaafanador3: Same as Map

Realm/RLMAccessor.mm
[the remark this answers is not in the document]
dianaafanador3: For any other mixed wrapped types besides list and dictionary the accessor is constructed without the current property because it is not needed to box the native value

Realm/RLMManagedDictionary.mm
```diff
@@ -102,6 +105,11 @@ - (RLMManagedDictionary *)initWithBackingCollection:(realm::object_store::Dictio
                                            property:(__unsafe_unretained RLMProperty *const)property {
     if (property.type == RLMPropertyTypeObject)
         self = [self initWithObjectClassName:property.objectClassName keyType:property.dictionaryKeyType];
+    else if (property.type == RLMPropertyTypeAny)
+        // Because the property is type mixed and we don't know if it will contain a dictionary when the schema
+        // is created, we set RLMPropertyTypeString by default.
+        // If another type is used for the dictionary key in a mixed dictionary context, this will thrown by core.
+        self = [self initWithObjectType:property.type optional:property.optional keyType:RLMPropertyTypeString];
```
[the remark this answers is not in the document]
dianaafanador3: When we create the schema we don't know what it is going to contain, so when we create the dictionary the property type will be of type Any, and property key type will be set to nil. There are two options here, or set the key type to string even though we don't know what the user's keys types are and let core throw any errors otherwise, or skip validation for key type if the property type is Any.

Realm/Tests/ObjectTests.m
dianaafanador3: we want to skip checking invalid value for anyCol, given that it accepts List and Dictionary

Realm/RLMManagedArray.mm
dianaafanador3: Are you sure that's the case, I was instructed by @jedelbo that this was the api for getting nested collections

Realm/RLMCollection.mm
```diff
@@ -58,11 +60,15 @@ @implementation RLMFastEnumerator {
...
                                collection:(id)collection
-                                classInfo:(RLMClassInfo&)info {
+                                classInfo:(RLMClassInfo&)info 
+                                parentInfo:(RLMClassInfo&)parentInfo
+                                 property:(RLMProperty *)property {
     self = [super init];
     if (self) {
         _info = &info;
         _realm = _info->realm;
+        _parentInfo = &parentInfo;
```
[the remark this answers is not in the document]
leemaguire: It's not more expensive, I just don't see the need to pass it as a reference if the caller already stores it as a pointer.

Realm/RLMQueryUtil.mm
dianaafanador3: Added comments and add it to the internal namespace

dianaafanador3:
@tgoyne and @leemaguire solved pr comments and there are some comments as well

Realm/RLMManagedDictionary.mm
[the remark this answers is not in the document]
dianaafanador3: We are allowing users to assign NSArrays and NSDictionary so they can initialise mixed values easier, which could have keys of different type.

Realm/RLMManagedDictionary.mm
dianaafanador3: For nested collections we want to be able to return the nested collection, which makes use of the get_dictionary and get_list API

elle-j (commented):
Nice to see the Swift implementation of this as well! 🙂

Realm/RLMManagedArray.mm
tgoyne: All of the tests pass with both of these just calling get() and there's code in get() which does exactly this.

Realm/RLMValue.h
```diff
@@ -52,11 +54,15 @@
...
+ `RLMDictionary`
+ `RLMArray`
+ `NSArray`
+ `NSDictionary`
  */
 @protocol RLMValue
 
 /// Describes the type of property stored.
-@property (readonly) RLMPropertyType rlm_valueType;
+@property (readonly) RLMMixedValueType rlm_valueType;
```
tgoyne, line 65: This is a breaking change. It needs to be a new property.

RealmSwift/Map.swift
tgoyne: ObjectIdentifier is not a valid definition either. Two managed dictionary objects which correspond to the same field are equal to each other but have different ObjectIdentifiers. All of our accessor types are fundamentally incompatible with Hashable due to that their identity changes when you promote an unmanaged object.

Realm/RLMValue.h
nirinchev: Since this is public and the wording we use is around "realm value" and not "mixed", that type should probably be `RLMValueType` or something similar - I don't think we should be throwing in "mixed" in a public API.

Realm/RLMUtil.mm
```diff
@@ -454,7 +476,15 @@ id RLMMixedToObjc(realm::Mixed const& mixed,
         case realm::type_UUID:
             return [[NSUUID alloc] initWithRealmUUID:mixed.get<realm::UUID>()];
         default:
-            @throw RLMException(@"Invalid data type for RLMPropertyTypeAny property.");
+            if (mixed.is_type(realm::type_Dictionary)) {
+                return [[RLMManagedDictionary alloc] initWithParent:obj property:property parentInfo:*classInfo];
+            }
+            else if (mixed.is_type(realm::type_List)) {
+                return [[RLMManagedArray alloc] initWithParent:obj property:property parentInfo:*classInfo];
+            }
```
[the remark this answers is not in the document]
dianaafanador3: `type_Dictionary` and `type_List` are not included in realm::DataType, in consequence this is showing a very annoying warning if used as a case, and that's the reason we are using a condition instead of adding a case here. @jedelbo any reason for this?

RealmSwift/Map.swift
dianaafanador3: @tgoyne we do need to conform to `Hashable` to be able to include Map or List as a an associated value in AnyRealmValue if we want to go that approach.
I can think of a few approaches for this:
* Use ObjectIdentifer and sacrifice equality for managed and unmanaged object
* Have associated values as Swift's native types Array/Dictionary, this will clearly will go bad with nested collections which will need to be instantiated completely to convert it (Discarded)
* Store some reference to the collection as the associated value and then and have some static function that returns the collection (Not need to conform to hashable, but worse developer experience).

I'm open to discuss different solutions

tgoyne: We can override the hash function on AnyRealmValue and return a constant value for the Map and List cases.

tgoyne:
There are a bunch of new warnings here that need to be fixed.

Realm/RLMQueryUtil.mm
```diff
@@ -1850,6 +1852,34 @@ bool is_self_value_for_key_path_function_expression(NSExpression *expression)
...
+            if (expression.arguments[i].expressionType == NSConstantValueExpressionType) {
+                id value = [expression.arguments[i] constantValue];
+                if ([value isKindOfClass:[NSNumber class]]) {
+                    NSNumber *index = (NSNumber *)value;
+                    if (index) {
+                        paths.push_back(PathElement{[index intValue]});
+                    }
+                } else if ([value isKindOfClass:[NSString class]]) {
+                    NSString *key = (NSString *)value;
+                    if ([key isEqual:@"all"]) {
```
tgoyne, line 1871: ???

If you have a field named "all" it just completely ignores the field name and does a wildcard search instead?

Realm/RLMManagedDictionary.mm
```diff
@@ -244,18 +250,27 @@ - (NSUInteger)countByEnumeratingWithState:(NSFastEnumerationState *)state
 - (nullable id)objectForKey:(id)key {
     return translateErrors([&]() -> id {
         [self.realm verifyThread];
-        RLMAccessorContext context(*_objectInfo);
+        RLMAccessorContext context(*_ownerInfo, *_objectInfo, _property);
```
tgoyne, line 253:
```
        RLMAccessorContext context(*_ownerInfo, *_objectInfo, _property);
        if (_property.type == RLMPropertyTypeAny) {
            return _backingCollection.get(context, context.unbox<realm::StringData>(key));
        }
        if (auto value = _backingCollection.try_get_any(context.unbox<realm::StringData>(key))) {
            return context.box(*value);
        }
        return nil;
```

RealmSwift/Object.swift
```diff
@@ -760,6 +760,8 @@ public final class DynamicObject: Object {
             return array.isOptional ? List<ObjectId?>(collection: array) : List<ObjectId>(collection: array)
         case .object:
             return List<DynamicObject>(collection: array)
+        default:
```
tgoyne, line 763: These defaults aren't needed.

Realm/RLMManagedArray.mm
```diff
@@ -198,9 +198,18 @@ - (NSUInteger)countByEnumeratingWithState:(NSFastEnumerationState *)state
...
+        RLMAccessorContext context(*_ownerInfo, *_objectInfo, _property);
+        if (value.is_type(realm::type_Dictionary)) {
+            return context.box(_backingList.get_dictionary(index));
+        }
+        else if (value.is_type(realm::type_List)) {
+            return context.box(_backingList.get_list(index));
+        }
+        else {
+            return context.box(value);
+        }
```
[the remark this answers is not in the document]
dianaafanador3: I don't think get is getting nested levels, this are done by using get_dictionary and get_list, which is going an extra level within the collection https://github.com/realm/realm-core/blob/970e558fbe1b3439a5c6d8fbdecfee35815d04ba/src/realm/list.cpp#L585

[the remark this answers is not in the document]
dianaafanador3: I don't know if I'm missing something

Realm/Tests/KVOTests.mm
dianaafanador3: @tgoyne I guess you are right, this function is overly optimistic about what to expect, this works every time the query is structured in a way that the subscripts are declared the expected way.  We do have to work with NSPredicate limitations and the only way I see we can declare subscripts in a field and not get an error from NSPredicate is this way (using subscripts)
If we for example use an NSPredicate with this structure `anyCol[0]['key']['all']`, NSPredicate will return a nested NSExpression where the first item is another NSExpression which nests as many NSExpressions we have as subscripts.
We are applying this paths to the query in case is a mixed value, we cannot know if the wrapped value is a collection or not from the property but only from the value at runtime, also even if we are querying against it, we cannot be sure the wrapped value is of that type. We are trusting core to do this verifications and not return any data if that's the case.

CHANGELOG.md
```diff
@@ -2,6 +2,35 @@ x.y.z Release notes (yyyy-MM-dd)
...
+
+  // You can build a AnyRealmValue from a Swift's Map
+  // and nest a collection within another collection.
+  let list: Array<AnyRealmValue> = [.int(12), .double(14.17), AnyRealmValue.fromDictionary(dictionary)]
+
+  let realm = realmWithTestPath()
+  try realm.write {
+    let obj = MixedObject()
+    obj.anyValue = AnyRealmValue.fromArray(list)
+    realm.add(o)
```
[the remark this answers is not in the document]
dianaafanador3: I removed the return

Realm/Tests/KVOTests.mm
dianaafanador3: Aso, fixed some hardcoded values in this function

Realm/RLMCollection.mm
```diff
@@ -146,10 +156,19 @@ - (NSUInteger)countByEnumeratingWithState:(NSFastEnumerationState *)state
...
+                _strongBuffer[batchCount] = _results->get(ctx, index);
+                batchCount++;
+            }
+        } else {
+            // This is used by Dicitonary and List for nested collections.
+            RLMAccessorContext ctx = RLMAccessorContext(*_parentInfo, *_info, _property);
+            for (NSUInteger index = state->state; index < count && batchCount < len; ++index) {
+                _strongBuffer[batchCount] = _results->get(ctx, index);
+                batchCount++;
+            }
```
tgoyne, line 171:
```suggestion
        auto ctx = _parentInfo ? RLMAccessorContext(*_parentInfo, *_info, _property) : 
        RLMAccessorContext(*_info);
        for (NSUInteger index = state->state; index < count && batchCount < len; ++index) {
            _strongBuffer[batchCount] = _results->get(ctx, index);
            batchCount++;
        }
```

Realm/RLMManagedDictionary.mm
```diff
@@ -454,9 +483,9 @@ - (BOOL)isFrozen {
 - (instancetype)resolveInRealm:(RLMRealm *)realm {
     auto& parentInfo = _ownerInfo->resolve(realm);
     return translateErrors([&] {
-        return [[self.class alloc] initWithBackingCollection:_backingCollection.freeze(realm->_realm)
+        return [[RLMManagedDictionary alloc] initWithBackingCollection:_backingCollection.freeze(realm->_realm)
```
tgoyne, line 486: Things which produce a copy of the current object use `self.class` rather than explicitly spelling out the class name  so that they work properly when called on subclasses. That's not directly relevant here, but it is the more generally correct pattern.

Realm/RLMQueryUtil.mm
```diff
@@ -1684,23 +1683,29 @@ bool is_self_value_for_key_path_function_expression(NSExpression *expression)
...
-    } else {
-        keyPath = [functionExpression.arguments.firstObject keyPath];
-        mapKey = [functionExpression.arguments[1] constantValue];
-    }
+    NSExpression *keyPathExpression = functionExpression;
 
+    // We consider only `NSKeyPathExpressionType` values to build the keypath.
+    for (unsigned long i = 0; i < pathElements.size(); i++) {
+        if (keyPathExpression.arguments[0].expressionType == NSKeyPathExpressionType) {
+            keyPath = [NSString stringWithFormat:@"%@", keyPathExpression.arguments[0]];
```
tgoyne, line 1695: This appears to be a very roundabout way of writing `keyPathExpression.arguments[0].description`. I think the actually correct thing here is `keyPathExpression.arguments[0].keyPath`

Realm/RLMQueryUtil.mm
tgoyne: Why is this using "all" as the wildcard search key?

Realm/RLMQueryUtil.mm
```diff
@@ -1850,6 +1855,31 @@ bool is_self_value_for_key_path_function_expression(NSExpression *expression)
...
+
+// This function returns the nested subscripts from a NSPredicate with the following format `anyCol[0]['key']['all']`
+// This will iterate each argument of the NSExpression and its nested NSExpression's and take the constant value
+// and create a PathElement for the query.
+void QueryBuilder::get_path_elements(std::vector<PathElement> &paths, NSExpression *expression) {
+    for (NSUInteger i = 0; i < expression.arguments.count; i++) {
+        if (expression.arguments[i].expressionType == NSFunctionExpressionType) {
+            get_path_elements(paths, expression.arguments[i]);
+        } else {
+            if (expression.arguments[i].expressionType == NSConstantValueExpressionType) {
```
tgoyne, line 1867: If this expression is any other type this just silently builds the wrong query? This function appears to be missing error checking.

Realm/RLMUtil.mm
```diff
@@ -454,7 +476,15 @@ id RLMMixedToObjc(realm::Mixed const& mixed,
         case realm::type_UUID:
             return [[NSUUID alloc] initWithRealmUUID:mixed.get<realm::UUID>()];
         default:
-            @throw RLMException(@"Invalid data type for RLMPropertyTypeAny property.");
+            if (mixed.is_type(realm::type_Dictionary)) {
```
tgoyne, line 479: Why are these ifs in the default case rather than switch cases?

Realm/RLMValue.mm
```diff
@@ -23,26 +23,42 @@
 
 @implementation NSData (RLMValue)
 
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Wdeprecated-declarations"
```
tgoyne, line 27: This is not the correct warning.

Realm/Tests/ObjectTests.m
```diff
@@ -785,7 +785,8 @@ - (void)testCreateInRealmValidationForDictionary {
...
 
         if ([keyToInvalidate isEqualToString:@"anyCol"]) {
-            obj = self;
+            // Mixed properties now can accept dictionary or list, so we are skipping testing invalid values for this.
```
tgoyne, line 788: This test is still applicable for Mixed properties and just needs to be supplying something which is not a valid value for them.

RealmSwift/Object.swift
```diff
@@ -789,6 +789,8 @@ public final class DynamicObject: Object {
             return set.isOptional ? MutableSet<ObjectId?>(collection: set) : MutableSet<ObjectId>(collection: set)
         case .object:
             return MutableSet<DynamicObject>(collection: set)
+        default:
+            fatalError()
```
tgoyne, line 793: These are not needed and will only lead to future bugs.

RealmSwift/ObjectiveCSupport+AnyRealmValue.swift
```diff
@@ -20,6 +20,21 @@
 import Foundation
 import Realm
 
+/// :nodoc:
+extension RLMDictionary: RLMValue {
+    /// :nodoc:
+    public var rlm_anyValueType: RLMAnyValueType { .dictionary }
```
tgoyne, line 26: This is part of the obj-c API and it doesn't make any sense to be redefining it here.

Realm/RLMValue.mm
```diff
@@ -60,65 +76,177 @@ - (RLMPropertyType)rlm_valueType {
...
+}
+
+@end
+
+@implementation RLMDictionary (RLMValue)
+
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Wdeprecated-declarations"
+- (RLMPropertyType)rlm_valueType {
+    REALM_UNREACHABLE();
```
tgoyne, line 214: This is not unreachable. `rlm_valueType` is part of the public API.

RealmSwift/Impl/ObjcBridgeable.swift
```diff
@@ -163,6 +163,27 @@ extension AnyRealmValue: BuiltInObjcBridgeable {
     }
 }
 
+
+extension RLMManagedDictionary: BuiltInObjcBridgeable {
```
tgoyne, line 167: These are not valid implementations of BuiltInObjcBridgeable, and declaring these types as BuiltInObjcBridgeable doesn't really make sense at all.

RealmSwift/Query.swift
```diff
@@ -289,6 +289,23 @@ extension Query where T == Bool {
...
+// MARK: Mixed
+
+extension Query where T == AnyRealmValue {
+    /// :nodoc:
+    public subscript(position: Int) -> Query<AnyRealmValue> {
+        .init(appendKeyPath("[\(position)]", options: [.isPath]))
+    }
+    /// :nodoc:
+    public subscript(key: String) -> Query<AnyRealmValue> {
+        .init(appendKeyPath("['\(key)']", options: [.isPath]))
```
tgoyne, line 301: This needs to handle `key` containing quotes.

Realm/RLMUtil.mm
dianaafanador3: `type_Dictionary` and `type_List` are not included in realm::DataType, in consequence this is showing a very annoying warning if used as a case, and that's the reason we are using a condition instead of adding a case here.

Realm/RLMValue.mm
dianaafanador3: @tgoyne This is a very particular case because we are deprecating the declaration but also changing the type to be used. You are right because this is part of the public, the user can ask for `rlm_valueType `, but because we don't want to add any new types to property type we are in some sort of limbo here.
Solutions
We can return RLMPropertyTypeAny like we do for Null, and this is actually true because this indeed a colection of mixed?

RealmSwift/Query.swift
dianaafanador3: Do you mean because of this 
```
Note
The expressions you write inside parentheses within an interpolated string can’t contain an unescaped backslash (\), a carriage return, or a line feed. However, they can contain other string literals.
```

https://docs.swift.org/swift-book/documentation/the-swift-programming-language/stringsandcharacters/

Realm/RLMQueryUtil.mm
```diff
@@ -1850,6 +1848,45 @@ bool is_self_value_for_key_path_function_expression(NSExpression *expression)
...
+        if (expression.arguments[i].expressionType == NSFunctionExpressionType) {
+            nestedKeyPath = get_path_elements(paths, expression.arguments[i]);
+        } else if (expression.arguments[i].expressionType == NSConstantValueExpressionType) {
+            id value = [expression.arguments[i] constantValue];
+            if ([value isKindOfClass:[NSNumber class]]) {
+                paths.push_back(PathElement{[(NSNumber *)value intValue]});
+            } else if ([value isKindOfClass:[NSString class]]) {
+                NSString *key = (NSString *)value;
+                if ([key isEqual:@"*"]) {
+                    paths.push_back(PathElement{});
```
tgoyne, line 1869: `paths.emplace_back()`

Realm/RLMQueryUtil.mm
```diff
@@ -1850,6 +1848,45 @@ bool is_self_value_for_key_path_function_expression(NSExpression *expression)
...
+                               @"Only `Strings` or index are allowed subscripts");
+            }
+        } else if (expression.arguments[i].expressionType == NSKeyPathExpressionType) {
+            nestedKeyPath = expression.arguments[i].keyPath;
+        } else {
+            throwException(@"Invalid expression type",
+                           @"Subscripts queries don't allow any other expression types");
+        }
+        if ([nestedKeyPath length] > 0) {
+            keyPath = ([keyPath length] > 0) ? [NSString stringWithFormat:@"%@.%@", keyPath, nestedKeyPath] : [NSString stringWithFormat:@"%@", nestedKeyPath];
```
tgoyne, line 1884: `nestedKeyPath` already is a string and `[NSString stringWithFormat:@"%@", nestedKeyPath]` is just a complicated no-op.

Realm/RLMQueryUtil.mm
```diff
@@ -1850,6 +1848,45 @@ bool is_self_value_for_key_path_function_expression(NSExpression *expression)
...
+            } else if ([value isKindOfClass:[NSString class]]) {
+                NSString *key = (NSString *)value;
+                if ([key isEqual:@"*"]) {
+                    paths.push_back(PathElement{});
+                } else {
+                    paths.push_back(PathElement{key.UTF8String});
+                }
+            } else {
+                throwException(@"Invalid subscript type",
+                               @"Only `Strings` or index are allowed subscripts");
```
tgoyne, line 1875: Each of these errors should include the expression which was a problem to help users figure out what they're doing wrong.

Realm/RLMQueryUtil.mm
```diff
@@ -1850,6 +1848,45 @@ bool is_self_value_for_key_path_function_expression(NSExpression *expression)
...
+        NSString *nestedKeyPath = @"";
+        if (expression.arguments[i].expressionType == NSFunctionExpressionType) {
+            nestedKeyPath = get_path_elements(paths, expression.arguments[i]);
+        } else if (expression.arguments[i].expressionType == NSConstantValueExpressionType) {
+            id value = [expression.arguments[i] constantValue];
+            if ([value isKindOfClass:[NSNumber class]]) {
+                paths.push_back(PathElement{[(NSNumber *)value intValue]});
+            } else if ([value isKindOfClass:[NSString class]]) {
+                NSString *key = (NSString *)value;
+                if ([key isEqual:@"*"]) {
```
tgoyne, line 1868: `'*'` is better, but it is still a legal key. I poked a bit at what NSPredicate will let us put here and the reasonable-seeming options are `anyCol['a'].@any['b']` and `anyCol['a'][#any]['b']` (note the lack of quotes). The latter seems better and also the parser is weird and doesn't allow `@` followed by any of the NSPredicate keywords, which happens to include "any" and "all".

This shows up in the parsed predicate as a NSKeyPathExpressionType with `[(id)expression.arguments[i] predicateFormat]` being `@"#any"` (but `keyPath` is `@"any").

RealmSwift/Query.swift
tgoyne: If you do `$0.anyCol["'"] == 0` to query on the dictionary key "'" this will produce invalid syntax.

RealmSwift/Query.swift
```diff
@@ -289,6 +289,23 @@ extension Query where T == Bool {
...
+    }
+    /// Query all indexes or keys in a mixed nested collecttion.
+    public var all: Query<AnyRealmValue> {
```
tgoyne, line 304: This function is `any`, not `all`. `all` would query for all of the values in the dictionary matching the predicate.

RealmSwift/RealmCollection.swift
```diff
@@ -84,9 +84,16 @@ public protocol _RealmMapValue {
     /// Advance to the next element and return it, or `nil` if no next element exists.
     public mutating func next() -> Element? {
         let next = generatorBase.next()
-        if let key = next as? Key,
-           let value = collection[key as AnyObject].map(dynamicBridgeCast) as? Value {
-            return (key: key, value: value)
+        if let key = next as? Key {
+            if let value = collection[key as AnyObject].map(dynamicBridgeCast) as? Value {
```
tgoyne, line 88: This is going to perform very poorly. Reading a List element is going to allocate and discard two extra List objects, along with do a very large number of dynamic casts which are pretty expensive in Swift.

RealmSwift/ObjectiveCSupport+AnyRealmValue.swift
```diff
@@ -50,6 +50,12 @@ public extension ObjectiveCSupport {
             return u as NSUUID
         case let .object(o):
             return o
+        case let .dictionary(d):
+            let rlmDictionary = Map<String, AnyRealmValue>._rlmFromObjc(d)?.rlmDictionary
```
tgoyne, line 54: Why is this wrapping the object in the Swift type and then immediately rereading the thing it just wrapped?

tgoyne (commented):
There are still several new warnings. I'm not sure why the implicit casts from id<RLMCollection> to RLMArray are producing warnings here but not on master, but the other warnings seem straightforward.

RealmSwift/Query.swift
dianaafanador3: `$0.anyCol["'"]` is not allowed for AnyRealmValue subscripts

dianaafanador3:
@tgoyne solved PR comments and removed the warnings

Realm/RLMSwiftValueStorage.mm
```diff
@@ -72,7 +72,7 @@ void attach(__unsafe_unretained RLMObjectBase *const obj, NSString *property) {
     : _realm(obj->_realm)
     , _object(obj->_realm->_realm, *obj->_info->objectSchema, obj->_row)
     , _columnName(prop.columnName.UTF8String)
-    , _ctx(*obj->_info)
+    , _ctx(obj, obj->_info->objectSchema->property_for_name(prop.columnName.UTF8String))
```
tgoyne, line 75:
```suggestion
    , _ctx(obj, &obj->_info->objectSchema->persisted_properties[prop.index])
```

Realm/RLMValue.mm
tgoyne: Returning RLMPropertyTypeAny seems like the best option here.

REALM_UNREACHABLE() marks a place that should be logically impossible to hit and if we get there it always indicates a bug in our code.

Realm/Tests/NotificationTests.m
```diff
@@ -1727,3 +1727,362 @@ - (void)testModifyUnobservedKeyPathArrayProperty {
...
 @end
+
+@interface MixedCollectionChangesetTests : RLMTestCase <ChangesetTestCase>
+@end
+
+@implementation MixedCollectionChangesetTests  {
+    MixedObject *_obj;
+}
+- (void)setUp {
+    RLMRealm *realm = [RLMRealm defaultRealm];
```
tgoyne, line 1738: This logic should be in `prepare` so that each of the `ExpectChange` calls has a clean state.

Realm/Tests/QueryTests.m
```diff
@@ -3805,11 +3805,23 @@ - (void)testDictionaryQueryKeySubscriptWithObjectCol {
 - (void)testDictionarySubscriptThrowsException {
     RLMRealm *realm = [self realm];
     RLMAssertThrowsWithReason(([realm objects:@"ArrayPropertyObject" where:@"array['invalid'] = NULL"]),
-                              @"Invalid keypath 'array[\"invalid\"]': only dictionaries support subscript predicates.");
+                              @"Invalid keypath 'array[\"invalid\"]': only dictionaries and mixed support subscript predicates.");
```
tgoyne, line 3808: "mixed" isn't a word that should show up anywhere in our public API, which includes error messages.

RealmSwift/AnyRealmValue.swift
```diff
@@ -156,4 +178,61 @@ public enum AnyRealmValue: Hashable {
...
+
+    /// Returns a `AnyRealmValue` storing a `Map`.
+    ///
+    /// - Parameter dictionary: A Swift's dictionary of `AnyRealmValue` values.
+    /// - Returns: Returns an `AnyRealmValue` storing a `Map`.
+    public static func fromDictionary(_ dictionary: Dictionary<String, AnyRealmValue>) -> AnyRealmValue {
+        let map = Map<String, AnyRealmValue>()
+        dictionary.forEach { (key, value) in
+            map.setValue(value, forKey: key)
+        }
```
tgoyne, line 190:
```suggestion
        map.merge(dictionary, uniquingKeysWith: { $1 })
```

`setValue(forKey:)` is the obj-c KVC function, so it involves a bunch of funky bridging compared to using the subscript operator.

RealmSwift/Object.swift
```diff
@@ -820,6 +820,8 @@ public final class DynamicObject: Object {
             return dictionary.isOptional ? Map<String, ObjectId?>(objc: dictionary) : Map<String, ObjectId>(objc: dictionary)
         case .object:
             return Map<String, DynamicObject?>(objc: dictionary)
+        default:
```
tgoyne, line 823: This doesn't need a default clause.

RealmSwift/Query.swift
tgoyne: I don't know what you mean. `any["'"] == %@` (or `any['\''] == %@`) is a perfectly valid query, which this currently fails to generate a valid string for.

RealmSwift/RealmCollection.swift
```diff
@@ -85,7 +85,7 @@ public protocol _RealmMapValue {
     public mutating func next() -> Element? {
         let next = generatorBase.next()
         if let key = next as? Key,
-           let value = collection[key as AnyObject].map(dynamicBridgeCast) as? Value {
+            let value = collection[key as AnyObject].map(Value._rlmFromObjc(_:)), let value {
```
tgoyne, line 88: Previous indentation was correct.

`RLMMapIterator` presumably needs to be changed as well (and tests added for that).

RealmSwift/Tests/MixedCollectionTest.swift
```diff
@@ -0,0 +1,990 @@
...
+            token.invalidate()
+        }
+
+        let o = AnyRealmTypeObject()
+        let realm = realmWithTestPath()
+        try realm.write {
+            realm.add(o)
+        }
+
+        assertObjectNotification(o, block: {
```
tgoyne, line 653: This is a place where it'd be usual to use trailing closure syntax (i.e. `assertObjectNotification(o) {`).

RealmSwift/RealmCollection.swift
tgoyne: This was not fixed.

Realm/RLMQueryUtil.mm
```diff
@@ -1850,6 +1848,46 @@ bool is_self_value_for_key_path_function_expression(NSExpression *expression)
...
+        NSString *nestedKeyPath = @"";
+        if (expression.arguments[i].expressionType == NSFunctionExpressionType) {
+            nestedKeyPath = get_path_elements(paths, expression.arguments[i]);
+        } else if (expression.arguments[i].expressionType == NSConstantValueExpressionType) {
+            id value = [expression.arguments[i] constantValue];
+            if ([value isKindOfClass:[NSNumber class]]) {
+                paths.push_back(PathElement{[(NSNumber *)value intValue]});
+            } else if ([value isKindOfClass:[NSString class]]) {
+                NSString *key = (NSString *)value;
+                if ([key isEqual:@"#any"]) {
```
tgoyne, line 1869: This is not the correct thing to be checking for. This is matching `['#any']` rather than `[#any]`. The whole point here is to be using something that *isn't* a string value which could be a legal dictionary key. The tests for this are also all incorrect. Creating this via substitution syntax requires using `%K`, not `%@`.

RealmSwift/Query.swift
```diff
@@ -1035,7 +1035,8 @@ private func buildPredicate(_ root: QueryNode, subqueryCount: Int = 0) -> (Strin
         case .mapAnySubscripts(let keyPath, let keys):
             build(keyPath)
             for key in keys {
-                formatStr.append("[%@]")
+                let arg = (key as? String == "#any") ? "[%K]" : "[%@]"
```
tgoyne, line 1038: This is incorrectly turning `.where { $0.anyCol["#any"] == ... }` into `.where { $0.anyCol.any == ... }`. The entire point of all this is that `.any` *doesn't* produce something that could be a string subscript.

RealmSwift/AnyRealmValue.swift
```diff
@@ -156,4 +178,59 @@ public enum AnyRealmValue: Hashable {
...
+        list.append(objectsIn: array)
+        return AnyRealmValue.list(list)
+    }
+
+    // MARK: - Hashable
+
+    public func hash(into hasher: inout Hasher) {
+        switch self {
+        case let .int(i):
+            hasher.combine(i.hashValue)
```
tgoyne, line 207: These should all be `hasher.combine(i)` without the `.hashValue`. This currently is hashing the values and then rehashing the hash.

tgoyne (approved):

PR #8549, opened 2024-04-15
RCOCOA-2334: Move initial subscription logic into Core

nirinchev:
Fixes https://github.com/realm/realm-swift/issues/8548

TODO:
- [x] Changelog
- [x] API docs


RealmSwift/Sync.swift
```diff
@@ -515,6 +515,36 @@ public typealias Provider = RLMIdentityProvider
...
 }
 
+/**
+ A configuration controlling how the initial subscriptions are populated when a Realm file is first opened.
+
+ - seeAlso ``Realm.subscriptions``
+ - seeAlso https://www.mongodb.com/docs/atlas/device-sdks/sdk/swift/sync/flexible-sync/ for more information
+   about managing subscriptions with flexible sync.
+ */
+@frozen public struct InitialSubscriptionsConfiguration: Sendable {
```
tgoyne, line 525: There doesn't appear to be any need for this wrapper as it's not exposing a different API from the obj-c type.

RealmSwift/Sync.swift
```diff
@@ -569,6 +599,13 @@ public typealias Provider = RLMIdentityProvider
         config.cancelAsyncOpenOnNonFatalErrors
     }
 
+    /**
+     A configuration that controls how initial subscriptions are populated when the Realm is opened.
+     */
+    public var initialSubscriptions: InitialSubscriptionsConfiguration? {
+        config.initialSubscriptions == nil ? nil : InitialSubscriptionsConfiguration(config: config.initialSubscriptions!)
```
tgoyne, line 606: `config.initialSubscriptions.map(InitialSubscriptionsConfiguration.init(config:))`

Realm/RLMInitialSubscriptionsConfiguration.h
```diff
@@ -0,0 +1,70 @@
...
+ */
+@property (nonatomic, readonly) BOOL rerunOnOpen;
+
+/**
+ Create a new initial subscriptions configuration.
+
+ @param callback Callback that will be invoked to update the subscriptions for this Realm file when it's first created or every time it's opened if `rerunOnOpen` is `true`.
+ @param rerunOnOpen A flag controlling whether to run the subscription callback every time the Realm is opened or only the first time.
+ */
+- (instancetype)initWithCallback:(RLMFlexibleSyncInitialSubscriptionsBlock)callback rerunOnOpen: (BOOL)rerunOnOpen;
```
tgoyne, line 58: No space after the colon in obj-c.

Realm/RLMInitialSubscriptionsConfiguration.h
```diff
@@ -0,0 +1,70 @@
...
+ */
+RLM_SWIFT_SENDABLE
+typedef void(^RLMFlexibleSyncInitialSubscriptionsBlock)(RLMSyncSubscriptionSet * _Nonnull subscriptions);
+
+/**
+ A configuration controlling how the initial subscriptions are populated when a Realm file is first opened.
+
+ @see `RLMSubscriptionSet`
+ */
+@interface RLMInitialSubscriptionsConfiguration : NSObject
```
tgoyne, line 37: This type looks like it should be Sendable.

RealmSwift/Sync.swift
nirinchev: Right. I originally planned for this to be constructible and settable on the `SyncConfiguration` class, but then abandoned it because it would require unfreezing the sync config.

Realm/RLMInitialSubscriptionsConfiguration.h
tgoyne: There's a bunch of spots with extra spaces between the colon and the argument and not just here.

(I think the historical reason behind this style is that it makes word-wrapping work much better and obj-c is prone to very long message sends that need wrapping).

tgoyne (approved):
Looks good other than the style nit.

Realm/RLMInitialSubscriptionsConfiguration.h
nirinchev: Oh, sorry, will go over those!

Issue #8008, opened 2022-10-27
Filters on String


Vanilil:
### Problem

'Invalid operator type', reason: 'Operator '>' not supported for type 'string''
I don't understand why operator like '>' or '<' are not supported for type 'string' since it works in realm studio.

let name = "James"
let query = try! Realm().objects(myQuery_Realm).filter("name < %@", name)

### How important is this improvement for you?

I'd like to see it, but have a workaround

PR #8557, opened 2024-04-23
RCOCOA-1874: Add support for lexicographical comparison for string queries

dianaafanador3:
Added support for string comparison queries, which allows building string 
   queries with the following operators (`>`, `>=`, `<`, `<=`). 
   ([#8008](https://github.com/realm/realm-swift/issues/8008)).


leemaguire (approved):
LGTM

tgoyne:
This is silently ignoring case-insensitive modifiers on comparison operators (e.g. `a >[c] b`).

You appear to have modified the generated test files directly rather than modify the template files.

tgoyne:
I don't think we should support case-insensitive comparisons, but we need to report an error rather than silently doing the wrong thing.

dianaafanador3:
@tgoyne solved comments

PR #8560, opened 2024-04-26
RCOCOA-2289: Update Base url API

dianaafanador3:
Added support for updating Atlas Device Sync's base url, in case the need to roam between servers (cloud and/or edge server). ([#8486](https://github.com/realm/realm-swift/issues/8486))

Notes: This is gonna be tested manually, for now there is just one simple test which just check the new url is set.


Realm/RLMApp_Private.h
```diff
@@ -31,6 +31,8 @@ typedef void(^RLMAppNotificationBlock)(RLMApp *);
 @end
 
 @interface RLMApp ()
+/// A custom base URL to request against. If not set or set to nil, the default base url for app services will be returned.
+@property (nonatomic, readonly) NSString *baseURL;
```
nirinchev, line 35: Should this have a `nonnull` annotation? I don't suppose we can ever return `nil` here.

dianaafanador3: When we add the header RLM_HEADER_AUDIT_BEGIN(nullability, sendability) which assumes that all pointers in the region are assumed to be '_Nonnull' by default, you have to specify otherwise

Realm/ObjectServerTests/SwiftObjectServerTests.swift
```diff
@@ -62,6 +62,45 @@ class SwiftObjectServerTests: SwiftSyncTestCase {
...
 
+    func testUpdateBaseUrl() {
+        let app = App(id: appId)
+        XCTAssertNotNil(app.baseURL)
+        XCTAssertEqual(app.baseURL, "http://localhost:9090")
+
+        let ex = expectation(description: "update base url")
+        app.updateBaseUrl(to: "http://localhost:8080", { result in
+            switch result {
+            case .success(let realm):
```
nirinchev, line 73: Should this have `let realm` here?

dianaafanador3:
Solved PR comments and changed the approach for the Swift API, we are now using `@_spi(Private)` to be able to have this as a private API.

CHANGELOG.md
```diff
@@ -9,6 +9,16 @@ x.y.z Release notes (yyyy-MM-dd)
   ([Core #7552](https://github.com/realm/realm-core/pull/7552)).
 * Improve perfomance of IN queries and chained OR equality queries for
   UUID/ObjectId types. ([.Net #3566](https://github.com/realm/realm-dotnet/issues/3566))
+  UUID/ObjectId types. ([.Net * #3566](https://github.com/realm/realm-dotnet/issues/3566))
```
tgoyne, line 12: Bad merge

CHANGELOG.md
```diff
@@ -9,6 +9,16 @@ x.y.z Release notes (yyyy-MM-dd)
...
+* Added support for updating Atlas Device Sync's base url, in case the need to roam between 
+  servers (cloud and/or edge server). This API is private and can only be imported using 
+  `@_spi(Private)`
```
tgoyne, line 15: SPI names are a global namespace, so this needs to be `RealmSwiftPrivate`. I'm also not sure why this is private, specifically? Is this *never* going to be a public part of the API for some reason? If it's just not GA then it should be something like `RealmSwiftUpcoming` or `RealmSwiftExperimental` or Unstable or something.

RealmSwift/App.swift
```diff
@@ -237,21 +237,51 @@ public extension App {
...
+     - note: Updating the base URL would trigger a client reset.
+     */
+    @preconcurrency
+    @available(macOS 10.15, iOS 13.0, tvOS 13.0, watchOS 6.0, *)
+    @_spi(Private) func updateBaseUrl(to url: String?, _ completion: @Sendable @escaping (Result<Void, Error>) -> Void) {
+        self.__updateBaseURL(url) { error in
+            if let error = error {
+                completion(.failure(error))
+            } else {
+                completion(.success(()))
```
tgoyne, line 252: There's a bunch of extra parentheses here.

RealmSwift/App.swift
```diff
@@ -237,21 +237,51 @@ public extension App {
...
-    /// @param credentials The credentials identifying the user.
-    /// @returns A publisher that eventually return `User` or `Error`.
+    /**
+    Updates the base url used by Atlas device sync, in case the need to roam between servers (cloud and/or edge server).
+     - parameter url: The new base url to connect to. Setting `nil` will reset the base url to the default url.
+     - note: Updating the base URL would trigger a client reset.
+     */
+    @preconcurrency
+    @available(macOS 10.15, iOS 13.0, tvOS 13.0, watchOS 6.0, *)
+    @_spi(Private) func updateBaseUrl(to url: String?, _ completion: @Sendable @escaping (Result<Void, Error>) -> Void) {
```
tgoyne, line 247: This is missing the basic callback version of the API.

CHANGELOG.md
dianaafanador3: Experimental is more like it

RealmSwift/App.swift
dianaafanador3: We are returning void in case of success

tgoyne: Oh, I thought Result had an extra initializer when Success = Void but I guess that got dropped after the initial proposal.

RealmSwift/App.swift
```diff
@@ -220,6 +220,20 @@ let credentials = Credentials.JWT(token: myToken)
...
+    /**
+    Updates the base url used by Atlas device sync, in case the need to roam between servers (cloud and/or edge server).
+     - parameter url: The new base url to connect to. Setting `nil` will reset the base url to the default url.
+     - parameter completion: A callback invoked after completion.
+     - note: Updating the base URL would trigger a client reset.
+     */
+    @preconcurrency
+    @available(macOS 10.15, iOS 13.0, tvOS 13.0, watchOS 6.0, *)
+    @_spi(RealmSwiftExperimental) func updateBaseUrl(to url: String?, _ completion: @Sendable @escaping (Error?) -> Void) {
+        self.__updateBaseURL(url, completion: { error in
```
tgoyne, line 232: Shouldn't need to wrap the completion block here since it's the same type.

tgoyne (approved):

nirinchev:
These failures seem to be all 502s, so I'll go ahead and merge this 😩

PR #8563, opened 2024-04-29
RCOCOA-2344 Drop support for Xcode 14

tgoyne:
Xcode 15.1 (Swift 5.9.2) is now the minimum supported version.


Configuration/Base.xcconfig
```diff
@@ -66,17 +66,12 @@ MACOSX_DEPLOYMENT_TARGET_1400 = 10.13;
 MACOSX_DEPLOYMENT_TARGET_1500 = 10.14;
 MACOSX_DEPLOYMENT_TARGET = $(MACOSX_DEPLOYMENT_TARGET_$(XCODE_VERSION_MAJOR));
 WATCHOS_DEPLOYMENT_TARGET_1400 = 4.0;
-WATCHOS_DEPLOYMENT_TARGET_1500 = 5.0;
+WATCHOS_DEPLOYMENT_TARGET_1500 = 4.0;
```
tgoyne, line 69: I double-checked exactly what the minimum versions are with Xcode 15 and for whatever reason it turned out that they bumped the minimum iOS and tvOS versions but not the minimum watchOS version.

I left all the logic for setting deployment targets based on Xcode major version in place even though it's currently not relevant because Xcode 16 is presumably going to bump the minimum supported versions and we'd just have to restore it in a month.

Configuration/Base.xcconfig
```diff
@@ -66,17 +66,12 @@ MACOSX_DEPLOYMENT_TARGET_1400 = 10.13;
...
 WATCHOS_DEPLOYMENT_TARGET = $(WATCHOS_DEPLOYMENT_TARGET_$(XCODE_VERSION_MAJOR));
 TVOS_DEPLOYMENT_TARGET_1400 = 11.0;
 TVOS_DEPLOYMENT_TARGET_1500 = 12.0;
 TVOS_DEPLOYMENT_TARGET = $(TVOS_DEPLOYMENT_TARGET_$(XCODE_VERSION_MAJOR));
 
-REALM_LD_CLASSIC_1400 = ;
-REALM_LD_CLASSIC_1500 = -Wl,-ld_classic;
```
tgoyne, line 76: This was specifically a workaround for a Xcode 15.0 bug that was fixed in 15.1 (and is why this is bumping to 15.1 minimum rather than 15.0).

tgoyne:
I think all of the CI failures here are networking issues; we're sometimes getting 502 errors from github when cloning repos.

nirinchev (approved):

PR #8572, opened 2024-05-03
Fix using initialSubscriptions with @AsyncOpen and @AutoOpen

tgoyne:
Moving the storage of initialSubscriptions from RealmConfiguration to SyncConfiguration means that AsyncOpen needs to explicitly copy them over to the newly created SyncConfiguration.

It turned out we had no tests for swiftui plus flexible sync, so I made the existing test suite run against both pbs and flx apps.

### DIFF
--- a/RealmSwift/SwiftUI.swift
+++ b/RealmSwift/SwiftUI.swift
@@ -1531,10 +1531,16 @@ private class ObservableAsyncOpenStorage: ObservableObject {
     }
 
     private func asyncOpenForUser(_ user: User) {
+        let initialSubscriptions = configuration?.syncConfiguration?.initialSubscriptions
+
         // Set the `syncConfiguration` depending if there is partition value (pbs) or not (flx).
         var config: Realm.Configuration
         if let partitionValue = partitionValue {
             config = user.configuration(partitionValue: partitionValue, cancelAsyncOpenOnNonFatalErrors: true)
+        } else if let initialSubscriptions {
+            config = user.flexibleSyncConfiguration(cancelAsyncOpenOnNonFatalErrors: true,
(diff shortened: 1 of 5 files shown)


RealmSwift/SwiftUI.swift
```diff
@@ -1531,10 +1531,16 @@ private class ObservableAsyncOpenStorage: ObservableObject {
...
+        let initialSubscriptions = configuration?.syncConfiguration?.initialSubscriptions
+
         // Set the `syncConfiguration` depending if there is partition value (pbs) or not (flx).
         var config: Realm.Configuration
         if let partitionValue = partitionValue {
             config = user.configuration(partitionValue: partitionValue, cancelAsyncOpenOnNonFatalErrors: true)
+        } else if let initialSubscriptions {
+            config = user.flexibleSyncConfiguration(cancelAsyncOpenOnNonFatalErrors: true,
+                                                    initialSubscriptions: ObjectiveCSupport.convert(block: initialSubscriptions.callback),
+                                                    rerunOnOpen: initialSubscriptions.rerunOnOpen)
```
nirinchev, line 1543: Do we have to do something similar for `clientResetMode` and the before/after callbacks? I haven't looked much into the SwiftUI code, but it appears like something that's not obviously handled in this method.

nirinchev (approved):
Looks good - it's nice to see the tests generalized to also cover FLX. Only one question, which is not a problem with the fix, but rather something that caught my eye during the review.

RealmSwift/SwiftUI.swift
tgoyne: Yeah, those are probably broken too. Will fix in a subsequent PR.

PR #8573, opened 2024-05-03
Fix -[RLMApp baseURL] always being nil

tgoyne:
It turns out that our sync tests have been silently failing to run on CI for quite a while. ~~Hopefully this is the only broken thing.~~ There's a bunch of failing things.

I had to rewrite the baseUrl tests because they were not even remotely close to working.


tgoyne:
The sync tests now pass for me locally†. Thankfully everything broken iutside of baseURL was bugs in the tests and not bugs in the library. The sync jobs are now failing on CI because it's no longer silently ignoring errors when installing BaaS. AFAICT, installing BaaS has been broken ever since they added another private dep that Xcode Cloud doesn't have permission to access in Sept 2023, and the reason the sync tests got so much more reliable when we migrated to Xcode Cloud is that they've never actually been running. The SPM tests were failing to actually invoke setup_baas.rb at all; I fixed this and then re-disabled it because we want to at least be running the non-sync tests with asan/tsan.

† Sometimes. Creating an app in BaaS fails ~5% of the time which means there's often at least one failure while running the full test suite.

ci_scripts/ci_post_clone.sh
```diff
@@ -16,10 +16,11 @@ install_dependencies() {
         brew install swiftlint
     elif [[ "$CI_WORKFLOW" == "cocoapods"* ]]; then
         install_ruby
-    elif [[ "$CI_WORKFLOW" == "objectserver"* ]] || [[ "$target" == "swiftpm"* ]]; then
+    elif [[ "$CI_WORKFLOW" == "sync"* ]]; then
+    # elif [[ "$CI_WORKFLOW" == "sync"* ]] || [[ "$CI_WORKFLOW" == "swiftpm"* ]]; then
```
dianaafanador3, line 20: Are we not running the sync tests in SPM?

dianaafanador3 (approved):
LGTM!, added some questions

ci_scripts/ci_post_clone.sh
tgoyne: We're not running the sync tests *anywhere*, and trying and failing to install baas now results in the non-sync spm tests not running.

Realm/Tests/SwiftUITestHostUITests/SwiftUITestHostUITests.swift
```diff
@@ -22,8 +22,10 @@ import RealmSwift
 @available(iOS 13.0, macOS 10.15, tvOS 13.0, watchOS 6.0, *)
 class SwiftUITests: XCTestCase {
     var realm: Realm!
+    @MainActor
```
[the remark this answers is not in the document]
tgoyne: All of the UI test stuff is `@MainActor` and produces warnings about cross-actor access if used from unconfined contexts.

PR #8597, opened 2024-05-20
Upgrade to core 14.9.0

dianaafanador3:
Upgrade to core 14.9.0


CHANGELOG.md
```diff
@@ -7,6 +7,16 @@ x.y.z Release notes (yyyy-MM-dd)
 * <How to hit and notice issue? what was the impact?> ([#????](https://github.com/realm/realm-swift/issues/????), since v?.?.?)
 * `@AutoOpen` and `@AsyncOpen` failed to use the `initialSubscriptions` set in
   the configuration passed to them ([PR #8572](https://github.com/realm/realm-swift/pull/8572), since v10.50.0).
+* Accessing `App.currentUser` from within a notification produced by `App.switchToUser()` 
+  (which includes notifications for a newly logged in user) would deadlock. 
+  ([Core #7670](https://github.com/realm/realm-core/issues/7670), since v14.6.0).
+* Inserting the same link to the same key in a dictionary more than once would incorrectly
+  create multiple backlinks to the object. This did not appear to cause any crashes later, but would 
+  have affecting explicit backlink count queries (eg: `$0.links.count`) and possibly notifications.
```
tgoyne, line 15: We either don't support this query at all or it has a different syntax.

CHANGELOG.md
```diff
@@ -7,6 +7,16 @@ x.y.z Release notes (yyyy-MM-dd)
...
 * `@AutoOpen` and `@AsyncOpen` failed to use the `initialSubscriptions` set in
   the configuration passed to them ([PR #8572](https://github.com/realm/realm-swift/pull/8572), since v10.50.0).
+* Accessing `App.currentUser` from within a notification produced by `App.switchToUser()` 
+  (which includes notifications for a newly logged in user) would deadlock. 
+  ([Core #7670](https://github.com/realm/realm-core/issues/7670), since v14.6.0).
+* Inserting the same link to the same key in a dictionary more than once would incorrectly
+  create multiple backlinks to the object. This did not appear to cause any crashes later, but would 
+  have affecting explicit backlink count queries (eg: `$0.links.count`) and possibly notifications. 
+  ([Core #7676](https://github.com/realm/realm-core/issues/7676) since v14.5.2).
+* Automatic client reset recovery would crash when recovering on an `AnyRealmValue` property if its type was
```
nirinchev, line 17: This is omitting some nuance, which actually makes this point irrelevant for the Swift SDK - the crash was happening when `AddInteger` is called on a Mixed/AnyRealmValue property, but Swift doesn't support counters, so users shouldn't have been affected by it.

nirinchev (approved):

tgoyne:
I don't think we can merge this separately from #8546 because this includes the protocol version bump which indicates that we're collections-in-mixed aware.

CHANGELOG.md
dianaafanador3: I think this is referring to queries on backlinks for example `$0.linkingObjects.count == 3` where 
```
class TestObject: Object {
    @Persisted var objectCol: TestObject?

    @Persisted(originProperty: "objectCol")
    var linkingObjects: LinkingObjects< TestObject >
}
```
And the return count will be wrong because it will create multiple backlinks instead of just one. 
Correct me if I'm wrong @tgoyne

PR #8608, opened 2024-05-27
RCOCOA-2273: Add support for Logger categories/ Improve current logger.

dianaafanador3:
* Added support for filtering logs by category. Users wil have more fine grained control over 
  the log level for each category as well.
  ```swift
  Logger.setLogLevel(.info, for: Category.Storage.transactions)
  ```

This has a proposal for the Categories enum which includes nested categories, for Objective-C we are using strings as it may be very complicated to have nested enums and it may be too mucho work for the number of users using Objective-C


Realm/RLMLogger.h
```diff
@@ -94,6 +114,22 @@ typedef void (^RLMLogFunction)(RLMLogLevel level, NSString *message);
...
+*/
+- (void)setLevel:(RLMLogLevel)level category:(NSString *)category NS_REFINED_FOR_SWIFT;
+
+/**
+ Gets the logger's associated level for the specified category.
+
+ @param category The log category which we need the level.
+ @returns The log level for the specified category
+*/
+- (RLMLogLevel)getLevelForCategory:(NSString *)category NS_REFINED_FOR_SWIFT;
```
tgoyne, line 131: `levelForCategory:`. Getters in obj-c and Swift are not prefixed with `get`.

Realm/RLMLogger.mm
```diff
@@ -140,11 +157,34 @@ - (void)logWithLevel:(RLMLogLevel)logLevel message:(NSString *)message, ... {
...
+- (void)logWithLevel:(RLMLogLevel)logLevel category:(NSString *)category message:(NSString *)message {
     auto level = levelForLogLevel(logLevel);
     if (_logger->would_log(level)) {
         _logger->log(level, "%1", message.UTF8String);
+        _logger->log(LogCategory::get_category(category.UTF8String), levelForLogLevel(logLevel), message.UTF8String);
+    }
+}
+
+- (void)setLevel:(RLMLogLevel)level category:(NSString *)category {
+    RLMLogger *defaultLogger = [RLMLogger defaultLogger];
```
tgoyne, line 169: Why is this modifying the default logger and not the logger it's called on?

Realm/RLMLogger.mm
```diff
@@ -140,11 +157,34 @@ - (void)logWithLevel:(RLMLogLevel)logLevel message:(NSString *)message, ... {
     }
 }
 
-- (void)logLevel:(RLMLogLevel)logLevel message:(NSString *)message {
+- (void)logWithLevel:(RLMLogLevel)logLevel category:(NSString *)category message:(NSString *)message {
     auto level = levelForLogLevel(logLevel);
     if (_logger->would_log(level)) {
         _logger->log(level, "%1", message.UTF8String);
+        _logger->log(LogCategory::get_category(category.UTF8String), levelForLogLevel(logLevel), message.UTF8String);
```
tgoyne, line 164: This is double-logging every message?

Realm/RLMLogger.mm
```diff
@@ -140,11 +157,34 @@ - (void)logWithLevel:(RLMLogLevel)logLevel message:(NSString *)message, ... {
...
+}
+
+- (RLMLogLevel)getLevelForCategory:(NSString *)category {
+    RLMLogger *defaultLogger = [RLMLogger defaultLogger];
+    return logLevelForLevel(defaultLogger->_logger->get_level_threshold(LogCategory::get_category(category.UTF8String)));
+}
+
+#pragma mark Testing
+
++ (NSArray<NSString *> *)getAllCategories {
```
tgoyne, line 180: `allCategories`

Realm/RLMSyncManager.mm
```diff
@@ -73,14 +74,14 @@ RLMSyncLogLevel logLevelForLevel(Level logLevel) {
 #pragma mark - Loggers
 
 struct CocoaSyncLogger : public realm::util::Logger {
-    void do_log(const realm::util::LogCategory&, Level, const std::string& message) override {
-        NSLog(@"Sync: %@", RLMStringDataToNSString(message));
+    void do_log(const realm::util::LogCategory& category, Level, const std::string& message) override {
+        NSLog(@"%@: %@", RLMStringDataToNSString(category.get_name()), RLMStringDataToNSString(message));
```
tgoyne, line 78:
```suggestion
        NSLog(@"%s: %s", category.get_name().c_str(), message.c_str());
```

RealmSwift/Logger.swift
```diff
@@ -0,0 +1,275 @@
...
+         print("Realm Log - \(level): \(message)")
+     })
+     logger.log(level: .info, message: "Info DB: Database opened succesfully")
+     ```
+
+     - parameter level: The log level for the message.
+     - parameter category: The log category for the message.
+     - parameter message: The message to log.
+     */
+    internal func log(level: LogLevel, category: LogCategory = Category.realm, message: String) {
```
tgoyne, line 58: Why is this here?

RealmSwift/Logger.swift
```diff
@@ -0,0 +1,275 @@
...
+    }
+
+    /**
+     Creates a logger with the associated log level, category and the logic function to define your own logging logic.
+
+     ```swift
+     let logger = Logger(level: .info, category: Category.All, logFunction: { level, category, message in
+         print("\(category.toString()) - \(level): \(message)")
+     })
+     logger.log(level: .info, category: Category.realm, message: "Info DB: Database opened succesfully")
```
tgoyne, line 69: This is not a use-case we want to support. There's no reason for users to be sending log messages through our mechanism for reporting logs to them.

RealmSwift/Logger.swift
```diff
@@ -0,0 +1,275 @@
...
+
+/// Defines a log category for the Realm `Logger`.
+public protocol LogCategory: Sendable {
+    /**
+     Returns the string represtation of the Log category.
+
+     - returns: A string representing the log category.
+     - SeeAlso: `LogCategory`
+     */
+    func toString() -> String
```
tgoyne, line 119: `toString()` is a very unidiomatic name for this. `var rawValue: String { get }` is the normal name.

tgoyne (commented):
I don't see any reason for the obj-c code to be using strings for category names. There's a fixed set of them, so it can just be a (non-nested) enum.

Realm/RLMLogger_Private copy.h
```diff
@@ -0,0 +1,51 @@
+////////////////////////////////////////////////////////////////////////////
```
nirinchev, line 1: This looks like a merge artifact.

Realm/RLMLogger_Private.h
```diff
@@ -30,7 +30,27 @@ RLM_HEADER_AUDIT_BEGIN(nullability)
...
+ @param message The message to log.
+ */
+- (void)logWithLevel:(RLMLogLevel)logLevel categoryName:(NSString *)categoryName message:(NSString *)message;
+
+#pragma mark Testing
+
+/**
+Gets all the categories from Core. This is to be used for testing purposes only.
+ */
++ (NSArray<NSString *> *)getAllCategories;
```
nirinchev, line 48:
```suggestion
+ (NSArray<NSString *> *)allCategories;
```

RealmSwift/Logger.swift
dianaafanador3: This just makes it easier for us internally to log  in the SDK side, we do have a category for this kind of logs Category.sdk in Swift

dianaafanador3: There is also a category for app logs Category.app, so we could expose this to the user in that matters, but I agree that we don't need to expose this to the users and add noise to our logging mechanism, but I do agree that we could have some logging in the sdk side, we should discuss how granular we want this because we don't want to contaminate Core's log

Realm/RLMLogger.mm
```diff
@@ -140,13 +208,44 @@ - (void)logWithLevel:(RLMLogLevel)logLevel message:(NSString *)message, ... {
     }
 }
 
-- (void)logLevel:(RLMLogLevel)logLevel message:(NSString *)message {
+- (void)logWithLevel:(RLMLogLevel)logLevel category:(RLMLogCategory)category message:(NSString *)message {
     auto level = levelForLogLevel(logLevel);
     if (_logger->would_log(level)) {
```
clementetb, line 213: It is not filtering by category, just by level. There is an alternative signature for `would_log` that takes the category as a parameter.

Realm/RLMLogger.mm
```diff
@@ -120,10 +172,26 @@ - (instancetype)initWithLogger:(std::shared_ptr<Logger>)logger {
...
+            logFunction(level, message);
+        };
+        logger->function = block;
+        self->_logger = logger;
+    }
+    return self;
+}
+
+- (instancetype)initWithLevel:(RLMLogLevel)level
+                     category:(RLMLogCategory)category
```
clementetb, line 190: The loggers created with this function are of the default category `realm`. If you like to filter by category you might need to instantiate a Logger with the category.

clementetb (commented):
The current logger implementation does not filter per category as per the documentation.

Realm/RLMLogger.mm
dianaafanador3: Changed the approach, and now you can only set the global log level for a category using setLogLevel. initWithLevel is gonna is deprecated

clementetb (approved):
LGTM

RealmSwift/Logger.swift
```diff
@@ -0,0 +1,346 @@
...
+ You can define your own custom logger creating an instance of `Logger` and defining the log function which will be
+ invoked whenever there is a log message.
+
+ ```swift
+ let logger = Logger(function: { level, category, message in
+    print("Realm Log - \(category.rawValue)-\(level): \(message)")
+ })
+ ```
+
+ Set this custom logger as you default logger using `Logger.shared`. This will replace the default logger.
```
nirinchev, line 44:
```suggestion
You can set this custom logger as the default logger using `Logger.shared`.
```

RealmSwift/Logger.swift
```diff
@@ -0,0 +1,346 @@
...
+         print("Realm Log - \(level): \(message)")
+     })
+     logger.log(level: .info, message: "Info DB: Database opened succesfully")
+     ```
+
+     - parameter level: The log level for the message.
+     - parameter category: The log category for the message.
+     - parameter message: The message to log.
+     */
+    internal func log(level: LogLevel, category: LogCategory = Category.realm, message: String) {
```
nirinchev, line 70: Should the default category here be `Category.sdk` since that method should supposedly only be used to log things from the SDK?

RealmSwift/Logger.swift
```diff
@@ -0,0 +1,346 @@
...
+
+         Category hierarchy:
+         ```
+          Storage
+          ├─► Transaction
+          ├─► Query
+          ├─► Object
+          └─► Notification
+         ```
+         */
```
nirinchev, line 249: This doc string is wrong

RealmSwift/Tests/RealmTests.swift
```diff
@@ -1953,66 +1954,239 @@ class LoggerTests: TestCase, @unchecked Sendable {
...
+
+        logger.log(with: .info, categoryName: Category.Storage.transaction.rawValue, message: "Transaction test entry")
+        XCTAssertTrue(logs.value.contains("Transaction test entry"))
+        logs.wrappedValue = ""
+
+        logger.log(with: .info, categoryName: Category.realm.rawValue, message: "REALM test entry")
+        XCTAssertFalse(logs.value.contains("REALM test entry"))
+    }
+
+    func categoryfromString(_ string: String) -> LogCategory? {
```
nirinchev, line 2179: This is copied from `Category.fromString` - is there no way to make that method available to tests?

RealmSwift/Tests/RealmTests.swift
```diff
@@ -1953,66 +1954,239 @@ class LoggerTests: TestCase, @unchecked Sendable {
...
+
+        logger.log(with: .info, categoryName: Category.Storage.all.rawValue, message: "Storage test entry")
+        XCTAssertTrue(logs.value.contains("Storage test entry"))
+        logs.wrappedValue = ""
+
+        logger.log(with: .info, categoryName: Category.Storage.transaction.rawValue, message: "Transaction test entry")
+        XCTAssertTrue(logs.value.contains("Transaction test entry"))
+        logs.wrappedValue = ""
+
+        logger.log(with: .info, categoryName: Category.realm.rawValue, message: "REALM test entry")
```
nirinchev, line 2175: Why is this passing the raw category name rather than the `LogCategory`?

nirinchev (commented):
Looks good - I only have minor comments/suggestions.

RealmSwift/Tests/RealmTests.swift
dianaafanador3: Because we cannot expose internal function to the test suite, instead we are using Realm.Private API to generate the logs for the tests

RealmSwift/Tests/RealmTests.swift
dianaafanador3: This is no longer used internally, so I removed it there

Realm/RLMLogger.h
```diff
@@ -48,33 +48,95 @@ typedef RLM_CLOSED_ENUM(NSUInteger, RLMLogLevel) {
...
+ │   │   ├─► Network
+ │   │   └─► Reset
+ │   └─► Server
+ ├─► App
+ └─► Sdk
+```
+*/
+typedef NS_ENUM(NSUInteger, RLMLogCategory) {
+    ///  Top level log category for Realm, updating this category level would set all other subcategories too.
+    RLMLogCategoryRealm,
```
nirinchev, line 75: [nit] This naming scheme is causing a bit of tautology as we have the `RLM` prefix and the `Realm` suffix/infix. One option would be to have

```
RLMLogCategoryAll,
RLMLogCategorySDK,
...
```

Essentially dropping the `Realm` as it's part of every category, thus not bringing any new information. Not a blocker though.

nirinchev (approved):

Realm/RLMLogger.h
dianaafanador3: Should we change RLMLogCategoryRealm as well?, given that this is the keyword we use for all categories in all SDKs?

nirinchev: Yeah, I feel a bit ambiguous about it - I'd be fine with either keeping it as `RLMLogCategoryRealm` or with replacing it with `RLMLogCategoryAll`. I think in other SDKs we can get away with it because we don't have the `RLM` prefix (similarly to what we do in Swift). That being said, perhaps we shouldn't overthink it as Obj C users are so few and whether we have `Realm` in there or not is unlikely to make a huge difference.

dianaafanador3: In Swift we have Category.realm so to keep uniformity in both APIs, we can keep it like that, and then remove Realm for the rest of the categories

dianaafanador3: Added changes

PR #8618, opened 2024-06-11
RCOCOA-2393 Add support for Xcode 16

tgoyne:
Addresses some of https://github.com/realm/realm-swift/issues/8628.


tgoyne:
Needs a proper core release before this can be merged but I think the initial set of code changes are done and can be reviewed.

bdkjones:
Is there a rough ETA on this release? Xcode 16 has many things I'd like to try, but I can't use it until Realm is buildable there.

nirinchev:
You can build the framework from source and work against that until we've done an official release.

RealmSwift/Realm.swift
```diff
@@ -1356,6 +1356,35 @@ extension Realm {
         return Unchecked(ret)
     }
 
+    /**
```
dianaafanador3, line 1359: wouldn't this be for swift less than 6, I'm a little bit confused because of the commit message

dependencies.list
```diff
@@ -1,3 +1,3 @@
 VERSION=10.51.0
-REALM_CORE_VERSION=v14.9.0
+REALM_CORE_VERSION=v14.10.0-8-g8d7134760
```
dianaafanador3, line 2: Core 14.10.0 was released, we probably want to point to that

dianaafanador3 (approved):
LGTM!, Just some minor comments

dependencies.list
tgoyne: `v14.10.0-8-g8d7134760` is newer than v14.10.0. git describe's format is `${tagName}-${numberOfCommitsSinceTag}-`${commitHash}`, so this is referring to commit g8d7134760, which is 8 commits after tag v14.10.0.

RealmSwift/Realm.swift
tgoyne: This is moving the old version of the function into the `#if compiler(<6)` block.

PR #8637, opened 2024-06-20
RCOCOA-2395: Code sign our published xcframeworks

dianaafanador3:
Code sign our published xcframeworks in Github actions.
You can verify the generated release package in this action run.
https://github.com/realm/realm-swift/actions/runs/9600027436

### DIFF
--- a/.github/workflows/master-push.yml
+++ b/.github/workflows/master-push.yml
@@ -185,12 +185,22 @@ jobs:
         - uses: actions/checkout@v4
         - name: Select Xcode Version
           run: sudo xcode-select -switch /Applications/Xcode_${{ env.RELEASE_VERSION }}.app
+        - name: Install the Apple certificate and provisioning profile
+          env:
+            DEVELOPMENT_CERTIFICATE_BASE64: ${{ secrets.DEVELOPMENT_CERTIFICATE_BASE64 }}
+            P12_PASSWORD: ${{ secrets.P12_PASSWORD }}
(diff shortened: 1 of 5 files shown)


nirinchev:
It looks like this is signed with your personal team certificate (team 74HX4QV25A rather than QX5CR2FTN2). We should probably update the certificates.

.github/workflows/master-push.yml
```diff
@@ -185,12 +185,37 @@ jobs:
...
+
+            # create temporary keychain
+            security create-keychain -p "$KEYCHAIN_PASSWORD" $KEYCHAIN_PATH
+            security set-keychain-settings -lut 21600 $KEYCHAIN_PATH
+            security unlock-keychain -p "$KEYCHAIN_PASSWORD" $KEYCHAIN_PATH
+
+            # import certificate to keychain
+            security import $CERTIFICATE_PATH -P "$P12_PASSWORD" -A -t cert -f pkcs12 -k $KEYCHAIN_PATH
+            security set-key-partition-list -S apple-tool:,apple: -k "$KEYCHAIN_PASSWORD" $KEYCHAIN_PATH
+            security list-keychain -d user -s $KEYCHAIN_PATH
```
tgoyne, line 209: This should all be a shell script which is invoked. The only thing which should ever be in the GHA yaml files are just the GHA configuration, and not the actual build steps.

tgoyne (commented):
Where does the signing cert come from and how do we update it? If this is a manual process it needs to be documented.

tgoyne (approved):

Naveen-C-Ramachandrappa (approved):

PR #8643, opened 2024-07-02
RCOCOA-1883: Add "Sync is not enabled for this App" as a catchable error in error handler

dianaafanador3:
Report any sync errors tagged with warning action from the server via the sync error handler.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,8 +1,8 @@
 x.y.z Release notes (yyyy-MM-dd)
 =============================================================
 ### Enhancements
-* Added support for filtering logs by category. Users wil have more fine grained control over 
-  the log level for each category as well.
+* Added support for setting the log level for a category in `Logger`. User will have more fine grained control over 
+  the log level for each category.
   ```swift
   Logger.setLogLevel(.info, category: Category.Storage.transactions)
   ```
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ x.y.z Release notes (yyyy-MM-dd)
   binaries so Xcode can validate it was signed by the same developer on every new version. 
   ([Apple](https://developer.apple.com/support/third-party-SDK-requirements/)).
 * Throw any sync errors tagged with warning action from the server via the sync error handler. 
+* Report sync warnings from the server such as sync being disabled server-side to the sync error handler.
   ([#8020](https://github.com/realm/realm-swift/issues/8020)).
 
 ### Fixed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,15 +1,14 @@
 x.y.z Release notes (yyyy-MM-dd)
 =============================================================
 ### Enhancements
-* Added support for setting the log level for a category in `Logger`. User will have more fine grained control over 
-  the log level for each category.
+* Added support for filtering logs by category. Users wil have more fine grained control over 
+  the log level for each category as well.
   ```swift
   Logger.setLogLevel(.info, category: Category.Storage.transactions)
   ```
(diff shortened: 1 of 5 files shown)


Realm/ObjectServerTests/SwiftFlexibleSyncServerTests.swift
```diff
@@ -1186,6 +1186,30 @@ class SwiftFlexibleSyncTests: SwiftSyncTestCase {
         XCTAssertEqual(downloadCount.value, 0)
         XCTAssertEqual(uploadCount.value, 0)
     }
+
+    func testFlexibleSyncNotEnabledError() throws {
+        let appServerId = try RealmServer.shared.retrieveAppServerId(appId)
```
tgoyne, line 1191: Creating a new app with sync disabled is probably faster than toggling sync off and on and eliminates the chance of something going wrong in this test causing problems for other tests. This could be done by adding a new value to `SyncMode` and exiting createApp() after setting up the auth providers.

Realm/ObjectServerTests/SwiftFlexibleSyncServerTests.swift
```diff
@@ -1186,6 +1186,30 @@ class SwiftFlexibleSyncTests: SwiftSyncTestCase {
...
+        let syncServerId = try RealmServer.shared.retrieveSyncServiceId(appServerId: appServerId)
+        _ = try RealmServer.shared.disableSync(appServerId: appServerId, syncServiceId: syncServerId)
+        var isEnable = try RealmServer.shared.isSyncEnabled(appServerId: appServerId, syncServiceId: syncServerId)
+        XCTAssertFalse(isEnable)
+
+        let ex = expectation(description: "Waiting for error handler to be called...")
+        ex.assertForOverFulfill = false // error handler can legally be called multiple times
+        app.syncManager.errorHandler = { @Sendable (error, _) in
+            assertSyncError(error, .clientInternalError, "Sync is not enabled for this app")
```
tgoyne, line 1200: This isn't an internal error. None of the existing error codes are a great match for this, so we should probably add a new one. Maybe `RLMSyncErrorServerWarning`?

CHANGELOG.md
```diff
@@ -1,11 +1,13 @@
...
 =============================================================
 ### Enhancements
-* Added support for filtering logs by category. Users wil have more fine grained control over 
-  the log level for each category as well.
+* Added support for setting the log level for a category in `Logger`. User will have more fine grained control over 
+  the log level for each category.
   ```swift
   Logger.setLogLevel(.info, category: Category.Storage.transactions)
   ```
+* Throw any sync errors tagged with warning action from the server via the sync error handler.
```
tgoyne, line 9: There's no throwing involved here.

```suggestion
* Report sync warnings from the server such as sync being disabled server-side to the sync error handler. 
```

kneth (approved):
I don't have anything meaningful to contribute but the tests indicate that the error is reported to the error handler

dianaafanador3:
Solved PR comments @tgoyne

Realm/RLMError.h
```diff
@@ -328,6 +328,11 @@ typedef RLM_ERROR_ENUM(NSInteger, RLMSyncError, RLMSyncErrorDomain) {
      Connecting to the server failed due to a TLS issue such as an invalid certificate.
      */
     RLMSyncErrorTLSHandshakeFailed = 13,
+    /**
+     An error indicating the server send down an error with a warning action, which
+     needs to be executed by the client or the end user.
```
tgoyne, line 333: Warnings don't have an action which needs to be executed by the client or end user.

dianaafanador3: @tgoyne Reworded, let me know if this fits

tgoyne (approved):

PR #8673, opened 2024-08-27
RCOCOA-2394 RCOCOA-2396 RCOCOA-2397 Build in Swift 6 language mode when supported

tgoyne:
This drops support for Xcode 15.1 and 15.2 as supporting both Swift 5.9 and 6 at the same time turned out to be an utter mess.

SPM and the Xcode project build in Swift 6 mode when using Xcode 16. CocoaPods always builds in Swift 5 mode since there doesn't appear to be any functional benefit to make it conditional, and it's a bit complicated to do.

Most of the library changes are just replacing `@_unsafeInheritExecutor` with `#isolation`, which is the proper official way to achieve the same effect. Sadly it was not added until Swift 6, so a lot of `#if` is involved. Some of the mongo client functions were incorrectly returning `[AnyHashable: String]` rather than `Document`, which now doesn't compile at all so I fixed the return type in Swift 6 mode. Isolated init never got implemented, so `try await Realm(actor: actor)` has been replaced with `try await Realm.open()`, since doing the same thing in a static function does work.

Swift used to have a thing where an isolated property wrapper would make the containing type implicitly isolated. This was very confusing and weird so it got removed and the SwiftUI types now need to be explicitly marked as `@MainActor`.

The changes to tests are mostly boring mechanic changes. Making XCTest subclasses `@MainActor` didn't actually work, and instead individual test functions need to be marked with it. `setUp` *can't* be isolated, so a bunch of test init and deinit had to be adjusted to become thread-safe. `waitForExpectations()` is main actor isolated while `wait(for:)` isn't (as it doesn't have the magic global list of expectations), so a bunch of tests were switched to that. `@ThreadSafe` just plain doesn't work in swift 6 mode due to sendability checking on property wrappers never being implemented, so those tests are disabled.

For some reason the compiler is convinced that `QuerySubscription`'s callback function needs to be Sendable, so a bunch of tests had to be updated to not capture `self` in that callback.

### DIFF
--- a/.github/workflows/build-binaries.yml
+++ b/.github/workflows/build-binaries.yml
@@ -16,13 +16,8 @@ jobs:
       core-version: ${{ steps.get-core-version.outputs.version }}
     strategy:
       matrix:
-        target: [macosx, iphoneos, iphonesimulator, appletvos, appletvsimulator, watchos, watchsimulator, maccatalyst]
-        xcode: ["15.1"]
-        include:
-          - target: xros
-            xcode: "15.2.0"
-          - target: xrsimulator
-            xcode: "15.2.0"
+        target: [macosx, iphoneos, iphonesimulator, appletvos, appletvsimulator, watchos, watchsimulator, maccatalyst, xros, xrsimulator]
+        xcode: ["15.3"]
     steps:
       - uses: actions/checkout@v4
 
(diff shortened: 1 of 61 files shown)


leemaguire (approved):
LGTM!

PR #8713, opened 2024-11-17
Update GHA workflows to work with the macOS image changes

tgoyne:


tgoyne:
https://github.com/realm/realm-swift/actions/runs/12510269430 successful packaging run.

PR #8722, opened 2024-11-26
Fix a crash that may occur when resolving a thread-safe reference

PoltoraIvana:
In some cases, as per documentation, resolveThreadSafeReference can return nil. It usually happens in highly-concurrent contexts. Checking for nil fixes this crash

### DIFF
--- a/Realm/RLMError.h
+++ b/Realm/RLMError.h
@@ -199,6 +199,10 @@ typedef RLM_ERROR_ENUM(NSInteger, RLMError, RLMErrorDomain) {
      subscription or the server will reject the write.
      */
     RLMErrorNoSubscriptionForWrite = 22,
+     / **
+    Denotes an error that occurs when resolving thread safe reference returns a nil object.
+     */
+    RLMSyncErrorPermissionDeniedError = 23
 };
 
 #pragma mark - RLMSyncError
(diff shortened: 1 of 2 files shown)


aehlke:
Who will review and merge this now that Realm is dead?

PoltoraIvana:
I thought there are some maintainers that are left?
Overall, we had this problem in our codebase, it was fixed like that and the problem was fixed with it as well

aehlke:
@tgoyne hi - can we get some community members authorized to merge PRs? Thank you

tgoyne:
Reporting an error here is incorrect. The TSR resolving to nil means that the object was deleted in between `observe()` being called and the notifier being set up, and we can just report that to the block.

aehlke:
I believe you can mitigate this crash by putting all writes on the same global actor

PoltoraIvana:
>I believe you can mitigate this crash by putting all writes on the same global actor, per realm file
that's an option, yes, but for us, we still have realm observations on queue, so we would need to migrate to take advantage of actors.

aehlke:
Yes, I migrated all my Realm usage to actors to resolve crashing instability issues. I recommend it.

Naveen-C-Ramachandrappa (approved):

PR #8725, opened 2024-12-03
- Fix `Datetime` rounding in `RLMConvertBsonToRLMBSON`

anton-plebanovich:
I noticed that the `Date` types returned from Sync Functions calls are rounded to seconds. This PR fixes the issue. Please note that I am not a C++ programmer so this probably could be done in a better way. I will gladly accept any fix requests with hints.


Naveen-C-Ramachandrappa (approved):

aehlke:
@anton-plebanovich are you able to merge or do we need to summon mongodb staff..?

aehlke:
Realm is dead and no one is assigned to review PRs. Probably we need to organize around a community fork

anton-plebanovich:
We are moving away from Realm Sync and are using Functions as an intermediate step. Thus, this change was needed to receive valid responses. I have created a fork for myself but this is only temporarily needed until we finish our transition so I don't want to spend more effort than needed here.

PR #8770, opened 2025-06-10
Add support for building with Xcode 26

tgoyne:


ugiacoman:
> Prebuilt binaries depend on GHA deploying the Xcode beta (see [actions/runner-images#12355](https://github.com/actions/runner-images/issues/12355)). If that hasn't happened in the next few days I'll tag a release without it.

Sounds good, thank you!

PR #8773, opened 2025-06-11
Add support for Xcode 16.4.

Stannieman:
I'm not sure if this is enough or if the runners are available. CI shall decide if it works I guess.
Looking at the PR for Xcode 16.3 it at least looks close to what's needed.
Do 16.2 builds need to be removed?


tgoyne:
Cocoapods errors are because those jobs don't work for forks.

PR #8783, opened 2025-06-24
Fix UIDevice support enabled for WatchOS 26 on Xcode 26

eriklamanna:
When building a WatchOS target with Xcode 26, the build fails due to a missing UIDevice call on the Watch SDK. 

It appears that in the initial WatchOS 26 betas that the check for `<UIKit/UIDevice.h>` in `RLMApp.mm` now passes for Watch targets when it shouldn't. 

This change adds a separate check for `TARGET_OS_WATCH` not being defined before defining the `REALM_UIDEVICE_AVAILABLE` flag.

### DIFF
--- a/Realm/RLMApp.mm
+++ b/Realm/RLMApp.mm
@@ -21,7 +21,9 @@
 #import <sys/utsname.h>
 #if __has_include(<UIKit/UIDevice.h>)
 #import <UIKit/UIDevice.h>
+#ifndef TARGET_OS_WATCH
 #define REALM_UIDEVICE_AVAILABLE
+#endif
 #endif
 
 #import "RLMAnalytics.hpp"


aehlke:
May I ask what your plans are for continued production use of Realm? I have several apps that still use it due to its local capabilities (I never used sync as I have my own iCloud & server sync layers for it) and am worried about the lack of non-Mongo maintainers. Are you transitioning off or prepared to help maintain it (within your means)?

There are still no satisfying alternatives available but I am worried about some iOS update that requires extensive change in this library. Thanks...

tgoyne:
To the best of my memory, iOS 15's file lock changes were the only time a new iOS version required significant changes on our end. We often had to update tests for new iOS version but it was always things like the test hardcoding a specific key order in a dictionary and the new version changing that. As such, my general expectation is that Realm will probably keep working in roughly its current state for as long as Swift 5† is supported by Xcode. I intend to do the bare minimum to keep things compiling with new Xcode versions and such, but that is something I'm doing in my free time.

We found it difficult to hire people to work on Realm (Swift plus obj-c++ plus lots of c++ is not a very common skillset) so barring one of the large companies that uses it deciding it's cheaper to pay someone to fix things than to migrate their apps off it I am not hopeful for substantial external contributions.

† One of my last projects on Realm was adding support for building as Swift 6, but Swift 6 continues to be a moving target and building in Swift 6 mode didn't work with Xcode 26.

aehlke:
@tgoyne thank you for the context & warning about Swift 6. I haven't migrated to Swift 6 yet so I wasn't aware it doesn't work. I will definitely find a way off Realm sooner than later in that case...

tgoyne:
Migrating your app to Swift 6 *should* be fine since you can mix and match swift versions.

PR #8829, opened 2026-06-10
Fix Xcode 27 Beta build issues

### LINKED ISSUE #8827: Build failure on XCode 27

Always

### Description

When building with Xcode 27 Beta 1 the build always fails

### Stacktrace & log output

```shell
/Applications/Xcode-beta.app/Contents/Developer/Platforms/iPhoneSimulator.platform/Developer/SDKs/iPhoneSimulator27.0.sdk/usr/include/c++/v1/__utility/lazy_synth_three_way_comparator.h:52:21: No matching function for call to object of type 'const std::less<void>'
```

Always

### Reproduction Steps

Build with Xcode 27 Beta 1

### Version

Xcode 27 Beta 1

### What Atlas Services are you using?

Local Database only

### Are you using encryption?

No

iOS 27

### Build environment

Xcode version: 27 Beta 1
Dependency manager and version: ...

TomasLinhart:
@State is no longer a property wrapper in Xcode 27. It's a macro that generates code that does not work with `where`. This replaces the name with one that is not a reserved keyword of the Swift compiler.

It also includes changes to `Base.xcconfig` and `TestBase.xcconfig` for Xcode 27. These values differ from previous ones because they are the minimum versions Xcode 27 supports.

This fixes https://github.com/realm/realm-swift/issues/8827 and https://github.com/realm/realm-swift/issues/8828


rjchatfield:
👍

kylebrowning (approved):

kylebrowning:
@tgoyne can we get this merged so our projects can compile?

tgoyne:
I should have time to do all the stuff to get core and swift releases out this weekend.

tgoyne:
I cherry-picked the xcconfig changes, but did something different for the `where` property that I hope will avoid being a breaking change.

samuelowino:
What is the fate of xcode 27 support, since this was not merged @tgoyne ?

